PR #9, opened 2022-09-19
add production push and update staging

cpanato:
- add production push and update staging

~~need to deploy the function in prod first, will stay in draft mode to avoid merge~~

function deployed good to go


cpanato:
do we want to push to both staging and prod when we merge to the main branch? or main => staging and tag => prod?

PR #10, opened 2022-09-21
add 'make dev' target to develop in a container

imjasonh:
The changes to `MELANGE` and `MELANGE_DIR` are almost certainly incorrect, but they unblocked me from getting a build going.

That build still fails with a `bwrap` error though, so there's probably more left to do.

Feel free to ignore this until after any large higher-priority events are behind us. 😄

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 ARCH := $(shell uname -m)
-MELANGE_DIR ?= ../melange
-MELANGE ?= ${MELANGE_DIR}/melange
+MELANGE ?= $(shell which melange)
+MELANGE_DIR ?= /usr/share/melange
 KEY ?= local-melange.rsa
 REPO ?= $(shell pwd)/packages
 
@@ -210,3 +210,13 @@ $(eval $(call build-package,giflib,5.2.1-r0))
 $(eval $(call build-package,jenkins,2.378-r0))
 
 .build-packages: ${PACKAGES}
+
+dev:
+	docker run \
+		-v $(shell pwd):/work \
+		--workdir=/work \
+		--rm -it \
+		--entrypoint=sh \
+		--platform=linux/amd64 \
+		--privileged \
+		cgr.dev/chainguard/sdk
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 packages/*
+local-melange.rsa
+local-melange.rsa.pub


Makefile
```diff
@@ -1,6 +1,6 @@
 ARCH := $(shell uname -m)
-MELANGE_DIR ?= ../melange
-MELANGE ?= ${MELANGE_DIR}/melange
+MELANGE ?= $(shell which melange)
+MELANGE_DIR ?= /usr/share/melange
```
kaniini, line 3: we should probe for this, because on my desktop setup, it really is `../melange`.

Makefile
```diff
@@ -133,3 +133,13 @@ $(eval $(call build-package,s6,2.11.1.2-r0))
...
 
 .build-packages: ${PACKAGES}
+
+dev:
+	docker run \
+		-v $(shell pwd):/work \
+		--workdir=/work \
+		--rm -it \
+		--entrypoint=sh \
+		--platform=linux/amd64 \
```
rawlingsj, line 143: BTW I was getting a bwrap error too, `bwrap: Creating new namespace failed: Invalid argument`

With the help of @amouat we figured I needed to use `--platform linux/arm64`.  Not sure if that's the same issue you had.  

Maybe we could use `--platform=linux/${ARCH}` ?

imjasonh: So I'm on an M1 mac (it sounds like you are too?).

If I run this with `--platform=linux/arm64` I get:

```
Sep 23 12:43:17.010 [DEBUG] [arch:aarch64] [cmd:apk] [use-proot:false] [use-qemu:] fetch https://packages.wolfi.dev/bootstrap/stage3/aarch64/APKINDEX.tar.gz
Sep 23 12:43:17.201 [DEBUG] [arch:aarch64] [cmd:apk] [use-proot:false] [use-qemu:] WARNING: Ignoring https://packages.wolfi.dev/bootstrap/stage3: No such file or directory
```

Because Wolfi packages aren't currently built for arm64 yet.

If I run this with `--platform=linux/amd64` I get further, but fail with bwrap:

```
2022/09/23 12:48:25 melange (gmp/x86_64): bwrap: Creating new namespace failed: Invalid argument
```

If I use an x86_64 machine and `--platform=linux/amd64` I get past the bwrap error and can actually build, so I think it's an issue with using Docker-for-M1's emulation layer.

If a generalized dev container isn't possible, we should document what's necessary (an x86_64 machine, what tools, etc.), and I'll get one of those instead.

amouat: Yeah, this is a pain.

I'm pretty sure it is an emulation thing. I'm not sure if it's 2x QEMU or rosetta & QEMU, but running the x86_64 image on m1 does seem to fail with bwrap.

I suspect what will work is running in a full blown Alpine VM on a Mac. Has anyone tried with Lima? https://github.com/chainguard-dev/apko/blob/main/mac/README.md

rawlingsj: > So I'm on an M1 mac (it sounds like you are too?). 

Ah sorry, yes and building x86_64

jdolitsky:
There is now a wolfi-based sdk image we can use:
```
cgr.dev/chainguard/sdk:latest-glibc
```

However, on M1 mac, it still fails builds with:
```
bwrap: Creating new namespace failed: Invalid argument
```

To reproduce this:

1. Get the Wolfi sources:
```
git clone https://github.com/wolfi-dev/os.git && cd os/
```

2. Open a shell in the SDK environment:
```
docker run --platform linux/amd64 --privileged --rm -it -v "${PWD}:${PWD}" -w "${PWD}" cgr.dev/chainguard/sdk:latest-glibc
```

3. Try building a package (e.g. Grype):
```
melange build --empty-workspace --arch x86_64 grype.yaml
```

rawlingsj:
@imjasonh is this work being done as part of apko + melanage on Mac?  Maybe we can close?

PR #15, opened 2022-09-26
icu: new package

jdolitsky:
Required dependency for node, based on https://git.alpinelinux.org/aports/tree/main/icu/APKBUILD

Relies on the following 2 open PRs in melange:

- https://github.com/chainguard-dev/melange/pull/124
- https://github.com/chainguard-dev/melange/pull/126


dlorenc (approved):
nice!

PR #18, opened 2022-09-28
add patch, patches

rawlingsj:

### DIFF
--- a/.github/workflows/push-production.yaml
+++ b/.github/workflows/push-production.yaml
@@ -174,6 +174,7 @@ jobs:
       - uses: chainguard-dev/actions/inky-build-pkg@main
         with:
           package-name: patch
+          source-dir: ${{ github.workspace }}/patch
 
       - uses: chainguard-dev/actions/inky-build-pkg@main
         with:
--- a/patch.yaml
+++ b/patch.yaml
@@ -13,6 +13,20 @@ package:
   dependencies:
     runtime:
 
+secfixes:
+  2.7.6-r7:
+    - CVE-2019-20633
+  2.7.6-r6:
+    - CVE-2018-1000156
+    - CVE-2019-13638
+    - CVE-2018-20969
+  2.7.6-r5:
+    - CVE-2019-13636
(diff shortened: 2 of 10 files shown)


cpanato (approved):

imjasonh (approved):

patch.yaml
```diff
@@ -13,6 +13,20 @@ package:
   dependencies:
     runtime:
 
+secfixes:
+  2.7.6-r7:
```
kaniini, line 17: `epoch` needs to be bumped to make the version `2.7.6-r3`.

These patches should be acknowledged using the Wolfi version, not the Alpine version.  So, all of these should be `2.7.6-r3`.

patch.yaml
```diff
@@ -30,6 +44,77 @@ pipeline:
     with:
       uri: https://ftp.gnu.org/gnu/patch/patch-${{package.version}}.tar.gz
       expected-sha256: 8cf86e00ad3aaa6d26aca30640e86b0e3e1f395ed99f189b06d4c9f74bc58a4e
+
+  - uses: fetch
```
kaniini, line 48: Please copy the patches into our repository, instead of fetching them from Alpine's cgit instance.

kaniini (changes requested):
`epoch` needs to be bumped, `secfixes` needs to reflect the wolfi version introducing the fix, patches need to be copied instead of fetched.

patch.yaml
rawlingsj: thanks - done

.github/workflows/push-production.yaml
```diff
@@ -182,6 +182,8 @@ jobs:
       - uses: chainguard-dev/actions/inky-build-pkg@main
         with:
           package-name: patch
+          empty-workspace: false
```
kaniini, line 185: This isn't needed anymore, the action now just creates an empty directory if one didn't already exist.

kaniini (changes requested):

kaniini (approved):

PR #24, opened 2022-10-11
add patch for zlib CVE-2022-37434

rawlingsj:


rawlingsj:
In draft as still verifying

zlib.yaml
```diff
@@ -13,6 +13,12 @@ package:
   dependencies:
     runtime:
 
+secfixes:
+  1.2.12-r3:
+    - CVE-2022-37434
+  1.2.11-r4:
```
rawlingsj, line 19: should this instead be the first version of wolfi zlib?  i.e. `1.2.12-r2`

rawlingsj: @kaniini just checking ^

kaniini: 1.2.12-r0

PR #26, opened 2022-10-12
Explain how to patch CVEs

luhring:


HOW_TO_PATCH_CVES.md
```diff
@@ -0,0 +1,63 @@
...
+    - uses: patch
+      with:
+        patches: CVE-2018-25032.patch
+    ```
+
+    For more information on how `patch` works, see [its definition](https://github.com/chainguard-dev/melange/blob/main/pipelines/patch.yaml).
+
+1. In the [Makefile](./Makefile), find the line that corresponds to this package, and update the package version to our new release version from the Melange file.
+
+1. Verify that our update package will build successfully by running Melange.
```
luhring, line 49: I'd love to include the command to run to do this (I just missed it during today's session)

kaniini: `doas make`, but that will build the entire world at the moment.  there is also `doas make packages/${ARCH}/foo-1.2.3-r4.apk` option to make an individual package for a given arch.

kaniini (approved):

HOW_TO_PATCH_CVES.md
luhring: Cool thanks — added in 3b3f0f1

imjasonh:
There was another case we covered where the vuln had already been fixed in the version we had. I think we just noted it in secfixes and didn't patch anything.

Is this case common enough in practice to warrant documenting it?

luhring:
@imjasonh I think that case is worth a mention, thanks for spotting that

PR #29, opened 2022-10-13
push-staging: do reposync before build, like CI

kaniini:


.github/workflows/push-staging.yaml
```diff
@@ -49,9 +49,15 @@ jobs:
...
       - name: 'Prepare package repository'
         run: |
-          mkdir "${{ github.workspace }}/packages"
+          mkdir -p "${{ github.workspace }}/packages"
           cp /etc/apk/keys/wolfi-signing.rsa.pub ${{ github.workspace }}/packages/wolfi-signing.rsa.pub
 
+      - name: 'Sync public package repository'
+        run: |
+          mkdir -p "${{ github.workspace }}/packages"
+          gsutil -m rsync -r gs://wolfi-production-registry-destination/os/ "${{ github.workspace }}/packages/"
```
jdolitsky, line 58: are the args reversed?

jdolitsky (approved):

luhring (approved):

puerco (approved):
LGTM, only thing I'm thinking is that perhaps we should find a way to make these runs more deterministic when rerun :thinking:

PR #34, opened 2022-10-17
bubblewrap: new package

jdolitsky:
Depends on #33 being released (libcap)

### DIFF
--- a/bubblewrap.yaml
+++ b/bubblewrap.yaml
@@ -1,0 +1,44 @@
+package:
+  name: bubblewrap
+  version: 0.6.2
+  epoch: 0
+  description: "Unprivileged sandboxing tool"
+  target-architecture:
+    - all
+  copyright:
+    - paths:
+      - "*"
+      attestation: TODO
+      license: LGPL-2.0-or-later
+  dependencies:
+    runtime:
+
+environment:
+  contents:
+    keyring:
+      - https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
+    repositories:
+      - https://packages.wolfi.dev/os
+    packages:
+      - ca-certificates-bundle
+      - busybox
+      - build-base
+      - bash
+      - meson
+      - libcap-dev
+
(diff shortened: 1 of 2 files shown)


bubblewrap.yaml
```diff
@@ -0,0 +1,44 @@
...
+  - runs: |
+      meson \
+        --prefix=/usr \
+        --sysconfdir=/etc \
+        --mandir=/usr/share/man \
+        --localstatedir=/var \
+        --buildtype=plain \
+        -Drequire_userns=true . output
+      meson compile -C output
+      DESTDIR="${{targets.destdir}}" meson install --no-rebuild -C output
```
kaniini, line 44: Should have docs subpackage, right?  It generates a manpage I think.

jdolitsky: no, no manpages:
```
$ tar -xvf packages/x86_64/bubblewrap-0.6.2-r0.apk
x .PKGINFO
x usr
x usr/bin
x usr/bin/bwrap
x usr/share
x usr/share/bash-completion
x usr/share/bash-completion/completions
x usr/share/bash-completion/completions/bwrap
x usr/share/zsh
x usr/share/zsh/site-functions
x usr/share/zsh/site-functions/_bwrap
```

jdolitsky: the meson command here is gleaned from the alpine build which uses `abuild-meson`, which appears to do just the following:
https://gitlab.alpinelinux.org/alpine/abuild/-/merge_requests/60/diffs

jdolitsky: maybe since we dont have the `docbook-xsl` dependency?

jdolitsky: https://git.alpinelinux.org/aports/tree/main/bubblewrap/APKBUILD#n11

kaniini: oh, right, i didn't package any of that stuff because it was annoying.

PR #38, opened 2022-10-19
add gperf, libbsd, libmd, libpthread-stubs

rawlingsj:

### DIFF
--- a/gperf.yaml
+++ b/gperf.yaml
@@ -1,0 +1,37 @@
+package:
+  name: gperf
+  version: 3.1
+  epoch: 0
+  description: Perfect hash function generator.
+  target-architecture:
+    - all
+  copyright:
+    - paths:
+        - "*"
+      attestation: TODO
+      license: GPL-3.0-or-later
+environment:
+  contents:
+    repositories:
+      - https://packages.wolfi.dev/os
+    keyring:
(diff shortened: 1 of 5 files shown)


rawlingsj:
this is the start of the JDK dependencies.

gperf.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - all
+  copyright:
+    - paths:
+        - "*"
+      attestation: TODO
+      license: GPL-3.0-or-later
+environment:
+  contents:
+    repositories:
+      - https://packages.wolfi.dev/bootstrap/stage3
```
kaniini, line 16: Stage3 should be dropped here.

gperf.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - https://packages.wolfi.dev/os
+    keyring:
+      - https://packages.wolfi.dev/bootstrap/stage3/wolfi-signing.rsa.pub
```
kaniini, line 19: Likewise.  Also, for future reference, the signing key is redundant here (both are the same key).

kaniini (changes requested):
These look mostly fine, except for the stage3 references, which need to be removed.

gperf.yaml
rawlingsj: great - will sort that now

kaniini (approved):

PR #39, opened 2022-10-19
add util-macros, xorgproto, libgpg-error, libgcrypt

rawlingsj:


rawlingsj:
@kaniini would you mind double checking `xorgproto` as [alpine](https://git.alpinelinux.org/aports/plain/main/xorgproto/APKBUILD) uses `replace` and `provides` keys which I wasn't sure if we need to handle?

kaniini:
We don't need the replaces/provides for Wolfi :)

kaniini (approved):

PR #43, opened 2022-10-20
more JDK deps, libxml2, perl-test-pod, perl-yaml-syck, libxslt, xmlto…

rawlingsj:
…, libxau, xtrans

### DIFF
--- a/libxml2.yaml
+++ b/libxml2.yaml
@@ -1,0 +1,75 @@
+package:
+  name: libxml2
+  version: 2.10.3
+  epoch: 0
+  description: XML parsing library, version 2
+  target-architecture:
+    - all
+  copyright:
+    - paths:
+        - "*"
+      attestation: TODO
+      license: MIT
+
+secfixes:
+  2.10.3-r0:
+    - CVE-2022-40303
+    - CVE-2022-40304
+
+environment:
+  contents:
+    repositories:
+      - https://packages.wolfi.dev/os
+    keyring:
+      - https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+      - automake
+      - autoconf
+      - zlib-dev
+      - xz-dev
+      - python3-dev
+      - python3
+pipeline:
+  - uses: fetch
+    with:
+      expected-sha256: 5d2cc3d78bec3dbe212a9d7fa629ada25a7da928af432c93060ff5c17ee28a9c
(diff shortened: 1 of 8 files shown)


libxml2.yaml
```diff
@@ -0,0 +1,75 @@
...
+  copyright:
+    - paths:
+        - "*"
+      attestation: TODO
+      license: MIT
+
+secfixes:
+  2.10.3-r0:
+    - CVE-2022-40303
+    - CVE-2022-40304
```
rawlingsj, line 17: Is this correct?  If we add a new package with a secfix for the same version we are publishing, should we include these here?  
Also I've not included secfixes for previous versions?

rawlingsj:
Sorry for the larger review.  There was a few related to each other so figured it best to push together.  I can push smaller or larger groups if preferred.

rawlingsj:
hmm CI just failed with an error I've not seen when testing
```
2022/10/20 21:27:23 melange (libxml2/x86_64): /usr/lib/gcc/x86_64-pc-linux-gnu/12.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: /lib/../lib64/libz.a(crc32.o): relocation R_X86_64_32 against `.rodata' can not be used when making a shared object; recompile with -fPIC
2022/10/20 21:27:23 melange (libxml2/x86_64): /usr/lib/gcc/x86_64-pc-linux-gnu/12.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: failed to set dynamic section sizes: bad value
```

rawlingsj:
specifically the error `recompile with -fPIC` 

I'm not sure if that's the result of something I've missed from this PR or if it's a valid problem?

rawlingsj:
tried a change but it didn't help so reverted back to the original commit + issue

libxml2.yaml
kaniini: yes, that seems fine

kaniini:
It's because libxml2 build system has decided it wants to staticly link zlib in.  I split out the zlib static library into its own package, to help give the libxml2 build system an attitude adjustment.

kaniini:
GNOME appears to have recut the libxslt release since you wrote libxslt.yaml: 2022/10/20 23:01:02 melange (libxslt/x86_64): sha256sum: WARNING: 1 of 1 computed checksums did NOT match

Sigh.

rawlingsj:
> It's because libxml2 build system has decided it wants to staticly link zlib in. I split out the zlib static library into its own package, to help give the libxml2 build system an attitude adjustment.

thank you!

PR #47, opened 2022-10-25
more JDK deps, add libxrandr, libxfixes, libxi, libxtst, check, libic…

rawlingsj:
…e, libsm

### DIFF
--- a/libsm.yaml
+++ b/libsm.yaml
@@ -1,0 +1,63 @@
+package:
+  name: libsm
+  version: 1.2.3
+  epoch: 0
+  description: X11 Session Management library
+  target-architecture:
+    - all
+  copyright:
+    - paths:
+        - "*"
+      attestation: TODO
+      license: MIT
+environment:
+  contents:
+    repositories:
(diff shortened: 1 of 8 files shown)


libsm.yaml
```diff
@@ -0,0 +1,63 @@
...
+      expected-sha256: 2d264499dcb05f56438dee12a1b4b71d76736ce7ba7aa6efbf15ebb113769cbb
+      uri: https://www.x.org/releases/individual/lib/libSM-${{package.version}}.tar.bz2
+  - uses: autoconf/configure
+    with:
+      opts: |
+        --enable-docs \
+        --with-xmlto \
+        --without-fop
+
+  # --with-libuuid TODO as above, would rather not build util-linux-dev yet
```
rawlingsj, line 44: is this ok?  avoiding configuring `--with-libuuid` so we don't need to package  `util-linux` yet.

kaniini: Yeah it’s fine for now!

PR #48, opened 2022-10-25
more JDK deps, libxt, libusb, libevent, dbus, libpaper, cups

rawlingsj:


cups.yaml
```diff
@@ -0,0 +1,119 @@
...
+      - libpaper-dev
+      - dbus
+      - dbus-dev
+      - libusb-dev
+      - libgcrypt-dev
+
+      # - poppler-utils do we need this?
+      # - libjpeg-turbo-dev do we need this?
+      # - avahi-dev do we need this?
+      # - gnutls-dev do we need this?
```
rawlingsj, line 40: Not sure about these four packages.  Compiling seems fine as well as downstream packages that uses this.

If we don't need these I'll remove them from this PR or add a TODO.  If we do need these I'll need to create them as I've not done that yet.

dbus.yaml
```diff
@@ -0,0 +1,82 @@
...
+      - build-base
+      - automake
+      - autoconf
+      - expat-dev
+      - libsm-dev
+      - libx11-dev
+      - meson
+      - xmlto
+      - bash
+      # - glib-dev # TODO do we need this?  This and downstream packages are compiling fine
```
rawlingsj, line 38: Also not sure about `glib-dev`?

cups.yaml
kaniini: We will need them for CUPS to be actually useful for printing.

dbus.yaml
kaniini: This is needed for the libdbus-glib dependency, which is obsolete but still used by some software.

cups.yaml
rawlingsj: ok cool - I'll do those too

dbus.yaml
rawlingsj: ok sounds good, will package glib-dev too

rawlingsj: for now we could unblock JDK and address TODOs in a followup https://github.com/wolfi-dev/os/issues/49

cups.yaml
rawlingsj: for now we could unblock JDK and address TODOs in a followup https://github.com/wolfi-dev/os/issues/49

PR #54, opened 2022-10-26
add open JDK + JRE 17

rawlingsj:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -182,6 +182,6 @@ $(eval $(call build-package,zip,3.0-r0))
 $(eval $(call build-package,libpng,1.6.38-r0))
 $(eval $(call build-package,freetype,2.12.1-r0))
 $(eval $(call build-package,fontconfig,2.14.0-r0))
-$(eval $(call build-package,openjdk17,17.0.4.1-r0))
+$(eval $(call build-package,openjdk,17.0.4.1-r0))
 
 .build-packages: ${PACKAGES}
--- a/Makefile
+++ b/Makefile
@@ -182,5 +182,6 @@ $(eval $(call build-package,zip,3.0-r0))
 $(eval $(call build-package,libpng,1.6.38-r0))
 $(eval $(call build-package,freetype,2.12.1-r0))
 $(eval $(call build-package,fontconfig,2.14.0-r0))
+$(eval $(call build-package,openjdk17,17.0.4.1-r0))
 
 .build-packages: ${PACKAGES}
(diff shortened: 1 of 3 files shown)


openjdk17.yaml
```diff
@@ -0,0 +1,104 @@
...
+      uri: https://github.com/google/googletest/archive/release-1.8.1.tar.gz
+      expected-sha256: 9bf1fe5182a604b4135edc1a425ae356c9ad15e9b23f9f12a02e80184c3a249c
+      strip-components: 0
+  - runs: chmod +x configure
+  - uses: autoconf/configure
+    with:
+      opts: |
+        --with-boot-jdk=./jdk-16.0.2+7 \
+        --with-vendor-name=Wolfi \
+        --with-vendor-url=https://github.com/wolfi-dev \
```
rawlingsj, line 63: are these vendor flags correct?

kaniini: i would use wolfi.dev here

rawlingsj: yeah that sounds better

Makefile
```diff
@@ -182,5 +182,6 @@ $(eval $(call build-package,zip,3.0-r0))
 $(eval $(call build-package,libpng,1.6.38-r0))
 $(eval $(call build-package,freetype,2.12.1-r0))
 $(eval $(call build-package,fontconfig,2.14.0-r0))
+$(eval $(call build-package,openjdk17,17.0.4.1-r0))
```
kaniini, line 185: lets just call it openjdk :)

openjdk17.yaml
```diff
@@ -0,0 +1,104 @@
...
+        --disable-precompiled-headers \
+        --enable-dtrace=no \
+        --with-zlib=system \
+        --with-native-debug-symbols=none \
+        --disable-warnings-as-errors \
+        --disable-precompiled-headers \
+        --with-jvm-variants=server \
+        --with-debug-level=release \
+        --with-jtreg=no
+  # TODO revisit below args
```
kaniini, line 77: these need to be packaged in the near future but we can proceed for now

kaniini (changes requested):

Makefile
rawlingsj: OK cool.  I wasn't sure if we needed to include the major version in the name so we could have multiple LTS version packages. Or maybe this Java package should track the latest (currently Java 19) though I suspect that may be a problem for some of the Java based applications we want to build as a lot are using 17.  

Or best to merge this package without the version in the name and figure out different major versions after?  The supported Java versions is interesting https://endoflife.date/java

kaniini (approved):

PR #56, opened 2022-10-28
nodejs: include npm

imjasonh:
This installs npm when nodejs is installed. https://nodejs.org/en/download/releases/ says nodejs 18.11 includes npm 8.19.2.

This differs from Alpine's approach, whose node package doesn't include npm, and instead it's provided by a [separate package in the community repo](https://pkgs.alpinelinux.org/packages?name=npm&branch=edge&repo=&arch=&maintainer=), which installs npm 9.0.0.

I can have a separate npm package as well if we'd like, and just require folks to install them both together.


imjasonh:
Some more context about Alpine's decision to use a separate package, from when they split it out last year: https://git.alpinelinux.org/aports/commit/main/nodejs/APKBUILD?id=25b10bd1a93e12a7e49fee38b0a229281ae49fb7

I don't have a strong opinion, and we can change to separate packages later if we decide we'd like to.

luhring:
@imjasonh Could you elaborate on the motivation for this? Is this for convenience for developers using the package?

At first glance I kind of lean toward the separated/minimalist approach, for the reasons mentioned in the Alpine notes. Also thinking about non-dev environments where there's no need for npm.

imjasonh:
That's reasonable. I'd also like to enable a minimal base image capable of running node apps, with none of the stuff needed to build them.

I couldn't find a good example of doing that though -- both the official [Node](https://nodejs.org/en/docs/guides/nodejs-docker-webapp/) and [Docker](https://docs.docker.com/language/nodejs/build-images/) tutorials for containerizing node apps recommend basing them on their `node` image, which includes `npm`. 😕 

That's not to say we shouldn't do better, we should. Not being a JS expert myself I just didn't want to do a more complicated thing without a bit more assurance I wasn't making things worse.

To get that assurance I'd want to find/build a runnable example using a multi-stage build and separate dev-time and run-time images, but that will take time. Having a usable node+npm image at all seems like a good first step, and we can build on it to separate out npm when it's all a bit more mature.

luhring:
That sounds reasonable to me!

luhring (approved):

PR #59, opened 2022-10-29
add maven

rawlingsj:


rawlingsj:
Marked as draft as wanted to confirm the approach of using binaries rather than compiling from source.  This is what alpine is doing and I'm assuming there's a reason for that as there's a lot of parts to maven.

rawlingsj:
closing this and will recreate in a followup once java changes are merged

PR #61, opened 2022-10-31
add postgres package

rawlingsj:


rawlingsj:
contributed by @rawlingsj @luhring @kaniini @patflynn @jvanzyl

rawlingsj:
hmm, that CI build finished very quickly.  I don't think it actually built the postgres package

postgres/postgresql-entrypoint.sh
```diff
@@ -0,0 +1,339 @@
+#!/usr/bin/env bash
```
luhring, line 1: Would it be helpful at all to put a note at the top here about where this script came from / what it was based on?

rawlingsj: good call, will do that now

postgresql.yaml
```diff
@@ -0,0 +1,93 @@
+# Generated from https://git.alpinelinux.org/aports/plain/main/postgresql/APKBUILD?h=3.3-stable
```
kaniini, line 1: uhh :)

postgresql.yaml
```diff
@@ -0,0 +1,93 @@
...
+
+  - name: libpq
+    description: PostgreSQL libraries
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/usr/lib
+          mv ${{targets.destdir}}/usr/lib/libpq.so* ${{targets.subpkgdir}}/usr/lib/
+
+  - name: postgresql-oci-entrypoint
+    description: PostgreSQL libraries
```
kaniini, line 85:
```suggestion
    description: Entrypoint for using PostgreSQL in OCI containers
```

kaniini (changes requested):

postgresql.yaml
rawlingsj: damn! removing now

kaniini (approved):

PR #19, opened 2022-09-28
add script to generate graphviz of deps

imjasonh:
- support subgraphs of only certain packages
- add README describing usage
- add generated dag.svg
- verify dag.svg is up to date in PRs

PR #64, opened 2022-11-01
Update go to 1.19.3

luhring:
Resolves CVE-2022-41716


go.yaml
```diff
@@ -15,8 +15,16 @@ package:
       - build-base
       - bash
 
+secfixes:
+  1.19.3-r0:
+    - CVE-2022-41716
+
 environment:
   contents:
+    repositories:
```
kaniini, line 24: This isn't needed and shouldn't be present.  Instead, use `gsutil -m rsync` to clone the repository locally.  I can add some helpers to the Makefile to help with this.

kaniini (changes requested):

kaniini (approved):

tuananh:
@luhring off topic question but may i ask why is epoch set to 1?

kaniini:
Because go binaries are staticly linked, we have to rebuild them when Go receives security updates, so we increase the epoch to trigger a rebuild.

tuananh:
that seems like a bit of manual, error-prone process, maybe it's better to have a check in CI for if a package like `go` get updated? which print out error message like "trivy is staticly linked with go, because this PR updates go so you need to bump epoch for trivy too"

kaniini:
We use `dag` to handle this.  https://github.com/wolfi-dev/dag

tuananh:
This is nice. I saw there's a PR for it but it's closed? https://github.com/wolfi-dev/os/pull/19

imjasonh:
That work moved to https://github.com/wolfi-dev/dag -- there are a few goals of that tool:

1. visualization of the dependency graph, just to be able to tell what depends on what, track over time, etc.
2. execution of the graph when things change -- e.g., when `go` is rebuilt, know to update and rebuild `trivy`, etc.
3. execution of the graph remotely -- instead of having the build graph manually maintained in `Makefile`, have `dag` create a K8s pod that runs the necessary builds in order on cloud VMs. This will also help us support arm64 builds better.

If this is interesting feel free to take it for a spin, file issues, and ask questions. We can discuss in more detail in a future public community call as well.

PR #66, opened 2022-11-02
add packages: ko, apko, melange

imjasonh:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -185,5 +185,8 @@ $(eval $(call build-package,fontconfig,2.14.0-r0))
 $(eval $(call build-package,openjdk,17.0.4.1-r0))
 $(eval $(call build-package,su-exec,0.2-r0))
 $(eval $(call build-package,postgresql,11.17-r0))
+$(eval $(call build-package,ko,0.12.0-r0))
+$(eval $(call build-package,apko,0.5.0-r0))
+$(eval $(call build-package,melange,0.1.0-r0))
 
 .build-packages: ${PACKAGES}
--- a/apko.yaml
+++ b/apko.yaml
@@ -19,6 +19,7 @@ environment:
     repositories:
       - https://packages.wolfi.dev/os
     packages:
+      - busybox
       - ca-certificates-bundle
       - go
 
--- a/apko.yaml
+++ b/apko.yaml
@@ -1,0 +1,32 @@
+package:
+  name: apko
+  version: 0.5.0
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -185,5 +185,8 @@ $(eval $(call build-package,fontconfig,2.14.0-r0))
 $(eval $(call build-package,openjdk,17.0.4.1-r0))
 $(eval $(call build-package,su-exec,0.2-r0))
 $(eval $(call build-package,postgresql,11.17-r0))
+$(eval $(call build-package,ko,0.12.0-r0))
+$(eval $(call build-package,apko,0.5.0-r0))
+$(eval $(call build-package,melange,0.1.0-r0))
```
kaniini, line 190: Is it worth packaging this version of melange, considering it is too old to build Wolfi?

imjasonh: I can drop it if you want. I thought it'd be nice to start having it packaged, and we can bump it when we do the next melange release. If you'd prefer to wait until 0.2.0 I can do that too.

kaniini: We can just hold until apko 0.6 and melange 0.2 are shipped, which i might do tomorrow. :)

tuananh:
@imjasonh should we copy the melange pipelines to /usr/share/melange as well?

imjasonh:
> @imjasonh should we copy the melange pipelines to /usr/share/melange as well?

Starting soon™️ , melange will compile in the built-in pipelines, so this won't need to install into `/usr/share/melange/pipelines` 🚀

tuananh:
> > @imjasonh should we copy the melange pipelines to /usr/share/melange as well?
> 
> Starting soon™️ , melange will compile in the built-in pipelines, so this won't need to install into `/usr/share/melange/pipelines` 🚀 

Niceee.

If i want to add new pipeline in melange, what's the recommended way now?

I still want to use the default pipelines, just add some new one

imjasonh:
> If i want to add new pipeline in melange, what's the recommended way now?
> 
> I still want to use the default pipelines, just add some new one

If you want to add a new built-in pipeline, for all users of Melange, propose them here: https://github.com/chainguard-dev/melange/tree/main/pkg/build/pipelines

If you want to define pipelines only for your own use, put them anywhere in your local filesystem, and invoke `melange build --pipeline-dir=/path/to/my/pipelines`.

imjasonh:
Closing for now.

PR #71, opened 2022-11-03
add cosign pkg

tuananh:


cosign.yaml
```diff
@@ -0,0 +1,33 @@
...
+  description: Container Signing
+  target-architecture:
+    - all
+  copyright:
+    - license: Apache-2.0
+      paths:
+        - "*"
+  dependencies:
+    runtime:
+      - apk-tools
```
imjasonh, line 13: We shouldn't need this at runtime. I think we will need CA certs at runtime though.

cosign.yaml
```diff
@@ -0,0 +1,33 @@
+package:
+  name: cosign
+  version: 1.13.1
+  description: Container Signing
+  target-architecture:
```
kaniini, line 5: Please add `epoch: 0` here, even if it is moot, it makes it easier to bump the epoch later if it is already present.

cosign.yaml
```diff
@@ -0,0 +1,33 @@
...
+      paths:
+        - "*"
+  dependencies:
+    runtime:
+      - ca-certificates-bundle
+
+environment:
+  contents:
+    keyring:
+      - https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
```
kaniini, line 18: `keyring`/`repositories` are redundant here.

kaniini (approved):

cosign.yaml
```diff
@@ -0,0 +1,33 @@
...
+      - go
+
+pipeline:
+  - uses: fetch
+    with:
+      uri: https://github.com/sigstore/cosign/archive/v${{package.version}}/cosign-v${{package.version}}.tar.gz
+      expected-sha256: 496583e9a3e07f16ff18f3734ca69398ef7091336fe74adaf8012715656bb920
+  - runs: |
+      CGO_ENABLED=0 go build \
+        -o "${{targets.destdir}}/usr/bin/cosign" ./
```
kaniini, line 33: probably needs to be `go build ./...`.

kaniini (changes requested):

cosign.yaml
imjasonh: I think this should be `go build ./cmd/cosign`

cpanato: Can we also inject the version in the ldflags?

tuananh: i injected the version but git state is not there in the release tar archive so I just leave it out.

cosign.yaml
tuananh: i dont need to use keyring if repository is there? i saw the other packages they all have both.

kaniini: Some do but it is a problem for a few reasons so trying to avoid it

cosign.yaml
```diff
@@ -0,0 +1,33 @@
...
+    - license: Apache-2.0
+      paths:
+        - "*"
+  dependencies:
+    runtime:
+      - ca-certificates-bundle
+
+environment:
+  contents:
+    repositories:
```
kaniini, line 18: The `repositories` entry is redundant.

tuananh:
@imjasonh can you help with debugging the new lint script

i got error with this but everything looks good to me

```
--- cosign.yaml
build-package,cosign
,1.13.1
-r0
 not found in Makefile
Error: Process completed with exit code 1.
```

tuananh:
@kaniini can you help? i'm not sure what i did that make the lint script failed here.

rawlingsj:
The issue is with the lineendings in the `cosign.yaml` file.  I can push a PR to your branch if you like but it might be worth addressing yourself so future PRs don't suffer from the same issue.

I was able to recreate the problem locally by running the lint script locally but I'm not 100% sure if that will fail for you, are you on windows maybe?
```
./lint.sh
```

When I open your fork and cosign branch in VSCode, in the bottom right hand corner I see `CRLF`

Screenshot 2022-11-16 at 10 09 14

If you compare that to other yamls in that folder they have `LF`

In VSCode you can click that `CRLF` in the bottom bar and then select `LF`

Screenshot 2022-11-16 at 10 11 49

Now rerunning the `./lint.sh` script should pass.

See if that helps?  Also you'll need to do a quick rebase for the makefile conflict but that should be a small diff.

tuananh:
> 

this is it. thank you :)

tuananh:
is there anything else you may want to change @kaniini ?

kaniini (approved):

PR #72, opened 2022-11-03
add postgres 15 and rename existing postgres package to handle multip…

rawlingsj:
…le major version streams


postgresql11.yaml
```diff
@@ -1,5 +1,5 @@
 package:
-  name: postgresql
+  name: postgresql11
```
rawlingsj, line 2: `postgresql11` or `postgresql-11 `?

amouat: Personally I like the "-" but I wonder if there's a reason Alpine don't have it...

amouat: debian (apt) seems to use a -

kaniini: We can use `-` or no dash.  `@` is restricted for tagged repositories.

rawlingsj: Do one of you two want to make a call as I don't have an opinion?  Then I'll leave or change this PR.  

I've got a followup for Java ready so wont push that until I know if we are using a dash or not.

kaniini (approved):

PR #76, opened 2022-11-04
new package: kubectl

imjasonh:

### DIFF
--- a/popt.yaml
+++ b/popt.yaml
@@ -1,0 +1,42 @@
+package:
+  name: popt
+  version: 1.19
+  epoch: 0
+  description: commandline option parser
+  target-architecture:
+    - all
+  copyright:
+    - license: MIT
+      paths:
+        - "*"
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+
+pipeline:
+  - uses: fetch
+    with:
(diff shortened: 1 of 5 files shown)


luhring (dismissed):

popt.yaml
```diff
@@ -0,0 +1,42 @@
...
+      expected-sha512: 5d1b6a15337e4cd5991817c1957f97fc4ed98659870017c08f26f754e34add31d639d55ee77ca31f29bb631c0b53368c1893bd96cf76422d257f7997a11f6466
+  - runs: |
+      ./configure \
+        --build=$CBUILD \
+        --host=$CHOST \
+        --prefix=/usr \
+        --libdir=/lib \
+        --disable-static
+      make
+      make DESTDIR="${{targets.destdir}}" install
```
luhring, line 33: (non-blocking, just something I'm wondering about) When would we want to use pipeline actions for this instead of `runs` scripting? I recently started looking through all our melange YAMLs, and see a lot using `uses: autoconf/make` (and make install, etc.)

luhring (approved):

popt.yaml
kaniini: we should prefer to use the pipelines whenever possible, since they allow us a central point to define policy.

imjasonh: I can make that a lint check if we want, even just to warn about pipeline-able `./configure`s at least.

I'll merge this now just to put a stake in the ground and come back later to standardize on using the pipeline more consistently.

Issue #55, opened 2022-10-28
Build libjpeg, giflib, lcms, and update Java package to include them


rawlingsj:
To unblock the JDK package we are not configuring libjpeg, giflib and lcms yet.  We need to build these packages and update the JDK package.

Related review comment https://github.com/wolfi-dev/os/pull/54#discussion_r1007212722

PR #77, opened 2022-11-04
Add giflib pkg

tuananh:
i'm still new to this so bear with me here :)

partially address #55 

btw is there a way to add shasum for patch too?


kaniini:
This looks alright but it needs to be added to the Makefile.

giflib.yaml
```diff
@@ -0,0 +1,66 @@
...
+      - libtool
+      - xmlto
+      - coreutils
+
+pipeline:
+  - uses: fetch
+    with:
+      uri: http://downloads.sourceforge.net/sourceforge/giflib/giflib-${{package.version}}.tar.gz
+      expected-sha256: 31da5562f44c5f15d63340a09a4fd62b48c45620cd302f77a6d9acf0077879bd
+  - uses: autoconf/configure
```
kaniini, line 36: The patch isn’t getting applied because the patch step is missing.

```yaml
- uses: patch
  with:
    patches: CVE-2022-28506.patch
```

kaniini:
It seems that the autoconf build files are missing from giflib?  Not sure, will look into this in the morning.

kaniini:
The `giflib-doc` package has no contents, which is why it is erroring.

tuananh:
when i run `make` locally, i can see doc folder generated. not sure why it's missing.

giflib.yaml
```diff
@@ -0,0 +1,65 @@
...
+      runtime:
+        - giflib
+
+  - name: "giflib-doc"
+    description: "giflib documentation"
+    pipeline:
+      - uses: split/manpages
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/usr/share
+          mv "${{targets.destdir}}"/doc "${{targets.subpkgdir}}"/usr/share/
```
kaniini, line 58:
```suggestion
          mkdir -p "${{targets.subpkgdir}}"/usr/share/doc/
          mv doc "${{targets.subpkgdir}}"/usr/share/doc/giflib

```

kaniini (approved):

PR #79, opened 2022-11-04
add maven package

rawlingsj:


rawlingsj:
Note that we are packaging maven binaries here and not building maven from source.  This seems to be the same approach Alpine has taken. 
```
mvn -v
Apache Maven 3.8.6 (84538c9988a25aec085021c365c560670ad80f63)
Maven home: /usr/share/java/maven
Java version: 17.0.5-internal, vendor: wolfi, runtime: /usr/lib/jvm/openjdk
Default locale: en_US, platform encoding: ANSI_X3.4-1968
OS name: "linux", version: "5.10.104-linuxkit", arch: "amd64", family: "unix"
```
@kaniini @jvanzyl is using maven binaries ok here?

kaniini:
Seems fine to me, but will wait on @jvanzyl to make his opinion known :)

kaniini (approved):

PR #83, opened 2022-11-04
build system: add packages/name alias for convenience

kaniini:
This allows, e.g. `make packages/tzdata` without having to build all of world, or type in the full package name.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ MELANGE_OPTS += ${MELANGE_EXTRA_OPTS}
 
 define build-package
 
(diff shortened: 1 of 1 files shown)


imjasonh:
Maybe I just don't know make well enough, but (how) does this detect the version+epoch necessary to populate `(2)`? Or does this just produce an apk named like `packages/tzdata.apk` which isn't meant to be published?

kaniini:
The $(1) and $(2) come from macro expansion and so are known at the time the macro is expanded.

PR #85, opened 2022-11-04
fix erroneus libice-static definitions

imjasonh:
The real libice-static-1.0.10-r0.apk should be unaffected and not need an epoch bump (AFAIK), but due to this issue there will be packages published that don't make sense:

- libice-static-2.12.1-r0.apk (from freetype.yaml)
- libice-static-1.6.38-r0.apk (from libpng.yaml)

### DIFF
--- a/freetype.yaml
+++ b/freetype.yaml
@@ -52,8 +52,8 @@ pipeline:
   - uses: autoconf/make-install
   - uses: strip
 subpackages:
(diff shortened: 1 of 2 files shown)


luhring (approved):
Thanks @imjasonh!

rawlingsj (approved):

PR #90, opened 2022-11-06
Add argon2 pkg

Shikachuu:
I'm still new to this melange build thing, but I had a really hard time since the documentation for certain features are missing (strip, strip-dev, git).
Yet I hope I did evetything correctly!

Added packages:
- argon2
- argon2-dev

The reasoning behing this pacakge that I'm actively work on a PHP melange script.
To build PHP from source with the `--with-password-argon2` flag it requires argon2-dev, and the README stated that we should not mix and match repositories.


argon2.yaml
```diff
@@ -0,0 +1,47 @@
...
+  dependencies:
+    runtime:
+
+environment:
+  contents:
+    repositories:
+      - https://packages.wolfi.dev/os
+    keyring:
+      - https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
+    environment:
```
kaniini, line 20: This part isn’t needed.

argon2.yaml
```diff
@@ -0,0 +1,47 @@
...
+      PREFIX: ${{targets.destdir}}
+      DESTDIR: ${{targets.destdir}}
+    packages:
+      - wolfi-base
+      - build-base
+      - git
```
kaniini, line 26: This part isn’t needed, the git checkout action will indicate it needs git in the build environment.

argon2.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/P-H-C/phc-winner-argon2.git
+      destination: /home/build/source
```
kaniini, line 32: destination isn’t needed here, it defaults to the current directory

argon2.yaml
```diff
@@ -0,0 +1,47 @@
...
+      depth: 1
+      branch: ${{package.version}}
+  - uses: autoconf/make
+    with:
+      dir: /home/build/source
```
kaniini, line 37: Since directory isn’t needed these aren’t needed too.

kaniini (changes requested):

argon2.yaml
Shikachuu: Can you ellaborate this please?
I'll ofc remove that part, but tbh I do not fully understand why.

rawlingsj: It's because existing packages are pulled from a storage bucket and mounted to the local filesystem during CI.  This means we don't need to configure the repository or keyring which is used to serve up the same packages from the mounted storage bucket.

So you can remove the `repository` and `keyring` sections above, e,g it can look like this https://github.com/wolfi-dev/os/blob/6370c077ce82c9e50ba52a304d783acbc832a67a/git.yaml#L19-L21

We'll clean the existing melange configs up so it's consistent.

Shikachuu: Thanks, everything is clear now.

rawlingsj:
what's the status of this?  will close if no updates in 2 days, we can always reopen.

Shikachuu:
> what's the status of this? will close if no updates in 2 days, we can always reopen.

I did every requested changes and did a rebase too, sorry for the delay

luhring (approved):

luhring (changes requested):
@Shikachuu Looks good in general, but check out the lint check failure

Makefile
```diff
@@ -245,5 +245,6 @@ $(eval $(call build-package,php,8.1.13-r0))
 $(eval $(call build-package,composer,2.4.4-r0))
 $(eval $(call build-package,docker-credential-ecr-login,0.6.0-r1))
 $(eval $(call build-package,wait-for-it,0.20200823-r1))
+$(eval $(call build-package,argon2,20190702))
```
curantes, line 248:
```suggestion
$(eval $(call build-package,argon2,20190702-r0))
```

rawlingsj (dismissed):

imjasonh (approved):

imjasonh:
@luhring you're still requesting changes, does this LGTY?

luhring (approved):

kaniini (approved):

eddiezane:
Hey @Shikachuu,

Thanks for this contribution! I'd like to send some swag your way. If interested could you please shoot an email to eddiezane at chainguard dot dev?

PR #92, opened 2022-11-07
postgres 12, 13, 14 packages

rawlingsj:


postgresql-12.yaml
```diff
@@ -0,0 +1,96 @@
...
+        - "*"
+      attestation: TODO
+      license: BSD
+
+environment:
+  contents:
+    repositories:
+      - https://packages.wolfi.dev/os
+    keyring:
+      - https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
```
imjasonh, line 19:
```suggestion
```

postgresql-13.yaml
```diff
@@ -0,0 +1,96 @@
...
+        - "*"
+      attestation: TODO
+      license: BSD
+
+environment:
+  contents:
+    repositories:
+      - https://packages.wolfi.dev/os
+    keyring:
+      - https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
```
imjasonh, line 19:
```suggestion
```

postgresql-14.yaml
```diff
@@ -0,0 +1,96 @@
...
+        - "*"
+      attestation: TODO
+      license: BSD
+
+environment:
+  contents:
+    repositories:
+      - https://packages.wolfi.dev/os
+    keyring:
+      - https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
```
imjasonh, line 19:
```suggestion
```

postgresql-12/postgresql-entrypoint.sh
```diff
@@ -0,0 +1,343 @@
+#!/usr/bin/env bash
```
imjasonh, line 1: Hard to tell at a glance how different these files are, and if there might be some way to share them. Is there any benefit to having one `postgresql-entrypoint.sh` that gets referenced by each pipeline?

kaniini (dismissed):

postgresql-12/postgresql-entrypoint.sh
rawlingsj: Yeah they're exactly the same - I did wonder whether to reference a single one or duplicated them for each version as the melange source directory mounts a folder for sources, that matches the package name.  In this postgres case it means a folder name with the version number in it.  

Having one is better though, will switch and have a hard link to the `postgresql-entrypoint.sh`

rawlingsj:
@kaniini  @imjasonh latest commit uses the same entrypoint file for multiple postgres packages.  I did a local test and the default source-dir remains the package name so not to break existing behaviour.

kaniini (approved):

kaniini:
Is this ready to go in?

rawlingsj:
Yes ready to go.

PR #95, opened 2022-11-08
make python3 an explicit runtime dependency for the python3-dev package

rawlingsj:


rawlingsj:
Is is correct to bump the `epoch` for all melange configs I edited in this PR?  Even though the resulting apk is the same?

fontconfig.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: fontconfig
   version: 2.14.0
-  epoch: 0
+  epoch: 1
```
kaniini, line 4: epoch bump is only needed for python3 part.

kaniini (changes requested):

fontconfig.yaml
rawlingsj: Thats great, thanks.  I'll update the PR now.

kaniini (approved):

PR #96, opened 2022-11-08
removes the mavenrc from the package, let apko configs handle defaulu…

rawlingsj:
…t env vars

I noticed by including the M2_HOME default env var via the package caused issues running downstream maven tests that have multiple maven versions in unit tests.

### DIFF
--- a/maven.yaml
+++ b/maven.yaml
@@ -1,7 +1,7 @@
 package:
   name: maven
   version: 3.8.6
-  epoch: 0
+  epoch: 1
   description: A Java project management and project comprehension tool.
   target-architecture:
     - all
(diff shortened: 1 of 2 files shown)


imjasonh (dismissed):

maven.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: maven
   version: 3.8.6
-  epoch: 0
+  epoch: 1
```
imjasonh, line 4: Need to update Makefile

rawlingsj:
> rawlingsj dismissed  imjasonh stale review

sorry - I went to say thank you but it closed as stale instead :)

kaniini (approved):

PR #98, opened 2022-11-09
add dejavu and dependent packages, enable Java packages to work with …

rawlingsj:
…fonts

### DIFF
--- a/fontconfig.yaml
+++ b/fontconfig.yaml
@@ -10,6 +10,14 @@ package:
         - "*"
       attestation: TODO
       license: MIT
+  scriptlets:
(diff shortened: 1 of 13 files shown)


fontconfig.yaml
```diff
@@ -10,6 +10,14 @@ package:
         - "*"
       attestation: TODO
       license: MIT
+  scriptlets:
+    trigger:
```
rawlingsj, line 14: @kaniini is this correct?  I'd missed these triggers before.  I've seen some examples already merged in wolfi but not 100% how triggers work to test this.

PR #100, opened 2022-11-09
add dejavu, encodings, font-util, libfontenc, mkfontscale, tini

rawlingsj:
also enables using fonts with Java packages


fontconfig.yaml
```diff
@@ -10,6 +10,13 @@ package:
         - "*"
       attestation: TODO
       license: MIT
+  scriptlets:
```
rawlingsj, line 13: I'm not 100% sure about these scriptlets. I've used examples already in wolfi, are they correct?

openjdk-11.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: openjdk-11
   version: 11.0.17.8
-  epoch: 0
+  epoch: 1
```
rawlingsj, line 4: Is this correct to bump the epoch as the package will include extra runtime dependencies.

mkfontscale.yaml
```diff
@@ -0,0 +1,48 @@
...
+      attestation: TODO
+      license: MIT
+  scriptlets:
+    trigger:
+      paths:
+        - /usr/share/fonts/*
+      script: |
+          #!/bin/sh
+          for i in "$@"; do
+              mkfontscale "$i"
```
kaniini, line 20: This one should also call `mkfontdir "$i"`.

fontconfig.yaml
kaniini: This one is fine.

openjdk-11.yaml
kaniini: Yes.

openjdk-11.yaml
```diff
@@ -10,6 +10,10 @@ package:
         - "*"
       attestation:
       license: GPL-2.0-only
+  dependencies:
+    runtime:
+      - freetype-dev
```
kaniini, line 15: However, freetype-dev and fontconfig-dev are runtime dependencies of the JDK?  Why is that?  The runtime depends will get detected automatically by Melange, or should be.

kaniini (changes requested):

openjdk-11.yaml
rawlingsj: I'll double check this in the morning. I was seeing a few instances of
```
libfontmanager.so: libfreetype.so.6: cannot open shared object file: No such file or directory
```
Though I now wonder if that was the JDK or the JRE subpackage.  I'll do a test and make sure.

mkfontscale.yaml
rawlingsj: good catch thanks

openjdk-11.yaml
rawlingsj: Just ran a test removing these runtime dependencies and get the error I was mentioning above:
```
LoadStatisticsTest.graph:94 » UnsatisfiedLink /usr/lib/jvm/openjdk/lib/libfontmanager.so: libfreetype.so.6: cannot open shared object file: No such file or directory
```

rawlingsj: Just testing replacing the runtime dependencies using
```
      - freetype
      - fontconfig
```
rather than
```
      - freetype-dev
      - fontconfig-dev
```

rawlingsj: I've ran some downstream tests and this is working.

kaniini (approved):

PR #104, opened 2022-11-15
add Jenkins package

rawlingsj:


jenkins.yaml
```diff
@@ -0,0 +1,53 @@
...
+  - uses: patch
+    with:
+      patches: ignoreArchiveNotReadableTest.patch
+  - runs: |
+      export JAVA_HOME=/usr/lib/jvm/openjdk
+      export LANG=en_US.UTF-8
+
+      # TODO lets see how long running unit tests in CI takes
+      # mvn install -DskipTests=true
+      mvn install
```
rawlingsj, line 50: let's see how long CI takes when running tests

cpanato (dismissed):
Yeah

rawlingsj:
The duration of these tests is going to be ~2-3 hours I'd guess.  Still I think it's worthwhile keeping at least for now.  Disabling tests doesn't seem like a good step especially as these Jenkins ones have helped catch a few issues during development.

jenkins.yaml
kaniini: is it possible to run the tests in parallel?

rawlingsj: I've kicked off a couple of alternatives playing with the maven `-DforkCount`, will know in a couple of hours if that helps.

jenkins/ignoreArchiveNotReadableTest.patch
```diff
@@ -0,0 +1,22 @@
+# Disabling single test which fails.  Manually checked that Java does set the read only file permissions correctly
+# Requires more investigation, early thoughts could be related to the jenkins test harness coming from maven central and the test not using the one built from source
```
daniel-beck, line 2: Running this as `root` would do this.

rawlingsj: Ah that explains it - many thanks @daniel-beck I was definitely scratching my head on that one :) Out of all the tests in Jenkins this was the only one that failed.

@kaniini do we have any precedence for running steps as different users?

jenkins.yaml
rawlingsj: Good: `-DforkCount=2`  finished in 1h03m in my tests compared with 2h20m without the fork.  

I did try `-DforkCount=$(nproc)` but got failures 
```
Exception: java.lang.NullPointerException thrown from the UncaughtExceptionHandler in thread "TCP agent connection handler #1 with /192.168.10.2:56910"
```
and
```
JNLPLauncherTest.testDelegatingComputerLauncher:129->launchJnlpAndVerify:190 Agent failed to go online
```

trying a last build with `-DforkCount=4`

hectorj2f:
👏

jenkins/ignoreArchiveNotReadableTest.patch
rawlingsj: Just testing using `su-exec` and a different user, looks promising.

kaniini (approved):

jenkins/ignoreArchiveNotReadableTest.patch
kaniini: Not really, we create a throwaway container which is unprivileged on the outside, so I haven't given it much thought.

rawlingsj: Ok thanks - makes sense.  We might need to revisit this and in the meantime keep the patch to ignore that particular test.

PR #110, opened 2022-11-18
wolfi-baselayout: generate os-release file rather than shipping it as a static file

kaniini:
VERSION_ID is intended to reflect the version of the base operating system data in the `/etc/os-release` specification.

Wolfi itself does not have formal versions, and so the base operating system data is versioned using a date code.  However, that date code had to be manually updated whenever the base operating system data was changed.  We can generate the `/etc/os-release` file at build time instead to solve that issue.

Importantly, it should be noted that the "base operating system data" only reflects the `wolfi-baselayout` package.  Inspection of customer image SBOMs or use of an SCA tool such as syft is necessary to understand the composition of image contents (and thus any vulnerability status conclusions that could be drawn from that inspection).

ref chainguard-images/jdk#16.


imjasonh (approved):

amouat (approved):

amouat:
We should also put an explanation of the version number in an FAQ or somewhere.

PR #111, opened 2022-11-18
add ninja pkg

tuananh:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -208,5 +208,7 @@ $(eval $(call build-package,bazel-5,5.3.2-r1))
 $(eval $(call build-package,clang-15,15.0.4-r0))
 $(eval $(call build-package,giflib,5.2.1-r0))
 $(eval $(call build-package,jenkins,2.378-r0))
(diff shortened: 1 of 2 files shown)


kaniini:
We use `samurai` to provide ninja.  Is there a reason why Ninja itself is required?

tuananh:
> 

i'm not well-versed in this topic. i just saw some of hte package required ninja when building. not sure what's the pros & cons of samurai vs ninja.

the top link on this topic is apaprently written by you :D 

https://lists.alpinelinux.org/~alpine/devel/%3CCAAOiGNwnzTomzOQ6YjH3t7VHmbfMRZUNXap8GM0Gy_mGS7aKhg%40mail.gmail.com%3E

kaniini:
I think we should avoid Ninja, personally, as samurai does everything functionally required for building the distribution.  Ninja does however have some tools that samurai does not yet implement, so there may be an argument for providing it in the future.

rawlingsj:
I've been using the `samurai` package when needing `Ninja` and it's been working great.  Let's close this PR and we can always reopen if there's a need.

PR #113, opened 2022-11-19
Bump curl to v7.86.0 to fix CVE-2022-42916.

dlorenc:


curl.yaml
```diff
@@ -11,6 +11,10 @@ package:
       attestation: TODO
       license: MIT
 
+secfixes:
```
kaniini, line 14: 7.86.0-r0

melange.rsa
```diff
@@ -0,0 +1,51 @@
+-----BEGIN RSA PRIVATE KEY-----
```
kaniini, line 1: this shouldn’t be here, or the public key :)

kaniini (changes requested):

curl.yaml
dlorenc: Fixed!

kaniini (approved):

melange.rsa
luhring: Maybe there's a setting or GH action we could use to check for this at some point 🤔  (e.g. gitleaks)

imjasonh: We can add these paths to .gitignore, and grep for PRIVATE KEY in lint.sh

PR #116, opened 2022-11-21
add regclient into wolfi os

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

/cc @sudo-bmitch @dlorenc

PR #121, opened 2022-11-22
use strip to shrink the size of it

developer-guy:

According to the @kaniini recommendations [here](https://github.com/wolfi-dev/os/pull/116#discussion_r1028378901)


kaniini:
`epoch` should be bumped so that the package gets rebuilt.

imjasonh:
I have to admit, I don't entirely understand what `strip` does.

What does it strip from the binary, and how much smaller is it as a result?

Is there an equivalent `go build` flag we should be passing instead?

developer-guy:
PTAL @kaniini

tuananh:
> I have to admit, I don't entirely understand what `strip` does.
> 
> What does it strip from the binary, and how much smaller is it as a result?
> 
> Is there an equivalent `go build` flag we should be passing instead?

i think it's equivalent to `go install -ldflags '-s'`

imjasonh:
> i think it's equivalent to `go build -ldflags '-s'`

In that case, should we just `go build` with those flags? Maybe we need a `go-build` built-in pipeline that sets that flag, and any other common flags we expect to find.

tuananh:
> > i think it's equivalent to `go build -ldflags '-s'`
> 
> In that case, should we just `go build` with those flags? Maybe we need a `go-build` built-in pipeline that sets that flag, and any other common flags we expect to find.

@puerco is already working on that here https://github.com/chainguard-dev/melange/pull/180

but no default ldflags is set.

kaniini (dismissed):

kaniini (approved):

PR #126, opened 2022-11-23
add llvm-libunwind and llvm-lld packages

rawlingsj:


rawlingsj:
I'm doing some more testing and I'm not 100% sure about all the `-D` build flags so a review would be great.

kaniini (approved):

luhring (approved):

PR #129, opened 2022-11-24
add envoy package

rawlingsj:

### DIFF
--- a/envoy.yaml
+++ b/envoy.yaml
@@ -1,0 +1,66 @@
+package:
+  name: envoy
+  version: 1.24.0 # TODO when upgrading to 1.25.0 remove all patches
+  epoch: 0
+  description: Cloud-native high-performance edge/middle/service proxy
+  target-architecture:
+    - all
+  copyright:
+    - paths:
+        - "*"
+      attestation: TODO
+      license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+      - git
+      - bazel-5
+      - openjdk-11
+      - bash
+      - libtool
+      - cmake
+      - samurai
+      - python3-dev
+      - clang-15
+      - libstdc++
+      - llvm15
+      - llvm15-dev
+      - llvm-lld
+      - coreutils
+      - patch
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/envoyproxy/envoy
+      branch: v${{package.version}}
+      destination: envoy
+
+  - runs: |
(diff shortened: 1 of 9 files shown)


envoy.yaml
```diff
@@ -0,0 +1,67 @@
...
+      branch: v${{package.version}}
+      destination: envoy
+
+  - runs: |
+      export JAVA_HOME=/usr/lib/jvm/openjdk
+      mkdir -p /builder/home/.cache/bazel/_bazel_root
+
+      cd envoy
+
+      # these patches are only required for 1.24.0, they are already merged upstream but 1.25.0 is not yet available.
```
rawlingsj, line 49: This is a shame but AFAICT necessary until 1.25.0 is available.  FWIW fetching [envoy](https://github.com/envoyproxy/envoy) `main` branch works with the same config but with ALL patches removed.  So this is untidy for now but we can't depend on a main upstream branch and forking for now doesn't seem sensible either.  So using the patch approach.

kaniini: It makes sense, even if I don't like it.

kaniini (commented):
I see there are a lot of external deps that it fetches and builds.  Is it possible to use the system-provided deps instead?  This would cut down the package size a lot.

rawlingsj:
> I see there are a lot of external deps that it fetches and builds. Is it possible to use the system-provided deps instead? This would cut down the package size a lot.

I'm not yet sure if / how Bazel handles system-provided deps.  FWIW I did just test the binary size after adding the `-strip:` step and the binary came down from 405M to 58M.

rawlingsj:
I suspect we'd need to start patching the bazel rules and take on the maintenance of them which I guess brings in it's own risks, as we're not running the test suites the envoy project is.

kaniini (approved):

kaniini (approved):

rawlingsj:
Maybe there is a way, not too sure TBH.

puerco (approved):

puerco:
The SBOM on this one is going to be a challenge.

dlorenc:
> I'm not yet sure if / how Bazel handles system-provided deps. FWIW I did just test the binary size after adding the `-strip:` step and the binary came down from 405M to 58M.

This is within range of the official build:
image
https://github.com/envoyproxy/envoy/releases/tag/v1.24.0

dlorenc (approved):

dlorenc:
> The SBOM on this one is going to be a challenge.

In general for bazel projects this will be challenging. Envoy has done a lot of work to make this easy though. They require manual CPE declaration for all dependencies: https://github.com/envoyproxy/envoy/blob/main/DEPENDENCY_POLICY.md#declaring-external-dependencies

PR #132, opened 2022-11-25
packages needed for bootstraping rust

Lunarequest:
I'm adding the packages needed for bootstraping rust before I add the rust package itself since it seems redundant to not upstream these very useful packages!


kaniini (approved):

Lunarequest:
Could we get this merged in? I'd like these upstream before I merge in rust

dlorenc (approved):

PR #138, opened 2022-11-29
Rust bootstrap

Lunarequest:
rust bootstrap


kaniini:
`/usr/bin/FileCheck` is provided by `llvm15` package, but that doesn't get pulled in by `llvm15-dev` (which just pulls in the libraries).

Lunarequest:
ah, will add it

Lunarequest:
@kaniini any idea what provides `llvm-c/BitReader.h`?

kaniini:
rust probably needs to call the proper llvm-config binary ;/

kaniini:
llvm has to be rebuilt first, but i don't think that change is correct

Lunarequest:
> llvm has to be rebuilt first, but i don't think that change is correct

What would be the right change?

rust.yaml
```diff
@@ -0,0 +1,124 @@
...
+      export ARCH=${{host.triplet.rust}}
+      cd rustc-${{package.version}}-src
+      ./configure \
+        --host="${ARCH}" \
+        --target="${ARCH},wasm32-wasi,wasm32-unknown-unknown" \
+        --prefix="/usr" \
+        --release-channel="stable" \
+        --enable-local-rust \
+        --local-rust-root="/usr" \
+        --llvm-root="/usr/lib/llvm15" \
```
tuananh, line 72: use `--llvm-config` flag for specifying llvm-config bin path

https://github.com/rust-lang/rust/blob/2682b88c526d493edeb2d3f2df358f44db69b73f/src/bootstrap/configure.py#L98

Lunarequest:
I've got no idea why this is happening, any suggestions would be helpful

kaniini:
Not sure, can dig into it this morning.

tuananh:
> I've got no idea why this is happening, any suggestions would be helpful

i think the problem is with llvm-config still

BitReader.h should be included, when you run llmv-config --cxxflags

it should show something like below, and then that output will be passed to the build command c++


-I/usr/lib/llvm-15/include -std=c++14   -fno-exceptions -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS

can you try doing llvm-config --cxxflags and see what it currently showing.

Lunarequest:
> can you try doing llvm-config --cxxflags and see what it currently showing.

cxxflags outputs the following
```
-I/usr/include -std=c++14   -fno-exceptions -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS
```
cflags the following 
```
 -I/usr/include  -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS
```

Lunarequest:
I fixed the build, I'm going to clean it up before I mark this as ready, I hope the following points can be agreed upon

- development utilities such as rustfmt or clippy won't be included(I don't believe we intend on people using wolfi for development so there's no need for these) 
- no docs(same as above)
- limited support for non-arch builds barring wasm(cross compiling to say aarch64 from x86_64 shouldn't be a thing once arch specific packages start being built. WASM however is an exception as there are use cases eg for javascript libraries)

Lunarequest:
looks like some minor fixes are needed, but it builds which is a lot more then before

tuananh:
> -I/usr/include

this is why it was failing in the first place i guess.

Lunarequest:
I've taken ill, could someone take over. The issue has to do with wasi-libc
https://github.com/rust-lang/rust/blob/897e37553bba8b42751c67658967889d11ecd120/src/bootstrap/compile.rs#L271-L280

Lunarequest:
this builds without wasm, I have no idea how to get wasm to build. If its needed I've got no idea how to get it to build

Lunarequest:
@kaniini since it was originally assigned to you, do you have any ideas how to get wasm to build?

kaniini (approved):

local-melange.rsa
```diff
@@ -0,0 +1,51 @@
+-----BEGIN RSA PRIVATE KEY-----
```
kaniini, line 1: Oh, this needs to be removed...

local-melange.rsa.pub
```diff
@@ -0,0 +1,14 @@
+-----BEGIN PUBLIC KEY-----
```
kaniini, line 1: As does this...

kaniini (dismissed):

patflynn:
Just holding on merging until post-ARM boostrap. Should be no more than a few days.

Lunarequest:
There seems to be a issue with wget not the rust package :/

dlorenc:
Hey @Lunarequest - I got a copy of this branch rebased and fixed up over here: https://github.com/wolfi-dev/os/compare/main...dlorenc:os:rust-bootstrap?expand=1

It looks like the issue was the version constraint you had on rust-stage0. It might have been the wrong syntax (=< instead of >=), but also it was unnecessary. That triggered some bug somewhere which led to the busybox version of wget appearing, instead of the normal one, which led to the strange failure mode with dns-timeout.

I can send that one and shepherd it in (with your commits included), or feel free to grab my commit and move it over here!

dlorenc:
Looks right now - I think you just have to rebase to fix up the Makefile!

dlorenc:
Ah whoops, yeah I also had to remove tar. It comes in as part of one of the other packages.

Lunarequest:
Ahh will do.

dlorenc:
It's alive!!!

dlorenc (approved):

imjasonh (approved):

PR #153, opened 2022-12-01
Prepare packages for wolfictl update generates PRs.  These package fe…

rawlingsj:
…tch URLs need to change to support future versions


pax-utils.yaml
```diff
@@ -30,7 +26,7 @@ environment:
 pipeline:
   - uses: fetch
     with:
-      uri: https://dev.gentoo.org/~xen0n/distfiles/pax-utils-${{package.version}}.tar.xz
+      uri: https://gitweb.gentoo.org/proj/pax-utils.git/snapshot/pax-utils-${{package.version}}.tar.gz
```
rawlingsj, line 29: `1.3.5` is now also available but not under the original uri above.  Moving to this uri means we can automate future version updates.

pax-utils.yaml
```diff
@@ -27,7 +27,7 @@ pipeline:
   - uses: fetch
     with:
       uri: https://gitweb.gentoo.org/proj/pax-utils.git/snapshot/pax-utils-${{package.version}}.tar.gz
-      expected-sha256: 8baed2f9c5ae8e0cda1b9c75990864101afc64fad0a4616e10f3ff8ef891040b
+      expected-sha256: 179f35678a0472e32315fc8725c62a6be9549f8765178b26486851c4bbe54590
```
rawlingsj, line 30: Not sure why the sha is different if we are fetching the same tar ball just using a different location.  Is this something that needs to be looked at?

PR #156, opened 2022-12-02
lets include an optional default config that matches the UX of upstre…

rawlingsj:
…am image


envoy.yaml
```diff
@@ -63,5 +63,23 @@ pipeline:
...
+subpackages:
+  - name: envoy-oci-entrypoint
+    description: Entrypoint for using Envoy in OCI containers
+    dependencies:
+      runtime:
+        - busybox
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/var/lib/envoy/init
+          cp envoy-entrypoint.sh ${{targets.subpkgdir}}/var/lib/envoy/init/
```
kaniini, line 78: is this part of envoy sources?

rawlingsj: great point and catch! yes it is :) will update the PR

rawlingsj: I'd missed the `git add --all` to add the files in the `./envoy` source-dir in the previous commit, in case it was a bit confusing.  I've switched the references to used the files from the fetched source instead.

kaniini (approved):

jdolitsky (approved):

Issue #86, opened 2022-11-04
Enforce best practices for melange YAMLs


imjasonh:
- checked-in packages shouldn't depend on `packages.wolfi.dev/os` or include its keyring; packages are fetched from that repo and depended on locally during the build
- every melange YAML should have a corresponding entry in `Makefile` (would have caught https://github.com/wolfi-dev/os/pull/84)
- maybe harder to catch without involving `dag`: no duplicate package or subpackage names (would have caught https://github.com/wolfi-dev/os/pull/85)
- less importantly, consistently format YAML (mostly to avoid formatting arguments)
- more generally, melange could fail if it encounters an unknown field -- this would have caught the `epcho` bug fixed in #84

These checks should be enforced at CI time to prevent bugs from being committed.

This is complicated somewhat by the fact that melange YAML files aren't actually parseable YAML files, they're templates that get Go-templated into parseable YAML. Maybe we could templateize them then YAML-format them, or something else smart we haven't thought of yet.

kaniini:
I have some thoughts based on updating `wolfi-secdb`, which caught a few things in #81.  I'm probably going to add a `melange lint` action, which can check some of these things, and then we can extend that into a wolfi-specific linting package.

Dentrax:
> I'm probably going to add a melange lint action

Any ongoing work on this? I'd be nice to have a tracking issue for this in the melange side. If you split into smaller tasks, we are ready to help! @developer-guy 

Additional to @.imjasonh ideas, here are some other ideas:

* smoke test first: check if given file a valid `build.Configuration` struct
* check if `copytight.license` is correct (external api call needed to check LICENSE file, handling just GitHub would be fine in the first place)
* traverse all `target-architecture`, ensure melange have a support for each
* enforce `package.version` in semantic-versioning format
* `pipeline.with.uri` is correct URI format (not sure how to handle `${{ foo }}` stuff)
* `pipeline.with.expected-sha256` is correct sha256 format (do same for other algorithms)

developer-guy:
It should be a separate command in the [wolfictl](https://github.com/wolfi-dev/wolfictl) CLI

kaniini:
@Dentrax says he wants to work on this, so assigning it to him!

rawlingsj:
@Dentrax have you started working on this?  I've got a need for a CI check so am thinking it best live in `wolfictl lint`.  If I make a start perhaps you can add extra lints to it once available?

Dentrax:
[...]

PR #163, opened 2022-12-05
wait-for-it pkg

developer-guy:


kaniini:
It would be better to get sources from a remote place rather than include the sources in the yaml file.

wait-for-it.yaml
```diff
@@ -0,0 +1,27 @@
...
+  contents:
+    packages:
+      - busybox
+      - bash
+      - ca-certificates-bundle
+      - build-base
+pipeline:
+  - uses: fetch
+    with:
+      uri: https://raw.githubusercontent.com/vishnubob/wait-for-it/master/wait-for-it.sh
```
kaniini, line 23: This should be pinned to a tagged version rather than `master` (and `package.version` should match the tagged version).  If you are taking a snapshot from git, you should specify the commit hash instead of `master`.

Dentrax: _off topic and nit: just add this concern as a check item in the [linter tracking issue](https://github.com/wolfi-dev/os/issues/86), so we don't forget!_

wait-for-it.yaml
```diff
@@ -0,0 +1,27 @@
...
+  description: Pure bash script to test and wait on the availability of a TCP host and port
+  target-architecture:
+    - all
+  copyright:
+    - paths:
+        - "*"
+      attestation: TODO
+      license: MIT
+environment:
+  contents:
```
developer-guy, line 14: @marcofranssen, do you think that we should include other packages to make wait-for-it work?

wait-for-it.yaml
developer-guy: I've updated the version information according to your feedbacks @kaniini, thx for your help

wait-for-it.yaml
```diff
@@ -0,0 +1,27 @@
...
+  version: 0_git200823
+  epoch: 0
+  description: Pure bash script to test and wait on the availability of a TCP host and port
+  target-architecture:
+    - all
+  copyright:
+    - paths:
+        - "*"
+      attestation: TODO
+      license: MIT
```
imjasonh, line 12: Does this have a runtime dependency on bash too?

marcofranssen: yes it does, as the shell script requires some bash things.

wait-for-it.yaml
```diff
@@ -0,0 +1,27 @@
...
+environment:
+  contents:
+    packages:
+      - busybox
+      - bash
+      - ca-certificates-bundle
+      - build-base
```
marcofranssen, line 19: Why do we need build-base?

PR #244, opened 2022-12-09
add jenkins-remoting & jenkins-inbound-agent package

tuananh:
this pr add 2 packages: jenkins-remoting & jenkins-inbound-agent

related pr: https://github.com/chainguard-images/images/pull/130

### DIFF
--- a/jenkins-inbound-agent.yaml
+++ b/jenkins-inbound-agent.yaml
@@ -1,0 +1,30 @@
+package:
+  name: jenkins-inbound-agent
+  version: 3077.vd69cf116da_6f-3
+  epoch: 0
+  description: Jenkins inbound agent
+  target-architecture:
+    - all
+  copyright:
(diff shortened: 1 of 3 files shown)


jenkins-inbound-agent.yaml
```diff
@@ -0,0 +1,30 @@
+package:
+  name: jenkins-inbound-agent
+  version: 3077.vd69cf116da_6f-3
```
kaniini, line 3: Version is invalid.  https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-260003.3 outlines what can and cannot be used in a version string.

luhring: That's a good question, we'll wait for @kaniini to get back to this. We'd want to make sure we produce a version that can be used downstream, e.g. in vuln matching.

PR #265, opened 2022-12-13
libarchive/3.6.2 package update

rawlingsj:
<p align="center">
  
</p>


rawlingsj:
need to review and remove current patches as we have a new version including those patches

dlorenc:
This one should be good to go right?

kaniini:
We need to remove the patches, it fails to build right now because it is patching things which are already patched.

dlorenc (approved):

puerco (approved):

PR #269, opened 2022-12-14
fix runtime deps by adding coreutils

developer-guy:
thx @rawlingsj for figuring this out!

### DIFF
--- a/wait-for-it.yaml
+++ b/wait-for-it.yaml
@@ -1,7 +1,7 @@
 package:
   name: wait-for-it
   version: 0.20200823
-  epoch: 0
+  epoch: 1
   description: Pure bash script to test and wait on the availability of a TCP host and port
   target-architecture:
     - all
@@ -13,6 +13,7 @@ package:
   dependencies:
     runtime:
       - bash
+      - busybox
 environment:
   contents:
     packages:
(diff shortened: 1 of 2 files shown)


rawlingsj:
I've tested this locally and seems to work.

rawlingsj (dismissed):

marcofranssen:
Resolves https://github.com/chainguard-images/images/issues/136

marcofranssen (dismissed):

marcofranssen:
Ready to test another release https://github.com/philips-labs/helm-charts/pull/80/files

rawlingsj:
Looks like the script needs busybox here https://github.com/vishnubob/wait-for-it/blob/master/wait-for-it.sh#L144-L149

rawlingsj:
@developer-guy so it turns out we can just use `coreutils` instead of adding `busybox` would you mind changing this PR and we'll get it merged in

wait-for-it.yaml
```diff
@@ -13,6 +13,7 @@ package:
   dependencies:
     runtime:
       - bash
+      - busybox
```
rawlingsj, line 16: this can be `coreutils` which avoids busybox

amouat (approved):

developer-guy:
Can we merge this one, @kaniini, because it will fix the wait-for-it OCI image? We have a bug right now 😩

kaniini (approved):

PR #279, opened 2022-12-19
Add vault package

tuananh:
build without ui because node 16 is required for building ui and we only got node 18 here.

more specifically, it use node-sass 7.0.3 and that module doens't have native prebuilt binaries for Node 18 (ABI 108) https://github.com/sass/node-sass/releases/tag/v7.0.3

so it fallbacks to rebuild and a couple of problems i haven't figured out (we need python, gcc, nodejs and couple more). Will attempt to fix this in another PR.


luhring:
@tuananh Could you clarify what you mean by this?

> we need python, gcc, nodejs and couple more

tuananh:
> @tuananh Could you clarify what you mean by this?
> 
> > we need python, gcc, nodejs and couple more

Vault's frontend component is a Node.js project. They have a build dependency on node-sass which is a native Node.js module.

Usually, project will provide native prebuilt binaries on GitHub releases for each combination of OS and Node ABI version.

The node-sass dependency version is 7.0.3 and that specific releases doesn't have prebuilt binary for Node version 18 (The one we have in Wolfi). It doesn't make sense to add Node 16 to Wolfi either because it's EOL this month.

You can see the list of prebuilt binary for this 7.0.3 version here: https://github.com/sass/node-sass/releases/tag/v7.0.3

What we can do is:
- Add Nodejs 16 to Wolfi and use that to build vault UI
- Or open a PR in Vault to upgrade node-sass dependency.
- Or build Vault package without UI component and wait for them to fix it upstream.

luhring:
> What we can do is:
> 
> * Add Nodejs 16 to Wolfi and use that to build vault UI. When Node 16 is detected, they will download the prebuilt binary for node-sass and proceed with build process.
> * Or open a PR in Vault to upgrade node-sass dependency.
> * Or build Vault package without UI component and wait for them to fix it upstream.

Thanks, @tuananh, this is helpful! I think it's fine to build the Vault package **without the UI** for now. And opening a PR to upgrade node-sass also sounds like a good idea — but we don't need to block this PR for that.

luhring (approved):

dlorenc (approved):

tuananh:
Submit another PR since we can build ui component now https://github.com/wolfi-dev/os/pull/931

PR #288, opened 2022-12-20
add CRON jobs to periodically check packages for new versions

rawlingsj:
This PR add's two CRON jobs because depending on which service we use to check for new packages the updater performs differently.

When using GitHub to check for updates the updater uses a graphql API and a single query compared with https://release-monitoring.org/ which needs to call their API for every package we are checking.  Therefore, checking via GitHubs API can be done more frequently.

The CRONS from this PR  will query GitHub every hour and https://release-monitoring.org/ daily.

### DIFF
--- a/.github/workflows/wolfictl-update-gh.yaml
+++ b/.github/workflows/wolfictl-update-gh.yaml
@@ -1,0 +1,24 @@
+name: Wolfictl Update From GitHub
+
+on:
+  # Triggers the workflow every hour
+  schedule:
+    - cron: "0 * * * *"
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  lint:
+    name: Wolfictl Update
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: wolfi-dev/wolfictl/.github/actions@main
+    - run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/wolfictl-update-gh.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  lint:
+    name: Wolfictl Update
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: wolfi-dev/wolfictl/.github/actions@main
+    - run: |
+        git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
luhring, line 20: Where does this email address value come from?

rawlingsj: it means we get a nice github actions bot name making the PR...

Screenshot 2022-12-19 at 21 52 42

luhring: Okay nice! May be worth a comment to explain the value, but consider that just a nit from me

luhring (approved):
Very exciting!

jdolitsky (approved):

PR #296, opened 2022-12-21
dag: add production build

kaniini:
This allows us to build and release arm64 packages into production.

### DIFF
--- a/.github/workflows/dag-push-production.yaml
+++ b/.github/workflows/dag-push-production.yaml
@@ -1,0 +1,177 @@
+name: Build Packages Production using dag
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+
+env:
+  PROJECT: prod-images-c6e5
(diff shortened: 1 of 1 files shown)


.github/workflows/dag-push-production.yaml
```diff
@@ -0,0 +1,177 @@
...
+    runs-on: ubuntu-latest
+
+    permissions:
+      id-token: write
+      contents: read
+
+    steps:
+      - uses: google-github-actions/auth@v0
+        with:
+          workload_identity_provider: "projects/618116202522/locations/global/workloadIdentityPools/prod-shared-e350/providers/prod-shared-gha"
```
luhring, line 30: It may be worth pulling this value out into `env:` since we use it a few times

imjasonh (approved):
🚀🚀🚀🚀

rawlingsj (approved):

PR #303, opened 2022-12-24
Add a Redis package.

dlorenc:
This applies two patches, one was taken from the Alpine redis distro (redis.confd.patch), and the other was taken from the Docker image of redis, which is required to disable protected mode. Protected mode prevents redis from listening on an external port by default, but we need that capability in containers.


dlorenc:
I'm not 100% sure we need the redis.confd.patch file...

rawlingsj (dismissed):

imjasonh (approved):

redis/0000-Disable-protected-mode.patch
```diff
@@ -0,0 +1,29 @@
+From b91e39701c7a71f229be34dcf078579d37367436 Mon Sep 17 00:00:00 2001
+From: Dan Lorenc <dlorenc@chainguard.dev>
+Date: Sat, 24 Dec 2022 12:15:14 -0500
+Subject: [PATCH] Disable protected mode.
+
+See https://github.com/docker-library/redis/blob/19f345b2056588c009396bcfc22ba9241a669491/Dockerfile-alpine.template#L43
+for more context - we disable protected mode because it's not needed
+in a docker image and we can't do this via a config option.
```
puerco, line 8: Thanks for the context here :+1:

puerco (approved):

PR #306, opened 2022-12-25
Add a go-1.20rc1 package.

dlorenc:
This doesn't actually change the existing go 1.19 build, should we switch that to be the same style of multi-version packages we use for ruby and postgres?


rawlingsj:
Yeah, I think having the major version in the package name makes sense.  I wonder if we'll need to do something in the new year to switch the current go package to go-19.

rawlingsj (approved):

imjasonh:
> Yeah, I think having the major version in the package name makes sense. I wonder if we'll need to do something in the new year to switch the current go package to go-19.

+1. I think we should also have a `go` package that's just a metapackage pointing to whatever version of `go` is the latest.

imjasonh (approved):

dlorenc:
> +1. I think we should also have a `go` package that's just a metapackage pointing to whatever version of `go` is the latest.

Ooooh I like this, do you have an example?

PR #309, opened 2022-12-26
Add etcd package.

dlorenc:


etcd.yaml
```diff
@@ -0,0 +1,36 @@
...
+    packages:
+      - ca-certificates-bundle
+      - busybox
+      - git
+      - bash
+      - go
+pipeline:
+  - uses: fetch
+    with:
+      uri: https://github.com/coreos/etcd/archive/v${{package.version}}.tar.gz
```
tuananh, line 26: nitpick: shouldn't it be https://github.com/etcd-io/etcd? coreos/etcd redirect to the new one anyway

etcd.yaml
```diff
@@ -0,0 +1,36 @@
...
+      expected-sha256: 0ef11c0cdc3cc693756a636d97d280ea63e222a7be38159fad0c8032ced0da42
+  - runs: |
+      bash -x ./build.sh
+      mkdir -p "${{targets.destdir}}"/var/lib/$pkgname
```
rawlingsj, line 30: `$pkgname` can change to the melange templated value `${{package.name}}` or just hard code `etcd`

etcd.yaml
dlorenc: Fixed!

etcd.yaml
dlorenc: Fixed!

imjasonh: I've definitely made this mistake before. We should add a check for this in linting.

imjasonh (approved):

jedsalazar (approved):

puerco (approved):

PR #315, opened 2022-12-28
Add libLLVM-15 as a runtime dependency for rust.

dlorenc:
This is needed to run rustc.


rust.yaml
```diff
@@ -22,7 +23,6 @@ environment:
       - busybox
       - ca-certificates-bundle
       - build-base
-      - rust-stage0
```
imjasonh, line 25: Don't we still need this?

dlorenc: Yeah, it was actually duplicated. See line 38/39.

imjasonh: That just means we _really_ need it.

imjasonh (approved):

puerco (approved):

PR #323, opened 2022-12-28
Add libmaxminddb.

dlorenc:
This is required for nginx.


puerco (dismissed):

libmaxminddb.yaml
```diff
@@ -0,0 +1,48 @@
...
+        - uses: split/dev
+      dependencies:
+        runtime:
+            - libmaxminddb
+            - curl
+      description: libmaxminddb dev
+    - name: libmaxminddb-doc
+      pipeline:
+        - uses: split/manpages
+      description: libmaxminddb manpages
```
rawlingsj, line 48: I wonder if we also need a libs subpackage like alpine does?
i.e.
```
    - name: libmaxminddb-libs
      pipeline:
        - runs: |
            mkdir -p ${{targets.subpkgdir}}/usr/lib
            mv ${{targets.destdir}}/usr/lib/libmaxminddb.so.* ${{targets.subpkgdir}}/usr/lib/
      description: libmaxminddb libs
```

dlorenc: Added!

rawlingsj (approved):

puerco (approved):

PR #328, opened 2022-12-29
Add tiff.

dlorenc:
This is required for nginx.

Signed-off-by: Dan Lorenc <dlorenc@chainguard.dev>

PR #327, opened 2022-12-29
Add libwebp.

dlorenc:
This is needed for nginx.

PR #321, opened 2022-12-28
Add zeromq package.

dlorenc:
This is required for nginx.

Signed-off-by: Dan Lorenc <dlorenc@chainguard.dev>

PR #333, opened 2022-12-30
Add nasm, dav1d, aom, gd, libavif and libxpm

rawlingsj:
Remaining dependencies needed for Nginx

### DIFF
--- a/aom.yaml
+++ b/aom.yaml
@@ -1,0 +1,59 @@
+package:
+  name: aom
+  version: 3.5.0
+  epoch: 0
+  description: Alliance for Open Media (AOM) AV1 codec SDK
+  target-architecture:
+    - all
+  copyright:
+    - paths:
+        - "*"
+      attestation: TODO
+      license: BSD-2-Clause AND custom
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+      - automake
+      - autoconf
+      - cmake
+      - perl
+      - python3
+      - yasm
+      - samurai
+      - tree
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://aomedia.googlesource.com/aom
+      branch: v${{package.version}}
+      destination: aom
+  - runs: |
+      cd aom
+      cmake -B build -G Ninja \
+        -DCMAKE_INSTALL_PREFIX=/usr \
+        -DCMAKE_INSTALL_LIBDIR=lib \
+        -DBUILD_SHARED_LIBS=True \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DENABLE_TESTS="OFF"
+      ninja -C build
+      DESTDIR=${{targets.destdir}} ninja -C build install
+  - uses: strip
+  - runs: tree ${{targets.destdir}}
+subpackages:
+  - name: aom-dev
+    pipeline:
+      - uses: split/dev
+    dependencies:
+      runtime:
+        - aom
+    description: aom dev
+
+  - name: aom-libs
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/usr/lib
+          mv ${{targets.destdir}}/usr/lib/libaom.so.* ${{targets.subpkgdir}}/usr/lib/
+    description: aom libs
--- a/gd.yaml
+++ b/gd.yaml
@@ -21,7 +21,7 @@ environment:
       - fontconfig-dev
       - freetype-dev
       - libavif-dev
-      - libjpeg-turbo-dev
+      - libjpeg-dev
       - libpng-dev
       - libwebp-dev
       - libxpm-dev
--- a/gd.yaml
+++ b/gd.yaml
@@ -1,0 +1,60 @@
+package:
+  name: gd
+  version: 2.3.3
+  epoch: 0
+  description: Library for the dynamic creation of images by programmers
+  target-architecture:
+    - all
+  copyright:
+    - paths:
+        - "*"
+      attestation: TODO
+      license: custom
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+      - automake
+      - autoconf
+      - fontconfig-dev
+      - freetype-dev
+      - libavif-dev
+      - libjpeg-turbo-dev
+      - libpng-dev
+      - libwebp-dev
+      - libxpm-dev
+      - tiff-dev
+      - zlib-dev
+pipeline:
+  - uses: fetch
+    with:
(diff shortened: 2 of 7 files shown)


rawlingsj:
Needs...

https://github.com/wolfi-dev/os/pull/321
https://github.com/wolfi-dev/os/pull/323
https://github.com/wolfi-dev/os/pull/327
https://github.com/wolfi-dev/os/pull/328

dlorenc:
Nice! I punted on nasm since yasm seems to be a fine substitute, but it's probably worth having both for compatibility. I'd suggest trying to use yasm first though.

rawlingsj:
> Nice! I punted on nasm since yasm seems to be a fine substitute, but it's probably worth having both for compatibility. I'd suggest trying to use yasm first though.

Ok yeah good call!  Will do.

rawlingsj:
I'm not yet able to test locally if we can only use yasm.  Recently started getting some build errors which the GHA on this PR is failing with too.

Not too sure what's up yet.  There was a python upgrade recently so possibly related?  I [noticed when there's a go upgrade](https://github.com/wolfi-dev/os/pull/176/files) we do an epoch update for downstream dependencies, perhaps python is the same?

Build error:
```
  File "/usr/lib/python3.11/importlib/metadata/__init__.py", line 563, in from_name
    return next(cls.discover(name=name))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
StopIteration

During handling of the above exception, another exception occurred:

  File "/usr/bin/meson", line 33, in <module>
    sys.exit(load_entry_point('meson==0.64.1', 'console_scripts', 'meson')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/bin/meson", line 22, in importlib_load_entry_point
    for entry_point in distribution(dist_name).entry_points
                       ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/importlib/metadata/__init__.py", line 981, in distribution
    return Distribution.from_name(distribution_name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/importlib/metadata/__init__.py", line 565, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: No package metadata was found for meson
```

dlorenc:
> Not too sure what's up yet. There was a python upgrade recently so possibly related? I [noticed when there's a go upgrade](https://github.com/wolfi-dev/os/pull/176/files) we do an epoch update for downstream dependencies, perhaps python is the same?

I thought with go the issue was that we had to do that to get them all to rebuild with go, which we wanted to do because there was a vulnerability in go itself. My (probably naive) understanding is that since python isn't "built", it's closer to dynamic linking - we don't need to bump every package that uses glibc when we update glibc.

rawlingsj:
Yeah that makes sense. I was having a very wild guess as not quite sure what's causing the build problem.

dlorenc:
> Yeah that makes sense. I was having a very wild guess as not quite sure what's causing the build problem.

it's definitely possible some of these libraries don't work on the new version of python though...

dlorenc:
I think this might help: https://github.com/wolfi-dev/os/pull/335

dlorenc (dismissed):

rawlingsj:
I "think" the last CI failure was because the release pipeline didn't complete properly because of two merges in close succession.  Going to retrigger a couple of GHA's.  They'll be more robust over time.

rawlingsj:
Ok got past the python related issues but latest CI failure is related to thw `nasm` change 

```
2022/12/30 22:02:28 melange (dav1d/x86_64): Compiler for C supports arguments -mpreferred-stack-boundary=6: YES 
2022/12/30 22:02:28 melange (dav1d/x86_64): Fetching value of define "__AVX__" :  
2022/12/30 22:02:28 melange (dav1d/x86_64): Program nasm found: NO
2022/12/30 22:02:28 melange (dav1d/x86_64): 
2022/12/30 22:02:28 melange (dav1d/x86_64): meson.build:412:4: ERROR: Program 'nasm' not found or not executable
```

Think I'm going to add nasm back in and see if that helps

dlorenc:
Ah weird, I thought it worked for me. Oh well, sorry!

rawlingsj:
No worries!  It was definitely worth a try and TBH I'm going to double check in the AM but wanted to see if this goes green.

gd.yaml
```diff
@@ -0,0 +1,60 @@
...
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+      - automake
+      - autoconf
+      - fontconfig-dev
+      - freetype-dev
+      - libavif-dev
+      - libjpeg-turbo-dev
```
dlorenc: Needs to change to libjpeg-dev

libavif.yaml
```diff
@@ -0,0 +1,56 @@
...
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+      - automake
+      - autoconf
+      - aom-dev
+      - cmake
+      - dav1d-dev
+      - libjpeg-turbo-dev
```
dlorenc: And here

rawlingsj:
yay - green!

dlorenc (approved):

aom.yaml
```diff
@@ -0,0 +1,59 @@
...
+  version: 3.5.0
+  epoch: 0
+  description: Alliance for Open Media (AOM) AV1 codec SDK
+  target-architecture:
+    - all
+  copyright:
+    - paths:
+        - "*"
+      attestation: TODO
+      license: BSD-2-Clause AND custom
```
puerco, line 12: Mmh this package has a special license to control how patents are used. In the SBOM you can add a custom license ref, but I'm not sure how to handle it in the apk, which will then flow back to the SBOM but without the context.

I think we will need to do an audit of all licensing and copyright entries in the packages to ensure we are keeping the notices and expressions correct.

puerco (approved):
OK, LGTM. Just one note about the license in one of them :point_down:

aom.yaml
rawlingsj: Sounds good to me!  Raised an issue to track https://github.com/wolfi-dev/os/issues/340

rawlingsj: @puerco are you thinking we need to resolve this before merging or would the audit be enough as a followup?

dlorenc: I think we can clean it up after.

puerco: @rawlingsj no I dind't mean to hold this one. I'm sure we have a bunch of inconsistent license entries, so one more will not change that. But we need to check them soon as the number of packages grows.

PR #337, opened 2022-12-30
Add multiple Python streams

### LINKED ISSUE #78: Support multiple python 3.x streams
It would be incredibly useful to have some way of having packages for each currently supported python 3 stream, so currently I think

- 3.11
- 3.10
- 3.9
- 3.8
- 3.7 (though I think this just has 7 months or so left)

I saw https://github.com/wolfi-dev/os/pull/75 and wondered if a similar approach is possible; however, python seems a bit more complicated since there are also a bunch of py3-* packages that'd probably have to be split out as well since they currently have runtime dependencies on `python3`

daewok:
Closes #78

Adds `python-3.11` and `python-3.10` packages. `python3` becomes a virtual package for `python-3.11`

Signed-off-by: Eric Timmons <etimmons@asylonrobotics.com>


daewok:
I'm trying to bring wolfi/melange/apko into my company and ran into this. Figured I try and crank it out.

I don't have a great dev environment set up for this yet, so it's only very lightly tested. I can also add more python versions if desired.

dlorenc:
Awesome! Thanks!

dlorenc:
We ran into some trouble earlier today with some of the bootstrap packages relying on a hardcoded version of python, so this might take a little longer to sort out, so please bear with us!

daewok:
Rebased to hopefully fix the CI issue.

Some questions I had about style: 

- Do you prefer the packages to be named `python-$VERSION` or `python$VERSION`? I see both styles in the repo.
- In #78, I see it was suggested to make a virtual package named `system-python`. Is that preferred? I made `python3` be the virtual package to minimize any breakage for folks already using it.
- Should the virtual package inherit the PSF-2.0 license? Or should I switch it to something like MIT?

kaniini:
We can just do something like:

```yaml
package:
  ...
  dependencies:
    provides:
      - system-python=3.11
```

And then packages like `meson` and the `py3-*` packages can depend on `system-python` and probably `system-python-dev`.

We should maybe do this for Ruby too if we start shipping gems to support Ruby-based tools.

kaniini:
> Do you prefer the packages to be named python-$VERSION or python$VERSION? I see both styles in the repo.

We are moving to `python-$VERSION` format.

daewok:
Nice, I didn't know about the provides keyword!

I just pushed a change that moves python3 to python-3.11 and provides system-python. I switched everything currently depending on python3 to system-python.

Makefile
```diff
@@ -53,7 +53,9 @@ $(eval $(call build-package,mpc,1.2.1-r2))
 $(eval $(call build-package,isl,0.24-r2))
 $(eval $(call build-package,zlib,1.2.13-r1))
 $(eval $(call build-package,flex,2.6.4-r2))
-$(eval $(call build-package,glibc,2.36-r3))
+$(eval $(call build-package,python-3.11,3.11.1-r0))
```
daewok, line 56: I'm not happy about moving this higher in the Makefile, but it was needed since glibc now depends on system-python to build.

py3-setuptools.yaml
```diff
@@ -29,13 +29,13 @@ environment:
...
       - build-base
-      - python3
+      - system-python
       - py3-appdirs
       - py3-more-itertools
       - py3-ordered-set
       - py3-packaging
       - py3-parsing
-      - py3-setuptools
+      - py3-setuptools-stage0
```
daewok, line 38: Did this otherwise there are package conflicts during the CI build. The already built `py3-setuptools` depends on `python3`, which overwrites files provided by `system-python` and makes apk unhappy.

rawlingsj:
@daewok I've got a similar need to this PR, would you mind if I try and help push this through today?  I could cherry-pick your commit and make any changes needed but you'll still have the contribution recognised.

rawlingsj:
@daewok I've cherry picked your commits, resolved conflicts and made the necessary changes to other packages that have landed since you originally created this PR.  Work is being done in this PR https://github.com/wolfi-dev/os/pull/545

If you don't mind I'm going to close this PR to avoid confusion but as mentioned your commits are moved over so you will be credited to the contribution.

PR #339, opened 2022-12-31
add flux CLI

developer-guy:


developer-guy:
PTAL @stefanprodan

flux.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - bash
+      - curl
+      - go
+pipeline:
+  - uses: fetch
+    with:
+      uri: https://github.com/fluxcd/flux2/archive/v${{package.version}}/v${{package.version}}.tar.gz
+      expected-sha256: 6f64ac68748137816e85b2258681782951b04135d00d8e69e0efa96438f47177
+  - runs: |
+      curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
```
dlorenc, line 27: We should probably package customize instead of curling it in here.

flux.yaml
```diff
@@ -0,0 +1,34 @@
...
+      echo "The current working directory is $(pwd)"
+      mkdir -p "${{targets.destdir}}"/usr/bin
+      install -Dm755 $(pwd)/kustomize /usr/bin/kustomize
+      echo "The kustomize version is $(kustomize version)"
+      make build
```
stefanprodan, line 32: We need to set the `VERSION` arg for `make build` to produce a valid binary, otherwise people can’t use it to bootstrap Flux.

developer-guy:
We can wait until https://github.com/wolfi-dev/os/pull/346 merges to use customize as an apk package.

flux.yaml
rawlingsj: kustomize has been added to wolfi now https://github.com/wolfi-dev/os/blob/main/kustomize.yaml so yeah we should try adding it as an `environment.contents.package`

developer-guy: yep, I've added kustomize to the package list.

developer-guy:
PTAL @stefanprodan @rawlingsj

rawlingsj:
LGTM though will let @stefanprodan confirm - BTW hey @stefanprodan!  I think our paths have crossed it the past, hope all is well :)

stefanprodan:
Hey James 👋  We've met at at KubeCon some years ago, I think it was in Copenhagen :)

stefanprodan (commented):
@developer-guy can we add a test please and check that `flux version --client` returns the expected string `flux: v${{package.version}}`

developer-guy:
> @developer-guy can we add a test please and check that `flux version --client` returns the expected string `flux: v${{package.version}}`

I don't think that we have a features like that PTAL @rawlingsj

rawlingsj:
You could add a pipeline step...
```
  - check
    runs: |
       ${{targets.destdir}}/usr/bin/flux | grep "flux: v${{package.version}}"
  - uses: strip
```

I wonder if something that would work?  You could try that or I can try it in the morning.

developer-guy:
yeah, great suggestion, added that check!

flux.yaml
```diff
@@ -0,0 +1,33 @@
...
+environment:
+  contents:
+    packages:
+      - ca-certificates-bundle
+      - busybox
+      - build-base
+      - bash
+      - curl
+      - go
+      - kustomize
```
rawlingsj, line 21: I was wondering if kustomize is needed at runtime rather than build time?  

Adding it here will mean it's only used in the build environment when compiling but figured if a go app needs a library at compile time it would likely use go modules or something.  So am wondering if there's a runtime dependency on kustomize for the flux CLI?  If that's the case we'd need to move this up to the package section.  

i.e.
```
package:
  name: flux
  version: 0.38.2
  epoch: 0
  description: Open and extensible continuous delivery solution for Kubernetes. Powered by GitOps Toolkit.
  target-architecture:
    - all
  copyright:
    - license: Apache-2.0
      paths:
        - "*"
   dependencies:
     runtime:
      - kustomize
```

If it's used at build / compile time, ignore me :)

stefanprodan: We embed the Flux CRDs and controllers manifests inside the binary so people can run the install behind the firewall. These manifests are pulled from each Flux controller repo and we use kustomize remote bases to do that. The kustomize binary is only needed at build time, Flux the CLI has no runtime deps as it doesn't shell exec to anything.

stefanprodan:
I've searched the logs for the test result and it did run successfully, thanks! But reading more of the logs, it seems that the Flux CLI is built only for x86_64 while we provide binaries for amd64/arm64/armv7. Is Wolfi available only on amd64?

rawlingsj:
Yeah currently we only build x86 in CI on PRs but the release pipelines also builds aarch64.

flux.yaml
rawlingsj: awesome - makes sense, thanks for clearing that up

stefanprodan (dismissed):
LGTM

Thanks @developer-guy and @rawlingsj

rawlingsj (dismissed):

rawlingsj (approved):

amouat (approved):

PR #341, opened 2022-12-31
Remove duplicated package entry.

dlorenc:
I built a lint rule for this as well over here: https://github.com/wolfi-dev/wolfictl/pull/14

### DIFF
--- a/rust.yaml
+++ b/rust.yaml
@@ -33,7 +33,6 @@ environment:
       - llvm15-dev
       - clang-15
       - openssl-dev
-      - libssh2-dev
       - python3
       - rust-stage0
       - wasi-libc


rust.yaml
```diff
@@ -33,7 +33,6 @@ environment:
       - llvm15-dev
       - clang-15
       - openssl-dev
-      - libssh2-dev
```
rawlingsj, line 36: was it definitely `libssh2-dev`?  I couldn't see the duplicate

dlorenc: Yeah it's here on line 30, just outside the preview.

image

rawlingsj: doh I was looking at the whole file using the commit sha - i.e. with the duplicate removed

rawlingsj (approved):

dlorenc:
Forcing merge since the new check is now failing on other PRs :)

PR #347, opened 2023-01-01
Trim some more ruby docs by moving them to the docs sub-package.

dlorenc:
This cuts the compressed package from 15MB to 10MB.


kaniini:
Should we do this to ruby-3.0 and 3.1 too?

kaniini (dismissed):

dlorenc:
> Should we do this to ruby-3.0 and 3.1 too?

Yea probably! Done!

kaniini (approved):

dlorenc:
This takes the built image from 72MB to 49MB!

luhring (approved):

PR #348, opened 2023-01-01
Trim the go packages by removing unneeded files.

dlorenc:
These were also inspired by the trimming in the Dockerfile: https://github.com/docker-library/golang/blob/master/1.19/bullseye/Dockerfile#L109.

This trims the package size (compressed) from 112MB to 90MB.


dlorenc:
The resulting image goes from 781MB to 638MB!

kaniini (approved):

luhring (approved):

PR #350, opened 2023-01-02
Bump trivy to v0.36.0.

dlorenc:
This one supports Wolfi!


trivy.yaml
```diff
@@ -19,7 +19,7 @@ pipeline:
   - uses: fetch
     with:
       uri: https://github.com/aquasecurity/trivy/archive/v${{package.version}}/trivy-${{package.version}}.tar.gz
-      expected-sha512: 6d8e828fa406518b4b3f55b0e5f62bbd5cf25cb5782d1884b9d5eaf61fb0614deaacad4236ab7420fa5b3868c79df226ae1aa5193bb136c556aa52853eeca553
+      expected-sha512: 52a9bf424934d68f04ff75a50e1daa792a3b733d80528c301c96f6249f46b47c
```
luhring, line 22: I think this is the sha256 digest. We can use this value if we update the key to `expected-sha256`.

luhring (changes requested):

trivy.yaml
dlorenc: Yea whoops, fixed back to 512

luhring (approved):
🎉

imjasonh (approved):

PR #145, opened 2022-12-01
Add jre headless subpackage

tuananh:
add jre-headless subpackage, addressing #142

PR #359, opened 2023-01-03
WIP: try and fix missing java from JDK and JREs

rawlingsj:
It seems we have an issue with our JDK packages.  This PR is failing https://github.com/wolfi-dev/os/pull/356, the reason is
```
The JAVA_HOME environment variable is not defined correctly,
```
while investigating found
```
javac: error while loading shared libraries: libjli.so: cannot open shared object file: No such file or directory
```

Which suggests Java isn't there.

My thinking is that this PR https://github.com/wolfi-dev/os/pull/145/files is moving java into a subpackage and therefore now doesn't exist in the main JDK package.

Because the JRE and JDK initially used different names to hold java in, i.e. /usr/lib/jvm/openjdk and /usr/lib/jvm/openjdk-jre, we cannot simply add a runtime dependency on the new headless package.  We need to keep the same directory structure for all, so changing to use /usr/lib/jvm/openjdk.

This is expected to break users that use export `JAVA_HOME=/usr/lib/jvm/openjdk-jre`, they will need to change to use `export JAVA_HOME=/usr/lib/jvm/openjdk`


rawlingsj:
Let's take care over this one as it contains a breaking change for users using the `-jre` suffix when setting JAVA_HOME, i.e. ```
JAVA_HOME=/usr/lib/jvm/openjdk-jre
````

rawlingsj:
Made WIP while testing.

PR #365, opened 2023-01-03
Run: wolfictl advisory sync-secfixes

luhring:
This PR adds our initial advisory data (needed for VEX) to the Wolfi distro. In particular, this:

1. Fixes various whitespace mistakes in the YAML files so that the YAML can be re-encoded without any goofiness
2. Adds a new lint check in CI to ensure that there is advisory data for all secfixes data, and vice versa
3. Runs `wolfictl advisory sync-secfixes` to populate advisory data using existing secfixes data
4. Adjusts this particular advisory timestamp data to match the `git blame` data for existing secfixes entries
5. Encodes the correct `justification` and `impact_statement` data via researching previously `nak`-ed vulnerabilities
6. Updates our `HOW_TO_PATCH_CVES.md` based on the new advisory system and `wolfictl advisory` commands

### DIFF
--- a/curl.yaml
+++ b/curl.yaml
@@ -7,10 +7,9 @@ package:
     - all
   copyright:
     - paths:
-      - "*"
+        - "*"
       attestation: TODO
       license: MIT
-
 secfixes:
   7.87.0-r0:
     - CVE-2022-43551
@@ -18,7 +17,6 @@ secfixes:
   7.86.0-r0:
     - CVE-2022-42916
     - CVE-2022-32221
-
 environment:
   contents:
     packages:
@@ -30,7 +28,6 @@ environment:
       - openssl-dev
       - zlib-dev
       - brotli-dev
-
 pipeline:
   - uses: fetch
     with:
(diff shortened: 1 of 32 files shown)


curl.yaml
```diff
@@ -30,26 +28,17 @@ environment:
...
     with:
-      opts: |
-        --enable-ipv6 \
-        --enable-unix-sockets \
-        --with-openssl \
-        --with-nghttp2 \
-        --with-pic \
-        --disable-ldap \
-        --without-libssh2    
+      opts: "--enable-ipv6 \\\n--enable-unix-sockets \\\n--with-openssl \\\n--with-nghttp2 \\\n--with-pic \\\n--disable-ldap \\\n--without-libssh2    \n"
```
luhring, line 38: This isn't great — wondering if anyone has suggestions for how to avoid this effect when operating on a `yaml.Node`...

rawlingsj, line 38: this is suffering from the same issues some `melange bump` calls make.  I think @kaniini suggested we could try and set the format of more nodes to https://github.com/chainguard-dev/melange/blob/bd4e1298b6f461ffcd203b2b4e0bb29107ecc6e6/pkg/renovate/bump/bump.go#L76

I did have a quick try of that but it didn't seem to work.  Maybe I was doing it wrong though.

puerco (approved):
Readddyyyy wooooot :rocket:

jdolitsky (approved):

PR #366, opened 2023-01-03
rust: drop dependency on rust-stage0

imjasonh:
This had worked because `go` was already built and available in Wolfi's packages.

We might want to keep `go` building using previous `go`s instead of `go-stage0`, in which case we should document this decision and its implications, and make `rust` consistent with that too.

### DIFF
--- a/go.yaml
+++ b/go.yaml
@@ -28,6 +28,10 @@ environment:
       - build-base
       - bash
       - go
+      # During bootstrapping, this must depend on go-stage0, which builds a Go
+      # package using a prebuilt Go toolchain provided by the Go team.
+      # After bootstrapping, go can depend on a previously built Go toolchain.
+      # - go-stage0
 pipeline:
   - uses: fetch
     with:
--- a/rust.yaml
+++ b/rust.yaml
@@ -34,11 +34,15 @@ environment:
       - clang-15
       - openssl-dev
       - python3
-      - rust-stage0
       - wasi-libc
       - xz-dev
       - zlib-dev
       - patch
+      - rust
+      # During bootstrapping, this must depend on rust-stage0, which builds a Rust
+      # package using a prebuilt Rust toolchain provided by the Rust team.
+      # After bootstrapping, rust can depend on a previously built Rust toolchain.
+      # - rust-stage0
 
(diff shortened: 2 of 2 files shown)


tuananh:
>  and its implications

can you share more about this? i would love to learn about this topic.

imjasonh:
> can you share more about this? i would love to learn about this topic.

Sure! For Go, and now Rust, we start by fetching a prebuilt Go/Rust toolchain ([`go-stage0`](https://github.com/wolfi-dev/os/blob/main/go-stage0.yaml), [`rust-stage0`](https://github.com/wolfi-dev/os/blob/main/rust-stage0.yaml)), then use that to build the "real" Go/Rust toolchains from source. In theory we could build that initial toolchain ourselves, and we might eventually, but for now we just use a prebuilt one provided by upstream releases.

The implication of having `go` depend on `go` is that we're implicitly assuming that package is already built and available in the Wolfi package repo (which is fine, since... it is). This works fine unless we somehow lose the `go` package in the repo, or need to rebuild everything from scratch, or someone else wants to reproduce our packages independently using their own repo -- `go` won't be there until it's built, so the `go` build will fail.

At some point in the past, `go.yaml` must have had a dependency on `go-stage0`, or else it wouldn't have been able to be built. The [oldest version of `go.yaml` we have](https://github.com/wolfi-dev/os/commit/43b655e03da6cbbb1aa4b19721e1daca4c0ed072) is from September, but it doesn't include a dependency on `go-stage0`, so `go` must have already been available in the repo at that time.

`rust.yaml` depends on `rust-stage0`, which means it doesn't require a pre-existing `rust` package. I think that's the general direction we want to go, and for consistency we should switch `go` to depend on `go-stage0`. But, presumably this change was made for a reason, so I wanted to get @kaniini's input in case there's something I'm missing.

It's also possible that this change means we should bump the `epoch` for `go`, since it changes the version of Go that's used to build Go.

imjasonh:
Updated this PR to instead keep go depending on go, and instead to update rust to depend on rust instead of rust-stage0.

kaniini (approved):

luhring (approved):

PR #367, opened 2023-01-03
add goreleaser,nerdctl,lima

developer-guy:

### DIFF
--- a/goreleaser.yaml
+++ b/goreleaser.yaml
@@ -1,0 +1,21 @@
+package:
+  name: goreleaser
+  version: 1.14.1
+  epoch: 0
+  description: Deliver Go binaries as fast and easily as possible
+  target-architecture:
+    - all
+  copyright:
+    - license: Apache-2.0
+      paths:
+        - "*"
+
+environment:
+  contents:
+    packages:
+      - git
+
+pipeline:
+  - uses: go/install
+    with:
+      package: github.com/goreleaser/goreleaser@v${{package.version}}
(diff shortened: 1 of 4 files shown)


goreleaser.yaml
```diff
@@ -0,0 +1,24 @@
...
+  description: Deliver Go binaries as fast and easily as possible
+  target-architecture:
+    - all
+  copyright:
+    - license: Apache-2.0
+      paths:
+        - "*"
+  dependencies:
+    runtime:
+      - apk-tools
```
imjasonh, line 14: I don't think these need apk-tools...

lima.yaml
```diff
@@ -0,0 +1,24 @@
+package:
+  name: lima
+  version: 0.14.2
+  epoch: 0
+  description: Linux virtual machines, typically on macOS, for running containerd
```
imjasonh, line 5: Is lima usable/useful from inside a container on (presumably) Linux? I think you know more about lima than I do, are you planning something by supporting this in Wolfi?

goreleaser.yaml
developer-guy: agree, yes.

lima.yaml
developer-guy: I'm not sure about that, good question 🤔 @akihirosuda can you please help me ?

imjasonh:
Closing this for now as we're waiting to hear whether limactl makes sense inside a container itself. Feel free to reopen.

PR #370, opened 2023-01-04
add tkn cli

developer-guy:


tkn.yaml
```diff
@@ -0,0 +1,16 @@
+package:
+  name: tkn
```
imjasonh, line 2: I should probably know this, but does tkn require kubectl at runtime?

developer-guy: not sure about that, but it looks like working, I tried with the following command:

>_docker container run --rm -ti nixery.dev/bash/tektoncd-cli bash_

imjasonh: That doesn't check that `tkn` actually runs, it only checks that `tkn` can be installed and that `bash` can run.

The docs don't mention needing `kubectl`. We could define a subpackage that enables [using `tkn` as a kubectl plugin](https://github.com/tektoncd/cli#tkn-as-a-kubectl-plugin) (`cp`s the binary to `kubectl-tkn` and requires `kubectl` at runtime), but I don't think we need to.

imjasonh (approved):

PR #379, opened 2023-01-05
add fuse3 and deps, cairo, eudev, gobject-introspection, kmod, pixman…

rawlingsj:
… and xcb-util

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -314,5 +314,12 @@ $(eval $(call build-package,krb5,1.20.1-r1))
 $(eval $(call build-package,libtirpc,1.3.3-r0))
 $(eval $(call build-package,spire-server,1.5.3-r0))
 $(eval $(call build-package,alpine-keys,2.4-r1))
+$(eval $(call build-package,xcb-util,0.4.1-r0))
+$(eval $(call build-package,pixman,0.42.2-r0))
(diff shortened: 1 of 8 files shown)


rawlingsj:
I'm not 100% if these are needed or useful in wolfi + some feel a bit complex so may need some discussion.  I was going to split out but as I don't know enough, I figured push as a draft PR to get visibility from folks next week and anyone considering the same packages

tuananh:
@rawlingsj should we have testing/community repo like alpine? new packages like this could be move to testing instead?

rawlingsj:
I'm not sure these are related to testing as they seem to be required for core dependencies.  But we have talked in the past about having different repos to core.  Maybe start a github discussion or bring it up at the wolfi office hours?

kaniini (approved):

kaniini:
> @rawlingsj should we have testing/community repo like alpine? new packages like this could be move to testing instead?

I don't want to do this for a number of reasons.  In Alpine, the purpose of `testing` is commonly confused and packages just wind up rotting there.  I also don't see the point in `community`, as it creates problems, and we do not have the same support model as Alpine.

dlorenc (approved):

PR #383, opened 2023-01-06
Bump wolfi-keys epoch

puerco:
We corrected the wolfi keys in https://github.com/wolfi-dev/os/pull/118 but the package was never rebuilt. This commit bumps the epoch to trigger a rebuild of the apk. The current apk has the keys in the wrong location (they are clobbering the arch directory name):

```
tar --list -f wolfi-keys-1-r2.apk
.PKGINFO
etc
etc/apk
etc/apk/keys
usr/share/apk/keys/aarch64
usr/share/apk/keys/x86_64
```

Signed-off-by: Adolfo García Veytia (Puerco) <puerco@chainguard.dev>


tuananh:
TIL: you can list files from apk like this `tar --list -f wolfi-keys-1-r2.apk` :)

dlorenc (approved):

luhring (approved):

PR #384, opened 2023-01-06
let users specifiy nodejs to use with yarn as it may not be the same …

rawlingsj:
…as current latest in wolfi


rawlingsj:
`yarn` is a new package that was introduced yesterday so this change should not break users

dlorenc (approved):

jdolitsky (approved):

PR #392, opened 2023-01-06
go-1.20: rename go-doc to avoid conflict

imjasonh:
Without this change there are two packages named `go-doc` which `dag` chokes on.

We should probably make this an actual wolfi lint check too.


puerco (approved):

luhring (approved):

luhring:
> We should probably make this an actual wolfi lint check too

💯

PR #408, opened 2023-01-09
go: add go-1.18, go-1.19

### LINKED ISSUE #385: Package go 1.18
Upstream Go supports the last two releases, which right now are 1.18 and 1.19. We only package 1.19 today.

I think we should follow the example set by go-1.20.yaml (an unreleased rc) and package go-1.18.yaml and go-1.19.yaml, with a `go` package in go.yaml that simply has a runtime dependency on the latest stable go release (currently `go-1.19`).

When go-1.20 is officially released ([expected "Feb 2023"](https://tip.golang.org/doc/go1.20)), go.yaml can update to point to `go-1.20`, go-1.18.yaml can be deleted as it's no longer supported upstream (the `go-1.18` package would continue to exist ~forever), and folks who want 1.19 can still get it by depending on the `go-1.19` package.

We'd follow this same process with go-1.21 has an rc, and later the full release.

Does this sound right @kaniini ?

imjasonh:
Fixes https://github.com/wolfi-dev/os/issues/385

This adds a package for `go-1.18`, which currently packages the latest release, `1.18.9`.

This also makes the `go` package provided by the `go-1.19` package.

When Go 1.20 is officially released (1.20 is currently an RC), we should update `go-1.20` to `provides=go-1.20`, and we can stop packaging `go-1.18` entirely at that time.


go-1.19.yaml
```diff
@@ -0,0 +1,91 @@
...
+  epoch: 0
+  description: "the Go programming language"
+  target-architecture:
+    - all
+  copyright:
+    - paths:
+        - "*"
+      attestation: TODO
+      license: BSD-3-Clause
+  dependencies:
```
imjasonh, line 13: Question: instead of having a `go` meta-package, should this `provides: go=1.19.4`?

go.yaml
```diff
@@ -12,8 +12,7 @@ package:
       license: BSD-3-Clause
   dependencies:
     runtime:
-      - build-base
-      - bash
+      - go-1.19
 secfixes:
```
imjasonh, line 16: Question: Do we need to keep `secfixes` and `advisories` for this meta-package? Or is it enough to have them in the `go-1.19` package this points to?

go-1.19.yaml
```diff
@@ -11,6 +11,8 @@ package:
       attestation: TODO
       license: BSD-3-Clause
   dependencies:
+    provides:
+      - go=1.19
```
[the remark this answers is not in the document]
imjasonh: I'm going to plead ignorance here about how `provides` works.

Why `.999`? Should this be `go=1.19.4`, and update to provide each new patch release? Should it do both, so that it counts as `go=1.19.<highest>` and `go=1.19.4` exactly?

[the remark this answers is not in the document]
kaniini: I would just do 1.19.999 tbh.  The version is just used to determine preference.  We need to do it specially for 1.19 because we have old packages.  In future we can just do `go=1.20`.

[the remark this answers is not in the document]
imjasonh: Makes sense. Done.

Makefile
```diff
@@ -146,7 +146,8 @@ $(eval $(call build-package,pcre2,10.42-r0))
 $(eval $(call build-package,git,2.39.0-r1))
 $(eval $(call build-package,bash,5.2_rc4-r0))
 $(eval $(call build-package,go-stage0,1.19.1-r0))
-$(eval $(call build-package,go,1.19.4-r1))
+$(eval $(call build-package,go-1.18,1.18.9-r0))
```
cpanato, line 149: do we need to build this one? 1.18.10 is out and 1.19.5 as well

kaniini (approved):

Makefile
imjasonh: 1.19.5 will be added in another PR.

We won't build any old packages as they're no supported upstream, but the packages for previously built versions are still available (e.g., 1.19.3, 1.19.1)

cpanato: thanks

cpanato (approved):

PR #413, opened 2023-01-10
ci(staging): create unique clusters for each dag run

imjasonh:
Instead of creating and destroying a cluster with a constant name `tmp-cluster-staging`, this will have us create clusters named like `tmp-cluster-staging-1673360663`, which means multiple builds can be ongoing at the same time.

There's a [quota of 100 clusters per zone](https://cloud.google.com/kubernetes-engine/quotas), but something would have to be going wrong to hit that.

### DIFF
--- a/.github/workflows/dag-push-staging.yaml
+++ b/.github/workflows/dag-push-staging.yaml
@@ -8,7 +8,6 @@ on:
 
 env:
   PROJECT: staging-images-183e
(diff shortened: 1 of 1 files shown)


luhring (approved):

cpanato (approved):

kaniini:
Why not use ${{github.run_id}} as i did in #414?

PR #414, opened 2023-01-10
actions: dag: use ${{github.run_id}} to differentiate cluster name

kaniini:
This solves the immediate problem of builds trampling on each other.  We will need to increase gcloud quotas to support the additional builders before this gets merged though.

Closes #413.

cc @dlorenc

### DIFF
--- a/.github/workflows/dag-push-production.yaml
+++ b/.github/workflows/dag-push-production.yaml
@@ -8,7 +8,7 @@ on:
 
 env:
   PROJECT: prod-images-c6e5
-  CLUSTER_NAME: tmp-cluster-prod
+  CLUSTER_NAME: tmp-cluster-prod-${{github.run_id}}
   CLUSTER_ZONE: us-central1-b
   SERVICE_ACCOUNT: prod-images-ci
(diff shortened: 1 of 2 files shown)


dlorenc:
> This solves the immediate problem of builds trampling on each other. We will need to increase gcloud quotas to support the additional builders before this gets merged though.

do you know which quotas? happy to try to bump them if i have permission

kaniini:
we need to probably upgrade the number of VMs (for GCE) and clusters (for GKE).  cc @imjasonh for more details I think

imjasonh (approved):

imjasonh:
> we need to probably upgrade the number of VMs (for GCE) and clusters (for GKE). cc @imjasonh for more details I think

GKE's cluster quota is 100 clusters per zone, I'm not worried about it. We have quota for 600 x86_64 CPUs in us-central1, but only 48 aarch64 CPUs.

If we hit a quota cap on CPUs we can still create clusters and request Pods on those clusters, they'll just go pending for longer and eventually time out.

jdolitsky (approved):

PR #417, opened 2023-01-11
haproxy: add haproxy-oci-entrypoint subpkg

tuananh:
- Add haproxy-oci-entrypoint subpkg in order to create haproxy images.
docker-entrypoint.sh is copied from the offical haproxy image https://github.com/docker-library/haproxy/blob/master/2.8/alpine/docker-entrypoint.sh

i use `${{targets.subpkgdir}}/usr/local/bin/docker-entrypoint.sh` because the haproxy official image does so. no surprise when switching. lemme know if you prefer it to be name sth different like `haproxy-entrypoint.sh` for example

also, when we add that entrypoint script, we also need to pull in busybox (sh). but doing this will make it compatible with the official image

- remove the default `haproxy.cfg` in /etc/haproxy/ because users should supply one anyway.


haproxy/docker-entrypoint.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
```
dlorenc, line 1: Should we link to the source this was copied from?

dlorenc (approved):

tuananh:
@rawlingsj can you take a look at this ?

rawlingsj (approved):

PR #423, opened 2023-01-11
Bind tools dependencies part4 .....

rawlingsj:
This PR includes an unusually large number of packages that relate to eventually building bind-tools.  Rather than split into a PR for each which makes it hard to manage CI, we're trying a single PR with multiple commits that should help the review process.

This PR includes `docbook-xml` and `docbook-xsl` which means we can recompile `xmlto` and `glib` as it was missing from each when originally added.

Packages being added:

- swig
- lua 5.4
- graphviz
- vala
- bump libgcrypt
- libsecret
- libtasn1
- strace
- wayland
- wayland-protocols
- at-spi2-core
- py3-docutils
- py3-pygments
- docbook-xml
- docbook-xsl
- rarian
- itstool
- shared-mime-info


kaniini:
CI and lint are failing atm

tuananh:
i'm surprised to see wayland here. do we need it in container workload?

rawlingsj:
AFAICT it's just used as a dependency to compile other packages, needed to eventually compile bind-tools.

docbook-xml.yaml
```diff
@@ -0,0 +1,316 @@
...
+        - "*"
+      attestation: TODO
+      license: MIT
+  dependencies:
+    runtime:
+      - libxml2-utils
+      - busybox
+  scriptlets:
+    post-install: |
+      #!/bin/busybox sh
```
kaniini, line 19: Instead of having a scriptlet here, lets merge docbook-xml and docbook-xsl into a single package and then build /etc/xml/catalog at build time.

kaniini (changes requested):

kaniini (approved):

PR #425, opened 2023-01-12
add kubescape

developer-guy:

PTAL @dwertent @slashben

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -366,6 +366,7 @@ $(eval $(call build-package,py3-s3transfer,0.5.1-r0))
 $(eval $(call build-package,py3-urllib3,1.26.13-r0))
 $(eval $(call build-package,py3-yaml,6.0-r0))
 $(eval $(call build-package,aws-cli,1.27.38-r0))
+$(eval $(call build-package,kubescape,2.0.183-r0))
 
 .build-packages: ${PACKAGES}
 
--- a/kubescape.yaml
+++ b/kubescape.yaml
@@ -1,0 +1,39 @@
+package:
+  name: kubescape
+  version: 2.0.183
+  epoch: 0
+  description: "Kubescape is an open-source Kubernetes security platform for your IDE, CI/CD pipelines, and clusters. It includes risk analysis, security, compliance, and misconfiguration scanning, saving Kubernetes users and administrators precious time, effort, and resources."
+  target-architecture:
+    - all
+  copyright:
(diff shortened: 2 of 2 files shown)


developer-guy:
I'm just not sure that kubescape needs `libgit2` as a run-time dependency; folx, @dwertent @slashben would you mind helping me with that?

slashben:
We need to parse it to parse Git objects. The problem we haven't found any Golang native alternative to libgit2

developer-guy:
so, it depends to libgit2 package at runtime, right?

matthyx:
> so, it depends to libgit2 package at runtime, right?

it's not a runtime dependency... the build is static so you only need the binary file

developer-guy:
> > so, it depends to libgit2 package at runtime, right?
> 
> it's not a runtime dependency... the build is static so you only need the binary file

thanks, @matthyx, I've updated the package accordingly ☝️

developer-guy:
> > so, it depends to libgit2 package at runtime, right?
> 
> it's not a runtime dependency... the build is static so you only need the binary file

thanks, @matthyx, I've updated the package accordingly ☝️

kubescape.yaml
```diff
@@ -0,0 +1,39 @@
+package:
+  name: kubescape
+  version: 2.0.182
```
dwertent, line 3:
```suggestion
  version: 2.0.183
```

Makefile
```diff
@@ -366,6 +366,7 @@ $(eval $(call build-package,py3-s3transfer,0.5.1-r0))
 $(eval $(call build-package,py3-urllib3,1.26.13-r0))
 $(eval $(call build-package,py3-yaml,6.0-r0))
 $(eval $(call build-package,aws-cli,1.27.38-r0))
+$(eval $(call build-package,kubescape,2.0.182-r0))
```
dwertent, line 369:
```suggestion
$(eval $(call build-package,kubescape,2.0.183-r0))
```

developer-guy: thanks @dwertent, I've upgraded it accordingly.

kubescape.yaml
developer-guy: thanks @dwertent, I've upgraded it accordingly.

dwertent (approved):

dlorenc (approved):

rawlingsj (approved):

PR #439, opened 2023-01-16
feat: add py3-magic

denhamparry:
Signed-off-by: Lewis Denham-Parry <lewis@chainguard.dev>


rawlingsj:
We also need to add an entry to the Makefile

rawlingsj:
I.e. adding an entry like..
```
$(eval $(call build-package,py3-magic,0.4.27-r0))
``` 

here which is after the last environment dependency this package needs

https://github.com/wolfi-dev/os/blob/f2e5a236c4c03a931a8f6dcf0eed26437208dc7b/Makefile#LL140-L140C51

denhamparry:
> I.e. adding an entry like..
> 
> ```
> $(eval $(call build-package,py3-magic,0.4.27-r0))
> ```
> 
> here which is after the last environment dependency this package needs
> 
> https://github.com/wolfi-dev/os/blob/f2e5a236c4c03a931a8f6dcf0eed26437208dc7b/Makefile#LL140-L140C51

Thanks for the feedback, I've updated the `Makefile` accordingly 🙌

rawlingsj (dismissed):

rawlingsj (approved):

dlorenc (approved):

PR #445, opened 2023-01-17
curl: add optional rustls backend

kaniini:
Add new `libcurl-rustls4` package which provides libcurl compiled against rustls backend.  Since curl itself does not encode any backend-specific data into its public ABI, we can provide multiple libcurl packages compiled against various backends and have them be interchangable.


dlorenc (dismissed):

curl.yaml
```diff
@@ -58,12 +74,26 @@ subpackages:
     description: "documentation for curl"
     pipeline:
       - uses: split/manpages
-  - name: "libcurl4"
-    description: "curl library"
+  - name: "libcurl-openssl4"
+    description: "curl library (openssl backend)"
+    dependencies:
+      provides:
+        - libcurl4=7.87.1
```
jdolitsky, line 81: why this version differs from main package?

kaniini: We want to provide some extra weight to these packages verses the original libcurl4.

jdolitsky (dismissed):

PR #456, opened 2023-01-19
feat: add fluent-bit pkg

### LINKED ISSUE #452: [Wolfi Package Request]: fluent-bit

Fluent Bit: https://github.com/fluent/fluent-bit

https://docs.fluentbit.io/

2.0.8+

Most recent supported version, includes logging, metrics and OTEL output now as well.

- [X] I am interested in adding this package to Wolfi OS myself.
- [X] I am willing to help maintain this package.

tuananh:
~~WIP, i want to do a bit more testing as well as getting feedback from @patrick-stephens~~

Fixes: #452


fluent-bit.yaml
```diff
@@ -0,0 +1,64 @@
...
+      - bison
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - cmake
+      - flex
+      - glibc
+      - openssl-dev
+      - yaml-dev
+      - zlib-dev
```
patrick-stephens, line 27: If you want the systemd/journald input plugin then we need to add the `libsystemd-dev` or equivalent.
It's useful to have for kubelet logs or other systemd services: https://docs.fluentbit.io/manual/pipeline/inputs/systemd

patrick-stephens, line 27: I notice we're enabling the postgresql output which probably means we should explicitly include the `libpq-dev` and `postgresql-server-dev-all` or similar headers & libraries.

fluent-bit.yaml
```diff
@@ -0,0 +1,64 @@
...
+pipeline:
+  - uses: fetch
+    with:
+      uri: https://github.com/fluent/fluent-bit/archive/v${{package.version}}.tar.gz
+      expected-sha256: 8ff5566389033669feabc9c69a5c6f417dad5c8b066454388e6a706507262acf
+  - runs: |
+      cmake -B build \
+            -DCMAKE_INSTALL_PREFIX=/usr \
+            -DCMAKE_INSTALL_LIBDIR=lib \
+            -DFLB_TESTS_INTERNAL=On \
```
patrick-stephens, line 37: You're not running the tests below so it will speed up the build to disable this. If you want the tests then I think we need to invoke the `make tests` target but you could also run them separately.

fluent-bit.yaml
```diff
@@ -0,0 +1,64 @@
...
+  name: fluent-bit
+  version: 2.0.8
+  epoch: 0
+  description: Fast and Lightweight Log processor and forwarder
+  target-architecture:
+    - all
+  copyright:
+    - paths:
+        - "*"
+      attestation: TODO
```
patrick-stephens, line 11: Do we need to do something for this btw?

fluent-bit.yaml
```diff
@@ -0,0 +1,64 @@
...
+      license: Apache-2.0
+environment:
+  contents:
+    packages:
+      - autoconf
+      - automake
+      - bison
+      - build-base
+      - busybox
```
patrick-stephens, line 20: What's this for? A shell to build with?

fluent-bit.yaml
patrick-stephens: There are also runtime tests but that may need more setup to execute: `-DFLB_TESTS_RUNTIME=On`

fluent-bit.yaml
```diff
@@ -0,0 +1,64 @@
...
+      - cmake
+      - flex
+      - glibc
+      - openssl-dev
+      - yaml-dev
+      - zlib-dev
+pipeline:
+  - uses: fetch
+    with:
+      uri: https://github.com/fluent/fluent-bit/archive/v${{package.version}}.tar.gz
```
patrick-stephens, line 31: We also have the source detained with checksums here:
- https://releases.fluentbit.io/2.0/source-2.0.8.tar.gz
- https://releases.fluentbit.io/2.0/source-2.0.8.tar.gz.sha256

(Other repo maintainers preferred that over Github tagged versions so just wanted to indicate it).

patrick-stephens (commented):
Looks good to me, I had a few queries.

patrick-stephens:
One other thing that may not be relevant, the `exec` plugin requires `/bin/sh` to exist: https://docs.fluentbit.io/manual/pipeline/inputs/exec#container-support

This means that for the distroless container deployments we disable it to provide an explicit error rather than some hidden snowball effect: https://github.com/fluent/fluent-bit/blob/3d4237dad1751028c5bc4513e4299850814954ca/dockerfiles/Dockerfile#L72 

I'm not sure if this package will be deployed into containers but something to be aware of.

fluent-bit.yaml
tuananh: I think i will disable tests. Lets leave those upstream

fluent-bit.yaml
tuananh: Those are build dependency. I havent declared any runtime dependency yet

fluent-bit.yaml
patrick-stephens: The `-dev` is not but the libraries are - same for the others. We use `cpack` to handle it currently, i.e. auto-require things.

tuananh: i did not see any errors. would it just skip if the dependency are missing?

patrick-stephens: Probably unless your base already has it. I just like to make it explicit .

tuananh:
@patrick-stephens how do you test the bartifact?

right now, i just test it by producing an image and test it like this

image

fluent-bit.yaml
tuananh: they are left TODO like this for now.

@rawlingsj what should we put here for attestation?

fluent-bit.yaml
tuananh: previously, it doesn't have libpq-dev and no error. how can i test with postgres?

patrick-stephens: We just need to confirm it's built and the easiest way is to check if the plugin is included in the output:

```shell
$ docker run --rm -it fluent/fluent-bit -h
Usage: /fluent-bit/bin/fluent-bit [OPTION]
...
Outputs
...
  pgsql                   PostgreSQL
```

tuananh: this is what i get 

```
Usage: /usr/bin/fluent-bit [OPTION]

[1mAvailable Options[0m
  -b  --storage_path=PATH specify a storage buffering path
  -c  --config=FILE       specify an optional configuration file
  -d, --daemon            run Fluent Bit in background mode
  -D, --dry-run           dry run
  -f, --flush=SECONDS     flush timeout in seconds (default: 1)
  -C, --custom=CUSTOM     enable a custom plugin
  -i, --input=INPUT       set an input
  -F  --filter=FILTER     set a filter
  -m, --match=MATCH       set plugin match, same as '-p match=abc'
  -o, --output=OUTPUT     set an output
  -p, --prop="A=B"        set plugin configuration property
  -R, --parser=FILE       specify a parser configuration file
  -e, --plugin=FILE       load an external plugin (shared lib)
  -l, --log_file=FILE     write log info to a file
  -t, --tag=TAG           set plugin tag, same as '-p tag=abc'
  -T, --sp-task=SQL       define a stream processor task
  -v, --verbose           increase logging verbosity (default: info)
  -vv                     trace mode (available)
  -Z, --enable-chunk-traceenable chunk tracing. activating it requires using the HTTP Server API.
  -w, --workdir           set the working directory
  -H, --http              enable monitoring HTTP server
  -P, --port              set HTTP server TCP port (default: 2020)
  -s, --coro_stack_size   set coroutines stack size in bytes (default: 24576)
  -q, --quiet             quiet mode
  -S, --sosreport         support report for Enterprise customers
  -V, --version           show version number
  -h, --help              print this help

[1mInputs[0m
  cpu                     CPU Usage
  mem                     Memory Usage
  thermal                 Thermal
  kmsg                    Kernel Log Buffer
  proc                    Check Process health
  disk                    Diskstats
  netif                   Network Interface Usage
  docker                  Docker containers metrics
  docker_events           Docker events
  node_exporter_metrics   Node Exporter Metrics (Prometheus Compatible)
  fluentbit_metrics       Fluent Bit internal metrics
  prometheus_scrape       Scrape metrics from Prometheus Endpoint
  tail                    Tail files
  dummy                   Generate dummy data
  head                    Head Input
  health                  Check TCP server health
  http                    HTTP
  collectd                collectd input plugin
  statsd                  StatsD input plugin
  opentelemetry           OpenTelemetry
  event_test              Event tests for input plugins
  event_type              Event tests for input plugins
  nginx_metrics           Nginx status metrics
  serial                  Serial input
  stdin                   Standard Input
  syslog                  Syslog
  udp                     UDP
  exec_wasi               Exec WASI Input
  tcp                     TCP
  mqtt                    MQTT, listen for Publish messages
  forward                 Fluentd in-forward
  random                  Random

[1mFilters[0m
  alter_size              Alter incoming chunk size
  aws                     Add AWS Metadata
  checklist               Check records and flag them
  ecs                     Add AWS ECS Metadata
  record_modifier         modify record
  throttle                Throttle messages using sliding window algorithm
  type_converter          Data type converter
  kubernetes              Filter to append Kubernetes metadata
  modify                  modify records by applying rules
  multiline               Concatenate multiline messages
  nest                    nest events by specified field values
  parser                  Parse events
  expect                  Validate expected keys and values
  grep                    grep events by specified field values
  rewrite_tag             Rewrite records tags
  lua                     Lua Scripting Filter
  stdout                  Filter events to STDOUT
  geoip2                  add geoip information to records
  nightfall               scans records for sensitive content
  wasm                    WASM program filter

[1mOutputs[0m
  azure                   Send events to Azure HTTP Event Collector
  azure_blob              Azure Blob Storage
  azure_kusto             Send events to Kusto (Azure Data Explorer)
  bigquery                Send events to BigQuery via streaming insert
  counter                 Records counter
  datadog                 Send events to DataDog HTTP Event Collector
  es                      Elasticsearch
  exit                    Exit after a number of flushes (test purposes)
  file                    Generate log file
  forward                 Forward (Fluentd protocol)
  http                    HTTP Output
  influxdb                InfluxDB Time Series
  logdna                  LogDNA
  loki                    Loki
  kafka                   Kafka
  kafka-rest              Kafka REST Proxy
  nats                    NATS Server
  nrlogs                  New Relic
  null                    Throws away events
  opensearch              OpenSearch
  plot                    Generate data file for GNU Plot
  pgsql                   PostgreSQL
  skywalking              Send logs into log collector on SkyWalking OAP
  slack                   Send events to a Slack channel
  splunk                  Send events to Splunk HTTP Event Collector
  stackdriver             Send events to Google Stackdriver Logging
  stdout                  Prints events to STDOUT
  syslog                  Syslog
  tcp                     TCP Output
  udp                     UDP Output
  td                      Treasure Data
  flowcounter             FlowCounter
  gelf                    GELF Output
  websocket               Websocket
  cloudwatch_logs         Send logs to Amazon CloudWatch
  kinesis_firehose        Send logs to Amazon Kinesis Firehose
  kinesis_streams         Send logs to Amazon Kinesis Streams
  opentelemetry           OpenTelemetry
  prometheus_exporter     Prometheus Exporter
  prometheus_remote_write Prometheus remote write
  s3                      Send to S3

[1mInternal[0m
 Event Loop  = epoll
 Build Flags = FLB_HAVE_IN_STORAGE_BACKLOG FLB_HAVE_CHUNK_TRACE FLB_HAVE_PARSER FLB_HAVE_RECORD_ACCESSOR FLB_HAVE_STREAM_PROCESSOR JSMN_PARENT_LINKS JSMN_STRICT FLB_HAVE_TLS FLB_HAVE_OPENSSL FLB_HAVE_METRICS FLB_HAVE_WASM FLB_HAVE_AWS FLB_HAVE_AWS_CREDENTIAL_PROCESS FLB_HAVE_SIGNV4 FLB_HAVE_SQLDB FLB_HAVE_TRACE FLB_HAVE_METRICS FLB_HAVE_HTTP_SERVER FLB_HAVE_FORK FLB_HAVE_TIMESPEC_GET FLB_HAVE_GMTOFF FLB_HAVE_UNIX_SOCKET FLB_HAVE_LIBYAML FLB_HAVE_ATTRIBUTE_ALLOC_SIZE FLB_HAVE_PROXY_GO FLB_HAVE_JEMALLOC JEMALLOC_MANGLE FLB_HAVE_LIBBACKTRACE FLB_HAVE_REGEX FLB_HAVE_UTF8_ENCODER FLB_HAVE_LUAJIT FLB_HAVE_C_TLS FLB_HAVE_ACCEPT4 FLB_HAVE_INOTIFY FLB_HAVE_GETENTROPY FLB_HAVE_GETENTROPY_SYS_RANDOM

```

tuananh: Looks like pqsql is there

patrick-stephens: Yeah so I'm guessing the dependency is already present on your builder or added by one of the others. Looks good anyway!

patrick-stephens:
> @patrick-stephens how do you test the artifact?
> 
> right now, i just test it by producing an image and test it like this
> 
> image

That's fine, I'd probably test with the default configuration as well - CPU input --> stdout. That way you don't need anything else.

For our enterprise stuff we have various tests but the easiest is using a dummy input to generate a fixed string with a stdout output so you can verify it appears in the output. I don't think you need to verify all the plugins work, etc. - that's the responsibility of upstream I would say.

fluent-bit.yaml
patrick-stephens: Ok, I just wanted to check if it needed anything upstream

kaniini: the attestation should ideally be a collection of the copyright lines plus the "this software is licensed under blabla" text.  similar to `debian/copyright`, but we are still kinda playing with the format, so it's not so important right now.

tuananh:
@rawlingsj could you help taking a look at this?

fluent-bit/fluent-bit.confd
```diff
@@ -0,0 +1 @@
+fluentbit_opts="-c /etc/fluent-bit/fluent-bit.conf"
```
kaniini, line 1: any particular reason for this file? :p

tuananh: gonna remove that.

Dentrax:
Is `postgresql` really necessary? That's because `-DFLB_SQLDB=On` feature gate? It'd be nice to strip some unnecessary packages by disabling not-widely-use feature flags. For example, is there any package we build more than one image for an app? I mean, create 2 different fluent-bit images: keep `fluent-bit` bare-minimum with everything-default, and `fluent-bit-full` with every single feature enabled. Just thinking...

rawlingsj:
What does the flag `-DFLB_SQLDB=On` do?  If it generates extra files used at runtime then we could move those out into a subpackage.  Or if it builds a different flavour of a binary maybe compile it a second time in a subpackage pipeline

i.e. 

```yaml
subpackages:
  - name: fluent-bit-sql
    pipeline:
      - runs: cmake -B build ...
      - runs: make -C build ...
      - runs: make -C build DESTDIR="${{targets. subpkgdir}}" install
      - ....
```
Then folks can install what combinations of packages they like, i.e. 
```
apk add fluent-bit
```
or 
```
apk add fluent-bit fluent-sql
```
Would that help?

rawlingsj:
BTW as a side note there's also some reusable pipelines for cmake https://github.com/chainguard-dev/melange/tree/main/pkg/build/pipelines/cmake not sure if they make sense using here but thought I'd mention just in case

tuananh:
From the output of it, its just a single binary. Maybe @patrick-stephens cna confirm?

patrick-stephens:
> From the output of it, its just a single binary. Maybe @patrick-stephens cna confirm?

Yes, it is a single binary with whatever combinations of components you enable compiled in. It does not support sub-packages in that sense although users can provide custom plugins via shared libraries they want to use but this is not part of the official package.

When we make an RPM/DEB we also provide a systemd unit and standard config file to use along with it. I'm not sure how Wolfi does it though - the binary needs some configuration to run with but this can be provided as a file or via command line arguments.

patrick-stephens:
> Is `postgresql` really necessary? That's because `-DFLB_SQLDB=On` feature gate? It'd be nice to strip some unnecessary packages by disabling not-widely-use feature flags. For example, is there any package we build more than one image for an app? I mean, create 2 different fluent-bit images: keep `fluent-bit` bare-minimum with everything-default, and `fluent-bit-full` with every single feature enabled. Just thinking...

SQLDB is required for the internal DB feature and I would say is mandatory - it's required to handle restarted and persisting your location in the log file. It is also the default: https://github.com/fluent/fluent-bit/blob/16eae107868bf03da2fce6ca4acd3f6ab2d3c4ff/CMakeLists.txt#L127

Postgresql output is a separate flag: https://github.com/fluent/fluent-bit/blob/master/CMakeLists.txt#L224

I would suggest aligning with what the upstream project does unless there is a good reason not to otherwise people will have migration issues if they want to adopt. By all means have variants but maybe not initially and I guarantee it'll become a complex matrix!

tuananh:
> BTW as a side note there's also some reusable pipelines for cmake https://github.com/chainguard-dev/melange/tree/main/pkg/build/pipelines/cmake not sure if they make sense using here but thought I'd mention just in case

i looked at it but it's not suitable in this case. maybe we need to update the pipeline

for example: we need `-C <dir>` flag in make commadn but the pipeline doesnt support it.

tuananh:
@rawlingsj can you take a look at this?

tuananh:
> > I would suggest aligning with what the upstream project does unless there is a good reason not to otherwise people will have migration issues if they want to adopt.
> 
> Oh, this is one of the important aspects. I love this thinking. I think Wolfi packages already respecting to that, but for what it's worth, we should bring contributor's attention more in CONTRIBUTING.md by pointing under `Write your first Wolfi package` section. And apply this mindset to every package during the PR reviews. WDYT @tuananh?

Totally agree. Feel free to update it. I'm not very good when it comes to copywriting & documentation as English is not my native language :(

rawlingsj (approved):

jdolitsky (approved):

fluent-bit.yaml
imjasonh: Filed https://github.com/wolfi-dev/os/issues/510 to track

imjasonh (approved):

fluent-bit.yaml
```diff
@@ -0,0 +1,63 @@
...
+      - glibc
+      - libpq-15
+      - openssl-dev
+      - postgresql-15-dev
+      - yaml-dev
+      - zlib-dev
+pipeline:
+  - uses: fetch
+    with:
+      uri: https://releases.fluentbit.io/2.0/source-${{package.version}}.tar.gz
```
amouat, line 33: won't the 2.0 break when it moves to 2.1? Can we add a var here somehow?

amouat (commented):
Looks good, just a minor problem with URL versioning.

fluent-bit.yaml
patrick-stephens: I mean I don't really know why it has to be in that sub-directory anyway - just because it was so we don't want to break existing builds. I can always update the Fluent Bit release process to make it top-level as the version is explicit in the package name anyway.

The other option is to use the Github Release/tag URL to get the source which was done originally.

tuananh: @amouat can we do some string manipulation here with melange? `2.0.8` => `2.0`?

patrick-stephens: If you can that should be fine, I do something (disgustingly) similar for the release: https://github.com/fluent/fluent-bit/blob/7bcb502ebdcd647d6b83c972805e734ad522eb1a/.github/workflows/staging-release.yaml#L59-L61

dlorenc (approved):

tuananh:
@rawlingsj this is weird. i can see the fluent-bit-dev package in the index but not the fluent-bit package. (https://packages.wolfi.dev/os/x86_64/APKINDEX.tar.gz)

any idea why?

kaniini:
It can take a little bit of time for newly uploaded packages to the repository to be indexed.

PR #469, opened 2023-01-24
Bump epoch for all packages

luhring:
This bumps the epoch value for all packages using `wolfictl bump ./*.yaml`.

The intended outcome here is:

1. We trigger rebuilds for all packages, which means we get the benefit of recent enhancements to Melange (such as SBOM improvements) to manifest for all packages.
2. We find out if we _can_ rebuild all packages at once.

homer-all-of-them


kaniini:
Rebuild hit the 6 hour time limit and was terminated.  I guess we will have to batch it.

luhring:
Closing this PR in order to approach this task via batch groups

luhring:
First batch: #477

PR #478, opened 2023-01-25
haproxy: upgrade to 2.6.8

curantes:
- Switch make TARGET to linux-glibc since wolfi uses glibc and not musl
- Add USE_GETADDRINFO to support resolve of IPv6 host names


haproxy.yaml
```diff
@@ -25,16 +24,15 @@ environment:
...
   - uses: fetch
     with:
       uri: https://www.haproxy.org/download/2.6/src/haproxy-${{package.version}}.tar.gz
-      expected-sha256: cff9b8b18a52bfec277f9c1887fac93c18e1b9f3eff48892255a7c6e64528b7d
+      expected-sha256: a02ad64550dd30a94b25fd0e225ba699649d0c4037bca3b36b20e8e3235bb86f
   - uses: autoconf/make
     with:
       opts: |
-        TARGET=linux-musl \
+        TARGET=linux-glibc \
```
rawlingsj, line 35: good catch!

rawlingsj:
We also need to update the Makefile to reference the new version https://github.com/wolfi-dev/os/blob/b5909080f31486a2fb2b326b971fb149562c2026/Makefile#L386

The use of the Makefile will go away so this is only temporary.

curantes:
> We also need to update the Makefile to reference the new version
> 
> https://github.com/wolfi-dev/os/blob/b5909080f31486a2fb2b326b971fb149562c2026/Makefile#L281
> 
> The use of the Makefile will go away so this is only temporary.

Yes I noticed I missed that, will fix

jdolitsky (approved):

rawlingsj (approved):

PR #528, opened 2023-02-03
janitor old clusters more aggressively

imjasonh:
Instead of janitoring clusters every day and deleting clusters >24 old, run every hour and delete clusters >7 hours old.

Since GitHub jobs are limited to 6 hours, any cluster older than that can be assumed to be dead.

I couldn't test this locally myself since `date -d` doesn't work on macOS, so please let me know if this is wrong.

### DIFF
--- a/scripts/janitor-clusters.sh
+++ b/scripts/janitor-clusters.sh
@@ -12,9 +12,9 @@ do
   echo "  NAME: ${NAME}"
   echo "  CREATE_TIME: ${CREATE_TIME} / ${FORMATED_CREATE_TIME}"
 
-  YESTERDAY_DATE=$(date -I -d "yesterday")
-  if [[ ${FORMATED_CREATE_TIME} < ${YESTERDAY_DATE} ]]; then
-    echo "Cluster ${NAME}(${FORMATED_CREATE_TIME}) is older than ${YESTERDAY_DATE} will terminate"
+  CUTOFF_DATE=$(date -I -d "7 hours ago")
+  if [[ ${FORMATED_CREATE_TIME} < ${CUTOFF_DATE} ]]; then
+    echo "Cluster ${NAME}(${FORMATED_CREATE_TIME}) is older than ${CUTOFF_DATE} will terminate"
     gcloud container clusters delete "${NAME}" \
     --zone "${ZONE}" \
     --quiet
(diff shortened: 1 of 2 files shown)


luhring (approved):

kaniini:
I feel like this is going to shoot us in the foot.  We want the ability to have builds which last longer than 6 hours, for example.

dlorenc:
Do we know why the clusters started leaking? I hadn't seen any until today.

cpanato:
> Do we know why the clusters started leaking? I hadn't seen any until today.

we had some clusters that i cleanup when I did the janitor job, but was two clusters in each env only

cpanato:
> Do we know why the clusters started leaking? I hadn't seen any until today.

reason: https://chainguard-dev.slack.com/archives/C0499BTMHPF/p1675458660417649?thread_ts=1675439705.042159&cid=C0499BTMHPF

scripts/janitor-clusters.sh
```diff
@@ -12,9 +12,9 @@ do
   echo "  NAME: ${NAME}"
   echo "  CREATE_TIME: ${CREATE_TIME} / ${FORMATED_CREATE_TIME}"
 
-  YESTERDAY_DATE=$(date -I -d "yesterday")
-  if [[ ${FORMATED_CREATE_TIME} < ${YESTERDAY_DATE} ]]; then
-    echo "Cluster ${NAME}(${FORMATED_CREATE_TIME}) is older than ${YESTERDAY_DATE} will terminate"
+  CUTOFF_DATE=$(date -I -d "7 hours ago")
```
cpanato, line 15: how about 12 hours?

imjasonh: I'm fine with anything. We can have them reaped after 7 hours for now, and once we have builds that actually successfully last longer than 7 hours we can run the janitor's cutoff accordingly.

Right now we have a pretty hard 6-hour cap, so keeping clusters around 6 hours after that seems wasteful, but I don't really feel strongly about it. 🤷

cpanato: 7 is fine for me as well, just trying to balance the comments 😅

cpanato (approved):

imjasonh:
Sounds like we're generally fine with the longer timeline, rather than bump up into actually still-running jobs. Closing for now.

PR #531, opened 2023-02-03
dont enable the node auto-repair

cpanato:
dont enable the node auto-repair


imjasonh (approved):
Background: these clusters are short-lived anyway, and if anything goes wrong we're just going to delete the cluster anyway. When a node is being auto-repaired it's actually blocked from being deleted, which hinders our ability to just delete and try again, and ends up with zombie clusters sitting around costing money.

cpanato:
thanks for the justification

rawlingsj (approved):

PR #534, opened 2023-02-05
Add syft package!

dlorenc:

### DIFF
--- a/syft.yaml
+++ b/syft.yaml
@@ -1,0 +1,16 @@
+package:
+  name: syft
+  version: 0.70.0
+  epoch: 0
+  description: CLI tool and library for generating a Software Bill of Materials from container images and filesystems
+  target-architecture:
+    - all
+  copyright:
+    - license: Apache-2.0
+      paths:
+        - "*"
+
+pipeline:
+  - uses: go/install
+    with:
+      package: github.com/anchore/syft/cmd/syft@v${{package.version}}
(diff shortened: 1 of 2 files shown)


developer-guy (dismissed):

syft.yaml
```diff
@@ -0,0 +1,24 @@
...
+  description: CLI tool and library for generating a Software Bill of Materials from container images and filesystems
+  target-architecture:
+    - all
+  copyright:
+    - license: Apache-2.0
+      paths:
+        - "*"
+  dependencies:
+    runtime:
+      - apk-tools
```
luhring, line 14: What's the thinking behind this?

syft.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+environment:
+  contents:
+    packages:
+      - git
```
luhring, line 19: Do we need `git`?

syft.yaml
dlorenc: Good question, I stole this from the apko sample of go/install, my mistake!

syft.yaml
dlorenc: Yea, go install fetches things with git.

dlorenc: Well, I guess it doesn't always, but it can. I'll try without and see if it works here!

dlorenc: nvm - it's installed anywhere here: https://github.com/chainguard-dev/melange/blob/main/pkg/build/pipelines/go/install.yaml#L8

luhring (approved):
Sweet! 🎉

rawlingsj (approved):

PR #536, opened 2023-02-05
Add nodejs 19 and 16, in addition to 18.

dlorenc:
This changes the package naming to be suffixed with the version, but keeps the LTS version as the default using a provdies clause.


kaniini (dismissed):

kaniini:
I wonder if we should also provide a `nodejs-latest` virtual for the non-LTS channel.

nodejs-18.yaml
```diff
@@ -10,6 +10,9 @@ package:
         - "*"
       attestation: TODO
       license: MIT
+  dependencies:
+    provides:
+      - nodejs=18.14.999
```
tuananh, line 15: what does `999` mean here? it will satifies any package require `18.14.x` going forward ?

dlorenc:
> I wonder if we should also provide a `nodejs-latest` virtual for the non-LTS channel.

Do you have an example of how to set this up?

nodejs-18.yaml
dlorenc: Not entirely sure honestly! I copied this from go: https://github.com/wolfi-dev/os/blob/main/go-1.20.yaml#L15

imjasonh: This means "when someone asks for `nodejs`, pretend this package provides `nodejs 18.14.999` -- this will always make it the highest-versioned `nodejs` and select it unless there's a `nodejs=19`+.

TBH I don't really love this convention, and I wish we'd just have it provide `nodejs=18.14.0` explicitly, and just update it when we update this package's version. It seems to only use `999` as a convenience to mean "the highest patch version possible". It does prevent us from forgetting to bump it when we update the package though.

rawlingsj (dismissed):

rawlingsj (approved):

imjasonh (approved):

PR #537, opened 2023-02-05
Bump php.

dlorenc:


luhring (approved):
Just curious why the bot didn't bump this? cc: @rawlingsj

rawlingsj:
It's because currently update data for packages lives in a separate repo which we need to keep up to date but has fallen behind.  For now I've added it here https://github.com/wolfi-dev/wolfi-update-mapper/commit/438652b32c2e92c818abb0f58a8afa0526897234 so should be picked up going forward.  

In the near future this config will likely move to the melange configs themselves, so we can make sure update data exists when new packages are added.

rawlingsj (approved):

luhring:
@rawlingsj TIL, thanks!

Issue #542, opened 2023-02-07
[Wolfi Package Request]: yq


kameshsampath:
### What software would you like us to add to wolfi-os. Ideally include a URL to the project and its source.

As YAML user I would like to install [yq](https://github.com/mikefarah/yq) allowing me to manipulate YAMLs with ease

### which versions of the software should we include?

4.30.8

### Add some justification for why this specific package and versions are important.

yq is a very handy YAML utility that allows manipulating of YAML with ease. As Kubernetes developer I find it very handy with kube manifests e.g. patching.

### Please check all that apply

- [X] This package has an un-restrictive license
- [X] I am interested in adding this package to Wolfi OS myself.
- [X] I am willing to help maintain this package.

kameshsampath:
https://github.com/wolfi-dev/os/pull/543

PR #543, opened 2023-02-07
fix: add yq package

kameshsampath:
Add [yq](https://github.com/mikefarah/yq) package 

#542


yq.yaml
```diff
@@ -0,0 +1,27 @@
...
+  packages:
+    - busybox
+    - go
+pipeline:
+  - uses: go/install
+    with:
+      package: github.com/mikefarah/yq/v4
+      version: v${{package.version}}
+  - runs: |
+      ${{targets.destdir}}/usr/bin/yq --version | grep "version v${{package.version}}"
```
tuananh, line 26: testing?

kameshsampath: yeah, just want to make sure the right version of package is installed.

@tuananh ^^

tuananh:
@rawlingsj `go/install` support installing from a branch, `v4` in this case but it's kind of a moving target. the version doesn't quite make sense in this case.

IMO, the version should point toward a specific, non-moving target (specific commit hash, a specific file with static hash, etc..)

wdyt?

rawlingsj:
yeah agreed - we only tend to use tags where they are fixed as that can lead to un expected versions

kameshsampath:
@tuananh  and @rawlingsj  

In this case I am installing yq v4.30.8 --> https://github.com/mikefarah/yq/blob/v4.30.8/go.mod and the module name is github.com/mikefarah/yq/v4, so I used as package name and version `${{package.version}}`.   so ideally it will be translated to `go install github.com/mikefarah/yq/v4@v4.30.8` 

Is that something not right? Sorry I did not understand the issue here.

tuananh:
> @tuananh and @rawlingsj
> 
> In this case I am installing yq v4.30.8 --> https://github.com/mikefarah/yq/blob/v4.30.8/go.mod and the module name is github.com/mikefarah/yq/v4, so I used as package name and version `${{package.version}}`. so ideally it will be translated to `go install github.com/mikefarah/yq/v4@v4.30.8`
> 
> Is that something not right? Sorry I did not understand the issue here.

my bad. i thought version is 4. sorry

tuananh (commented):
lgtm

kameshsampath:
> > @tuananh and @rawlingsj
> > In this case I am installing yq v4.30.8 --> https://github.com/mikefarah/yq/blob/v4.30.8/go.mod and the module name is github.com/mikefarah/yq/v4, so I used as package name and version `${{package.version}}`. so ideally it will be translated to `go install github.com/mikefarah/yq/v4@v4.30.8`
> > Is that something not right? Sorry I did not understand the issue here.
> 
> my bad. i thought version is 4. sorry

no worries.

imjasonh (dismissed):

imjasonh (approved):

eddiezane:
Hey @kameshsampath,

Thanks for this contribution! I'd like to send some swag your way. If interested could you please shoot an email to eddiezane at chainguard dot dev?

PR #549, opened 2023-02-07
change python to use a version stream and add python 3.10

rawlingsj:
Signed-off-by: James Rawlings <jrawlings@chainguard.dev>
Signed-off-by: Eric Timmons <etimmons@asylonrobotics.com>

PR #550, opened 2023-02-07
fix python3-dev issue

imjasonh:
https://github.com/wolfi-dev/os/pull/549 should have updated `python-3.11`'s `-dev` subpackage to by named `python-3.11-dev` instead of `python3-dev`.


kaniini:
Needs epoch bump.  We also need to purge python 3.11.1-r4 packages.

kaniini (approved):

PR #567, opened 2023-02-11
add golangci-lint

developer-guy:

PTAL @dlorenc @jirfag


developer-guy:
@jirfag, it looks like golangci-lint requires `go` to be installed onto the environment, am I right? I'm asking because I've added `go` as a runtime dep of the golangci-lint package in wolfi.

dlorenc:
> @jirfag, it looks like golangci-lint requires `go` to be installed onto the environment, am I right? I'm asking because I've added `go` as a runtime dep of the golangci-lint package in wolfi.

That sounds plausible, it probably shells out to go tools.

ldez:
Hello, yes golangci-lint requires Go and also `diff`.

FYI, it's useless to ping Denis: he has not been active on the project for several years.
I'm the current active core maintainer.

I recommend not pinging directly maintainers especially when you don't know the person, as this can be considered harassment. Remember that you are not alone to do that, and managing a popular OSS is already taking a lot of time.

kaniini:
I do not believe there is any intent to annoy any upstream maintainers, but rather a desire to make them aware of the intent to include their package in the distribution, and to solicit any opinions about that intention.

ldez:
Don't take me wrong, I just trying to explain a recurring problem, it is neither against a person nor a project nor this PR. There is no major problem here, I'm trying to explain this just to improve the sanity of OSS maintainers.

Just a quick demo of the problem with this PR: there 3 messages each of them is pinging Denis.
After the first one, he can decide to unsubscribe but as the other messages are still continuing to ping him, he is forced to follow the topic.

I'm sure there are no bad intentions here but it is bad behavior.
I'm also sure that you can understand what I mean here.

Sorry for the off-topic, I hope I answered the original question in my previous message.

golangci-lint.yaml
```diff
@@ -0,0 +1,32 @@
...
+  description: Fast linters Runner for Go
+  target-architecture:
+    - all
+  copyright:
+    - license: Apache-2.0
+      paths:
+        - "*"
+  dependencies:
+    runtime:
+      - go
```
imjasonh, line 14: Did we also need to include `diff`?

Makefile
```diff
@@ -407,6 +407,7 @@ $(eval $(call build-package,vim,9.0.1275-r0))
 $(eval $(call build-package,policy-controller,0.6.2-r1))
 $(eval $(call build-package,syft,0.70.0-r0))
 $(eval $(call build-package,delve,1.20.1-r0))
+$(eval $(call build-package,golangci-lint,1.51.1-r0))
```
imjasonh, line 410: Pro-tip: I've found that you can avoid merge conflicts here if you put new packages randomly in the bottom ~5-10 lines, instead of the last line.

golangci-lint.yaml
imjasonh: It sounds like we should include `diff`.

imjasonh (changes requested):

PR #584, opened 2023-02-15
bump packages that depend on go

imjasonh:
Forgotten in https://github.com/wolfi-dev/os/pull/581


rawlingsj (dismissed):

kaniini (dismissed):

kaniini:
i think git-lfs is another example of `-extldflags` needing to be quoted.  See commit a22a370099a639718d882f49b8f35f0aaf6376c3 for how I did it.  Or we can jump on a call and figure it out this morning.

imjasonh:
> i think git-lfs is another example of `-extldflags` needing to be quoted. See commit [a22a370](https://github.com/wolfi-dev/os/commit/a22a370099a639718d882f49b8f35f0aaf6376c3) for how I did it. Or we can jump on a call and figure it out this morning.

Can you suggest an edit? I'm not sure I understand where the quotes should go for git-lfs.

tuananh:
I think it's because of this line in git-lfs source https://github.com/git-lfs/git-lfs/blob/77deabdf9a18f8a07ecfd3e0f4aa572ffbbab8d4/Makefile#L39

they do some manipulation with the LDFLAGS variable so if you want to use `make` to build it, you will have to override the LDFLAGS again like this, i think

```
export LDFLAGS="'$LDFLAGS'" 
```

another option is to not use `make` but keep do go build in `git-lfs.yaml`

tuananh:
@imjasonh you can pull this fix for git-lfs first and then this will work i think

https://github.com/wolfi-dev/os/pull/585

i need to maek this change because if EXTRA_LD_FLAGS is not passed in, they will use LDFLAGS as extldflags but do not quote it properly. so here I passed it in and quote it myself.

i also change from fetch to git-checkout so that buildinfo can have proper gitcommit information

```
EXTRA_LD_FLAGS="-extldflags '${LDFLAGS}'"
```

jdolitsky (dismissed):

jdolitsky (dismissed):

jdolitsky (approved):

PR #585, opened 2023-02-15
fix(git-lfs): quote extldflags

tuananh:
- switch from fetch to git-checkout so that we can use proper gitcommti in buildinfo
- pass EXTRA_LD_FLAGS to make so that we can proper quote extldflags ourselves.
- remove GIT_LFS_SHA="v${{package.version}}"  from make because we have proper gitcommit hash with `git-checkout`.


git-lfs.yaml
```diff
@@ -22,18 +22,22 @@ environment:
...
+      cd git-lfs
       # mitigate CVE-2021-38561
       go get golang.org/x/text
 
       # mitigate CVE-2022-27191
       go get golang.org/x/crypto/ssh
   - runs: |
-      make EXTRA_GO_FLAGS="-v -tags netcgo" GIT_LFS_SHA="v${{package.version}}" VERSION="v${{package.version}}"
-  - runs: |
+      cd git-lfs
```
imjasonh, line 40: It would be nice if git-checkout `cd`ed into there for us... 🤔

imjasonh (approved):

git-lfs.yaml
tuananh: i think whenever we create a new step with `runs`, it will be change back to workspace root?

kaniini: Yeah, new pipeline steps start at whatever `working-directory` is, which is by default the workspace root.  The packaging I did for .NET 6 shows how you can leverage this in interesting ways: https://gist.github.com/kaniini/0e7e90f23dc04b6db7d3cd30a030f1f7

PR #589, opened 2023-02-15
add withdraw-packages workflow

kaniini:


luhring:
Can we document somewhere "what this is" and "why it exists"?

.github/workflows/withdraw-packages.yaml
```diff
@@ -0,0 +1,50 @@
...
+            for pkg in $(cat withdrawn-packages.txt); do
+              echo "=> $pkg"
+              gsutil -m rm gs://wolfi-production-registry-destination/os/$arch/$pkg
+            done
+          done
+
+      - name: 'Sync public package repository'
+        run: |
+          mkdir "${{ github.workspace }}/packages"
+          gsutil -m rsync -r gs://wolfi-production-registry-destination/os/ "${{ github.workspace }}/packages/"
```
imjasonh, line 39: This copies a lot of bytes out of the repo when we really just need to scan the first few KB and get the total object size, which will make this a lot faster. Did you try `wolfictl generate-index`? It should work (🤞) and works against remote objects in GCS.

.github/workflows/withdraw-packages.yaml
```diff
@@ -0,0 +1,50 @@
...
+          find "${{ github.workspace }}/packages" -print -exec touch \{} \;
+
+      - name: 'Reconcile Wolfi index'
+        run: |
+          for arch in x86_64 aarch64; do
+            pushd "${{ github.workspace }}/packages/"$arch
+              melange index -o APKINDEX.tar.gz *.apk
+              melange sign-index --signing-key="${{ github.workspace }}/wolfi-signing.rsa" APKINDEX.tar.gz
+              gsutil -m rsync -r "${{ github.workspace }}/packages/${arch}/APKINDEX.tar.gz" gs://wolfi-production-registry-destination/os/${arch}/APKINDEX.tar.gz
```
imjasonh, line 48: I think we can just `gsutil cp APKINDEX.tar.gz` instead of rsync, which feels easier to footgun into writing other stuff we don't want.

.github/workflows/withdraw-packages.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+      - run: echo "${{ secrets.MELANGE_RSA }}" > ./wolfi-signing.rsa
+      - run: |
+          sudo mkdir -p /etc/apk/keys
+          sudo cp ./wolfi-signing.rsa.pub /etc/apk/keys/wolfi-signing.rsa.pub
+
+      - name: 'Delete withdrawn packages'
+        run: |
+          for arch in x86_64 aarch64; do
+            for pkg in $(cat withdrawn-packages.txt); do
```
imjasonh, line 30: FR: ignore empty lines and lines starting with # so we can include comments about why they're withdrawn, links to issues, etc.

kaniini:
> Can we document somewhere "what this is" and "why it exists"?

Added instructions.

.github/workflows/withdraw-packages.yaml
kaniini: It ran out of file descriptors while scanning the bucket.  While I agree that it would be better, this way works right now.

imjasonh (approved):

PR #593, opened 2023-02-16
fix: spire-server pkg

tuananh:
- switch to `git-checkout` because `make` use git tag & commit to add to buildinfo
- add `curl` in build deps because make try to install go with `curl`. i did not dig too deep into this to why spire wants to download another go version but if we want to keep using make, then curl is required.

Related: #584 . the issue is found on #584


imjasonh:
> * add `curl` in build deps because make try to install go with `curl`. i did not dig too deep into this to why spire wants to download another go version but if we want to keep using make, then curl is required.

The way I tried to fix this in https://github.com/wolfi-dev/os/pull/584 was to pin the go dep to the version they expect to build with, so they don't need `curl` to get another pre-built Go version. Another way would be to patch upstream's `.go-version` file to expect the one we provide, which might be better since they'll get Go fixes faster.

PR #611, opened 2023-02-19
attempt to fix the janitor

imjasonh:


.github/workflows/janitor-clusters.yaml
```diff
@@ -41,3 +40,12 @@ jobs:
         working-directory: ./scripts
         run: ./janitor-clusters.sh ${{ matrix.project }}
         shell: bash
+
+      # Notify on failures
```
imjasonh, line 44: This, and the `set -euo pipefail` in the script, prevents us from missing this in the future.

.github/workflows/janitor-clusters.yaml
```diff
@@ -26,12 +26,11 @@ jobs:
             workload_identity_provider: "projects/567187841907/locations/global/workloadIdentityPools/staging-shared-9bd2/providers/staging-shared-gha"
 
     steps:
-      - uses: google-github-actions/auth@ef5d53e30bbcd8d0836f4288f5e50ff3e086997d # v1.0.0
```
imjasonh, line 29: My theory is the v1 one didn't write the file to where the gcloud action looked for it. These versions match the one used in the workflow that creates the cluster.

dlorenc (approved):

PR #617, opened 2023-02-20
Bump the x/net dependency in git-lfs to mitigate a GHSA.

dlorenc:
I've built a git image with this and verified it fixes the GHSA in a grype scan.


imjasonh (approved):

imjasonh:
Are there other packages that depend on x/net that should also be updated?

git-lfs.yaml
```diff
@@ -36,6 +36,9 @@ pipeline:
 
       # mitigate CVE-2022-27191
       go get golang.org/x/crypto/ssh
+
+      # mitigate GHSA-vvpx-j8f3-3w6h
+      go get golang.org/x/net@v0.7.0
```
imjasonh, line 41: Do we need a `secfixes`/`advisory` entry for this? And the above?

luhring, line 41: Is there a risk to using the version here? Will this cause us to accidentally downgrade in the future?

dlorenc: I don't think we do secfixes for non-CVEs, but I could be mistaken.

dlorenc: I added a comment to the version number to tell the next person to try removing all of these. You're right we'd risk downgrading, but IMO that's better than getting a potentially new, unpinned version of the package each time we build.

luhring: Secfixes/advisories are only for vulns for `apk` packages, not `go` packages (i.e. modules). So no need to add anything here!

kaniini (approved):

PR #632, opened 2023-02-22
Add container-entrypoint.

amouat:
Please check:
 - `/bin/container-entrypoint.sh` is an ok location (I want it to be as short as possible)
 - does position in Makefile matter?

### DIFF
--- a/container-entrypoint.yaml
+++ b/container-entrypoint.yaml
@@ -23,7 +23,6 @@ pipeline:
       uri: https://github.com/wolfi-dev/container-entrypoint/archive/refs/tags/${{package.version}}.tar.gz
       expected-sha512: 32cea7963406d75cb6023626ac6a5969a6178ae94a49cc0400c80ff223a8d3874b85e4a67a8276faaf0bd39e4ae26c2708f6b0f4aef1064020490cf79643fd8c
   - runs: |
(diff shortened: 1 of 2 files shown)


container-entrypoint.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - ca-certificates-bundle
+      - busybox
+pipeline:
+  - uses: fetch
+    with:
+      uri: https://github.com/wolfi-dev/container-entrypoint/archive/refs/tags/${{package.version}}.tar.gz
+      expected-sha512: 32cea7963406d75cb6023626ac6a5969a6178ae94a49cc0400c80ff223a8d3874b85e4a67a8276faaf0bd39e4ae26c2708f6b0f4aef1064020490cf79643fd8c
+  - runs: |
+      mkdir -p ${{targets.destdir}}/bin
+      cp entrypoint.sh ${{targets.destdir}}/bin/
```
kaniini, line 27: why not

```
- runs: |
    install -Dm755 entrypoint.sh ${{targets.destdir}}/bin/
```

?

amouat: Because @rawlingsj didn't tell me to do that 😂 

There's no reason, I'll change it now.

kaniini (approved):

PR #640, opened 2023-02-23
Mitigate the x/net and x/text GHSA's in gomplate and dex.

dlorenc:
This gives us a clean dex image! Both of these bumps were done upstream but they haven't been released yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -419,8 +419,8 @@ $(eval $(call build-package,gnutar,1.34-r0))
 $(eval $(call build-package,dpkg,1.20.12-r0))
 $(eval $(call build-package,calicoctl,3.25.0-r0))
 $(eval $(call build-package,container-entrypoint,0.1.0-r0))
-$(eval $(call build-package,gomplate,3.11.3-r0))
-$(eval $(call build-package,dex,2.35.3-r0))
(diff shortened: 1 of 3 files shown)


rawlingsj:
This is probably a fine way to do this.  FWIW In the past I have made local changes to a `go.mod` file and created a `cve.patch` file using `git diff > cve.patch` then added that to the melange patch list as normal.  

@dlorenc just in case you prefer that approach, I don't have a preference either way but figured I'd mention it just in case.

rawlingsj:
Shall we go with the current approach in this PR for now?

dlorenc:
> This is probably a fine way to do this. FWIW In the past I have made local changes to a `go.mod` file and created a `cve.patch` file using `git diff > cve.patch` then added that to the melange patch list as normal.
> 
> @dlorenc just in case you prefer that approach, I don't have a preference either way but figured I'd mention it just in case.

I stole this pattern from `git-lfs` here: https://github.com/wolfi-dev/os/blob/main/git-lfs.yaml#L35 - I don't really have a strong preference either way either.

rawlingsj (approved):

rawlingsj:
Ah cool - yeah lets' go with that

PR #650, opened 2023-02-24
Add Python 3.12 prerelease.

dlorenc:


python-3.12.yaml
```diff
@@ -0,0 +1,106 @@
...
+  target-architecture:
+    - all
+  copyright:
+    - paths:
+        - "*"
+      attestation: TODO
+      license: PSF-2.0
+  dependencies:
+    provides:
+      - python3=3.12.999
```
strongjz, line 16: Will this break 3.11 as the default?

dlorenc: Yes! Good catch. I thought we had a system-python alias somewhere for the default but I don't see it.

python-3.12.yaml
```diff
@@ -0,0 +1,103 @@
+package:
+  name: python-3.12
+  # When bumping to a real non-prerelease you'll have to change the URL pattern on line 45
+  version: 3.12.0a5
```
kaniini, line 4: Version here should be `3.12.0_alpha5`.

kaniini (approved):

PR #665, opened 2023-02-26
add cilium-cli

developer-guy:
PTAL @tgraf


developer-guy:
I got the following error in pipeline:

```
x86_64-pc-linux-gnu/bin/ld: cannot find -lgcc: No such file or directory
```

it worked on my local machine but I couldn't make it work in pipeline, can you please help me folx @tgraf @rawlingsj @kaniini?

tuananh:
> I got the following error in pipeline:
> 
> ```
> x86_64-pc-linux-gnu/bin/ld: cannot find -lgcc: No such file or directory
> ```
> 
> it worked on my local machine but I couldn't make it work in pipeline, can you please help me folx @tgraf @rawlingsj @kaniini?

i think it has something to do with the 32bit arch. i didn't dig deep enough but maybe libgcc not build for 32 bit?

also , maybe not using `make local-release` because it's doing release for all combination of OS and arch but Wolfi is only doing amd64 and arm64 for now.

try this instead https://gist.github.com/tuananh/d5715ffab6af5eb7c9908c00f956a47b

cilium-cli.yaml
```diff
@@ -0,0 +1,48 @@
+package:
+  name: cilium-cli
+  version: 0.13.0
+  epoch: 0
+  description: cert-manager CLI tool to manage and configure cert-manager resources
```
margamanterola, line 5: This should be cilium rather than cert-manager

developer-guy:
> > I got the following error in pipeline:
> > ```
> > x86_64-pc-linux-gnu/bin/ld: cannot find -lgcc: No such file or directory
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > it worked on my local machine but I couldn't make it work in pipeline, can you please help me folx @tgraf @rawlingsj @kaniini?
> 
> i think it has something to do with the 32bit arch. i didn't dig deep enough but maybe libgcc not build for 32 bit?
> 
> also , maybe not using `make local-release` because it's doing release for all combination of OS and arch but Wolfi is only doing amd64 and arm64 for now.
> 
> try this instead [gist.github.com/tuananh/d5715ffab6af5eb7c9908c00f956a47b](https://gist.github.com/tuananh/d5715ffab6af5eb7c9908c00f956a47b)

that super make sense, thanks @tuananh !

cilium-cli.yaml
developer-guy: thank you @margamanterola, copy-paste issues 🤦

cilium-cli.yaml
```diff
@@ -0,0 +1,51 @@
...
+          GOARCH=amd64 CGO_ENABLED=0 go build \
+            -ldflags "-w -s -X 'github.com/cilium/cilium-cli/internal/cli/cmd.Version=${VERSION}'" \
+            -o release/cilium ./cmd/cilium
+          install -m755 -D release/cilium "${{targets.destdir}}"/usr/bin/cilium-cli
+      - if: ${{build.arch}} == 'aarch64'
+        runs: |
+          cd cilium-cli
+          export VERSION=$(git describe --tags --always)
+          mkdir -p release
+          GOARCH=amd64 CGO_ENABLED=0 go build \
```
lizrice, line 47:
```suggestion
          GOARCH=arm64 CGO_ENABLED=0 go build \

```

More generally, does this really need arch-specific builds? The [Makefile for Cilium CLI](https://github.com/cilium/cilium-cli/blob/master/Makefile) doesn’t have them

developer-guy: thanks @lizrice, I forgot the replace the GOARCH for the ARM build. 

> The [Makefile for Cilium CLI](https://github.com/cilium/cilium-cli/blob/master/Makefile?rgh-link-date=2023-02-28T15%3A58%3A20Z) doesn’t have them

https://github.com/cilium/cilium-cli/blob/25031f4bafb4d15e100d3a0fc3f66ad8d210a4a6/Makefile#L44, it seems that it has 🤔

lizrice: You're right on that, I guess I was looking at an old version, sorry!

developer-guy: no worries, thanks for helping me!

Makefile
```diff
@@ -453,6 +453,10 @@ $(eval $(call build-package,gradle-stage0,8.0.1-r0))
 $(eval $(call build-package,gradle-8,8.0.1-r0))
 $(eval $(call build-package,python-3.12,3.12.0_alpha5-r0))
 
+
+
+$(eval $(call build-package,cilium-cli,0.13.0-r0))
+
```
imjasonh, line 459:
```suggestion
$(eval $(call build-package,cilium-cli,0.13.0-r0))
$(eval $(call build-package,gradle-8,8.0.1-r0))
$(eval $(call build-package,python-3.12,3.12.0_alpha5-r0))

```

cilium-cli.yaml
imjasonh: Should this just run `make cilium` instead of duplicating its behavior here? If the Cilium project changes their Makefile I'd rather just get those updates for free rather than have to keep these in sync.

developer-guy: we tried to use `make cilium but it didn't work as we expected due to the 32-bit arch build, you can find these details in the below comments on why we switched from `make` to the present change.

cilium-cli.yaml
```diff
@@ -0,0 +1,51 @@
...
+  - uses: git-checkout
+    with:
+      repository: https://github.com/cilium/cilium-cli
+      tag: v${{package.version}}
+      destination: cilium-cli
+      expected-commit: ac2eca8005b7bdd677272845b5d0a6261e0e8a37
+  - assertions:
+      required-steps: 1
+    pipeline:
+      - if: ${{build.arch}} == 'x86_64'
```
imjasonh, line 33: Can we simplify this so it doesn't require conditionals, similar to https://github.com/wolfi-dev/os/pull/735 ? That might even help us with issues we had with `make cilium` 🤞

developer-guy: Of course, cool idea 👌

developer-guy: there is no target specific to the os/arch: _https://github.com/cilium/cilium-cli/blob/master/Makefile_

So, we cannot make it at the time of doing this.

cilium-cli.yaml
```diff
@@ -0,0 +1,51 @@
...
+          install -m755 -D release/cilium "${{targets.destdir}}"/usr/bin/cilium-cli
+      - if: ${{build.arch}} == 'aarch64'
+        runs: |
+          cd cilium-cli
+          export VERSION=$(git describe --tags --always)
+          mkdir -p release
+          GOARCH=arm64 CGO_ENABLED=0 go build \
+            -ldflags "-w -s -X 'github.com/cilium/cilium-cli/internal/cli/cmd.Version=${VERSION}'" \
+            -o release/cilium ./cmd/cilium
+          install -m755 -D release/cilium "${{targets.destdir}}"/usr/bin/cilium-cli
```
imjasonh, line 50:
```suggestion
  - runs: |
        cd cilium-cli
        export VERSION=$(git describe --tags --always)
        mkdir -p release
        GOARCH=$(go env GOARCH)
        CGO_ENABLED=0 go build \
          -ldflags "-w -s -X 'github.com/cilium/cilium-cli/internal/cli/cmd.Version=${VERSION}'" \
          -o release/cilium ./cmd/cilium
        install -m755 -D release/cilium "${{targets.destdir}}"/usr/bin/cilium-cli
```

Makefile
```diff
@@ -501,6 +501,10 @@ $(eval $(call build-package,json-c,0.16-r0))
 $(eval $(call build-package,opensearch-2,2.6.0-r0))
 $(eval $(call build-package,libmspack,0.11_alpha-r0))
 
+
+
+$(eval $(call build-package,cilium-cli,0.13.0-r0))
```
imjasonh, line 506: Can you add this to packages.txt instead?

imjasonh:
I don't think we need this anymore after https://github.com/wolfi-dev/os/blob/main/cilium-cli.yaml

PR #673, opened 2023-02-27
WIP: Add a scala3 package.

dlorenc:
This doesn't build from source yet, but follows the pattern from maven. I'll switch this to be the bootstrap compiler then built a real one.

Fixes: 

Related: 

#### For new package PRs only

- [ ] The package is available under an OSI-approved or FSF-approved license
- [ ] The version of the package is still receiving security updates

#### For security-related PRs

#### For version bump PRs


Makefile
```diff
@@ -454,6 +454,7 @@ $(eval $(call build-package,gradle-8,8.0.1-r0))
 $(eval $(call build-package,python-3.12,3.12.0_alpha5-r0))
 $(eval $(call build-package,doxygen,1.9.6-r0))
 $(eval $(call build-package,jbig2dec,0.19-r0))
+$(eval $(call build-package,scala3,3.2.2-r0))
```
imjasonh, line 457: You're going to get a merge conflict.

imjasonh (dismissed):

scala3.yaml
```diff
@@ -0,0 +1,39 @@
...
+  version: 3.2.2
+  epoch: 0  
+  description: The scala compiler
+  target-architecture:
+    - all
+  copyright:
+    - paths:
+        - "*"
+      attestation: TODO
+      license: Apache-2.0
```
imjasonh, line 12:
```suggestion
  copyright:
      license: Apache-2.0
```

scala3.yaml
```diff
@@ -0,0 +1,34 @@
+package:
+  name: scala3
+  version: 3.2.2
+  epoch: 0  
+  description: The scala compiler
+  copyright:
+      license: Apache-2.0
+  dependencies:
+    runtime:
+      - openjdk-17-jre
```
imjasonh, line 10: Should this be something we enforce, or something the user who's installing scala also needs to be responsible for?

Like, if someone wants to run scala3 with openjdk-11-jre (or openjdk-19 or something), is that a valid configuration we should allow?

PR #675, opened 2023-02-27
set tag and expected-commit for git-checkout pipelines

imjasonh:
https://github.com/wolfi-dev/wolfictl/pull/81 adds a lint check that requires `tag` and `expected-commit` when using `git-checkout`.

Also bump packages:
- rustup
- envoy
- vault
- aom
- kubescape
- gomplate

@found-it `ruby3.2-yajl-ruby` makes this a little annoying since it doesn't have the tag in the repo. We can either:
1. checkout some other tag (checking commit) then `git checkout` to the commit we want
2. use a `run` that `git clone`s instead of `uses: git-checkout`
3. relax the linter so that a missing `tag` is a warning instead of an error
4. maybe something else, ideas welcome

#### For version bump PRs


aom.yaml
```diff
@@ -28,7 +28,8 @@ pipeline:
   - uses: git-checkout
     with:
       repository: https://aomedia.googlesource.com/aom
-      branch: v${{package.version}}
+      tag: v${{package.version}}
+      expected-commit: b5fcaffbfd423cf49854805e5a6716e0c75e0c60
```
imjasonh, line 32:
```suggestion
      expected-commit: 3c65175b1972da4a1992c1dae2365b48d13f9a8d
```

luhring:
@imjasonh I see some packages with a `git-checkout` action that still don't have the `expected-commit` set, is that okay?

imjasonh:
> @imjasonh I see some packages with a `git-checkout` action that still don't have the `expected-commit` set, is that okay?

It's okay in that they won't fail any lint checks, because the lint check isn't enabled yet. But it's not okay in that I want to get rid of them before I enable the lint check 😆 

I went through again and found `terraform` lacking, were there others you found? (ruby yajl is expected, and will be `#nolint`ed)

luhring:
> I went through again and found `terraform` lacking, were there others you found? (ruby yajl is expected, and will be `#nolint`ed)

I thought I had seen more, but no, that's it! Looks good. 😄

luhring (approved):

PR #678, opened 2023-02-27
run withdraw-packages workflow automatically when list changes

imjasonh:


imjasonh:
Closing while we discuss whether we want this to be automatic.

PR #687, opened 2023-02-28
Add terraform package

nsmith5:
Terraform!

### DIFF
--- a/terraform.yaml
+++ b/terraform.yaml
@@ -1,0 +1,31 @@
+package:
+  name: terraform 
+  version: 1.3.9
+  epoch: 0
+  description: Terraform is a tool for building, changing, and versioning infrastructure safely and efficiently 
+  target-architecture:
+    - all
+    - !riscv64
+  copyright:
+    - license: MPL-2.0
+      paths:
+        - "*"
+environment:
+  contents:
+    packages:
+      - ca-certificates-bundle
+      - busybox
+      - go
+pipeline:
+  - uses: git-checkout
+    with:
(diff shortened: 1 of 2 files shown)


imjasonh (dismissed):

luhring (dismissed):

terraform.yaml
```diff
@@ -0,0 +1,32 @@
...
+    packages:
+      - ca-certificates-bundle
+      - busybox
+      - go
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/hashicorp/terraform
+      branch: v${{package.version}}
+      destination: terraform
```
tuananh, line 25: use tag instead of branch and `expected-commit`?

terraform.yaml
```diff
@@ -0,0 +1,32 @@
...
+  - runs: |
+      cd terraform
+      go generate ./...
+      CGO_ENABLED=0 go build -v
+  - runs: |
+      install -m755 -D ./terraform/terraform "${{targets.destdir}}/usr/sbin/terraform"
```
tuananh, line 31: why `sbin`?

terraform.yaml
nsmith5: Good idea!

terraform.yaml
nsmith5: Just some copy pasta from your Vault example. bin makes sense

imjasonh (approved):

terraform.yaml
tuananh: olps :D

PR #689, opened 2023-02-28
move lint for advisory to wolfictl lint

rawlingsj:


cpanato (approved):

developer-guy (dismissed):

cpanato (changes requested):
ok, looks weird

what is the difference between the wolfi-lint and the lint?

if both are the same you need to drop the entire job and not only the line

rawlingsj:
> if both are the same you need to drop the entire job and not only the line

They are slightly different, the `./lint.sh` script was added first but not all checks have been ported over yet.  There was a `wolfictl advisory sync` added since then but rather than using the `./lint.sh` script and compiling wolfictl from source in the GHA, lets use the image instead.

cpanato (dismissed):

.github/workflows/wolfictl-lint.yaml
```diff
@@ -18,4 +18,10 @@ jobs:
...
         entrypoint: wolfictl
-        args: lint
\ No newline at end of file
+        args: lint
+    - name: Advisory sync
+      id: advisory
+      uses: docker://ghcr.io/wolfi-dev/wolfictl:latest
+      with:
+        entrypoint: wolfictl
+        args: "advisory sync-secfixes --warn *.yaml"
```
luhring, line 27: @rawlingsj I think we can just drop the file part — the command should default to looking at all files in the current directory

```suggestion
        args: "advisory sync-secfixes --warn"
```

rawlingsj: YAY it passed and was super fast!

jdolitsky (approved):

PR #698, opened 2023-03-03
withdrawn-packages: withdraw python 3.12.0_alpha5-r{0,1}

kaniini:
These packages provided python3=3.12.999 and were subtly broken.


dlorenc (approved):
Sorry again for causing the break!

kaniini:
Happens to the best of us!

PR #709, opened 2023-03-06
all: simplify YAMLs

imjasonh:
This is not expected to have any functional changes, and **epochs are not updated.**

- remove `.copyright.attestation: TODO`, `""`, and unset -- only actual attestation values are retained
- remove `target-architecture` which is AFAICT [not used by melange at all](https://github.com/search?q=repo%3Achainguard-dev%2Fmelange%20TargetArchitecture&type=code), and was always `- "all"`.
- remove `copyright.paths: ["*"]`, which also seem to be unused by melange.
- remove empty runtime dependencies when I found them
- remove unnecessary whitespace when I found it

This change was _mostly_ automated, with some spot-checking where lint/parse errors occurred. In general this kind of change would probably have been easier using `yq` rather than `sed`. 🤷‍♂️ 

Assigning @kaniini to make sure these are safe changes.


jdolitsky (dismissed):

kaniini:
> * remove target-architecture which is AFAICT [not used by melange at all](https://github.com/search?q=repo%3Achainguard-dev%2Fmelange%20TargetArchitecture&type=code), and was always - "all".
> * remove copyright.paths: ["*"], which also seem to be unused by melange.

Lets not remove these.  Melange will make use of `target-architecture` eventually.  Likewise, the `paths` entries will be used in SBOMs, and also in APKv3 metadata.

luhring (dismissed):
YES!

imjasonh:
> Lets not remove these. Melange will make use of `target-architecture` eventually. Likewise, the `paths` entries will be used in SBOMs, and also in APKv3 metadata.

From our discussion offline it sounded like we'd be fine waiting until `target-architecture` is used, and assume/hope that we implement it so that the absence of a target architecture was assumed to mean "all".

For paths, it sounded like we were going to rely on SCA to detect licenses and the paths (#710), which means we won't need to set these at all eventually, and we won't need to set them for now.

Is there anything else I should do for this PR?

luhring (approved):

PR #711, opened 2023-03-06
Reject CVE-2010-4756 for glibc

luhring:
Thanks to @kaniini for her analysis!

Command run to generate this diff:

```shell
wolfictl advisory create ./glibc.yaml --sync \
  --status 'not_affected' \
  --justification 'vulnerable_code_not_present' \
  --impact 'Not a vulnerability. POSIX does not guarantee memory safety; resource limits should be appropriately set on applications which use untrusted inputs to glob. BSD GLOB_LIMIT extension is unreliable and was rejected by the GNU project.' \
  --vuln 'CVE-2010-4756'
```

### DIFF
--- a/glibc.yaml
+++ b/glibc.yaml
@@ -31,6 +31,7 @@ secfixes:
     - CVE-2019-1010023
     - CVE-2019-1010024
     - CVE-2019-1010025
+    - CVE-2010-4756
   2.36-r1:
     - CVE-2022-39046
   2.37-r1:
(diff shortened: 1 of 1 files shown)


glibc.yaml
```diff
@@ -436,6 +437,11 @@ subpackages:
             && mv "${{targets.destdir}}"/usr/lib/locale/${{range.key}} "${{targets.subpkgdir}}"/usr/lib/locale/
 
 advisories:
+  CVE-2010-4756:
+    - timestamp: 2023-03-06T12:47:28.379717-05:00
+      status: not_affected
+      justification: vulnerable_code_not_present
```
kaniini, line 443: I don't think this is a correct justification for this specific case (though I admit it is the closest one).  There is a vulnerability, but the vulnerability is in the consumer of the API in this case.  Maybe we should add additional justification types to OpenVEX?

luhring: This has come up before: https://github.com/wolfi-dev/os/blob/main/flex.yaml#L75

tl;dr: Yes there should be a new enum value here, but that's a larger issue. We have a few of these already, and we'll want to come back and address them all once there's a new status.

I'd vote for not letting that block this PR.

kaniini: Sounds good, just as long as we have an issue tracking adding an additional enum.

kaniini (approved):

PR #715, opened 2023-03-07
glibc: rebuild 2.37-r3 to remove final dependency cycle

kaniini:


dlorenc (approved):

luhring:
@kaniini Just so I'm following correctly, which exact Melange change/PR will result in this rebuild fixing the dependency cycle?

PR #716, opened 2023-03-07
Bump some go modules in traefik to fix two CVEs.

dlorenc:
I sent this same PR upstream: https://github.com/traefik/traefik/pull/9763

Fixes: 

Related: 

#### For new package PRs only

- [ ] The package is available under an OSI-approved or FSF-approved license
- [ ] The version of the package is still receiving security updates

#### For security-related PRs

#### For version bump PRs


kaniini (approved):

luhring:
This set the fixed version to a later version (`-r2`) than the version of the package itself (`-r1`). We should make it easier in `wolfictl` to avoid this. I've opened a bug for this: https://github.com/wolfi-dev/wolfictl/issues/97

PR #717, opened 2023-03-07
push packages to wolfi-production-registry-destination directly

kaniini:
the current approach with using a source bucket + reconcilation is not really getting us anything useful, but it *is* getting us a 2 hour waiting period on builds while the repository reconciles.

we should move to using `wolfictl` to reconcile the index and upload artifacts directly, but that requires further discussion.  this provides an immediate quality of life improvement for now.

### DIFF
--- a/.github/workflows/dag-push-production.yaml
+++ b/.github/workflows/dag-push-production.yaml
@@ -12,7 +12,7 @@ env:
   CLUSTER_ZONE: us-central1-b
   SERVICE_ACCOUNT: prod-images-ci
   FQ_SERVICE_ACCOUNT: prod-images-ci@prod-images-c6e5.iam.gserviceaccount.com
-  BUCKET: wolfi-production-registry-source/os/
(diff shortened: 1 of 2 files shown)


imjasonh:
What will be responsible for updating the APKINDEX in this case? 🤔

kaniini:
Melange itself updates the index, and signs it.  It has to in order to handle pre-requisites.

imjasonh:
> Melange itself updates the index, and signs it. It has to in order to handle pre-requisites.

Is that true even after https://github.com/wolfi-dev/os/pull/552 ?

kaniini:
> Is that true even after https://github.com/wolfi-dev/os/pull/552 ?

Yes, it is still generating the index.  The fact that #552 had no effect seems to be a bug in Melange, but that PR should not have been merged into Wolfi.  I wish I had known about it, as I would have pointed out that the index is necessary for prerequisite dependencies to be used in builds.

jdolitsky (approved):

PR #732, opened 2023-03-10
Bump Go versions and note vulnerability fixes

luhring:
### Note existing fixes for CVE-2022-41723

```shell
wolfictl advisory create ./go-1.19.yaml --sync --status 'fixed' --fixed-version '1.19.6-r1' --vuln 'CVE-2022-41723'
wolfictl advisory create ./go-1.20.yaml --sync --status 'fixed' --fixed-version '1.20.1-r1' --vuln 'CVE-2022-41723'
```

### Bump Go versions and note fixes for CVE-2023-24532

```shell
melange bump ./go-1.19.yaml 1.19.7
melange bump ./go-1.20.yaml 1.20.2
# plus Makefile updates

wolfictl advisory create ./go-1.19.yaml --sync --status 'fixed' --fixed-version '1.19.7-r0' --vuln 'CVE-2023-24532'
wolfictl advisory create ./go-1.20.yaml --sync --status 'fixed' --fixed-version '1.20.2-r0' --vuln 'CVE-2023-24532'
```


imjasonh (dismissed):

imjasonh:
Will we also want to bump all the packages that depend on go?

luhring:
>One thing I'm not sure on is if we need advisories for an upstream version update? I.e. we're not patching the vulnerability so didn't think we need to add an advisory as scanners will detect the correct library version and know it's not present?

The tl;dr here is that we **should** include these advisories in our data. For Trivy in particular: prior versions of our Go package were vulnerable, and these CVEs should be reported to users to they know to upgrade. But if we don't add this data to our secdb, Trivy won't surface those CVEs when it should.

With the upcoming work on systematizing vuln mgmt for Wolfi, it'll also be important to note these advisories for our metrics, so we can determine when exactly our packages had vulnerabilities.

luhring:
>Will we also want to bump all the packages that depend on go?

Done! Hopefully I did that right.

imjasonh (dismissed):

rawlingsj:
> The tl;dr here is that we should include these advisories in our data.

Ok perfect - thanks @luhring!

luhring:
(Kubescape's build might have been broken since #675 when its version was changed but not its `expected-commit`. Should be fixed in this PR... 🤞 )

imjasonh (approved):

PR #735, opened 2023-03-10
zot,oras: simplify build processes to avoid conditionals

imjasonh:
I'm not totally sure this will work, but I thought it would be worth a shot, to avoid having to duplicate config with conditionals.

Simplifying this way will help us expand to other archs more easily, hopefully without any new config changes required.

cc @developer-guy


oras.yaml
```diff
@@ -1,37 +1,31 @@
...
-          install -m755 -D ./bin/linux/amd64/oras "${{targets.destdir}}"/usr/bin/oras
-      - if: ${{build.arch}} == 'aarch64'
-        runs: |
-          cd oras
-          make build-linux-arm64
-          install -m755 -D ./bin/linux/arm64/oras "${{targets.destdir}}"/usr/bin/oras
+      destination: oras
+  - runs: |
+      cd oras
+      GOARCH=$(go env GOARCH)
```
tuananh, line 28: nice :)

developer-guy: nice trick!

developer-guy (approved):

PR #740, opened 2023-03-11
nodejs-18: fix provides for 18.5.0

imjasonh:
Missed this in #738


dlorenc (approved):
Do we need to withdraw the old one?

PR #750, opened 2023-03-13
Add nats nats-server and nsc.

dlorenc:
These are all required to build the full nats image.

Fixes: 

Related: 

#### For new package PRs only 

### DIFF
--- a/nats-server.yaml
+++ b/nats-server.yaml
@@ -1,0 +1,29 @@
+package:
+  name: nats-server
+  version: 2.9.15
+  epoch: 0
+  description: High-Performance server for NATS.io, the cloud and edge native messaging system.
+  copyright:
+    - license: Apache-2.0
+environment:
+  contents:
+    packages:
+      - wolfi-baselayout
+      - busybox
+      - go
+      - ca-certificates-bundle
+pipeline:
+  # We can't use go/install because this requires a specific go version
+  - uses: git-checkout
+    with:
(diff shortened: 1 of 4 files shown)


nats-server.yaml
```diff
@@ -0,0 +1,29 @@
...
+pipeline:
+  # We can't use go/install because this requires a specific go version
+  - uses: git-checkout
+    with:
+      repository: https://github.com/nats-io/nats-server
+      tag: v${{package.version}}
+      expected-commit: b91fa85462d42c2f988170aee27955773e68c56d
+  - runs: |
+      mkdir -p ${{targets.destdir}}/usr/bin
+      mkdir -p ${{targets.destdir}}/nats/conf
```
jdolitsky, line 24: /nats ? or should it be something like /etc/nats ?

dlorenc: This is the path used in the upstream docker image: https://github.com/nats-io/nats-server/blob/main/docker/Dockerfile.nightly

I'm not sure why, I'll have to dig in to see if it's a hardcoded default. It's definitely weird.

jdolitsky: sorry confused - are we saying switch to /etc or keep?

jdolitsky: (it looks like still in the root)

dlorenc: d'oh, never pushed the branch. nice catch.

jdolitsky (approved):

PR #751, opened 2023-03-13
ruby3.2: Update all packages to include runtime dependencies

found-it:
This PR updates all the ruby packages to pull in their runtime dependencies when installed.

Fixes: 

Related: 

#### For new package PRs only 

#### For security-related PRs

#### For version bump PRs


strongjz (dismissed):

ruby3.2-fluentd15.yaml
```diff
@@ -1,14 +1,25 @@
 package:
-  name: ruby3.2-fluentd
+  name: ruby3.2-fluentd15
   version: 1.15.3
-  epoch: 1
+  epoch: 2
```
jdolitsky, line 4: hm. I think epoch 0 makes sense here since we are changing the package name

jdolitsky (changes requested):

ruby3.2-fluentd15.yaml
found-it: Fair, it is technically a new package. Should I also withdraw the old fluentd packages since those will no longer be maintained?

jdolitsky: good question. probably shouldnt hurt

jdolitsky (dismissed):

jdolitsky (approved):

PR #756, opened 2023-03-14
bump openjdk versions

rawlingsj:
also moved the version metadata to the package.version so this action will be automated in the future by the wolfi bot

#### For version bump PRs


jdolitsky (approved):

openjdk-11.yaml
```diff
@@ -1,7 +1,8 @@
+#nolint:bad-version
```
luhring, line 1: What does this mean? Are we using invalid versions now?

(And if not, should we adjust how this lint rule works?)

luhring: @rawlingsj  (just curious)

imjasonh: This lint check ensures the version string is one apk understands.

luhring: Okay, thanks — so what does it mean for us to use a version that apk doesn't understand? (My naive mind says we shouldn't do this, actually. Is that a fair conclusion here?)

rawlingsj: Yeah this is how the JDKs are being versioned https://github.com/openjdk/jdk11u/tags 

Also unsure what to do here, my test last night before pushing the PR did see apk resolve a `+1` suffix, but just tested with more now and indeed do see subsequent suffix increments do not resolve.

i.e. this worked, apk does resolve 1.2.3+1 
1.2.3
1.2.3+1

this does not work, apk resolves 1.2.3+1 and not 1.2.3+2
1.2.3
1.2.3+1
1.2.3+2

Yeah I'm also not sure what our options are.  We could move the suffix into a var block which the wolfi bot / belange bump could also update when updating.
```
vars:
  version_suffix: "+5"
```
and the fetch URL then uses 
```
uri: https://github.com/openjdk/jdk11u/archive/refs/tags/jdk-${{package.version}}${{vars.version_suffix}}.tar.gz
```
But not sure what to do with the `package.version` in that case, so that apk resolves the correct version.

@kaniini do you have any ideas?
```

rawlingsj: Maybe we shouldn't be including versions that have this extra `+5` build metadata, into wolfi?  Though looking at adoptium they have the latest version as `jdk-17.0.6+10` https://adoptium.net/en-GB/temurin/releases/

PR #767, opened 2023-03-17
WIP enable so name CI check

rawlingsj:
This enables a CI check from wolfictl which compares proposed apk changes with it's latest version in wolfi to see if there is a breaking ABI change based on `.so` name files.

It also moved `wolfictl svg` to using a container to avoid compiling speed up builds.


rawlingsj:
hmm think I made an incorrect assumption here, cool, will need to fix this up.

PR #773, opened 2023-03-18
Change location of vault.hcl file in the vault package.

dlorenc:
I was trying to build a vault image, and placing this directly in /etc makes it hard to configure permissions for this file for a non-root user.

Fixes: 

Related:


dlorenc:
cc @tuananh - i think you set this up originally, do you have your own image for this somewhere? Will this cause you any issues? I'm wondering how you configured the permissions.

tuananh:
I did not use this yet since we're still skipping build for webui

https://github.com/wolfi-dev/os/pull/279#issuecomment-1360683067

the reason is because we did not have nodejs16 at the time. maybe we can start building webui now that nodejs16 pkg is available now.

kaniini (approved):

Issue #763, opened 2023-03-16
[Wolfi Package Request]: clamav


curantes:
### What software would you like us to add to wolfi-os. Ideally include a URL to the project and its source.

ClamAV® is an open source antivirus engine for detecting trojans, viruses, malware & other malicious threats

web: https://www.clamav.net/
source: https://github.com/Cisco-Talos/clamav
licence: GPL-2.0

### which versions of the software should we include?

1.0.1

### Add some justification for why this specific package and versions are important.

This is a well known antivirus and malware scanner, which is nice to have in an container to stream files through to scan for virus or malware.

### Please check all that apply

- [ ] This package has an un-restrictive license
- [X] The package/versions proposed are actively maintained upstream
- [X] I am interested in adding this package to Wolfi OS myself
- [ ] I am willing to help maintain this package

PR #782, opened 2023-03-18
libmilter: new package

curantes:
dependency for clamav

Related: #763 

#### For new package PRs only 


libmilter.yaml
```diff
@@ -0,0 +1,40 @@
...
+        DESTDIR="${{targets.destdir}}" \
+        install
+  - uses: strip
+subpackages:
+  - name: libmilter-dev
+    pipeline:
+      - uses: split/dev
+    dependencies:
+      runtime:
+        - libmilter
```
kaniini, line 39: This shouldn't be needed anymore.  Melange figures out the dependencies of -dev packages on its own now.

PR #783, opened 2023-03-18
Add mariadb 10.11.2 and set it as the default mariadb package.

dlorenc:
10.11 is the latest LTS now, we had 10.6 previously. That's still LTS, but 10.11 will now last longer and is more recent.

To make this swap, I added a new package and renamed the old one. The new package provides mariadb with no version specifier.

Fixes: 

Related: 

#### For new package PRs only 

#### For version bump PRs


mariadb-10.11.yaml
```diff
@@ -0,0 +1,219 @@
+package:
+  name: mariadb-10.11
+  # Latest LTS
+  version: 10.11.2 # When bumping to 10.6.13, the patch for CVE-2022-47015 can be removed.
```
imjasonh, line 4: This comment confuses me. When would we bump 10.11.2 (down) to 10.6.13?

edit: oh it's just copied.

mariadb-10.6.yaml
```diff
@@ -1,6 +1,5 @@
 package:
-  name: mariadb
-  # Latest LTS
+  name: mariadb-10.6
```
imjasonh, line 2: Do we want this to also provide `mariadb=10.6.999`? (and throughout for the subpackages)

mariadb-10.11.yaml
dlorenc: D'oh! The comment worked too (I removed the patch), but I didn't remove the comment :)

mariadb-10.6.yaml
dlorenc: Fixed!

imjasonh (approved):

PR #787, opened 2023-03-19
Fix golangci-lint config by changing diff to cmd:diff.

dlorenc:
We don't have a diff package, but we have diffutils which provides diff.


luhring:
> We don't have a diff package, but we have diffutils which provides diff.

Interesting, did this cause an issue? I had been thinking this field mapped to the "depends" field in the apk spec, where I thought the items didn't necessarily have to be the name of the package itself.

dlorenc:
> > We don't have a diff package, but we have diffutils which provides diff.
> 
> Interesting, did this cause an issue? I had been thinking this field mapped to the "depends" field in the apk spec, where I thought the items didn't necessarily have to be the name of the package itself.

Yeah, I ran into it here:

```shell
% docker run -it cgr.dev/chainguard/wolfi-base
ed53cc914829:/#
ed53cc914829:/# apk add golangci-lint
fetch https://packages.wolfi.dev/os/aarch64/APKINDEX.tar.gz
  diff (no such package):
    required by: golangci-lint-1.51.1-r1[diff]
```

dlorenc:
> > Interesting, did this cause an issue? I had been thinking this field mapped to the "depends" field in the apk spec, where I thought the items didn't necessarily have to be the name of the package itself.

Oooh, so I dug in a bit. It looks like it does get picked up by melange in the provides section of diffutils, but with a cmd: prefix:

```shell
% tar -xOf packages/aarch64/diffutils-3.9-r2.apk .PKGINFO
# Generated by melange.
pkgname = diffutils
pkgver = 3.9-r2
arch = aarch64
size = 341317
origin = diffutils
pkgdesc = GNU diff utilities
url =
commit = 1ec5b6ec792043fbcf3915dfc416e04907807fe6
license = GPL-3.0-or-later
depend = so:ld-linux-aarch64.so.1
depend = so:libc.so.6
provides = cmd:cmp=3.9-r2
provides = cmd:diff3=3.9-r2
provides = cmd:diff=3.9-r2
provides = cmd:sdiff=3.9-r2
datahash = f2a2ac312e3745d299c3dcabb919887880d57b131ea292fc9ddd3922f2de829f
```

But `apk add` doesn't support these prefixes directly: https://docs.alpinelinux.org/user-handbook/0.1a/Working/apk.html

So we can either change this to depend on diffutils directly, or to something like `cmd:diff`, not 100% sure if apko/melange support that though.

```shell
 % docker run -it cgr.dev/chainguard/wolfi-base
32ed63655325:/#
32ed63655325:/# apk add cmd:diff
fetch https://packages.wolfi.dev/os/aarch64/APKINDEX.tar.gz
(1/1) Installing diffutils (3.9-r2)
OK: 11 MiB in 12 packages
```

luhring:
>So we can either change this to depend on diffutils directly, or to something like cmd:diff, not 100% sure if apko/melange support that though.

Yeah "cmd:diff" is what I was thinking of! Having that be `golanci-lint`'s runtime dependency in the apk metadata seems like the "correct" solution to me.

It'd also be great if Melange could auto-detect and add command dependencies for packages like `golangci-lint`.

It seems like your fix would work, too. Maybe we merge this as-is but open an issue to look into this, WDYT?

dlorenc:
Swapped it to `cmd:`. We don't have any other packages using this format here in Wolfi so I'll give it a try to make sure it works.

dlorenc:
> Swapped it to `cmd:`. We don't have any other packages using this format here in Wolfi so I'll give it a try to make sure it works.

It works! I built a test image with apko and golangci-lint.

luhring (approved):

PR #799, opened 2023-03-20
check: fix license and split/infodir

curantes:
I got an error when trying to build clamav. Seems like a conflict in infodir for check and libffi packages. I noticed that the infodir files where not located in doc sub-package but in main package of check. This will fix the split of infodir. 

Also I noticed that the check package have the wrong license so correct this also.

```
Mar 20 20:19:20.461 [DEBUG] [arch:x86_64] [cmd:apk] [use-proot:false] [use-qemu:] (65/73) Installing libffi (3.4.4-r0)
Mar 20 20:19:20.461 [DEBUG] [arch:x86_64] [cmd:apk] [use-proot:false] [use-qemu:] ERROR: libffi-3.4.4-r0: trying to overwrite usr/share/info/dir owned by check-0.15.2-r2.
```


kaniini:
Actually, `/usr/share/info/dir` should just always be deleted.  We should add a check to melange for this to automatically do it.

kaniini (approved):

PR #836, opened 2023-03-24
DON'T MERGE: Minimize jre sub-packages

patflynn:
It looks like we're including too many modules for the JRE subpackage.

Related:
https://github.com/chainguard-images/images/issues/297

Only including the jenkins version bump to test the change.


kaniini (approved):

patflynn:
For the record. I don't think this is enough modules to cover the full set of JRE APIs. Jenkins works but I'm not sure we wouldn't end up breaking existing users.

PR #848, opened 2023-03-25
clamav: add clamav user and group

curantes:
- add clamav user and group since its required with scriptlets
- add clamav-scanner and clamav-daemon as runtime dependencies


curantes:
I am not user this is the correct way to add users/groups. Let me know if there is a better way to handle this.

imjasonh:
> I am not user this is the correct way to add users/groups. Let me know if there is a better way to handle this.

In our own usage of Wolfi in Chainguard Images, we're creating users/groups when we create the image using apko (for example, [here](https://github.com/chainguard-images/images/blob/cabe9ef9884a175f86102bd55fdc7050587f52d5/images/node/configs/latest.apko.yaml#L35-L44)). Apko doesn't run scriptlets at all, they're just ignored, so these are fine if you happen to be packaging using apko, you just need to make sure to _also_ configure those users/groups in the apko.yaml.

(If you're not using apko at all, that's fine too, and I think scriptlets are the best way to do this. @kaniini might have better ideas)

curantes:
> > I am not user this is the correct way to add users/groups. Let me know if there is a better way to handle this.
> 
> In our own usage of Wolfi in Chainguard Images, we're creating users/groups when we create the image using apko (for example, [here](https://github.com/chainguard-images/images/blob/cabe9ef9884a175f86102bd55fdc7050587f52d5/images/node/configs/latest.apko.yaml#L35-L44)). Apko doesn't run scriptlets at all, they're just ignored, so these are fine if you happen to be packaging using apko, you just need to make sure to _also_ configure those users/groups in the apko.yaml.
> 
> (If you're not using apko at all, that's fine too, and I think scriptlets are the best way to do this. @kaniini might have better ideas)

As of now I don't use apko but simply wolfi-base and apk add clamav and then this failes since it does not have the user and group installed. So if scriptlets are fine I guess this could be merged or are there any other issues with this pr?

PR #854, opened 2023-03-26
Create busybox-full package with mount/umount applets.

dlorenc:
I'm packaging the kubernetes secret-store-sci-driver and it needs "mount" available at runtime. I tested this with the latest melange/apko and it works.

Fixes:

Related:


dlorenc:
```shell
% docker run -it --entrypoint=sh test:latest-arm64
/ $
/ $ mount
overlay on / type overlay (rw,relatime,lowerdir=/var/lib/docker/overlay2/l/HKO3Y5A36V2ETS6NHDGFZJMZ2I:/var/lib/docker/overlay2/l/SJOQKDPG7LU3WCUPHJUTJZSMIH,upperdir=/var/lib/docker/overlay2/48aa32819f2f92f69b4f710cd7de7b5af92e498f8b789a4ca8e241080d720487/diff,workdir=/var/lib/docker/overlay2/48aa32819f2f92f69b4f710cd7de7b5af92e498f8b789a4ca8e241080d720487/work)
proc on /proc type proc (rw,nosuid,nodev,noexec,relatime)
tmpfs on /dev type tmpfs (rw,nosuid,size=65536k,mode=755)
devpts on /dev/pts type devpts (rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=666)
sysfs on /sys type sysfs (ro,nosuid,nodev,noexec,relatime)
cgroup on /sys/fs/cgroup type cgroup2 (ro,nosuid,nodev,noexec,relatime)
mqueue on /dev/mqueue type mqueue (rw,nosuid,nodev,noexec,relatime)
shm on /dev/shm type tmpfs (rw,nosuid,nodev,noexec,relatime,size=65536k)
/dev/vda1 on /etc/resolv.conf type ext4 (rw,relatime)
/dev/vda1 on /etc/hostname type ext4 (rw,relatime)
/dev/vda1 on /etc/hosts type ext4 (rw,relatime)
devpts on /dev/console type devpts (rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=666)
proc on /proc/bus type proc (ro,nosuid,nodev,noexec,relatime)
proc on /proc/fs type proc (ro,nosuid,nodev,noexec,relatime)
proc on /proc/irq type proc (ro,nosuid,nodev,noexec,relatime)
proc on /proc/sys type proc (ro,nosuid,nodev,noexec,relatime)
proc on /proc/sysrq-trigger type proc (ro,nosuid,nodev,noexec,relatime)
tmpfs on /proc/kcore type tmpfs (rw,nosuid,size=65536k,mode=755)
tmpfs on /proc/keys type tmpfs (rw,nosuid,size=65536k,mode=755)
tmpfs on /proc/timer_list type tmpfs (rw,nosuid,size=65536k,mode=755)
tmpfs on /sys/firmware type tmpfs (ro,relatime)
```

imjasonh (dismissed):

kaniini:
I had originally dropped mount/umount because of its usefulness in breaking out of a container.  Perhaps we should have an additional busybox package which contains these more dangerous applets?

dlorenc:
> I had originally dropped mount/umount because of its usefulness in breaking out of a container. Perhaps we should have an additional busybox package which contains these more dangerous applets?

That works for me - is there a way to package just the applets, or do we need to have two busybox-es?

dlorenc:
I did some playing around and apko currently hardcodes the name `busybox`: https://github.com/chainguard-dev/apko/blob/fab93a165dcd81b06923a3dd6943a88223c19407/pkg/build/busybox.go#L66

so if we want a separate package we'd need to refactor that too.

dlorenc:
I couldn't get it to work with apko without this change: https://github.com/chainguard-dev/apko/pull/604

dlorenc:
This requires https://github.com/chainguard-dev/apko/pull/604

imjasonh (approved):

PR #856, opened 2023-03-26
Add the GCP CSI driver package.

dlorenc:
We use this to build Wolfi itself!

Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/secrets-store-csi-driver-provider-gcp.yaml
+++ b/secrets-store-csi-driver-provider-gcp.yaml
@@ -1,0 +1,32 @@
+package:
+  name: secrets-store-csi-driver-provider-gcp
+  # When bumping the version check if the GHSA mitigations below can be removed.
+  version: 1.2.0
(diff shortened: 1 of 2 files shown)


secrets-store-csi-driver-provider-gcp.yaml
```diff
@@ -0,0 +1,32 @@
...
+environment:
+  contents:
+    packages:
+      - wolfi-baselayout
+      - busybox
+      - build-base
+      - go
+      - ca-certificates-bundle
+pipeline:
+  # We can't use go/install because this requires a specific go version
```
imjasonh, line 18: It does? Which version?

dlorenc: Whoops, fixed the comment. It was about specific ldflags to set the right version.

imjasonh (approved):

PR #861, opened 2023-03-27
use .tar.gz extension as .tar.bz2 are not generate in future versions

rawlingsj:
This means these packages can be updated by the wolfi bot in the future.


imjasonh (approved):
Is this worth a lint warning or something?

imjasonh (approved):
Is this worth a lint warning or something?

rawlingsj:
> Is this worth a lint warning or something?

Not sure TBH, I think it's still a valid extension for some packages, just not these three.  The main reason it was caught was because initially `.tar.bz2` were being created, but the bot updater fails for new versions.  When folks contribute new packages `.tar.bz2` wont be available so they would only use `.tar.gz`

rawlingsj:
> Is this worth a lint warning or something?

Not sure TBH, I think it's still a valid extension for some packages, just not these three.  The main reason it was caught was because initially `.tar.bz2` were being created, but the bot updater fails for new versions.  When folks contribute new packages `.tar.bz2` wont be available so they would only use `.tar.gz`

kaniini (approved):

kaniini (approved):

PR #866, opened 2023-03-27
don't cache APKINDEX objects in GCS

imjasonh:
Currently when we publish our APKINDEX, it gets cached for up to an hour, meaning that downstream tools like apko trying to install a new package won't see that new package, since they get an older cached index.

This changes our GCS upload to set the `cache-control` header to tell GCS and Cloud CDN not to cache these objects. This will affect both APKINDEX and the apks themselves, even though it should be fine to cache the apks as they don't change.

If we want to target this change better to only the index, we can (either now, or in the future).

Also uses `gcloud storage cp` instead of the `alpha` variant.

When we're done with this change we can do the same to the arm64 builds happening on GKE.

### DIFF
--- a/.github/workflows/push-production.yaml
+++ b/.github/workflows/push-production.yaml
@@ -72,7 +72,10 @@ jobs:
       - name: 'Upload the repository to a bucket'
         run: |
           cp /etc/apk/keys/wolfi-signing.rsa.pub ${{ github.workspace }}/packages/wolfi-signing.rsa.pub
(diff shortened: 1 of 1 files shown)


dlorenc (commented):
I dug this one up from sigstore: https://github.com/sigstore/root-signing/pull/168/files

imjasonh:
> I think caching apk content is desirable. Just not the APKINDEX.

Done.

This removes the `--recursive` and should only `cp` the APKINDEX (without caching) and `**/*.apk` (with caching). This means the public key won't be re-uploaded, but I don't think we need to do that anyway, right?

cpanato (approved):

kaniini (approved):

PR #872, opened 2023-03-28
OpenJDK: Add GA version

### LINKED ISSUE #598: [Wolfi Package Request]: openjdk-19

Introduce openjdk 19: https://github.com/openjdk/jdk19u

19

Like wolfi already provides openjdk 11 and 17: https://github.com/wolfi-dev/os/blob/main/openjdk-17.yaml

Dropping the idea here so I don't forget.

- [ ] This package has an un-restrictive license
- [X] I am interested in adding this package to Wolfi OS myself.
- [ ] I am willing to help maintain this package.

vr:
This is copied from openjdk-17 package to release more recent GA release

Fix https://github.com/wolfi-dev/os/issues/598

#### For new package PRs only
- [X] REQUIRED - The version of the package is still receiving security update

### DIFF
--- a/openjdk-19.yaml
+++ b/openjdk-19.yaml
@@ -1,0 +1,124 @@
+#nolint:bad-version
+package:
+  name: openjdk-19
+  version: 19.0.2.7
+  epoch: 1
+  description:
+  copyright:
+    - license: GPL-2.0-only
+  dependencies:
+    runtime:
+      - freetype
+      - fontconfig
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+      - automake
+      - autoconf
+      - file
+      - freetype-dev
+      - cups-dev
+      - libx11-dev
+      - libxext-dev
+      - libxrender-dev
+      - libxrandr-dev
+      - libxtst-dev
+      - libxt-dev
+      - alsa-lib-dev
+      - libffi-dev
+      - bash
+      - zip
+      - fontconfig-dev
+      - libxi-dev
+      - libjpeg-dev
+      - giflib-dev
+      - lcms2-dev
+
+var-transforms:
+  - from: ${{package.version}}
+    match: \.(\d+)$
+    replace: +$1
+    to: mangled-package-version
+
+pipeline:
+  - uses: fetch
+    with:
+      uri: https://github.com/openjdk/jdk19u/archive/refs/tags/jdk-${{vars.mangled-package-version}}.tar.gz
+      expected-sha256: 5903efd527dd08e9c235c8822e3d5699c3d18a8618c3e533307e8d6491ffbbf0
+
+# We need pre-built binaries to use as Boot JDK https://hg.openjdk.org/jdk/jdk12/raw-file/tip/doc/building.html#boot-jdk-requirements
+  - uses: fetch
+    with:
+      uri: https://github.com/adoptium/temurin19-binaries/releases/download/jdk-19.0.2%2B7/OpenJDK19U-jdk_x64_linux_hotspot_19.0.2_7.tar.gz
+      expected-sha256: 3a3ba7a3f8c3a5999e2c91ea1dca843435a0d1c43737bd2f6822b2f02fc52165
+      strip-components: 0
+  - uses: fetch
+    with:
+      uri: https://github.com/google/googletest/archive/release-1.8.1.tar.gz
+      expected-sha256: 9bf1fe5182a604b4135edc1a425ae356c9ad15e9b23f9f12a02e80184c3a249c
+      strip-components: 0
+  - runs: chmod +x configure
+  - uses: autoconf/configure
+    with:
+      opts: |
+        --with-boot-jdk=./jdk-19.0.2+7/ \
+        --with-vendor-name=wolfi \
+        --with-vendor-url=https://wolfi.dev \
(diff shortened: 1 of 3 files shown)


imjasonh:
`alsa-lib-dev` is defined [here](https://github.com/wolfi-dev/os/blob/382e2f90fdd099d140c187ef9bb9f06635c40203/alsa-lib.yaml#L37), and I see it exists in the repo at `alsa-lib-dev-1.2.8-r2.apk`.

In your local build, are you passing `--repository-append` and `--keyring-append` to add the Wolfi package repo to your builds? During CI builds we fetch packages locally so they don't need to be specified in the file, but when testing yourself you need these flags, which should be [set by the Makefile](https://github.com/wolfi-dev/os/blob/382e2f90fdd099d140c187ef9bb9f06635c40203/Makefile#L20).

vr:
OK this goes a bit further locally but i still can't test locally because it seems `var-transforms` block is not interpreted, so fetch url is not good

rawlingsj:
@vr do you have the latest melange?  `var-transforms` was added fairly recently.

vr:
I launch it with 
```
docker run --privileged --rm -v "${PWD}":/work cgr.dev/chainguard/melange build openjdk-19.yaml --signing-key melange.rsa --arch amd64 --repository-append=https://packages.wolfi.dev/os
```

vr:
OK i clone the `melange` repository to have it in `../melange` like the Makefile wants, i have `make melange` to have the binary, and now i launch `make packages/openjdk-19 BUILDWORLD=no ARCH=x86_64` and it ends up with :
```
2023/03/29 17:17:40 melange (openjdk-19/x86_64): --2023-03-29 15:17:40--  https://github.com/openjdk/jdk19u/archive/refs/tags/jdk-19.0.2+7.tar.gz
2023/03/29 17:17:41 melange (openjdk-19/x86_64): Resolving github.com... failed: Temporary failure in name resolution.
2023/03/29 17:17:41 melange (openjdk-19/x86_64): wget: unable to resolve host address 'github.com'
```

it just install everything before this so yes, i have internet connectivity 😞

vr:
Everything is ready i think, can we move on this ?

openjdk-19.yaml
```diff
@@ -0,0 +1,124 @@
...
+pipeline:
+  - uses: fetch
+    with:
+      uri: https://github.com/openjdk/jdk19u/archive/refs/tags/jdk-${{vars.mangled-package-version}}.tar.gz
+      expected-sha256: 5903efd527dd08e9c235c8822e3d5699c3d18a8618c3e533307e8d6491ffbbf0
+
+# We need pre-built binaries to use as Boot JDK https://hg.openjdk.org/jdk/jdk12/raw-file/tip/doc/building.html#boot-jdk-requirements
+  - uses: fetch
+    with:
+      uri: https://github.com/adoptium/temurin19-binaries/releases/download/jdk-19.0.2%2B7/OpenJDK19U-jdk_x64_linux_hotspot_19.0.2_7.tar.gz
```
rawlingsj, line 55: I wonder if we can use the existing `openjdk-17` package for the bootstrap rather than downloading the temurin binaries?  The idea is we want to build everything in wolfi from source.  There has been a need at times to use a binary to bootstrap a build but I'm wondering if we don't need that here and adding `openjdk-17` to the `environment.contents.packages` list would be enough?

rawlingsj:
BTW sorry for the slow feedback and also the latest merge conflicts.

openjdk-19.yaml
vr: We can use openjdk-17 to build openjdk-18, then use 18 to build 19, and so on, because you can't use 17 to build 19 or 20

vr: What we can do is : 
- I can add openjdk-18 the same way
- I can prepare another PR for when we will have proper package, to use them instead of temurin one.

What i'm wondering is if we really need to rebuild the Java binaries instead of just adding it to the image, is there some value added on this ?

patflynn: It's important to us that we have the capability to patch and customize if need be, and that the build environment for our packages is Wolfi-ized (up-to-date and minimal). That's not to say that we are against potentially providing binary packages of certain JDK distributions. For now though we're going to continue building one ourselves.

@rawlingsj WDYT about bootstrapping using the Temurin JDK n-1 binary for now, and once we get to packaging N-1 we can swap that out? 

@vr could you swap the jlink based JRE to using the submodule approach [here](https://github.com/wolfi-dev/os/pull/906/files)?

openjdk-20.yaml
```diff
@@ -0,0 +1,123 @@
...
+
+pipeline:
+  - uses: fetch
+    with:
+      uri: https://github.com/openjdk/jdk20u/archive/refs/tags/jdk-${{vars.mangled-package-version}}.tar.gz
+      expected-sha256: 215767be88a18af00440d5241960a3937237e77b938db4485bdb1c9bb414972e
+
+  - uses: fetch
+    with:
+      uri: https://github.com/adoptium/temurin20-binaries/releases/download/jdk-20%2B36/OpenJDK20U-jdk_x64_linux_hotspot_20_36.tar.gz
```
patflynn, line 54: presumably this won't work for the ARM builds :thinking:

kaniini:
We have OpenJDK 20 now.  Closing.

PR #897, opened 2023-03-29
Nack a CVE in ruby 3.1 and 3.0.

dlorenc:
The CVE is actually in rails, not ruby, but the CPE entry is a bit strange in the NVD.

Fixes:

Related:

#### For security-related PRs

- [X] The security fix is recorded in `annotations` and `secfixes`

### DIFF
--- a/ruby-3.0.yaml
+++ b/ruby-3.0.yaml
@@ -58,7 +58,6 @@ pipeline:
   - uses: autoconf/make
   - uses: autoconf/make-install
   - uses: strip
-
   # Remove gem bundler installed with ruby. It's provided by it's own package
   - runs: |-
       # Ignore the patch version of ruby since ruby will install under the
@@ -95,10 +94,17 @@ advisories:
     - timestamp: 2023-03-10T10:57:16.957642-05:00
       status: fixed
       fixed-version: 3.0.5-r0
+  CVE-2023-22795:
+    - timestamp: 2023-03-28T20:54:15.650462-04:00
+      status: not_affected
+      justification: component_not_present
 
 secfixes:
+  "0":
+    - CVE-2023-22795
   3.0.5-r0:
     - CVE-2021-33621
+
 update:
   enabled: false
   release-monitor:
--- a/ruby-3.1.yaml
+++ b/ruby-3.1.yaml
@@ -58,7 +58,6 @@ pipeline:
   - uses: autoconf/make
   - uses: autoconf/make-install
   - uses: strip
-
   # Remove gem bundler installed with ruby. It's provided by it's own package
   - runs: |-
       # Ignore the patch version of ruby since ruby will install under the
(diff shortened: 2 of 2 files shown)


ruby-3.0.yaml
```diff
@@ -58,8 +58,6 @@ pipeline:
   - uses: autoconf/make
   - uses: autoconf/make-install
   - uses: strip
-
-  # Remove gem bundler installed with ruby. It's provided by it's own package
```
luhring, line 62: (This is a bug in https://github.com/go-yaml/yaml that I want to fix for our usage)

luhring (dismissed):

luhring:
@dlorenc How did you find this?

dlorenc:
> @dlorenc How did you find this?

I clicked the rumble PDF and saw CVEs in these images: 
image


then ran grype - there are a bunch of false positives in gems then this one at the bottom: 
image

ruby-3.0.yaml
dlorenc: Oh whoops, should I try to put the comment back>? What's the easiest way to keep it without messing up automation?

luhring: I don't think we have a great answer.

If it's not too hard, it'd be nice to keep the comment, but if you run any of our YAML tools (wolfictl or melange) it'll happen again — they all use that YAML library.

I'm going to try to get this fixed in a fork in the next few days.

jdolitsky (approved):

luhring (changes requested):
I'm actually not seeing those vulns from the screenshot show up now (including the NAK-ed vuln).

```console
$ grype -q registry:cgr.dev/chainguard/ruby:3.0
NAME  INSTALLED  FIXED-IN  TYPE  VULNERABILITY   SEVERITY
cgi   0.2.2                gem   CVE-2021-41816  Critical
```

What are you seeing now @dlorenc?

dlorenc:
```shell
% grype cgr.dev/chainguard/ruby:3.0
 ✔ Vulnerability DB        [no update available]
New version of grype is available: 0.60.0 (currently running: 0.59.0)
 ✔ Loaded image
 ✔ Parsed image
 ✔ Cataloged packages      [107 packages]
 ✔ Scanning image...       [11 vulnerabilities]
   ├── 2 critical, 4 high, 4 medium, 1 low, 0 negligible
   └── 0 fixed
NAME      INSTALLED  FIXED-IN  TYPE  VULNERABILITY   SEVERITY
cgi       0.2.2                gem   CVE-2021-41816  Critical
find      0.1.0                gem   CVE-2015-10065  Critical
find      0.1.0                gem   CVE-2020-24550  Medium
readline  0.0.2                gem   CVE-2014-2524   Low
```

Looks like it got cleaned up. Maybe an upstream CPE fix?

luhring:
>Looks like it got cleaned up. Maybe an upstream CPE fix?

Yeah that's probably it!

(Also — I think you'll see more of those CVEs disappear from the Grype scan too if you update to latest)

Should we close this PR?

PR #912, opened 2023-03-30
fluent-bit: update to version 2.0.10

curantes:
- also add update section and point to release-monitoring.org

#### For version bump PRs


rawlingsj:
thanks @curantes! Love seeing the update config too :)

rawlingsj (approved):

PR #913, opened 2023-03-30
update makefile

deitch:
The existing makefile has a lot of repetitive calls to `$(eval $(call build-package,<pkgName>,<version>))`

issues:

* that is a lot of boilerplate to maintain for an ordered list, which can lead to errors
* It includes the version, which only exists to create the target `.apk` name, but already exists in the `.yaml`, which means it need to be edited in 2 places
* Includes `MELANGE_DIR` just to add the pipelines, which have long since been built into melange

Changes:

1. Replaces the long list of `$(eval $(call ...))` with a few functions and a list, which is easier to maintain
2. Includes custom directories as comma-separated, like before
3. Uses melange to determine the version from the yaml file, rather than hand-maintaining it
4. Removes `MELANGE_DIR` and pipelines; only thing left is to set `MELANGE` variable


kaniini (dismissed):

luhring (dismissed):
What a great idea 🤯

kaniini:
Sadly, CI is red.

Makefile
```diff
@@ -3,9 +3,9 @@ ARCH ?= $(shell uname -m)
 ifeq (${ARCH}, arm64)
 	ARCH = aarch64
 endif
+TARGETDIR = packages/${ARCH}
 
-MELANGE_DIR ?= ../melange
-MELANGE ?= ${MELANGE_DIR}/melange
+MELANGE ?= ../melange/melange
```
tuananh, line 8: why dont we just use $(which melange) here?

kaniini: Historical reasons.  Those of us using Melange natively are usually running it alongside the wolfi checkout.

tuananh: I'm just afraid this will be a bit unfriendly for new contributors. where they will need to have `melange` repo at the same level as `wolfi` repo. also, if their melange repo is staled, it could become problem for them, no?

deitch: I agree it is a bit unfriendly, while also being convenient for using melange natively, although this PR makes that neither worse nor better.

I see two options:

A. Default to `MELANGE=$(which melange)` and have people using it natively override that with `make MELANGE=../melange/melange`
B. Default to `MELANGE=../melange/melange` and have people who do not have it checked it override it with `make MELANGE=$(which melange)`

My personal preference is B, but I didn't want to change too much UX in here without agreement from other folks.

There is a third possibility, i.e.

1. If `MELANGE` is set use that; else
2. If `../melange/melange` exists use that; else
3. If `$(which melange)` exists use that; else
4. Error out

Thoughts?

deitch:
> Sadly, CI is red.

Sadly indeed. Looking.

deitch:
Rebased to catch latest Makefile (in order, of course).

I updated the lint as well.

The remaining piece is going to be `wolfictl lint`, which is hard-coded to the format of the Makefile. Having that in a separate repo and hard-coded to the makefile format is not great either. 

I changed lint to, instead, run `make list-yaml` and then it can use the output of that to see if the package is there or not. This decouples the internal structure of the Makefile from the listing of what packages it supports.

I propose that we accept that one `wolfictl lint` failure one time and, when everything else is ready, merge this in. Then we can change `wolfictl` to do the same thing, i.e. call `make list` or `list-yaml` and check the package list that way.

deitch:
I was also thinking: do we want to keep that list of packages in the Makefile? Or should we just have a separate file with them, so that changes to the Makefile don't risk upsetting the order, and changes to the list don't impact the Makefile.

It is pretty easy to include the list by:

```makefile
PKGS := $(shell cat packages.txt)
```

Makefile
deitch: Actually, as I think about it, every time something gets updated with the Makefile packages, I have to do a manual rebase. 

I would like to get this in as is; happy to address the `MELANGE` var separately afterwards.

deitch:
`wolfictl lint` still is failing. I stand by my earlier suggestion: let it sail for now, as soon as this is in, I will update `wolfictl` to run `make list` and check packages that way.

deitch:
Does anyone understand the `wolfctl check update` CI failures? There are 10 of them:

* 2 are about malformed versions from python-3.12 (`3.12.0_alpha6`) and openssh (`9.3_p1`)
* 8 are about `found newer version _____ compared with package.version in melange config`

rawlingsj:
similar to this https://github.com/wolfi-dev/os/pull/956#issuecomment-1494144059 

the update check is not a required check for now while we rollout the auto update configs

deitch:
So I can ignore it @rawlingsj ? Cool. Then all that is left is the lint, with my proposal above.

Well, a maintainer does need to agree that this should go in and approve it. That too. 😁

kaniini (approved):

PR #918, opened 2023-03-31
restic: patch CVE-2022-41723

jamonation:
Patches CVE-2022-41723 in restic 0.15.1. Upstream has [updated `go.mod` package versions](https://github.com/restic/restic/blob/5f13bbc118232c93c1c3bab57d2723b08a58161f/go.mod#L28) but haven't cut a release yet.

#### For security-related PRs
- [x] The security fix is recorded in `annotations` and `secfixes`


restic.yaml
```diff
@@ -18,9 +18,22 @@ pipeline:
     with:
       uri: https://github.com/restic/restic/archive/refs/tags/v${{package.version}}.tar.gz
       expected-sha256: fce382fdcdac0158a35daa640766d5e8a6e7b342ae2b0b84f2aacdff13990c52
+  - uses: patch
```
imjasonh, line 21: I think we can do this even easier. Elsewhere in Wolfi we'll do a `go get -u ...` to pick up new versions, without having to manually patch the `go.mod`.

For example: https://github.com/wolfi-dev/os/blob/7a8f38ba7044ac453c20d8c5c0930166cc4717b6/dex.yaml#L26-L29

If you do this, it's also nice to put a note near the version that once a new version is released, you can remove the `go get`s below, e.g., https://github.com/wolfi-dev/os/blob/7a8f38ba7044ac453c20d8c5c0930166cc4717b6/dex.yaml#L3

jamonation: Ah that's a great help, thanks for the pointers! I'm assuming I should still leave in the advisories and secfixes?

luhring: Yep! Thanks @jamonation!

luhring (approved):

PR #931, opened 2023-03-31
vault: Build ui + bump vault version to 1.13.1

tuananh:
- Start building ui component in vault since we have `nodejs-16` now.
- Bump version to 1.13.1

historical reason why we didn't do it before: https://github.com/wolfi-dev/os/pull/279#issuecomment-1360683067

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs


vault.yaml
```diff
@@ -30,8 +32,8 @@ pipeline:
         make "$plugin"
       done
 
-      # Skip with web ui build since it needs Nodejs ver 16 explicitly.
-      _gotags="vault"
```
imjasonh, line 34: Genuinely curious: is it possible to package the ui-having variant as a subpackage? Does that even make sense? I could imagine someone only wanting the server components and not the UI, but I don't know if that's a real scenario since I've never used Vault.

I'll defer to you completely, whatever your answer is 😄

tuananh: I'm not sure how to split them out since they build it and bundle it into the binary :-/

dlorenc (approved):

vault.yaml
dlorenc: This matches upstream, so I'm fine with it.

PR #933, opened 2023-03-31
Add envoy ratelimit package.

dlorenc:
They don't actually tag upstream releases, so I'm not sure I did the version right. This is the latest commit.

Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/ratelimit.yaml
+++ b/ratelimit.yaml
@@ -1,0 +1,15 @@
+package:
+  name: ratelimit
+  # This project doesn't do releases and everything is commit based.
+  # This corresponds to commit 008b66a
+  version: "0.0_git20230331"
+  epoch: 0
+  description: Go/gRPC service designed to enable generic rate limit scenarios from different types of applications.
+  copyright:
+    - license: Apache-2.0
+pipeline:
+  - uses: go/install
+    with:
+      package: github.com/envoyproxy/ratelimit/src/service_cmd@008b66a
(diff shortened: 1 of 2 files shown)


ratelimit.yaml
```diff
@@ -0,0 +1,15 @@
...
+  # This corresponds to commit 008b66a
+  version: "0.0_20230331"
+  epoch: 0
+  description: Go/gRPC service designed to enable generic rate limit scenarios from different types of applications.
+  copyright:
+    - license: Apache-2.0
+pipeline:
+  - uses: go/install
+    with:
+      package: github.com/envoyproxy/ratelimit/src/service_cmd@008b66a
```
luhring, line 13: Can we avoid hardcoding the version here to help with our auto-updating?

dlorenc: Hmm.. how else would we do it? If we pin to nothing we'd get non-reproducible builds.

luhring: Yeah, I was thinking there'd be a way to pin to a Melange variable instead of to a literal commit.

But it seems like we have a limitation preventing us from encoding the commit information in the version, so we'd need a way to resolve the commit when we do updates, which isn't a thing yet AFAIK.

So in short, nvm!

luhring (approved):

imjasonh (approved):

PR #937, opened 2023-03-31
Nack CVE-2023-24329 in Python 3.10.

dlorenc:
This one is an interesting read. The cypython maintainers argue that this is not a security issue in Python itself, but could present vulnerabilities in libraries that use the function in an unsafe manner.

Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/python-3.10.yaml
+++ b/python-3.10.yaml
@@ -10,6 +10,7 @@ package:
 secfixes:
   "0":
     - CVE-2007-4559
+    - CVE-2023-24329
   3.10.9-r0:
     - CVE-2020-10735
 environment:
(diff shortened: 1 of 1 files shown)


imjasonh:
There appears to be some fairly active discussion on this (https://github.com/python/cpython/issues/102153), let's keep this un-NACKed for now and see where that ends up.

luhring (changes requested):
Let's hold this from merging for now, and check back soon to see if there's been progress upstream. If there's no progress on the upstream discussion soon, I think this would be reasonable to NAK.

luhring (dismissed):
I think we can proceed with this now: we've waited a very reasonable amount of time but it looks like the upstream conversation isn't progressing meaningfully, unfortunately.

luhring:
@dlorenc Think we need to get those last two required CI checks to run somehow

dlorenc:
> @dlorenc Think we need to get those last two required CI checks to run somehow

I'll rebase and force push, thanks!

luhring (approved):

PR #948, opened 2023-04-02
Withdraw gradle-stage0 and rebuild it with gradle 7 instead of 8.

dlorenc:
Gradle 8 can't be used to bootstrap 7, so we need to drop our bootstrap build back. If we don't withdraw the existing package, version 8 will take precendence over 7.

Fixes:

Related:

#### For version bump PRs


dlorenc:
After this, I'll send a PR to rebuild gradle-8 and add a gradle-7 package, bootstrapped from this.

kaniini (dismissed):

Makefile
```diff
@@ -467,7 +467,7 @@ $(eval $(call build-package,ruby3.2-protocol-hpack,1.4.2-r1))
 $(eval $(call build-package,ruby3.2-yajl-ruby,1.4.3-r1))
 $(eval $(call build-package,gomplate,3.11.4-r1))
 $(eval $(call build-package,dex,2.35.3-r2))
-$(eval $(call build-package,gradle-stage0,8.0.1-r0))
+$(eval $(call build-package,gradle-stage0,7.6.1-r0))
```
imjasonh, line 470: This change shouldn't be needed now that we're using packages.txt

gradle-stage0.yaml
```diff
@@ -9,7 +9,7 @@ package:
     runtime:
       - openjdk-11-jre
     provides:
-      - gradle=8.0.1 # hack for self bootstrapping
+      - gradle=${{package.version}} # hack for self bootstrapping
```
imjasonh, line 12:
```suggestion
      - gradle=${{package.version}}
```

PR #949, opened 2023-04-02
Add missing python3 dependency to vault package.

dlorenc:
Some how arm builds are failing without this.

Fixes:

Related:


dlorenc:
After https://github.com/wolfi-dev/os/pull/931 all arm builds started failing.

imjasonh (approved):

PR #956, opened 2023-04-02
Also add no-store to the withdraw job.

dlorenc:
Follow up to https://github.com/wolfi-dev/os/pull/955

Fixes:

Related:

### DIFF
--- a/.github/workflows/withdraw-packages.yaml
+++ b/.github/workflows/withdraw-packages.yaml
@@ -55,6 +55,6 @@ jobs:
             pushd "${{ github.workspace }}/packages/"$arch
               melange index -o APKINDEX.tar.gz *.apk
               melange sign-index --signing-key="${{ github.workspace }}/wolfi-signing.rsa" APKINDEX.tar.gz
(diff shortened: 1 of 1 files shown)


dlorenc:
cc @rawlingsj - I think this failure is a bug.

rawlingsj:
yeah we can ignore these failures
```
	* failed to create a version from package python-3.12: 3.12.0_alpha6: Malformed version: 3.12.0_alpha6
	* failed to create a version from package openssh: 9.3_p1.  Error: Malformed version: 9.3_p1
```
there's two open issues to cover them
https://github.com/wolfi-dev/os/issues/920
https://github.com/wolfi-dev/os/issues/923

The check is not a github required check so this should merge...

rawlingsj:
hmm that's annoying - even though the update check isn't a "required" check in the branch protection rules it still seems to block.  Let's merge this by overriding and will need to find a better way.

rawlingsj (approved):

rawlingsj:
Oh maybe it would have auto merged - it just wasn't approved.

PR #957, opened 2023-04-03
Add upstream packaged nginx.confs

amouat:
Adds configs found in the upstream pacakges, but not in the source repo. See https://hg.nginx.org/pkg-oss/

Related: https://github.com/chainguard-images/images/issues/382

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/nginx.yaml
+++ b/nginx.yaml
@@ -1,7 +1,7 @@
 package:
   name: nginx
   version: 1.23.4
-  epoch: 1
+  epoch: 0
   description: HTTP and reverse proxy server (stable version)
   copyright:
     - license: BSD-2-Clause
--- a/nginx.yaml
+++ b/nginx.yaml
@@ -132,3 +132,10 @@ subpackages:
                 install -m644 -D stream.conf ${{targets.subpkgdir}}/etc/nginx/stream.d/
               ;;
             esac
+    - name: nginx-package-config
+      description: Config supplied in packaged nginx.org bundles at https://hg.nginx.org/pkg-oss/
+      pipeline:
+        - runs: |
+            install -m644 -D nginx.conf ${{targets.subpkgdir}}/etc/nginx/
+            mkdir -p ${{targets.subpkgdir}}/etc/nginx/conf.d
+            install -m644 -D default.conf ${{targets.subpkgdir}}/etc/nginx/conf.d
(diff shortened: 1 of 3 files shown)


nginx.yaml
```diff
@@ -132,3 +134,15 @@ subpackages:
...
               ;;
             esac
+    - name: nginx-package-config
+      description: Config supplied in packaged nginx.org bundles at https://hg.nginx.org/pkg-oss/
+      pipeline:
+        - runs: |
+            mkdir -p ${{targets.subpkgdir}}/etc/nginx/conf.d
+            install -m644 -D nginx.conf ${{targets.subpkgdir}}/etc/nginx/
+            install -m644 -D nginx.default.conf ${{targets.subpkgdir}}/etc/nginx/conf.d
+            # Follow Docker Image and link logs to to stdout/stderr
```
imjasonh, line 144: Can we link to a source of truth that the Docker image does this? This seems like it would be helpful if we're debugging this a year from now.

nginx/nginx.conf
```diff
@@ -0,0 +1,32 @@
+
```
imjasonh, line 1:
```suggestion
# Based on https://hg.nginx.org/pkg-oss/file/fd9484abcae4/alpine/nginx.conf
```

nginx/nginx.default.conf
```diff
@@ -0,0 +1,44 @@
+server {
```
imjasonh, line 1:
```suggestion
# https://hg.nginx.org/pkg-oss/file/fd9484abcae4/alpine/nginx.default.conf

server {
```

nginx.yaml
amouat: Argh, you've led me to another rabbit hole. This happens in the Debian (latest) image:

https://github.com/nginxinc/docker-nginx/blob/73a5acae6945b75b433cafd0c9318e4378e72cbb/mainline/debian/Dockerfile#L103-L104

It doesn't happen in the alpine image it seems, so I don't currently understand how it works.

amouat: Done

amouat:
Note I have combined a patch update with this PR. Most changes are to a new subpackage, hence non-breaking. That being said, the default access low routes to /dev/stdout, which I consider a bugfix.

rawlingsj (approved):

PR #997, opened 2023-04-04
add update config for more packages

rawlingsj:


luhring (dismissed):
🆙 📅

rawlingsj:
The red `wolfi check updates` is ok, packages that have older versions reported will be updated once this merges, the one that fails will have an issue created.

This is ok to merge.

imjasonh (approved):

PR #1010, opened 2023-04-05
another 97 package update configs, that's all github related ones

rawlingsj:


rawlingsj:
The failing "check updates" are ok as the versions will be auto created when the bot runs next.  The other one failure is fixed in https://github.com/wolfi-dev/wolfictl/pull/145 so will not cause issues.

libspf2.yaml
```diff
@@ -48,3 +47,5 @@ subpackages:
             mv "${{targets.destdir}}"/usr/bin/spfquery "${{targets.subpkgdir}}"/usr/bin
             mv "${{targets.destdir}}"/usr/bin/spfd "${{targets.subpkgdir}}"/usr/bin
             rm -fr "${{targets.destdir}}"/usr/bin
+update:
+  enabled: false # to tags or releases
```
luhring, line 51: Just for my education, what does this comment mean?

rawlingsj: ah yeah - so just means there's no tags or releases available https://github.com/shevek/libspf2/releases.  The package is using a commit sha https://github.com/wolfi-dev/os/blob/26a31dcbabffc3bac873363c7759511e117cca47/libspf2.yaml#LL21C14-L21C47 so not able to auto update it.

imjasonh (approved):

libspf2.yaml
luhring: Ah okay thanks! so maybe `s/to/no/`?

PR #1105, opened 2023-04-08
Nack CVE-2023-0466 in openssl.

dlorenc:
The upstream maintainers decide to not change the behavior here and instead update the documentation. Ref: https://www.openssl.org/news/secadv/20230328.txt

Fixes:

Related:

#### For security-related PRs

### DIFF
--- a/openssl.yaml
+++ b/openssl.yaml
@@ -9,6 +9,8 @@ package:
     runtime:
 
 secfixes:
+  "0":
+    - CVE-2023-0466
   3.0.7-r0:
     - CVE-2022-3358
     - CVE-2022-3602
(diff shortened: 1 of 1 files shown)


openssl.yaml
```diff
@@ -165,6 +167,12 @@ advisories:
     - timestamp: 2023-03-28T07:54:27.093515-07:00
       status: fixed
       fixed-version: 3.1.0-r2
+  CVE-2023-0466:
+    - timestamp: 2023-04-08T12:32:54.797413-04:00
+      status: not_affected
+      justification: component_not_present
```
luhring, line 173: Just so I'm following, why did you choose this justification?

openssl.yaml
```diff
@@ -165,6 +167,12 @@ advisories:
...
+      impact: The upstream behavior was deemed correct, and the documentation was updated instead of a code change
```
luhring, line 174: IIUC, something like this (feel free to tweak) would be more accurate in this case:

```suggestion
      impact: The upstream maintainers decided to document a workaround rather than make a code change that would potentially break existing users.
```

luhring (changes requested):

openssl.yaml
luhring: Got it!

openssl.yaml
luhring: Nice work 👏

luhring (approved):
LGTM

PR #1164, opened 2023-04-11
Add elfutils and libbpf packages.

dlorenc:
These caught an issue in apko which we're working on tests for now, so these should build but not necessarily work yet until that's fixed.

Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/libbpf.yaml
+++ b/libbpf.yaml
@@ -1,0 +1,40 @@
+package:
+  name: libbpf
+  version: 1.1.0
+  epoch: 0
+  description: "A library for interacting with the Linux kernel's Berkeley Packet Filter (BPF) facility from user space"
+  copyright:
+    - license: GPL-2.0-only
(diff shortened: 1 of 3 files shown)


jdolitsky (dismissed):

libbpf.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - ca-certificates-bundle
+      - build-base
+      - elfutils-dev
+      - linux-headers
+      - zlib-dev  
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/libbpf/libbpf
+      tag: v${{package.version}}
```
jdolitsky, line 23: maybe need to drop the "v" here?
```suggestion
      tag: ${{package.version}}
```

jdolitsky (approved):

PR #1165, opened 2023-04-11
Filter dotnet update versions to 1xx series

luhring:
Based on comment here: https://github.com/wolfi-dev/os/pull/1127#issuecomment-1503509550

cc: @kaniini @rawlingsj

### DIFF
--- a/dotnet-7.yaml
+++ b/dotnet-7.yaml
@@ -135,4 +135,4 @@ update:
     identifier: dotnet/installer
     strip-prefix: v
     use-tag: true
-    tag-filter: "v7."
+    tag-filter: "v7.0.1" # We specifically the 7.0.1xx releases since those are intended for prod.


rawlingsj (approved):

rawlingsj:
the wolfii check update job has the correct version
```
2023/04/11 14:54:35 attempting to bump version to 7.0.104
2023/04/11 14:54:35 processing git-checkout node
2023/04/11 14:54:35   expected-commit: 5c12d17216f411669c0beb40956249ec48bcb818
2023/04/11 14:54:35 wolfictl check update: cloning sources from https://github.com/dotnet/installer tag v7.0.104 into a temporary directory, this may take a while
```

tuananh:
the tag filter is just prefix check right now right? how about we parse them as semver properly?

of if you're worried that upstream may not do proper semver tagging, maybe we can do regex constraint? it probably looks more complicated though.

luhring:
> the tag filter is just prefix check right now right? how about we parse them as semver properly?
> 
> of if you're worried that upstream may not do proper semver tagging, maybe we can do regex constraint? it probably looks more complicated though.

Good questions. I defer to @rawlingsj. My take is that the prefix approach is a reasonable start, and seems to be handling all the cases I've seen so far. But I'm sure we'll run into cases where we need to iterate more on this.

rawlingsj:
> > the tag filter is just prefix check right now right? how about we parse them as semver properly?
> > of if you're worried that upstream may not do proper semver tagging, maybe we can do regex constraint? it probably looks more complicated though.
> 
> Good questions. I defer to @rawlingsj. My take is that the prefix approach is a reasonable start, and seems to be handling all the cases I've seen so far. But I'm sure we'll run into cases where we need to iterate more on this.

Yeah, the tag filter is passed as a string to github's graphql API as a query filter.  Any extra logic would mean complicating wolfi bot.  The filter is working for the usecases we've found so far and makes it easy for folks to maintain.  Probably best to leave it for now and see how things go.

PR #1188, opened 2023-04-11
Prepare for YAML formatting

luhring:
This PR:

1. Fixes a minor YAML comment issue in `py3-yaml.yaml` to fix how YAML is decoded by our tools.

2. But more importantly, adds the `.yam.yaml` configuration file that governs how `yam`—as well as our tools that are now using yam's encoder (`melange` and `wolfictl`)—will format our package YAML files. This also determines what _the "correct answer" is_ when we add a lint check into our CI process.

Of note is the "gap" section, which specifies yq-style expressions for nodes whose children elements should be separated by gaps (i.e. a blank line).

### DIFF
--- a/.yam.yaml
+++ b/.yam.yaml
@@ -1,0 +1,6 @@
+gap:
+  - "."
+  - ".subpackages"
+  - ".data"
+
+indent: 2
--- a/.yam.yaml
+++ b/.yam.yaml
@@ -2,5 +2,6 @@ gap:
   - "."
   - ".subpackages"
   - ".data"
+  - ".pipeline"
 
 indent: 2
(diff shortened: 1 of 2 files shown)


.yam.yaml
```diff
@@ -0,0 +1,6 @@
+gap:
+  - "."
+  - ".subpackages"
+  - ".data"
```
luhring, line 4: This is based on the most common formatting I saw among our config files. **Very open to feedback!** 😃 

🤔  Should we also add an item for `.pipeline`? Just eyeballing, it seemed roughly 50/50 on configs that had blank lines between pipeline items.

kaniini: I like the idea of using `.pipeline` :)

luhring: Added!

kaniini (approved):

PR #1201, opened 2023-04-12
add update config for remaining packages

rawlingsj:


cmake.yaml
```diff
@@ -45,6 +45,7 @@ pipeline:
       DESTDIR="${{targets.destdir}}" ninja install
   - uses: strip
 update:
-  enabled: false
+  enabled: true
+  manual: true # be careful upgrading Make as it is a core package
```
luhring, line 49: nit:
```suggestion
  manual: true # be careful upgrading cmake as it is a core package
```

luhring (dismissed):
So cool... And the explanations in the comments are super useful 💯

rawlingsj:
a different CI failure
```
error getting package dependencies: could not find package libbsd=${{package.version}}-${{package.epoch}} in indexes
```

rawlingsj:
Looks like this https://github.com/wolfi-dev/os/blob/bcf13d9e18be38be8d4e84d9864da00b4621aa26/libbsd.yaml#L33

maybe that's not needed

kaniini:
Yeah we can get rid of that libbsd line, Melange already solves those kinds of dep relationships automatically :)

rawlingsj:
green! just need a LGTM

kaniini (approved):

PR #1215, opened 2023-04-12
flux: add remaining components

joshrwolf:
Fixes:

Related: https://github.com/chainguard-dev/internal/issues/1177

#### For new package PRs only


flux.helm-controller.yaml
```diff
@@ -0,0 +1,29 @@
+package:
+  name: flux-helm-controller
```
dlorenc, line 2: Nit - I think the file names have to match the package name (so flux-helm-controller, instead of flux.helm-controller)

joshrwolf: ack, was wondering whether that would break things...

luhring (approved):

PR #1226, opened 2023-04-12
Don't run update cron for forks

luhring:
I noticed when I updated my own fork of Wolfi, I started getting failed job GitHub notifications every hour 😱  (e.g. [this](https://github.com/luhring/wolfi-os/actions/runs/4683324124/jobs/8298280373)). It seems like this shouldn't run on forks.


rawlingsj:
I tend to disable actions on my own forks and enable them when I want to test. Not sure if this is a common approach for others but I don't normally see GitHub actions tied to a repo like this?

kaniini (approved):

kaniini:
Most Github repos don't have a bunch of scheduled tasks running.  I think it makes sense to lock the scheduled tasks to the main repo.

PR #1239, opened 2023-04-13
Add systemd package, and add it as a dependency to fluent-bit.

dlorenc:
Fluent-bit has a systemd plugin that's required for the default helm chart. We comfigured it in our build, but didn't actually have systemd-dev so it ignored that part, making our fluent-bit image crash in a normal helm install.

I've verified that this fixes our image and the helm tests now pass!

Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/systemd.yaml
+++ b/systemd.yaml
@@ -1,0 +1,56 @@
+package:
+  name: systemd
+  version: 253
+  epoch: 0
+  description: The systemd System and Service Manager
+  copyright:
+    - license: LGPL-2.1-or-later AND GPL-2.0-or-later
+
+environment:
+  contents:
+    packages:
+      - busybox-full
+      - ca-certificates-bundle
+      - build-base
+      - gperf
+      - coreutils
+      - meson
+      - libcap-dev
+      - ninja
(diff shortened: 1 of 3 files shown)


dlorenc:
This will need https://github.com/wolfi-dev/os/pull/1228 as well.

kaniini:
Do we know which systemd components are actually needed by systemd?  There are alternative implementations that don't require buy-in to systemd itself for most of them.

dlorenc:
> Do we know which systemd components are actually needed by fluent-bit? There are alternative implementations that don't require buy-in to systemd itself for most of them.

systemd-dev is what I needed to make the plugin work - it has a journald plugin that's configured by default.

systemd.yaml
```diff
@@ -0,0 +1,56 @@
+package:
+  name: systemd
+  version: 253
+  epoch: 0
+  description:
```
luhring, line 5: Should we have a description?

dlorenc: Fixed! Systemd is indescribable!

luhring (approved):

PR #1278, opened 2023-04-13
apk-tools: upgrade to 2.14.0_rc1

kaniini:
#### For version bump PRs

### DIFF
--- a/apk-tools.yaml
+++ b/apk-tools.yaml
@@ -1,13 +1,14 @@
 package:
   name: apk-tools
-  version: 2.12.11
-  epoch: 2
+  version: 2.14.0_rc1
+  epoch: 0
   description: "apk-tools (Wolfi package manager)"
   copyright:
     - license: GPL-2.0-only
   dependencies:
     runtime:
       - ca-certificates-bundle
+
 environment:
   contents:
     repositories:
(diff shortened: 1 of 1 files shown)


apk-tools.yaml
```diff
@@ -50,12 +57,14 @@ subpackages:
...
+
   - name: "lua5.3-apk"
     description: "Lua module for libapk"
     pipeline:
       - runs: |
           mkdir -p "${{targets.subpkgdir}}"/usr
           mv "${{targets.destdir}}"/usr/lib "${{targets.subpkgdir}}"/usr/lib
+
 update:
   enabled: true
```
luhring, line 69: Depending on how long we want to use this version, we should (IIUC) consider setting this to `false` now to avoid auto updates

kaniini: 2.14.0 final is due out in a couple of weeks.  I don't think that we should disable auto-updates, right?

luhring: I just mean that since the `Wolfictl Check Updates` check failed, I think it will automatically open a PR to "bump" this to `2.12.11` as soon as this PR gets merged, which might be confusing to folks and even accidentally get merged

kaniini: Got it.  Now done :)

jdolitsky (approved):

PR #1400, opened 2023-04-17
enable wolfi check diff on CI builds so we can see what files change …

rawlingsj:
…for a package via a pull request comment


rawlingsj:
Comments will look similar to this:

Screenshot 2023-04-17 at 12 54 54

kaniini (approved):

PR #1406, opened 2023-04-17
add mysql-8

joshrwolf:
mysql-8: new package

Related: #272, https://github.com/chainguard-dev/internal/issues/951

#### For new package PRs only

### DIFF
--- a/mysql-8.yaml
+++ b/mysql-8.yaml
@@ -1,0 +1,76 @@
+package:
+  name: mysql
+  version: 8.0.32
+  epoch: 0
+  description: MySQL Server, the world's most popular open source database
+  copyright:
+    - license: GPL-2.0-only
+  dependencies:
+    runtime:
+
+environment:
+  contents:
+    packages:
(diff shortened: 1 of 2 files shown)


mysql-8.yaml
```diff
@@ -0,0 +1,76 @@
+package:
+  name: mysql
```
rawlingsj, line 2: I think this needs to be mysql-8 to match the filename

rawlingsj:
BTW I'll fix the panic in the check above, looks like we need to check if c is nil https://github.com/wolfi-dev/wolfictl/blob/e7b07485cca8b00720ce66cf164bcf0440a3a842/pkg/checks/update.go#L157-L161

But the reason this happened I think is because the file name is different from the package name.  I added an inline comment  aove

joshrwolf:
> But the reason this happened I think is because the file name is different from the package name. I added an inline comment above

thanks, you are quick! I was just about to ask

PR #1410, opened 2023-04-17
Numactl dev

joshrwolf:
numactl: add `-dev` package


joshrwolf:
does the `epoch` need to be bumped when adding a new `subpackage`?

numactl.yaml
```diff
@@ -40,6 +40,11 @@ subpackages:
     pipeline:
       - uses: split/manpages
 
+  - name: "numactl-dev"
+    description: "autoconf dev"
```
dlorenc, line 44: typo: numactl dev

dlorenc (commented):
Yea, you'll need to bump the epoch to add a new package.

jdolitsky (approved):

PR #1429, opened 2023-04-18
Revert "Makefile: make mkdir -p step for source dir an explicit step …

rawlingsj:
…so it can be overridden"

We're getting failed release builds, think it's due to this change. Let's roll back to get green again and reimplement this.

This reverts commit 6e8612ae46f378d66d5efea1e56838bd266afaf9.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,9 +38,8 @@ $(eval sourcedir = $(call comma-split,$(1),2))
 $(eval sourcedir = $(or $(sourcedir),$(pkgname)))
 $(eval pkgtarget = $(TARGETDIR)/$(shell $(MELANGE) package-version $(pkgname).yaml).apk)
 packages/$(pkgname): $(pkgtarget)
-./${sourcedir}/:
-	mkdir -p ./${sourcedir}/
-$(pkgtarget): ${KEY} ./${sourcedir}/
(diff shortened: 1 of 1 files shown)


dlorenc:
My theory for the problem is that the change here: 
image

is making it so that the directory `${sourcedir}` is getting created for every package, and because it's listed as a dependency for for `$(pkgtarget)`, it's busting the cache and triggering every package to get rebuilt (unless it already has a directory). This can be observed in the failing logs, where we attempt to build 61 packages before failing on cmake: https://pipelines.actions.githubusercontent.com/serviceHosts/969e5abc-ffab-48b3-bd54-b87797129033/_apis/pipelines/1/runs/15821/signedlogcontent/2?urlExpires=2023-04-18T11%3A00%3A03.3656729Z&urlSigningMethod=HMACV1&urlSignature=NPcXAbGBpnFvOfSLTI1lxAySHjfug2RnBsjU0%2F38BtM%3D

These builds are ordered by their order in packages.txt, but with packages that already have directories missing:

`mpfr` is first, when `gmp` is before it. Then `mpc`, `isl`, `zlib`, `flex`, but no `glibc`.

dlorenc (approved):

PR #1469, opened 2023-04-19
openjdk-17: update to 17.0.7.7

patflynn:


patflynn:
the update checks will always fail due to the jdk version schema.

rawlingsj (approved):

PR #1475, opened 2023-04-19
gradle-8: Fix broken build

patflynn:
Fix gradle build due to openjdk-11 in build deps and coming from transitive runtime deps

Related:
review thread on https://github.com/wolfi-dev/os/pull/1203

I suspect this might be an apko regression as I don't understand how this was working before.

One thing that I'm wondering about is why gradle pulls in JRE as runtime deps while maven doesn't. I don't know what the right thing but it's definitely inconsistent right now.

### DIFF
--- a/gradle-8.yaml
+++ b/gradle-8.yaml
@@ -1,7 +1,7 @@
 package:
   name: gradle-8
   version: 8.0.2
-  epoch: 0
+  epoch: 1
   description: A Java project management and project comprehension tool.
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


patflynn:
I tested this with a new gradle image and seems 👍

jdolitsky (approved):

jdolitsky (commented):
looks like new version of gradle too
```
2023/04/19 20:39:10 error during command execution: 1 error occurred:
	* package gradle-8: update found newer version 8.1.0 compared with package.version 8.0.2 in melange config
	```

patflynn:
@jdolitsky yeah this fixes the build so that update PR will pass.

PR #1477, opened 2023-04-19
Fix chartmuseum pulling in glibc

jdolitsky:


dlorenc:
What's the problem with pulling it in? IIUC melange should auto-detect this then add it to the image as a runtime dependency.

imjasonh:
> What's the problem with pulling it in? IIUC melange should auto-detect this then add it to the image as a runtime dependency.

How does melange auto-detect it? It sounded like `CGO_ENABLED=0` might be the hint that it needed, but I don't even pretend to understand how that works.

jdolitsky:
good question for @kaniini

kaniini:
What is the question, exactly?

kaniini (approved):

PR #1480, opened 2023-04-20
fix maven dependency to be stage-0 rather than itself

rawlingsj:


maven.yaml
```diff
@@ -12,7 +12,7 @@ environment:
       - ca-certificates-bundle
       - build-base
       - busybox
-      - maven
+      - maven-stage0
```
imjasonh, line 15: I think we want stage0 to say it provides maven, like the other stage0s do.

This will mean maven always builds with the stage0 and we never move off of it.

imjasonh (changes requested):

deitch (approved):
I can confirm that this solves the dependency issue.

Whether we want to do it this way, or via `maven-stage0.yaml` having a `provides:` as @imjasonh recommended, I am fine either way.

rawlingsj:
Going to close this and re-create with the provides approach

maven.yaml
imjasonh: It stops depending on stage0 when there's another `maven` built from source. The progression goes:

1. the repo is empty; `melange build maven.yaml` finds the package `provides`ed by `maven-stage0` as the only `maven` package, and uses `stage0` to build `maven`.
2. the repo contains `maven`; `melange build maven.yaml` finds the `maven` built in (1) and uses that to build `maven`.

If `maven.yaml` says "I depend on `maven-stage0` specifically" then even after `maven` is built, it will always use the `stage0`-built `maven` to build subsequent packages.

In the ideal steady state, after (2), no packages depend on any `stage0`. We could even consider making this a validation check done by `wolfictl`, perhaps.

deitch: That's what I thought you meant, and that's the issue I had. How do you make this reproducible? The declarative file `maven.yaml` has a dependency declared (`maven`) whose source for fulfillment is unpredictable. It will change based on what I have built already. It also makes reproducing from scratch impossible (besides the fact that building from scratch will lead to distinct results from building while using the cache).

You are describing something of a laddered approach. First build A, then B on that, then another B based on that, then another B based on that, etc.

It makes a lot of sense (heck, don't we compile things based on compilers that were built based on etc.?), but to do so, we need a way to reproduce it reliably. At no point in time, does the state of config files represent something that will allow someone to build the build graph in a way that will work each time.

If we could get to that, I think it would be pretty great.

PR #1483, opened 2023-04-20
make stage0 usage more consistent

imjasonh:
⚠️ This PR only bumps epochs for non`-stage0` packages. Please let me know if you disagree that this is what we want to do. ⚠️ 

This aligns usage of stage0 packages across the repo (hopefully correctly 🤞)

When `{package}` depends on itself, there is a `{package}-stage0.yaml` that builds that package from pre-built binaries and `provides: package-${{package.version}}-${{package.epoch}}` at an early version.

Then, when `{package}.yaml` has a dependency on itself, the first time it's built it will use the package `provides`ed by `{package}-stage0.yaml`, and subsequent versions will use the package `provides`ed by `{package}.yaml` _and not the `-stage0`_.

After `{package}` is successfully built, the intention is that the `stage0` packages are never depended on again, except when bootstrapping from scratch (e.g., a new architecture, or someone reproducing and checking our work).

The following packages are `stage0`'ed in this way:

- `go`
- ~`rust`~ (https://github.com/wolfi-dev/os/pull/1485)
- `maven`
- `gradle`
- `sbt`

`openjdk-{11,17}` are not built this way, but should be. This PR adds `TODO`s.

@joshrwolf @rawlingsj @patflynn @deitch

### DIFF
--- a/go-fips-1.20.yaml
+++ b/go-fips-1.20.yaml
@@ -7,12 +7,15 @@ package:
     - license: BSD-3-Clause
   dependencies:
     provides:
-      - go=1.20.999
+      - go-fips=1.20.999 # This should be go-fips=${{package.version}}-${{package.epoch}}
     runtime:
       - build-base
       - bash
       - binutils-gold # Needed for cgo linking due to upstream issue #15696 which forces use of the gold linker.
       - openssl-dev # Needed for building against cryptographic packages.
+      # The first time this is built, it depends on go-stage0.yaml being built first,
+      # using pre-built go binaries, which `provides` a go package at an earlier version.
+      # Afterwards it depends on previous versions of this package built from source.
       - '!go-1.20'
 
 secfixes:
--- a/sbt-stage0.yaml
+++ b/sbt-stage0.yaml
@@ -15,7 +15,7 @@ package:
       - openjdk-11-jre
       - bash
     provides:
-      - sbt=1.8.2-r0
+      - sbt=1.8.2
 
 environment:
   contents:
--- a/sbt-stage0.yaml
+++ b/sbt-stage0.yaml
@@ -14,6 +14,8 @@ package:
     runtime:
       - openjdk-11-jre
       - bash
+    provides:
+      - sbt=1.8.2-r0
 
 environment:
   contents:
--- a/busybox-full.yaml
+++ b/busybox-full.yaml
@@ -1,13 +1,10 @@
 package:
   name: busybox-full
   version: 1.36.0
-  epoch: 0
+  epoch: 1
   description: "swiss-army knife for embedded systems, with more applets"
   copyright:
     - license: GPL-2.0-only
-  dependencies:
-    provides:
-      - busybox=${{package.version}}-${{package.epoch}}
 
 secfixes:
   1.35.0-r3:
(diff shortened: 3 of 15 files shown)


go-fips-1.20.yaml
```diff
@@ -7,12 +7,15 @@ package:
     - license: BSD-3-Clause
   dependencies:
     provides:
-      - go=1.20.999
+      - go=1.20.999  # This should be go=${{package.version}}-${{package.epoch}}
```
imjasonh, line 10: This should be `go-fips=1.20.999` right?

kaniini:
I don't think `${{package.version}}` and friends are substituted in `provides` entries yet.

go-fips-1.20.yaml
imjasonh: Did this.

deitch:
> I don't think ${{package.version}} and friends are substituted in provides entries yet.

They weren't, I was wondering about that when parsing `busybox-full.yaml`, so I pulled it in as a lib from melange for when building the graph.

deitch:
> Then, when {package}.yaml has a dependency on itself, the first time it's built it will use the package providesed by {package}-stage0.yaml, and subsequent versions will use the package providesed by {package}.yaml and not the -stage0.
>
> After {package} is successfully built, the intention is that the stage0 packages are never depended on again, except when bootstrapping from scratch (e.g., a new architecture, or someone reproducing and checking our work).

We had a bit of a discussion elsewhere on this, so it is good to centralize it.

As I understand this PR, the mechanism by which we indicate that something should depend on stage0 vs some prior build of the same package, is`provides`.

Let's say `a-1.3` depends on `a`; `a-stage0.yaml` (v1.0) provides `a=1.0`, so `a-1.3` can build depending on it. Now I have `a-1.3` in the repo. I modify `a.yaml` to get `a-1.4`. Now when I build, it still looks for `a`, but because it was already built, it find `a-1.3`, which is higher than stage0 (`a-1.0`), so it uses that?

If this is the case, then a few comments/questions:

* what happens when we rebuild from scratch? We now get `a-1.4` built based on `a-1.0`. That is not the same as `a-1.4` based on `a-1.3` which was based on `a-1.0`. Maybe we don't care, maybe we do, but that is not identical.
* If we ever actually have a need to reproduce from scratch, how do we do so?

kaniini:
I think we should not merge this, and wait for melange to support variable substitutions in `provide`.

sbt-stage0.yaml
```diff
@@ -14,6 +14,8 @@ package:
     runtime:
       - openjdk-11-jre
       - bash
+    provides:
+      - sbt=1.8.2-r0
```
kaniini, line 18:
```suggestion
      - sbt=1.8.2
```

kaniini (changes requested):

kaniini (approved):

deitch:
@imjasonh that does clarify. We are saying that what we need to be able to do is:

* rebuild any current yaml based on current state of the apk repository
* rebuild any current yaml from scratch, i.e. nothing in the apk repository

We accept that those two might not give identical results. Both will produce a package `a` whose version is 1.4, but they might not be binary identical. The first might be based on 1.3 based on 1.2 based on 1.0, some of which interim stages might not exist, while the latter will be based directly on 1.0, which is either here in the repo as `a-stage0` or upstream in bootstrap.

As long as we both accept that those two are our build requirements, and accept that the rebuilds could be different at a binary level, then we are fine here.

I think we should be able to rebuild at any point in direct reproduction of what was there, but that may very well be come with too much cost.

PR #1485, opened 2023-04-20
rust: clean up stage0, upgrade to 1.69.0

kaniini:


rust-stage0.yaml
```diff
@@ -1,14 +1,14 @@
 package:
   name: rust-stage0
-  version: 1.65.0
-  epoch: 2
+  version: 1.68.2
```
imjasonh, line 3: I'm surprised we're updating the stage0 package. Is that because 1.68 is needed to be able to build 1.69? If so, can we document that in rust.yaml somewhere?

dlorenc (dismissed):

rust-stage0.yaml
```diff
@@ -1,14 +1,14 @@
...
+  epoch: 0
   description: "bootstrap package for rust"
   copyright:
-    - license: blessing
+    - license: Apache-2.0 AND MIT
   dependencies:
     runtime:
     provides:
-      - rust=1.65.0
+      - rust=1.68.2
```
imjasonh, line 11: Since `${{package.version}}` isn't supported here yet, can you add a comment up by the `version:` to update this `provides` when we do this in the future? I can see this being easy to miss.

rust-stage0.yaml
kaniini: Yes, 1.68 is needed to build 1.69.

rust-stage0.yaml
kaniini: Done.

rawlingsj (approved):

PR #1506, opened 2023-04-21
Add a workflow to close stale issues and PRs

imjasonh:
This will close PRs that haven't had activity after 120 days, warning about this closure after 90 days.

We can tweak these timelines or wording if we want to.

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -1,0 +1,30 @@
+name: 'Close stale'
+
+on:
+  schedule:
+  - cron: '0 1 * * *'
+
+jobs:
+  stale:
+    runs-on: 'ubuntu-latest'
+    steps:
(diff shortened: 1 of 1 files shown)


kaniini:
Not in favor of this.  https://drewdevault.com/2021/10/26/stalebot.html

luhring:
> Not in favor of this. https://drewdevault.com/2021/10/26/stalebot.html

This article seems to just be about _issues_, not PRs. I'm not sure I entirely agree with the author there, but, I think there's more value to us in closing stale _PRs_ than issues, since we've seen so many problems with stale PRs.

Would it make sense to close only stale _PRs_ for now?

kaniini:
Yes, we can close stale PRs.  Just not in favor of closing issues.

luhring (approved):

kaniini (approved):

PR #1508, opened 2023-04-21
Update llvm-lld.yaml

imjasonh:
Configure llvm-ld to upgrade to latest major version 15 only, and not try to upgrade to 16, which requires manual work.

There isn't a newer 15 at this time, but this pattern will help us when we do manually upgrade it to 16.

### DIFF
--- a/llvm-lld.yaml
+++ b/llvm-lld.yaml
@@ -66,5 +66,7 @@ subpackages:
 
 update:
   enabled: true
-  release-monitor:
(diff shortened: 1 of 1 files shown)


imjasonh:
```
	* failed to create a version slice for llvm/llvm-project: failed to create a version from llvmorg-15.0.0: Malformed version: llvmorg-15.0.0
```

rawlingsj (approved):

PR #1545, opened 2023-04-23
add external secrets operator

ajayk:
Add External secrets operator 

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/external-secrets-operator.yaml
+++ b/external-secrets-operator.yaml
@@ -1,0 +1,36 @@
+package:
+  name: external-secrets-operator
+  version: 0.8.1
+  epoch: 0
+  description: Integrate external secret management systems with Kubernetes
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - ca-certificates-bundle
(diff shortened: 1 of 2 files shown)


external-secrets-operator.yaml
```diff
@@ -0,0 +1,38 @@
...
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/external-secrets/external-secrets
+      tag: v${{package.version}}
+      expected-commit: a0d2aef2e35c259c9ee75d65f7587e6ed71ef2ad
+
+  - runs: |
+      # Our global LDFLAGS conflict with a Makefile parameter
+      unset LDFLAGS
```
rawlingsj, line 28: we needed to add this for a previous package but it's not always needed.  Just wanted to check that unsetting the global LDFLAGS was needed for external secrets?

ajayk: I don't think we need to reset the LDFLAGS here ,

rawlingsj (approved):

PR #1563, opened 2023-04-24
busybox: add symlink manifest

kaniini:
BusyBox symlinks are not tracked by the apkdb, so we need to track them somehow for SBOMs and other tasks like simulating `busybox --install`.

### DIFF
--- a/busybox.yaml
+++ b/busybox.yaml
@@ -1,7 +1,7 @@
 package:
   name: busybox
   version: 1.36.0
-  epoch: 2
+  epoch: 3
   description: "swiss-army knife for embedded systems"
   copyright:
     - license: GPL-2.0-only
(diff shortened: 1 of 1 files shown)


busybox.yaml
```diff
@@ -67,6 +67,9 @@ pipeline:
       install -m755 busybox "${{targets.destdir}}"/bin/busybox
       install -m644 securetty "${{targets.destdir}}"/etc/securetty
 
+      mkdir -p "${{targets.destdir}}"/etc/busybox-paths.d
+      ./busybox --list-path > "${{targets.destdir}}"/etc/busybox-paths.d/busybox
```
jdolitsky, line 71:
```suggestion
      "${{targets.destdir}}"/bin/busybox --list-path > "${{targets.destdir}}"/etc/busybox-paths.d/busybox
```
?

jdolitsky (approved):

busybox.yaml
kaniini: I don't really think it matters which we do.  We have `busybox` in the working dir which gets installed.

deitch:
Do we need this for `busybox-full.yaml` as well?

PR #1596, opened 2023-04-26
disable auto updates for py3-urllib3 as upstream is releasing pre-rel…

rawlingsj:
…eases using a strange version suffix `2.0.0a4`

I'll add a `strip-suffix:` config to melange and update the bot to work with it


rawlingsj:
`Wolfictl Check Updates` is expected to fail until the bot handles the proposed new `strip-siffix` key

patflynn (approved):

PR #1619, opened 2023-04-27
Rename libjpeg to libjpeg-turbo

luhring:


libjpeg-turbo.yaml
```diff
@@ -1,13 +1,13 @@
...
+  name: libjpeg-turbo
   version: 2.1.91
   epoch: 1
   description: "Accelerated baseline JPEG compression and decompression library"
   copyright:
     - license: BSD-3-Clause AND IJG AND Zlib
   dependencies:
     provides:
-      - libjpeg-turbo=2.1.91
+      - libjpeg
```
kaniini, line 10: If you remove the `libjpeg` provide in the `-dev` package, then you can delete this provide entry too.  Melange will deduce that the two packages are related automatically using SCA.

libjpeg-turbo.yaml
```diff
@@ -1,13 +1,13 @@
 package:
-  name: libjpeg
+  name: libjpeg-turbo
   version: 2.1.91
   epoch: 1
```
kaniini, line 4: The epoch should be bumped just to be sure.

kaniini (changes requested):

libjpeg-turbo.yaml
luhring: Done!

libjpeg-turbo.yaml
luhring: Done! (I think!)

kaniini (approved):

PR #1620, opened 2023-04-27
libxcursor: new package

kaniini:
Dependency of GTK, which is a dependency of OpenJDK 7, because it does not have headless mode as a build option yet.

#### For new package PRs only

### DIFF
--- a/libxcursor.yaml
+++ b/libxcursor.yaml
@@ -1,0 +1,51 @@
+package:
+  name: libxcursor
+  version: 1.2.1
+  epoch: 0
+  description: X cursor management library
+  copyright:
+    - license: MIT
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+      - automake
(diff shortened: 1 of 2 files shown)


libxcursor.yaml
```diff
@@ -0,0 +1,51 @@
...
+      expected-sha256: 46c143731610bafd2070159a844571b287ac26192537d047a39df06155492104
+      uri: https://www.x.org/releases/individual/lib/libXcursor-${{package.version}}.tar.xz
+
+  - uses: autoconf/configure
+
+  - uses: autoconf/make
+
+  - uses: autoconf/make-install
+
+  - uses: strip
```
imjasonh, line 35: (not blocking) Is the spacing intentional? Is `yam` doing that? 🤔

imjasonh (approved):

libxcursor.yaml
kaniini: Yes

PR #1627, opened 2023-04-27
bump cups following libpaper major version update and soname update

rawlingsj:


kaniini:
No changes to package, why is this rebuild needed?

jdolitsky (approved):

PR #1631, opened 2023-04-27
mpc/1.3.1 also switch to automatic updates

rawlingsj:
@kaniini mcp was originally skipping auto updates, I wonder if this was because of possible soname files changed?  If so I've switched it to enable automated updates now but can keep manual if you prefer?

### DIFF
--- a/mpc.yaml
+++ b/mpc.yaml
@@ -1,7 +1,7 @@
 package:
   name: mpc
-  version: 1.2.1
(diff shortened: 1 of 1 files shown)


rawlingsj:
fixes https://github.com/wolfi-dev/os/issues/1254

kaniini (approved):

PR #1634, opened 2023-04-27
icu/73.1 manually update as melange bump doesnt work with var-transfo…

### LINKED ISSUE #1259: icu/73-1 new package update

rawlingsj:
…rms yet

fixes #1259

### DIFF
--- a/dotnet-7.yaml
+++ b/dotnet-7.yaml
@@ -1,7 +1,7 @@
 package:
   name: dotnet-7
   version: 7.0.105
-  epoch: 0
+  epoch: 1
   description: ".NET SDK"
   target-architecture:
     - all
--- a/harfbuzz.yaml
+++ b/harfbuzz.yaml
@@ -1,7 +1,7 @@
 package:
   name: harfbuzz
   version: 7.2.0
-  epoch: 0
+  epoch: 1
   description: Text shaping library
   copyright:
     - license: MIT
--- a/icu.yaml
+++ b/icu.yaml
@@ -1,7 +1,7 @@
 package:
   name: icu
-  version: 71.1
-  epoch: 3
+  version: 73.1
+  epoch: 0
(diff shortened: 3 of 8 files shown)


rawlingsj:
CI soname check failure:
```
soname files differ, this can cause an ABI break.  

Existing soname files 
libicudata.so.71.1,libicui18n.so.71.1,libicuio.so.71.1,libicutest.so.71.1,libicutu.so.71.1,libicuuc.so.71.1
New soname files 
libicudata.so.73.1,libicui18n.so.73.1,libicuio.so.73.1,libicutest.so.73.1,libicutu.so.73.1,libicuuc.so.73.1

soname version check failed, libicudata has an existing version 71.1.0 while new package contains a different version 73.1.0.  This can cause ABI failures
```

kaniini:
We will need to rebuild all the ICU dependents then when we merge this.  Thanks to the way Wolfi works, the old ICU won't get deleted, but still :)

rawlingsj:
I've bumped the epoch for dependent packages in a following commit in this PR to be sure.

rawlingsj:
`Wolfictl Check Updates` is expected to fail as melange bump doesn't work with var-transforms yet

rawlingsj:
CI failure:
```
EXEC : error : RPC failed; curl 18 Recv failure: Connection reset by peer [/home/build/installer/src/SourceBuild/tarball/BuildSourceBuildTarball.proj]
```
going to retry

rawlingsj:
wolfictl update check expected to fail as melange bump doesn't work with var transforms yet

rawlingsj:
CI failure was related to the new soname version
```
soname version check failed, libicudata has an existing version 71.1.0 while new package contains a different version 73.1.0.
```
But as I've bumped all the dependent packages epoch to ensure they work I think this is ok?  

@kaniini WDYT?

PR #1663, opened 2023-04-27
gradle-8: update to 8.1.1

patflynn:
This temporarily swaps build to use gradlew instead of the previous version of Gradle.  Building 8.1.1 with 8.0.1 results in:

```
  x86_64    | * What went wrong:
  x86_64    | Script compilation errors:
  x86_64    |
  x86_64    |   Line 27:         languageVersion = JavaLanguageVersion.of(11)
  x86_64    |                    ^ Val cannot be reassigned
  x86_64    |
  x86_64    |   Line 27:         languageVersion = JavaLanguageVersion.of(11)
  x86_64    |                                      ^ Type mismatch: inferred type is JavaLanguageVersion! but Property<JavaLanguageVersion!>! was expected
  x86_64    |
  x86_64    |   Line 28:         vendor = JvmVendorSpec.ADOPTIUM
  x86_64    |                    ^ Val cannot be reassigned
  x86_64    |
  x86_64    |   Line 28:         vendor = JvmVendorSpec.ADOPTIUM
  x86_64    |                             ^ Type mismatch: inferred type is JvmVendorSpec! but Property<JvmVendorSpec!>! was expected
  x86_64    |
  x86_64    | 4 errors
  x86_64    |
```

### DIFF
--- a/gradle-8.yaml
+++ b/gradle-8.yaml
@@ -1,15 +1,13 @@
 package:
   name: gradle-8
-  version: 8.0.2 # When bumping this, also bump the provides below!
-  epoch: 2
+  version: 8.1.1 # When bumping this, also bump the provides below!
+  epoch: 0
   description: A Java project management and project comprehension tool.
   copyright:
     - license: Apache-2.0
   dependencies:
-    runtime:
-      - openjdk-11-jre
     provides:
-      - gradle=8.0.2
+      - gradle=8.1.1
 
 environment:
   contents:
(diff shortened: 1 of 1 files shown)


rawlingsj:
I can fix up the failed check above later today after a change that I'm working on lands in wolfictl.

rawlingsj (approved):

gradle-8.yaml
```diff
@@ -1,15 +1,13 @@
 package:
   name: gradle-8
-  version: 8.0.2 # When bumping this, also bump the provides below!
-  epoch: 2
+  version: 8.1.1 # When bumping this, also bump the provides below!
```
luhring, line 3: I know the comment wasn't added in this PR, but I'm not sure what to make of it. Does the update automation handle updating the provides? If so I'd say we can remove the comment. If the automation doesn't handle this, and it's enabled right now, this seems like a footgun. WDYT?

imjasonh: I added the comment, in an attempt to at least warn about the footgun.

Ideally we could use vars in provides and it'd stay up to date itself. I haven't made that change in melange yet though.

patflynn: The updater doesn't as AFAIK. @kaniini @imjasonh to check that I'm doing it right.

luhring: I think the comment is good @imjasonh 👍  — perhaps "footgun" is too strong, I just meant that I could see someone (like myself) overlooking this and merging a messed up state.

Issue #1588, opened 2023-04-26
the openjdk packages should include their versions in the install path


patflynn:
It should be possible to have multiple versions of the jdk or jre installed in the same image. Specifically this is important for builder images that may use a different jvm to run the build tool from the jdk used to do the actual build.

This should mean changing our install paths from 
`/usr/lib/jvm/openjdk/` to something like `/usr/lib/jvm/openjdk-11`

PR #1671, opened 2023-04-28
openjdk-17: Install jdk17 in version specific folder

patflynn:
WIP towards #1588. I'm thinking the safest thing to do is roll this out to 17 first and wait a few days to be sure there's no user impact before doing the same for 11.

### DIFF
--- a/openjdk-17.yaml
+++ b/openjdk-17.yaml
@@ -1,7 +1,7 @@
 package:
   name: openjdk-17
   version: 17.0.7.7
-  epoch: 0
+  epoch: 1
   description:
   copyright:
     - license: GPL-2.0-only
@@ -9,6 +9,8 @@ package:
     runtime:
       - freetype
       - fontconfig
+  options:
+    no-provides: true
 
 environment:
   contents:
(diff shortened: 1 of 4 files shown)


openjdk-17.yaml
```diff
@@ -89,49 +91,53 @@ pipeline:
...
   - runs: |
-      mkdir -p ${{targets.destdir}}/usr/lib/jvm/openjdk
-      cp -r ./build/linux-*-server-release/images/jdk/* ${{targets.destdir}}/usr/lib/jvm/openjdk
+      mkdir -p ${{targets.destdir}}/usr/lib/jvm/java-17-openjdk
+      cp -r ./build/linux-*-server-release/images/jdk/* ${{targets.destdir}}/usr/lib/jvm/java-17-openjdk
 
       mkdir -p ${{targets.destdir}}/usr/bin
-      ln -sf /usr/lib/jvm/openjdk/bin/java ${{targets.destdir}}/usr/bin/java
-      ln -sf /usr/lib/jvm/openjdk/bin/javac ${{targets.destdir}}/usr/bin/javac
+      ln -sf /usr/lib/jvm/java-17-openjdk/bin/java ${{targets.destdir}}/usr/bin/java
```
kaniini, line 98: It would be better to integrate this into the `java-common` work I did.  If you create a `-default-jvm` package, with a symlink `/usr/lib/jvm/default-jvm` -> `/usr/lib/jvm/java-17-openjdk`, then Java 17 will automatically take ownership of all of the Java-related symlinks when the `-default-jvm` package is installed.  See `java-gcj-compat` and `openjdk-7` for examples of how the `-default-jvm` stuff works.

PR #1681, opened 2023-04-28
openjdk-7: new package

kaniini:
OpenJDK 7 provides the next chain in our source-bootstrapped Java story.

#### For new package PRs only

  - Exception: bootstrap dependency for OpenJDK 8

### DIFF
--- a/openjdk-7.yaml
+++ b/openjdk-7.yaml
@@ -1,0 +1,281 @@
+package:
+  name: openjdk-7
+  version: 7.321.2.6.28
+  epoch: 1
+  description: "IcedTea distribution of OpenJDK 7 (UNSUPPORTED)"
+  copyright:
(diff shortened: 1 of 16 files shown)


joshrwolf (approved):

openjdk-7.yaml
```diff
@@ -0,0 +1,281 @@
+package:
+  name: openjdk-7
```
patflynn, line 2: should we put `unsupported` or `bootstrap` in the name here to avoid folks using it without realizing it's not getting security updates?

kaniini: it's in the description.  i'd rather not rename it, because in the near future it will be moved to a special repo for bootstrap components.  that repo just doesn't presently exist.

PR #1682, opened 2023-04-28
Add advisories for gcc-6 vulnerabilities

luhring:
Creates advisories for `gcc-6` vulnerabilities just found by `wolfictl advisory discover`.

cc: @kaniini 

#### For security-related PRs

- [x] The security fix is recorded in `advisories` and `secfixes`

### DIFF
--- a/gcc-6.yaml
+++ b/gcc-6.yaml
@@ -242,3 +242,30 @@ subpackages:
 update:
   # EOL upstream, needed for Java 8 bootstrap only.
   enabled: false
+
+advisories:
+  CVE-2018-12886:
+    - timestamp: 2023-04-28T14:46:17.894915-04:00
+      status: under_investigation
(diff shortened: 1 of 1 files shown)


gcc-6.yaml
```diff
@@ -242,3 +242,30 @@ subpackages:
...
   # EOL upstream, needed for Java 8 bootstrap only.
   enabled: false
+
+advisories:
+  CVE-2018-12886:
+    - timestamp: 2023-04-28T14:46:17.894915-04:00
+      status: under_investigation
+    - timestamp: 2023-04-28T15:03:15.115547-04:00
+      status: affected
+      action: We are affected.
```
luhring, line 252: @kaniini Anything we can do about this at the moment?

jdolitsky (approved):

gcc-6.yaml
luhring: We can follow up post-merge.

PR #1691, opened 2023-04-29
fix some packages' provides to match their version

imjasonh:
For example, if a user asked for `- nodejs~18.15` they'd have gotten node 18.16, because it provided 18.15.999.

I found a couple other examples using this and some eyeballing:

```
yq '. | .package.name + "-" + .package.version + " : " + .package.dependencies.provides[]' *.yaml
```

I've added comments to hopefully remind us to bump the version when node 18.17 or whatever comes out and we get an auto-upgrade PR. These should really just be done with variable replacements, I should get around to doing that in melange.

### DIFF
--- a/nodejs-18.yaml
+++ b/nodejs-18.yaml
@@ -1,13 +1,13 @@
 package:
   name: nodejs-18
-  version: 18.16.0
-  epoch: 0
+  version: 18.16.0 # When bumping this, also bump the provides below!
+  epoch: 1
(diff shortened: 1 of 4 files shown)


dlorenc (approved):

tuananh:
@imjasonh is there any case we want to provide a lower version like that?

imjasonh:
> @imjasonh is there any case we want to provide a lower version like that?

Not that I can imagine, but I might not be thinking creatively enough. :)

PR #1705, opened 2023-04-30
Add CVE-2023-0264 to keycloak fixes.

dlorenc:
This was fixed in 21.0.1, we started packaging in 21.0.2.

Fixes:

Related:

#### For security-related PRs

### DIFF
--- a/keycloak.yaml
+++ b/keycloak.yaml
@@ -76,8 +76,18 @@ advisories:
     - timestamp: 2023-04-09T09:54:02.782617-04:00
       status: not_affected
       justification: vulnerable_code_not_in_execute_path
+  CVE-2023-0264:
+    - timestamp: 2023-04-30T07:15:20.601606-04:00
+      status: not_affected
+      justification: component_not_present
+      impact: The CVE was fixed in a version of keycloak prior to our packaging
+      fixed-version: "0"
+    - timestamp: 2023-04-30T07:15:51.679089-04:00
+      status: not_affected
+      justification: vulnerable_code_not_in_execute_path
 
 secfixes:
   "0":
     - CVE-2017-12159
     - CVE-2017-12158
+    - CVE-2023-0264
--- a/keycloak.yaml
+++ b/keycloak.yaml
@@ -61,33 +61,21 @@ advisories:
   CVE-2017-12158:
     - timestamp: 2023-04-09T09:54:02.778547-04:00
       status: not_affected
-      justification: component_not_present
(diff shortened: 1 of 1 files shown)


keycloak.yaml
```diff
@@ -76,8 +76,18 @@ advisories:
     - timestamp: 2023-04-09T09:54:02.782617-04:00
       status: not_affected
       justification: vulnerable_code_not_in_execute_path
+  CVE-2023-0264:
+    - timestamp: 2023-04-30T07:15:20.601606-04:00
+      status: not_affected
+      justification: component_not_present
+      impact: The CVE was fixed in a version of keycloak prior to our packaging
```
luhring, line 83: Just curious, was a scanner reporting this CVE? (Normally we don't need to comment on CVEs that were remediated before we started packaging, but curious what's going on here)

keycloak.yaml
```diff
@@ -76,8 +76,18 @@ advisories:
...
+      fixed-version: "0"
+    - timestamp: 2023-04-30T07:15:51.679089-04:00
+      status: not_affected
+      justification: vulnerable_code_not_in_execute_path
```
luhring, line 87: Is having both justification's here intentional?

dlorenc: It shows up during sync secfixes, I was confused too but it happens above for the others as well.

keycloak.yaml
dlorenc: No, but it started getting press yesterday so I researched it to make sure we weren't affected: https://www.offensity.com/en/blog/user-impersonation-via-stolen-uuid-code-in-keycloak-cve-2023-0264/

keycloak.yaml
luhring: Okay, got it. That could be a bug then... 🤔  I'll take a look soon.

My suggestion here would be to have this for all three advisories:

```yaml
status: not_affected
justification: vulnerable_code_not_present
impact: The CVE was fixed in a version of keycloak prior to our packaging
```

When I delete the `secfixes:` section, and update the advisories to just this ^, sync-secfixes gives me something that looks right. **I'll push a commit** that's just a suggestion, but **feel free to drop** the commit if you disagree.

keycloak.yaml
luhring: Got it, cool 👍

luhring (approved):

keycloak.yaml
dlorenc: Ah nice!

PR #1708, opened 2023-04-30
Actually fix perl regexps in grep.

dlorenc:
The flag is harmless but the build just silently continues even if it doesn't have the regexp library. Swap it to install the library and remove the flag.

I tested this time in a full container and it actually works now.

Fixes:

Related:

### DIFF
--- a/grep.yaml
+++ b/grep.yaml
@@ -1,7 +1,7 @@
 package:
   name: grep
   version: "3.10"
-  epoch: 1
+  epoch: 2
   description: "GNU grep implementation"
   copyright:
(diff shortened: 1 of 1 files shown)


dlorenc:
Now:
```
/ # grep -P
Usage: grep [OPTION]... PATTERNS [FILE]...
Try 'grep --help' for more information.
```

Before:

```
/# grep -P
grep: Perl matching not supported in a --disable-perl-regexp build
```

luhring (approved):

luhring:
For reference: #1706

PR #1731, opened 2023-05-02
libxml2/2.11.1 package update - also add var transforms to automatica…

rawlingsj:
…lly use major.minor in fetch URL

fixes #1677


rawlingsj:
Wolfictl Check Updates check expected to fail as melange bump doesn't work with var-transforms yet

luhring (approved):
This is cool!

PR #1732, opened 2023-05-02
fluent-bit/2.1.2 package update - switch to manual bump as melange bu…

rawlingsj:
…mp doesnt work with var-transforms yet

#### For version bump PRs

### DIFF
--- a/fluent-bit.yaml
+++ b/fluent-bit.yaml
@@ -1,6 +1,6 @@
 package:
   name: fluent-bit
-  version: 2.1.1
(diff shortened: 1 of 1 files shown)


rawlingsj:
`Wolfictl Check Updates` check expected to fail as melange bump doesn't work with `var-transforms` yet

marcofranssen:
Looking forward to this release to test and swap out the image from the fluent-bit helm chart.

joshrwolf (approved):

PR #1734, opened 2023-05-02
Add clickhouse package.

dlorenc:
Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/clickhouse.yaml
+++ b/clickhouse.yaml
@@ -1,0 +1,78 @@
+package:
+  name: clickhouse
+  version: "23.4.1.1943"
+  epoch: 0
+  description: ClickHouse is the fastest and most resource efficient open-source database for real-time apps and analytics.
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - git
+      - ninja
+      - build-base
+      - cmake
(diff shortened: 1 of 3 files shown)


rawlingsj:
I'm sure it's fine but just double checking - I noticed the build took a while and then saw clickhouse uses git submodules to pull in a bunch of other repos, some we have already in wolfi.  Example https://github.com/ClickHouse/ClickHouse/blob/517fcb1d02696817efe3398ae72232f1ca53d91b/.gitmodules#L58

Not 100% sure but wondered if the build is using those rather than wolfi and if that's what we expect / want?

dlorenc:
> I'm sure it's fine but just double checking - I noticed the build took a while and then saw clickhouse uses git submodules to pull in a bunch of other repos, some we have already in wolfi. Example https://github.com/ClickHouse/ClickHouse/blob/517fcb1d02696817efe3398ae72232f1ca53d91b/.gitmodules#L58
> 
> Not 100% sure but wondered if the build is using those rather than wolfi and if that's what we expect / want?

Yea, they require strict vendoring. This is similar to anything that uses bazel, or any java or Go apps.

kaniini:
I wonder if we should record these vendored dependencies somehow so we can keep track of them.

kaniini (approved):

PR #1769, opened 2023-05-03
Revert "logstash/8.7.1 package update"

rawlingsj:
This is failing on arm
```
ℹ️  aarch64   | Building logstash-core using gradle
⚠️  aarch64   | ./gradlew assemble
ℹ️  aarch64   |
ℹ️  aarch64   | > Task :installDevelopmentGems FAILED
⚠️  aarch64   |
⚠️  aarch64   | FAILURE: Build failed with an exception.
```

### DIFF
--- a/logstash.yaml
+++ b/logstash.yaml
@@ -58,6 +58,7 @@ pipeline:
 
 update:
   enabled: true
+  manual: true # avoid creating PRs while build fails on arm
   github:
     identifier: elastic/logstash
     strip-prefix: v
--- a/logstash.yaml
+++ b/logstash.yaml
@@ -1,6 +1,6 @@
 package:
   name: logstash
-  version: "8.7.0"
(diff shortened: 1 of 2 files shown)


luhring (dismissed):

logstash.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: logstash
-  version: 8.7.1
```
imjasonh, line 3: Can you add a comment that 8.7.1-r0 has been withdrawn, so when bumping to 8.7.1 again (if we do that) we should make it r1?

We don't want to republish a withdrawn package, or the next time the withdrawal workflow runs it'll re-withdraw it.

rawlingsj: yeah good call

imjasonh (approved):

PR #1777, opened 2023-05-03
Bump the JDK toolchain in logstash without bumping the version.

dlorenc:
The 8.7.1 update failed to build on arm. I wasn't able to reproduce it.

This is somewhat of a test to try changing the toolchain and see if it works, then we can try again to bump the version.

Fixes:

Related: https://github.com/wolfi-dev/os/pull/1769


jdolitsky (commented):
is it safe to merge even though the lint fails? @rawlingsj

rawlingsj (approved):

PR #1782, opened 2023-05-03
Add CVE-2020-29511 to go 1.19 as nacked.

dlorenc:
We nacked this in 1.20, so nack it back to 1.19.

Fixes:

Related:

#### For security-related PRs

### DIFF
--- a/go-1.19.yaml
+++ b/go-1.19.yaml
@@ -16,6 +16,7 @@ package:
 secfixes:
   "0":
     - CVE-2020-29509
+    - CVE-2020-29511
   1.19.3-r0:
     - CVE-2022-41716
   1.19.4-r0:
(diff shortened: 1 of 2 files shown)


luhring (commented):
Since this was addressed with a code update upstream, should we say "fixed" now instead?

dlorenc:
> Since this was addressed with a code update upstream, should we say "fixed" now instead?

Maybe - so this one was fixed back in go 1.17. I thought that if something was fixed prior to wolfi packaging it, then we used vulnerable_code_not_present, since we don't have a fixed version.

kaniini (approved):

luhring:
>I thought that if something was fixed prior to wolfi packaging it, then we used vulnerable_code_not_present, since we don't have a fixed version.

@dlorenc You're exactly right. I forgot this was an old fix. Thanks!

PR #1783, opened 2023-05-03
new package: openjdk-8 iced tea distribution

joshrwolf:
new package: openjdk-8

builds off the work done in #1681 to bootstrap openjdk-8 using the icedtea distribution

related: https://github.com/chainguard-dev/internal/issues/1575

#### For new package PRs only

### DIFF
--- a/openjdk-8.yaml
+++ b/openjdk-8.yaml
@@ -1,0 +1,222 @@
+package:
+  name: openjdk-8
+  version: 8u372-b07
+  epoch: 0
+  description: "IcedTea distribution of OpenJDK 8 (UNSUPPORTED)"
+  copyright:
+    - license: GPL-2.0-or-later
+  dependencies:
+    runtime:
+      - openjdk-8-jre
+      - java-cacerts
+
+environment:
+  contents:
+    packages:
+      - wolfi-base
+      - build-base
+      - bash
+      - findutils
+      - zip
+      - file
+      - gawk
+      - util-linux-dev
+      - libxslt-dev
+      - autoconf
+      - automake
+      - libtool
+      - linux-headers
+      - coreutils
+      - ca-certificates
+      - freetype-dev
+      - cups-dev
+      - libx11-dev
+      - libxext-dev
+      - libxrender-dev
+      - libxrandr-dev
+      - libxtst-dev
+      - libice-dev
+      - libxt-dev
+      - alsa-lib-dev
+      - libffi-dev
+      - krb5-dev
+      - pkgconf-dev
+      - attr-dev
+      - expat-dev
+      - fontconfig-dev
+      - libxi-dev
+      - libjpeg-dev
+      - giflib-dev
+      - lcms2-dev
+      - java-gcj-compat-default-jvm
+      - posix-libc-utils
+      - glib-dev
+      - gtk-2.0-dev
+      - libxinerama-dev
+      - libxcomposite-dev
+      - harfbuzz-dev
+      - pango-dev
+      - cairo-dev
+      - libxft-dev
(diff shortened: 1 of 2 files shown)


openjdk-8.yaml
```diff
@@ -0,0 +1,222 @@
+package:
+  name: openjdk-8
+  version: 8u372-b07
```
joshrwolf, line 3: seems like this isn't parseable by `wolfictl`, but I'm not sure this is the version naming we want to use given this is based on icedtea? follow up to that is how the icedtea versions translate to the openjdk versions...

patflynn: @rawlingsj may have advice.

openjdk-8.yaml
```diff
@@ -0,0 +1,222 @@
...
+  epoch: 0
+  description: "IcedTea distribution of OpenJDK 8 (UNSUPPORTED)"
```
patflynn, line 5: We do want to support JDK 8 since it's still supported upstream

openjdk-8.yaml
```diff
@@ -0,0 +1,222 @@
...
+      rm -f "${{targets.destdir}}"/usr/lib/jvm/java-1.8-openjdk/server/classes.jsa
+
+      # symlink to shared java cacerts store
+      rm -f "${{targets.destdir}}"/usr/lib/jvm/jvm-1.8-openjdk/jre/lib/security/cacerts
+      ln -sf /etc/ssl/certs/java/cacerts \
+        "${{targets.destdir}}"/usr/lib/jvm/java-1.8-openjdk/jre/lib/security/cacerts
+
+  - uses: strip
+
+subpackages:
```
patflynn, line 137: @kaniini I think will have suggestions related to -default-jvm

openjdk-8.yaml
```diff
@@ -0,0 +1,222 @@
+package:
+  name: openjdk-8
```
patflynn, line 2: should we call out the icedtea provenance? @kaniini

openjdk-8.yaml
joshrwolf, line 5:
```suggestion
  description: "IcedTea distribution of OpenJDK 8"
```

openjdk-8.yaml
joshrwolf: the default-jvm for 7 was done as a subpackage, I can do the equivalent here if that's how we want to keep things!

https://github.com/wolfi-dev/os/blob/main/openjdk-7.yaml#L267-L277

openjdk-8.yaml
kaniini: We do in the description.

openjdk-8.yaml
kaniini: Yes, that's how all JVMs should be packaged in Wolfi moving forward.  We will do similar for dotnet when 6 lands.

openjdk-8.yaml
kaniini: This should probably be 8.372.07.

openjdk-8.yaml
```diff
@@ -0,0 +1,222 @@
...
+      - krb5-dev
+      - pkgconf-dev
+      - attr-dev
+      - expat-dev
+      - fontconfig-dev
+      - libxi-dev
+      - libjpeg-dev
+      - giflib-dev
+      - lcms2-dev
+      - java-gcj-compat-default-jvm
```
kaniini, line 51: This should probably be openjdk-7-default-jvm.

```suggestion
      - openjdk-7-default-jvm
```

kaniini (approved):

PR #1796, opened 2023-05-04
patch pulumi to fix race condition which is breaking release builds

rawlingsj:
- [x] Patch source: https://patch-diff.githubusercontent.com/raw/pulumi/pulumi/pull/12821.patch


rawlingsj:
patch from @dlorenc

dlorenc (approved):

dlorenc:
image

PR #1802, opened 2023-05-04
gomplate: depend on latest go

imjasonh:
When gomplate was added in [three months ago](https://github.com/wolfi-dev/os/commit/8bdd1e0092448436f9fee86bf95df2dad9091724), it was 3.11.3 and required Go 1.19 specifically for some reason. 

It was bot-updated to 3.11.5 [two weeks ago](https://github.com/wolfi-dev/os/commit/f0b4930f9191862cb54c85eb3eb461f8c753f424) and seems to no longer require Go 1.19.

This PR makes it depend on the floating latest go, which is currently 1.20.4.

### DIFF
--- a/gomplate.yaml
+++ b/gomplate.yaml
@@ -32,6 +32,11 @@ pipeline:
         -X github.com/hairyhenderson/gomplate/v3/version.GitCommit=$(git rev-parse --short HEAD)" \
         github.com/hairyhenderson/gomplate/v3/cmd/gomplate
 
+      # Test that the binary can be run, since there have been issues with the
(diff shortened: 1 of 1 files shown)


imjasonh:
> It's because it imports this thing which blows up at runtime on new versions of Go, until someone tells it that they're safe: https://github.com/hairyhenderson/gomplate/blob/main/go.mod#L121
> 
> You might want to explicitly add a "test" in here to execute gomplate after it's built to make sure it doesn't fail in the future.

Wow, interesting.

I added a lil smoke test, let's see how it goes. 💣

dlorenc (approved):

PR #1803, opened 2023-05-04
bump epoch of go-1.20 dependents

joshrwolf:
bump epoch for all dependents of 1.20

Related: #1750, #1797

### DIFF
--- a/apko.yaml
+++ b/apko.yaml
@@ -1,7 +1,7 @@
 package:
   name: apko
   version: 0.8.0
-  epoch: 0
+  epoch: 1
   description: Build OCI images using APK directly without Dockerfile
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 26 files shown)


imjasonh:
Did you do this with a script? I hope so 😆 

In the past I've done this:

```
grep -l " - go$" *.yaml | xargs wolfictl bump
```

...and then spot-checked to see things that don't need to be bumped, like go-1.20.yaml etc.

It might be worth putting something like that in a script so we don't have to rely on bash history 😄

joshrwolf:
a few false positives I'm cleaning up now

joshrwolf:
> Did you do this with a script? I hope so 😆

definitely scripted :) I missed `$` in my regex, cleaning that up now

kaniini (approved):

PR #1806, opened 2023-05-04
apko: bump go deps to mitigate a GHSA

imjasonh:
This shows up in `cgr.dev/chainguard/apko`:

```
$ grype registry:cgr.dev/chainguard/apko
 ✔ Cataloged packages      [143 packages]
 ✔ Scanning image...       [3 vulnerabilities]
   ├── 0 critical, 1 high, 2 medium, 0 low, 0 negligible
   └── 1 fixed
NAME                       INSTALLED  FIXED-IN  TYPE       VULNERABILITY        SEVERITY
github.com/sigstore/rekor  v1.1.0     1.1.1     go-module  GHSA-2h5h-59f5-c5x9  Medium
```

### DIFF
--- a/apko.yaml
+++ b/apko.yaml
@@ -61,7 +61,7 @@ advisories:
     - timestamp: 2023-04-05T10:22:32.318046-04:00
       status: fixed
       fixed-version: 0.7.3-r1
-  GHSA-2h5h-59f5-c5x9:
+  CVE-2023-30551:
     - timestamp: 2023-05-04T12:30:46.874822-04:00
       status: fixed
       fixed-version: 0.8.0-r1
--- a/apko.yaml
+++ b/apko.yaml
@@ -26,7 +26,7 @@ pipeline:
   - runs: |
       cd apko
 
-      # GHSA-2h5h-59f5-c5x9
+      # GHSA-2h5h-59f5-c5x9 (CVE-2023-30551)
       go get -u github.com/sigstore/rekor@v1.1.1
       go mod tidy
 
--- a/apko.yaml
+++ b/apko.yaml
@@ -1,7 +1,8 @@
 package:
   name: apko
+  # When bumping the version check if the CVE/GHSA mitigations below can be removed.
   version: 0.8.0
-  epoch: 0
+  epoch: 1
   description: Build OCI images using APK directly without Dockerfile
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


apko.yaml
```diff
@@ -1,7 +1,8 @@
 package:
   name: apko
+  # When bumping the version check if the CVE/GHSA mitigations below can be removed.
```
kaniini, line 3: Experience in Alpine shows me that comments like this never really get read.  Maybe we need something other than a comment for these cases, like gating the patching with an `if` expression.

luhring: https://github.com/chainguard-dev/melange/issues/384

dlorenc: FWIW I sent this today which might help here: https://github.com/chainguard-dev/melange/commit/ece24c2854a2956bf48c539b312cde654765e6d5#diff-e9720019c7b001a10aa87ca89ef44a50a3942cfb22ea8c16ea3dd009d9992a0c

apko.yaml
```diff
@@ -24,6 +25,11 @@ pipeline:
 
   - runs: |
       cd apko
+
+      # GHSA-2h5h-59f5-c5x9
```
luhring, line 29:
```suggestion
      # GHSA-2h5h-59f5-c5x9 (CVE-2023-30551)
```

apko.yaml
```diff
@@ -53,3 +61,7 @@ advisories:
     - timestamp: 2023-04-05T10:22:32.318046-04:00
       status: fixed
       fixed-version: 0.7.3-r1
+  GHSA-2h5h-59f5-c5x9:
```
luhring, line 64:
```suggestion
  CVE-2023-30551:
```

luhring (commented):
Sorry for the drive-by review! I'd like us to be able to use GHSAs, but my suggestion comments point us at the corresponding CVE ID until we can add more intentional GHSA support to the tooling. WDYT?

apko.yaml
luhring: I think Dan's update to `go/build` helps us move this idea forward. I think what's left is to ensure that the `go get` is applied only if the specified module version is _later_ than what you'd see with `go list -m all`.

IMHO, that's the point where I think it makes sense to tie in automation that removes the version bumps. I think that functionality would probably go into `wolfictl update`.

Down the road we could also have dependabot-like automation (e.g. in `wolfictl`) that can automatically _add_ those version bump snippets — and add corresponding "fix" advisory entries — if we can determine that such a bump would resolve the vulnerability.

apko.yaml
```diff
@@ -39,6 +45,8 @@ secfixes:
     - CVE-2023-28840
     - CVE-2023-28841
     - CVE-2023-28842
+  0.8.0-r1:
+    - GHSA-2h5h-59f5-c5x9
```
luhring, line 49: I missed this one 🤦 1,000 apologies!!!!
```suggestion
    - CVE-2023-30551
```

luhring (approved):

PR #1836, opened 2023-05-05
Fix double-quoting of ldflags in go/build pipeline.

dlorenc:
This was spitting out a build-warning at the start like: `sh: github.com/kyverno/kyverno/pkg/version.BuildVersion=1.9.2: unknown operand`

I checked and the variable wasn't actually set either in the binary. Removing this quoting fixes it.

Fixes:

Related:

### DIFF
--- a/docker-credential-ecr-login.yaml
+++ b/docker-credential-ecr-login.yaml
@@ -1,7 +1,7 @@
 package:
   name: docker-credential-ecr-login
   version: 0.7.0
-  epoch: 1
+  epoch: 2
   description: Credential helper for Docker to use the AWS Elastic Container Registry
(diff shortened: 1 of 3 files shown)


rawlingsj (approved):

dlorenc:
Now:

```
% docker run -it cgr.dev/chainguard/test:latest-arm64 sh
/ #
/ # kyverno-cli version
Version: 1.9.2
Time: --
Git commit ID: --
```

Before:

```
% docker run -it cgr.dev/chainguard/test:latest-arm64 sh
/ #
/ # kyverno-cli version
Version: --
Time: --
Git commit ID: --
```

PR #1839, opened 2023-05-05
Refactor a bunch of go packages to use go/build.

dlorenc:
Fixes:

Related:

### DIFF
--- a/cosign.yaml
+++ b/cosign.yaml
@@ -1,7 +1,7 @@
 package:
   name: cosign
   version: 2.0.2
-  epoch: 1
+  epoch: 2
   description: Container Signing
   copyright:
     - license: Apache-2.0
@@ -15,6 +15,8 @@ environment:
       - busybox
       - ca-certificates-bundle
       - go
+  environment:
+    CGO_ENABLED: "0"
 
 pipeline:
   - uses: fetch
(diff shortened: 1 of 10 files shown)


dlorenc:
I think there are some tradeoffs with this approach, but overall I think it's probably correct long term. The risk of moving go-based builds to a standard pipeline is that we'll eventually "drift" from the exact build invocations from upstream projects, but the benefit (as I see it), is that we can more easily roll out changes to all go-based builds.

These could come in the form of new flags passed, or extra features like a go module proxy, automatic dependency sbom generation, etc...

cosign.yaml
```diff
@@ -15,17 +15,20 @@ environment:
...
 
-  - runs: |
-      CGO_ENABLED=0 go build \
-        -trimpath -ldflags "-buildid= -X sigs.k8s.io/release-utils/version.gitVersion=${{package.version}}" \
-        -o "${{targets.destdir}}/usr/bin/cosign" ./cmd/cosign
+  - uses: go/build
+    with:
+      packages: ./cmd/cosign
+      output: cosign
+      ldflags: -X sigs.k8s.io/release-utils/version.gitVersion=${{package.version}}
```
imjasonh, line 31: Just checking does variable expansion work as we expect here? I don't think cosign would complain if it was told its git version was the string `${{package.version}}`.

dlorenc: Yep, I tested that!

dlorenc:
> > These could come in the form of new flags passed, or extra features like a go module proxy, automatic dependency sbom generation, etc...
> 
> ...possibly shared build cache, remote hermetic builds, etc...
> 
> I love it. ❤️

If you loved it then you would have put an LGTM on it!

image

jonjohnsonjr (approved):

PR #1868, opened 2023-05-06
Start using CHECK-WHEN-VERSION-CHANGES lint rule.

dlorenc:
Let's try this new rule out in buildkitd. We can expand from there if it works well.

Fixes:

Related:

### Pre-review Checklist


kaniini (dismissed):

dlorenc:
This should fail first, then we can make it pass.

We have to wait for the next wolfictl sdk build to start.

dlorenc:
It worked!

image

kaniini (approved):

PR #1899, opened 2023-05-08
Bump some go dependencies to resolve CVEs in pulumi.

dlorenc:
Fixes:

Related:

#### For new package PRs only

- [ ] REQUIRED - The package is added to `packages.txt`

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/pulumi.yaml
+++ b/pulumi.yaml
@@ -1,7 +1,7 @@
 package:
   name: pulumi
   version: 3.66.0
-  epoch: 1
+  epoch: 2
   description: Infrastructure as Code in any programming language
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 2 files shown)


kaniini (approved):

jdolitsky (approved):

pulumi.yaml
```diff
@@ -41,6 +41,17 @@ pipeline:
           export GOBIN="${PULUMI_ROOT}/bin"
           mkdir -p "${{targets.destdir}}/usr/bin"
 
+          # Mitigate CVE-2022-41723
```
imjasonh, line 44: Do you want one of those blocking comments if the package gets bumped to 3.67.0?

dlorenc: I wanted to test those a bit more before using them widely. They don't work in the go/build setup since they only parse inside a pipeline (so it would work here, but not in the other yaml where it's structured) :(

PR #1907, opened 2023-05-08
fix: create dex db folder and add destination

developer-guy:
Fixes:

This is going to fix the "initialize storage: failed to perform migrations: creating migration table: unable to open database file: no such file or directory" error. They created a "/var/dex" directory in their [Dockerfile#56](https://github.com/dexidp/dex/blob/83c644755bfa1649bb5e73ca61dfb6d84f7f8a18/Dockerfile#L56).

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/dex.yaml
+++ b/dex.yaml
@@ -2,7 +2,7 @@ package:
   name: dex
   # When bumping the version check if the GHSA mitigations below can be removed.
   version: 2.36.0
-  epoch: 2
+  epoch: 3
   description: OpenID Connect (OIDC) identity and OAuth 2.0 provider with pluggable connectors
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


dex.yaml
```diff
@@ -43,6 +45,7 @@ pipeline:
       mkdir -p ${{targets.destdir}}/usr/bin/
       mkdir -p ${{targets.destdir}}/srv/dex/
       mkdir -p ${{targets.destdir}}/etc/dex
+      mkdir -p ${{targets.destdir}}/var/dex
```
rawlingsj, line 48: I'm not 100% sure what the trade-off's are for creating empty folders in the apk verses downstream when folks create an image.  I.e. if folks use apko like chainguard images, they tend to create folders + the necessary permissions in the apko.yaml.

kaniini (approved):

dex.yaml
kaniini: It is better to create required layout with the package than in the image config.  Consider non-apko uses of the packages, for example :)

PR #1937, opened 2023-05-09
add eol java 9

joshrwolf:
new package for java 9 and java 10. although EOL, they will be used for building java 11

#### For new package PRs only

### DIFF
--- a/openjdk-9.yaml
+++ b/openjdk-9.yaml
@@ -1,0 +1,213 @@
+package:
+  name: openjdk-9
+  version: 9.0.4
+  epoch: 0
+  description: "Oracle OpenJDK 9"
+  copyright:
+    - license: GPL-2.0-with-classpath-exception
+  dependencies:
+    runtime:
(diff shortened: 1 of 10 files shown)


kaniini (dismissed):

kaniini:
I would suggest lets focus on Java 9 first and then we can sort out the Java 10 build failure

joshrwolf:
I'm hitting weird errors building java 10 in CI that I can't repro "locally":

```console
Error occurred during initialization of boot layer
java.nio.file.InvalidPathExcepti��������������������������������home������������������������������������������������������������les
```

not sure if it's related to 9 _and_ 10 going in at the same time 🤔 but holding this PR for now

joshrwolf:
> I would suggest lets focus on Java 9 first and then we can sort out the Java 10 build failure

just saw this 😅 I'll change this PR to 9 only, and see if the error with 10 persists

kaniini (approved):

PR #1973, opened 2023-05-10
Fix for metric-server GOARCH defaulting to amd64

jdolitsky:


metrics-server.yaml
```diff
@@ -26,6 +26,14 @@ pipeline:
...
   - runs: |
+      # The Makefile sets GOARCH from the ARCH env var which defaults to amd64
+      set -x
+      case $(uname -m) in
+      aarch64) export ARCH="arm64" ;;
+      x86_64) export ARCH="amd64" ;;
+      *) export ARCH="$(uname -m)" ;;
+      esac
+
       make metrics-server
```
kaniini, line 37: Why not

```suggestion
      make metrics-server ARCH=$(go env GOARCH)
```

kaniini (dismissed):

metrics-server.yaml
jdolitsky: brilliant ......... BRILLIANT

jdolitsky: done

kaniini (approved):

PR #1985, opened 2023-05-12
consul helm chart compat subpackage

found-it:
Fixes:

Some [upstream helm charts ](https://github.com/hashicorp/consul-k8s/blob/bd16ab83383dbec5f8db680f39d85edd66b25b62/charts/consul/templates/client-daemonset.yaml#L283)require the binary to sit in `/usr/local/bin`. This PR adds a compat subpackage to install consul and the entrypoint in `/usr/local/bin`

Built locally and tested with

```sh
docker run -it --rm --entrypoint '/bin/sh' consul:local-amd64 -ec 'exec /usr/local/bin/docker-entrypoint.sh'
```

### DIFF
--- a/consul.yaml
+++ b/consul.yaml
@@ -36,11 +36,9 @@ pipeline:
 
           go mod tidy
           make linux
-
       - runs: |
           mkdir -p ${{targets.destdir}}/usr/bin
           mv ./pkg/bin/linux_*/consul ${{targets.destdir}}/usr/bin/consul
-
       - uses: strip
 
 subpackages:
--- a/consul.yaml
+++ b/consul.yaml
@@ -38,6 +38,8 @@ pipeline:
           make linux
       - runs: |
           mkdir -p ${{targets.destdir}}/bin
+
+          # The docker-entrypoint.sh expects the binary to be in /bin so put it there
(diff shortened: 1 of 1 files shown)


consul.yaml
```diff
@@ -20,32 +20,51 @@ pipeline:
...
+        - busybox
+        - dumb-init
+        - su-exec
+
+  - name: ${{package.name}}-compat
+    description: "Compatiblity package for locating binaries according to upstream helm charts"
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}/usr/local/bin"
+          cp "${{targets.destdir}}/usr/bin/consul" "${{targets.subpkgdir}}/usr/local/bin"
```
dlorenc, line 61: Should this be a symlink instead of copying the binary?

found-it: Yep good idea

found-it: Actually, I sorta take it back. You're right, it should prefer a symlink. However the entrypoint is expecting `/bin/consul` and it can't be a symlink because the entrypoint sets capabilities on it. On direction from @kaniini I just put the binary in `/bin` in the main package

imjasonh (approved):

PR #2047, opened 2023-05-15
don't cache security.json

imjasonh:
The security.json feed doesn't need to be cached -- it's not requested often, it's small, and caching can obscure important updates.

### DIFF
--- a/.github/workflows/secdb-production.yaml
+++ b/.github/workflows/secdb-production.yaml
@@ -50,4 +50,8 @@ jobs:
 
       - name: 'Upload the security database to a bucket'
         run: |
-          gcloud --quiet alpha storage cp --recursive ${{ github.workspace }}/security.json gs://wolfi-production-registry-destination/os/
+          # Don't cache the security.json.
+          gcloud --quiet alpha storage cp \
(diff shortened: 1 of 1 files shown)


.github/workflows/secdb-production.yaml
```diff
@@ -50,4 +50,8 @@ jobs:
 
       - name: 'Upload the security database to a bucket'
         run: |
-          gcloud --quiet alpha storage cp --recursive ${{ github.workspace }}/security.json gs://wolfi-production-registry-destination/os/
+          # Don't cache the security.json.
+          gcloud --quiet alpha storage cp \
+              --cache-control=no-store \
+              ${{ github.workspace }}/security.json \
+              gs://wolfi-production-registry-destination/os/
```
luhring, line 57: I think this change needs to go in the `advisories` repo, since it does the secdb build/publish now

luhring (changes requested):

.github/workflows/secdb-production.yaml
luhring: How does this look? https://github.com/wolfi-dev/advisories/pull/2

PR #2061, opened 2023-05-15
Feature: set `SOURCE_DATE_EPOCH` based on the config file.

### LINKED ISSUE #535: Feature: set `SOURCE_DATE_EPOCH` based on the mtime of melange configs
_This is related to [this](https://github.com/chainguard-dev/melange/issues/421) issue._

I was talking to @kaniini about how we get better `SOURCE_DATE_EPOCH` for our APKs (to feed into a smarter default for our images), and the `tl;dr` is:

> We should set the `SOURCE_DATE_EPOCH` for our `melange` builds to match the last-modified-time of the melange configuration.

Initially, I was overcomplicating this, thinking that it should be some combination of the _proper_ `SOURCE_DATE_EPOCH=$(git log -1 --pretty=%ct)` and the melange config's `mtime`, but unlike `apko` (which floats forward APKs), `melange` essentially always pins to one of a release tag or commit SHA.  So in essence, the melange config's `mtime` is _**always**_ after the commit timestamp because it references it!

I believe that this should result in us getting timestamps on our APKs, which change exactly when they should.

mattmoor:
:gift: This sets the `SOURCE_DATE_EPOCH` for a particular APK to be [the standard `git` command](https://reproducible-builds.org/docs/source-date-epoch/) with the qualification: `--follow {pkg}.yaml`.

The logic here is simple yet subtle: Our SDE should reflect the last change that influenced the APK, and the melange config fixes the version used to build, so it is in essence the last node in a sort of virtual merkle DAG.

If the above weren't true, then it would somewhat fundamentally indicate that our builds are not reproducible and that is something we should then fix.

Fixes: https://github.com/chainguard-images/images/issues/535

/kind feature

This also "touches" the `ko` package in an effort to observe this behavior in action when this change lands. 🤞

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ $(pkgtarget): ${KEY}
 	mkdir -p ./$(sourcedir)/
 ifdef SOURCE_DATE_EPOCH
 	$(eval CONFIG_DATE_EPOCH := $(SOURCE_DATE_EPOCH))
-oelse
+else
 	$(eval CONFIG_DATE_EPOCH := $(shell git log -1 --pretty=%ct --follow $(pkgname).yaml))
 endif
 	SOURCE_DATE_EPOCH=${CONFIG_DATE_EPOCH} ${MELANGE} build $(pkgname).yaml ${MELANGE_OPTS} --source-dir ./$(sourcedir)/ --log-policy builtin:stderr,${TARGETDIR}/buildlogs/$(pkgfullname).log
--- a/Makefile
+++ b/Makefile
@@ -40,8 +40,12 @@ $(eval pkgtarget = $(TARGETDIR)/$(pkgfullname).apk)
 packages/$(pkgname): $(pkgtarget)
 $(pkgtarget): ${KEY}
 	mkdir -p ./$(sourcedir)/
-	$(eval SOURCE_DATE_EPOCH ?= $(shell git log -1 --pretty=%ct --follow $(pkgname).yaml))
-	SOURCE_DATE_EPOCH=${SOURCE_DATE_EPOCH} ${MELANGE} build $(pkgname).yaml ${MELANGE_OPTS} --source-dir ./$(sourcedir)/ --log-policy builtin:stderr,${TARGETDIR}/buildlogs/$(pkgfullname).log
+ifdef SOURCE_DATE_EPOCH
+	$(eval CONFIG_DATE_EPOCH := $(SOURCE_DATE_EPOCH))
+oelse
+	$(eval CONFIG_DATE_EPOCH := $(shell git log -1 --pretty=%ct --follow $(pkgname).yaml))
+endif
+	SOURCE_DATE_EPOCH=${CONFIG_DATE_EPOCH} ${MELANGE} build $(pkgname).yaml ${MELANGE_OPTS} --source-dir ./$(sourcedir)/ --log-policy builtin:stderr,${TARGETDIR}/buildlogs/$(pkgfullname).log
 
 endef
 
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ $(eval pkgtarget = $(TARGETDIR)/$(pkgfullname).apk)
 packages/$(pkgname): $(pkgtarget)
 $(pkgtarget): ${KEY}
 	mkdir -p ./$(sourcedir)/
-	$(eval SOURCE_DATE_EPOCH := $(shell git log -1 --pretty=%ct --follow $(pkgname).yaml))
+	$(eval SOURCE_DATE_EPOCH ?= $(shell git log -1 --pretty=%ct --follow $(pkgname).yaml))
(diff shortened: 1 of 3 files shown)


deitch:
Try doing it in 2 lines. I am not sure it will solve the CI problem, but it does ensure that `SOURCE_DATE_EPOCH` is set, and that you can override it:

```make
$(pkgtarget): ${KEY}
	mkdir -p ./$(sourcedir)/
	$(eval SOURCE_DATE_EPOCH := $(shell git log -1 --pretty=%ct --follow $(pkgname).yaml))
	SOURCE_DATE_EPOCH=${SOURCE_DATE_EPOCH} ${MELANGE} build $(pkgname).yaml ${MELANGE_OPTS} --source-dir ./$(sourcedir)/ --log-policy builtin:stderr,${TARGETDIR}/buildlogs/$(pkgfullname).log
```

deitch:
That having been said, I really like the approach. So once we have CI clean and the make working, 👍

deitch (dismissed):

deitch:
Something strange about how the command is being constructed.

Makefile
```diff
@@ -41,6 +40,7 @@ $(eval pkgtarget = $(TARGETDIR)/$(pkgfullname).apk)
 packages/$(pkgname): $(pkgtarget)
 $(pkgtarget): ${KEY}
 	mkdir -p ./$(sourcedir)/
+	$(eval SOURCE_DATE_EPOCH ?= $(shell git log -1 --pretty=%ct --follow $(pkgname).yaml))
```
mattmoor, line 43: @kaniini will this work if this target it invoked for multiple packages?  I am unfamiliar with the scoping rules of Make.

mattmoor:
So my anecdotal check with `crane` didn't work, both got the same `SOURCE_DATE_EPOCH`, but I suspect it is because of a merge commit being created for the PR, so I am retrying now with a rebase. 🤞

mattmoor:
Hmm, it is still getting a merge commit `aab59f4848f6197cb3b571e5ee9cf1666a7dfdba` (in the logs) vs. `475c82ba6eec9b2cb12dbecb23fda6a6ab9181c7` (in my branch)

mattmoor:
Nevermind, my "Dr. Empirical" check isn't going to prove anything anyways because I see we invoke `make` separately for things presubmit but not (it seems) post-submit.  I'm going to back out the `crane` commit and defer to @kaniini here on the best way to allow overrides safely.

mattmoor:
Ok, I crafted a small repro for this locally and it's busted:
```Makefile
all: foo bar baz
	echo ${BLAH}

foo:
	$(eval BLAH ?= "foo")
	echo ${BLAH}

bar:
	$(eval BLAH ?= "bar")
	echo ${BLAH}

baz:
	$(eval BLAH ?= "baz")
	echo ${BLAH}
```

yields:
```
make all
echo "foo"
foo
echo "foo"
foo
echo "foo"
foo
echo "foo"
foo
```

I'll devise something that passes this test and supports overrides.

mattmoor:
Ok, this Makefile seems to exhibit the behavior we want, I am updating the PR now to match.

```
all: foo bar baz
	echo ${BLAH_OVERRIDE}
	echo ${BLAH}

foo:
ifdef BLAH_OVERRIDE
	$(eval BLAH := $(BLAH_OVERRIDE))
else
	$(eval BLAH := "foo")
endif
	echo ${BLAH}

bar:
ifdef BLAH_OVERRIDE
	$(eval BLAH := $(BLAH_OVERRIDE))
else
	$(eval BLAH := "bar")
endif
	echo ${BLAH}

baz:
ifdef BLAH_OVERRIDE
	$(eval BLAH := $(BLAH_OVERRIDE))
else
	$(eval BLAH := "baz")
endif
	echo ${BLAH}
```

Makefile
```diff
@@ -41,7 +40,12 @@ $(eval pkgtarget = $(TARGETDIR)/$(pkgfullname).apk)
 packages/$(pkgname): $(pkgtarget)
 $(pkgtarget): ${KEY}
 	mkdir -p ./$(sourcedir)/
-	SOURCE_DATE_EPOCH=${SOURCE_DATE_EPOCH} ${MELANGE} build $(pkgname).yaml ${MELANGE_OPTS} --source-dir ./$(sourcedir)/ --log-policy builtin:stderr,${TARGETDIR}/buildlogs/$(pkgfullname).log
+ifdef SOURCE_DATE_EPOCH
+	$(eval CONFIG_DATE_EPOCH := $(SOURCE_DATE_EPOCH))
+oelse
```
jdolitsky, line 45:
```suggestion
else
```

jdolitsky (approved):

mattmoor:
This needs review from @wolfi-dev/wolfi-build-maintainers

kaniini (approved):

PR #2065, opened 2023-05-15
add testing to redis package builds

joshrwolf:
Adds upstream tests to redis package builds


tuananh:
i thought we're going to leave test to upstream?

PR #2067, opened 2023-05-15
add wavefront-proxy

ajayk:
Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/wavefront-proxy.yaml
+++ b/wavefront-proxy.yaml
@@ -28,7 +28,7 @@ pipeline:
 
   - runs: |
       export LANG=en_US.UTF-8
-      export JAVA_HOME=/usr/lib/jvm/openjdk
+      export JAVA_HOME=/usr/lib/jvm/java-11-openjdk
       mvn -f proxy --batch-mode clean package -DskipTests -DskipFormatCode
       mkdir -p ${{targets.destdir}}/usr/share/java/wavefront/wavefront-proxy
       mv /home/build/proxy/target/proxy-${{package.version}}-spring-boot.jar ${{targets.destdir}}/usr/share/java/wavefront/wavefront-proxy/wavefront-proxy.jar
--- a/wavefront-proxy.yaml
+++ b/wavefront-proxy.yaml
@@ -1,0 +1,74 @@
+package:
+  name: wavefront-proxy
+  version: 12.4
+  epoch: 0
+  description: Wavefront Proxy Project
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - openjdk-11-jre
+
+environment:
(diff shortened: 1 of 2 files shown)


wavefront-proxy.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+  - runs: |
+      export LANG=en_US.UTF-8
+      export JAVA_HOME=/usr/lib/jvm/openjdk
+      mvn -f proxy --batch-mode clean package -DskipTests -DskipFormatCode
+      mkdir -p ${{targets.destdir}}/opt/wavefront/wavefront-proxy
+      mv /home/build/proxy/target/proxy-${{package.version}}-spring-boot.jar ${{targets.destdir}}/opt/wavefront/wavefront-proxy/wavefront-proxy.jar
+
+subpackages:
+  - name: ${{package.name}}-config
```
kaniini, line 37: Why not `name: wavefront-proxy-config`?

imjasonh (dismissed):

wavefront-proxy.yaml
```diff
@@ -0,0 +1,74 @@
...
+    with:
+      repository: https://github.com/wavefronthq/wavefront-proxy
+      tag: proxy-${{package.version}}
+      expected-commit: 7170d26b802e5f4b2201c3a38c2d11e6e3ad719a
+
+  - runs: |
+      export LANG=en_US.UTF-8
+      export JAVA_HOME=/usr/lib/jvm/openjdk
+      mvn -f proxy --batch-mode clean package -DskipTests -DskipFormatCode
+      mkdir -p ${{targets.destdir}}/opt/wavefront/wavefront-proxy
```
kaniini, line 33: Packages should not go into `/opt`.

kaniini (changes requested):

wavefront-proxy.yaml
ajayk: the [run.sh](https://github.com/wavefrontHQ/wavefront-proxy/blob/master/docker/run.sh#L62) has a reference /opt , so we would need to override/re-write the run.sh when we do an image with a new reference ?

The only other package reference I found was from conda where the package was going to /opt

dlorenc: We could do it in a compat subpackage if they hard code it.

ajayk: I moved the packages to `/usr/share/java/wavefront/wavefront-proxy` add a replace in the compat subpackage

rawlingsj (dismissed):

rawlingsj (dismissed):

imjasonh (approved):

kaniini (approved):

PR #2075, opened 2023-05-16
openjdk-11: upgrade to 11.0.20.2, align with other JDK packaging

kaniini:
- Upgrade to 11.0.20.2
- Use OpenJDK 10 as bootstrap JDK
- Align with the new `default-jvm` contract.

PR #2081, opened 2023-05-16
add openjdk 12, 13, and 14

joshrwolf:
java 12, 13, and 14: new package

Related: #2075 

#### For new package PRs only

### DIFF
--- a/openjdk-12.yaml
+++ b/openjdk-12.yaml
@@ -1,0 +1,220 @@
+package:
+  name: openjdk-12
+  version: 12.0.2.10
+  epoch: 0
+  description:
+  copyright:
+    - license: GPL-2.0-only
+  dependencies:
+    runtime:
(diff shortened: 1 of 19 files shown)


openjdk-12.yaml
```diff
@@ -0,0 +1,219 @@
+package:
+  name: openjdk-12
+  version: 12.0.2.10
+  epoch: 0
+  description:
+  copyright:
+    - license: GPL-2.0-only
+  dependencies:
+    runtime:
+      - freetype
```
joshrwolf, line 10:
```suggestion
      - openjdk-12-jre
      - freetype
```

joshrwolf: should we have the jdk depend on the jre so `$_java_home/bin/java` is available from `apk add openjdk-12`?

kaniini (approved):

PR #2092, opened 2023-05-16
ensure openjdk-11 jdk depends on jre

joshrwolf:
Fixes: ensures jdk depends on jre for openjdk 11 to make `apk add openjdk-11` work as one would expect


joshrwolf:
`wolfictl check updates` is expected to fail for openjdk-11

kaniini (approved):

PR #2095, opened 2023-05-16
add openjdk-15, 16, and updated 17

joshrwolf:
new packages: openjdk-15 and 16

updated packages: openjdk-17, zookeeper, keycloak, jenkins, cloudwatch-exporter

this brings openjdk17 up to par with the updated way we're packaging openjdk

Related:

#### For new package PRs only

- [ ] This PR is marked as fixing a pre-existing package request bug
  - [ ] Alternatively, the PR is marked as related to a pre-existing package request bug, such as a dependency
- [x] REQUIRED - The package is available under an OSI-approved or FSF-approved license
- [x] REQUIRED - The version of the package is still receiving security updates
- [x] REQUIRED - The package is added to `packages.txt`

PR #2108, opened 2023-05-17
remove custom source dir in packages.txt

deitch:
There are a few packages which have shared source directories. These have been recorded as custom fields (comma-separated) in `packages.txt`, along with the logic in Makefile to separate them.

That logic is a little complex to parse, leading to potential failures.

More importantly, it isn't something that the graph can understand. As soon as the last openjdk-17 issue is resolved with #2095, we intend to get rid of `packages.txt` entirely, and replace it with an auto-generated dependency graph, which, as above, will not understand these custom directories.

This PR:

* removes the custom directories from `packages.txt`
* removes the logic for processing custom directories from `Makefile`
* replaces those custom directories (which really are just shared directories among multiple packages) with symlinks from the proper name to the shared directory
* updates the documentation to describe how to do this

Eventually, melange itself is likely to understand a "source directory" command in the yaml config files, which will make all of this go away. This is a faster fix, allowing us to get here immediately in a well-understood way.

Once this and #2095 are in, we will eliminate `packages.txt`.


deitch:
Testing. I ran the following commands on current `main` and this branch, and compared them:

* `make list` - this just lists all of the packages in order. The two were identical.
* `make -n all` - this lists what build commands would be executed. The two were _almost_ identical, except for the changed directories for the 9 which switched from custom directory to symlinks

imjasonh (approved):

kaniini (approved):

PR #2110, opened 2023-05-17
use -default-jvm for all openjdk-11 dependencies

joshrwolf:
Fixes: 

Related:

### DIFF
--- a/bazel-6.yaml
+++ b/bazel-6.yaml
@@ -1,7 +1,7 @@
 package:
   name: bazel-6
   version: 6.2.0
-  epoch: 0
+  epoch: 1
   description: Bazel is an open-source build and test tool
   copyright:
     - license: Apache-2.0
@@ -15,7 +15,7 @@ environment:
       - python3
       - zip
       - bash
-      - openjdk-11
+      - openjdk-11-default-jvm
 
 pipeline:
   - uses: fetch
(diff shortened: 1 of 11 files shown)


bazel-6.yaml
```diff
@@ -15,7 +15,7 @@ environment:
       - python3
       - zip
       - bash
-      - openjdk-11
+      - openjdk-11-default-jvm
```
patflynn, line 18: here too

gradle-stage0.yaml
```diff
@@ -1,13 +1,13 @@
...
   version: 8.0.1
-  epoch: 0
+  epoch: 1
   description: A Java project management and project comprehension tool.
   copyright:
     - license: Apache-2.0
   dependencies:
     runtime:
-      - openjdk-11-jre
+      - openjdk-11-default-jvm
```
patflynn, line 10: meta: we need to decide whether we runtime include jvm or not.  @kaniini what happens if two packages in the dependency tree provide default-jvm but for different jdk versions?

patflynn:
@kaniini @joshrwolf I think we want to close this out right? My understanding is that we should set JAVA_HOME for each of these packages.

joshrwolf:
closing this in favor of `JAVA_HOME` and piece meal `-default-jvm` alternatives

PR #2121, opened 2023-05-17
try to improve contributing docs

imjasonh:
- set `MELANGE=$(shell which melange)` by default in the Makefile, remove guidance to set this yourself
- remove mention of `MELANGE_DIR`
- update old docs that mentioned a Chainguard `sdk` image
- use latest `ghcr.io/wolfi-dev/sdk` image, which supports `--log-policy` flag -- I don't know why the digestabot isn't updating this

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,7 @@
 
 ## Setup development environment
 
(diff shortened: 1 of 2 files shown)


deitch (approved):
Yes! Every time I try and run, it fails with melange not found because I forgot to set the env var, I am surprised. And with it `MELANGE ?=`, you always can override it.

jonjohnsonjr (approved):

PR #2123, opened 2023-05-17
document new use of openjdk packages

joshrwolf:
Add documentation for the usage of `openjdk` packages following the new subpackage structure.

### DIFF
--- a/docs/java.md
+++ b/docs/java.md
@@ -1,0 +1,17 @@
+# Java
+
+This doc outlines how Java is packaged within Wolfi, and how to determine which package is right for your needs.
+
+Each version of Java provided by Wolfi consists of the following packages:
+
+- `openjdk-#-jre`: the Java Runtime Environment (JRE)
+- `openjdk-#`: the Java Development Kit (JDK) and the JRE
+
+The installation location for these packages is `/usr/lib/jvm/java-#-openjdk`. It is important to note that this path is **not** added to the default `$PATH` when installed. Instead, we strongly recommend that you configure `JAVA_HOME=/usr/lib/jvm/java-#-openjdk` post installation, and if you require the Java tooling on your default `$PATH`, to add it post installation. An example of this being done is below:
+
+```bash
+export JAVA_HOME=/usr/lib/jvm/java-#-openjdk
+export PATH=$JAVA_HOME/bin:$PATH
+```
+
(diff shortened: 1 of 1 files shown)


docs/java.md
```diff
@@ -0,0 +1,17 @@
...
+- `openjdk-#`: the Java Development Kit (JDK) and the JRE
+
+The installation location for these packages is `/usr/lib/jvm/java-#-openjdk`. It is important to note that this path is **not** added to the default `$PATH` when installed. Instead, we strongly recommend that you configure `JAVA_HOME=/usr/lib/jvm/java-#-openjdk` post installation, and if you require the Java tooling on your default `$PATH`, to add it post installation. An example of this being done is below:
+
+```bash
+export JAVA_HOME=/usr/lib/jvm/java-#-openjdk
+export PATH=$JAVA_HOME/bin:$PATH
+```
+
+Alternatively, a compatibility package: `openjdk-#-default-jvm` is provided that will provide both the JDK, JRE, and ensure they are added to the default `$PATH`.
```
patflynn, line 17: the default-jvm only provides the JRE. You'll need install the JDK package separately if you want all the JDK tools to be available (and work through the symlinks)

docs/java.md
```diff
@@ -0,0 +1,17 @@
+# Java
+
+This doc outlines how Java is packaged within Wolfi, and how to determine which package is right for your needs.
+
+Each version of Java provided by Wolfi consists of the following packages:
+
+- `openjdk-#-jre`: the Java Runtime Environment (JRE)
+- `openjdk-#`: the Java Development Kit (JDK) and the JRE
```
patflynn, line 8: JAVA_HOME only works for use-cases where the jdk isn't being used directly.  I don't think we should present ..default-jvm as a compatibility thing. I'd list it front and center as the third Java package available for each version, and describe how it makes it the default JVM and connects the /usr/bin/symlinks to your JDK of choice.

patflynn:
@joshrwolf let's address feedback and merge :pray:

patflynn:
or I guess you're waiting for 17 to hit? If so perhaps we can copy/paste the wording into an announcement now?

docs/java.md
patflynn: We should also describe the default-jvm symlink pattern so that folks understand what's going on there.

kaniini:
AFAIK we have an announcement in Zendesk that is ready to go that has basically incorporated the wording here.

PR #2154, opened 2023-05-18
add newrelic/newrelic-fluent-bit-output

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/newrelic-fluent-bit-output.yaml
+++ b/newrelic-fluent-bit-output.yaml
@@ -1,0 +1,38 @@
+package:
+  name: newrelic-fluent-bit-output
+  version: 1.16.0
+  epoch: 0
+  description: A Fluent Bit output plugin that sends logs to New Relic
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - wolfi-baselayout
+      - busybox
+      - build-base
+      - go
+      - ca-certificates-bundle
+
+pipeline:
+  # We can't use go/install because this requires specific ldflags to set the version
+  - uses: git-checkout
+    with:
+      repository: https://github.com/newrelic/newrelic-fluent-bit-output
+      tag: v${{package.version}}
+      expected-commit: 8e3aa7c9d829d2c36d8007c2d44295cefd5ee284
+
+  - runs: |
+      ARCH=$(go env GOARCH)
+      make VERSION=v${{ package.version }} linux/${ARCH}
+      mkdir -p ${{targets.destdir}}/fluent-bit/bin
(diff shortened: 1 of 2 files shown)


newrelic-fluent-bit-output.yaml
```diff
@@ -0,0 +1,44 @@
...
+    runs: |
+      make VERSION=v${{ package.version }} linux/amd64
+      mkdir -p ${{targets.destdir}}/newrelic/newrelic-fluent-bit-output
+      cp ./out_newrelic-linux-amd64-v${{ package.version }}.so ${{targets.destdir}}/newrelic/newrelic-fluent-bit-output
+
+  - if: ${{build.arch}} == 'aarch64'
+    runs: |
+      make VERSION=v${{ package.version }} linux/arm64
+      mkdir -p ${{targets.destdir}}/newrelic/newrelic-fluent-bit-output
+      cp ./out_newrelic-linux-arm64-v${{ package.version }}.so ${{targets.destdir}}/newrelic/newrelic-fluent-bit-output
```
imjasonh, line 36: I think we should make this one step, that sets `ARCH=$(go env GOARCH)` and passes that around instead of `amd64` and `arm64`.

The reason is, when we add a third arch, we'll want to make sure we actually produce a useful apk for it, without having to check every package for `if`s like this.

It looks like it should be possible to do, let me know if you hit a snag.

developer-guy: brilliant idea, thanks!

newrelic-fluent-bit-output.yaml
```diff
@@ -0,0 +1,38 @@
...
+    with:
+      repository: https://github.com/newrelic/newrelic-fluent-bit-output
+      tag: v${{package.version}}
+      expected-commit: 8e3aa7c9d829d2c36d8007c2d44295cefd5ee284
+
+  - runs: |
+      ARCH=$(go env GOARCH)
+      make VERSION=v${{ package.version }} linux/${ARCH}
+      mkdir -p ${{targets.destdir}}/newrelic/newrelic-fluent-bit-output
+      cp ./out_newrelic-linux-${ARCH}-v${{ package.version }}.so ${{targets.destdir}}/newrelic/newrelic-fluent-bit-output
```
imjasonh, line 30: In the upstream image this seems to be [`/fluent-bit/bin/out_newrelic.so`](https://github.com/newrelic/newrelic-fluent-bit-output/blob/8e3aa7c9d829d2c36d8007c2d44295cefd5ee284/Dockerfile#LL24C83-L24C117), can we do there here too?

imjasonh: I think the upstream filename is better, out_newrelic.so

imjasonh (approved):

PR #2155, opened 2023-05-18
Bring back "build date epoch"

### LINKED ISSUE #535: Feature: set `SOURCE_DATE_EPOCH` based on the mtime of melange configs

mattmoor:
The first commit here is a pure revert of the revert commit.

The second commit here drops the `ko` epoch bump which we withdrew.  I will bump this to `6` in a subsequent change once this lands.

The main difference is that we have swapped `kontext` implementations to one that now supports [what we need](https://github.com/chainguard-dev/kontext/blob/347f191bd8d465631acc627f8a8309e01016fce0/.github/workflows/e2e-test.yaml#L64C21-L65).

Fixes: https://github.com/chainguard-images/images/issues/535

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -59,6 +59,12 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: 'Trust the github workspace'
+        run: |
+          # This is to avoid fatal errors about "dubious ownership" because we are
+          # running inside of a container action with the workspace mounted in.
+          git config --global --add safe.directory "$GITHUB_WORKSPACE"
+
       - name: 'Generate local signing key'
         run: |
           make MELANGE="melange" local-melange.rsa
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,6 @@ TARGETDIR = packages/${ARCH}
 MELANGE ?= $(shell which melange)
 KEY ?= local-melange.rsa
 REPO ?= $(shell pwd)/packages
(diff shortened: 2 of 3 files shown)


mattmoor:
I'm opening this as a draft because I think I realized we will need the following for the `log` command to work:
```
git config --global --add safe.directory /workspace
```

imjasonh (approved):

mattmoor:
I think we also need this: https://github.com/wolfi-dev/wolfictl/pull/222

mattmoor:
The above should be pulled in by: https://github.com/wolfi-dev/os/pull/2157 🤞

deitch (approved):

PR #2182, opened 2023-05-19
rabbitmq: bump epoch to rebuild with latest erlang

imjasonh:


kaniini (dismissed):

kaniini (approved):

imjasonh:
```
⚠️  x86_64    | ** (Mix) httpc request failed with: {:failed_connect, [{:to_address, {'objects.githubusercontent.com', 443}}, {:inet, [:inet], {:tls_alert, {:handshake_failure, 'TLS client: In state wait_cert_cr at ssl_handshake.erl:2111 generated CLIENT ALERT: Fatal - Handshake Failure\n {bad_cert,hostname_check_failed}'}}}]}
```

This reads like a flake, retrying

imjasonh:
After https://github.com/wolfi-dev/os/pull/2184 this should have no effect for `rabbitmq` (I think), but it should fix elixir, which wasn't noting its runtime dependency on erlang.

PR #2206, opened 2023-05-21
Add new packages for kubeflow-katib

rawlingsj:
This includes:

katib-cert-generator
katib-db-manager
katib-earlystopping
katib-katib-controller
katib-metricscollector
katib-suggestion-skopt-enas
katib-suggestion-goptuna
katib-suggestion-hyperband
katib-suggestion-hyperopt
katib-suggestion-nas-darts
katib-suggestion-nas-enas
katib-suggestion-optuna-enas
katib-suggestion-pbt-enas

#### For new package PRs only

### DIFF
--- a/kubeflow-katib.yaml
+++ b/kubeflow-katib.yaml
@@ -1,0 +1,80 @@
+package:
+  epoch: 0
+  name: kubeflow-katib
+  version: 0.15.0
+  description: Kubeflow Katib services
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - wget
+      - ca-certificates
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - go
+      - build-base
+      - python-3.10-dev
+      - py3.10-pip
+      - gfortran
+      - pcre-dev
+      # libopenblas-dev liblapack-dev not sure if these are needed, builds fine without.  From https://github.com/kubeflow/katib/blob/b9dc63efb55bec19bdd80ca6f2930a65795f7c41/cmd/earlystopping/medianstop/v1beta1/Dockerfile#L10
+
+data:
+  - name: go-builds
+    items:
+      cert-generator: cert-generator/v1beta1
+      controller: katib-controller/v1beta1
+      db-manager: db-manager/v1beta1
+      file-metricscollector: metricscollector/v1beta1/file-metricscollector
+      suggestion-goptuna: suggestion/goptuna/v1beta1
+
+  - name: python-builds
+    items:
+      earlystopping: earlystopping/medianstop/v1beta1
+      tfevent-metricscollector: metricscollector/v1beta1/tfevent-metricscollector
+      suggestion-hyperband: suggestion/hyperband/v1beta1
+      suggestion-hyperopt: suggestion/hyperopt/v1beta1
+      suggestion-nas-darts: suggestion/nas/darts/v1beta1
+      suggestion-nas-enas: suggestion/nas/enas/v1beta1
+      suggestion-optuna-enas: suggestion/optuna/v1beta1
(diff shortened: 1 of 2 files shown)


rawlingsj:
Build will fail, depends on https://github.com/chainguard-dev/melange/pull/455

kubeflow-katib.yaml
```diff
@@ -0,0 +1,80 @@
...
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - go
+      - build-base
+      - python-3.10-dev
+      - py3.10-pip
+      - gfortran
+      - pcre-dev
+      # libopenblas-dev liblapack-dev not sure if these are needed, builds fine without.  From https://github.com/kubeflow/katib/blob/b9dc63efb55bec19bdd80ca6f2930a65795f7c41/cmd/earlystopping/medianstop/v1beta1/Dockerfile#L10
```
rawlingsj, line 24: Upstream has libopenblas-dev and liblapack-dev during the build but not sure if we need them in wolfi as the builds are all working?

kubeflow-katib.yaml
```diff
@@ -0,0 +1,80 @@
...
+      suggestion-skopt-enas: suggestion/skopt/v1beta1
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/kubeflow/katib
+      expected-commit: 05ab6f012ab6723609700a47a6cfa1845080e9b6
+      tag: v${{package.version}}
+
+subpackages:
```
rawlingsj, line 54: I initially added pipelines to the subpackages, but now wondering if we should build all the packages in the main melange pipeline, `strip`  them and then `mv` the binaries in the subpackages from the `destdir` to `subpkgdir`

kubeflow-katib.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+  - range: python-builds
+    name: katib-${{range.key}}
+    description: Kubeflow Katib ${{range.key}}
+    pipeline:
+      - runs: |
+          cd ./cmd/${{range.value}}
+          pip install --prefer-binary -r requirements.txt --prefix=/usr --root=${{targets.subpkgdir}}
+
+          install -Dm755 ./main.py ${{targets.subpkgdir}}/usr/bin/
```
imjasonh, line 74: So one thing about this is that we won't be able to install `katib-a` and `katib-b` in the same image, since they'll conflict on which gets to provide `/usr/bin/main.py`. Should we name them uniquely to avoid this?

kubeflow-katib.yaml
rawlingsj: range only works for for subpackages

kubeflow-katib.yaml
rawlingsj: yeah good point!

rawlingsj: looking into this a bit more - I wonder where we should put these python builds, should this even go into `/usr/bin`. upstream they use `/opt` but for wolfi we avoid using this folder. 

additionally the auto diff comment failed as the message body failed to POST to the github comment.  A manual tree for one of the python related packages shows a huge number of files https://gist.github.com/rawlingsj/0102e404248ba1381153900997b2c192

so I think there may be something else wrong with these python based subpackages.

After looking at the upstream image

```
docker run -ti --entrypoint bash kubeflowkatib/tfevent-metrics-collector:latest
```
I can see the core python related files are written to `/usr/local/bin/`

the main.py is indeed namespaced in a folder 
```
/opt/katib/cmd/metricscollector/v1beta1/tfevent-metricscollector# ls
Dockerfile  Dockerfile.ppc64le	main.py  requirements.txt
```
in fact `/opt/katib/` seems to contain the whole katib git repo.

developer-guy:
we are continuing to this PR over [here](https://github.com/wolfi-dev/os/pull/4438)

PR #2217, opened 2023-05-22
add coredns

joshrwolf:
coredns: new package
#### For new package PRs only

### DIFF
--- a/coredns.yaml
+++ b/coredns.yaml
@@ -1,0 +1,55 @@
+package:
+  name: coredns
+  version: 1.10.1
+  epoch: 0
+  description: CoreDNS is a DNS server that chains plugins
+  target-architecture:
+    - all
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - ca-certificates-bundle
+
+environment:
+  contents:
+    packages:
+      - wolfi-baselayout
+      - build-base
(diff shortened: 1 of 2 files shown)


coredns.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+  - runs: |
+      # Ensures plugins get included
+      make check
+
+  - uses: go/build
+    with:
+      packages: .
+      output: coredns
+      ldflags: -s -w -X github.com/coredns/coredns/coremain.GitCommit=v${{package.version}}
```
dlorenc, line 36: Do we need the setcap bit, or does that come later in apko? https://github.com/coredns/coredns/blob/master/Dockerfile#L13

joshrwolf: that's a great question, I don't know... I don't even know if melange's solver will track/keep that? I also don't know if apko supports `setcap`ing.

do we have precedence for this anywhere?

dlorenc: No, I think this package was blocked on some recent work to make setcap work somehow so this will likely be the first to use it! cc @kaniini @deitch

kaniini: I think we need to update melange to use a newer version of the go-apk component first.

imjasonh (dismissed):

imjasonh (approved):

PR #2280, opened 2023-05-23
Add redis_exporter

ajayk:
#### For new package PRs only

### DIFF
--- a/packages.txt
+++ b/packages.txt
@@ -685,3 +685,4 @@ kaf
 dask-gateway
 http-echo
 src
+prometheus-redis-exporter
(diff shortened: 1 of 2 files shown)


redis_exporter.yaml
```diff
@@ -0,0 +1,39 @@
+package:
+  name: redis_exporter
```
kaniini, line 2: might it be helpful to call this `prometheus-redis_exporter`?

ajayk: the upstream has redis_exporter all over , just wanted to keep the same parity , happy to change it

prometheus-redis_exporter.yaml
```diff
@@ -0,0 +1,39 @@
+package:
+  name: prometheus-redis_exporter
```
imjasonh, line 2: Should we name this -exporter? I don't think apk cares, but just for consistency

ajayk: Addressed

imjasonh (approved):

PR #2281, opened 2023-05-24
fix policy violations in YAML files

kaniini:
These policy violations were found as a result of chainguard-dev/melange#465, which made it an explicit policy violation to use undefined keywords in the melange package definitions.

### DIFF
--- a/exim.yaml
+++ b/exim.yaml
@@ -27,7 +27,6 @@ environment:
       - build-base
       - ca-certificates-bundle
   accounts:
-    runs: exim
     users:
       - username: exim
         uid: 1001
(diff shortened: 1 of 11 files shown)


exim.yaml
```diff
@@ -27,7 +27,6 @@ environment:
       - build-base
       - ca-certificates-bundle
   accounts:
-    runs: exim
```
imjasonh, line 30: run-as?

kaniini: This setting is ignored by Melange.  No point in having it.

mattmoor (approved):

PR #2319, opened 2023-05-25
add ingress defaultbackend

cpanato:
add ingress defaultbackend

#### For new package PRs only

### DIFF
--- a/ingress-default-backend.yaml
+++ b/ingress-default-backend.yaml
@@ -1,0 +1,37 @@
+package:
+  name: ingress-default-backend
+  version: 1.23.1
+  epoch: 0
+  description: 404-server is a simple webserver that satisfies the ingress and serve pages 404 at root and 200 at /healthz
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - wolfi-baselayout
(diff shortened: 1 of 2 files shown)


ingress-default-backend.yaml
```diff
@@ -0,0 +1,44 @@
...
+  - uses: git-checkout
+    with:
+      repository: https://github.com/kubernetes/ingress-gce
+      tag: v${{package.version}}
+      expected-commit: e4201e49ca197573e2a319cbb9fbb595ded31a31
+
+  - runs: |
+      set -x
+      export CGO_ENABLED=0 GO111MODULE=on
+      go build \
```
rawlingsj, line 32: looks great - just one comment, it might be better to use the default melange `go/build` pipeline.  Here's an example https://github.com/wolfi-dev/os/blob/22607f91e0464d3e39322b2962a5dcf324d31a4e/cosign.yaml#L27-L30.  That will then run this https://github.com/chainguard-dev/melange/blob/main/pkg/build/pipelines/go/build.yaml#L62

cpanato: thank you <3

rawlingsj (approved):

PR #2321, opened 2023-05-25
Vault: Add entrypoint subpackage 

amouat:
Should allow us to create an image that is close to upstream and usable as Helm replacement.

Also made symlink of vault binary to /usr/bin/, so we can later remove it from /usr/sbin

### DIFF
--- a/vault.yaml
+++ b/vault.yaml
@@ -58,7 +58,7 @@ pipeline:
       # We can later delete the sbin one.
       mkdir "${{targets.destdir}}/usr/bin"
       chmod 755 "${{targets.destdir}}/usr/bin"
-      ln -s "${{targets.destdir}}/usr/sbin/vault" "${{targets.destdir}}/usr/bin/vault"
+      ln -s /usr/sbin/vault "${{targets.destdir}}/usr/bin/vault"
 
       install -m644 -D "./vault.hcl" "${{targets.destdir}}/etc/vault/vault.hcl"
 
--- a/vault.yaml
+++ b/vault.yaml
@@ -54,11 +54,30 @@ pipeline:
       install -m644 -D "./vault.confd" "${{targets.destdir}}/etc/conf.d/vault"
 
       install -m755 -D ./vault/bin/vault "${{targets.destdir}}/usr/sbin/vault"
+      # Really want to move the binary out of sbin. Let's have it in two locations temporarily
+      # We can later delete the sbin one.
+      mkdir "${{targets.destdir}}/usr/bin"
+      chmod 755 "${{targets.destdir}}/usr/bin"
+      ln -s "${{targets.destdir}}/usr/sbin/vault" "${{targets.destdir}}/usr/bin/vault"
+
       install -m644 -D "./vault.hcl" "${{targets.destdir}}/etc/vault/vault.hcl"
 
       install -m750 -d "${{targets.destdir}}/var/lib/vault"
 
   - uses: strip
+
+subpackages:
+  - name: "vault-entrypoint"
+    description: "Container entrypoint script for vault and required dependencies"
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}/usr/bin"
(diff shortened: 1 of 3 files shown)


amouat:
One moment, forgot to bump epoch and want to check the tarfile...

tuananh:
i think the reason i put it in sbin is because I ported it from alpine package.

https://git.alpinelinux.org/aports/tree/community/vault/APKBUILD#n99

amouat:
ahhh, that explains a few things! Thanks @tuananh !

amouat:
I think we can do the setcap now as well then, I'll move this back to draft 🤦 

Apologies for the back and forth if you're following along.

vault.yaml
```diff
@@ -53,12 +54,31 @@ pipeline:
       install -m644 -D "./vault.confd" "${{targets.destdir}}/etc/conf.d/vault"
 
       install -m755 -D ./vault/bin/vault "${{targets.destdir}}/usr/sbin/vault"
+      setcap cap_ipc_lock=+ep  "${{targets.destdir}}/usr/sbin/vault"
+      # Binary doesn't require root, so let's move it out of sbin - for a temporary period have it
+      # in both locations. We can later delete the sbin one.
+      mkdir "${{targets.destdir}}/usr/bin"
+      chmod 755 "${{targets.destdir}}/usr/bin"
+      ln -s /usr/sbin/vault "${{targets.destdir}}/usr/bin/vault"
```
imjasonh, line 62: Is it in /usr/sbin/vault for compatibility with an upstream Helm chart? If so we might want to make that symlinking part of a separate compat package.

amouat: The upstream Helm chart doesn't work at all with our current image. This PR adds the entrypoint package, which effectively adds Helm compatibility.

This change is separate though - I just wanted to move the binary out of sbin, as I don't think it should be there.

amouat:
I'm actually going to move this back to draft _again_.

The upstream entrypoint doesn't work very well with our package/image. I'm going to write a new entrypoint based on the upstream.

vault.yaml
imjasonh: I think we should just move it to `/usr/bin/vault`, if it sounds like we don't need to worry about breaking anybody since this was already not functional. WDYT?

amouat: It was functional, just not compatible with the helm chart or upstream image. There was actually another breakage where the config file wasn't being picked up.

amouat: I guess I'm being stupid though, that's literally a one char change to the apko that we can put through immediately.

amouat: Ok, created https://github.com/chainguard-images/images/pull/731 and will resolve this conversation.

amouat:
A few things to note about this PR:
 - the major addition is an entrypoint package to add upstream compatibility with the Helm chart
 - the vault binary was moved to /usr/bin from /usr/sbin, which requires an update to the apko
 - the config file was removed. Originally it wasn't being picked up due to a mistake in the melange config, and when it was picked up it conflicted with the entrypoint package config. The simplest solution seemed to be to remove it, as it wasn't being used.

vault.yaml
```diff
@@ -52,13 +53,37 @@ pipeline:
   - runs: |
       install -m644 -D "./vault.confd" "${{targets.destdir}}/etc/conf.d/vault"
 
-      install -m755 -D ./vault/bin/vault "${{targets.destdir}}/usr/sbin/vault"
-      install -m644 -D "./vault.hcl" "${{targets.destdir}}/etc/vault/vault.hcl"
+      install -m755 -D ./vault/bin/vault "${{targets.destdir}}/usr/bin/vault"
+      setcap cap_ipc_lock=+ep  "${{targets.destdir}}/usr/bin/vault"
```
imjasonh, line 57: Do we need to setcap after strip?

amouat: Yes, testing that change at the moment

amouat: Haven't actually got this working locally - not sure if it's because melange or apko are out of date?

amouat: I've pushed the update to check if it works here...

imjasonh (dismissed):

imjasonh (approved):

PR #2359, opened 2023-05-26
hugo: support extended build

marshall007:
#### For new package PRs only

### DIFF
--- a/hugo-extended.yaml
+++ b/hugo-extended.yaml
@@ -6,9 +6,8 @@ package:
   copyright:
     - license: Apache-2.0
   dependencies:
-    # required for extended SASS libraries
-    # - libc6-compat
-    - libstdc++
+    runtime:
+      - libstdc++
 
 environment:
   contents:
--- a/hugo-extended.yaml
+++ b/hugo-extended.yaml
@@ -1,6 +1,6 @@
 package:
   name: hugo-extended
-  version: 0.112.3
+  version: 0.112.5
   epoch: 0
   description: The world's fastest framework for building websites.
   copyright:
--- a/hugo-extended.yaml
+++ b/hugo-extended.yaml
@@ -1,10 +1,14 @@
 package:
-  name: hugo
+  name: hugo-extended
   version: 0.112.3
   epoch: 0
   description: The world's fastest framework for building websites.
   copyright:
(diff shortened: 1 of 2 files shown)


hugo-extended.yaml
```diff
@@ -0,0 +1,31 @@
+package:
+  name: hugo-extended
+  version: 0.112.3
+  epoch: 0
+  description: The world's fastest framework for building websites.
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    # required for extended SASS libraries
+    # - libc6-compat
```
marshall007, line 10: `libc6-compat` isn't available yet, may be required per: https://github.com/gohugoio/hugo/blob/3297b395d8d5fadb45b9055e55458967fb9f4a76/Dockerfile#L33

hugo-extended.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - busybox
+      - ca-certificates-bundle
+      - git
+
+pipeline:
+  - uses: go/install
+    with:
+      package: github.com/gohugoio/hugo
+      version: v${{package.version}}
+      tags: extended
```
marshall007, line 25: Is it possible to do the extended build in the same pipeline as `hugo.yaml`? Is it appropriate to use subpackages in this case?

johnfosborneiii (approved):

hugo-extended.yaml
johnfosborneiii: It could be a sub-package, since it's more of a build tag than a related package that's self-contained I'd say keep as two-separate packages. I noticed other Linux distros had one monolithic package for extended only, but I think two separate packages makes most sense here

johnfosborneiii (dismissed):

imjasonh (dismissed):

imjasonh:
The change looks good! Some notable lint findings:

```
missing hugo-extended.yaml in packages.txt
```

and 

```
package hugo-extended: update found newer version 0.112.5 compared with package.version 0.112.3 in melange config
```

johnfosborneiii (approved):

marshall007:
@imjasonh fixed, thanks. Can you approve the workflow execution again?

ajayk (approved):

PR #2364, opened 2023-05-26
add refactored calico package, remove old calicoctl configs

joshrwolf:
replaces `calicoctl` with a meta package for `calico`, with all required subpackages. this also adds the missing components from the previous `calicoctl` config file, namely it's core package: `node`

Related: #627


calico.yaml
```diff
@@ -0,0 +1,371 @@
...
+  - runs: |
+      go mod download
+  - working-directory: felix
+    pipeline:
+      # Equivalent to target: "build-bpf"
+      - runs: |
+          # TODO: this is a terrible hack to prevent the error:
+          #   /usr/include/gnu/stubs.h:7:11: fatal error: 'gnu/stubs-32.h' file not found
+          # I have no idea why it's wanting stubs-32.h and not correctly recognizing stubs-64.h instead
+          sed -i 's/#include <gnu\/stubs\.h>/\n/' /usr/include/features.h
```
joshrwolf, line 61: I'm not proud of this, and open to suggestions on how to do it the right way

calico.yaml
```diff
@@ -0,0 +1,371 @@
+package:
+  name: calico
+  version: 3.25.1
+  epoch: 3
```
mattmoor, line 4: FYI there's 3.26.0 now.  I just fixed up the automation and merged that for `calicoctl`, but we should update things here as well when you get a chance.

calico.yaml
```diff
@@ -0,0 +1,371 @@
...
+          CGO_ENABLED=1 \
+            CGO_LDFLAGS="$CGO_LDFLAGS" \
+            CGO_CFLAGS="$CGO_CFLAGS" \
+            go build -v -tags cgo -v -buildvcs=false \
+            -o "${{targets.subpkgdir}}"/usr/bin/test-workload \
+            ./felix/fv/test-workload/test-workload.go
+
+  - name: "calico-cni"
+    pipeline:
+      - uses: go/build
```
mattmoor, line 240: Can you do a pass and make sure all of the `go/build` steps have `ldflags: -s -w`? 🙏 

see also: https://github.com/wolfi-dev/os/pull/2377

calico.yaml
```diff
@@ -0,0 +1,371 @@
...
+      - bison
+      - autoconf
+      - automake
+      # TODO: Can remove this when cni-plugins are built from source
+      - curl
+
+# The calico build process is heavily tailored towards using Makefiles to
+# orchestrate docker commands. This makes for a less than ideal
+# experience replicating it in melange. To try to maintain some level of
+# sanity while packaging and updating calico, the approach here choses to
```
jonjohnsonjr, line 40:
```suggestion
# sanity while packaging and updating calico, the approach here chooses to
```

calico.yaml
```diff
@@ -0,0 +1,442 @@
...
+
+          mkdir -p bpf-gpl/include/libbpf/src/$_arch
+          export CFLAGS="-O2 -Wall -fomit-frame-pointer -march=x86-64-v2 -Wp,-D_FORTIFY_SOURCE=3"
+          CFLAGS="$CFLAGS" make -C bpf-gpl/include/libbpf/src BUILD_STATIC_ONLY=2 OBJDIR=$_arch
+  # This fetches and builds the statically compiled calico fork of BIRD
+  - working-directory: bird
+    pipeline:
+      - uses: fetch
+        with:
+          uri: https://github.com/projectcalico/bird/tarball/v0.3.3-202-g7a77fb73
```
imjasonh, line 81: Is this updated, and do we need to account for updating it to pick up new releases?

calico.yaml
```diff
@@ -0,0 +1,442 @@
...
+        - libnetfilter_conntrack
+        - iproute2
+        - procps
+        - conntrack-tools
+        - libnetfilter_cthelper
+        - libnetfilter_cttimeout
+        - libnetfilter_queue
+    pipeline:
+      # Equivalent to target: "$(NODE_CONTAINER_BINARY)"
+      - if: ${{build.arch}} == 'x86_64' || ${{build.arch}} == 'aarch64' # Essentially a guard to make sure we never try to build beyond x86_64 and aarch64
```
imjasonh, line 131: In the event that we add a third supported architecture, what will happen? Will this fail, or will it silently produce an invalid/empty package?

calico.yaml
kaniini: I think this is because its using the system headers when it should only be using the libbpf ones.

calico.yaml
```diff
@@ -0,0 +1,442 @@
...
+    pipeline:
+      # Equivalent to target: "build-bpf"
+      - runs: |
+          # TODO: this is a terrible hack to prevent the error:
+          #   /usr/include/gnu/stubs.h:7:11: fatal error: 'gnu/stubs-32.h' file not found
+          # I have no idea why it's wanting stubs-32.h and not correctly recognizing stubs-64.h instead
+          sed -i 's/#include <gnu\/stubs\.h>/\n/' /usr/include/features.h
+
+          # TODO: Alternatively, we could patch the unused variable out...
+          export CFLAGS="-O2 -Wall -fomit-frame-pointer -march=x86-64-v2 -Wp,-D_FORTIFY_SOURCE=3"
```
kaniini, line 66:
```suggestion
          export CFLAGS="$CFLAGS -Wno-error=unused-but-set-variable"
```

Though since this is BPF bytecode we probably don't really want the host CFLAGS anyway.

runit/run-service-dir.patch
```diff
@@ -0,0 +1,11 @@
+--- ./sv.c.orig
++++ ./sv.c
+@@ -32,7 +32,7 @@
+ char *progname;
+ char *action;
+ char *acts;
+-char *varservice ="/service/";
++char *varservice ="/etc/service/";
```
kaniini, line 8: why?  it's usually `/service` or `/sv` when using djb-style supervision trees.

calico.yaml
```diff
@@ -0,0 +1,442 @@
...
+          mkdir -p "${{targets.subpkgdir}}"/licenses
+          cp ./typha/LICENSE "${{targets.subpkgdir}}"/licenses
+
+  - name: "calico-typha-client"
+    description: "The calico typha client"
+    dependencies:
+      runtime:
+        - glibc
+    pipeline:
+      # TODO: I'm not sure yet if this actually needs CGO, or is just enabled upstream because of go-fips
```
kaniini, line 425: I would say CGO is a hard requirement for all Calico components tbh

calico.yaml
imjasonh: I'd be fine if it failed for that arch, since it would give us a hint that we need to do something in order to proceed. Silently producing an empty package is bad, but silently not producing a package could also lead to downstream confusion, if an image expects to use this package and be built for some third arch.

calico.yaml
kaniini: You might also try `-ffreestanding` option for CFLAGS here to avoid the stubs-32.h hack.

calico.yaml
kaniini: In that case, an assert + subpipeline is needed

runit/run-service-dir.patch
joshrwolf: good catch, I blindly took this from alpine, thinking `calico` also hardcoded it, but it looks like calico is setting it using `SVDIR`.

I removed this patch

calico.yaml
joshrwolf: I added an `assertions` guard on the arch specific subpackages. I think that, along with the `target-architecture` in the package config should be good enough 🤞

calico.yaml
joshrwolf: I followed the stack trace down to the `__x86_64__` macro not being set properly during our build, causing `/usr/include/gnu/stubs.h` to choose the `gnu/stubs-32.h` leg instead of the appropriate `x86_64` leg. to get around this I instead manually set `-D__x86_64__` as a compiler arg. this feels less hacky to me than removing the import `gnu/stubs` import entirely, but correct me if I'm wrong!

calico.yaml
joshrwolf: this still needs to be updated manually for now. in a future PR I'll make a `bird` package which we can leverage in `calico`. but as of today I don't know all the deltas between calico's fork.

since this changes out of band from calico proper (relevant [blame](https://github.com/projectcalico/calico/blame/e40707c5d30c353a78bed67a8c6596656b183121/metadata.mk#L26)), we hopefully have some time though

dlorenc:
Is this one almost ready?

calico.yaml
```diff
@@ -0,0 +1,446 @@
+package:
+  name: calico
+  version: 3.26.0
+  epoch: 2
```
joshrwolf, line 4: this is +1 of the old `calicoctl` package to avoid any conflicts with `calicoctl`

imjasonh (approved):

PR #2377, opened 2023-05-27
Pass `-ldflags '-s -w'` for all Go builds.

mattmoor:
These options strip debug and symbol information to get the smallest possible Go binaries.  Many of our binaries were already passing one or both of these, but this pass tries to add them basically everywhere else.


dlorenc (approved):

mattmoor:
impact on the `crane` binary size:
image


about 10% smaller

PR #2394, opened 2023-05-28
fix external-secrets package location and use strip

ajayk:
fix external-secrets package location and use strip


imjasonh (approved):
Thanks!

PR #2403, opened 2023-05-29
add kaniko

ajayk:
* "Kaniko: new package"

#### For new package PRs only

### DIFF
--- a/kaniko.yaml
+++ b/kaniko.yaml
@@ -36,6 +36,26 @@ subpackages:
           install -m755 -D out/warmer "${{targets.subpkgdir}}"/usr/bin/warmer
       - uses: strip
 
+  - name: kaniko-compat
+    pipeline:
+      - runs: |
+          # Symlink the binary from usr/bin to /kaniko/
+          mkdir -p ${{targets.subpkgdir}}/kaniko/
+          ln -sf /usr/bin/executor ${{targets.subpkgdir}}/kaniko/
+    dependencies:
+      runtime:
+        - kaniko
+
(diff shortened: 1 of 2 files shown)


kaniko.yaml
```diff
@@ -0,0 +1,45 @@
...
+  - uses: git-checkout
+    with:
+      repository: https://github.com/GoogleContainerTools/kaniko
+      tag: v${{package.version}}
+      expected-commit: 43d2eafcf210c6ff220e9cfad52bb674f6d27fd8
+
+  - runs: |
+      mkdir -p "${{targets.destdir}}"/usr/bin
+      make out/executor
+      install -m755 -D out/executor "${{targets.destdir}}"/usr/bin/executor
```
dlorenc, line 26: I think the file path is important here: https://github.com/GoogleContainerTools/kaniko/blob/83311d7a8f31928e315baae541d4cea4c4a926a2/deploy/Dockerfile#L67

Iirc it has to be /kaniko

ajayk: should that go into compat package ?

dlorenc (approved):

kaniko.yaml
imjasonh: This isn't a compat issue AIUI. Compat means "it should work anywhere, but helm chart assumes it's at <path>"

In kaniko's case *kaniko* expects to be run at that location, and anything else will result in failure.

There was work to make kaniko able to run from anywhere (so it could build itself) but it petered out and nobody maintains it anymore.

PR #2409, opened 2023-05-29
Add newrelic-infrastructure-bundle

Dentrax:
Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/newrelic-infrastructure-bundle.yaml
+++ b/newrelic-infrastructure-bundle.yaml
@@ -1,0 +1,56 @@
+package:
+  name: newrelic-infrastructure-bundle
+  version: 3.2.2
+  epoch: 0
+  description: New Relic Infrastructure containerised agent bundle
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - wolfi-baselayout
+      - busybox
+      - build-base
+      - go
+      - ca-certificates-bundle
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/newrelic/infrastructure-bundle
+      tag: v${{package.version}}
+      expected-commit: ece3dc47380c31d131a17d2b3b9f42f9811545fc
+      destination: ${{package.name}}
+
+  - working-directory: ${{package.name}}
+    pipeline:
+      - runs: |
+          go run downloader.go
+
+          case "${{build.arch}}" in
+              x86_64)   export TARGETARCH="amd64" ;;
+              aarch64)  export TARGETARCH="arm64" ;;
+              *)        echo "Unsupported architecture: ${{build.arch}}"; exit 1 ;;
+          esac
+
+          cd ./out/$TARGETARCH
+
+          mkdir -p ${{targets.destdir}}/var
+          mkdir -p ${{targets.destdir}}/etc
+          mkdir -p ${{targets.destdir}}/usr
+
+          mv ./var/db "${{targets.destdir}}"/var
+          mv ./etc/newrelic-infra "${{targets.destdir}}"/etc
+
+          mv ./usr/bin "${{targets.destdir}}"/usr
+          mv ./usr/share "${{targets.destdir}}"/usr
+      - uses: strip
+
+update:
+  enabled: true
+  github:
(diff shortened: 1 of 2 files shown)


newrelic-infrastructure-bundle.yaml
```diff
@@ -0,0 +1,56 @@
...
+      tag: v${{package.version}}
+      expected-commit: ece3dc47380c31d131a17d2b3b9f42f9811545fc
+      destination: ${{package.name}}
+
+  - working-directory: ${{package.name}}
+    pipeline:
+      - runs: |
+          go run downloader.go
+
+          case "${{build.arch}}" in
```
imjasonh, line 31: Can this be based on `go env GOARCH`? Then when we add a new arch we don't have to remember to add a case here.

newrelic-infrastructure-bundle.yaml
```diff
@@ -0,0 +1,56 @@
...
+    with:
+      repository: https://github.com/newrelic/infrastructure-bundle
+      tag: v${{package.version}}
+      expected-commit: ece3dc47380c31d131a17d2b3b9f42f9811545fc
+      destination: ${{package.name}}
+
+  - working-directory: ${{package.name}}
+    pipeline:
+      - runs: |
+          go run downloader.go
```
imjasonh, line 29: Just checking: Does this download the same contents every time it's run? I don't want to let a source of irreproducible builds to sneak in.

newrelic-infrastructure-bundle.yaml
Dentrax: Unfortunately, that `downloader.go` [does not allow to pass](https://github.com/newrelic/infrastructure-bundle/blob/ece3dc47380c31d131a17d2b3b9f42f9811545fc/downloader.go#L57-L63) custom arch and not respect the `GOARCH`. Force downloads the all archs: `amd64`, `arm64` and `arm`.

newrelic-infrastructure-bundle.yaml
Dentrax: Actually I was thinking the same thing. As long as the [bundle.yml](https://github.com/newrelic/infrastructure-bundle/blob/ece3dc47380c31d131a17d2b3b9f42f9811545fc/bundle.yml#L25-L87) doesn't change we can expect the same content. But the "trust" and reproduciblity is questionable.

But maybe we can bake that file in the repo to ensure that any changes to the upstream repository will not break reproducibility.

Another idea is to adopt that ~20+ package in Wolfi to create packages individually. However, it would increase the lead time as it will require a considerable of time to adopt in Wolfi. Wdyt?

dlorenc: Would it be possible to pin the downloaded contents by digest somehow?

Dentrax: I've just filed a proposal to implement digest pinning support in the `bundle.yaml` file: https://github.com/newrelic/infrastructure-bundle/issues/305

However, we don't have to wait for this request to be fulfilled. Here are some ideas:
1. We can manage a custom fork that include digest pinning support.
2. I can create a `.patch` file that adds digest pinning support to `downloader.go` file. Means that we have to keep `bundle.yaml` in Wolfi and manually upgrade all the versions and their digests.
3. Create all the packages (20+) in Wolfi. However, one drawback is, it would increase the lead time as it will require a considerable of time to adopt.
4. Merge this with known drawbacks and file a tracking issue to resolve this known issues later on.

Any thoughts?

imjasonh: The upstream issue you filed sounds like the best possible solution, let's see how that goes. Otherwise (4) sounds good for now. We don't want to maintain a fork, unless it's very short term so that we can fix the issue upstream. Forking the downloader to support this has the same drawbacks.

If there's no movement on the upstream issue we may have to do (3) and package each integration ourselves.

Thanks for looking into this @Dentrax!

dlorenc:
I think we can move forward with 4) for now, they seem to be interested in improving it but nothing imminent.

imjasonh (approved):

PR #2426, opened 2023-05-30
build-env: add additional CFLAGS/LDFLAGS from the OpenSSF hardening guide

kaniini:
Not all of the suggestions in the hardening guide are appropriate for distributions, either because they are case-specific, or the performance hit is too great, but we can enable some of them for Wolfi.

The flags enabled are:

- `-Wextra`: make the compiler warn about more things like integer conversion issues
- `-Wp,-D_GLIBCXX_ASSERTIONS`: like FORTIFY, but for C++ containers
- `-Wtrampolines`: warn when trampolines are generated, which cause text relocations
- `-z noexecstack`: require the stack be non-executable
- `-z noexecheap`: require the heap be non-executable


imjasonh (approved):
Nice.

dlorenc:
Nice!

PR #2446, opened 2023-05-31
Reintroduce Makefile rules caching

kaniini:
Solve regressions in generated makefile rules that were caused by differences between modern GNU make and the old version of GNU make included in macOS... 🙃


joshrwolf (dismissed):
worked for me with gnu `make 4.3`.

it still bombs on untracked files without explicitly setting `SOURCE_DATE_EPOCH`, but that should be fixed with a new build of melange with https://github.com/chainguard-dev/melange/pull/480

mattmoor (approved):

PR #2471, opened 2023-06-01
Add libapr : new package

ajayk:
#### For new package PRs only


libapr.yaml
```diff
@@ -0,0 +1,78 @@
...
+      expected-commit: 22867ddac49b7dcf356556577461231acdff6445
+
+  - uses: patch
+    with:
+      patches: apr-1.7.2-autoconf-2.72.patch
+
+  - runs: |
+      ./buildconf
+
+  - if: ${{build.arch}} == 'x86_64'
```
imjasonh, line 39: If we ever add a third arch, this will fail since configure hasn't run. Can we add another step that fails if neither of the previous steps ran, to get a clearer error? There's a precedent for go-stage0 I think.

libapr.yaml
```diff
@@ -0,0 +1,78 @@
...
+  - name: "libapr-dev"
+    description: "headers for libapr"
+    pipeline:
+      - uses: split/dev
+    dependencies:
+      runtime:
+        - libapr
+
+update:
+  enabled: false
```
imjasonh, line 75: Why enabled: false?

packages.txt
```diff
@@ -714,3 +714,5 @@ newrelic-prometheus-configurator
 prometheus-nats-exporter
 docker-credential-gcr
 kyverno
+alsa-lib
```
imjasonh, line 717: Where's this come from?

libapr.yaml
ajayk:
```could get the GraphQL to work with the tags here  [1.7.4-rc1-candidate](https://github.com/apache/apr/releases/tag/1.7.4-rc1-candidate)

libapr.yaml
ajayk: removed we don't need this anymore

imjasonh (approved):

PR #2501, opened 2023-06-02
dogfood our own gcloud image

imjasonh:
This has the usual potential bootstrapping issues, which we already have by dogfooding our CSI secrets-store images, and which has the same solution if it ever becomes a problem -- remove the tag and we'll use the official upstream gcloud image instead.

### DIFF
--- a/.github/workflows/dag-push-production.yaml
+++ b/.github/workflows/dag-push-production.yaml
@@ -136,6 +136,7 @@ jobs:
             --bucket=${BUCKET} \
             --src-bucket=${SRC_BUCKET} \
             --sdk-image ghcr.io/wolfi-dev/sdk:latest@sha256:ec70f835c3885d831ae1e5ba08425e0ec08ad2e55ed9b46f5dd5faff459e713c \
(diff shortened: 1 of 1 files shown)


ajayk (dismissed):

.github/workflows/dag-push-production.yaml
```diff
@@ -136,6 +136,7 @@ jobs:
             --bucket=${BUCKET} \
             --src-bucket=${SRC_BUCKET} \
             --sdk-image ghcr.io/wolfi-dev/sdk:latest@sha256:ec70f835c3885d831ae1e5ba08425e0ec08ad2e55ed9b46f5dd5faff459e713c \
+            --gcloud-image cgr.dev/chainguard/google-cloud-sdk:latest \
```
mattmoor, line 139: Can you pin this and let digestabot float it?  This gives us a rollback story when stuff breaks 🙏

imjasonh:
I'd still like to get @kaniini 's signoff on this before we merge please 🙏

kaniini (approved):

PR #2514, opened 2023-06-02
dogfood our CSI secrets-store images in this repo

imjasonh:
We're already dogfooding two of these images, the two most likely to break:

https://github.com/wolfi-dev/wolfictl/blob/0e133a790f792c2877c1d8c674d5473f516afffc/scripts/setup-cluster.sh#L27-L29
https://github.com/wolfi-dev/wolfictl/blob/0e133a790f792c2877c1d8c674d5473f516afffc/scripts/setup-cluster.sh#L37-L39

But, since we define them in wolfictl's setup-cluster.sh, we don't get the benefit of digestabot to pin and bump updates. Moving these dogfood configs into this repo will give us that benefit, and we can remove the change from setup-cluster.sh

As before, if these images break, we should comment them out and use the upstream to unbreak them, then reuse them once they work.

Once this merges we can drop the `kubectl set image`s from setup-cluster.sh (https://github.com/wolfi-dev/wolfictl/pull/242)

Supercedes https://github.com/wolfi-dev/wolfictl/pull/227/

### DIFF
--- a/.github/workflows/dag-push-production.yaml
+++ b/.github/workflows/dag-push-production.yaml
@@ -82,6 +82,31 @@ jobs:
       - working-directory: ${{github.workspace}}/wolfictl
         run: ./scripts/setup-cluster.sh ${SERVICE_ACCOUNT}
 
+      # Dogfood our own CSI secrets-store images.
+      # These can introduce a bootstrapping issue: if the images are broken,
+      # we won't be able to use them to build a fixed replacement.
+      # In that case, comment this out to use the upstream directly.
+      - run: |
+          kubectl set image daemonset/csi-secrets-store \
+            -n kube-system \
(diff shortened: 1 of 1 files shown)


mattmoor:
Haven’t looked at the change, but should we just make digestabot rewrite shell scripts too?-M(obile)On Jun 2, 2023, at 6:18 AM, Jason Hall ***@***.***> wrote:﻿
@imjasonh requested review from @wolfi-dev/wolfi-build-maintainers on: #2514 dogfood our CSI secrets-store images in this repo as a code owner.

imjasonh:
> Looks like Dockerfile and yaml? 
> 
> 
> 
> https://github.com/chainguard-dev/actions/blob/00f5af095173935ab7ab5a7931ef5c1d2e3c9b85/digesta-bot/action.yml#L75
> 
> 
> 
> cc @cpanato 

I think it's easier just to do this here in YAML and not have more places to have to look.

We could just move setup-cluster here too, honestly.

This reminds me I need to make this change in enterprise-packages too!

kaniini (approved):

PR #2527, opened 2023-06-02
don't use our own csi helper images, they need tests

imjasonh:
I'll make the same change in enterprise-packages too.


kaniini (approved):

imjasonh:
```
MountVolume.SetUp failed for volume "melange-key" : kubernetes.io/csi: mounter.SetUpAt failed to get CSI client: driver name secrets-store.csi.k8s.io not found in the list of registered CSI drivers
```

https://secrets-store-csi-driver.sigs.k8s.io/troubleshooting.html#volume-mount-fails-with-secrets-storecsik8sio-not-found-in-the-list-of-registered-csi-drivers

It sounds like the registrar isn't registraring

PR #2533, opened 2023-06-02
fix bazel 6 build

ajayk:
Regression introduced as apart of  https://github.com/chainguard-dev/melange/pull/486 , 
when extract is false deleted should be overridden or set to false


imjasonh (approved):

tuananh:
should this be fixed in melange fetch pipeline? 

that default doesn't make a lot of sense to me (do not extract and then delete right away)

PR #2552, opened 2023-06-03
Setcap for haproxy.

dlorenc:
The default configurations for haproxy can require it to listen on privileged ports.

Fixes:

Related:


dlorenc:
Looks like there's a new version so that check is failing, but IMO we should merge this first then bump.

jonjohnsonjr (approved):

imjasonh (approved):

PR #2740, opened 2023-06-12
don't BUILDWORLD unless asked

imjasonh:
This switches the default for `BUILDWORLD` from `yes` to `no`. This means to build a package you just need to:

```
make package/<package-name>
```


kaniini:
The builder workflows need to use `BUILDWORLD=yes`.

imjasonh:
> The builder workflows need to use `BUILDWORLD=yes`.

Hmm, that's going to be annoying for `wolfictl pod`. I'll drop this for now and maybe come back later.

PR #2747, opened 2023-06-12
add caddy package

joshrwolf:
adds `caddy` package

Related:

#### For new package PRs only

### DIFF
--- a/caddy.yaml
+++ b/caddy.yaml
@@ -1,0 +1,46 @@
+package:
+  name: caddy
+  version: 2.6.4
+  epoch: 0
+  description: Fast and extensible multi-platform HTTP/1-2-3 web server with automatic HTTPS
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - busybox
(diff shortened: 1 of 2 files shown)


dlorenc:
I looked into this awhile ago, is it worth using `xaddy`? https://github.com/caddyserver/caddy#with-version-information-andor-plugins

joshrwolf:
> I looked into this awhile ago, is it worth using `xaddy`? https://github.com/caddyserver/caddy#with-version-information-andor-plugins

I thought about it, but since the plugins all require a rebuild, there's still the problem of what plugins we package, and how we ship those packaged plugins...

how about for now we add a `xcaddy` (sub)package alongside this one? that way we at least have a usable default server (`caddy`), and users can install `xcaddy` if they wish to rebuild with any [plugins](https://caddyserver.com/download)

dlorenc (approved):

joshrwolf:
I'm going to land this as is for now and add `xcaddy` as a separate package for those users wanting to build their own caddy with our shipped `xcaddy`

PR #2751, opened 2023-06-12
use wolfictl from tools image

imjasonh:
This should make our builds (slightly) faster, and (slightly) more reliable since they don't use wolfictl from head.


ajayk (dismissed):

imjasonh:
I think this will need @kaniini's approval

PR #2788, opened 2023-06-13
Reduce CVE count and add prometheus patch for npd

Dentrax:
Add a prometheus patch to fix high CVEs

### DIFF
--- a/node-problem-detector.yaml
+++ b/node-problem-detector.yaml
@@ -1,13 +1,14 @@
 package:
   name: node-problem-detector
   version: 0.8.13
-  epoch: 1
+  epoch: 2
   description: node-problem-detector aims to make various node problems visible to the upstream layers in the cluster management stack.
   copyright:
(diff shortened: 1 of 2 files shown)


node-problem-detector.yaml
```diff
@@ -30,18 +31,28 @@ pipeline:
     with:
       patches: fix-archs.patch
 
+  - uses: patch
```
imjasonh, line 34:
```suggestion
  # This is needed to account for breaking changes in prometheus/client_golang
  - uses: patch
```

Correct?

This will help us know to remove the patch after we remove the `go get` bump.

Dentrax: Correct!

dlorenc (approved):

dlorenc:
You can send a patch upstream for this one too!

luhring:
@Dentrax let me know if you need help w/ the advisories PR!

PR #2801, opened 2023-06-14
Switch fluent-bit to build from git instead of the release URL.

dlorenc:
This allows us to turn on auto-updates.

Fixes:

Related:


dlorenc:
The update check should fail - we can rebuild here then let the updater bump us to 2.1.4

ajayk (approved):

PR #2853, opened 2023-06-15
add iputils

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/iputils.yaml
+++ b/iputils.yaml
@@ -1,0 +1,54 @@
+package:
+  name: iputils
+  version: 20221126
+  epoch: 0
+  description: IP Configuration Utilities
+  copyright:
+    - license: BSD-3-Clause AND GPL-2.0-or-later
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+      - automake
+      - autoconf
+      - libcap-dev
+      - meson
+
+pipeline:
+  - uses: fetch
+    with:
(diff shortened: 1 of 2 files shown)


dlorenc (approved):

iputils.yaml
```diff
@@ -0,0 +1,54 @@
+package:
+  name: iputils
+  version: '20221126'
```
dlorenc, line 3: Oh wait hmm... This might need to be 0_ prefixed.

dlorenc (dismissed):

iputils.yaml
dlorenc: Yeah I think we need to start this with 0.

https://github.com/wolfi-dev/os/blob/main/minio.yaml#L5

developer-guy: I just saw the argon.yaml which uses this syntax: _https://github.com/wolfi-dev/os/blob/main/argon2.yaml#L3_

So, should I fix also the argon package version?

developer-guy: I added '0.' prefix to the version.

developer-guy: this change caused an error like this: 
_https://github.com/wolfi-dev/os/actions/runs/5287579349/jobs/9568238586?pr=2853#step:5:10_

kaniini (approved):

kaniini:
I think we should use the date as the version, it is consistent with the upstream version.  We only do the `0.` thing when there are no real "releases."

kaniini (changes requested):
Please change the version to `20221126`.

kaniini (approved):

PR #2886, opened 2023-06-16
Revert "patch cve for acl" 

luhring:
Reverts #2854. In talking with @developer-guy, it doesn't look like we know of a CVE that this patch is addressing after all.


kaniini:
The patch should be kept, as it is related to the deprecation of LFS64 functions, but it does not fix a CVE.

kaniini (approved):

Issue #873, opened 2023-03-28
[Wolfi Package Request]: k3s


marcofranssen:
### What software would you like us to add to wolfi-os. Ideally include a URL to the project and its source.

https://k3s.io/

### which versions of the software should we include?

Starting with latest stable release should be good

### Add some justification for why this specific package and versions are important.

Being able to run k3s clusters for CI / test purposes.

Basically kubernetes in docker or kubernetes in kubernetes.

### Please check all that apply

- [X] This package has an un-restrictive license
- [X] The package/versions proposed are actively maintained upstream
- [ ] I am interested in adding this package to Wolfi OS myself
- [ ] I am willing to help maintain this package

PR #2959, opened 2023-06-20
add klipper-{lb,helm} packages

joshrwolf:
adds `klipper-{lb,helm}` packages.

This does not add the optionally dependent `helm-mapkubeapis` or `helm-set-status` plugins. These can/will be added in future packages, and pulled in via the `apko` config.

Related: #873 

#### For new package PRs only

### DIFF
--- a/klipper-helm.yaml
+++ b/klipper-helm.yaml
@@ -1,0 +1,36 @@
+package:
+  name: klipper-helm
+  version: 0.8.0-build20230510
+  epoch: 0
+  description: Klipper helm integration job image
+  copyright:
+    - license: Apache-2.0
+  dependencies:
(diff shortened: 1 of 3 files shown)


klipper-helm.yaml
```diff
@@ -0,0 +1,36 @@
+package:
+  name: klipper-helm
+  version: 0.8.0-build20230510
```
joshrwolf, line 3: cc @rawlingsj if you know any wizardry for getting updates working with this version schema...

rawlingsj: fwiw the lint that's failing can be recreated locally with
```
wolfictl lint klipper-helm.yaml
```

PR #2968, opened 2023-06-20
Revert "mongo-tools/100.7.2 package update" and withdraw 

rawlingsj:


dlorenc (approved):

rawlingsj:
Wolfictl Check Updates - this will fail as we are reverting to a previous working version

PR #3059, opened 2023-06-23
fix: bump apks with bad signatures and archives and withdraw those packages

ajayk:
```
/tmp/OpenSearch # apk add openjdk-11
WARNING: opening /github/workspace/packages: No such file or directory
(1/5) Installing java-cacerts (20230106-r1)
(2/5) Installing java-common (0.1-r0)
(3/5) Installing openjdk-11-jre-base (11.0.20.4-r1)
(4/5) Installing openjdk-11-jre (11.0.20.4-r1)
(5/5) Installing openjdk-11 (11.0.20.4-r1)
ERROR: openjdk-11-11.0.20.4-r1: BAD signature

```

### DIFF
--- a/cadvisor.yaml
+++ b/cadvisor.yaml
@@ -1,7 +1,7 @@
 package:
   name: cadvisor
   version: 0.47.2
-  epoch: 0
+  epoch: 1
   description: Analyzes resource usage and performance characteristics of running containers.
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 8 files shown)


dlorenc:
Do we have any idea where these keep coming from?

ajayk:
> Do we have any idea where these keep coming from?

one of the reasons could be when we sync to the buckets , I still need to dig down exactly when this started failing

luhring (approved):

PR #3069, opened 2023-06-23
add a meta kubernetes package with relevant subpackages

joshrwolf:
This consolidates the way we package components that are distributed via `kubernetes/kubernetes`. The idea here is for a central place to manage wolfi related things pertaining to `kubernetes/kubernetes`.

This means the addition of several `kube*` components ([here](https://github.com/wolfi-dev/os/pull/3069/files#diff-e2db87b5982b2194e3cae1eeafedf9e0a9246caec2b1dc4871c28dd892dbef1eR34)), and the deprecation of our existing `kubectl.yaml` package in favor of this one.

While this works, it significantly changes how we package an existing thing (`kubectl`), so I'm leaving this as draft/RFC for now.

#### For new package PRs only


amdawson:
heads up @mattmoor in case this has any impact on the kots build

kubernetes.yaml
```diff
@@ -0,0 +1,153 @@
...
+
+# "transform" the kubernetes version into the corresponding pause version, these don't always line up
+var-transforms:
+  - from: ${{package.version}}
+    match: ^(.+)$
+    replace: 3.9
+    to: pause-version
+
+vars:
+  components: "kubectl kubeadm kubelet kube-scheduler kube-proxy kube-controller-manager kube-apiserver"
```
imjasonh, line 34: Do we want this to be a `range` with subpackages that `make` each one, etc.? Or is the idea that this gives us a `kubernetes` package with allll the things, and subpackages for each individual thing if we want that.

Either way, +1 to this change.

joshrwolf: yeah +1 to your point. I started down the `range` route then fell down an overengeered templating rabbit hole.

I think while this is more verbose, it's easier to understand at first glance, and easier to maintain long term.

kubernetes.yaml
```diff
@@ -0,0 +1,153 @@
+package:
+  name: kubernetes
+  version: 1.27.3
+  epoch: 2
```
joshrwolf, line 4: start this at `2` so it doesn't squash the existing `kubectl-x-r1` we already have

imjasonh:
Unless there are any remaining concerns I'd like to push this forward, so we can add the other supported k8s versions (1.25, 1.26). It might be worth landing this as `kubernetes-1.27.yaml` (with `provides: kubernetes=1.27`) directly so I don't have to do it later and we have less churn.

kubernetes-1.25.yaml
```diff
@@ -0,0 +1,145 @@
...
+      make WHAT="$WHAT"
+
+  - runs: |
+      mkdir -p "${{targets.destdir}}"/usr/bin/
+      install -m755 _output/bin/* ${{targets.destdir}}/usr/bin/
+
+      mkdir -p "${{targets.destdir}}"/etc/kubernetes
+
+subpackages:
+  - name: kubectl
```
imjasonh, line 63: I think we want each of these to have a `provides: ['kubectl=1.25']` so that folks downstream don't have to care about the version to find the package.

joshrwolf: thanks, just caught this :D

joshrwolf: actually... do we edo this with a `provides: ['kubectl=1.15']` or with just renaming the subpackage with the variant? `name: kubectl-1.25`

joshrwolf: I went with the approach we use for `python` (both)

kubernetes-1.24.yaml
```diff
@@ -0,0 +1,164 @@
...
+      - findutils # needed for non busybox version of `xargs`
+      - rsync
+      - apk-tools
+
+# "transform" the kubernetes version into the corresponding pause version, these don't always line up
+var-transforms:
+  - from: ${{package.version}}
+    match: ^(.+)$
+    replace: 3.9
+    to: pause-version
```
imjasonh, line 31: This seems to be defined in each version and only used in 1.27?

kubernetes-1.24.yaml
```diff
@@ -0,0 +1,164 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/kubernetes/kubernetes
+      tag: v${{package.version}}
+      expected-commit: 2c67202dc0bb96a7a837cbfb8d72e1f34dfc2808
+
+  - runs: |
+      ./hack/pin-dependency.sh github.com/docker/distribution v2.8.2
```
imjasonh, line 44: Why do we pin this?

imjasonh (commented):
Lookin good, just a couple questions

imjasonh (approved):

kubernetes-1.24.yaml
imjasonh: This gets us past a vuln in distribution.

kubernetes-1.24.yaml
imjasonh: This is different per release.

PR #3081, opened 2023-06-23
new package: kots

imjasonh:


kots.yaml
```diff
@@ -0,0 +1,89 @@
+package:
+  name: kots
+  version: 1.100.3
+  epoch: 2
```
denhamparry, line 4: Should the `epoch` be 0?

denhamparry (dismissed):

kots.yaml
imjasonh: Yes indeed!

ajayk (approved):

dlorenc (approved):

PR #3102, opened 2023-06-24
add xmerl-1.3.32

developer-guy:
Signed-off-by: Batuhan Apaydin <batuhan.apaydin@chainguard.dev>

Related:
- https://github.com/chainguard-dev/image-requests/issues/102
- https://pkgs.alpinelinux.org/contents?file=&path=&name=erlang&branch=edge&repo=community&arch=x86

Screenshot 2023-06-24 at 19 45 08

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs


kaniini (changes requested):
The subpackage names need to describe their relationship to erlang.   Also the versions of the erlang applications should probably be omitted from the package name.

developer-guy:
> The subpackage names need to describe their relationship to erlang. 

So, we should add smth like "erlang-" prefix on front of them.

> Also the versions of the erlang applications should probably be omitted from the package name.

As we have erlang-25 (25.3.2.2-r0) and erlang-26 (26.0.1-r0) packages, so, I believe, they have both different version of the sub-packages, for example, for erlang-26 the sub-package xmerl* version is 1.3.2 but for erlang-25, it is 1.3.31.1, so, in that situtation, what should we do if we shouldn't provide the version information in the sub-package name? Thanks

kaniini:
We should use erlang-26-xmerl etc.

imjasonh (commented):
This change lgtm but I'd like to wait for @kaniini's approval since she had comments.

erlang-26.yaml
```diff
@@ -64,6 +64,118 @@ subpackages:
       provides:
         - erlang-dev=26.999.0
 
+  - name: "erlang-26-xmerl"
```
kaniini, line 67: Why not use range here?

kaniini (dismissed):
Please flatten and use range instead of repeating the same block over and over again.

erlang-26.yaml
developer-guy: great suggestion!

imjasonh (approved):

PR #3115, opened 2023-06-25
add ghostscript & openjpeg

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/ghostscript.yaml
+++ b/ghostscript.yaml
@@ -1,0 +1,128 @@
+package:
+  name: ghostscript
+  version: 10.01.2
+  epoch: 0
+  description: Interpreter for the PostScript language and for PDF
+  copyright:
+    - license: AGPL-3.0-or-later
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+      - autoconf
+      - automake
+      - cups-dev
+      - expat-dev
+      - freetype-dev
+      - gtk-2.0-dev
+      - jbig2dec-dev
+      - lcms2-dev
+      - libjpeg-turbo
+      - libpng-dev
+      - libtool
+      - openjpeg-dev
+      - tiff-dev
+      - zlib-dev
+
+var-transforms:
+  - from: ${{package.version}}
+    match: \.
+    replace: ''
+    to: mangled-package-version
+
+pipeline:
+  - uses: fetch
+    with:
+      expected-sha512: 11887850d440b8bbd8b83b8b58df8c743cf94be8fdbcfbda1a5fce55ce7c6a963d94b41a4ebcf1ba48f64f8af01e76dec789e2711212b1f7a32f98c42d75e0ff
+      uri: https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs${{vars.mangled-package-version}}/ghostscript-${{package.version}}.tar.gz
+
+  - uses: patch
+    with:
+      patches: fix-sprintf.patch
+
+  - uses: patch
+    with:
+      patches: ghostscript-system-zlib.patch
+
+  - runs: |
+      rm -r jpeg libpng zlib tiff lcms2mt cups/libs jbig2dec \
+      freetype
+
+      # fix parallel builds
+      sed -i -e 's/ECHO_XE/ECHOGS_XE/g' \
+      	-e 's/^\($(GLOBJ)md5.$(OBJ) :.*\)/\1 $(ECHOGS_XE)/' \
+      	base/lib.mak
+      aclocal && autoconf --force
+
+      cd ./ijs
+      libtoolize --force && aclocal && autoconf && automake --add-missing
+
+  - uses: autoconf/configure
+    with:
+      dir: ijs
+      opts: |
+        CFLAGS="$CFLAGS -flto=auto" \
+        --enable-shared \
+        --disable-static
+
+  - uses: autoconf/make
+    with:
+      dir: ijs
+
+  - uses: autoconf/make-install
+    with:
+      dir: ijs
+
+  - uses: autoconf/configure
+    with:
+      dir: .
+      opts: |
+        --docdir=/usr/share/doc/ghostscript \
+        --with-system-libtiff \
+        --with-ijs \
+        --with-jbig2dec \
+        --without-libpaper \
+        --without-versioned-path \
+        --enable-gtk \
+        --with-drivers=ALL \
+        --with-fontpath=/usr/share/fonts/Type1:/usr/share/fonts \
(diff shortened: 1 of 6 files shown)


imjasonh (dismissed):
The change lgtm but I think the update check is telling us wolfibot won't be able to update it for us:

```
time="2023-06-25T17:37:34Z" level=fatal msg="error during command execution: 1 error occurred:\n\t* package ghostscript: failed to validate update config, melange bump: got 404 Not Found when fetching [https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs${{vars.mangled-package-version}}/ghostscript-10.01.2.tar.gz\n\n](https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs$%7B%7Bvars.mangled-package-version%7D%7D/ghostscript-10.01.2.tar.gz/n/n)"
```

developer-guy:
> The change lgtm but I think the update check is telling us wolfibot won't be able to update it for us:
> 
> ```
> time="2023-06-25T17:37:34Z" level=fatal msg="error during command execution: 1 error occurred:\n\t* package ghostscript: failed to validate update config, melange bump: got 404 Not Found when fetching [https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs${{vars.mangled-package-version}}/ghostscript-10.01.2.tar.gz\n\n](https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs$%7B%7Bvars.mangled-package-version%7D%7D/ghostscript-10.01.2.tar.gz/n/n)"
> ```

is this some sort of a bug in wolfictl lint command? Because it seems that it couldn't replace this part `${{vars.mangled-package-version}}`🤔 PTAL @Dentrax

rawlingsj:
The cause of the issue is the var substitution isn’t taking place during a melange bump.  The reason for that is the way the melange packages are structured as a cyclic package dependency occurs when melange bump calls the var replace code.

Until this is resolved we've been merging with that failed check as it isn’t a “required” check.

What you can do is change the update section to `manual: true`

ghostscript.yaml
```diff
@@ -0,0 +1,135 @@
...
+  - name: ghostscript-dev
+    pipeline:
+      - uses: split/dev
+    description: ghostscript dev
+    dependencies:
+      runtime:
+        - ghostscript
+
+update:
+  enabled: true
```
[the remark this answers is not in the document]
developer-guy: it seems that adding `manual: true` didn't fix the issue :/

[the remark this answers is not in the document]
rawlingsj: yeah it wont fix the issue, it just means the wolfi bot wont attempt to do the update, instead an issue will be created so a manual update can occur.  

the failed check is not a required check so we can ignore it

ghostscript.yaml
```diff
@@ -0,0 +1,136 @@
...
+
+  - runs: |
+      export CFLAGS="$CFLAGS -flto=auto"
+
+      # build ijs
+      cd ./ijs
+
+      ./configure \
+      	--build=$CBUILD \
+      	--host=$CHOST \
```
rawlingsj, line 71: I think these need to be 
```
        --host=${{inputs.host}} \
        --build=${{inputs.build}} \
```
i.e. https://github.com/chainguard-dev/melange/blob/58d5b8b3a565bb012ff37fde1c38e4b359e79c96/pkg/build/pipelines/autoconf/configure.yaml#L23C1-L24C36

I wonder if we should us the reusable autoconf pipeline instead?

i.e. like this https://github.com/wolfi-dev/os/blob/d46144988d2bb6f40c37a603285d2935f01c2941/tomcat-native.yaml#L44-L55

and for the others below?

developer-guy: thanks @rawlingsj, fixed.

ghostscript.yaml
```diff
@@ -0,0 +1,128 @@
...
+      - cups-dev
+      - expat-dev
+      - freetype-dev
+      - gtk-2.0-dev
+      - jbig2dec-dev
+      - lcms2-dev
+      - libjpeg-turbo
+      - libpng-dev
+      - libtool
+      # - openjpeg-dev
```
rawlingsj, line 26: Is openjpeg needed?  I noticed that we're adding the melange config below but `openjpeg-dev` is commented out above.

ghostscript/fix-sprintf.patch
```diff
@@ -0,0 +1,17 @@
+diff --git a/cups/gdevcups.c b/cups/gdevcups.c
```
rawlingsj, line 1: Could we add a comment at the top please, of where the original patch comes from?

ghostscript/ghostscript-system-zlib.patch
```diff
@@ -0,0 +1,14 @@
+diff --git a/base/lib.mak b/base/lib.mak
```
rawlingsj, line 1: Same as above, can we add a comment at the top please, of where the original patch comes from?

openjpeg/fix-cmakelists.patch
```diff
@@ -0,0 +1,11 @@
+--- a/CMakeLists.txt
```
rawlingsj, line 1: Another comment needed top where the patch comes from please.

rawlingsj (changes requested):

ghostscript.yaml
developer-guy: yeah, I'll send a complementary fix PR for adding openjpeg-dev afterwards we merge this PR since I'm adding packages in single PR but I can separate them but this time I need to wait openjpeg to be built first, then I can add ghostscript package, WDYT?

ghostscript/fix-sprintf.patch
developer-guy: of course 🙉

ghostscript/ghostscript-system-zlib.patch
developer-guy: of course 🙉

openjpeg/fix-cmakelists.patch
developer-guy: of course 🙉

ghostscript.yaml
rawlingsj: As long as openjpeg comes before ghostscript in the `packages.txt` which it does, it should be fine to uncomment the dependency above and we can make sure CI passed?

developer-guy: great, lets give it a try!

developer-guy: it worked ✅

rawlingsj (approved):

PR #3172, opened 2023-06-28
add gsm

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/gsm.yaml
+++ b/gsm.yaml
@@ -1,0 +1,74 @@
+package:
+  name: gsm
+  version: 1.0.22
+  epoch: 0
+  description: Shared libraries for GSM speech compressor
+  copyright:
+    - license: TU-Berlin-2.0
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+      - automake
+      - autoconf
+
+pipeline:
+  - uses: fetch
+    with:
+      expected-sha512: f4284f096ed4d606ddece5c1b28edb63feacc26a0b664f1350c040dbf50113d27bedfd9896bf1620cb4fc3a6c8f6a5244955a6c625ded833dcee607c63c97dd2
+      uri: https://www.quut.com/gsm/gsm-${{package.version}}.tar.gz
+      strip-components: 1
+
+  - uses: patch
+    with:
+      patches: gsm-64bit.patch
+
+  - uses: patch
+    with:
+      patches: gsm-makefile.patch
+
+  - runs: |
+      export RPM_OPT_FLAGS="$CFLAGS -fPIC"
+      make all
+
+  - runs: |
+      mkdir -p "${{targets.destdir}}"/usr/bin \
+      	"${{targets.destdir}}"/usr/include/gsm \
+      	"${{targets.destdir}}"/usr/lib \
+      	"${{targets.destdir}}"/usr/share/man/man1 \
+      	"${{targets.destdir}}"/usr/share/man/man3
+
+      make install \
+      	INSTALL_ROOT="${{targets.destdir}}"/usr \
+      	GSM_INSTALL_INC="${{targets.destdir}}"/usr/include/gsm \
+      	GSM_INSTALL_LIB="${{targets.destdir}}"/usr/lib
+
+      ls -latr ./lib
+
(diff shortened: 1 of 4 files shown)


gsm.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+  - uses: strip
+
+subpackages:
+  - name: "gsm-tools"
+    description: "GSM speech compressor tools"
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/usr
+          mv "${{targets.destdir}}"/usr/bin/* "${{targets.subpkgdir}}"/usr/
```
imjasonh, line 69: This doesn't look right. Won't this put everything in `/usr/bin` into the subpackage's `/usr/`?

gsm.yaml
```diff
@@ -0,0 +1,74 @@
...
+      uri: https://www.quut.com/gsm/gsm-${{package.version}}.tar.gz
+      strip-components: 1
+
+  - uses: patch
+    with:
+      patches: gsm-64bit.patch
+
+  - uses: patch
+    with:
+      patches: gsm-makefile.patch
```
imjasonh, line 31: Where do these patches come from? I'd love to understand what they're doing a bit more, or at least have a pointer if things go wrong.

gsm.yaml
kaniini: Yeah, it should be

```suggestion
          mv "${{targets.destdir}}"/usr/bin "${{targets.subpkgdir}}"/usr/
```

gsm/gsm-makefile.patch
```diff
@@ -0,0 +1,126 @@
...
++
+ # Choose a compiler.  The code works both with ANSI and K&R-C.
+ # Use -DNeedFunctionPrototypes to compile with, -UNeedFunctionPrototypes to
+ # compile without, function prototypes in the header files.
+@@ -44,7 +48,7 @@ WAV49	= -DWAV49
+ # CCFLAGS 	= -c -O
+ 
+ CC		= gcc -ansi -pedantic
+-CCFLAGS 	= -c -O2 -DNeedFunctionPrototypes=1 -Wall -Wno-comment
++CCFLAGS 	= -c $(RPM_OPT_FLAGS) -D_REENTRANT -DNeedFunctionPrototypes=1
```
kaniini, line 21: This seems wrong.  There will be no `RPM_OPT_FLAGS` in our environment.

gsm.yaml
developer-guy: https://git.alpinelinux.org/aports/tree/main/gsm

gsm/gsm-makefile.patch
developer-guy: I took them from here: https://git.alpinelinux.org/aports/tree/main/gsm

imjasonh: I think Ariadne's point stands. We need to understand what this patch does, and if it does nothing, we don't need it.

kaniini: The patch is necessary, it makes the build system more friendly for packaging purposes, but this needs to be documented.  I also think the `RPM_OPT_FLAGS` should be renamed to something more generic.

gsm/gsm-64bit.patch
```diff
@@ -0,0 +1,15 @@
+--- gsm-1.0-pl12/inc/private.h.64bit	2006-04-26 21:17:01.000000000 +0200
++++ gsm-1.0-pl12/inc/private.h	2007-05-11 00:25:23.000000000 +0200
+@@ -10,10 +10,10 @@
+ #define	PRIVATE_H
+ 
+ typedef short			word;		/* 16 bit signed int	*/
```
kaniini, line 6: I think it would be better to refactor this patch in terms of using `stdint.h` types like `int16_t` and `uint16_t` and so on.  This makes the patch more robust and stops encoding toolchain-specific assumptions into the types.

developer-guy: I tried to do that refactoring on patch but the CI failed :/

> _https://github.com/wolfi-dev/os/actions/runs/5413419072/jobs/9838942778?pr=3172#step:7:288_

imjasonh (approved):
Merging this to unblock, but we need to come back to this and explain/fix the patches next week.

PR #3179, opened 2023-06-28
add cdparanoia

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/cdparanoia/fix-includes.patch
+++ b/cdparanoia/fix-includes.patch
@@ -1,0 +1,7 @@
+--- cdparanoia-III-10.2.orig/utils.h
++++ cdparanoia-III-10.2/utils.h
+@@ -1,3 +1,4 @@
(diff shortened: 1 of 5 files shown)


cdparanoia/fix-includes.patch
```diff
@@ -0,0 +1,7 @@
+--- cdparanoia-III-10.2.orig/utils.h
++++ cdparanoia-III-10.2/utils.h
```
imjasonh, line 2: Are these also borrowed from alpine? Do we know if we need them, and what they did for alpine?

kaniini: In this case, all of these patches are needed to make it compile cleanly under modern GCC.  But they need to be documented.

imjasonh (approved):

PR #3186, opened 2023-06-28
add libtheora

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs


libtheora/automake.patch
```diff
@@ -0,0 +1,11 @@
+--- ./configure.ac.orig	2012-12-31 20:15:29.384600257 +0000
++++ ./configure.ac	2012-12-31 20:15:38.461347599 +0000
+@@ -9,7 +9,7 @@
+ AC_CANONICAL_HOST
+ AC_CANONICAL_TARGET
+ 
+-AM_CONFIG_HEADER([config.h])
++AC_CONFIG_HEADER([config.h])
```
imjasonh, line 8: I'd love to have a better understanding of these patches in general. Where do they come from? What do they do? When should we feel confident enough to remove them?

kaniini: This one changes the use of an erroneous macro which was removed in Automake 1.15: https://www.gnu.org/software/automake/manual/1.12/html_node/Obsolete-Macros.html

dlorenc (approved):

PR #3211, opened 2023-06-29
Compiling PHP with shared extensions

erikaheidi:
The current php build in Wolfi is very simple, with just a few built-in, common extensions. This makes it hard to use Wolfi-based PHP images as replacement for Alpine and Debian PHP images.

This PR improves the PHP build to include several extensions as shared libraries that will live in subpackages such as `php-curl`, `php-openssl`, etc.

With this new build in place it will be easier to add more extensions when dependencies are met (for instance, the `php-zip` package depends on [libzip](https://libzip.org/download/) which is not yet available in Wolfi).

Full list of extensions now available as subpackages `php-${EXTENSION}`:

- bz2: Bzip2
- curl: cURL
- gd: GD imaging
- gmp: GNU GMP support
- ldap: LDAP
- mysqlnd: MySQLnd
- openssl: OpenSSL
- pdo_mysql: MySQL driver for PDO
- pdo_sqlite: SQLite 3.x driver for PDO
- soap: SOAP
- sodium: Sodium
- calendar: Calendar
- exif: EXIF
- gettext: GetText
- intl: Internationalization
- mbstring: Multibyte String Functions
- opcache: Opcache
- pcntl: pcntl
- pdo: PHP Data Objects
- sockets: Sockets
- xsl: XSL
- bcmath: BC Math
- ctype: ctype
- iconv: Iconv
- dom: DOM
- pgsql: PostgreSQL
- posix: Posix
- simplexml: SimpleXML
- mysqli: MySQLi
- xmlreader: XMLReader
- xmlwriter: XMLWriter
- fileinfo: fileinfo

This build was largely based on the [Alpine build](https://git.alpinelinux.org/aports/tree/community/php82/APKBUILD) with the goal of creating drop-in replacements to Alpine PHP images.

## Paths Change
Bin paths changed slightly so I created an alias to not break current images that set `/bin/php` as entry point.


php.yaml
```diff
@@ -12,43 +12,178 @@ package:
...
+            --with-avif \
+            --with-freetype \
+            --with-jpeg \
+            --with-webp \
+            --with-xpm \
+            --disable-gd-jis-conv \
+        --with-libxml \
+        --enable-phar=shared \
+        --enable-mbstring=shared \
+      	--with-mysqli=shared \
```
imjasonh, line 90: Weird indenting here and throughout

imjasonh (approved):

php.yaml
erikaheidi: It's because some of these `with-` options are linked to a parent option (such as `with-pdo-X` are dependent on `--enable-pdo`) so this is an attempt to keep the related options together (as in the [Alpine build](https://git.alpinelinux.org/aports/tree/community/php82/APKBUILD#n244)).

imjasonh (approved):

PR #3243, opened 2023-06-30
Fix guacd telnet and vnc plugins.

dlorenc:
Fixes:

Related:

#### For new package PRs only

- [ ] This PR is marked as fixing a pre-existing package request bug
  - [ ] Alternatively, the PR is marked as related to a pre-existing package request bug, such as a dependency
- [ ] REQUIRED - The package is available under an OSI-approved or FSF-approved license
- [ ] REQUIRED - The version of the package is still receiving security updates
- [ ] REQUIRED - The package is added to `packages.txt`

#### For security-related PRs

- [ ] The security fix is recorded in `advisories` and `secfixes`

#### For version bump PRs

- [ ] The `epoch` field is reset to 0
- [ ] Patch source: _patch source here_

PR #3244, opened 2023-06-30
guacamole-server: refactor plugin subpackages, delete unusable static libraries

kaniini:


guacamole-server.yaml
```diff
@@ -56,30 +56,38 @@ pipeline:
...
   - uses: strip
 
+  - runs: |
+      find ${{targets.destdir}}/usr/lib -name '*.a' -type f -exec rm -f '{}' +
+
+data:
+  - name: plugins
+    items:
+      vnc: VNC
+      telnet: Telnet
```
developer-guy, line 66: maybe we can add `ssh` here as the @dlorenc did in his PR: https://github.com/wolfi-dev/os/pull/3243/files

developer-guy: then I can add rdp here as well

PR #3255, opened 2023-06-30
set concurrency on x86 - push to production

ajayk:


.github/workflows/push-production.yaml
```diff
@@ -6,6 +6,11 @@ on:
     paths-ignore: ['README.md']
   workflow_dispatch:
 
+concurrency:
+  group: x86-${{ github.ref }}
+  # cancel-in-progress: true
+
+
 jobs:
```
imjasonh, line 14:
```suggestion
# Only run one build at a time, to prevent bad signature errors.
concurrency:
  group: x86-${{ github.ref }}

jobs:
```

imjasonh (approved):

PR #3274, opened 2023-06-30
deconflict kubernetes, and use a kubernetes-$v-default package choose

joshrwolf:
Fixes: allows multiple kubernetes versions to be installed on the same system without conflicting

### DIFF
--- a/kubernetes-1.24.yaml
+++ b/kubernetes-1.24.yaml
@@ -1,7 +1,7 @@
 package:
   name: kubernetes-1.24
   version: 1.24.15
-  epoch: 1
+  epoch: 2
   description: Production-Grade Container Scheduling and Management
   copyright:
     - license: Apache-2.0
@@ -69,7 +69,7 @@ subpackages:
     pipeline:
       - runs: |
           mkdir -p ${{targets.subpkgdir}}/usr/bin
-          install -m755 _output/bin/kubectl ${{targets.subpkgdir}}/usr/bin/kubectl
+          install -m755 _output/bin/kubectl ${{targets.subpkgdir}}/usr/bin/kubectl-1.24
 
           mkdir -p "${{targets.subpkgdir}}"/usr/share/bash-completion/completions
           _output/bin/kubectl completion bash > "${{targets.subpkgdir}}"/usr/share/bash-completion/completions/kubectl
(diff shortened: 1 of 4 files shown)


kubernetes-1.24.yaml
```diff
@@ -167,7 +167,27 @@ subpackages:
         runs: |
           mkdir -p ${{targets.subpkgdir}}/usr/bin/
           CFLAGS="$CFLAGS -static -DVERSION=${{vars.pause-version}}-${{package.version}}"
-          gcc -o "${{targets.subpkgdir}}"/usr/bin/pause linux/pause.c
+          gcc -o "${{targets.subpkgdir}}"/usr/bin/pause-3.7 linux/pause.c
+
+  - name: kubernetes-1.24-default
```
joshrwolf, line 172: the `pause` binary is intentionally left out of the `-default` subpackage since it's not pinned to any specific kubernetes version, and typically only installed standalone

kubernetes-1.24.yaml
```diff
@@ -167,7 +167,27 @@ subpackages:
...
+    description: "Compatibility package to set 1.24 as the default kubernetes, and add packages to their shortened path"
+    dependencies:
+      runtime:
+        - kubectl=1.24
+        - kubelet=1.24
+        - kube-scheduler=1.24
+        - kube-proxy=1.24
+        - kube-controrller-manager=1.24
+        - kube-apiserver=1.24
```
joshrwolf, line 181:
```suggestion
```

I'm second guessing myself here, I think these should be empty, so that things like:


```yaml
environment:
  contents:
    packages:
      - kubectl-1.24
      - kubernetes-1.24-default
```

can give the desired effect of defaulting `kubectl`, without pulling in all the other components?

kaniini (dismissed):

kubernetes-1.24.yaml
kaniini: Yeah I think I agree with that.

denhamparry:
Jumping in on this PR to track it. When the PR is merged, I'll look at updating the `kots` image to make sure the binaries are where Replicated wants them. This doesn't impact this PR from merging. Thanks for your work on this 🐺 🙌 :raised_paws:

rawlingsj:
@joshrwolf is there anything left to do on this PR or can we merge?

rawlingsj (approved):

PR #3279, opened 2023-06-30
calico: add symlink to calico-cni-compat

imjasonh:
The upstream `calico/cni` image has another copy of `calico-cni` available at `/opt/cni/bin/calico`, which gets called by the calico-node daemonset.

https://oci.dag.dev/layers/calico/cni@sha256:eb518dae10e9969596f9fda4f68af32f9bd6a66f95aad810054540ebd955a09c/opt/cni/bin/

### DIFF
--- a/calico.yaml
+++ b/calico.yaml
@@ -1,7 +1,7 @@
 package:
   name: calico
   version: 3.26.1
-  epoch: 1
+  epoch: 2
   description: "Cloud native networking and network security"
   target-architecture:
     - x86_64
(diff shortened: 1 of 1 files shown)


rawlingsj:
You might want an epoch bump?

imjasonh:
This might not be necessary, holding.

imjasonh:
Nevermind I think this is needed, bumped epoch. Thanks for catching that!

rawlingsj (approved):

PR #3307, opened 2023-07-02
graphviz switch libjpeg-dev to libjpeg-turbo-dev

deitch:
Fixes the graphviz dependency on `libjpeg-dev` to be our `libjpeg-turbo-dev`. Follow-up to #3302 and #3131

### DIFF
--- a/graphviz.yaml
+++ b/graphviz.yaml
@@ -20,7 +20,7 @@ environment:
       - freetype-dev
       - gd-dev
       - gmp-dev
-      - libjpeg-dev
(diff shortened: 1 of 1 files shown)


deitch:
This is breaking because (thanks to @dlorenc for pointing it out), `pango` includes entire `gtk`, but with different file contents, so by the time it installs `gtk`, it sees a conflict and fails.

ajayk (approved):

ajayk (changes requested):
Epoch needs to be bumped

PR #3310, opened 2023-07-02
feat: add corepack

rawkode:
In NodeJS land, people often use alternative package managers, with yarn and pnpm being the leading two.

Both of these can actually be enabled/configured via corepack, which is an experimental from the NodeJS project: it's a CLI that shims/proxies calls.

I believe this should be added to the nodejs images, so this was step 1 :)

### Pre-review Checklist

### DIFF
--- a/corepack.yaml
+++ b/corepack.yaml
@@ -1,0 +1,40 @@
+package:
+  name: corepack
+  version: 0.19.0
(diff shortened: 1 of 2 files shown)


dlorenc:
Nice!

imjasonh (approved):

PR #3311, opened 2023-07-02
Re-add gtk-3 package.

dlorenc:
This was missing a few deps which caused it to ignore pango, and then use it's own version of pango. This led to a conflict with pango, which broke builds that depended on both.

Fixes:

Related:

### DIFF
--- a/gtk-3.yaml
+++ b/gtk-3.yaml
@@ -1,0 +1,129 @@
+package:
+  name: gtk-3
+  version: 3.24.38
+  epoch: 0
+  description: The GTK+ Toolkit (v3)
+  copyright:
+    - license: LGPL-2.1-or-later
+
+# creates a new var that contains only the major and minor version to be used in the fetch URL
+# e.g. 2.46.0 will create a new var mangled-package-version=2.46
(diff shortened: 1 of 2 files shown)


dlorenc:
I also renamed the package from gtk-3.0 to gtk-3, the 3.0 part seemed redundant (and wrong) since we're actually using 3.24

dlorenc:
The update check error is expected

gtk-3.yaml
```diff
@@ -0,0 +1,129 @@
...
+      # but throw in a test here in case.
+      if [ -f ${{targets.destdir}}/usr/lib/libpango-1.0.so ]; then
+        echo "ERROR: libpango-1.0.so found in /usr/lib. This is a bug in the build system."
+        exit 1
+      fi
+
+  - uses: strip
+
+subpackages:
+  - name: gtk-3.0-dev
```
luhring, line 99: Should the subpackages be changed to `-3-` as well?

dlorenc: Good catch!

luhring (approved):

PR #3314, opened 2023-07-02
fix: ensure corepack cli available on path

rawkode:
First, the wildcard expansion didn't work on the original link command (pipelines not executed in a shell?), but we actually only need the corepack file anyway.

Secondly, we now link directly to corepack.js instead of the shim; due to the shim trying to require (relative path) the dist file. The shims seem to be more of an inconvenience than convenience, so bypassing altogether. 

I've tested this by `apk add --update corepack` in `wolfi-base` and I run the `ln` command myself. Everything works great now.

Fixes:

Related:

#### For new package PRs only

#### For version bump PRs

### DIFF
--- a/corepack.yaml
+++ b/corepack.yaml
@@ -1,7 +1,7 @@
 package:
   name: corepack
   version: 0.19.0
-  epoch: 0
+  epoch: 1
   description: Zero-runtime-dependency package acting as bridge between Node projects and their package managers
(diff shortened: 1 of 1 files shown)


rawkode:
cc @dlorenc @imjasonh as they reviewed the previous PR

ajayk (approved):

PR #3335, opened 2023-07-03
update ingress-nginx build to work with ingress-nginx image

strongjz:
#### For new package PRs only


rawlingsj:
Might be good for @amouat to take a look at this one too as I know you've spent some time looking at nginx.

ingress-nginx/etc/nginx/lua/balancer.lua
```diff
@@ -0,0 +1,374 @@
+local ngx_balancer = require("ngx.balancer")
```
rawlingsj, line 1: @strongjz the *.lua files that are being added here, I wonder if they are coming from an existing git repo that we could clone in the melange pipeline rather than duplicating here?  Just thinking it might make it hard to maintain these if / when they change?

strongjz: I was going to rethink how this was done with @imjasonh, havent had that conversation yet. It does come the ingress-nginx repo, here https://github.com/kubernetes/ingress-nginx/tree/main/images/nginx/rootfs

joshrwolf (approved):

PR #3350, opened 2023-07-04
Address vulnerabilities in zot

luhring:
Before:

```console
$ ./scripts/grype-scan-apk.sh ./packages/aarch64/zot-1.4.3-r5.apk json | jq '.matches | length'
28
```

After:

```console
$ ./scripts/grype-scan-apk.sh ./packages/aarch64/zot-1.4.3-r6.apk json | jq '.matches | length'
4
```

What's remaining is 2 GHSAs (multiplied by two binaries: `zot` and `zli`) in `github.com/notaryproject/notation-go`, where zot's code base depends directly on symbols in its current version of notation-go that are removed in the very next version — so it's not possible to resolve these vulnerabilities without breaking zot or reimplementing the affected piece of zot's code.

### DIFF
--- a/zot.yaml
+++ b/zot.yaml
@@ -1,7 +1,7 @@
 package:
   name: zot
-  version: 1.4.3
-  epoch: 5
+  version: 1.4.3 # Check if the Go module patching can be removed in the next release.
+  epoch: 6
   description: A production-ready vendor-neutral OCI-native container image registry (purely based on OCI Distribution Specification)
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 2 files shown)


zot.yaml
```diff
@@ -16,17 +16,44 @@ environment:
...
+      go mod tidy
+
+  - runs: |
+      # Zot hasn't been updated to handle the latest version of the image-spec module.
+      # Revert a particular breaking commit in image-spec (63b8bd0) until zot does upgrade.
+
+      moduleDir="$(go env GOPATH)/pkg/mod/github.com/opencontainers/image-spec@v1.1.0-rc3"
+      chmod -R u+w "$moduleDir"
+      patch -p1 -d "$moduleDir" -N < ./revert-removal-of-symbols.patch
+      go clean -cache
```
luhring, line 54: This reverts an upstream commit from @jonjohnsonjr 😁  — temporarily! We have reason to believe this won't be necessary for much longer.

dlorenc (approved):

PR #3357, opened 2023-07-04
Fix envoy build by pinning to gcc 12.

dlorenc:
The new gcc-12 package itself doesn't install, but this works and unblocks envoy updates.

Fixes:

Related:


dlorenc:
The update check should fail - I want to rebuild the existing version then let the update system bump.

rawlingsj (approved):

PR #3360, opened 2023-07-04
Pin elixir inside rabbitmq.

dlorenc:
New builds of elixir break the rabbitmq build.

Fixes:

Related:


dlorenc:
Update should fail, automation will bump after we fix the existing version.

luhring (approved):

PR #3369, opened 2023-07-05
move bash completions into separate subpackage

joshrwolf:
on the off chance we want the bash completions, keep them around but in a sub package.

this fixes the version conflicts between the bash completions, and brings them in line with the way we deconflict the clis


denhamparry (approved):
LGTM 🚪 🪓 🐙 🐺

PR #3375, opened 2023-07-05
update libjpeg-turbo to 3.0.0

ajayk:


dlorenc (commented):
Do we expect any breakages as part of the major version bump? Maybe we should test a few dependent builds too.

ajayk:
> Do we expect any breakages as part of the major version bump? Maybe we should test a few dependent builds too.

2.1.91 was also (3.0 beta 2) https://github.com/libjpeg-turbo/libjpeg-turbo/releases/tag/2.1.91

and there are  7 significant changes compared to 3.0 beta2 here https://github.com/libjpeg-turbo/libjpeg-turbo/releases/tag/3.0.0

ajayk:
I ran some dependent builds for libjpeg-turbo
- tiff
- libwebp
- guacamole-server

 https://gist.github.com/ajayk/366377c8a40857fd4ff0c52235fe174b

dlorenc (approved):

dlorenc:
Nice!

kaniini:
Should be fine, no ABI breaks as far as I can tell.

PR #3446, opened 2023-07-07
Adding php-amqp

erikaheidi:
Adding the `php-amqp` extension.

#### For new package PRs only

### DIFF
--- a/php-amqp.yaml
+++ b/php-amqp.yaml
@@ -1,0 +1,50 @@
+package:
+  name: php-amqp
+  version: 1.11.0
+  epoch: 0
+  description: "PHP extension to communicate with any AMQP compliant server"
+  copyright:
+    - license: PHP-3.01
+  dependencies:
+    runtime:
+      - php
+      - rabbitmq-c
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - autoconf
+      - busybox
+      - php
+      - php-dev
+      - rabbitmq-c-dev
+
+pipeline:
+  - uses: git-checkout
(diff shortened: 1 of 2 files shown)


php-amqp.yaml
```diff
@@ -0,0 +1,50 @@
...
+    runs: |
+      set -x
+      INSTALL_ROOT="${{targets.destdir}}" DESTDIR="${{targets.destdir}}" make install
+
+update:
+  enabled: true
+  github:
+    identifier: php-amqp/php-amqp
+    strip-prefix: v
+    tag-filter: v
```
rawlingsj, line 50:
```suggestion
    tag-filter: v
    use-tag: true
```
think adding this will fix the CI check

erikaheidi:
This package seems to handle releases differently, they use tags but don't make releases? Not quite sure how to handle the update here

php-amqp.yaml
erikaheidi: cool, let me try that!

rawlingsj (dismissed):

erikaheidi:
Awesome, thanks @rawlingsj for the quick 👍🏼 :)

rawlingsj:
> Awesome, thanks @rawlingsj for the quick 👍🏼 :)

Sure thing!  Looks like it may be removed from the merge queue because of a conflict of another change that just went in.  I tried to time them to go in together.  I'll sort out the conflict though and make sure it get's merged.

rawlingsj (approved):

PR #3476, opened 2023-07-07
Fix CVE-2023-32731 in pulumi packages

luhring:


rawlingsj (approved):

imjasonh (approved):

pulumi.yaml
```diff
@@ -29,25 +32,29 @@ pipeline:
...
   - working-directory: ${{package.name}}
     pipeline:
+      - runs: |
+          for d in pkg sdk; do
+            (
+              cd $d
+
+              # Mitigate CVE-2022-41723
+              go mod edit -dropreplace golang.org/x/text
+              go get golang.org/x/text@v0.9.0
```
imjasonh, line 42: (Not blocking, just for my own future edification) Is there a reason to prefer this over `go mod edit -replace golang.org/x/text@v0.9.0` ? Is there a difference?

luhring: IIUC the difference is that in the command you mention, there'd be a `replace` leftover for that module, whereas there wouldn't be in what's shown above. (FWIW these lines were here before.) I have more thoughts on `replace`s and I'd love other folks' thoughts too!

dlorenc: Fwiw I've used replace but only when there's an existing one that we need to override

PR #3505, opened 2023-07-10
Remedidate CVE-2023-1428 and CVE-2023-32732 in kube-fluentd-operator

rawlingsj:


luhring (changes requested):
I think we can close this and consider those CVEs to be false positives! 😃

rawlingsj:
yay!

PR #3621, opened 2023-07-14
CI: adds `wolfictl scan --require-zero` results as comment to PRs

mritunjaysharma394:
Since wolfictl now provides with `wolfictl scan <pkg_path> --require-zero`, we can for now comment on PRs, currently I have doubt on what should be the appropriate comment on the PR, that's one of the reasons to keep this still a draft PR, would love your inputs @luhring on this, this is very initial state of PR and has inspiration from that of @rawlingsj, I would love to know your suggestions and feedback to further improve it and reach it to the desired state, thanks!


scripts/wolfi-scan-apk.sh
```diff
@@ -0,0 +1,37 @@
...
+  apk_file="packages/${arch}/${package}-${version}.apk"
+  if [[ -f "$apk_file" ]]; then
+    echo "Processing ${apk_file}"
+    tmpdir=$(mktemp -d)
+
+    trap 'rm -rf "$tmpdir"' EXIT
+
+    tar -xf "$apk_file" -C "$tmpdir"
+
+    wolfictl "$tmpdir" --require-zero
```
luhring, line 33: One difference about how `wolfictl scan` works from how [grype-scan-apk.sh](https://github.com/wolfi-dev/os/blob/main/scripts/grype-scan-apk.sh) worked is that you pass the path of the apk file directly to `wolfictl scan`. It will fail if you pass it a directory path, and we also don't need to create a temp dir or do any untarring any more. That was just a hack I did for the shell script 😳

luhring (commented):
This looks mostly ready to me! @rawlingsj is more of an expert on the GitHub tricks here, but from my vantage point, I just had one comment about the shell script and then I think we're ready 🚀 

Is there a way to test this prior to merging?

scripts/wolfi-scan-apk.sh
mritunjaysharma394: Thanks for the review @luhring, since I haven't seen how packages.log looks like as I feel its created during the CI build, I wasn't sure earlier too about why we were using tmpdir here, but that makes sense now, so I should I should just go with `wolfictl "$apk_file" --require-zero` right? 

And yes you are right about testing too, I was wondering how to do that, maybe I can try that in my fork if that is possible by allowing github actions to run there? AFAIK, they are allowed for public repos, thanks and let me know the idea!

mritunjaysharma394: Also, how would you like the PR comment to look like, currently I guess it will just comment the error the above command shows? If that is enough, I think it is near to the shape we want then 🚀

rawlingsj: package.log will look like this
```
aarch64|ruby3.2-fluent-plugin-splunk-hec|ruby3.2-fluent-plugin-splunk-hec|1.3.2-r3
aarch64|ruby3.2-fluent-plugin-splunk-hec|ruby3.2-fluent-plugin-splunk-hec|1.3.2-r4
aarch64|deno|deno|1.35.1-r0
aarch64|gradle-8|gradle-8|8.2.1-r0
```
you can have that built locally when including `MELANGE_EXTRA_OPTS="--create-build-log"` as an option when running the `make package/foo` in your clone of wolfi os

On testing - yeah I use my fork, enable GH actions though you need to change like [this](https://github.com/rawlingsj/wolfi-os/commit/b312cde418ab921030336bc9872a4eafce59ccff#diff-b8790dc873e80c2f7f63bf198b4da0147787ad8954767b839967d29f65155e2cR46-R48) so that you can run on the free tier

scripts/wolfi-scan-apk.sh
```diff
@@ -0,0 +1,37 @@
...
+
+if ! command -v wolfictl > /dev/null; then
+  echo "This script requires Wolfictl to be installed. To install Wolfictl, check out https://github.com/wolfi-dev/wolfictl"
+  exit 1
+fi
+
+# Check if packages.log file exists
+
+if [[ ! -f "packages.log" ]]; then
+  echo "Cannot find packages.log file.  No apks to scan."
```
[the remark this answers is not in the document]
rawlingsj: unless you want this to help with running the script locally.

scripts/wolfi-scan-apk.sh
rawlingsj: One other thought is we add a flag to the new `wolfi scan` so that it uses the packages.log file directly, this would mean we can move the looping logic out of a script and directly into the go code which is easier to maintain and cover with tests.  Just an idea.  Then we can remove the script and the action would look similar to

```
      - name: Scan
        if: steps.file_check.outputs.exists == 'true'
        run: |
          wolfictl scan --use-package-log
```

Though using a script is fine too, just giving some options.  


**One important note** creating a PR comment in CI will only work if the PR is created from a branch on the main repo.  Not from forks, this is because secrets in GHA's are not available to PRs from forks.  Quite a few maintainers use branches rather than forks but for external maintainers we'll not see the comment be added and so will need to manually inspect GHA logs for scan results.

mritunjaysharma394: Thanks a lot @rawlingsj, that really helps a lot! I think I really like your idea of having a flag in wolfictl that will use-package-log and I can create a PR for the same if we want that, would love to know @luhring opinion on that too. 

The other point is quite interesting though, the current PR that I have created is from my fork and I will test it on my fork (not sure if GHA secrets will mess up), since I am not a maintainer here yet, I am also not sure if I can create a PR from the branch because I might not have write permissions but I will try how things work on my fork, thanks so much!

scripts/wolfi-scan-apk.sh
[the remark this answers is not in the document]
mritunjaysharma394: Oh got it, thanks for the heads up!

mritunjaysharma394:
Hi folks, I have tried to implement most of your suggestions and I liked the idea that @rawlingsj presented, I have tested the script locally and it works, next attempt I will try to see if I can test this in my fork or not otherwise I might need your help in getting it tested with GitHub Actions because we of course don't want PRs to be spammed, let me know if we have any other suggestions! Thanks

kaniini:
What is the current status of this?  Can it go in?

scripts/wolfi-scan-apk.sh
luhring: >One other thought is we add a flag to the new wolfi scan so that it uses the packages.log file directly, this would mean we can move the looping logic out of a script and directly into the go code which is easier to maintain and cover with tests. Just an idea.

This is an awesome idea @rawlingsj! I'm going to add this flag!

rawlingsj:
Going to close as we now have wolfi scan CI check

PR #3631, opened 2023-07-15
Cni static

joshrwolf:
`calico-cni` on boot installs the cni plugins onto the host via a host mount, since we can't dictate the host's GLIBC, this PR ensures we always build the cni plugin statically.

this also updates calico to use the wolfi built cni-plugins, which is also updated to be built statically.

comments are left in the package config for the next poor soul who goes down this 🐰 🕳️

### DIFF
--- a/calico.yaml
+++ b/calico.yaml
@@ -1,7 +1,7 @@
 package:
   name: calico
   version: 3.26.1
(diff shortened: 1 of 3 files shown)


denhamparry (approved):
LGTM

PR #3636, opened 2023-07-15
Add ngrep

Dentrax:
Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/ngrep.yaml
+++ b/ngrep.yaml
@@ -1,0 +1,53 @@
+package:
+  name: ngrep
+  version: 1.47
+  epoch: 0
+  description: A grep-like utility that allows you to search for network packets on an interface
+  copyright:
+    - license: custom
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+      - automake
+      - autoconf
+      - libpcap-dev
+      - pcre-dev
+
+# transform melange version 1.47 => 1_47
+var-transforms:
+  - from: ${{package.version}}
+    match: \.
+    replace: _
+    to: mangled-package-version
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/jpr5/ngrep
+      tag: V${{vars.mangled-package-version}}
+      expected-commit: 351ac491358f89a7860bbbe9963cb4bc7d60000f
+
+  - runs: autoreconf -vif
+
+  - uses: autoconf/configure
+    with:
+      opts: |
+        --enable-pcre \
+        --enable-ipv6 \
+        --with-pcap-includes=/usr/include/pcap \
+        --disable-pcap-restart
+
+  - runs: make
+
+  - uses: autoconf/make-install
+
+  - uses: strip
+
+update:
+  enabled: true
+  release-monitor:
+    identifier: 2084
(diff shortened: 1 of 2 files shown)


ngrep.yaml
```diff
@@ -0,0 +1,52 @@
...
+  - from: ${{package.version}}
+    match: \.
+    replace: _
+    to: mangled-package-version
+
+pipeline:
+  - uses: fetch
+    with:
+      uri: https://github.com/jpr5/ngrep/archive/V${{vars.mangled-package-version}}.tar.gz
+      expected-sha512: 47ba65878df6b555701c866721a8a935eabdcce636d398284cbfe5f63baf68c62d994a2f373ea4fc8f44fbed3eecee149f2ee48d39c71c04d34e5a088db8c657
```
luhring, line 31: I think we prefer `git-checkout` whenever we know where to find the source code repository

ngrep.yaml
```diff
@@ -0,0 +1,52 @@
...
+  - runs: make
+
+  - uses: autoconf/make-install
+
+  - uses: strip
+
+update:
+  enabled: true
+  release-monitor:
+    identifier: 2084
```
luhring, line 52: I think we prefer to use the GitHub update functionality over release-monitor when we're getting the source code from GitHub.

But also this project doesn't look maintained any more 🤔

Dentrax: It seems that the last was made in 2019. Also I couldn't see an indication about whether it's maintaining or not. What should we do?

luhring: Yeah, good question. I didn't find a definitive answer either. For now I wouldn't let that block us, but I probably would switch to using GitHub tags as the update trigger

luhring: I also saw this in the CI logs:

```
level=fatal msg="error during command execution: 1 error occurred:\n\t* package ngrep: failed to validate update config, melange bump: got 404 Not Found when fetching [https://github.com/jpr5/ngrep/archive/V${{vars.mangled-package-version}}.tar.gz\n\n](https://github.com/jpr5/ngrep/archive/V$%7B%7Bvars.mangled-package-version%7D%7D.tar.gz/n/n)"
```

Dentrax: Sure, i will update the `github` section. That error because of `wolfictl check update` does not respect `var-transforms` so it didn't evaluate `vars.`.

luhring: Oh shoot, okay gotcha. Do we have a GitHub issue tracking that need?

Dentrax: Thanks, not needed IMO.

Dentrax: Hmmm, I don't think so, but not 100% sure.

Dentrax: Pushed a commit to use `git-checkout`, but I think we can't use `github:` since ngrep uses different versioning semantics. I got the following error: `update found newer version 1461 compared with package.version 1.47 in melange config`. IMHO, `release-monitor` way better. WDYT?

luhring: Yeah agreed. Let's not block on this — but this is unexpected behavior (to me at least 😄 ) so @rawlingsj maybe we can revisit how this works at some point.

Thanks @Dentrax!

luhring (approved):

rawlingsj (approved):

PR #3637, opened 2023-07-15
add dstat

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/dstat.yaml
+++ b/dstat.yaml
@@ -1,0 +1,45 @@
+package:
+  name: dstat
+  version: 0.7.4
+  epoch: 0
+  description: A versatile resource statistics tool
+  copyright:
+    - license: GPL-2.0-only
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+      - python3-dev
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/dagwieers/dstat
+      tag: v${{package.version}}
+      expected-commit: 6f5db0aed26bf8cf2700d4ffe90a9bd3436ac728
+
+  - uses: patch
+    with:
+      # The patch source is https://github.com/efexgee/dstat/commit/220a785321b13b6df92a536080aca6ef1cb644ad.patch
+      patches: 220a785321b13b6df92a536080aca6ef1cb644ad.patch
+
+  - runs: |
+      make DESTDIR=${{targets.destdir}} install
+      make DESTDIR="${{targets.destdir}}" docs
+      # Get rid of deprecation warning.
+      sed -i "s/import collections/import collections.abc/g" -i "${{targets.destdir}}"/usr/bin/dstat
+      sed -i "s/collections.Sequence/collections.abc.Sequence/g" -i "${{targets.destdir}}"/usr/bin/dstat
+
+  - uses: strip
+
+subpackages:
+  - name: dstat-doc
+    description: dstat manpages
+    pipeline:
+      - uses: split/manpages
+
+update:
+  enabled: false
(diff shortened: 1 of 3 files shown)


dstat.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+      - python3-dev
+
+pipeline:
+  - uses: fetch
+    with:
+      uri: https://github.com/dagwieers/dstat/archive/v${{package.version}}.tar.gz
+      expected-sha512: d100223887ebd83d0fd1259436f39419a85df9002556abcfc8e4195aa63be73d85707dcd5c4fb56a979b29131485dce6e97c177fbfca08dc50bd0f92b15cc6f5
```
luhring, line 21: I think we prefer `git-checkout` whenever we know where to find the source code repository

dstat.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+  - uses: patch
+    with:
+      # The patch source is https://github.com/efexgee/dstat/commit/220a785321b13b6df92a536080aca6ef1cb644ad.patch
+      patches: 220a785321b13b6df92a536080aca6ef1cb644ad.patch
```
luhring, line 26: Why are we including this patch?

dstat.yaml
```diff
@@ -0,0 +1,46 @@
...
+subpackages:
+  - name: dstat-doc
+    description: dstat manpages
+    pipeline:
+      - uses: split/manpages
+
+update:
+  enabled: true
+  release-monitor:
+    identifier: 630
```
luhring, line 46: I think we prefer to use the GitHub update functionality over release-monitor when we're getting the source code from GitHub.

But also this project is no longer maintained according to its README

dstat.yaml
```diff
@@ -0,0 +1,46 @@
+package:
+  name: dstat
+  version: 0.7.4
+  epoch: 0
+  description: A versatile resource statistics tool
+  copyright:
+    - license: GPL
```
luhring, line 7: Is this right? Should it be `GPL-2.0-only`?

dstat.yaml
developer-guy: I used Arch Linux PKGBUILD to build this: _https://gitlab.archlinux.org/archlinux/packaging/packages/dstat/-/blob/main/PKGBUILD_

dstat.yaml
developer-guy: it should be yes, thank you.

dstat.yaml
developer-guy: I think we should disable the updates as you said this project is not actively maintained any more, thanks.

dstat.yaml
luhring: Right — but that doesn't explain what the purpose of the patch is. We need to make that clear here in this file, or we should leave the patch out. Friendly reminder, Wolfi's reason for including a patch should not be "because another distro did it". 😃 

At face value, I see taking a tag from 2019 and adding in a patch from 2001, and it doesn't make sense. If we have a story to make it make sense, that'd be great!

developer-guy: it says _"Subject: [PATCH] Fixed bug when specifying a delay"_ in the patch file, is that enough information for us to keep it?

developer-guy: I think the bug seems not fixed in the upstream, so I assume that we may need that patch still.

_https://github.com/dstat-real/dstat/blob/e428c7dc7137f81f8ca6bef8854b37a4ddd4f337/dstat#L2700_

luhring (dismissed):

luhring (approved):

PR #3641, opened 2023-07-16
add ltrace

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/ltrace.yaml
+++ b/ltrace.yaml
@@ -1,0 +1,48 @@
+package:
+  name: ltrace
+  version: 0.7.3
+  epoch: 0
+  description: Tracks runtime library calls in dynamically linked programs
+  copyright:
+    - license: GPL-2.0-or-later
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+      - autoconf
+      - automake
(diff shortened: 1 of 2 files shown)


ltrace.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+  - uses: patch
+    with:
+      # https://git.alpinelinux.org/aports/plain/main/ltrace/musl.patch
+      patches: musl.patch
+
+  - uses: patch
+    with:
+      # https://git.alpinelinux.org/aports/plain/main/ltrace/ltrace_aarch64.patch
+      patches: ltrace_aarch64.patch
```
luhring, line 39: Why do we need these?

developer-guy: they were available in APKBUILD

luhring: See explanation here -> https://github.com/wolfi-dev/os/pull/3637#discussion_r1264675856

esp.:
>Wolfi's reason for including a patch should not be "because another distro did it"

developer-guy: I confirmed that the build was succeed without these patches, so, I removed them 🟢

luhring (approved):

PR #3643, opened 2023-07-16
add mycli python deps

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs


py3-cffi.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - build-base
+      - python-3.11
+      - python-3.11-dev
+      - py3.11-setuptools
+      - libffi-dev
+
+pipeline:
+  - uses: fetch
+    with:
+      README: 'CONFIRM WITH: curl -L https://files.pythonhosted.org/packages/2b/a8/050ab4f0c3d4c1b8aaa805f70e26e84d0e27004907c5b8ecc1d31815f92a/cffi-1.15.1.tar.gz | sha256sum'
```
rawlingsj, line 29: `README` isn't a valid option for the [fetch](https://github.com/chainguard-dev/melange/blob/main/pkg/build/pipelines/fetch.yaml) pipeline.  I'm guessing this comes from the `melange convert` tool when creating python related packages?  We don't validate the schema which would help automatically help catch things like this TBH.

I think it's worth dropping the `README:` option for all these packages as they will become out of date after the next wolfi auto update.

rawlingsj (changes requested):
One more comment - sorry I should have spotted this when adding the last review.  It looks like the `melange convert python` is creating the fetch URLs to pull from https://pypi.org and include a sha in the URL.  This means we can't auto update these package with wolfictl.  

After checking a few of these packages I think all the sources come from github repos?  If we use `git-checkout` repo rather than fetch from https://pypi.org  we can keep these packages up-to-date with our automation.

py3-cffi.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - python-3.11
+      - python-3.11-dev
+      - py3.11-setuptools
+      - libffi-dev
+
+pipeline:
+  - uses: fetch
+    with:
+      expected-sha256: d400bfb9a37b1351253cb402671cea7e89bdecc294e8016a707f6d1d8ac934f9
+      uri: https://files.pythonhosted.org/packages/2b/a8/050ab4f0c3d4c1b8aaa805f70e26e84d0e27004907c5b8ecc1d31815f92a/cffi-${{package.version}}.tar.gz
```
rawlingsj, line 30: I think we can switch this to remove the sha and auto upgrades should work
```suggestion
      uri: https://foss.heptapod.net/pypy/cffi/-/archive/v${{package.version}}/cffi-v${{package.version}}.tar.gz
```

py3-cffi.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+environment:
+  contents:
+    packages:
+      - ca-certificates-bundle
+      - wolfi-base
+      - busybox
+      - build-base
+      - python-3.11
+      - python-3.11-dev
```
rawlingsj, line 22: do these need to be python 3.11?  Or can they track the latest python? i.e. dropping the `.11` similar to other examples
```
https://github.com/wolfi-dev/os/blob/d28eeb644d789dd48dfe39aa70fc29ba44672ddb/py3-pluggy.yaml#L19-L21
```

py3-pyperclip.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - name: Python Build
+    runs: python setup.py build
+
+  - name: Python Install
+    runs: python setup.py install --prefix=/usr --root="${{targets.destdir}}"
+
+  - uses: strip
+
+update:
+  enabled: true
```
rawlingsj, line 38:
```suggestion
  enabled: true
  manual: true # fetch URL contains a sha so will need to be manually updated
```

py3-cffi.yaml
developer-guy: thank you!

py3-cffi.yaml
developer-guy: let's give it a try, actually this content generated by the `melange convert python` then I didn't change any of the packages afterwards.

rawlingsj (approved):

PR #3657, opened 2023-07-17
add the multiple k3s maintained versions

joshrwolf:
`k3s` maintains the same versioning support scheme as `k8s`, so  the strategy here is copied from how we did `kubernetes-1.2{4,5,6,7}`


k3s-1.26.yaml
```diff
@@ -0,0 +1,173 @@
...
+                image=$(crane digest --full-ref cgr.dev/chainguard/coredns:latest)
+                ;;
+              docker.io/rancher/mirrored-library-busybox)
+                image=$(crane digest --full-ref cgr.dev/chainguard/busybox:latest)
+                ;;
+              docker.io/rancher/mirrored-library-traefik)
+                image=$(crane digest --full-ref cgr.dev/chainguard/traefik:latest)
+                ;;
+              *)
+                image="$repo:$tag@$(crane digest $repo:$tag)"
```
mattmoor, line 150: I'm wondering if we should fail on unknown do signal that we have gaps?

joshrwolf: these haven't changed in years, but I like that idea

PR #3658, opened 2023-07-17
remove unneeded `apk-tools` from kubernetes build environment

joshrwolf:
this is a nit, but lets be perfectionists?


luhring (approved):
💯

PR #3692, opened 2023-07-18
add a CI action to build the stateful world

joshrwolf:
Adds an action to use our fancy new self hosted runners to build the world.

This current iteration ironically keeps `BUILDWORLD=no` for now until we're confident the dag is correct.

### DIFF
--- a/.github/workflows/build-world.yaml
+++ b/.github/workflows/build-world.yaml
@@ -1,0 +1,69 @@
+name: Build the world
+
+on:
+  workflow_dispatch:
+    
+jobs:
+  build:
+    name: Build OS packages
+    runs-on:
+      - self-hosted
+      - ${{ matrix.arch }}
+
+    # Only run one build per arch at a time
+    concurrency:
+      group: build-world-${{ matrix.arch }}
+
+    strategy:
+      matrix:
+        arch: [ "x86_64", "aarch64" ]
+      fail-fast: false
+
+    container:
+      image: ghcr.io/wolfi-dev/sdk:latest@sha256:78751faebed64630b64351fff0431e38cf08c70997c19a72146698ced651aad0
+      options: |
+        --cap-add NET_ADMIN --cap-add SYS_ADMIN --security-opt seccomp=unconfined --security-opt apparmor:unconfined
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: 'Trust the github workspace'
+        run: |
+          # This is to avoid fatal errors about "dubious ownership" because we are
+          # running inside of a container action with the workspace mounted in.
+          git config --global --add safe.directory "$GITHUB_WORKSPACE"
+
+      - name: 'Generate local signing key'
+        run: |
+          make MELANGE="melange" local-melange.rsa
+
+      - name: 'Build Wolfi World'
+        run: |
+          # Run all and allow failures, all we really care about is the buildlogs for failures
+          make MELANGE="melange" MELANGE_EXTRA_OPTS="--create-build-log" REPO="$GITHUB_WORKSPACE/packages" BUILDWORLD=no ARCH="${{ matrix.arch }}" all -j1 -k
+
+          # Move the failures into another directory (for ease of uploading)
+          mkdir -p buildlogs-failures
+          grep -L -r "generating apk index from packages in packages/" packages/**/buildlogs | xargs -I '{}' cp '{}' buildlogs-failures
+
+      - name: Upload failed build logs
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: buildlogs-failures
+          path: ${{ github.workspace }}/buildlogs-failures/*
+          retention-days: 7
+
+  # TODO: Enable the notification
+  # postrun:
+  #   name: Notify failures
+  #   runs-on: ubuntu-latest
+  #   if: failure()
+  #   steps:
+  #     - uses: slackapi/slack-github-action@007b2c3c751a190b6f0f040e47ed024deaa72844 # v1.23.0
+  #       id: slack
+  #       with:
+  #         payload: '{"text": "[build world] failure: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"}'
+  #       env:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-world.yaml
```diff
@@ -0,0 +1,96 @@
...
+        run: |
+          # Copy wolfictl out of the wolfictl image and onto PATH
+          TMP=$(mktemp -d)
+          docker run --rm -i -v $TMP:/out --entrypoint /bin/sh ghcr.io/wolfi-dev/sdk:latest@sha256:78751faebed64630b64351fff0431e38cf08c70997c19a72146698ced651aad0 -c "cp /usr/bin/wolfictl /out"
+          echo "$TMP" >> $GITHUB_PATH
+
+      - name: 'Build Wolfi'
+        run: |
+          # Run all and allow failures, all we really care about is the buildlogs for failures
+          make MELANGE="sudo -E melange" MELANGE_DIR=/usr/share/melange KEY=wolfi-signing.rsa BUILDWORLD=no REPO="${{ github.workspace }}/packages" -j1 -k all
```
joshrwolf, line 71: keeping this `BUILDWORLD=no` intentionally since the goal for now is to get through a build of _all_ our packages, and `BUILDWORLD=no` + `-k` won't get us very far if we can't rely on prior state

.github/workflows/build-world.yaml
```diff
@@ -0,0 +1,96 @@
...
+
+on:
+  workflow_dispatch:
+    
+jobs:
+  build:
+    name: Build OS packages
+    runs-on:
+      - self-hosted
+      - ${{ matrix.arch }}
```
mattmoor, line 11: Are we sure this works?

.github/workflows/build-world.yaml
```diff
@@ -0,0 +1,96 @@
...
+
+    # Only run one build per arch at a time
+    concurrency:
+      group: build-world-${{ matrix.arch }}
+
+    strategy:
+      matrix:
+        arch: [ "x86_64", "arm64" ]
```
mattmoor, line 19: Maybe the ship has sailed on this, but aarch64 might be more consistent with APK's sense of architectures.

joshrwolf: I can/will change it!

.github/workflows/build-world.yaml
joshrwolf: I stole this from: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-using-a-multi-dimension-matrix

.github/workflows/build-world.yaml
```diff
@@ -0,0 +1,96 @@
...
+
+      - uses: actions/checkout@v3
+
+      - run: echo "${{ secrets.MELANGE_RSA }}" > ./wolfi-signing.rsa
+
+      - run: |
+          sudo mkdir -p /etc/apk/keys
+          sudo cp ./wolfi-signing.rsa.pub /etc/apk/keys/wolfi-signing.rsa.pub
+
+      - uses: chainguard-dev/actions/setup-melange@main
```
mattmoor, line 38: Not to be disruptive to this change, but I wonder if we should migrate to copying this out of the SDK image as well 🤔

.github/workflows/build-world.yaml
```diff
@@ -0,0 +1,96 @@
...
+          # Move the failures into another directory (for ease of uploading)
+          mkdir -p buildlogs-failures
+          grep -L -r "generating apk index from packages in packages/" packages/**/buildlogs | xargs -I '{}' cp '{}' buildlogs-failures
+
+      - name: Upload failed build logs
+        uses: actions/upload-artifact@v3
+        with:
+          name: buildlogs-failures
+          path: ${{ github.workspace }}/buildlogs-failures/*
+          retention-days: 7
```
mattmoor, line 82: neat TIL

.github/workflows/build-world.yaml
```diff
@@ -0,0 +1,96 @@
...
+      - name: 'Build Wolfi'
+        run: |
+          # Run all and allow failures, all we really care about is the buildlogs for failures
+          make MELANGE="sudo -E melange" MELANGE_DIR=/usr/share/melange KEY=wolfi-signing.rsa BUILDWORLD=no REPO="${{ github.workspace }}/packages" -j1 -k all
+
+          # Move the failures into another directory (for ease of uploading)
+          mkdir -p buildlogs-failures
+          grep -L -r "generating apk index from packages in packages/" packages/**/buildlogs | xargs -I '{}' cp '{}' buildlogs-failures
+
+      - name: Upload failed build logs
```
mattmoor, line 77: I think this needs an `if: failure()`

.github/workflows/build-world.yaml
joshrwolf: afaiu `setup-melange` does a bunch of other prereq related things (ie: bwrap setup) that make it harder than just copying the binary out like we did for `wolfictl`, but if there's precedence let me know

.github/workflows/build-world.yaml
joshrwolf: I added an `if: always()` just in case the prior `make -k` doesn't end up wai

mattmoor (dismissed):

.github/workflows/build-world.yaml
kaniini: If this is a bare build, which isn't uploading anything, why not just use the SDK image directly?

kaniini (changes requested):
Lets use the SDK image instead of setup-melange here

.github/workflows/build-world.yaml
imjasonh: We do this in our regular builds too: https://github.com/wolfi-dev/os/blob/main/.github/workflows/push-production.yaml#L38

kaniini: Yes, I know.  What I am saying is that we should stop doing that, and use the SDK image instead.

kaniini: (In other words, we should avoid adding additional variations of this particular technical debt.)

kaniini (approved):

PR #3731, opened 2023-07-19
Remove escaping from replacement strings in icu

mattmoor:
I could not build this locally, and I am fairly certain it was due to the escaping in the replacement side, which doesn't look right to me.

I'm building this locally on arm64 now, and it's past the fetch step, but still running.

### DIFF
--- a/icu.yaml
+++ b/icu.yaml
@@ -1,7 +1,7 @@
 package:
   name: icu
   version: 73.1
-  epoch: 0
+  epoch: 1
   description: "International Components for Unicode library"
   copyright:
     - license: MIT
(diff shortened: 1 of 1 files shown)


mattmoor:
@rawlingsj is the bit about this still requiring manual updates still true?
https://github.com/wolfi-dev/os/commit/6684cefd840aaa4437fc934b894e0747701d940e

If so, then I'm not sure why wolfictl is able to figure things out, but we should probably fix the presubmit check or re-enable updates 🤔

mattmoor:
Ah I think I just misread the error, since it's a 404 with the templated URL.

The aarch64 build completed for me with this change, so hopefully the x64 build in actions agrees 🤞

kaniini:
wolfictl still does not understand variable transforms, i'm going to work on applying them in melange in a more generic way that should automatically fix wolfictl

mattmoor:
@kaniini should we merge this around the presubmit since it's not blocking?  I can also just bump the package if we think that's the "right" way to fix this.

kaniini (approved):

kaniini:
The wolfictl update check presubmit isn't required to pass, so go ahead and merge. ✅

PR #3735, opened 2023-07-19
add kubeflow pipelines

developer-guy:
Fixes:

Related:

- [x] - ~persistenceagent~
- [x] - ~scheduledworkflow~
- [x] - ~cache-server~
- [x] - ~apiserver~
- [x] - ~viewer-crd-controller~
- [x] - ~cache-deployer~
- [x] - ~frontend~
- [X] - metadata-writer
- [ ] - visualization-server - needs tensorflow to be package (will submit another PR)

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs


kubeflow-pipelines-scheduledworkflow.yaml
```diff
@@ -0,0 +1,64 @@
+package:
+  name: kubeflow-pipelines-scheduledworkflow
```
rawlingsj, line 2: I think we can put all kubeflow-pipelines related packages under one main one and then have subpackages for all the pipeline components?

I.e. I had a tinker a little while back at using data ranges, something like this https://github.com/wolfi-dev/os/pull/2206

kubeflow-pipelines.yaml
```diff
@@ -0,0 +1,123 @@
...
+
+          name=$(echo "${{range.value}}" | cut -d ';' -f 1)
+          path=$(echo "${{range.value}}" | cut -d ';' -f 2-)
+
+          make build/$name
+          install -Dm644 ./build/$name ${{targets.subpkgdir}}/usr/bin/${{range.key}}
+
+          go-licenses check ./src/$path
+          go-licenses csv ./src/$path > ${{targets.subpkgdir}}/third_party/$name.csv
+          go-licenses save ./src/$path --save_path ${{targets.subpkgdir}}/third_party/NOTICES
```
rawlingsj, line 84: I'm assuming we need to do the same and add go-licenses?

kubeflow-pipelines.yaml
```diff
@@ -0,0 +1,123 @@
...
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/kfp/cache/deployer/
+          mkdir -p ${{targets.subpkgdir}}/third_party
+          cp third_party/license.txt ${{targets.subpkgdir}}/third_party/
+          cd backend/src/cache/deployer
+          chmod -R 777 ${{targets.subpkgdir}}/kfp/cache/deployer/
+          cp deploy-cache-service.sh ${{targets.subpkgdir}}/kfp/cache/deployer/
+          cp webhook-create-signed-cert.sh ${{targets.subpkgdir}}/kfp/cache/deployer/
+          cp webhook-patch-ca-bundle.sh ${{targets.subpkgdir}}/kfp/cache/deployer/
```
rawlingsj, line 104: I wonder if it would be easier to 
```
install -m755 -D third_party/license.txt ${{targets.subpkgdir}}/third_party/license.txt
install -m755 -D backend/src/cache/deployer/deploy-cache-service.sh ${{targets.subpkgdir}}/kfp/cache/deployer/deploy-cache-service.sh
install -m755 -D backend/src/cache/deployer/webhook-create-signed-cert.sh ${{targets.subpkgdir}}/kfp/cache/deployer/webhook-create-signed-cert.sh
install -m755 -D backend/src/cache/deployer/webhook-patch-ca-bundle.sh ${{targets.subpkgdir}}/kfp/cache/deployer/webhook-patch-ca-bundle.sh

```

kubeflow-pipelines.yaml
```diff
@@ -0,0 +1,123 @@
...
+            pip3 install --prefer-binary -r requirements.txt --prefix=/usr --root=${{targets.subpkgdir}}
+            cp -r ../samples/* ${{targets.subpkgdir}}/samples
+            cp ./src/$path/config/sample_config.json ${{targets.subpkgdir}}/samples/sample_config.json
+          fi
+
+  - name: "kubeflow-pipelines-cache-deployer"
+    description: "Kubeflow Pipelines cache-deployer"
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/kfp/cache/deployer/
```
rawlingsj, line 97: `/kfp/cache/deployer/` doesn't seem standard, same with `/sample` above and the other target folders in the yaml

Maybe we should add config to `/etc/kfp` and then add a compat subpackage with symlinks to the original?  

@kaniini what's your thoughts as I think you'd like to have more of a standard approach here but maybe not?

rawlingsj (dismissed):

remove-twitter-artifactory.patch
rawlingsj, line 1: empty file can be removed

kubeflow-pipelines.yaml
Dentrax: Makes sense, liked the idea!

kubeflow-pipelines.yaml
Dentrax: Done!

kubeflow-pipelines.yaml
Dentrax: I'm not sure if I followed you here. Could you please rephrase the question?

Dentrax (commented):
Thanks for the reviews, @rawlingsj! Comments are addressed. PTAL when possible.

kubeflow-pipelines.yaml
rawlingsj: it was my poorly worded comment sorry.  I was wondering if we really needed to include these licenses but let's leave as.

kubeflow-pipelines.yaml
```diff
@@ -0,0 +1,132 @@
...
+            cp -r ../samples/* ${{targets.subpkgdir}}/samples
+            cp ./src/$path/config/sample_config.json ${{targets.subpkgdir}}/samples/sample_config.json
+          fi
+
+  - name: "kubeflow-pipelines-cache-deployer"
+    description: "Kubeflow Pipelines cache-deployer"
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/usr/bin
+          mkdir -p ${{targets.subpkgdir}}/third_party
```
rawlingsj, line 98: non blocking but we don't need to mkdir when using `install`

kubeflow-pipelines.yaml
```diff
@@ -0,0 +1,132 @@
...
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/server
+          mkdir -p ${{targets.subpkgdir}}/client
+          cd frontend/
+          npm ci
+          npm run postinstall
+          npm run build
+          cp -r ./server/* ${{targets.subpkgdir}}/server
+          cp -r ./build/* ${{targets.subpkgdir}}/client
```
rawlingsj, line 126: should this be in the main pipeline and the compat subpackage just be symlinks?  I'm thinking that by default we'd want everything to work using our standard folder structure and use a compat subpackage for symlinks where there are non standard folder layouts.

Dentrax: Definitely agree with you. But in this case, I couldn't find a way out since `npm` outputs multiple files under both of `/server/` and `/build/` folders. There is no single binary that we can put under `/usr/bin/`. On the other hand, creating a folder under `/usr/bin/` would look weird. How should we proceed here?

kubeflow-pipelines.yaml
```diff
@@ -0,0 +1,178 @@
...
+        - py3-requests
+        - py3-rsa
+        - py3-six
+        - py3-urllib3
+        - py3-websocket-client
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/opt/metadata_writer
+          cd backend/metadata_writer
+          pip3 install -r requirements.txt --ignore-installed --prefix=/usr --root="${{targets.destdir}}"
```
Dentrax, line 164: Is this the right way to use? since all deps has been included in `runtime`.

kubeflow-pipelines.yaml
rawlingsj: could we have both a frontend and compat package?

i.e. something like

```yaml
  - name: "kubeflow-pipelines-frontend"
    description: "Kubeflow Pipelines frontend"
    pipeline:
      - runs: |
          cd frontend/
          npm ci
          npm run postinstall
          npm run build
          cp -r ./server/* ${{targets.subpkgdir}}/usr/bin/
          cp -r ./build/* ${{targets.subpkgdir}}/usr/bin/
```
and then the compat

```yaml
  - name: "kubeflow-pipelines-frontend-compat"
    description: "Kubeflow Pipelines frontend compatibility package"
    pipeline:
      - runs: |
          mkdir -p ${{targets.subpkgdir}}/server
          mkdir -p ${{targets.subpkgdir}}/client

          ln -sf /usr/bin ${{targets.subpkgdir}}/server
          ln -sf /usr/bin ${{targets.subpkgdir}}/client
```

I wonder if something like that would work?  This way we are installing the binaries into the standard wolfi location and using symlinks to support upstream non standard folders?

rawlingsj:
just the one comment and will need to be rebased so that the new arm CI build triggeres

kubeflow-pipelines.yaml
rawlingsj: To be sure I wonder if we need `--find-links /usr/bin`or something?

https://stackoverflow.com/questions/11091623/how-to-install-packages-offline

Dentrax: Actually, I'm not familiar with that flag, and it seems no actively used in the codebase 🤔

Dentrax:
> just the one comment and will need to be rebased so that the new arm CI build triggeres

Thanks! Rebased and all CI is green.

kubeflow-pipelines.yaml
rawlingsj: It would be good to confirm we are using the local packages and not downloading them here.  Do the build logs show anything?

Also I wonder if @strongjz or @kaniini know more?

kaniini: We can check the APK to see if any unexpected files are there

Dentrax: All the site-packages are matching with what included in the `dependencies.runtime`: https://pastebin.com/rxThunQL

developer-guy: I removed this line because we install all the packages, we should not need to do "pip install" again.

Dentrax: migrated `pip3 install` to `gpep517`.

kaniini (approved):

PR #3737, opened 2023-07-20
add a *temporary* large timeout for build world's CI

joshrwolf:
it appears this needs to be manually set, even for self hosted runners 💆

### DIFF
--- a/.github/workflows/build-world.yaml
+++ b/.github/workflows/build-world.yaml
@@ -24,6 +24,8 @@ jobs:
       options: |
         --cap-add NET_ADMIN --cap-add SYS_ADMIN --security-opt seccomp=unconfined --security-opt apparmor:unconfined
 
+    timeout-minutes: 2880 # 2 days
+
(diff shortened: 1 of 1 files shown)


.github/workflows/build-world.yaml
```diff
@@ -24,6 +24,8 @@ jobs:
       options: |
         --cap-add NET_ADMIN --cap-add SYS_ADMIN --security-opt seccomp=unconfined --security-opt apparmor:unconfined
 
+    timeout-minutes: 2880 # 2 days
```
joshrwolf, line 27: purposely over estimated until we see green

mattmoor (approved):

.github/workflows/build-world.yaml
kaniini: I mean, eventually it will take 2 days to build Wolfi from scratch anyway.  The more packages we add, the longer it will take ;)

PR #3756, opened 2023-07-20
move ko-fips into ko subpackage

imjasonh:
This change makes the FIPS go packages `replaces: go`, to (I think) resolve this error that we get otherwise:

```
2023/07/20 13:33:58 error during command execution: failed to build package: unable to build guest: unable to generate image: installing apk packages: installing packages: installing go-fips-1.20: unable to install files for pkg go-fips-1.20: unable to install file over existing one, different contents: usr/lib/go/bin/go
```


kaniini:
Why not just add `provides: go=1.19.0`, etc?  Then the solver will just select `go-fips` as a total replacement for `go` and not install both packages at once.

kaniini (approved):

go-fips-1.19.yaml
```diff
@@ -1,13 +1,14 @@
...
   version: 1.19.11
-  epoch: 0
+  epoch: 1
   description: "the Go programming language with OpenSSL cryptography"
   copyright:
     - license: BSD-3-Clause
   dependencies:
     provides:
       - go-fips=1.19.999 # This should be go-fips=${{package.version}}-${{package.epoch}}
+      - go=1.19.999
```
kaniini, line 11:
```suggestion
      - go=1.19
```

go-fips-1.20.yaml
```diff
@@ -1,13 +1,14 @@
...
   version: 1.20.6
-  epoch: 0
+  epoch: 1
   description: "the Go programming language with OpenSSL cryptography"
   copyright:
     - license: BSD-3-Clause
   dependencies:
     provides:
       - go-fips=1.20.999 # This should be go-fips=${{package.version}}-${{package.epoch}}
+      - go=1.20.999
```
kaniini, line 11:
```suggestion
      - go=1.20
```

kaniini (changes requested):

PR #3784, opened 2023-07-21
update nginx data

strongjz:
Update data from nginx and controller being copied into the final nginx directory


joshrwolf (dismissed):

joshrwolf (dismissed):

rawlingsj:
I think we can close this now as there's been more work on nginx since.

PR #3798, opened 2023-07-21
Apply a patch to fix libwebsocket build

mattmoor:
Currently this is broken at HEAD because of a type mismatch with openssl headers.

This picks up the upstream patch linked in the comment to correct things.

### DIFF
--- a/libwebsockets.yaml
+++ b/libwebsockets.yaml
@@ -29,7 +29,7 @@ pipeline:
 
   # This is from:
   # https://patch-diff.githubusercontent.com/raw/warmcat/libwebsockets/pull/2824.patch
-  # To fix a type mismatch error with openssl types.
+  # To fix an issue where an integer was being used instead of an enumerable type.
   - uses: patch
     with:
       patches: 2824.patch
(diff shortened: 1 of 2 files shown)


libwebsockets.yaml
```diff
@@ -27,6 +27,13 @@ pipeline:
       expected-sha256: 6a85a1bccf25acc7e8e5383e4934c9b32a102880d1e4c37c70b27ae2a42406e1
       uri: https://github.com/warmcat/libwebsockets/archive/v${{package.version}}.tar.gz
 
+  # This is from:
+  # https://patch-diff.githubusercontent.com/raw/warmcat/libwebsockets/pull/2824.patch
+  # To fix a type mismatch error with openssl types.
```
luhring, line 32: 👨‍🍳 💋

luhring (dismissed):

libwebsockets.yaml
mattmoor: 😁

kaniini, line 32:
```suggestion
  # To fix an issue where an integer was being used instead of an enumerable type.
```

luhring (approved):

PR #3800, opened 2023-07-21
gettext: add glibc-iconv dependency

kaniini:
The glibc iconv plugins are needed to convert non-unicode translation data to unicode.
Fixes: xkeyboard-config failure to rebuild

cc @mattmoor


mattmoor (approved):
Awesome 🙏

PR #3801, opened 2023-07-21
new package: vscode

imjasonh:
#### For new package PRs only

### DIFF
--- a/vscode.yaml
+++ b/vscode.yaml
@@ -1,0 +1,47 @@
+package:
+  name: vscode
+  version: 1.80.1
+  epoch: 0
+  description: "Visual Studio Code"
+  copyright:
+    - license: MIT
+
+environment:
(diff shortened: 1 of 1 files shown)


vscode.yaml
```diff
@@ -0,0 +1,47 @@
...
+      expected-commit: 74f6148eb9ea00507ec113ec51c489d6ffb4b771
+      destination: vscode
+
+  - working-directory: /home/build/vscode
+    runs: |
+      yarn
+
+      node build/lib/preLaunch.js
+      mkdir -p "${{targets.destdir}}"/usr/bin
+      install -dm755 .build/electron ${{targets.destdir}}/usr/bin/
```
kaniini, line 41: Shouldn't we build our own copy of `electron`, and use that?

PR #3810, opened 2023-07-22
cmake: bump to v3.27.0

### LINKED ISSUE #3713: cmake/3.27.0 new package update

Elizafox:
Tested with multiple packages:
- systemd
- jsonc
- libogg
- tcpdump

Fixes: #3713

### DIFF
--- a/cmake.yaml
+++ b/cmake.yaml
@@ -30,6 +30,8 @@ pipeline:
       uri: https://www.cmake.org/files/v3.27/cmake-${{package.version}}.tar.gz
       expected-sha256: aaeddb6b28b993d0a6e32c88123d728a17561336ab90e0bf45032383564d3cb8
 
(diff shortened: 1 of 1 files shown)


cmake.yaml
```diff
@@ -37,6 +37,7 @@ pipeline:
         --datadir=/share/cmake \
         --docdir=/share/doc/cmake \
         --system-libs \
+        --no-system-cppdap \
```
kaniini, line 40:
```suggestion
        --no-system-cppdap \
```

It would be good to add a comment explaining why the bundled dependencies are used (since they themselves depend on CMake, creating a cycle).

Elizafox: See [f10be8b](https://github.com/wolfi-dev/os/pull/3810/commits/f10be8b1a40fd146d1839033e216fb51ba1c049d)

kaniini (approved):

PR #3824, opened 2023-07-22
`metacontroller` requires `goreleaser<1.19`

mattmoor:
This is because of this feature deprecation at the latest version tag: https://goreleaser.com/deprecations/#archivesreplacements

### DIFF
--- a/goreleaser-1.18.yaml
+++ b/goreleaser-1.18.yaml
@@ -1,0 +1,27 @@
+package:
+  name: goreleaser-1.18
+  version: 1.18.2
+  epoch: 0
+  description: Deliver Go binaries as fast and easily as possible
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    provides:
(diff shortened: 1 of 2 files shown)


mattmoor:
The error we see is:
```
line 31: field replacements not found in type config.Archive
```

mattmoor:
Ugh, it appears that `wolfictl text` is dumb about the APK package constraint syntax and supports neither `<1.19` or `=1.18.2`, so this is more involved than it should be 😞

kaniini:
Instead of having a build constraint here, goreleaser 1.18 should be packaged as an alternative, and it should depend on goreleaser 1.18 directly.

mattmoor:
@kaniini you mean create a `goreleaser-1.18` package, right?

mattmoor:
@kaniini ok, I added `gorelease-1.18` based on `goreleaser`, and building that and the updated metacontroller also works for me 🤞

ajayk (approved):

PR #3853, opened 2023-07-24
`ruby3.2-validate_url`: unbreak at HEAD.

mattmoor:
This fixes an issue with `ruby3.2-validate_url` where the `git clone {url} .` was failing at HEAD because the directory isn't empty (it has `lost+found`).

To address this, I played around with a few options, but ended up adapting a forked version of our current `git-checkout` pipeline.

This also moves 1.0.15 one commit ahead to let us drop the `sed` command, since they updated the upstream repo with a superset of this change.

### DIFF
--- a/ruby3.2-validate_url.yaml
+++ b/ruby3.2-validate_url.yaml
@@ -1,7 +1,7 @@
 package:
   name: ruby3.2-validate_url
   version: 1.0.15
-  epoch: 0
+  epoch: 1
   description: Library for validating urls in Rails.
   dependencies:
     runtime:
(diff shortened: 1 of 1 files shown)


rawlingsj (dismissed):

mattmoor:
Removing from merge queue because I just had an idea that might clean this up a little bit.

I want to see if it works, and if not let's go with this, but if it does work then I'll be 💯 happier 🤞

mattmoor:
Ok, I switched to `fetch` and I think it is muuuuuch cleaner.

ajayk (approved):

PR #3856, opened 2023-07-24
bazelisk: init at 1.17.0

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/bazelisk.yaml
+++ b/bazelisk.yaml
@@ -1,0 +1,38 @@
+package:
+  name: bazelisk
+  version: 1.17.0
+  epoch: 0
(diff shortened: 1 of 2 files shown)


ajayk (approved):

kaniini:
Why are we packaging this?  We should not use it for our own internal builds using Bazel, instead we should use our bazel packages for that.

kaniini:
Converting to draft until more information is provided as to why this package is needed, as it is most likely not actually needed.

rawlingsj:
Looks like this is not needed and stale.  Closing.

PR #3866, opened 2023-07-24
Pull in the latest `ncurses`

mattmoor:
The previous archive was no longer available, so builds failed.


mattmoor:
I believe the `wolfictl check updates` error is erroneous, but @rawlingsj can check me on that.

ajayk (approved):

mattmoor:
We're going to switch back to the other mirror.  @kaniini is taking this over.

PR #4088, opened 2023-07-31
Refactor node packaging

kaniini:
* Convert `nodejs` into a proper version stream, e.g. `apk add nodejs~19`, will pick nodejs 19, and so on.
* Drop bundled `npm`, replace it with a common npm across all versions that reflects the latest release.
* Add `nodejs-lts` version stream which contains node 16 and 18 at the moment.

### DIFF
--- a/nodejs-18.yaml
+++ b/nodejs-18.yaml
@@ -1,11 +1,13 @@
 package:
   name: nodejs-18
   version: 18.17.0 # When bumping this, also bump the provides below!
-  epoch: 0
+  epoch: 2
   description: "JavaScript runtime built on V8 engine - LTS version"
   copyright:
     - license: MIT
   dependencies:
+    runtime:
+      - npm
     provides:
       - nodejs=18.17.999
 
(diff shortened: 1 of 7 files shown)


npm/npmrc
```diff
@@ -0,0 +1,253 @@
+<!DOCTYPE html>
```
ajayk, line 1: Probably this needs to be replaced ?

kaniini: Yeah, haha, oops.

nodejs-18.yaml
```diff
@@ -1,13 +1,16 @@
 package:
   name: nodejs-18
   version: 18.17.0 # When bumping this, also bump the provides below!
-  epoch: 0
```
ajayk, line 4: epoch should be set to 1 ?

nodejs-18.yaml
```diff
@@ -1,13 +1,16 @@
...
+  epoch: 2
```
ajayk, line 4:
```suggestion
  epoch: 1
```

ajayk (approved):

PR #4106, opened 2023-08-01
remove k3s-embedded and k3s patches in favor of more upstreamy build

joshrwolf:
this refactors our `k3s` build process to remove the confusing notion of `k3s-embedded` and removes our dependency on upstream patches.

the new build process fully leverages all the embedded components that upstream uses (`containerd`, `crictl`, etc...), rather than replacing them at runtime. this ends up having a significant size reduction (~120Mb --> ~60Mb), mainly because of the shared go runtime.

this also has the benefit of only keeping 1 `k3s`, which now functions identical (from an end user perspective) to upstream `k3s`.

the one downside with this approach is it ties us to upstreams release cadence for the various components, which may be slow to react for CVEs. however, we already deal with this with `go get`'s


k3s.yaml
```diff
@@ -118,15 +112,10 @@ subpackages:
...
-              # docker.io/rancher/klipper-helm)
-              #   image=$(crane digest --full-ref cgr.dev/chainguard/klipper-helm:latest)
-              #   ;;
-              # docker.io/rancher/klipper-lb)
-              #   image=$(crane digest --full-ref cgr.dev/chainguard/klipper-lb:latest)
-              #   ;;
-              # docker.io/rancher/local-path-provisioner)
-              #   image=$(crane digest --full-ref cgr.dev/chainguard/local-path-provisioner:latest)
-              #   ;;
+              docker.io/rancher/klipper-helm|docker.io/rancher/klipper-lb|docker.io/rancher/local-path-provisioner|docker.io/rancher/mirrored-pause)
```
joshrwolf, line 115: I pulled this in from https://github.com/wolfi-dev/os/pull/3657, which I will land separately once this new packaging structure lands

imjasonh (approved):

PR #4124, opened 2023-08-01
bump epochs on packages dependent on Go

imjasonh:
Go had a security release: https://github.com/wolfi-dev/os/pull/4121

This bumps packages dependent on Go to rebuild them with this fix.

> * crypto/tls: restrict RSA keys in certificates to <= 8192 bits
>
> Extremely large RSA keys in certificate chains can cause a client/server
to expend significant CPU time verifying signatures. Limit this by
restricting the size of RSA keys transmitted during handshakes to <=
8192 bits.

https://groups.google.com/g/golang-announce/c/X0b6CsSAaYI

### DIFF
--- a/apko.yaml
+++ b/apko.yaml
@@ -2,7 +2,7 @@ package:
   name: apko
   # When bumping the version check if the CVE/GHSA mitigations below can be removed.
   version: 0.10.0
-  epoch: 0
+  epoch: 1
   description: Build OCI images using APK directly without Dockerfile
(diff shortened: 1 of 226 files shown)


ajayk (approved):

kaniini:
There are 200+ packages being rebuilt in this update.  Given that the CI build keeps failing due to network timeouts, I think we are going to want to do this in much smaller batches.

PR #4126, opened 2023-08-01
Remove manual from various packages

Elizafox:
Now that Melange supports var transforms, we don't need to do this for most packages anymore.

Packages left alone at present are openjdk and ncurses. Their version schemes are not supported yet.

### Pre-review Checklist
Pre-review checklist not applicable. No new packages, security issues, or versions are being bumped.

### DIFF
--- a/7zip.yaml
+++ b/7zip.yaml
@@ -1,6 +1,6 @@
 package:
   name: 7zip
-  version: 22.01
+  version: 2201
   epoch: 0
   description: "File archiver with a high compression ratio"
   copyright:
@@ -12,12 +12,6 @@ environment:
       - busybox
       - ca-certificates-bundle
       - build-base
-
-var-transforms:
-  - from: ${{package.version}}
-    match: \.
-    replace: ''
-    to: mangled-package-version
 
 pipeline:
   - uses: fetch
--- a/7zip.yaml
+++ b/7zip.yaml
@@ -53,6 +53,5 @@ subpackages:
 
 update:
   enabled: true
-  manual: true
   release-monitor:
     identifier: 265148
--- a/icu.yaml
+++ b/icu.yaml
@@ -66,6 +66,5 @@ subpackages:
 
 update:
   enabled: true
-  manual: true # melange bump doesn't work with var-transforms yet, manual updates required
   release-monitor:
     identifier: 16134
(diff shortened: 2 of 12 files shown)


Elizafox:
I'm not sure what is with the failure with check updates.

https://github.com/wolfi-dev/os/actions/runs/5732597185/job/15535843750?pr=4126 seems spurious.

7zip.yaml
```diff
@@ -53,6 +53,5 @@ subpackages:
 
 update:
   enabled: true
-  manual: true
```
kaniini, line 56: I think this one needs to stay, but really we could just change the version to remove the separator.

icu.yaml
```diff
@@ -66,6 +66,5 @@ subpackages:
 
 update:
   enabled: true
-  manual: true # melange bump doesn't work with var-transforms yet, manual updates required
```
kaniini, line 69: Can't do this one yet without backwards mapping either.

tcptraceroute.yaml
```diff
@@ -54,4 +54,3 @@ subpackages:
 
 update:
   enabled: false
-  manual: true
```
kaniini, line 57: This one should remain as it is, as there is no update source connected for monitoring.

kaniini (changes requested):

7zip.yaml
Elizafox: > I think this one needs to stay, but really we could just change the version to remove the separator.

Should I implement that now?

icu.yaml
Elizafox: Alright, noted.

7zip.yaml
kaniini: If you want to.  2301 > 23.01 so should be fine.

tcptraceroute.yaml
Elizafox: Noted.

7zip.yaml
Elizafox: Done.

kaniini (approved):

kaniini:
The CI test failure is spurious, and actually due to a bug in wolfictl; overriding.

PR #4144, opened 2023-08-02
go-1.21 rc4 update

ajayk:


go-1.21.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: go-1.21
-  version: 1.21_rc3
-  epoch: 2
+  version: 1.21_rc4
+  epoch: 3
```
luhring, line 4: Do we need the epoch bump here if we're bumping the version?

imjasonh (dismissed):
Are wolfibot updates not working here? 🤔

go-1.21.yaml
ajayk: it makes me think that the `epoch`  should be `0` here

ajayk:
> Are wolfibot updates not working here? 🤔

I Ran the bot but no luck , even the update check fails with no new revision 
@rawlingsj : ^^

kaniini (approved):

PR #4152, opened 2023-08-02
Pure source bootstrap for Go

kaniini:
Like with Java, this PR introduces a pure-source bootstrap for Go.

It works by enabling the `gccgo` frontend in GCC 13, which provides an implementation of Go 1.18.  This implementation is packaged as `gcc-go` and provides `go-bootstrap=1.18`.

Go 1.19 is adjusted to explicitly depend on `go-bootstrap~1.18`, which pins the bootstrap Go to the Go 1.18 implementation provided by GCC 13.  It also now provides `go-bootstrap=1.19`.

Go 1.20 is adjusted to explicitly depend on `go-bootstrap~1.19`, which pins the bootstrap Go to the Go 1.19 package.  It also now provides `go-bootstrap=1.20`.

Go 1.21 is adjusted to explicitly depend on `go-bootstrap~1.20`.  It does not provide `go-bootstrap=1.21` yet, as there is no Go 1.21 final release yet.

The `go-stage0` package is dropped, as it is no longer used.  It was quite out of date, anyway.


imjasonh (approved):

kaniini:
The FIPS variants need to be adjusted to depend on `go~1.19` and `go~1.20` respectively, but this can happen in a followup PR.  At present, they just build with latest go.

joshrwolf (approved):
this is 🔥

PR #4185, opened 2023-08-03
Switch consul to leverage version streams

mattmoor:
This change is largely `s/consul/consul-1.16/` which enables us to track `1.16` directly, but also more easily other versions (1.14 and 1.15 are also both currently supported upstream).

### DIFF
--- a/consul-1.16.yaml
+++ b/consul-1.16.yaml
@@ -1,11 +1,14 @@
 package:
-  name: consul
+  name: consul-1.16
   version: 1.16.0
   # When bumping the version check if the CVE/GHSA mitigations below can be removed.
-  epoch: 1
+  epoch: 2
   description: Consul is a distributed, highly available, and data center aware solution to connect and configure applications across dynamic, distributed infrastructure.
   copyright:
     - license: MPL-2.0
+  dependencies:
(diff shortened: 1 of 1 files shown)


mattmoor:
https://endoflife.date/consul

consul-1.16.yaml
```diff
@@ -1,11 +1,14 @@
 package:
-  name: consul
+  name: consul-1.16
   version: 1.16.0
   # When bumping the version check if the CVE/GHSA mitigations below can be removed.
-  epoch: 1
+  epoch: 2
```
mattmoor, line 5: Wasn't sure if this was needed because we have `consul-1.16.0-r1` already.  Maybe this can be `0` and the `provides` should have `.99`?

kaniini: Yes, the `.999` is needed when transitioning to a version stream, but should otherwise not be done.

mattmoor: Done!

denhamparry:
I've sent an email to Jonas for confirmation. I've cc'ed Matt and Jed to unblock in my GMT absence.

imjasonh (approved):

PR #4202, opened 2023-08-04
gitlab: add `gitlab-runner`, `gitlab-runner-helper` and `glab` packages

marshall007:
For reference these packages are currently available in Alpine as a community package: https://pkgs.alpinelinux.org/packages?name=gitlab-runner

It is also available officially from these repos: https://docs.gitlab.com/runner/install/linux-repository.html

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/gitlab-runner.yaml
+++ b/gitlab-runner.yaml
@@ -1,7 +1,7 @@
 package:
   name: gitlab-runner
   version: 16.2.0
-  epoch: 1
+  epoch: 0
   description: GitLab Runner is the open source project that is used to run your CI/CD jobs and send the results back to GitLab
   copyright:
     - license: MIT
--- a/gitlab-runner.yaml
+++ b/gitlab-runner.yaml
@@ -1,0 +1,86 @@
+package:
+  name: gitlab-runner
+  version: 16.2.0
+  epoch: 1
+  description: GitLab Runner is the open source project that is used to run your CI/CD jobs and send the results back to GitLab
+  copyright:
+    - license: MIT
+
+environment:
+  contents:
+    packages:
+      - ca-certificates-bundle
+      - busybox
+  environment:
+    CGO_ENABLED: "1"
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://gitlab.com/gitlab-org/gitlab-runner
+      tag: v${{package.version}}
+      expected-commit: 782e15dae28e70213e21b29f2fed257d2cf2998c
+
+  # - uses: go/build
+  #   with:
+  #     packages: .
+  #     output: gitlab-runner
+  #     # TODO marshall: there are a few more: https://gitlab.com/gitlab-org/gitlab-runner/-/blob/v16.2.0/Makefile?ref_type=tags#L33-36
+  #     ldflags: -s -w -X gitlab.com/gitlab-org/gitlab-runner/common.NAME=${{package.name}} -X gitlab.com/gitlab-org/gitlab-runner/common.VERSION=v${{package.version}}
+
+subpackages:
+  - name: gitlab-runner-fips
+    description: GitLab Runner is the open source project that is used to run your CI/CD jobs and send the results back to GitLab
+    dependencies:
+      provides:
+        - gitlab-runner
+      runtime:
+        - libcrypto3
+    pipeline:
+      - uses: go-fips/build
+        with:
+          packages: .
+          output: gitlab-runner-fips
+          strict: true
+          ldflags: -s -w -X gitlab.com/gitlab-org/gitlab-runner/common.NAME=${{package.name}} -X gitlab.com/gitlab-org/gitlab-runner/common.VERSION=v${{package.version}}
+      
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/usr/bin
+          mv "${{targets.destdir}}"/usr/bin/gitlab-runner-fips "${{targets.subpkgdir}}"/usr/bin/gitlab-runner
+
+  # - name: gitlab-runner-helper
+  #   description: GitLab Runner Helper
+  #   pipeline:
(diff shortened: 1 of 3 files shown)


marshall007:
Is there a way to combine the build configs for `gitlab-runner` and `gitlab-runner-helper`? They do need to be published as separate packages.

gitlab-runner.yaml
```diff
@@ -0,0 +1,38 @@
+package:
+  name: gitlab-runner
+  version: 16.2.0
+  epoch: 1
+  description: GitLab Runner is the open source project that is used to run your CI/CD jobs and send the results back to GitLab
+  copyright:
+    - license: MIT
```
marshall007, line 7: I think `openssl` is a runtime dependency if you need FIPS enabled?

imjasonh:
> Is there a way to combine the build configs for `gitlab-runner` and `gitlab-runner-helper`? They do need to be published as separate packages.

You can do this with `subpackages`, which are separate pipelines that run after the main pipeline on the same source, and any contents written to `${{targets.subpkgdir}}` end up in the subpackage's package, and not the main package.

There are examples all over the place, for example https://github.com/wolfi-dev/os/blob/be449bc683c36165c23939cb69ec3059d75665dd/gc.yaml#L58

gitlab-runner.yaml
```diff
@@ -0,0 +1,38 @@
...
+  - uses: git-checkout
+    with:
+      repository: https://gitlab.com/gitlab-org/gitlab-runner
+      tag: v${{package.version}}
+      expected-commit: 782e15dae28e70213e21b29f2fed257d2cf2998c
+
+  - uses: go/build
+    with:
+      packages: .
+      tags: fips
```
imjasonh, line 29: To stay consistent with other fipsey packages, we should have separate `gitlab-runner` and `gitlab-runner-fips` packages. You can do this with subpackages too.

gitlab-runner.yaml
imjasonh: You can specify runtime dependencies for this package, or any subpackages, e.g., https://github.com/wolfi-dev/os/blob/be449bc683c36165c23939cb69ec3059d75665dd/gc.yaml#L44

kaniini: The packages would need to be built with `go-fips` to use system OpenSSL, and would need to depend on `libcrypto3` at runtime.

gitlab-runner.yaml
marshall007: @imjasonh and do you generally only include `openssl` as a runtime dependency on the fipsy variants?

marshall007: Answered above: https://github.com/wolfi-dev/os/pull/4202#discussion_r1284749997

gitlab-runner.yaml
marshall007: @kaniini we set `GOEXPERIMENT=boringcrypto` in our own FIPS builds. Is that necessary if I use your `go-fips` package?

marshall007:
> You can do this with `subpackages`, which are separate pipelines that run after the main pipeline on the same source, and any contents written to `${{targets.subpkgdir}}` end up in the subpackage's package, and not the main package.

Thanks! I've tried using `go/build` and various other things in subpackage pipelines before, I guess assuming that `${{targets.subpkgdir}}` would be the default target in that context. I kept getting confused that nothing was in the subpackage.

gitlab-runner.yaml
marshall007: Neat! My fork was behind and I hadn't seen the new inline melange pipeline stuff.

I tried building with the `go-fips/build` pipeline and it looks like we import `crypto/boring` where you are patching it  in as `crypto/internal/boring`.

We perform the `boring.Enabled()` check here: https://gitlab.com/gitlab-org/labkit/-/blob/master/fips/fips.go#L32

kaniini: The `go-fips` package implements a wrapper around OpenSSL which allows the use of the OpenSSL FIPS module.  If Gitlab only supports FIPS via BoringSSL, then we can use `GOEXPERIMENT=boringcrypto` instead, but in general our FIPS plan is centered around OpenSSL and its FIPS provider module.

kaniini:
The `go/build` pipeline is known to have some hardcoded assumptions that should be fixed.  See also chainguard-dev/melange#461 for more about the problems with `go/build`.

gitlab-runner.yaml
marshall007: @kaniini I don't think we need BoringSSL explicitly, we just referenced the module for runtime checks. I have tried to replicate this behavior by compiling with your `goexperiment.strictfipsruntime` build tag instead of our own `fips` tag.

Does this approach make sense? I wasn't able to get the runtime check to fail (i.e. with or without `libcrypto3` specified as a runtime dependency). How can I validate that it is working?

marshall007: I've removed the FIPS builds for now. We can followup on that in a subsequent PR.

marshall007:
@imjasonh @kaniini this should be ready for review. I dropped the FIPS variants for now and will followup on that later.

imjasonh (dismissed):

gitlab-runner.yaml
```diff
@@ -0,0 +1,44 @@
+package:
+  name: gitlab-runner
+  version: 16.2.0
+  epoch: 1
```
ajayk, line 4:
```suggestion
  epoch: 0
```

glab.yaml
```diff
@@ -0,0 +1,31 @@
+package:
+  name: glab
+  version: 1.31.0
+  epoch: 1
```
ajayk, line 4:
```suggestion
  epoch: 0
```

ajayk (approved):

PR #4209, opened 2023-08-04
Add consul-1.15 version stream

mattmoor:
I mirrored the exact commit info where we last built things as `consul` prior to picking up `1.16`, so this should be identical.

#### For new package PRs only


kaniini (approved):

mattmoor:
I'm going to ignore the `1.15.4` bump and let automation stage a PR instead.

PR #4224, opened 2023-08-06
Add llvm16 and clang16

dlorenc:
Fixes:

Related:

### Pre-review Checklist

cc @kaniini what kinda slotting do we need here?

PR #4231, opened 2023-08-07
Update llvm-ldd to 16.

dlorenc:
Fixes:

Related:

#### For version bump PRs

### DIFF
--- a/llvm-lld-15.yaml
+++ b/llvm-lld-15.yaml
@@ -1,10 +1,13 @@
 package:
-  name: llvm-lld
+  name: llvm-lld-15
   version: 15.0.7
   epoch: 0
   description: The LLVM Linker
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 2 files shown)


dlorenc:
Requires https://github.com/wolfi-dev/os/pull/4224

dlorenc:
> I would rather introduce slotting for lld: keeping the LLVM 15 version, and packaging the LLVM 16 version too.

I figured we'd have to do something like that for this one, that's why I separated it.

Any recent slotting PRs I can look at for inspiration?

kaniini:
Looks ok.  You should add llvm-lld-15 as a provider of the slot though:

```
dependencies:
  provides:
    - llvm-lld=15.0.999
```

llvm-lld-16 can just provide `llvm-lld=16`, since it's not superseding a previous version.

kaniini (approved):

PR #4279, opened 2023-08-08
Add kubernetes-csi-node-driver-registrar-compat

jonjohnsonjr:
This fixes an issue with helm charts that hardcode the path of the binary to be in / instead of in /usr/bin.


joshrwolf (approved):

jonjohnsonjr:
More context: https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/main/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver.yaml#L49C34-L49C43

PR #4320, opened 2023-08-09
add single workflow for multiarch HEAD build

joshrwolf:
Use a single workflow for building multiarch with our self hosted runners.

This intentionally isolates the `build` and `upload` step into 2 jobs that run on different workers.

**`build`**

- runs on our self hosted runners
- stores no secrets and requires no GHA permissions
- leverages a RO gcsfuse mount (setup by the self hosted runner) to build missing packages

**`upload`**

- runs on GH hosted runners
- mounts the signing key
- requires GH token permissions to exchange for GCS RW
- syncs, re-indexes/signs, and uploads the repository

This hoists the built apks out of `build -> upload`, which does introduce reliance on GHA to ensure the bits are correct, but seems worth it to me rather than storing keys on our self hosted runners.

This leaves the upload/notification commented out until we can validate the build is WAI. It's also strictly a `workflow_dispatch` action until then.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,176 @@
+name: Build Wolfi OS
+
+on:
+  # TODO: Enable this when we're ready to go live
+  # push:
+  #   branches: ['main']
+  #   paths-ignore:
+  #     - README.md
+
+  workflow_dispatch:
+
+# Only run one build at a time to prevent out of sync signatures
+concurrency:
+  group: build-${{ github.ref }}
+
+jobs:
+  build:
+    name: Build packages
+    if: github.repository == 'wolfi-dev/os'
+
+    strategy:
+      matrix:
+        arch: [ "x86_64", "aarch64" ]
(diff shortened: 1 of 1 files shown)


imjasonh (commented):
If we merge this as-is we'll get two builds of every package right? (Not that that's a bad thing, it might be helpful to debug diffs, just curious)

joshrwolf:
> If we merge this as-is we'll get two builds of every package right? (Not that that's a bad thing, it might be helpful to debug diffs, just curious)

the `build` phase fans out for each arch, so each package is built twice (1 for each arch). the `upload` phase just resigns and uploads, no additional build required

kaniini:
I think if we go this route, we should re-sign the APKs too.

joshrwolf:
> I think if we go this route, we should re-sign the APKs too.

ack, I'll plan on addressing that in the follow ups before this goes live

imjasonh (approved):

PR #4324, opened 2023-08-09
go: drop 1.19 as unsupported

imjasonh:
Go 1.21 was released, which means 1.19 is no longer supported upstream.


amdawson:
Let's hold off before we paywall 1.19

luhring (dismissed):

imjasonh:
Nevermind! 1.19 is needed to bootstrap 1.20, and is otherwise not expected to get updates.

I'll open a separate PR to update thanos and remove updates for 1.19, to avoid confusion.

PR #4344, opened 2023-08-10
trino init at 422

developer-guy:
Fixes:

/cc @johnfosborneiii

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/trino.yaml
+++ b/trino.yaml
@@ -1,0 +1,132 @@
+package:
+  name: trino
+  version: 422
+  epoch: 0
+  description: Official repository of Trino, the distributed SQL query engine for big data, formerly known as PrestoSQL
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - bash # some helper scripts use bash
+      - openjdk-17-default-jvm
+      - python3 # required by trino launcher script
+
+environment:
+  contents:
+    packages:
+      - wolfi-base
+      - busybox
+      - ca-certificates-bundle
+      - curl
+      - maven
+      - build-base
+      - openjdk-17
+      - git
+      - rsync
+      - jvmkill
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/trinodb/trino.git
+      tag: ${{package.version}}
+      expected-commit: 2d1e9d7914234cd7b7af752571e66ced645618fc
+
+  - runs: |
+      export JAVA_HOME=/usr/lib/jvm/java-17-openjdk
+      export LANG=en_US.UTF-8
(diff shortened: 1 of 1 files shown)


trino.yaml
```diff
@@ -0,0 +1,165 @@
...
+      tar -C /home/build/trino-all/ --strip-components 1 -xf core/trino-server/target/trino-server-${{package.version}}.tar.gz
+
+      mvn package -q -DskipTests -T $(nproc) -pl ':trino-cli'
+      chmod +x client/trino-cli/target/trino-cli-${{package.version}}-executable.jar
+
+      # rsync -av --force /home/build/trino-all/plugin/jmx /home/build/plugins/default/
+      # rsync -av --force /home/build/trino-all/plugin/memory /home/build/plugins/default/
+      # rsync -av --force /home/build/trino-all/plugin/tpcds /home/build/plugins/default/
+      # rsync -av --force /home/build/trino-all/plugin/tpch /home/build/plugins/default/
+      # rsync -av --exclude memory --exclude tpcds --exclude tpch --exclude jmx /home/build/trino-all/plugin/* /home/build/plugins/optional/
```
joshrwolf, line 70:
```suggestion
```

can these be removed?

trino.yaml
```diff
@@ -0,0 +1,165 @@
...
+      rsync -av /home/build/trino-all/plugin/* /home/build/plugins/optional/
```
joshrwolf, line 71: just curious why `rsync` instead of `cp`? do these come from some upstream build?

trino.yaml
```diff
@@ -0,0 +1,165 @@
+package:
+  name: trino
+  version: 422
+  epoch: 0
+  description:
```
imjasonh, line 5: Needs a description

trino.yaml
```diff
@@ -0,0 +1,165 @@
...
+  target-architecture:
+    - all
```
imjasonh, line 7:
```suggestion
```

trino.yaml
developer-guy: for sure!

trino.yaml
developer-guy: to be honest, don't know much about that package, maybe @johnfosborneiii can answer that question, also I didn't find any `rsync` command anywhere on that project.

johnfosborneiii: I switched from mv/cp to rsync due to some "rename" errors that I was getting, I ended up resolving the root cause and it was unrelated to mv/cp so you could switch it back if that would be more consistent or preferred.

imjasonh (approved):

PR #4379, opened 2023-08-11
fix trino client

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/trino.yaml
+++ b/trino.yaml
@@ -1,7 +1,7 @@
 package:
   name: trino
   version: "423"
-  epoch: 0
+  epoch: 1
   description: Official repository of Trino, the distributed SQL query engine for big data, formerly known as PrestoSQL
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


joshrwolf (approved):

trino.yaml
```diff
@@ -64,7 +64,7 @@ pipeline:
 
       rsync -av --ignore-existing core/docker/default/etc/* ${{targets.destdir}}/etc/trino/
       rsync -av --force core/docker/bin/* ${{targets.destdir}}/usr/lib/trino/bin/
-      rsync -av --ignore-existing $trino_client ${{targets.destdir}}/usr/bin/
+      rsync -av --ignore-existing client/trino-cli/target/trino-cli-${{package.version}}-executable.jar ${{targets.destdir}}/usr/bin/
```
imjasonh, line 67: Since this updated to 423, this fix won't be applied to 422. Is that expected?

imjasonh (approved):

trino.yaml
developer-guy: yes, this was my bad, I tried smth different and forgot to replace that variable, there is no $trino_client variable anywhere in that file.

imjasonh: Does that mean 422 doesn't have a functioning trino client? Is that something we have to address, or is it fine for now since folks can upgrade to 423 to get a working client?

developer-guy: thanks for asking, I should I say nope, I thought that the package was not working when I first started working on that package, so I tried smth different to make it worked, this is why added smth like trino_client environment variable, that was my bad, then I forgot to change it back, this is not an intended move. There shouldn't be any problem at all.

PR #4383, opened 2023-08-11
support wolfi update bot in k3s

joshrwolf:
Not bumping the epoch since this shouldn't trigger a new build.


kaniini:
This will need the new wolfictl image to pass CI

PR #4410, opened 2023-08-12
fix --modules-path in nginx package

lpcalisi:
This PR fix:

`modules-path` was setted by default on */var/lib/nginx* while module subpackages install modules over */usr/local/nginx/modules*.

When you to add module have this error:

image

`--modules-path` is not taking [$_modules_dir](https://github.com/wolfi-dev/os/blob/main/nginx.yaml#L57) var

image

### DIFF
--- a/nginx.yaml
+++ b/nginx.yaml
@@ -1,7 +1,7 @@
 package:
   name: nginx
-  version: 1.25.1
-  epoch: 1
+  version: 1.25.2
+  epoch: 0
   description: HTTP and reverse proxy server (stable version)
   copyright:
     - license: BSD-2-Clause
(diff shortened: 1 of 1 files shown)


lpcalisi:
@joshrwolf could run again pipeline? I've rebuilt locally for x86_64 and aarch64, and it's works ok.

joshrwolf:
thanks for catching this!

nginx.yaml
```diff
@@ -138,8 +138,6 @@ subpackages:
           mkdir -p ${{targets.subpkgdir}}/usr/lib/nginx/modules
           mkdir -p ${{targets.subpkgdir}}/etc/nginx/modules
 
-          mv ${{targets.destdir}}/ngx_${{range.key}}_module.so ${{targets.subpkgdir}}/usr/lib/nginx/modules/
```
joshrwolf, line 141: does this `mv` need to be removed?

lpcalisi: yes, because in the compilation the modules are generated in `/usr/lib/nginx/modules`

lpcalisi: you are right josh, the mv changes but it's needed. I've push the change.

rawlingsj (approved):

PR #4455, opened 2023-08-15
Move newrelic agent back to stable version

amouat:
The current build doesn't work as the tag doesn't exist. This moves the version _back_ to the current stable version.

### DIFF
--- a/newrelic-infrastructure-agent.yaml
+++ b/newrelic-infrastructure-agent.yaml
@@ -43,4 +43,3 @@ update:
   enabled: true
   github:
     identifier: newrelic/infrastructure-agent
-    use-tag: true
--- a/newrelic-infrastructure-agent.yaml
+++ b/newrelic-infrastructure-agent.yaml
@@ -1,6 +1,6 @@
 package:
   name: newrelic-infrastructure-agent
-  version: 1.46.1
+  version: 1.45.0
   epoch: 0
   description: New Relic Infrastructure Agent
   copyright:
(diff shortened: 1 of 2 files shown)


newrelic-infrastructure-agent.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: newrelic-infrastructure-agent
-  version: 1.46.1
+  version: 1.45.0
```
rawlingsj, line 3: let's keep the current version as apks were built and published

newrelic-infrastructure-agent.yaml
```diff
@@ -20,7 +20,7 @@ pipeline:
     with:
       repository: https://github.com/newrelic/infrastructure-agent
       tag: ${{package.version}}
-      expected-commit: 5f63f460103e9f8a705144ca0f62fc8953de82fb
+      expected-commit: 1cfd539a4796a48638fcb243bc8a713847b1fc6f
```
rawlingsj, line 23: let's keep the current sha as apks were built and published

rawlingsj (changes requested):

newrelic-infrastructure-agent.yaml
rawlingsj: ignore

newrelic-infrastructure-agent.yaml
rawlingsj: ignore

rawlingsj:
we also need to withdraw the 1.46.n packages, I'll push to this branch

rawlingsj (approved):

PR #4472, opened 2023-08-15
py3-certifi: use actual versions and transform when needed

luhring:
This adjusts `py3-certifi` to use _GitHub_ to retrieve the source code, and for package update information.

This means the "correct" package versions (in the form of `2023.07.22`) can now be used with the APK.

This also improves vulnerability scanner results, because some scanners have been sorting `2023.07.22` and `2023.7.22` inconsistently from one another. Vulnerability data is filed with the former version format (e.g. [here](https://github.com/advisories/GHSA-xqr8-7jwr-rhp7)), but the Python installation step "normalizes" the version to the latter version format, which introduces confusion, particularly with scanners.

I also bumped the epoch because of said version comparison confusion.

Before:

```console
$ w scan ./packages/aarch64/py3-certifi-2023.7.22-r0.apk
Will process: py3-certifi-2023.7.22-r0.apk
└── 📄 /usr/lib/python3.11/site-packages/certifi-2023.7.22-py3.11.egg-info/PKG-INFO, /usr/lib/python3.11/site-packages/certifi-2023.7.22-py3.11.egg-info/top_level.txt
        📦 certifi 2023.7.22 (python)
            High CVE-2023-37920 GHSA-xqr8-7jwr-rhp7 fixed in 2023.07.22
```

After:

```console
$ w scan ./packages/aarch64/py3-certifi-2023.07.22-r1.apk
Will process: py3-certifi-2023.07.22-r1.apk
✅ No vulnerabilities found
```

### DIFF
--- a/py3-certifi.yaml
+++ b/py3-certifi.yaml
@@ -49,6 +49,7 @@ pipeline:
   - runs: |
       python3 setup.py install --prefix=/usr --root="${{targets.destdir}}"
 
+      # TODO: Need to update this path when building using python-3.12
       sed -i 's/Version: ${{vars.normalized-package-version}}/Version: ${{package.version}}/' "${{targets.destdir}}"/usr/lib/python3.11/site-packages/certifi-${{vars.normalized-package-version}}-py3.11.egg-info/PKG-INFO
 
   - uses: strip
--- a/py3-certifi.yaml
+++ b/py3-certifi.yaml
@@ -1,7 +1,7 @@
 package:
   name: py3-certifi
-  version: 2023.7.22
-  epoch: 0
+  version: 2023.07.22
+  epoch: 1
   description: "Python3 package for providing Mozilla's CA Bundle"
   copyright:
     - license: MPL-2.0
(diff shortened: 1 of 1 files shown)


py3-certifi.yaml
```diff
@@ -36,9 +49,13 @@ pipeline:
   - runs: |
       python3 setup.py install --prefix=/usr --root="${{targets.destdir}}"
 
+      sed -i 's/Version: ${{vars.normalized-package-version}}/Version: ${{package.version}}/' "${{targets.destdir}}"/usr/lib/python3.11/site-packages/certifi-${{vars.normalized-package-version}}-py3.11.egg-info/PKG-INFO
```
imjasonh, line 52: Would this need to change under py3.12+ ?

imjasonh: I don't have any ideas, especially not any good ones.

Thinking it through, when 3.12 is released we won't automatically rebuild this package to use it, so we won't notice until its epoch gets bumped later, which will fail pretty mysteriously.

Can you add a comment like `TODO: Need to update this path when building using python-3.12`? That'd be enough of a breadcrumb I think.

jonjohnsonjr (approved):

PR #4508, opened 2023-08-17
remove runtime dependency on ca-certificates-bundle.

vaikas:
There is no reason to have the ca-certificates-bundle as a runtime dependency, so remove it.

Fixes:

Related:

### DIFF
--- a/alpine-keys.yaml
+++ b/alpine-keys.yaml
@@ -1,13 +1,10 @@
 package:
   name: alpine-keys
   version: 2.4
-  epoch: 1
+  epoch: 2
   description: Public keys for Alpine Linux packages
   copyright:
     - license: MIT
-  options:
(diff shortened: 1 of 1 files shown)


alpine-keys.yaml
```diff
@@ -5,9 +5,9 @@ package:
   description: Public keys for Alpine Linux packages
   copyright:
     - license: MIT
-  dependencies:
-    runtime:
-      - ca-certificates-bundle
+  options:
+    # There are no executables in this package.
```
dlorenc, line 9: I don't think you need to add this section. IIUC, melange manages these fields in the built APK and APK index automatically.

This will instruct melange not to search for commands, the end result should be the same but the default behavior in melange would be safer.

dlorenc (approved):

PR #4512, opened 2023-08-17
just-1.14.0: new package

### LINKED ISSUE #4511: [Wolfi Package Request]: Just

Requesting [Just](https://github.com/casey/just), a command-runner and rough Make equivalent.

1.14.0

I use Just in all my projects to provide useful entrypoints for things like build, test, melange, apko.

I'd like my melange builds to use the same build command that eg GH actions does, so I'm standardising everything to use one Just entrypoint. This means running Just under Melange (yes, I use wolfi as my melange os).

This should be simple: I currently install rust and have melange do `cargo install just` as its first step.
The only issue with this is that Rust builds everything from source, so this is slow, especially in throttled environments like GH actions.

- [X] The package/versions proposed are actively maintained upstream
- [X] I am interested in adding this package to Wolfi OS myself
- [ ] I am willing to help maintain this package

mt-inside:
Fixes: #4511 

Related:

#### For new package PRs only

### DIFF
--- a/just.yaml
+++ b/just.yaml
@@ -1,0 +1,36 @@
+package:
+  name: just
+  version: 1.14.0
+  epoch: 0
+  description: "just is a handy way to save and run project-specific commands."
+  copyright:
+    - license: CC0
+
+environment:
+  contents:
+    packages:
+      - rust
+      - wolfi-base
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+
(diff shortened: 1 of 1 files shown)


mt-inside:
The failing check seems to be checking github.com/gradle/gradle, not just?

ajayk (approved):

PR #4522, opened 2023-08-17
alphabetize runtime / build deps for [a-c]*yaml.

vaikas:
As discussed in: https://github.com/chainguard-dev/yam/issues/27

run the following mechanical cleanup for packages sorting for [a-c]*yaml:
```
yam --sort .environment.contents.packages
yam --sort .package.dependencies.runtime
```

No functional changes, only cosmetic.

PR #4526, opened 2023-08-17
alphabetize a*yaml deps.

vaikas:
As discussed in: https://github.com/chainguard-dev/yam/issues/27

run the following mechanical cleanup for packages sorting for [a-c]*yaml:

```
yam --sort .environment.contents.packages
yam --sort .package.dependencies.runtime
```

No functional changes, only cosmetic.

Trying to see if having too many changes is causing grief in #4522

### DIFF
--- a/acl.yaml
+++ b/acl.yaml
@@ -9,11 +9,11 @@ package:
 environment:
   contents:
     packages:
-      - wolfi-base
+      - attr-dev
+      - build-base
       - busybox
       - ca-certificates-bundle
-      - build-base
-      - attr-dev
+      - wolfi-base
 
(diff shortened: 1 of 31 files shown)


rawlingsj:
Not too sure what to do about this one. Reckon it's enough to let yam fix these up over time as updates are made to packages?

I wonder... maybe we could add yam when the update bot finds new versions and creates a PR?

Or we try and do a big update but we'd need to plan it so our infra is able to handle the load?

joshrwolf:
I'm going to close this as is since it's baked into `yam` now, and let us naturally roll forward as packages update 🙏

PR #4536, opened 2023-08-17
enable it on pushes to HEAD without actually publishing anything yet

joshrwolf:
enables the workflow to run on HEAD, but doesn't publish anything yet.

plan is to let this run for a while now that it's ~stable and see how we fare.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,11 +1,11 @@
 name: Build Wolfi OS
 
 on:
-  # TODO: Enable this when we're ready to go live
-  # push:
-  #   branches: ['main']
-  #   paths-ignore:
-  #     - README.md
+  push:
+    branches: ['main']
+    paths-ignore:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -1,11 +1,10 @@
...
 on:
-  # TODO: Enable this when we're ready to go live
-  # push:
-  #   branches: ['main']
-  #   paths-ignore:
-  #     - README.md
+  push:
+    branches: ['main']
+    paths-ignore:
+      - README.md
```
imjasonh, line 7: Can we also ignore withdrawn-packages 🙏

imjasonh (dismissed):

.github/workflows/build.yaml
joshrwolf: done, with one better!

imjasonh (approved):

PR #4551, opened 2023-08-18
kubernetes: use our Go to build

imjasonh:
K8s prefers to control the version of Go used to build it, and keeps track of that in `.go-version` in the root of their repo.

If the current Go version doesn't match that version, `make` will download and install the required version of Go. We don't want this, since it means any fixes we apply to our Go packages won't be used -- the upstream Go will be used instead.

This change sets `FORCE_HOST_GO=true` so that our version is used, and pins the version used to the one used upstream, to avoid inconsistencies and surprises. If we're feeling 🕺 saucy 💃 we can try to use a newer Go than upstream, but it might lead to ✨ surprises ✨ later.

This change is not expected to have any real changes -- we'd been building each version with the same Go version anyway, we'd just been doing it with the upstream `go` and not ours.

cc @eddiezane @joshrwolf

### DIFF
--- a/kubernetes-1.24.yaml
+++ b/kubernetes-1.24.yaml
@@ -13,7 +13,7 @@ environment:
       - bash
       - ca-certificates-bundle
       - build-base
-      - go
+      - go=1.20.6 # Pin to the version used upstream: https://github.com/kubernetes/kubernetes/blob/v1.24.16/.go-version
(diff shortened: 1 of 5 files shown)


joshrwolf (approved):
A+ for emoji's 🎉

PR #4557, opened 2023-08-18
katib python packages

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/py3-cloudpickle.yaml
+++ b/py3-cloudpickle.yaml
@@ -1,0 +1,45 @@
+package:
+  name: py3-cloudpickle
+  version: 2.2.1
+  epoch: 0
+  description: Extended pickling support for Python objects
+  copyright:
+    - license: BSD 3-Clause License
+  dependencies:
+    runtime:
+      - python3
+
+environment:
+  contents:
+    packages:
+      - ca-certificates-bundle
+      - wolfi-base
+      - busybox
+      - build-base
+      - python3
+      - py3-setuptools
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/cloudpipe/cloudpickle
+      expected-commit: 917bed6bc63355f50aeb4f001683be7f172b6b88
+      tag: v${{package.version}}
+
+  - name: Python Build
+    runs: python setup.py build
+
+  - name: Python Install
+    runs: python setup.py install --prefix=/usr --root="${{targets.destdir}}"
+
+  - uses: strip
+
+update:
+  ignore-regex-patterns:
+    - ^\.
+  enabled: true
+  github:
+    identifier: cloudpipe/cloudpickle
+    strip-prefix: v
+    use-tag: true
+    tag-filter: v
--- a/py3-snowballstemmer.yaml
+++ b/py3-snowballstemmer.yaml
@@ -1,0 +1,39 @@
+package:
+  name: py3-snowballstemmer
+  version: 2.2.0
+  epoch: 0
+  description: This package provides 29 stemmers for 28 languages generated from Snowball algorithms.
+  copyright:
+    - license: BSD-3-Clause
+  dependencies:
+    runtime:
+      - python3
+
+environment:
+  contents:
+    packages:
+      - ca-certificates-bundle
+      - wolfi-base
+      - busybox
+      - build-base
+      - python3
+      - py3-setuptools
+
+pipeline:
+  - uses: fetch
+    with:
+      expected-sha256: 09b16deb8547d3412ad7b590689584cd0fe25ec8db3be37788be3810cbf19cb1
+      uri: https://files.pythonhosted.org/packages/44/7b/af302bebf22c749c56c9c3e8ae13190b5b5db37a33d9068652e8f73b7089/snowballstemmer-${{package.version}}.tar.gz
+
+  - name: Python Build
+    runs: python setup.py build
+
+  - name: Python Install
+    runs: python setup.py install --prefix=/usr --root="${{targets.destdir}}"
+
+  - uses: strip
+
+update:
+  enabled: true
+  release-monitor:
+    identifier: 7853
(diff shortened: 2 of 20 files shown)


py3-cloudpickle.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - wolfi-base
+      - busybox
+      - build-base
+      - python3
+      - py3-setuptools
+
+pipeline:
+  - uses: fetch
+    with:
+      README: 'CONFIRM WITH: curl -L https://files.pythonhosted.org/packages/5f/51/913ecca3970a2227cf4d5e8937df52cc28f465ac442216110b8e3323262d/cloudpickle-2.2.1.tar.gz | sha256sum'
```
rawlingsj, line 26: I think we can remove all the `README: 'CONFIRM WITH: curl -L ...` lines in this PR, they're not needed

rawlingsj: Other than that I think it looks good!

py3-cloudpickle.yaml
```diff
@@ -0,0 +1,41 @@
+# Generated from https://pypi.org/project/cloudpickle/
```
rawlingsj, line 1: not sure if we need these at the top of each either, not a blocker but feels like we we're tidying up our configs and making them more consistent.  So maybe drop all the `# Generated from...`

py3-cloudpickle.yaml
developer-guy: done! thank you!

py3-cloudpickle.yaml
developer-guy: removed!

rawlingsj (changes requested):
similar to https://github.com/wolfi-dev/os/pull/4558#issuecomment-1684480938 I just noticed that there's sha's in the fetch URL.

I'm sorry for the extra work but I think these need to be git-checkouts and have the update config track github rather than release monitor

developer-guy:
> similar to [#4558 (comment)](https://github.com/wolfi-dev/os/pull/4558#issuecomment-1684480938) I just noticed that there's sha's in the fetch URL.
> 
> I'm sorry for the extra work but I think these need to be git-checkouts and have the update config track github rather than release monitor

Yes, I tried using `git-checkout` for those who work with `git-checkout` but some of them didn't work, so I kept using fetch from `files.pythonhosted.org` since APKBUILD does the same thing.

rawlingsj:
> Yes, I tried using git-checkout for those who work with git-checkout but some of them didn't work, so I kept using fetch from files.pythonhosted.org since APKBUILD does the same thing.

do you have more details of what wasn't working?  I just tried here switching the first two packages in this PR to github and the builds worked.  I.e.

```yaml
package:
  name: py3-cloudpickle
  version: 2.2.1
  epoch: 0
  description: Extended pickling support for Python objects
  copyright:
    - license: BSD 3-Clause License
  dependencies:
    runtime:
      - python3

environment:
  contents:
    packages:
      - ca-certificates-bundle
      - wolfi-base
      - busybox
      - build-base
      - python3
      - py3-setuptools

pipeline:
  - uses: git-checkout
    with:
      repository: https://github.com/cloudpipe/cloudpickle
      expected-commit: 917bed6bc63355f50aeb4f001683be7f172b6b88
      tag: v${{package.version}}

  - name: Python Build
    runs: python setup.py build

  - name: Python Install
    runs: python setup.py install --prefix=/usr --root="${{targets.destdir}}"

  - uses: strip

update:
  ignore-regex-patterns:
    - ^\.
  enabled: true
  github:
    identifier: cloudpipe/cloudpickle
    strip-prefix: v
    use-tag: true
    tag-filter: v
```

and 

```yaml
package:
  name: py3-future
  version: 0.18.3
  epoch: 0
  description: Clean single-source support for Python 3 and 2
  copyright:
    - license: MIT
  dependencies:
    runtime:
      - python3

environment:
  contents:
    packages:
      - ca-certificates-bundle
      - wolfi-base
      - busybox
      - build-base
      - python3
      - py3-setuptools

pipeline:
  - uses: git-checkout
    with:
      repository: https://github.com/PythonCharmers/python-future
      expected-commit: af1db970b0879b59e7aeb798c27a623144561cff
      tag: v${{package.version}}

  - name: Python Build
    runs: python setup.py build

  - name: Python Install
    runs: python setup.py install --prefix=/usr --root="${{targets.destdir}}"

  - uses: strip

update:
  enabled: true
  github:
    identifier: PythonCharmers/python-future
    strip-prefix: v
```

> since APKBUILD does the same thing.

The problem here is that we use different tooling to alpine, using APKBUILDs as a reference can help but we don't always want to do the exact same thing as wolfi is different.

py3-snowballstemmer.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - busybox
+      - build-base
+      - python3
+      - py3-setuptools
+
+pipeline:
+  - uses: fetch
+    with:
+      expected-sha256: 09b16deb8547d3412ad7b590689584cd0fe25ec8db3be37788be3810cbf19cb1
+      uri: https://files.pythonhosted.org/packages/44/7b/af302bebf22c749c56c9c3e8ae13190b5b5db37a33d9068652e8f73b7089/snowballstemmer-${{package.version}}.tar.gz
```
rawlingsj, line 26: can this also be a `git-checkout` https://github.com/snowballstem/snowball/releases/tag/v2.2.0?

developer-guy: I tried but the "setup.py" does not exist on GitHub source but it exists on files.pythonhosted source, so, I couldn't understand how to build with by using GitHub source

kaniini: Lets stick with the tarball for now.

rawlingsj (approved):

PR #4558, opened 2023-08-18
Add some deps of metadata-writer and kubernetes

Dentrax:
This PR adds:

- py3-absl-py.yaml
- py3-asn1-modules.yaml
- py3-blinker.yaml
- py3-cachetools.yaml
- py3-lru-dict.yaml
- py3-oauthlib.yaml
- py3-protobuf.yaml
- py3-pyjwt.yaml
- py3-websocket-client.yaml

Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/py3-absl-py.yaml
+++ b/py3-absl-py.yaml
@@ -1,0 +1,41 @@
+package:
+  name: py3-absl-py
+  version: 1.4.0
+  epoch: 0
+  description: Abseil Python Common Libraries, see https://github.com/abseil/abseil-py.
+  copyright:
+    - license: Apache 2.0
+  dependencies:
+    runtime:
+      - python3
+
+environment:
+  contents:
+    packages:
+      - ca-certificates-bundle
+      - wolfi-base
+      - busybox
+      - build-base
+      - python3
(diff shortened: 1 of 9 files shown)


rawlingsj:
I’m guessing these were generated using melange convert python ?  That’s something that was added after the initial apkbuild convert and I think there’s a little issue with it.
The generated URIs? i.e [here](https://github.com/wolfi-dev/os/pull/4558/files#diff-47f21455e84b1af2f0b1bbb1d06ae5c01f249f21e5384853555c3f585583eedbR28) contains a sha which means the wolfi bot wont be able to auto update them. 

Most if not all projects are on GitHub i.e. [this](https://github.com/websocket-client/websocket-client)  and so for the update bot to be able to maintain these new packages I think we need to switch them to git-checkout  and have the update config use github rather than release monitor

rawlingsj (changes requested):
comments above, we need to avoid sha's in the download URI, we can use git-checkout instead

rawlingsj (approved):

PR #4592, opened 2023-08-20
Fix the provides of the consul subpackages

mattmoor:
We did this for the main package's provides already, but not for the subpackages, so I see us pulling in the older entrypoint in our images:
```
-rw-r--r--  0 root   root      2098 Dec 31  1969 var/lib/db/sbom/consul-1.16-1.16.1-r1.spdx.json
-rw-r--r--  0 root   root      2239 Dec 31  1969 var/lib/db/sbom/consul-oci-entrypoint-1.16.0-r1.spdx.json
-rw-r--r--  0 root   root      1305 Dec 31  1969 var/lib/db/sbom/consul-oci-entrypoint-compat-1.16.0-r1.spdx.json
```

### DIFF
--- a/consul-1.15.yaml
+++ b/consul-1.15.yaml
@@ -2,7 +2,7 @@ package:
   name: consul-1.15
   version: 1.15.5
   # When bumping the version check if the CVE/GHSA mitigations below can be removed.
-  epoch: 1
+  epoch: 2
   description: Consul is a distributed, highly available, and data center aware solution to connect and configure applications across dynamic, distributed infrastructure.
   copyright:
     - license: MPL-2.0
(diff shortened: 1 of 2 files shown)


mattmoor:
FWIW I think this is why our `consul:latest` image is reporting https://nvd.nist.gov/vuln/detail/CVE-2023-3518 despite having 1.16.1 of the `consul` package itself.

mattmoor:
Scratch that, syft is stating that this is from `/bin/consul` itself, so I'm not entirely sure why this is showing up.  Let me also make sure I'm on the latest `grype` 🤞

mattmoor:
I _**think**_ the CVE being reported was a bug in `syft`'s classification logic, which just hasn't yet made it into `grype`. 🤞

jonjohnsonjr (approved):

PR #4600, opened 2023-08-21
add scipy deps

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/py3-gast.yaml
+++ b/py3-gast.yaml
@@ -1,0 +1,41 @@
+package:
+  name: py3-gast
+  version: 0.5.4
+  epoch: 0
+  description: Python AST that abstracts the underlying Python version
+  copyright:
+    - license: BSD 3-Clause
+  dependencies:
+    runtime:
+      - python3
+
+environment:
+  contents:
+    packages:
+      - ca-certificates-bundle
+      - wolfi-base
+      - busybox
+      - build-base
+      - python3
+      - py3-setuptools
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/serge-sans-paille/gast
+      tag: ${{package.version}}
+      expected-commit: 1f872364a9964bdb739c4f76d0be463104cc4757
+
+  - name: Python Build
+    runs: python setup.py build
+
+  - name: Python Install
+    runs: python setup.py install --prefix=/usr --root="${{targets.destdir}}"
+
+  - uses: strip
+
+update:
+  enabled: true
+  github:
+    identifier: serge-sans-paille/gast
+    use-tag: true
--- a/py3-ply.yaml
+++ b/py3-ply.yaml
@@ -1,0 +1,45 @@
+package:
+  name: py3-ply
+  version: 3.11_git20180215
+  epoch: 0
+  description: Python Lex & Yacc
+  copyright:
+    - license: BSD
+  dependencies:
+    runtime:
+      - python3
+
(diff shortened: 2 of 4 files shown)


py3-gast.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+  - name: Python Build
+    runs: python setup.py build
+
+  - name: Python Install
+    runs: python setup.py install --prefix=/usr --root="${{targets.destdir}}"
+
+  - uses: strip
+
+# no versions returned from github graphql api
```
rawlingsj, line 37: I can see the tags matching the version [here](https://github.com/serge-sans-paille/gast/releases/tag/0.5.4) 

so we should be able to use
```
update:
  enabled: true
  github:
    identifier: serge-sans-paille/gast
    use-tag: true
```
which I validated using 

```
wolfictl check update  --override-version 0.0.1 py3-gast
```

py3-ply.yaml
```diff
@@ -0,0 +1,39 @@
...
+    runs: python setup.py build
+
+  - name: Python Install
+    runs: python setup.py install --prefix=/usr --root="${{targets.destdir}}"
+
+  - uses: strip
+
+# some sort of a deprecated project
+update:
+  enabled: false
```
rawlingsj, line 39: this and [related noticed](https://github.com/dabeaz/ply#important-notice---october-27-2022) may be a problem as we can only add maintained projects to wolfi

py3-pythran.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+  - name: Python Build
+    runs: python setup.py build
+
+  - name: Python Install
+    runs: python setup.py install --prefix=/usr --root="${{targets.destdir}}"
+
+  - uses: strip
+
+# no versions returned from github graphql api
```
rawlingsj, line 41: we should be able to use github 

```
update:
  ignore-regex-patterns:
    - .post1
  enabled: true
  github:
    identifier: serge-sans-paille/pythran
    use-tag: true
```
validated using `wolfictl check update  --override-version 0.0.1 py3-pythran`

rawlingsj (dismissed):

py3-ply.yaml
developer-guy: it might be yes but some packages depend on it "py3-pythran" to be precise

py3-gast.yaml
developer-guy: I didn't use `use-tag`, let me try that again with this option.

developer-guy: thanks James!

py3-ply.yaml
kaniini: We can package it, but we should indicate the date of the git snapshots we use.

developer-guy: thanks, I've updated it as `3.11_git20180215`

py3-ply.yaml
```diff
@@ -0,0 +1,45 @@
...
+  description: Python Lex & Yacc
+  copyright:
+    - license: BSD
+  dependencies:
+    runtime:
+      - python3
+
+var-transforms:
+  - from: ${{package.version}}
+    match: _git20180215
```
kaniini, line 14:
```suggestion
    match: _git(.*)
```

kaniini (approved):

kaniini (approved):

PR #4614, opened 2023-08-21
add a compat package to cover our helm chart basis

joshrwolf:
https://artifacthub.io/packages/helm/cluster-autoscaler/cluster-autoscaler

vs

https://github.com/kubernetes-sigs/cluster-proportional-autoscaler/tree/master/charts/cluster-proportional-autoscaler

which are different for some reason 🤷

### DIFF
--- a/cluster-proportional-autoscaler.yaml
+++ b/cluster-proportional-autoscaler.yaml
@@ -1,7 +1,7 @@
 package:
   name: cluster-proportional-autoscaler
   version: 1.8.9
-  epoch: 1
+  epoch: 2
   description: Kubernetes Cluster Proportional Autoscaler Container
   copyright:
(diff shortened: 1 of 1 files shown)


cluster-proportional-autoscaler.yaml
```diff
@@ -30,6 +30,17 @@ pipeline:
 
   - uses: strip
 
+subpackages:
+  - name: ${{package.name}}-compat
+    pipeline:
+      - runs: |
```
kaniini, line 36: You can get rid of the `mkdir -p` by doing:

```suggestion
      - working-directory: ${{targets.subpkgdir}}
        runs: |
```

ajayk (approved):

PR #4617, opened 2023-08-21
add scipy

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/py3-scipy.yaml
+++ b/py3-scipy.yaml
@@ -1,0 +1,53 @@
+package:
+  name: py3-scipy
+  version: 1.11.2
+  epoch: 0
+  description: Fundamental algorithms for scientific computing in Python
+  copyright:
+    - license: BSD-3-Clause
+  dependencies:
+    runtime:
+      - numpy
+      - python3
+
+environment:
+  contents:
+    packages:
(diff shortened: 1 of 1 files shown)


py3-scipy.yaml
```diff
@@ -0,0 +1,53 @@
+package:
+  name: py3-scipy
+  version: 1.11.2
+  epoch: 0
+  description: Fundamental algorithms for scientific computing in Python
+  copyright:
+    - license: BSD-3-Clause
+  dependencies:
+    runtime:
+      - numpy
```
imjasonh, line 10: Does this need py3-pythran at runtime too?

imjasonh (approved):

py3-scipy.yaml
developer-guy: don't know much about the package but I used `melange convert python` to generate the configuration file, it didn't add that package to runtime deps

imjasonh: That might just be a bug in melange though. We'll find out when we make an image out of this whether we need it as a runtime dep too.

PR #4619, opened 2023-08-21
Chase melange ${{targets.contextdir}} changes

kaniini:
Some packages used `mv` to move files into the subpkgdir after building them.  Now that we have ${{targets.contextdir}}, this extra `mv` step is no longer needed: the go/build pipeline now follows the context dir instead.


kaniini:
These CI results are to be expected, as they depend on the new Melange changes

rawlingsj (approved):

PR #4625, opened 2023-08-21
Create 1.26 and 1.27 version streams for telegraf

mattmoor:
#### For new package PRs only

### DIFF
--- a/telegraf-1.26.yaml
+++ b/telegraf-1.26.yaml
@@ -1,7 +1,7 @@
 package:
   name: telegraf-1.26
   version: 1.26.3
-  epoch: 1
+  epoch: 0
   description:
   copyright:
     - license: MIT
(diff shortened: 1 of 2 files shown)


telegraf-1.26.yaml
```diff
@@ -0,0 +1,55 @@
+package:
+  name: telegraf-1.26
+  version: 1.26.3
+  epoch: 1
```
ajayk, line 4:
```suggestion
  epoch: 0
```

ajayk (dismissed):

telegraf-1.26.yaml
ajayk, line 4: Does the epoch needs to be set to `0` as this is a considered as a new package

mattmoor, line 4:
```suggestion
  epoch: 0
```

mattmoor: I think you are right as this is effectively a new package, but I used the original epoch out of an abundance of caution.  Changed to zero now.

ajayk (approved):

PR #4710, opened 2023-08-22
Switch vault to version streams

mattmoor:
🚨 using CI to test 1.14 before I add a stream for 1.13


mattmoor:
Added 1.13 at the last commit/tag we had it.  It looks like 1.12 and 1.13 are both still supported upstream: https://endoflife.date/hashicorp-vault

rawlingsj (approved):

PR #4743, opened 2023-08-23
adds wildfly-openssl

mritunjaysharma394:
Adds OpenSSL-based Java Secure Socket Extension implementation

Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs


wildfly-openssl.yaml
```diff
@@ -0,0 +1,51 @@
...
+          expected-commit: 9e58cff606974ec33918b7874951364fe02c58e5
+      - runs: |
+          mvn package
+
+  # Do something with the WildFly JAR, located at /home/build/final/combined/target/wildfly-openssl-${{package.version}}.Final.jar
+  - runs: |
+      ls -alh /home/build/final/combined/target/wildfly-openssl-${{package.version}}.Final.jar
+
+update:
+  enabled: true
```
kaniini, line 48: Since there are two unrelated source artifacts that do not use the same version, we need to manually handle updates here.

mritunjaysharma394: Thanks a lot @kaniini just disabled it

wildfly-openssl.yaml
```diff
@@ -0,0 +1,52 @@
...
+        with:
+          repository: https://github.com/wildfly-security/wildfly-openssl
+          tag: 2.2.5.Final
+          expected-commit: 9e58cff606974ec33918b7874951364fe02c58e5
+      - runs: |
+          mvn package
+
+  # Do something with the WildFly JAR, located at /home/build/final/combined/target/wildfly-openssl-${{package.version}}.Final.jar
+  - runs: |
+      ls -alh /home/build/final/combined/target/wildfly-openssl-${{package.version}}.Final.jar
```
stormqueen1990, line 45: Should this line place the final jar at some system path like `${{targets.destdir}}/usr/share/java/` perhaps?

kaniini (approved):

PR #4866, opened 2023-08-28
Add xmlsec

johnfosborneiii:
#### For new package PRs only

### DIFF
--- a/xmlsec.yaml
+++ b/xmlsec.yaml
@@ -96,3 +96,4 @@ update:
     identifier: lsh123/xmlsec
     strip-prefix: xmlsec_
     use-tag: true
+    tag-filter: xmlsec_
--- a/xmlsec.yaml
+++ b/xmlsec.yaml
@@ -94,6 +94,5 @@ update:
   enabled: true
   github:
     identifier: lsh123/xmlsec
-    strip-prefix: v
+    strip-prefix: xmlsec_
     use-tag: true
-    tag-filter: v
--- a/xmlsec.yaml
+++ b/xmlsec.yaml
@@ -92,6 +92,7 @@ subpackages:
 
 update:
   enabled: true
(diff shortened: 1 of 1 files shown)


dlorenc:
I think you have to disable the update check here - we don't really support the mangled-package-version stuff in that direction, it's getting confused. cc @rawlingsj

xmlsec.yaml
```diff
@@ -0,0 +1,99 @@
...
+      runtime:
+    description: xmlsec dev
+
+  - name: xmlsec-doc
+    pipeline:
+      - uses: split/manpages
+    description: xmlsec manpages
+
+update:
+  enabled: true
```
rawlingsj, line 94:
```suggestion
  enabled: true
  version-separator: _
```
Adding this should work.  I just did a quick check using
```
wolfictl check update  --override-version 0.0.1 xmlsec.yaml
```

rawlingsj: I did notice that there's a new `version-transform` config available [example](https://github.com/wolfi-dev/os/blob/0ca90c76b501d45c35e13e74049a54c18c39bfb3/openjdk-20.yaml#L221C1-L223C19),  in the update config but I don't think it's wired up correctly yet. I can help do that early next week unless it's already being looked at.  Once that does work we can switch all places that use `version-separator` config to `version-transform`.  

For now the suggestion above should make CI green and mean auto updates work.

kaniini: `version-transform` should be working fine.  What issues are you seeing with it?

kaniini (approved):

kaniini:
> I think you have to disable the update check here - we don't really support the mangled-package-version stuff in that direction, it's getting confused. cc @rawlingsj

We do, with `version-transform`

PR #4867, opened 2023-08-28
place the multicall links in $PATH

joshrwolf:
the packaged multicalls were previously in the k3s `data` dir, which were unpacked into `/var/lib/rancher/k3s/data`.

this change places them instead directly in `/bin/$i`, so that callers can access them directly without adjusting their $PATH.

worth noting that accessing them via the outer `k3s` (`k3s kubectl ...`) will still work

### DIFF
--- a/k3s.yaml
+++ b/k3s.yaml
@@ -1,7 +1,7 @@
 package:
   name: k3s
   version: 1.27.4
-  epoch: 3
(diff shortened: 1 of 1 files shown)


joshrwolf:
nevermind, this means `k3s` conflicts with `containerd` which I don't really like

PR #4876, opened 2023-08-28
Add ollama cli tool.

dlorenc:
Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/ollama.yaml
+++ b/ollama.yaml
@@ -1,0 +1,37 @@
+package:
+  name: ollama
+  version: 0.0.16
+  epoch: 0
+  description: Get up and running with Llama 2 and other large language models locally
+  copyright:
+    - license: MIT
+
+environment:
+  contents:
+    packages:
+      - ca-certificates-bundle
+      - busybox
+      - build-base
+      - go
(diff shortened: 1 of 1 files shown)


ollama.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/jmorganca/ollama
+      tag: v${{package.version}}
+      expected-commit: 177b69a211755537dc9bb8fe548ba3f596e8e3c5
+
+  - runs: |
+      CGO_ENABLED=1 go build -ldflags '-linkmode external -extldflags "-static"' .
```
kaniini, line 25: I assume we don't use `go/build` here because we want CGO?  Maybe we should change `go/build` to allow cgo to be enabled?

dlorenc: Yeah it can do ldflags but not env vars (that I know of)

dlorenc: or extldflags

kaniini (approved):

PR #4877, opened 2023-08-28
Add chroma package.

dlorenc:
Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/chroma.yaml
+++ b/chroma.yaml
@@ -1,0 +1,47 @@
+package:
+  name: chroma
+  version: 0.4.7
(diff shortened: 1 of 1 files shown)


kaniini (changes requested):
This is pulling in vendored dependencies from PyPI, which makes it impossible for us to service CVEs in the dependencies.

The `no-provides` option was set to avoid conflicts at the package manager level, but that isn't the correct fix here.  Instead, we need to avoid vendoring the dependencies.

PR #4885, opened 2023-08-28
 redis: add version stream and move to subpackages

ajayk:
- adds version streams for redis-7 and moves the cli , benchmark and utilities as sub-packages


redis.yaml
```diff
@@ -37,6 +40,64 @@ pipeline:
...
+  - name: redis-cli
+    pipeline:
+      - uses: split/dev
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/usr/bin
+          mv "${{targets.destdir}}"/usr/bin/redis-cli "${{targets.subpkgdir}}"/usr/bin/redis-cli
+    description: redis-cli is the command line interface utility to talk with Redis.
+    dependencies:
+      provides:
+        - redis-cli=7.2.99
```
mattmoor, line 53: I don't think we want these?

redis-7.0/0000-Disable-protected-mode.patch
```diff
@@ -0,0 +1,29 @@
+From b91e39701c7a71f229be34dcf078579d37367436 Mon Sep 17 00:00:00 2001
```
mattmoor, line 1: If these are all the same, then we can create it once and use directory symlinks for each of the copies.

mattmoor (approved):

PR #4896, opened 2023-08-28
Convert nginx into version streams

kaniini:
This introduces two version streams, `nginx-stable` and `nginx-mainline`, which track the aforementioned supported branches of nginx.

Reopened because Github is glitching out.


nginx-mainline.yaml
```diff
@@ -109,11 +112,13 @@ pipeline:
...
 subpackages:
-  - name: nginx-package-config
+  - name: nginx-mainline-package-config
     description: Config supplied in packaged nginx.org bundles at https://hg.nginx.org/pkg-oss/
     dependencies:
       runtime:
-        - nginx
+        - nginx-mainline
+      provides:
+        - nginx-package-config=${{package.version}}-r${{package.epoch}}
```
ajayk, line 121:
```suggestion
        - nginx-mainline-package-config=${{package.version}}-r${{package.epoch}}
```

ajayk (approved):

PR #4902, opened 2023-08-29
cmake: rebuild for rhash-1.4.4 upgrade

kaniini:
rhash SONAME changed, thus we need to rebuild cmake


dlorenc (approved):

dlorenc:
Failure is due to an update check, we should let this merge then let the update run.

PR #4906, opened 2023-08-29
Bump activesupport in kube-fluent-operator to mitigate CVE-2023-38037.

dlorenc:
Fixes:

Related:

#### For new package PRs only


dlorenc:
The update failure is intended, The new version has a lot of changes and we probably need to keep both versions in a stream.

ajayk (approved):

PR #4966, opened 2023-08-30
rebuild 124 packages to remove unwanted security.csm xattrs

kaniini:
the Google Container Threat Detection feature adds these security.csm xattrs to store its own internal state.  we do not want to capture this state into our packages.

### DIFF
--- a/aactl.yaml
+++ b/aactl.yaml
@@ -1,7 +1,7 @@
 package:
   name: aactl
   version: 0.4.12
-  epoch: 0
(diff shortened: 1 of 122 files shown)


kaniini:
@jonjohnsonjr fetched every single apk we have built in the past few weeks and scanned them for the `security.csm` content.

ajayk (dismissed):

jonjohnsonjr (approved):

PR #4971, opened 2023-08-30
Add bpftool

Dentrax:
> **Note**
> **Before:** `19 vulnerabilities (6 medium, 10 low, 3 negligible)`
> **After:** `0-CVE`

Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/bpftool.yaml
+++ b/bpftool.yaml
@@ -1,0 +1,68 @@
+package:
+  name: bpftool
+  version: 7.2.0
+  epoch: 0
+  description: Linux kernel bpf manipulation tool
+  copyright:
+    - license: GPL-2.0-only OR BSD-2-Clause
+  dependencies:
+    runtime:
+      - libLLVM-16
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+      - automake
+      - autoconf
+      - pkgconf
+      - linux-headers
+      - libelf
+      - elfutils-dev
+      - libcap-dev
+      - clang-16
+      - llvm16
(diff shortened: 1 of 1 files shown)


bpftool.yaml
```diff
@@ -0,0 +1,68 @@
+package:
+  name: bpftool
+  version: 7.2.0
+  epoch: 0
+  description: Linux kernel bpf manipulation tool
+  copyright:
+    - license: GPL-2.0-only OR BSD-2-Clause
+  dependencies:
+    runtime:
+      - libLLVM-16
```
joshrwolf, line 10: I may be wrong, but off the top of my head this doesn't seem necessary?

joshrwolf:
other than the runtime dep, overall lgtm, but I defer to @kaniini on whether we should add this to `util-linux.yaml` or keep this separate

kaniini:
It should be part of util-linux.

Dentrax:
I just checked but I couldn't find `bpftool` in [util-linux](https://github.com/util-linux/util-linux). It seems Alpine put it into [linux-tools](https://git.alpinelinux.org/aports/tree/community/linux-tools/APKBUILD#n65), which is fetching linux kernel. Should I continue using [mirror upstream](https://github.com/libbpf/bpftool) or directly using the kernel?

bpftool.yaml
Dentrax: Hmmm, otherwise binary complains with `bpftool: error while loading shared libraries: libLLVM-16.so: cannot open shared object file: No such file or directory`

PR #4972, opened 2023-08-30
feat: add py3-gcovr package

stormqueen1990:
Add py3-gcovr and py3-lxml.

#### For new package PRs only

- [ ] This PR is marked as fixing a pre-existing package request bug
  - [ ] Alternatively, the PR is marked as related to a pre-existing package request bug, such as a dependency
- [x] REQUIRED - The package is available under an OSI-approved or FSF-approved license
- [x] REQUIRED - The version of the package is still receiving security updates

PR #4994, opened 2023-08-31
feat: add version stream for Cython

stormqueen1990:
Create a version stream for Cython 0, since lxml does not work with Cython 3.

Related: required by #4972

#### For new package PRs only

### DIFF
--- a/cython-0.yaml
+++ b/cython-0.yaml
@@ -7,7 +7,7 @@ package:
     - license: Apache-2.0
   dependencies:
     provides:
-      - cython=0.29.36
+      - cython=${{package.version}}-r${{package.epoch}}
 
 environment:
   contents:
@@ -24,7 +24,7 @@ pipeline:
     with:
       expected-commit: 702bbc80fe2e346c4ddc58f074e0054da56e0ced
       repository: https://github.com/cython/cython
-      tag: 0.29.36
+      tag: ${{package.version}}
 
   - runs: |
       python setup.py build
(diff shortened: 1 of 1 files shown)


cython0.yaml
```diff
@@ -0,0 +1,37 @@
+package:
+  name: cython
```
mattmoor, line 2: Generally we call these `cython-0` and we can have it `provides: cython=0.x.y`

cython0.yaml
```diff
@@ -0,0 +1,37 @@
...
+      python setup.py install --prefix=/usr --root="${{targets.destdir}}"
+
+  - uses: strip
+
+update:
+  enabled: true
+  github:
+    identifier: cython/cython
+    use-tag: true
+    tag-filter: 0.
```
mattmoor, line 37: I'm getting the `-0` above from this scoping.

cython0.yaml
stormqueen1990: oof, forgot to change `name:` 🤦‍♀️

stormqueen1990: Is this the general format for the provides clause?
```yaml
dependencies:
  provides:
    - cython=0.29.36
```

stormqueen1990: Updated

cython0.yaml
stormqueen1990: Makes sense, just updated

kaniini (approved):

PR #5002, opened 2023-08-31
crossplane: rewrite package.yaml to depend on our image

imjasonh:
The package.yaml we include tells crossplane to install the upstream xpkg.upbound.io provider for AWS, instead of using ours.

By rewriting the config, installing our S3 provider (for example) will tell Crossplane to install our AWS provider.


imjasonh:
nvm, we don't want this in Wolfi. I'll see if there's another way to configure this from outside the image.

PR #5021, opened 2023-09-01
Add neon database.

dlorenc:
This vendors postgres, but they actually fork it and modify it so it doesn't seem possible to use our own, this is just how the neon stuff works.

Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/neon.yaml
+++ b/neon.yaml
@@ -1,0 +1,71 @@
+package:
+  name: neon
+  version: 3819
+  epoch: 0
+  description: "Serverless Postgres. We separated storage and compute to offer autoscaling, branching, and bottomless storage."
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - rust
+      - openssl-dev
+      - build-base
+      - libtool
+      - readline-dev
+      - zlib-dev
+      - flex
+      - bison
+      - libseccomp-dev
+      - clang
+      - pkgconf
+      - libpq-15
+      - cmake
+      - postgresql-15-client
+      - protoc
+      - libcurl-openssl4
+      - poetry
+      - perl
+      - bash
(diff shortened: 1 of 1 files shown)


neon.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+      make -j$(nproc) -s BUILD_TYPE=release
+
+      mkdir -p "${{targets.destdir}}"/usr/bin
+      for f in pg_sni_router pageserver pagectl safekeeper storage_broker proxy neon_local; do
+        mv target/release/$f "${{targets.destdir}}"/usr/bin
+      done
+
+      mkdir -p "${{targets.destdir}}"/usr/local
+      mv pg_install/v14 "${{targets.destdir}}"/usr/local/v14
```
kaniini, line 58: I don’t think we should install files into /usr/local.  Can we patch in a different install location?

dlorenc: Yeah that part is simple, it's just for the final container image so I think we can symlink it via a compat sub package.

neon.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+      mkdir -p "${{targets.destdir}}"/usr/bin
+      for f in pg_sni_router pageserver pagectl safekeeper storage_broker proxy neon_local; do
+        mv target/release/$f "${{targets.destdir}}"/usr/bin
+      done
+
+      # This has to be used in conjunction with the POSTGRES_DISTRIB_DIR env var to tell
+      # neon where to find the postgres binaries.
+      # POSTGRES_DISTRIB_DIR=/usr/share/neon
+      mkdir -p "${{targets.destdir}}"/usr/share/neon
```
kaniini, line 60: Using `/usr/libexec/neon` for this seems more appropriate.

```suggestion
      mkdir -p "${{targets.destdir}}"/usr/libexec/neon
```

dlorenc: Nice!

joshrwolf (approved):

PR #5064, opened 2023-09-02
Create more virtual java packages.

mattmoor:
Today we have a virtual `default-jvm` package, which installs its JRE, and sets up that JRE as the default.

The above package is of somewhat limited use for us in images (currently) because we track LTS and we don't have a similar virtual that JUST tracks LTS.  So the first piece of this change is introducing a new LTS virtual `default-jvm-lts` which is provided by the same packages, but only on LTS versions (8, 11, 17).

We are also missing a virtual package that aggregates the JDK into a single logical package.  So the next two pieces of this change are to introduce a similar concept to the above for installing a default JDK (and a similar LTS form) called `default-jdk` and `default-jdk-lts` respectively.  These are provided by a new empty subpackage, which has a runtime dependency on the appropriately versioned `default-jvm` and `openjdk-X` packages.

For context, I would love to have `default-jdk-lts` replace [this](https://github.com/chainguard-images/images/blob/fc964c7de0ebdaa17d242efb517b3052d11b0c78/images/jdk/config/main.tf#L7-L13) and `default-jvm-lts` replace [this](https://github.com/chainguard-images/images/pull/1305/files#diff-b5e0524f13f5fd6bbd21cfbe9edf03c8697dfee7906329c87ecc562f5eff5efcR7-R13) so that we don't have to touch our images repo when a new LTS version of Java drops.

### DIFF
--- a/openjdk-10.yaml
+++ b/openjdk-10.yaml
@@ -1,7 +1,7 @@
 package:
   name: openjdk-10
   version: 10.0.2
-  epoch: 3
+  epoch: 4
   description: "Oracle OpenJDK 10"
   copyright:
     - license: GPL-2.0-with-classpath-exception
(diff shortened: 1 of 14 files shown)


mattmoor:
Looks like we already do something similar for `nodejs-lts` TIL.

dlorenc (approved):

mattmoor:
FWIW, I am waiting to merge this because I lack the confidence that this won't break all our Java packages on a holiday (in the US) weekend.  Once it's a more reasonable time to make riskier changes, I will land this and update our public images to consume these 🤞

kaniini (approved):

PR #5068, opened 2023-09-02
Patch more CVEs in loki.

dlorenc:
Fixes:

Related:

### DIFF
--- a/loki.yaml
+++ b/loki.yaml
@@ -1,7 +1,7 @@
 package:
   name: loki
   version: 2.8.4
-  epoch: 5
+  epoch: 6
   description: Like Prometheus, but for logs.
   copyright:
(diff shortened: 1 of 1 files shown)


loki.yaml
```diff
@@ -22,6 +22,8 @@ pipeline:
       expected-commit: 89d282c4399acc12d31bf8d0a288be7fd8e3e42d
 
   - runs: |
+      go get github.com/docker/docker@v20.10.24
```
ajayk, line 25: Can we please add the CVE-IDS as a comment , so in the next version bump it makes it easier to review if they are fixed in upstream

dlorenc: Fixed!

imjasonh (approved):

PR #5076, opened 2023-09-02
Refactor llvm-libunwind to a 15 and a 16 version.

dlorenc:
Fixes:

Related:

### DIFF
--- a/llvm-libunwind-15.yaml
+++ b/llvm-libunwind-15.yaml
@@ -1,0 +1,71 @@
+package:
+  name: llvm-libunwind-15
+  version: 15.0.7
+  epoch: 0
+  description: LLVM version of libunwind library
+  copyright:
+    - license: MIT
+  dependencies:
+    provides:
+      - llvm-libunwind=15
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+      - gcc-12-default # needed for -nostdlib++
+      - llvm15
+      - llvm15-dev
+      - clang
+      - cmake
+      - samurai
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/llvm/llvm-project
+      tag: llvmorg-${{package.version}}
+      expected-commit: 8dfdcc7b7bf66834a761bd8de445840ef68e4d1a
+
+  - runs: |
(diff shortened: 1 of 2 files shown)


dlorenc:
cc @kaniini

llvm-libunwind-16.yaml
```diff
@@ -0,0 +1,70 @@
+package:
+  name: llvm-libunwind-16
+  version: 16.0.6
+  epoch: 1
```
vaikas, line 4: Since this was already epoch 1 in the original one and now there's a provides a new version section, should this be bumped?

llvm-libunwind-16.yaml
```diff
@@ -0,0 +1,70 @@
...
+  description: LLVM version of libunwind library
+  copyright:
+    - license: MIT
+  dependencies:
+    provides:
+      - llvm-libunwind=16
```
vaikas, line 10: Just for my own edification, how come this doesn't have something like 16.999

llvm-libunwind-16.yaml
```diff
@@ -0,0 +1,70 @@
...
+        -DLIBUNWIND_INSTALL_HEADERS=YES
+
+  - runs: |
+      cmake --build build
+
+  - runs: |
+      DESTDIR="${{targets.destdir}}" cmake --install build
+
+subpackages:
+  - name: llvm-libunwind--16-static
```
vaikas, line 48: is the extra `-` intentional?

llvm-libunwind-15.yaml
```diff
@@ -50,20 +45,26 @@ pipeline:
       DESTDIR="${{targets.destdir}}" cmake --install build
 
 subpackages:
-  - name: llvm-libunwind-static
+  - name: llvm-libunwind--15-static
```
vaikas, line 48: Is the extra `-` intentional?

llvm-libunwind-16.yaml
dlorenc: Fixed

llvm-libunwind-16.yaml
dlorenc: I don't really know, I copied another llvm one :)

vaikas (dismissed):

llvm-libunwind-16.yaml
dlorenc: Good call!

rawlingsj (approved):

PR #5085, opened 2023-09-03
Remove curl from the cmake package, it triggers a circular dependency.

dlorenc:
Fixes:

Related:


dlorenc:
Ref https://github.com/wolfi-dev/os/pull/5074

kaniini (approved):

PR #5088, opened 2023-09-03
Add tigerbeetle binary!

dlorenc:
Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/tigerbeetle.yaml
+++ b/tigerbeetle.yaml
@@ -1,0 +1,44 @@
+package:
+  name: tigerbeetle
+  version: 0.13.123
+  epoch: 0
+  description: "The distributed financial accounting database designed for mission critical safety and performance."
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - zig
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+
+pipeline:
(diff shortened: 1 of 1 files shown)


tigerbeetle.yaml
```diff
@@ -0,0 +1,44 @@
...
+  - uses: strip
+
+update:
+  enabled: true
+  github:
+    identifier: tigerbeetledb/tigerbeetle
+    strip-prefix: v
+    use-tag: true
+    # There are some prereleases with a weird WEEKLY format.
+    tag-filter: 0.
```
rawlingsj, line 44: If we want to only want versions that start with `v` we should use `tag-filter: v`

rawlingsj: you can test locally by running
```
wolfictl check update  --override-version 0.0.1 tigerbeetle.yaml
```

rawlingsj: Oh sorry - I was assuming the `strip-prefix: v` above meant this repo had tag prefixes of `v`.  In this case the strip prefix is a NOOP

rawlingsj: I can clean that up in a followup PR as am fixing up some failing PRs ATM.

rawlingsj (approved):

PR #5090, opened 2023-09-03
add dask-gateway-server

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For security-related PRs

#### For version bump PRs

### DIFF
--- a/dask-gateway.yaml
+++ b/dask-gateway.yaml
@@ -1,7 +1,7 @@
 package:
   name: dask-gateway
   version: 2023.1.1
-  epoch: 0
+  epoch: 1
   description: "A multi-tenant server for securely deploying and managing Dask clusters."
   copyright:
     - license: BSD-3-Clause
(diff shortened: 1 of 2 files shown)


dask-gateway.yaml
```diff
@@ -37,6 +40,21 @@ pipeline:
...
         --prefix=/usr --root=${{targets.destdir}}
 
+subpackages:
+  - name: dask-gateway-server
+    description: A multi-tenant server for securely deploying and managing Dask clusters
+    pipeline:
+      - name: Python Build
+        runs: |
+          cd dask-gateway-server
+          pip install -r requirements.txt --prefix=/usr --root=${{targets.subpkgdir}}
```
rawlingsj, line 50: I'm not sure how previous discussions went but does `pip install -r requirements.txt` mean we are installing python dependencies on the fly rather than wolfi packaged python dependencies?  If so, just checking we want to do this here?  I know we do this for maven rather than having to package up maven central but not sure if this is the same for python?

developer-guy: fixed, thank you!

dlorenc (approved):

PR #5098, opened 2023-09-03
Add virtual packages for dotnet

mattmoor:
This change adds a new virtual package that we can use to track the latest dotnet release in public images downstream, so that we don't have to bump the version there when a new version is released here.

ref: https://github.com/chainguard-images/images/pull/1314

### DIFF
--- a/dotnet-7.yaml
+++ b/dotnet-7.yaml
@@ -1,7 +1,7 @@
 package:
   name: dotnet-7
   version: 7.0.110
-  epoch: 0
+  epoch: 1
   description: ".NET SDK"
   target-architecture:
     - all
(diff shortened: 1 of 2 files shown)


dotnet-7.yaml
```diff
@@ -94,6 +94,8 @@ subpackages:
     dependencies:
       runtime:
         - dotnet-7
+      provides:
+        - dotnet-runtime=7
```
kaniini, line 98: This will technically prevent both branches of dotnet from being installed at the same time, even though Microsoft designed it so you can.  Are we sure we want that?

mattmoor: I can split out a shim subpackage for this.

mattmoor: ok, these are distinct subpackages now

kaniini (approved):

PR #5099, opened 2023-09-03
Create `default-{component}` packages.

mattmoor:
These virtual packages are provided by versioned forms to install that version's components as the latest.

I believe that these packages largely render the subpackages of `kubernetes-latest` redundant, and that we could achieve what `kubernetes-latest` itself does by having the `kubernetes-X.Y-default` packages provide the `kubernetes-latest` package (or something else suitably named, e.g. `default-kubernetes` :shrug:).

I did this because I wanted something similar to the versioned `default-jvm` packages, so that I can install a kubectl at any version as the default, and not just "latest".

### DIFF
--- a/kubernetes-1.25.yaml
+++ b/kubernetes-1.25.yaml
@@ -183,13 +183,13 @@ subpackages:
           gcc -o "${{targets.subpkgdir}}"/usr/bin/pause-3.8 linux/pause.c
 
   - range: components
-    name: "default-${{range.key}}-1.25"
+    name: "${{range.key}}-1.25-default"
     description: "Makes this version of ${{range.key}} the default."
     dependencies:
       runtime:
         - ${{range.key}}-1.25
       provides:
-        - default-${{range.key}}=1.25
+        - ${{range.key}}-default=1.25
     pipeline:
       - runs: |
           mkdir -p ${{targets.subpkgdir}}/usr/bin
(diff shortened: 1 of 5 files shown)


mattmoor:
@kaniini This is my first time playing with the `range` stuff, so let me know if there are better ways of achieving this.

kubernetes-1.25.yaml
```diff
@@ -182,17 +182,42 @@ subpackages:
...
           mkdir -p ${{targets.subpkgdir}}/usr/bin
-          ln -s kubectl-1.25 ${{targets.subpkgdir}}/usr/bin/kubectl
-          ln -s kubelet-1.25 ${{targets.subpkgdir}}/usr/bin/kubelet
-          ln -s kube-scheduler-1.25 ${{targets.subpkgdir}}/usr/bin/kube-scheduler
-          ln -s kube-proxy-1.25 ${{targets.subpkgdir}}/usr/bin/kube-proxy
-          ln -s kube-controller-manager-1.25 ${{targets.subpkgdir}}/usr/bin/kube-controller-manager
-          ln -s kube-apiserver-1.25 ${{targets.subpkgdir}}/usr/bin/kube-apiserver
+          ln -s ${{range.key}}-1.25 ${{targets.subpkgdir}}/usr/bin/${{range.key}}
+
+  - name: kubernetes-1.25-default
```
joshrwolf, line 198: this feels a little nitty, but do we want `default-kubernetes-1.25` instead to match the component defaults? or vice versa for some sort of standardization?

mattmoor: If anything I'd flip the others around because this package isn't new (see L185 in the baseline).

I biased this way because we also have `default-jvm` and friends in JRE land, but it's probably better to be consistent with things more locally in this case.

I wish we had better guidance on naming packages because I feel like we have loads of little inconsistencies like this. 😞

kaniini (approved):

PR #5139, opened 2023-09-05
Add three more python deps from hatch.

dlorenc:
Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/py3-SecretStorage.yaml
+++ b/py3-SecretStorage.yaml
@@ -1,0 +1,46 @@
+# Generated from https://pypi.org/project/SecretStorage/
+package:
+  name: py3-SecretStorage
+  version: 3.3.3
+  epoch: 0
+  description: Python bindings to FreeDesktop.org Secret Service API
+  copyright:
+    - license: BSD 3-Clause License
+  dependencies:
+    runtime:
+      - py3-cryptography
+      - py3-jeepney
+      - python3
+
+environment:
+  contents:
+    packages:
+      - ca-certificates-bundle
+      - wolfi-base
+      - busybox
+      - build-base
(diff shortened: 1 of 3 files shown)


rawlingsj:
update check failed because of a newer version of brotli - but that's been [merged](https://github.com/wolfi-dev/os/blob/8794c10d0ee221d8931bd8ad0305c9c629d348ed/brotli.yaml#L3) since so let's ignore the failed check.

rawlingsj (dismissed):

py3-SecretStorage.yaml
```diff
@@ -0,0 +1,46 @@
+# Generated from https://pypi.org/project/SecretStorage/
+package:
+  name: py3-SecretStorage
```
rawlingsj, line 3: Is the camel case here intentional?

dlorenc: Yeah it's the real package name.

rawlingsj (approved):

py3-SecretStorage.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: 3993fc7162fe50852852913b90b203ba13727c38
+      repository: https://github.com/mitya57/secretstorage
+      tag: ${{package.version}}
+
+  - name: Python Build
+    runs: |
```
vaikas, line 35: Should this and all the others use:
python/build-wheel pipeline instead?

dlorenc: Oh fancy! I didn't know we had that one yet.

PR #5159, opened 2023-09-05
feat: add dependencies for Perl DateTime module

stormqueen1990:
Add part of the required dependencies for the Perl `DateTime` module. This PR includes 12 dependencies that do not depend on other modules. More PRs to come with the remaining dependencies.

Related: #5051

#### For new package PRs only

### DIFF
--- a/perl-class-data-inheritable.yaml
+++ b/perl-class-data-inheritable.yaml
@@ -1,0 +1,51 @@
+# Generated from https://git.alpinelinux.org/aports/plain/main/perl-class-data-inheritable/APKBUILD
+package:
+  name: perl-class-data-inheritable
+  version: "0.09"
+  epoch: 0
+  description: Inheritable, overridable class data
+  copyright:
+    - license: GPL-1.0-or-later OR Artistic-1.0-Perl
+  dependencies:
+    runtime:
+      - perl
+
+environment:
+  contents:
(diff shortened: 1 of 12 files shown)


perl-class-data-inheritable.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - autoconf
+      - perl
+
+pipeline:
+  - uses: fetch
+    with:
+      expected-sha512: 2e76aaf12c8d26442d53746e5d74636bb35c55461efb6d9b5ea50a635e3f781707b2d7f9cb3da9113ed31de464d3931f9734c29ace2a7ee5c6e111392b6a97cf
+      uri: https://cpan.metacpan.org/authors/id/R/RS/RSHERER/Class-Data-Inheritable-${{package.version}}.tar.gz
+
+  - runs: |
```
vaikas, line 29: If we think (and it's fine to say too early), this is how things work in perl land, then maybe we should look into making this into perl/build-make and L37 perl/clean-build (or something like that?)
Seems to be the same in all of these?

stormqueen1990: @vaikas yes! I created https://github.com/chainguard-dev/melange/pull/657 for that 😁

vaikas (approved):

PR #5215, opened 2023-09-06
sysstat: switch to use GitHub

luhring:
Sysstat hadn't been receiving updates, because it was using release-monitoring, and presumably because it has recent releases marked as pre-release (although I haven't figured out why).

Sysstat was also using `fetch` even though it was pulling source from GitHub.

This PR switches both the source checkout and the updates to use GitHub.

🤞  The hope is that this triggers the update bot to move this package to the latest release: `12.7.4`.

### DIFF
--- a/sysstat.yaml
+++ b/sysstat.yaml
@@ -1,7 +1,7 @@
 package:
   name: sysstat
   version: 12.6.2
-  epoch: 0
+  epoch: 1
   description: Performance monitoring tools
   copyright:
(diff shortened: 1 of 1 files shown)


ajayk (approved):

luhring:
>The hope is that this triggers the update bot to move this package to the latest release: 12.7.4.

Yay! https://github.com/wolfi-dev/os/pull/5218

PR #5226, opened 2023-09-07
add sorting of packages/runtime deps alpha to yam

vaikas:
Signed-off-by: Ville Aikas <vaikas@chainguard.dev>

Now when you ran `yam` it will sort the:
`.environment.contents.packages` and  `.package.dependencies.runtime` for you.

### DIFF
--- a/.github/workflows/wolfictl-check-update.yaml
+++ b/.github/workflows/wolfictl-check-update.yaml
@@ -24,6 +24,7 @@ jobs:
       with:
         separator: ' '
         files: "*.yaml"
+        files-ignore: ".yam.yaml"
 
     - name: Check
       id: check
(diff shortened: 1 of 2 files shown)


vaikas:
This should ignore just .yam.yaml

https://github.com/tj-actions/changed-files#input_files_ignore

imjasonh (approved):
We should take a big pass through all the yamls and alphabetize them, so we don't get caught on this in each future dep bump or package change.

The change is a no-op though so it doesn't need new epochs / builds.

vaikas:
> We should take a big pass through all the yamls and alphabetize them, so we don't get caught on this in each future dep bump or package change.
> 
> The change is a no-op though so it doesn't need new epochs / builds.

Yes, that would be nice: https://github.com/wolfi-dev/os/pull/5227

But looks like there are now lints that are catching problems that maybe weren't there when the files were introduced, so I'll tackle that once I get little more room on my plate.

Second problem however is that it is apparently building the entire world for no good reason, so everybody was assuming it would be a nop, but for some reason that assumption is not exactly right (or I screwed something else up?)

PR #5227, opened 2023-09-07
Sort: .environment.contents.packages .package.dependencies.runtime

vaikas:
manual run of yam to sort things.


rawlingsj:
I think this PR is similar to https://github.com/wolfi-dev/os/pull/4526#issuecomment-1776948291 where we need to figure out a way to roll this out.  Though wonder if we can close this one in favour of the other or are they both valid?

joshrwolf:
similar logic to https://github.com/wolfi-dev/os/pull/4526#issuecomment-1776948291

please feel free to disagree with me an reopen if so!

PR #5249, opened 2023-09-07
go dependency epoch bump [batch 9]

rawlingsj:


workerd.yaml
dlorenc, line 1: Accident?

rawlingsj: indeed :) thanks

joshrwolf (approved):

PR #5253, opened 2023-09-07
Add lksctp-tools and libsctp-dev packages

jamonation:
Useful to have this library for any higher layer protocols or applications that use SCTP e.g. [WebRTC](https://webrtcforthecurious.com/docs/07-data-communication/) and [Diameter](https://www.f5.com/glossary/diameter-protocol) for IP telephony respectively.

#### For new package PRs only


vaikas (dismissed):

lksctp-tools.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+  - uses: strip
+
+subpackages:
+  - name: libsctp-dev
+    pipeline:
+      - uses: split/dev
+    description: libsctp dev
+
+update:
```
vaikas, line 54: They do not seem to use releases, so you probably have to use use-tags for the update section.
```
  github:
    use-tag: true
```

fwiw, you can test the update section locally like this:
```
GITHUB_TOKEN="<YOUR TOKEN HERE> wolfictl check update  --override-version 0.0.1 <yourfile>
```

jamonation: Ah nice, TIL!

jamonation:
@vaikas if you're inclined to review this again, I added a couple sub-packages. One for docs, and another for shared library itself. I also fixed the auto-update portion with the `use-tag` magic.

rawlingsj (approved):

PR #5257, opened 2023-09-07
explicitly enable/disable bootstrap-awareness

imjasonh:
Opening this to gather comments

This changes our interactions with the bootstrap/stage3 repo to be an opt-in at `make`-time, rather than something we implicitly support if available for packages that need bootstrapping.

When `BOOTSTRAP=yes`, the bootstrap/stage3 repo+keyring are added to all `melange build` invocations.

When `BOOTSTRAP=no` (or non-`yes`), the repo+keyring are not appended, and if as a result the package is not available, the build will not be able to resolve dependencies and fail.

The canonical case is: `busybox` needs `busybox` to build itself. When `BOOTSTRAP=yes`, `melange build busybox.yaml` is taught about the existence of the bootstrap/stage3 repo, where it finds a `busybox`, and it can happily build itself. When `BOOTSTRAP=no`, `melange build busybox.yaml` only knows about `packages.wolfi.dev/os`, where we better hope `busybox` is available already.

When `BOOTSTRAP=yes`, package builds are _not aware that `packages.wolfi.dev/os` exists_. This is fine, because `wolfictl text` constructs a build order that is bootstrap-aware (wolfictl PR incoming), and packages will be built in order -- `busybox` is built from bootstrap-`busybox`, that's used to build `go`, then that's used to build `crane` from the local 
`go` package. They don't need to know that `wolfi` exists during bootstrap mode.

What we lose in this is a signal in our melange YAMLs that a package _might_ depend on bootstrapping. `apk-tools` now no longer has any indication that it depends on bootstrap packages. Improvements to `wolfictl`'s graph UX can make this clearer in a more generalized, automated way, based on observed dep requirements.

This also removes `BUILDWORLD` and replaces it with the opposite-semantics `BOOTSTRAP`. We can change this back so that `BUILDWORLD=yes` means `BOOTSTRAP=yes`, if that seems like a smaller lift.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,9 +11,6 @@ KEY ?= local-melange.rsa
 REPO ?= $(shell pwd)/packages
 CACHE_DIR ?= gs://wolfi-sources/
 
-WOLFI_SIGNING_PUBKEY ?= https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
-WOLFI_PROD ?= https://packages.wolfi.dev/os
-
 MELANGE_OPTS += --repository-append ${REPO}
 MELANGE_OPTS += --keyring-append ${KEY}.pub
 MELANGE_OPTS += --signing-key ${KEY}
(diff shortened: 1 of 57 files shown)


kaniini:
I like this, it will allow us to more easily jettison the legacy bootstrap path later on when we start working on rebootstrapping the C world.

Makefile
```diff
@@ -28,17 +25,24 @@ ifeq (${USE_CACHE}, yes)
...
+ifeq (${BOOTSTRAP}, yes)
+	MELANGE_OPTS += -k https://packages.wolfi.dev/bootstrap/stage3/wolfi-signing.rsa.pub
+	MELANGE_OPTS += -r https://packages.wolfi.dev/bootstrap/stage3
+	PKGLISTCMD += -k https://packages.wolfi.dev/bootstrap/stage3/wolfi-signing.rsa.pub
+	PKGLISTCMD += -r https://packages.wolfi.dev/bootstrap/stage3
+else
+	MELANGE_OPTS += -k https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
+	MELANGE_OPTS += -r https://packages.wolfi.dev/os
+	PKGLISTCMD += -k https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
+	PKGLISTCMD += -r https://packages.wolfi.dev/os
```
stormqueen1990, line 43: Very likely a silly question but I am curious: since these values are all repeated twice, is there a specific reason for not keeping the variable approach?

imjasonh: Not silly at all! It's just a stylistic choice, and I don't feel strongly about it either way. I'll flip it the other way, that seems less error prone.

joshrwolf:
is it worth checking before this lands that all the changed packages work with the new `make BOOTSTRAP=yes ...`?

imjasonh:
> is it worth checking before this lands that all the changed packages work with the new `make BOOTSTRAP=yes ...`?

I think we can if you want, but my hope was we'd try building everything with `BOOTSTRAP=yes` next (either with `make` here or with concurrent `wolfictl build`), and find out if something's broken as a result then.

joshrwolf (approved):

kaniini (approved):

PR #5286, opened 2023-09-08
Add patch for CVE-2023-4807 to openssl.

dlorenc:
This is probably unnecessary given it only seems to apply to windows, but it's harmless.

Fixes:

Related:

#### For security-related PRs

### DIFF
--- a/openssl.yaml
+++ b/openssl.yaml
@@ -1,7 +1,7 @@
 package:
   name: openssl
   version: 3.1.2
-  epoch: 0
+  epoch: 1
   description: "the OpenSSL cryptography suite"
   copyright:
     - license: Apache-2.0
@@ -37,6 +37,10 @@ pipeline:
     with:
       uri: https://www.openssl.org/source/openssl-${{package.version}}.tar.gz
       expected-sha256: a0ce69b8b97ea6a35b96875235aa453b966ba3cba8af2de23657d8b6767d6539
+
+  - uses: patch
+    with:
+      patches: CVE-2023-4807.patch
 
   - name: Configure and build
     runs: |
--- a/openssl/CVE-2023-4807.patch
+++ b/openssl/CVE-2023-4807.patch
@@ -1,0 +1,45 @@
+From 4bfac4471f53c4f74c8d81020beb938f92d84ca5 Mon Sep 17 00:00:00 2001
+From: Bernd Edlinger <bernd.edlinger@hotmail.de>
+Date: Tue, 22 Aug 2023 16:07:30 +0200
+Subject: [PATCH] Avoid clobbering non-volatile XMM registers
+
+This affects some Poly1305 assembler functions
+which are only used for certain CPU types.
+
+Remove those functions for Windows targets,
+as a simple interim solution.
+
+Fixes #21522
+
+Reviewed-by: Tomas Mraz <tomas@openssl.org>
+Reviewed-by: Paul Dale <pauli@openssl.org>
+(Merged from https://github.com/openssl/openssl/pull/21808)
+
+(cherry picked from commit 7b8e27bc2e02238986d89ef0ece067ec1b48e165)
+---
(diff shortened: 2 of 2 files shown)


dlorenc:
Ref: https://github.com/wolfi-dev/advisories/pull/228

dlorenc:
Advisory here: https://www.openssl.org/news/secadv/20230908.txt

Patch source: https://github.com/openssl/openssl/commit/4bfac447.patch

openssl.yaml
```diff
@@ -38,6 +38,10 @@ pipeline:
       uri: https://www.openssl.org/source/openssl-${{package.version}}.tar.gz
       expected-sha256: a0ce69b8b97ea6a35b96875235aa453b966ba3cba8af2de23657d8b6767d6539
 
+  - uses: patch
+    with:
+      patches: CVE-2023-4807.patch
```
luhring, line 43: Are you sure this is the right CVE ID?

luhring: Ah, I see the link now, thanks!

luhring:
> This is probably unnecessary given it only seems to apply to windows, but it's harmless

So my two cents here is we should standardize how we handle this kind of situation, and I'd vote for not carrying patches when we're known to not be affected. If we're not affected, I'd rather mark it as a false positive rather than say we fixed it.

Curious for other thoughts though!

openssl.yaml
dlorenc: Yea, another case of NVD missing the entry...

mattmoor:
This calls for my favorite quote about theory and practice.

Personally I’d love to live in a world where marking something as a false positive is as effective as applying a benign patch, but the reality is that until something like VEX is further along that simply applying the patch is more effective at dropping the CVE count from scanners.

I think that until VEX has reached a level of ubiquity we should do both in the name of customer obsession.  Make sure as many scanners report zero as possible, but also document the false positive so that folks with sufficiently enlightened scanners can see that there is no great rush to upgrade to that patched version.  This feels like a win win to me.  My $0.02

luhring:
> but the reality is that until something like VEX is further along that simply applying the patch is more effective at dropping the CVE count from scanners.

Not in this particular case. NACKing this would have an equal effect on scanner result count.

dlorenc:
> So my two cents here is we should standardize how we handle this kind of situation, and I'd vote for not carrying patches when we're known to not be affected. If we're not affected, I'd rather mark it as a false positive rather than say we fixed it.

I convinced myself that not_affected is the right answer, but I got there a different way - I don't think carrying patches is something we should shy away from if it'll quiet scanners, but I think marking this as fixed will actually have the opposite affect.

IIUC, marking this as fixed in 3.1.2-r1 will quiet scanners for this version, but will actually cause some scanners to start reporting this vulnerability as present in versions < 3.1.2-r1 and will cause a lot of noise and churn for users that we believe don't actually have to upgrade.

@luhring - is my interpretation on the behavior this will trigger in scanners for lower versions correct?

luhring:
> @luhring - is my interpretation on the behavior this will trigger in scanners for lower versions correct?

Yes! And that's a great point to mention here. So in short, **NACKing will never cause new findings to appear.**

I'm inclined to take that route (and close this PR), marking this as "not affected" (false positive) in the advisory data, both because of the point you raise here, and because we are not affected.

> I don't think carrying patches is something we should shy away from if it'll quiet scanners

I think this is a reasonable opinion to have here, but if it's okay with you, I'd rather we discuss this more before standardizing on this approach, because I'm not sure I agree with it quite yet.

mattmoor:
Fwiw the position I was advocating above was to mark cases like this not affected (for exactly this reason), but (upon discretion) applying patches where that is insufficient to quiet scanners

kaniini:
> I'd vote for not carrying patches when we're known to not be affected

Yeah, we should minimize our patch load, as we have to manually review them on package updates.

PR #5287, opened 2023-09-08
Switch jq to github updates instead of release-monitor.

dlorenc:
We now build from a new github repo which is "official", use that for update checks!

Avoiding epoch bump on purpose.

Fixes:

Related:

### DIFF
--- a/jq.yaml
+++ b/jq.yaml
@@ -32,9 +32,6 @@ subpackages:
     description: "headers for libjq"
     pipeline:
       - uses: split/dev
-    dependencies:
-      runtime:
-        - jq
 
   - name: "jq-doc"
     description: "jq documentation"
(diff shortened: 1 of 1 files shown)


jq.yaml
```diff
@@ -43,5 +40,8 @@ subpackages:
 
 update:
   enabled: true
-  release-monitor:
-    identifier: 13252
+  github:
+    identifier: stedolan/jq
```
luhring, line 44: Should this be updated?

```suggestion
    identifier: jqlang/jq
```

jq.yaml
```diff
@@ -43,5 +40,8 @@ subpackages:
...
+    use-tag: true
```
luhring, line 45: And it looks like they're doing releases now, in case we want that...

```suggestion
```

luhring (approved):

PR #5291, opened 2023-09-08
Add velero.yaml

johnfosborneiii:
#### For new package PRs only

### DIFF
--- a/velero.yaml
+++ b/velero.yaml
@@ -25,12 +25,8 @@ pipeline:
       tag: v${{package.version}}
       expected-commit: bdbe7eb242b0f64d5b04a7fea86d1edbb3a3587c
       repository: https://github.com/vmware-tanzu/velero
-      destination: velero
 
   - runs: |
-      cd velero
-      mkdir -p ${{targets.destdir}}/usr/bin/
-
       CGO_ENABLED=0
       GO111MODULE=on
       GOOS=$(go env GOOS)
--- a/velero.yaml
+++ b/velero.yaml
@@ -1,0 +1,88 @@
+package:
+  name: velero
+  version: 1.11.1
+  epoch: 0
+  description: Backup and migrate Kubernetes applications and their persistent volumes
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - wolfi-base
+      - restic
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+      - go
+      - git
+
+pipeline:
+  - uses: git-checkout
+    with:
+      tag: v${{package.version}}
+      expected-commit: bdbe7eb242b0f64d5b04a7fea86d1edbb3a3587c
+      repository: https://github.com/vmware-tanzu/velero
+      destination: velero
+
+  - runs: |
+      cd velero
+      mkdir -p ${{targets.destdir}}/usr/bin/
+
+      CGO_ENABLED=0
+      GO111MODULE=on
+      GOOS=$(go env GOOS)
+      GOARCH=$(go env GOARCH)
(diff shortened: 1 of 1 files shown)


velero.yaml
```diff
@@ -0,0 +1,88 @@
...
+      LDFLAGS="$LDFLAGS -X github.com/vmware-tanzu/velero/pkg/buildinfo.Version=v${{package.version}}"
+
+      # Go flags
+      # -x Print commands
+      # -v Print names of packages
+      # -o output file location
+      # -trimpath Remove all system file paths
+      # -ldflags Args passed to the go link tool
+
+      go build \
```
joshrwolf, line 53: can we use our `go/build` pipeline for these builds?

example: https://github.com/wolfi-dev/os/blob/main/calico.yaml#L408

velero.yaml
```diff
@@ -0,0 +1,88 @@
...
+      - go
+      - git
+
+pipeline:
+  - uses: git-checkout
+    with:
+      tag: v${{package.version}}
+      expected-commit: bdbe7eb242b0f64d5b04a7fea86d1edbb3a3587c
+      repository: https://github.com/vmware-tanzu/velero
+      destination: velero
```
joshrwolf, line 28:
```suggestion
```

if you remove this, you don't need the below:

```
cd velero
```

velero.yaml
johnfosborneiii: Ariadne says not to because of an existing bug that is still open https://github.com/chainguard-dev/melange/issues/461

velero.yaml
```diff
@@ -0,0 +1,88 @@
...
+  - name: velero-restore-helper
+    description: Velero Restore Helper
+    pipeline:
+      - runs: |
+          set -x
+          mkdir -p "${{targets.subpkgdir}}/usr/bin/"
+          mv ${{targets.destdir}}/usr/bin/velero-restore-helper ${{targets.subpkgdir}}/usr/bin/
+    dependencies:
+      runtime:
+        - velero
```
joshrwolf, line 81: is this an actual runtime dependency? my understanding is `velero-restore-helper` is an independent cli, but happy to be told otherwise

johnfosborneiii: You are correct, that's why I moved it to a subpkg

velero.yaml
```diff
@@ -0,0 +1,88 @@
+package:
+  name: velero
+  version: 1.11.1
+  epoch: 0
+  description: Backup and migrate Kubernetes applications and their persistent volumes
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - wolfi-base
```
joshrwolf, line 10:
```suggestion
```

we generally leave this out since it's implied during the image builds dependency on `wolfi-baselayout`

velero.yaml
```diff
@@ -0,0 +1,88 @@
...
+
+  - uses: strip
+
+subpackages:
+  - name: velero-restore-helper
+    description: Velero Restore Helper
+    pipeline:
+      - runs: |
+          set -x
+          mkdir -p "${{targets.subpkgdir}}/usr/bin/"
```
joshrwolf, line 77: nit, but here's a neat little trick we've been using more and more of to avoid needing the `mkdir -p` step: https://github.com/wolfi-dev/os/blob/main/libaec.yaml#L59

velero.yaml
joshrwolf: I don't think that affects you?

the following should be wai:

```
  - uses: go/build
    with:
      packages: ./cmd/velero
      output: velero
      ldflags: -s -w
```

velero.yaml
joshrwolf: if `velero-restore-helper` is not dependent on `velero`, lets remove it

joshrwolf (approved):

PR #5301, opened 2023-09-08
feat: replace repetitive Perl commands with pipelines

stormqueen1990:
Replace the repetitive Perl make and cleanup with the new built-in pipelines.


joshrwolf (dismissed):
❤️

joshrwolf (approved):

PR #5325, opened 2023-09-08
feat: bump perl-uri to v5.20 and use built-in pipelines

stormqueen1990:
Bump version of perl-uri to v5.20 (latest on release-monitoring.org) and use built-in pipelines instead of repetitive commands.

#### For version bump PRs

- [ ] ~Patch source: _patch source here_~ _(not applicable)_

### DIFF
--- a/perl-uri.yaml
+++ b/perl-uri.yaml
@@ -1,6 +1,6 @@
 package:
   name: perl-uri
-  version: 5.19
+  version: 5.20
   epoch: 0
   description: Uniform Resource Identifiers (absolute and relative)
   copyright:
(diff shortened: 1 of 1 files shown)


perl-uri.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: perl-uri
-  version: 5.19
+  version: 5.20
   epoch: 0
```
joshrwolf, line 4:
```suggestion
  epoch: 1
```

luhring: Why this?

joshrwolf: you're right, I'm an idiot lost in severance

joshrwolf (approved):

PR #5329, opened 2023-09-08
q 0.12.0: new version and updated build instructions

brunomiguel:
#### For version bump PRs
- [x] Patch source: https://github.com/natesales/q/archive/refs/tags/v0.12.0.tar.gz


ajayk (changes requested):
go can be set to the latest version as it is supported by upstream now  

`- go `

brunomiguel:
> go can be set to the latest version as it is supported by upstream now
> 
> `- go `

Thanks for the feedback, @ajayk. I just updated the build file.

ajayk (approved):

PR #5356, opened 2023-09-10
modsecurity: split the static libs into sub package

tcnghia:
* "modsecurity: split the static libs to subpackage"

### DIFF
--- a/modsecurity.yaml
+++ b/modsecurity.yaml
@@ -1,7 +1,7 @@
 package:
   name: modsecurity
   version: 3.0.10
-  epoch: 1
+  epoch: 2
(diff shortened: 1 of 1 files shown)


modsecurity.yaml
```diff
@@ -58,6 +58,10 @@ pipeline:
 
   - uses: strip
 
+  - runs: |
+      # We don't need the static libraries
+      find ${{targets.destdir}} -name '*.a' -print | xargs rm
```
dlorenc, line 63: Could you try using the split/static pipeline to put it into a sub package?

tcnghia: Done!

dlorenc (approved):

PR #5374, opened 2023-09-11
remove unused CI workflows

joshrwolf:
removes the un used workflows displaced by `{ci}-build.yaml`


.github/workflows/build-world.yaml
joshrwolf, line 1: this one might make a comeback, but for now it is unused (and doesn't work)

imjasonh (approved):

PR #5376, opened 2023-09-11
Update cosign.yaml

johnfosborneiii:
Striping debug info and symbol table from cosign binary and bumping the epoch


rawlingsj (changes requested):
Looks like this is missing the `- uses: strip`

johnfosborneiii:
> Looks like this is missing the `- uses: strip`

My mistake, crossed the streams with PRs, fixed now

rawlingsj (approved):

PR #5404, opened 2023-09-11
Add a "version streams" section to the PR template.

mattmoor:
This change adds a section to the PR checklist to ensure folks are considering certain things when creating new version streams.

### DIFF
--- a/.github/pull-request-template.md
+++ b/.github/pull-request-template.md
@@ -32,6 +32,13 @@ addressed, and remove any items that are not relevant to this PR.
   - [ ] Alternatively, the PR is marked as related to a pre-existing package request bug, such as a dependency
 - [ ] REQUIRED - The package is available under an OSI-approved or FSF-approved license
 - [ ] REQUIRED - The version of the package is still receiving security updates
+- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`)
+
+#### For new version streams
+<!-- remove if unrelated -->
+- [ ] The upstream project actually supports multiple concurrent versions.
+- [ ] Any subpackages include the version string in their package name (e.g. `name: ${{package.name}}-compat`)
+- [ ] The package (and subpackages) `provides:` logical unversioned forms of the package (e.g. `nodejs`, `nodejs-lts`)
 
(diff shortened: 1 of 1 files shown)


kaniini (approved):

.github/pull-request-template.md
```diff
@@ -32,6 +32,13 @@ addressed, and remove any items that are not relevant to this PR.
   - [ ] Alternatively, the PR is marked as related to a pre-existing package request bug, such as a dependency
 - [ ] REQUIRED - The package is available under an OSI-approved or FSF-approved license
 - [ ] REQUIRED - The version of the package is still receiving security updates
+- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`)
+
+#### For new version streams
+<!-- remove if unrelated -->
+- [ ] The upstream project actually supports multiple concurrent versions.
+- [ ] Any subpackages include the version string in their package name (e.g. `name: ${{package.name}}-compat`)
```
vaikas, line 40: We should probably also say something about the package.name having a version in there explicitly. Without this there can be problems with solver not being able to do the right things. For example, the php had originally the php as the `package.name` so when we added php-8.1 for example, it caused some grief, that we wouldn't have had if the name of the package was originally, say php-8.2, and then later php-8.1.  I'm just using this as an example, there may be better ones, but that's the _only_ one that I'm familiar with 🤣 So, I'm tagging @kaniini to clarify this, but it plays a part in this imho as well.

kaniini: Yeah we should require that each implementation in a version stream has a distinct `package.name`.

mattmoor: @vaikas yeah, I sort of saw that as implied by it being under version streams, which generally use `-X` or `-X.Y` suffixes, but we can be more explicit too.

vaikas: I'd rather make it explicit 🤣

mattmoor: @vaikas unless there's a specific change you have in mind, let's merge and iterate?

mattmoor: I'm going for it

PR #5430, opened 2023-09-12
libwebp: CVE-2023-4863 use specific commit until latest upstream is tagged + released

rawlingsj:
this is so we can pull in this commit https://github.com/webmproject/libwebp/commit/902bc9190331343b2017211debcec8d2ab87e17a

#### For PRs that add patches

### DIFF
--- a/libwebp.yaml
+++ b/libwebp.yaml
@@ -1,7 +1,9 @@
+# remove nolint when switching back to released tarball
+#nolint:valid-pipeline-git-checkout-commit,valid-pipeline-git-checkout-tag
 package:
   name: libwebp
-  version: 1.3.1
-  epoch: 1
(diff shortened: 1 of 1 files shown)


rawlingsj:
strange - the error in CI fails with
```
fatal: reference is not a tree: 902bc9190331343b2017211debcec8d2ab87e17a
```

I might try with the github mirror

rawlingsj:
Wolfictl Check Updates is expected to fail until we get the new update from upstream

luhring (commented):
It'd be nice to have a comment in there about why this was done (i.e. to patch a CVE).

Also don't forget the advisories PR, please 😃 🙏

luhring (approved):
🙏

PR #5470, opened 2023-09-12
ingress-nginx-controller fixes needed for package to built and run

rawlingsj:
ingress-nginx-controller still needs some work to pull out a number of embedded dependencies from its melange pipeline and into separate packages.  That is not part of this PR and will need to be done as a followup.

The aim of this PR is to
- include a working set of runtime dependencies that have been tested
- avoid git clone of main branches and use fixed commits
- switch to modsecurity wolfi package rather than git clone in the ingress-nginx-controller pipeline
- add compat subpackage for modsecurity
- move fetch sha's from melange variables to avoid confusion when melange bump runs
- removes some unnecessary pipeline cleanup steps as files in the $BUILD_PATH are not added to the end apk package
- update embedded nginx module versions to match current list in wofli - these embedded fetches will be moved out in a followup PR
- updates ingress-nginx-controller to 1.8.2 

@tcnghia and @strongjz contributed to this PR


ingress-nginx-controller.yaml
```diff
@@ -2,10 +2,50 @@
...
+      - lua-resty-memcached
+      - lua-resty-redis
+      - lua-resty-string
+      - lua-resty-upload
+      - brotli-dev
+      - gd-dev
+      - geoip-dev
+      - libmaxminddb-dev
+      - libxml2-dev
+      - libxslt-dev
```
rawlingsj, line 46: These -dev packages came in from one of the earlier commits but they don't look quite right to me.  I might try one more test and remove these or at least drop the `-dev` suffix

kaniini (approved):

PR #5471, opened 2023-09-12
rebuild and withdraw defective neon packages

kaniini:
neon 3819 and 3836 packages have erroneous `provides: so:libpq.so.5` which cause problems with other packages.  Rebuild with latest melange to fix this problem.


vaikas (approved):
Thank you!

imjasonh (approved):

PR #5479, opened 2023-09-12
new package: openrefine 3.7.5

brunomiguel:
#### For new package PRs only


brunomiguel:
The openrefine/openrefine the CI action is complaining about is there, but I forgot to include it when I did the PR. A second commit fixed it, but the CI action still can't see it.

brunomiguel:
> The openrefine/openrefine the CI action is complaining about is there, but I forgot to include it when I did the PR. A second commit fixed it, but the CI action still can't see it.

Fixed with the 6155a81 commit

openrefine.yaml
```diff
@@ -0,0 +1,61 @@
...
+      mv graphics "${{targets.destdir}}"/usr/share/${{package.name}}/
+      mv IDEs "${{targets.destdir}}"/usr/share/${{package.name}}/
+      mv main "${{targets.destdir}}"/usr/share/${{package.name}}/
+      mv packaging "${{targets.destdir}}"/usr/share/${{package.name}}/
+      mv server "${{targets.destdir}}"/usr/share/${{package.name}}/
+      mv LICENSE.txt "${{targets.destdir}}"/usr/share/licenses/${{package.name}}/
+      cp pom.xml "${{targets.destdir}}"/usr/share/${{package.name}}/
+      cp refine.ini "${{targets.destdir}}"/usr/share/${{package.name}}/
+      cp refine "${{targets.destdir}}"/usr/share/${{package.name}}/
+      install -Dm755 openrefine.sh "${{targets.destdir}}"/usr/bin/openrefine
```
rawlingsj, line 53: Thanks for the contribution!

In Wolfi we prefer to compile software and copy built artifacts into the `${{targets.destdir}}/usr/bin/`.

I'm not familiar with refine but after a quick look I wonder if we can add `openjdk-20` and a compatible version of `nodejs`, maybe `maven` too as I see a pom.xml, to the build environment package list and run
```
./refine build
```
to compile the necessary artifacts and move those to `${{targets.destdir}}/usr/bin/`?

brunomiguel: I updated the YAML file following your suggestions, @rawlingsj. However, it gets stuck on `Downloading from central: https://repo.maven.apache.org/maven2/org/python/jython-standalone/2.7.3/jython-standalone-2.7.3.jar`. Some file downloads also stay inactive for a while but eventually proceed. I'm running `mvn --debug` to see if it shows any useful debug info, but I have had no luck with it. I guess something on their side blocks after x requests in y time.

I looked at how Debian (Stable branch) packages it, and they include a bunch of dependencies that would be otherwise downloaded. This works for Debian Stable, but I don't know if it's the best approach for a moving project.

I appreciate any feedback and suggestions you may have on how to proceed.

brunomiguel: OpenRefine has an official Linux distribution of the software, available on the releases page from GitHub. I updated the openrefine.yaml to use that. Is this acceptable?

rawlingsj:
Sorry for the slow response.  Unfortunately we can only accept PRs that build from source as this is a requirement for Wolfi.  As jython only supports python 2.x which is EOL I think we will need to wait until the python 3 support lands I'm afraid.

I'm going to close this PR as it looks like we need to wait for the upstream changes mentioned above but happy to help with getting openrefine in once we can build from source by maintained versions.

PR #5489, opened 2023-09-13
feat: add version/commit/date to boring-registry binary

stormqueen1990:
Update the `boring-registry` package to include the version, commit, and date metadata in the binary.

### DIFF
--- a/boring-registry.yaml
+++ b/boring-registry.yaml
@@ -1,7 +1,7 @@
 package:
   name: boring-registry
   version: 0.11.4
-  epoch: 0
+  epoch: 1
   description: Terraform Provider and Module Registry
   copyright:
     - license: MIT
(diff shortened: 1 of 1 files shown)


dlorenc:
> Update the `boring-registry` package to include the version, commit, and date metadata in the binary.

Nice! I always find it funny when folks don't set their own information in their builds: https://github.com/TierMobility/boring-registry/blob/main/Dockerfile#L9

dlorenc (approved):

stormqueen1990:
> > Update the `boring-registry` package to include the version, commit, and date metadata in the binary.
> 
> Nice! I always find it funny when folks don't set their own information in their builds: https://github.com/TierMobility/boring-registry/blob/main/Dockerfile#L9

That's true! It seems they set it only on `.goreleaser.yml`: https://github.com/TierMobility/boring-registry/blob/62960773833a865f8f7b4b02414a1281c512b1f2/.goreleaser.yml#L14

PR #5492, opened 2023-09-13
feat(php-pecl-mcrypt): Improve update section to allow auto updates of mcrypt

carstendietrich:
The update section of the PHP PECL mcrypt extension had an open todo, this PR adds the needed information to allow automatic updates via the GitHub check done by the wolfi-bot.

Mcrypt has some weird 2.x tags that are not relevant, therefore I filter down to only take the 1.x tags into consideration when upgrading.


carstendietrich:
@vaikas this is maybe interesting for upcoming PECL-based PHP extensions.

rawlingsj:
Sorry for the _really_ long delay on this and thanks for the PR!

rawlingsj (approved):

PR #5536, opened 2023-09-14
ingress-nginx-controller: update package dependencies to be the lates…

rawlingsj:
…t versions available, followup change to split them into their own packages.


rawlingsj:
completed via https://github.com/wolfi-dev/os/pull/5836

PR #5571, opened 2023-09-14
Fluentd plugin for http-out

cjohannsen81:
Attempt to add ruby gem for fluentd to enable http out traffic for /metrics

#### For version bump PRs

### DIFF
--- a/fluent-plugin-out-http.yaml
+++ b/fluent-plugin-out-http.yaml
@@ -31,7 +31,7 @@ pipeline:
   - uses: fetch
     with:
       expected-sha256: 821f4468c2db1ddcfaad8520b1d9a5b33f5d6eda89a2fc931da735b3d42a6f44
-      uri: https://github.com/fluent-plugins-nursery/fluent-plugin-out-http/archive/refs/tags/${{package.version}}.tar.gz
+      uri: https://github.com/fluent-plugins-nursery/fluent-plugin-out-http/archive/refs/tags/v${{package.version}}.tar.gz
 
   - uses: ruby/unlock-spec
 
@@ -50,4 +50,5 @@ update:
   enabled: true
   github:
     identifier: fluent-plugins-nursery/fluent-plugin-out-http
+    strip-prefix: v
     use-tag: true
--- a/fluent-plugin-out-http.yaml
+++ b/fluent-plugin-out-http.yaml
@@ -49,4 +49,4 @@ pipeline:
 update:
   enabled: true
   github:
-    identifier: splunk/fluent-plugin-out-http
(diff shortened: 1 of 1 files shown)


fluent-plugin-out-http.yaml
```diff
@@ -0,0 +1,52 @@
...
+    with:
+      gem: ${{package.name}}
+      version: ${{package.version}}
+
+  - uses: ruby/clean
+
+update:
+  enabled: true
+  github:
+    identifier: splunk/fluent-plugin-out-http
```
found-it, line 52: I think this should be `fluent-plugins-nersery/fluent-plugin-out-http` so the auto-updater can locate the repo on github

found-it (dismissed):

cjohannsen81 (commented):
You are right. Pipeline failed because of that.

fluent-plugin-out-http.yaml
```diff
@@ -0,0 +1,52 @@
...
+    with:
+      gem: ${{package.name}}
+      version: ${{package.version}}
+
+  - uses: ruby/clean
+
+update:
+  enabled: true
+  github:
+    identifier: fluent-plugins-nursery/fluent-plugin-out-http
```
found-it, line 52: Looks like it's defaulting to releases which the upstream repo doesn't have. Probably need to add this flag

```suggestion
    identifier: fluent-plugins-nursery/fluent-plugin-out-http
    use-tag: true
```

found-it (changes requested):

fluent-plugin-out-http.yaml
```diff
@@ -0,0 +1,52 @@
...
+      - ruby-3.2-dev
+      - build-base
+      - busybox
+      - git
+
+pipeline:
+  - uses: fetch
+    with:
+      expected-sha256: 821f4468c2db1ddcfaad8520b1d9a5b33f5d6eda89a2fc931da735b3d42a6f44
+      uri: https://github.com/fluent-plugins-nursery/fluent-plugin-out-http/archive/refs/tags/${{package.version}}.tar.gz
```
found-it, line 34:
```suggestion
      uri: https://github.com/fluent-plugins-nursery/fluent-plugin-out-http/archive/refs/tags/${{package.version}}.tar.gz
      uri: https://github.com/fluent-plugins-nursery/fluent-plugin-out-http/archive/refs/tags/v${{package.version}}.tar.gz
```

found-it: I _think_ you need a `v` in front of the version here

found-it: That looks to be why the build is failing

found-it (approved):

PR #5576, opened 2023-09-14
Version stream `cluster-autoscaler`.

mattmoor:
This component is expected to be co-versioned with Kubernetes components, so this should track the same streams as Kubernetes itself.

### DIFF
--- a/cluster-autoscaler-1.28.yaml
+++ b/cluster-autoscaler-1.28.yaml
@@ -1,0 +1,58 @@
+package:
+  name: cluster-autoscaler-1.28
+  version: 1.28.0
+  epoch: 2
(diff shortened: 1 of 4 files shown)


cluster-autoscaler-1.28.yaml
```diff
@@ -1,10 +1,13 @@
 package:
-  name: cluster-autoscaler
+  name: cluster-autoscaler-1.28
   version: 1.28.0
-  epoch: 1
+  epoch: 2
```
mattmoor, line 4: This bumped so that the `package.full-version` wins the virtual over the pre-existing package.

ajayk (approved):

PR #5586, opened 2023-09-15
Update mongo-tools.yaml to 100.8.0 and turn on autoupdate

jamonation:
#### For version bump PRs

### DIFF
--- a/mongo-tools/release-platform.patch
+++ b/mongo-tools/release-platform.patch
@@ -1,0 +1,19 @@
+diff --git a/release/platform/platform.go b/release/platform/platform.go
+index df1712d..9fed41f 100644
+--- a/release/platform/platform.go
++++ b/release/platform/platform.go
+@@ -548,6 +548,14 @@ var platforms = []Platform{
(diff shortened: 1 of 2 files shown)


mongo-tools/release-platform.patch
```diff
@@ -0,0 +1,19 @@
+diff --git a/release/platform/platform.go b/release/platform/platform.go
+index df1712d..9fed41f 100644
+--- a/release/platform/platform.go
++++ b/release/platform/platform.go
+@@ -548,6 +548,14 @@ var platforms = []Platform{
+ 		Repos:     []Repo{RepoEnterprise, RepoOrg},
+ 		BuildTags: []string{"failpoints", "ssl"},
+ 	},
++	{
```
jamonation, line 9: This whole stanza is added because by default on Linux, the platform detection only picks up Arm64 or X86_64 from `uname -sm`, whereas Aarch64 is what Docker on macOS returns, as well as Github runners I think.

joshrwolf (approved):

PR #5507, opened 2023-09-13
further dep packages for lcov added along with perl-devel-cover

mritunjaysharma394:
Fixes:

Related:

#### For new package PRs only

- [ ] This PR is marked as fixing a pre-existing package request bug
  - [ ] Alternatively, the PR is marked as related to a pre-existing package request bug, such as a dependency
- [x] REQUIRED - The package is available under an OSI-approved or FSF-approved license
- [x] REQUIRED - The version of the package is still receiving security updates
- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`)

#### For new version streams

- [ ] The upstream project actually supports multiple concurrent versions.
- [ ] Any subpackages include the version string in their package name (e.g. `name: ${{package.name}}-compat`)
- [ ] The package (and subpackages) `provides:` logical unversioned forms of the package (e.g. `nodejs`, `nodejs-lts`)

#### For security-related PRs

- [ ] The security fix is recorded in the [advisories](https://github.com/wolfi-dev/advisories) repo

#### For version bump PRs

- [ ] The `epoch` field is reset to 0

#### For PRs that add patches

- [ ] Patch source is documented

PR #5598, opened 2023-09-15
adds nested perl deps of lcov

mritunjaysharma394:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/perl-algorithm-diff.yaml
+++ b/perl-algorithm-diff.yaml
@@ -1,0 +1,46 @@
+# Generated from https://git.alpinelinux.org/aports/plain/community/perl-algorithm-diff/APKBUILD
+package:
+  name: perl-algorithm-diff
+  version: "1.201"
+  epoch: 0
+  description: Compute 'intelligent' differences between two files / lists
+  copyright:
+    - license: GPL-1.0-or-later OR Artistic-1.0-Perl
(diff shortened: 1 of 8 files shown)


joshrwolf (approved):

perl-algorithm-diff.yaml
```diff
@@ -0,0 +1,46 @@
...
+  - uses: strip
+
+subpackages:
+  - name: perl-algorithm-diff-doc
+    pipeline:
+      - uses: split/manpages
+    description: perl-algorithm-diff manpages
+
+update:
+  enabled: false # latest releases mismatches in release monitor and github release is not maintained to latest version
```
rawlingsj, line 46: I wonder if this means the project is not maintained anymore? If not should we be adding it to wolfi?

mritunjaysharma394: I think I discussed this with @kaniini and we moved to enabling it false but it might be not maintained because is completed? It is needed for `perl-text-diff` package which is in turn along with others is needed for accp package so not sure would should be the next step here

mritunjaysharma394:
Closing this as already handled with the now merging #5507

PR #5693, opened 2023-09-19
tooling: Cache root folder of dev-container to preserve bash history

pdeslaur:
This change allows us to maintain bash history between different invocations!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,9 @@ KEY ?= local-melange.rsa
 REPO ?= $(shell pwd)/packages
 CACHE_DIR ?= gs://wolfi-sources/
 
+# Assign a default XDG_CACHE_HOME value for darwin.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -75,9 +75,10 @@ packages/$(ARCH)/%.apk: $(KEY)
 dev-container:
 	docker run --privileged --rm -it \
 	    -v "${PWD}:${PWD}" \
+			-v "${HOME}/.cache/wolfictl/dev-container-wolfi/root:/root" \
```
joshrwolf, line 78: nit: should we `XDG_CACHE_HOME` this?

pdeslaur: Very good idea! Using `XDG_CACHE_HOME` now. I added a line a bit higher to set a default value if it's undefined.

PR #5730, opened 2023-09-20
libucontext: new package

cpanato:
- libucontext: new package

Related: https://github.com/chainguard-dev/image-requests/issues/359

#### For new package PRs only

#### For new version streams

### DIFF
--- a/libucontext.yaml
+++ b/libucontext.yaml
@@ -41,3 +41,4 @@ update:
     identifier: kaniini/libucontext
     strip-prefix: v
     tag-filter: v
+    use-tag: true
--- a/libucontext.yaml
+++ b/libucontext.yaml
@@ -1,0 +1,43 @@
+package:
+  name: libucontext
+  version: 1.2
+  epoch: 0
+  description: "ucontext function implementations"
+  copyright:
+    - license: "ISC"
+
+environment:
+  contents:
+    packages:
+      - wolfi-base
+      - cmake
+      - build-base
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/kaniini/libucontext
+      tag: v${{package.version}}
+      expected-commit: 4dde3417b4bb4b1b1545bd913be337680b5e28c3
+
+  - runs: |
+      make
+      make DESTDIR="${{targets.destdir}}" install
+
+  - uses: strip
+
+subpackages:
+  - name: "libucontext-dev"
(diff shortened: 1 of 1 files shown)


libucontext.yaml
```diff
@@ -0,0 +1,44 @@
...
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/kaniini/libucontext
+      tag: v${{package.version}}
+      expected-commit: 4dde3417b4bb4b1b1545bd913be337680b5e28c3
+
+  - runs: |
+      make
+      make install
```
rawlingsj, line 25:
```suggestion
      make DESTDIR="${{targets.destdir}}" install
```
I think we may need this else we won't get the binary in the end apk package.   It looks like `DESTDIR` is supported https://github.com/kaniini/libucontext/blob/be80075e957c4a61a6415c280802fea9001201a2/Makefile#L210C41-L210C48

cpanato: done

libucontext.yaml
```diff
@@ -0,0 +1,43 @@
...
+    dependencies:
+      provides:
+        - libucontext-dev=${{package.full-version}}
+
+update:
+  enabled: true
+  github:
+    identifier: kaniini/libucontext
+    strip-prefix: v
+    tag-filter: v
```
rawlingsj, line 43:
```suggestion
    use-tag: true
```
sorry - we do still need use-tags on this package though as https://github.com/kaniini/libucontext/releases only uses tags.  Releases are preferred but only if the upstream project uses github releases.

rawlingsj, line 43:
```suggestion
    tag-filter: v
    use-tag: true
```

sorry - we do still need use-tags on this package though as https://github.com/kaniini/libucontext/releases only uses tags.  Releases are preferred but only if the upstream project uses github releases.

rawlingsj (approved):

libucontext.yaml
```diff
@@ -0,0 +1,44 @@
...
+  - uses: strip
+
+subpackages:
+  - name: "libucontext-dev"
+    description: "ucontext function implementations (development files)"
+    pipeline:
+      - uses: split/dev
+    dependencies:
+      provides:
+        - libucontext-dev=${{package.full-version}}
```
rawlingsj, line 36: this should be
```
    dependencies:
      runtime:
        - libucontext
```

PR #5735, opened 2023-09-20
glog: new package

cpanato:
- glog: new package

Related: https://github.com/chainguard-dev/image-requests/issues/359

#### For new package PRs only

#### For new version streams

### DIFF
--- a/glog.yaml
+++ b/glog.yaml
@@ -1,0 +1,59 @@
+package:
+  name: glog
+  version: 0.6.0
+  epoch: 0
+  description: "C++ implementation of the Google logging module"
+  copyright:
+    - license: "BSD-3-Clause"
+
+environment:
+  contents:
+    packages:
+      - wolfi-base
+      - cmake
+      - build-base
+      - libucontext-dev
+      - libunwind-dev
+      - samurai
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/google/glog
+      tag: v${{package.version}}
+      expected-commit: b33e3bad4c46c8a6345525fd822af355e5ef9446
+
+  - uses: cmake/configure
+    with:
+      opts: |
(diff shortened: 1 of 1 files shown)


glog.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+  - uses: cmake/build
+
+  - uses: cmake/install
+
+  - uses: strip
+
+subpackages:
+  - name: "glog-dev"
+    description: "multi-language toolbox for accelerated data interchange and in-memory processing (development files)"
```
dlorenc, line 47: Is this right?

glog.yaml
```diff
@@ -0,0 +1,60 @@
...
+    pipeline:
+      - uses: split/dev
+    dependencies:
+      provides:
+        - glog-dev=${{package.full-version}}
```
rawlingsj, line 52:
```suggestion
    dependencies:
      runtime:
        - glog
```
melange will auto add the provides

glog.yaml
cpanato: no, copy and paste issue

fixed

glog.yaml
cpanato: fixed

glog.yaml
```diff
@@ -0,0 +1,60 @@
...
+      - uses: split/dev
+    dependencies:
+      runtime:
+        - glog
+
+update:
+  enabled: true
+  github:
+    identifier: google/glog
+    use-tag: true
```
stormqueen1990, line 58: It seems this project uses releases -- should `use-tag` be omitted so the releases are pulled instead?

cpanato: ahh, i see, i will drop that

cpanato: done

stormqueen1990 (approved):

PR #5740, opened 2023-09-20
use goreleaser pipeline

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/grpc-health-probe.yaml
+++ b/grpc-health-probe.yaml
@@ -1,7 +1,7 @@
 package:
   name: grpc-health-probe
   version: 0.4.19
-  epoch: 4
+  epoch: 5
   description: A command-line tool to perform health-checks for gRPC applications in Kubernetes and elsewhere
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


grpc-health-probe.yaml
```diff
@@ -23,12 +23,9 @@ pipeline:
...
       expected-commit: 1089f40639043ee46cc47cbea1bcb0cc48cfc284
 
-  - uses: go/build
+  - uses: goreleaser/build
     with:
-      packages: .
-      output: grpc_health_probe
-      tags: netgo
-      ldflags: -s -w
+      args: --snapshot
```
rawlingsj, line 28: do we need `--snapshot` or can we use the release?  I think this will change the version displayed by the `--version` and we'd want to indicate these are real releases rather than development ones?

developer-guy: snapshot option is just for avoiding to release these artifacts in GitHub releases

developer-guy: there was an option for snapshot builds, I've updated the build to use that.

rawlingsj: though using `snapshot` will still mean the version of the binary will include a `-devel` suffix?  I wonder if there's a way to avoid publishing the artifacts and still use a release version?

developer-guy: let me try with that, thanks James!

developer-guy: without snapshot option the pipeline is failing due to git dirty state

developer-guy: I used --skip-validate to avoid having git dirt state issues, does that make sense?

imjasonh (approved):

PR #5751, opened 2023-09-20
fluent-plugin-kubernetes_metadata_filter: new package

cjohannsen81:
Package needed to ensure liveness/readiness probes for fluentd and solve error when using fluentd.

Fixes:

2023-09-20 15:45:41 +0000 [error]: config error file="/etc/fluent/fluent.conf" error_class=Fluent::NotFoundPluginError error="Unknown filter plugin 'kubernetes_metadata'. Run 'gem search -rd fluent-plugin' to find plugins"

Related:

#### For new package PRs only

### DIFF
--- a/fluent-plugin-kubernetes_metadata_filter.yaml
+++ b/fluent-plugin-kubernetes_metadata_filter.yaml
@@ -1,0 +1,54 @@
+package:
+  name: fluent-plugin-kubernetes_metadata_filter
+  version: 3.3.0
+  epoch: 0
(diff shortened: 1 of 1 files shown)


found-it:
@cjohannsen81 I suspect this package isn't building the necessary files, it needs to use git-checkout because the [gemspec](https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/blob/master/fluent-plugin-kubernetes_metadata_filter.gemspec#L16) uses `git ls-files` and the `fetch` pipeline downloads the archive which does not preserve the git metadata

PR #5752, opened 2023-09-20
opentofu : move opentf package to opentofu

ajayk:


cpanato (approved):
do we need to witdraw to old one?

PR #5769, opened 2023-09-20
add docker and docker-cli

joshrwolf:
adds `docker` metapackage that includes `dockerd` and `docker-cli`.

also includes `pigz` which is a dependency for `dockerd`.

There are a bunch of additional subpackages that will eventually be used in the image building phase.

#### For new package PRs only

### DIFF
--- a/docker-cli.yaml
+++ b/docker-cli.yaml
@@ -1,0 +1,69 @@
+package:
+  name: docker-cli
+  version: 24.0.6
+  epoch: 0
+  description: The Docker CLI
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - bash
+      - btrfs-progs-dev
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - containerd
+      - coreutils
+      - go
+      - libseccomp-dev
+      - libtool
+      - linux-headers
+      - lvm2-dev
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/docker/cli
+      tag: v${{package.version}}
+      expected-commit: ed223bc820ee9bb7005a333013b86203a9e1bc23
+
+  - runs: |
+      export DISABLE_WARN_OUTSIDE_CONTAINER=1
+
+      mkdir -p "$GOPATH/src/github.com/docker/"
+      ln -sf /home/build "$GOPATH"/src/github.com/docker/cli
+      LDFLAGS="" make VERSION=${{package.version}} dynbinary
+
+      install -Dm755 build/docker-linux-* ${{targets.destdir}}/usr/bin/docker
+
+  - runs: |
+      make manpages
+      install -Dm644 man/man1/* -t ${{targets.destdir}}/usr/share/man/man1/
+
+  # Docker is vehemenetly against stripping the resulting binary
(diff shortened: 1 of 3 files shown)


docker-cli.yaml
```diff
@@ -0,0 +1,69 @@
...
+      install -Dm755 build/docker-linux-* ${{targets.destdir}}/usr/bin/docker
+
+  - runs: |
+      make manpages
+      install -Dm644 man/man1/* -t ${{targets.destdir}}/usr/share/man/man1/
+
+  # Docker is vehemenetly against stripping the resulting binary
+  # Ref: https://github.com/moby/moby/blob/d8a51d2887cbc465ab8d76ed98f7a86996ab3c22/project/PACKAGERS.md#stripping-binaries
+  # - uses: strip
+  - runs: |
```
imjasonh, line 48: why does yam want this?

docker.yaml
```diff
@@ -0,0 +1,141 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/moby/moby
+      tag: v${{package.version}}
+      expected-commit: 1a7969545d73537545645f5cd2c79b7a77e7d39f
+
+  - runs: |
+      # Repliace the environment setup performed by docker
```
imjasonh, line 37:
```suggestion
      # Replicate the environment setup performed by docker
```

docker.yaml
```diff
@@ -0,0 +1,141 @@
...
+      export DOCKER_BUILDTAGS="seccomp"
+
+      ./hack/make.sh dynbinary
+
+  # Independently managed from moby/moby, but contains many utility scripts needed for the official docker images.
```
imjasonh, line 42: Should this be a subpackage? Or another top-level package? That might make it easier to keep up to date.

docker.yaml
```diff
@@ -0,0 +1,141 @@
...
+      - runs: |
+          install -Dm755 /home/build/hack/dind "${{targets.subpkgdir}}"/usr/bin/dind
+
+  # Ref: https://github.com/docker-library/docker/blob/master/Dockerfile-git.template
+  - name: docker-git
+    description: "Docker in Docker with Git"
+    dependencies:
+      runtime:
+        - docker-dind
+        - git
```
imjasonh, line 125: Does this need to be a meta-package? This seems like something we could just replicate at image-build-time.

joshrwolf: that's fair, I got carried away with the other `docker-*` and didn't stop to think. incoming

docker-cli.yaml
joshrwolf: yam doesn't seem to understand empty comment blocks, and squishes this comment block into the previous node resulting in bad yaml.

if there's a better way to warn us from `strip` I'm all 👂!

docker.yaml
joshrwolf: I included it in here since the repo doesn't publish any tags or releases, so either way it's going to be difficult to auto update. Tying it to this package at least reminds us to check when `docker` (which has tags) bumps.

docker-cli.yaml
joshrwolf: ref: https://github.com/wolfi-dev/os/actions/runs/6253508252/job/16979074829?pr=5769

imjasonh (dismissed):

ajayk (approved):

PR #5781, opened 2023-09-21
ingress-nginx-controller: move binaries to use /usr/bin, remove opent…

rawlingsj:
…racing-cpp, aeger-client-cpp and nginx_ajp_module

The following projects have been deprecated so removing from wolfi and from being used in ingress-nginx-controller.

https://github.com/opentracing/opentracing-cpp
https://github.com/jaegertracing/jaeger-client-cpp

AJP support has been deprecated from the upstream ingress controller project https://github.com/kubernetes/ingress-nginx/pull/10158


rawlingsj:
Looks like [datadog-cpp](https://github.com/wolfi-dev/os/blob/c6e9ec483ffefa4036f16d6d2892d882853944e4/datadog-cpp.yaml#L18) and [zipkin-cpp](https://github.com/wolfi-dev/os/blob/c6e9ec483ffefa4036f16d6d2892d882853944e4/zipkin-cpp.yaml#L19)   
 are both dependant on the deprecated https://github.com/opentracing/opentracing-cpp

Probably need to remove these too.  I've removed the open tracing module from the ingress-nginx-controller in his PR which is the only package that uses them.

https://github.com/kubernetes/ingress-nginx is moving to [opentelemetry](https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/third-party-addons/opentelemetry.md) rather than opentracing.

Elizafox (approved):

PR #5800, opened 2023-09-21
Apply patch to istio-envoy to use make 4.4

tcnghia:
* "istio-envoy-1.8,1.9: patch to use make 4.4"

This is the same with https://github.com/wolfi-dev/os/pull/5779/. However this time Envoy is a dependency (of istio/proxy): we need to apply patch of a patch.

See https://github.com/envoyproxy/envoy/pull/29261 , https://github.com/envoyproxy/envoy/commit/1b576a103463ca008c76800d1f67929c2a2ffaeb 

#### For PRs that add patches

### DIFF
--- a/istio-envoy-1.18.yaml
+++ b/istio-envoy-1.18.yaml
@@ -52,10 +52,11 @@ pipeline:
 
       echo "build --config=clang" >> user.bazelrc
 
-      # When patching to 1.18.2, the PR
-      # https://github.com/istio/proxy/commit/3c27a1b0cf381ca854ccc3a2034e88c206928da2 updated
-      # the Envoy commit without updating its checksum.
-      sed -i s/0a932fd090feed587d82b4b943a1032a5a424587f65455a37236cffefc8648ef/8795a73871ce8bd4138191f746d4f1f5d0935d5e65b2b6f1803f966f13065211/g WORKSPACE
+      # Patch Envoy to be able to use make 4.4
+      #
+      # This patch is from https://github.com/envoyproxy/envoy/commit/1b576a103463ca008c76800d1f67929c2a2ffaeb.diff
+      sed -i "/sha256 = ENVOY_SHA256/a\\
+          patches = [\"//:1b576a103463ca008c76800d1f67929c2a2ffaeb.diff\"], patch_args = [\"-p1\"]," WORKSPACE
 
       bazel build --verbose_failures -c opt envoy
 
--- a/istio-envoy-1.18.yaml
+++ b/istio-envoy-1.18.yaml
@@ -1,5 +1,6 @@
 package:
   name: istio-envoy-1.18
+  # On the next version bump, please remove the make 4.4 patch.
   version: 1.18.3
   epoch: 1
   description: Envoy with additional Istio plugins (wasm, telemetry, etc)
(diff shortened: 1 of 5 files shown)


istio-envoy-1.19.yaml
```diff
@@ -50,6 +50,12 @@ pipeline:
       export JAVA_HOME=/usr/lib/jvm/java-11-openjdk
       mkdir -p .cache/bazel/_bazel_root
 
+      # Patch Envoy to be able to use make 4.4
+      #
+      # This patch is from https://github.com/envoyproxy/envoy/commit/1b576a103463ca008c76800d1f67929c2a2ffaeb.diff
+      sed -i "/sha256 = ENVOY_SHA256/a\\
+          patches = [\"//:1b576a103463ca008c76800d1f67929c2a2ffaeb.diff\"], patch_args = [\"-p1\"]," WORKSPACE
```
mattmoor, line 57: bump epochs?

tcnghia: Done, thanks!

istio-envoy-1.18.yaml
```diff
@@ -52,10 +52,11 @@ pipeline:
...
 
-      # When patching to 1.18.2, the PR
-      # https://github.com/istio/proxy/commit/3c27a1b0cf381ca854ccc3a2034e88c206928da2 updated
-      # the Envoy commit without updating its checksum.
-      sed -i s/0a932fd090feed587d82b4b943a1032a5a424587f65455a37236cffefc8648ef/8795a73871ce8bd4138191f746d4f1f5d0935d5e65b2b6f1803f966f13065211/g WORKSPACE
+      # Patch Envoy to be able to use make 4.4
+      #
+      # This patch is from https://github.com/envoyproxy/envoy/commit/1b576a103463ca008c76800d1f67929c2a2ffaeb.diff
+      sed -i "/sha256 = ENVOY_SHA256/a\\
+          patches = [\"//:1b576a103463ca008c76800d1f67929c2a2ffaeb.diff\"], patch_args = [\"-p1\"]," WORKSPACE
```
rawlingsj, line 59: It's probably fine but just checking that you knew about the builtin melange `patch` [pipeline](https://github.com/chainguard-dev/melange/blob/main/pkg/build/pipelines/patch.yaml)

i.e. like this
https://github.com/wolfi-dev/os/blob/876ce91c1d584f8d5b8d9b96d0b7951b1a9452aa/gsm.yaml#L29-L31 

It may be that this is doing something extra?

rawlingsj: other than the sha256 thing as I think that's not necessary if it's committed to this repo?

tcnghia: This is changing the Bazel rule to apply the patch for Envoy after Bazel downloads Envoy.

I think we can also do this as a patch file itself but it is getting too meta (patch of a patch of a patch).

tcnghia: The SHA256 is part of Bazel's `http_archive` rule, which downloads the source of its depedencies. It is similar to our fetch+patch pipeline. 

We are inserting the patch rule into this line right here https://github.com/istio/proxy/blob/master/WORKSPACE#L49 to apply the Envoy patch to Envoy after Bazel downloads Envoy (as a dependency of istio/proxy)

tcnghia:
Thanks @rawlingsj for the review! I've added also some comments to the `version:` line to make sure we remember to remove the patching when updating.

rawlingsj (approved):

tcnghia:
This PR has too many big Bazel build and taking 3-4 hours on Presubmit so I'm splitting it up

PR #5803, opened 2023-09-21
PHP 8.1 update: including extensions php-odbc and pdo_dblib

erikaheidi:
This PR updates the PHP 8.1 build to compile two additional PHP extensions:

- `php-odbc`, via `unixodbc`
- `php-pdo_dblib`, via `freetds`

PR #5809, opened 2023-09-21
PHP 8.2 update: including extensions php-odbc and pdo_dblib 

erikaheidi:
This PR updates the PHP 8.2 build to compile two additional PHP extensions:

- `php-odbc`, via `unixodbc`
- `php-pdo_dblib`, via `freetds`

Related: #5803

### DIFF
--- a/php-8.2.yaml
+++ b/php-8.2.yaml
@@ -7,7 +7,7 @@ package:
     - license: PHP-3.01
   dependencies:
     provides:
-      - php=${{package.version}}-r${{package.epoch}}}
+      - php=${{package.version}}-r${{package.epoch}}
     runtime:
       - libxml2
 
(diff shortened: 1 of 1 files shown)


php-8.2.yaml
```diff
@@ -7,7 +7,7 @@ package:
     - license: PHP-3.01
   dependencies:
     provides:
-      - php=${{package.version}}-r${{package.epoch}}}
+      - php=${{package.version}}-r${{package.epoch}}
```
vaikas, line 10: FYI: You can replace this with ${{package.full-version}}

vaikas (dismissed):
Thank you!!!

php-8.2.yaml
erikaheidi: thanks! i just realized I forgot to bump the epoch, will do it now!

imjasonh (approved):

PR #5810, opened 2023-09-21
Fix freetds 'provides' section

erikaheidi:
This PR removes the "provides" section from freetds since it doesn't use version streams.

### DIFF
--- a/freetds.yaml
+++ b/freetds.yaml
@@ -1,13 +1,10 @@
 package:
   name: freetds
   version: "1.3.20"
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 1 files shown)


freetds.yaml
```diff
@@ -7,7 +7,7 @@ package:
     - license: GPL-2.0
   dependencies:
     provides:
-      - freetds=${{package.version}}-r${{package.epoch}}}
+      - freetds=${{package.version}}-r${{package.epoch}}
```
ajayk, line 10:
```suggestion
```

ajayk: and please bump epoch

ajayk (approved):

ajayk (approved):

PR #5818, opened 2023-09-21
Use full-version substitutions

jonjohnsonjr:


kaniini (dismissed):

jonjohnsonjr:
This hits too many lints, so I'm giving up. Latest one is that nginx by default has a PID file under `/run/nginx` which runs afoul of the tempdir linter.

PR #5827, opened 2023-09-22
Fix wolfictl semgrep update build PR #5680

goproslowyo:
Fixes: Wolfictl semgrep update build PR #5680

Related:

#### For version bump PRs

- [x] ~~The `epoch` field is reset to 0~~

### DIFF
--- a/semgrep.yaml
+++ b/semgrep.yaml
@@ -40,7 +40,7 @@ pipeline:
       opam init --compiler=ocaml-base-compiler.4.14.0 -y
       opam switch create 4.14.0+static ocaml-base-compiler.4.14.0
       eval $(opam env --switch=4.14.0+static)
-      make dev-setup
+      make setup
       make core
       make copy-core-for-cli
 


semgrep.yaml
```diff
@@ -40,7 +40,7 @@ pipeline:
       opam init --compiler=ocaml-base-compiler.4.14.0 -y
       opam switch create 4.14.0+static ocaml-base-compiler.4.14.0
       eval $(opam env --switch=4.14.0+static)
-      make dev-setup
```
mattmoor, line 43: This needs to bump the epoch above

rawlingsj: yeah we just realised this was a PR to another PR branch that has the epoch bump

goproslowyo: Sorry for the trouble! I didn't realize the epoch was supposed to be bumped on the version rev from `1.40.0 --> 1.41.0`. I figured that was usually done when you have to patch a CVE/fix/rebuild a package where the version hasn't changed.

Since this was already merged, should I open another PR to the wolfictl branch? Let me know how you'd like me to fix it!

PR #5833, opened 2023-09-22
move py3-semgrep to a subpackage of semgrep

imjasonh:
The two packages are built from the same source, so this will mean less update noise.

There's also some indication that semgrep@$version needs to be used with py3-semgrep@$version, so this helps prevent the two package versions drifting out of sync in image builds.


semgrep.yaml
```diff
@@ -1,14 +1,14 @@
...
-  epoch: 0
+  epoch: 1
   description: "Lightweight static analysis for many languages. Find bug variants with patterns that look like source code."
   copyright:
     - license: LGPL-2.1
   dependencies:
     runtime:
       - python3
-      - py3-semgrep
+      - py3-semgrep=${{package.version}}
```
imjasonh, line 11:
```suggestion
      - py3-semgrep
```

Melange doesn't support this yet, but at least having them co-versioned should make this less necessary.

stormqueen1990 (approved):

imjasonh:
FYI @goproslowyo since you'd added these originally. This shouldn't have any impact on either package, it should just make it easier to keep them up to date together.

PR #5991, opened 2023-09-27
Remove a bad tempdir in cups to unblock package updates.

dlorenc:
Fixes:

Related:

#### For new package PRs only

Ref https://github.com/wolfi-dev/os/pull/5878


dlorenc:
Failure here is the update check, that's intentional.

kaniini (approved):

PR #6022, opened 2023-09-28
Add trillian logsigner and logserver packages

priyawadhwa:
Adds trillian logsigner and logserver packages

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches


trillian-logserver.yaml
```diff
@@ -0,0 +1,24 @@
...
+  version: 1.5.2
+  epoch: 0
+  description: Merkle tree implementation used in Sigstore
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - go
```
imjasonh, line 12: Not blocking, but this isn't needed since using `go/install` brings this dep in by itself. Specifying it again doesn't hurt it's just not needed.

trillian-logsigner.yaml
```diff
@@ -0,0 +1,24 @@
...
+pipeline:
+  - uses: go/install
+    with:
+      package: github.com/google/trillian/cmd/trillian_log_signer
+      version: v${{package.version}}
+
+update:
+  enabled: true
+  github:
+    identifier: google/trillian
```
imjasonh, line 23: Does it make sense for one of these to be a subpackage of the other, so they can both be updated as one update when the repo has a new release?

I think we'd just pick one to promote and make the other a subpackage.

trillian-logserver.yaml
```diff
@@ -0,0 +1,24 @@
+package:
+  name: trillian-logserver
```
rawlingsj, line 2: this isn't necessarily a blocker but I wondered if we should have one top level package `trillian` and then multiple subpackages, one for each `./cmd/...` in [here](https://github.com/google/trillian/tree/master/cmd)

mattmoor: Yeah, I'd recommend we do this, I just refactored `rekor` to do exactly this.

imjasonh: https://github.com/wolfi-dev/os/pull/6021/files

mattmoor: I wouldn't use a range for 2, I'd do what rekor does :)

priyawadhwa: thanks, i think i fixed it!

trillian.yaml
```diff
@@ -0,0 +1,55 @@
...
+      expected-commit: 8ba85b59c6eca840b1b673b7afc6c9516329a690
+
+  - runs: |
+      # ensure clean git tree
+      rm .gitconfig
+      export GOCACHE=$(mktemp -d)/.cache
+      export GOMODCACHE=$(mktemp -d)/mod
+
+      go build -o trillian_log_server ./cmd/trillian_log_server
+      go build -o trillian_log_signer ./cmd/trillian_log_signer
```
mattmoor, line 34: We should use `-trimpath -ldflags "-s -w"`

priyawadhwa: thanks, done!

mattmoor (approved):

PR #6083, opened 2023-09-29
librdkafka: Update GitHub organization

lyoung-confluent:
The `librdkafka` repository has been moved from https://github.com/edenhill/librdkafka to https://github.com/confluentinc/librdkafka, all of the tags/releases are still present and redirect from the old URL so this PR is a no-op but will ensure future releases are detected and downloaded from the correct URL.

I wasn't sure if I should bump `epoch` given this context or leave it the same.

### DIFF
--- a/librdkafka.yaml
+++ b/librdkafka.yaml
@@ -1,7 +1,7 @@
 package:
   name: librdkafka
   version: 2.2.0
-  epoch: 2
+  epoch: 3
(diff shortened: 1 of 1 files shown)


ajayk:
@lyoung-confluent epoch changes are not required

dlorenc (approved):

PR #6088, opened 2023-09-29
Fix broken ruby gems using git ls-files

found-it:
These gems were missing files due to missing git metadata when the gem was built

### DIFF
--- a/fluent-plugin-label-router.yaml
+++ b/fluent-plugin-label-router.yaml
@@ -1,7 +1,7 @@
 package:
   name: fluent-plugin-label-router
   version: "0.2.10_git20230928"
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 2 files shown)


imjasonh (dismissed):

fluent-plugin-label-router.yaml
```diff
@@ -13,31 +13,31 @@ package:
...
-      version: 0.2.10
+  # Hack until they cut a release
+  - runs: |
+      git clone https://github.com/kube-logging/fluent-plugin-label-router
+      cd fluent-plugin-label-router
+      git checkout 695d289f3b170d4739e79b21ce062700b10acf16
 
-  - uses: ruby/clean
+  - working-directory: fluent-plugin-label-router
+    pipeline:
```
mattmoor, line 31: I didn’t realize you could in-line this

mattmoor (approved):

PR #6121, opened 2023-10-01
python: add seaborn package

jspeed-meyers:
This PR adds the `seaborn` python package. This package is often used by data scientists and other data professionals. I think it could be useful for Wolfi users that want to build data science-related applications.

I should mention that this is my first time submitting a PR to Wolfi. So if it looks like I am confused, I am, or at least probably am. Helpful tips and constructive feedback encouraged. TY!

#### For new package PRs only
- [x] REQUIRED - The version of the package is still receiving security updates (I think so, at least. There are frequent updates.)
- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`): I can't find one.

### DIFF
--- a/py3-seaborn.yaml
+++ b/py3-seaborn.yaml
@@ -1,37 +1,39 @@
 # Generated from https://pypi.org/project/seaborn/
 package:
-    name: py3-seaborn
-    version: 0.13.0
-    epoch: 0
-    description: Statistical data visualization
-    copyright:
-        - license: BSD 3-Clause
-    dependencies:
-        runtime:
-            - py3-numpy
-            - py3-pandas
-            - py3-matplotlib
-            - python-3.11
+  name: py3-seaborn
+  version: 0.13.0
+  epoch: 0
+  description: Statistical data visualization
+  copyright:
+    - license: BSD 3-Clause
+  dependencies:
+    runtime:
+      - py3-numpy
+      - py3-pandas
+      - py3-matplotlib
+      - python3
+
 environment:
-    contents:
-        packages:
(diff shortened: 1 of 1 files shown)


py3-seaborn.yaml
```diff
@@ -0,0 +1,39 @@
...
+            - py3-pandas
+            - py3-matplotlib
+            - python-3.11
+environment:
+    contents:
+        packages:
+            - ca-certificates-bundle
+            - wolfi-base
+            - build-base
+            - python-3.11
```
imjasonh, line 21: Since 3.12 is coming out soon, we should think about whether this needs to be pinned to 3.11 or whether it can float to 3.12 when that becomes the latest py3.

If the latter, this should depend on `python3` and `py3-pip`, and when 3.12 comes out we'll bump this epoch to rebuild with 3.12.

If the former, we should understand why it needs to stay on 3.11 and document that somewhere, so we can check in some future version upgrade PR whether it can get on 3.12.

jspeed-meyers: Ah, I don't think this needs to be pinned to 3.11. I'll change the PR as you suggest.

jspeed-meyers: Fixed via commit 7b4550c

imjasonh (dismissed):

jspeed-meyers:
Sorry, @imjasonh. I seem to be too good at making the CI fail. 😢

jspeed-meyers:
I tested the update code locally and it works. 🤞

imjasonh (approved):

PR #6123, opened 2023-10-01
new package: python 3.12

imjasonh:
Also updating pip, setuptools and installer to be built with 3.12.

This doesn't make these `provide` python-3 yet, since that will make it the default python version for future packages, and will need to be coordinated with python package rebuilds.

### DIFF
--- a/python-3.12.yaml
+++ b/python-3.12.yaml
@@ -5,12 +5,12 @@ package:
   description: "the Python programming language"
   copyright:
     - license: PSF-2.0
-  # TODO(jason): Uncomment when we're ready for py-3.12 to be default everywhere.
-  #dependencies:
-  #  provides:
-  #    - python3=${{package.version}}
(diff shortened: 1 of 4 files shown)


ajayk (dismissed):

python-3.12.yaml
```diff
@@ -90,10 +94,13 @@ subpackages:
           # pyconfig.h is needed at runtime... ugh.
           mkdir -p "${{targets.destdir}}"/usr/include/python3.12
           mv "${{targets.subpkgdir}}"/usr/include/python3.12/pyconfig.h "${{targets.destdir}}"/usr/include/python3.12
+    dependencies:
+      provides:
+        - python3-dev=${{package.version}}
+        - python-3-dev=${{package.version}}
```
mattmoor, line 100: this needs to go

mattmoor (approved):

PR #6124, opened 2023-10-01
bump py3 deps to rebuild with 3.12

imjasonh:
After https://github.com/wolfi-dev/os/pull/6123 we'll need to rebuild everything that's built with python3 to be built with 3.12. This is an attempt to collect those bumps and get feedback about what else I've missed.

I did this pretty naively, with `git grep -l -- "- py.*3" | xargs wolfictl bump`, then eyeballed probably 20 or so packages that weren't named `py3-*`, I only found one that wasn't a genuine build-time dep (AFAIK)

### DIFF
--- a/aom.yaml
+++ b/aom.yaml
@@ -5,7 +5,7 @@
 package:
   name: aom
   version: 3.7.0
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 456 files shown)


imjasonh:
Gonna close to batch this up.

PR #6148, opened 2023-10-02
vault: fix lint errors

imjasonh:
This package has an `-entrypoint` package for compatibility with the OCI image, which should be named `-compat`. We didn't have this idiom back when this subpackage was created.

### DIFF
--- a/vault-1.14.yaml
+++ b/vault-1.14.yaml
@@ -1,7 +1,7 @@
 package:
   name: vault-1.14
   version: 1.14.3
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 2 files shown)


tcnghia:
May be rebase this on https://github.com/wolfi-dev/os/pull/5954 ?

vault-1.14.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: vault-1.14
   version: 1.14.3
```
tcnghia, line 3:
```suggestion
  version: 1.14.4
```

rawlingsj (approved):

PR #6165, opened 2023-10-02
Add vexctl package

jspeed-meyers:
This PR packages `vexctl`, a Go tool for creating, transforming, and attesting VEX metadata. It is part of the OpenVEX project.

#### For new package PRs only
- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`) (no support policy present)

### DIFF
--- a/vexctl.yaml
+++ b/vexctl.yaml
@@ -1,0 +1,32 @@
+package:
+  name: vexctl
+  version: 0.2.3
+  epoch: 0
+  description: A tool to create, transform and attest VEX metadata
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - go
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - go
+
+pipeline:
+  - uses: go/install
+    with:
+      package: github.com/openvex/vexctl
+      version: v${{package.version}}
+
+  - uses: strip
+
+update:
+  enabled: true
+  github:
+    identifier: openvex/vexctl
+    strip-prefix: v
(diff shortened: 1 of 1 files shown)


vexctl.yaml
```diff
@@ -0,0 +1,32 @@
+package:
+  name: vexctl
+  version: 0.2.3
+  epoch: 0
+  description: A tool to create, transform and attest VEX metadata
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - go
```
rawlingsj, line 10:
```suggestion
```
We can remove the runtime dependency on go

vexctl.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - go
```
rawlingsj, line 18:
```suggestion
```
This is automagically added by the [go/install](https://github.com/chainguard-dev/melange/blob/a4590d130eb51801d2819183e5cdc19fbcd079db/pkg/build/pipelines/go/install.yaml#L5) pipeline so can be removed

rawlingsj (changes requested):
looks great! just two minor comments

vexctl.yaml
imjasonh, line 10: Just checking: does this actually need go at runtime? Most Go apps don't need Go as a runtime dep, unless maybe they shell out to `go build` or something (like `ko` does).

This is as much an admission of my inexperience about vexctl as anything else.

jspeed-meyers: Will remove. Cooking up a commit! TY

vexctl.yaml
jspeed-meyers: Ohhh, too cool. Will remove.

jspeed-meyers: See commit b17865a

jspeed-meyers:
Thank you, @rawlingsj and @imjasonh.

rawlingsj (approved):

PR #6188, opened 2023-10-02
add cloud-sql-proxy package

k4leung4:
#### For new package PRs only


cloud-sql-proxy.yaml
```diff
@@ -0,0 +1,39 @@
+package:
+  name: cloud-sql-proxy
+  version: 2.7.0
+  epoch: 0
+  description: The Cloud SQL Auth Proxy is a utility for ensuring secure connections to your Cloud SQL instances
+  target-architecture:
```
imjasonh, line 6: Don't need this

cloud-sql-proxy.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - all
+  copyright:
+    - license: Apache-2.0
+  checks:
+    disabled:
+      - empty
```
imjasonh, line 12: What's empty?

imjasonh (changes requested):

cloud-sql-proxy.yaml
imjasonh: You'll need to write the binary to ${{targets.destdir}} to get it in the package and you can remove this.

cloud-sql-proxy.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/GoogleCloudPlatform/cloud-sql-proxy
+      tag: v${{package.version}}
+      expected-commit: 2527e5aedbca617579a528070e88a6e02c4d88d1
+
+  - runs: |
+      CGO_ENABLED=0 go build -ldflags "-X github.com/GoogleCloudPlatform/cloud-sql-proxy/v2/cmd.metadataString=container"
```
stormqueen1990, line 30: Any reason not to use the `go/build` pipeline here?

cloud-sql-proxy.yaml
k4leung4: this is the linter
the binary isn't placed in /usr/bin

cloud-sql-proxy.yaml
k4leung4: removed

cloud-sql-proxy.yaml
stormqueen1990: Also, this is going to output the binary in the workspace dir. IIRC that doesn't remain in the final package.

k4leung4: used go/build

cloud-sql-proxy.yaml
k4leung4: used go/build instead

k4leung4:
ptal

cloud-sql-proxy.yaml
```diff
@@ -0,0 +1,39 @@
...
+    with:
+      repository: https://github.com/GoogleCloudPlatform/cloud-sql-proxy
+      tag: v${{package.version}}
+      expected-commit: 2527e5aedbca617579a528070e88a6e02c4d88d1
+
+  - uses: go/build
+    with:
+      packages: .
+      output: cloud-sql-proxy
+      ldflags: -X github.com/GoogleCloudPlatform/cloud-sql-proxy/v2/cmd.metadataString=container
```
mattmoor, line 30: Can we add `-w`?

k4leung4: done

cloud-sql-proxy.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+  - uses: strip
```
mattmoor, line 32: Can you add a `compat` subpackage that depends on this, and makes the binary available with the expected name in `/`?

We should then add the compat package as a runtime dependency of https://github.com/wolfi-dev/os/blob/8f59ae1ede83ca7fce746b9610b09594b026ff9e/sigstore-scaffolding.yaml#L27

k4leung4: is the expected name `cloud-sql-proxy` or `cloudsqlproxy`?

is this what you had in mind?

cloud-sql-proxy.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+subpackages:
+  - name: ${{package.name}}-compat
+    description: Compatibility package to place binaries in the location expected by upstream helm charts
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}
+          ln -sf /usr/bin/cloud-sql-proxy ${{targets.subpkgdir}}/cloud-sql-proxy
```
mattmoor, line 40: Can you add a runtime dependency on `cloud-sql-proxy`?

Otherwise this is perfect: https://github.com/sigstore/scaffolding/blob/76f6bf19030792b4fa88fcffd188975ffe84514e/cmd/cloudsqlproxy/main.go#L34

k4leung4: added runtime dependency.

mattmoor (approved):

cloud-sql-proxy.yaml
```diff
@@ -0,0 +1,50 @@
...
+    dependencies:
+      provides:
+        - cloud-sql-proxy=${{package.full-version}}
```
imjasonh, line 43: I don't think this should provide the non-compat package, since it means we won't know whether we're getting the compat one or not.

Should this be `runtime` instead of `provides`?

mattmoor: oh absolutely, good catch.

mattmoor (dismissed):

cloud-sql-proxy.yaml
k4leung4: thanks, 
changed to `runtime`

cloud-sql-proxy.yaml
```diff
@@ -0,0 +1,50 @@
...
+subpackages:
+  - name: ${{package.name}}-compat
+    description: Compatibility package to place binaries in the location expected by upstream helm charts
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}
+          ln -sf /usr/bin/cloud-sql-proxy ${{targets.subpkgdir}}/cloud-sql-proxy
+    dependencies:
+      runtime:
+        - cloud-sql-proxy-compat
```
mattmoor, line 43: As is this is a dep on itself.
```suggestion
        - cloud-sql-proxy
```

k4leung4: thanks, fixed

imjasonh (approved):

Issue #6203, opened 2023-10-03
[Wolfi Sub Package Request]: $cargo


debasishbsws:
### What software would you like us to add to wolfi-os. Ideally include a URL to the project and its source.

Cargo is The Rust package manager. Currently Cargo comes within rust.
Rust is the core language compiler and runtime, whereas cargo is primarily a build tool used during development.
By separating the two, we enable more efficient image management, allowing users to opt for a smaller image without cargo when they only need the runtime.

### which versions of the software should we include?

Same as Rust version `1.72.1`

### Please check all that apply

- [X] This package has an un-restrictive license
- [X] The package/versions proposed are actively maintained upstream
- [X] I am interested in adding this package to Wolfi OS myself
- [ ] I am willing to help maintain this package

PR #6206, opened 2023-10-03
Cargo-1.72.1 new package(sub package of rust)

debasishbsws:
Related: #6203 

#### For new package PRs only

- [ ] The PR is marked as related to a pre-existing package request bug, such as a dependency

### DIFF
--- a/rust.yaml
+++ b/rust.yaml
@@ -112,6 +112,19 @@ pipeline:
       ln -srft usr/lib   usr/lib/rustlib/x86_64-unknown-linux-gnu/lib/*.so
       ln -srft usr/lib32 usr/lib/rustlib/i686-unknown-linux-gnu/lib/*.so
 
+subpackages:
+  - name: cargo
+    description: "Rust's package manager"
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/usr/bin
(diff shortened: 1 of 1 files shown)


debasishbsws:
I saw that in order to build rust package we need rust in the env packages. Isn't it a circular dependency? What is actually happening there?

debasishbsws:
Okay https://github.com/wolfi-dev/os/commit/14693edc2222086d3197800c4dd2fc8f6a9894f3 this explains my doubt.
> During bootstrapping, this must depend on rust-stage0, which builds a Rust  
  package using a prebuilt Rust toolchain provided by the Rust team.  
  After bootstrapping, rust can depend on a previously built Rust toolchain.  
  rust-stage0  
  
I am not sure, but I think here I need to create another PR adding `cargo` into the environment.contents.packages as envireonment dependencies, 
because we need cargo to build the `rust` and `cargo` binaries.  

And if this PR get merged we will not have the `cargo` binary inside `rust` package.

debasishbsws:
Hey, @imjasonh I am not sure why the tests are failing. I am not sure the reason for that is related to this or not.
Can I get a review at this?

debasishbsws:
It seems we will have a lot of problem as every other package that needs cargo and only have rust as a dependency will fail.

PR #6210, opened 2023-10-03
remove duplicate timestamp-authority-server package now that we have …

rawlingsj:
…a timestamp-authority which includes both cli and server subpackages

https://github.com/wolfi-dev/os/blob/702eb02a1ccfc5611f17d4e42cf083a56b5d0e3e/timestamp-authority.yaml#L35-L46


found-it:
This makes sense to me @rawlingsj. I realize you're just fixing the issue with duplicate packages but on the `timestamp-authority` package, won't `apk add timestamp-authority` just install nothing? Should it contain both binaries?

found-it (approved):

rawlingsj:
yeah - I agree, I don't think we have any rules around this but having top level runtime dependencies on the subpackages makes sense to me

PR #6228, opened 2023-10-03
Configure keda virtual packages

found-it:
Add support for KEDA version streams following upstream policy: https://github.com/kedacore/keda/blob/main/SECURITY.md#supported-versions

#### For new version streams

### DIFF
--- a/keda.yaml
+++ b/keda.yaml
@@ -38,12 +38,12 @@ pipeline:
 
 subpackages:
   - name: "${{package.name}}-adapter"
+    description: "Metrics adapter for Keda"
     dependencies:
       runtime:
         - tzdata
       provides:
         - keda-adapter=${{package.full-version}}
-    description: "Metrics adapter for Keda"
     pipeline:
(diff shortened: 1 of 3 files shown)


keda.yaml
```diff
@@ -1,13 +1,16 @@
 package:
   name: keda
```
mattmoor, line 2: Do you want to version stream this?

mattmoor: otherwise the provides here are not necessary.

found-it: This is where it's a little unclear to me. Yes I want to version stream it, but I also want it to track latest. So I don't want to call it `keda-2.12` since it will eventually switch to `2.13` or `3.0` right?

found-it: So this is the "latest version stream"

mattmoor: Yeah, it just makes it more complicated in images-private

mattmoor (approved):

PR #6281, opened 2023-10-04
fix: add WildFly OpenSSL runtime dependency

stormqueen1990:
Fix WildFly OpenSSL Wolfi package to include the `so:libssl.so.3` runtime dependency required by WildFly OpenSSL.

### DIFF
--- a/wildfly-openssl.yaml
+++ b/wildfly-openssl.yaml
@@ -1,10 +1,13 @@
 package:
   name: wildfly-openssl
   version: 2.2.5
-  epoch: 1
+  epoch: 2
   description: OpenSSL-based Java Secure Socket Extension implementation
   copyright:
     - license: Apache-2.0
+  dependencies:
+    runtime:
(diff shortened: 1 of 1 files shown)


pdeslaur (dismissed):

wildfly-openssl.yaml
```diff
@@ -1,10 +1,13 @@
...
   name: wildfly-openssl
   version: 2.2.5
-  epoch: 1
+  epoch: 2
   description: OpenSSL-based Java Secure Socket Extension implementation
   copyright:
     - license: Apache-2.0
+  dependencies:
+    runtime:
+      - so:libssl.so.3
```
pdeslaur, line 10: Actually, I think this needs to be `libssl3`

stormqueen1990: Fixed!

pdeslaur (approved):

PR #6295, opened 2023-10-05
build caddy with `CGO_ENABLED: 0`

mohammed90:
Caddy can be built without CGO. I noticed the caddy.yaml did not set the CGO_ENABLED variable, which leaves uncertainty on the table. Explicitly setting CGO_ENABLED to 0 to ensure Caddy build are static and CGO-free.

### DIFF
--- a/caddy.yaml
+++ b/caddy.yaml
@@ -1,7 +1,7 @@
 package:
   name: caddy
   version: 2.7.4
-  epoch: 2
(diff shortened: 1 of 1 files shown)


mohammed90:
The failure doesn't seem to be related to my change. Am I missing something?

ajayk (approved):

PR #6361, opened 2023-10-05
Revert "gitlab-runner/16.4.1 package update"

imjasonh:
They seem to have retracted this tag.


ajayk (dismissed):
We need to withdraw the package as well

ajayk (approved):

PR #6364, opened 2023-10-06
Change extracted filepath to align with original image

mamccorm:
"Unpack the files into a consistently named 'keycloak' directory. This approach mirrors the structure of the original image. Additionally, it enables us to utilize the 'paths:' feature in apko for setting file permissions, which requires a static path to the directory.

### DIFF
--- a/keycloak.yaml
+++ b/keycloak.yaml
@@ -40,16 +40,13 @@ pipeline:
 
       ./mvnw clean install -DskipTests=true -Dnode.version=$(node --version) -Pdistribution -q
 
-      # Mirrors the original images file structure.
-      mkdir -p ${{targets.destdir}}/opt
-      unzip -d ${{targets.destdir}}/opt quarkus/dist/target/keycloak-*.zip
-      mv ${{targets.destdir}}/opt/keycloak-* ${{targets.destdir}}/opt/keycloak
+      mkdir -p ${{targets.destdir}}/usr/share/java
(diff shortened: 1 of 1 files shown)


keycloak.yaml
```diff
@@ -40,12 +40,16 @@ pipeline:
...
+      unzip -d ${{targets.destdir}}/opt quarkus/dist/target/keycloak-*.zip
+      mv ${{targets.destdir}}/opt/keycloak-* ${{targets.destdir}}/opt/keycloak
 
+      # We don't use these symlinks in our images, but leaving in place so not
+      # to introduce a breaking change.
       mkdir -p ${{targets.destdir}}/usr/bin
       for i in kc.sh kcadm.sh kcreg.sh; do
-        ln -sf /usr/share/java/keycloak-${{package.version}}/bin/$i ${{targets.destdir}}/usr/bin/$i
+        ln -sf /opt/keycloak/bin/$i ${{targets.destdir}}/usr/bin/$i
       done
```
imjasonh, line 53: This will probably need to go in a `-compat` package, to appease the linter. Our stance is that non-`compat` packages shouldn't write to `/opt`

mamccorm: meh, figured it'd be better to keep in line with the pattern used by the source image. but if this will cause issues later with linters then not worth the hassle. I'll update to use /usr/local/bin instead

mamccorm: Sure, in that case i've backed out that particular change to avoid future linter issues

imjasonh (approved):

PR #6398, opened 2023-10-06
fix: php version streams

luhring:
This adjusts the names of subpackages for `php-8.1` and `php-8.2`, so that the subpackage names use the version stream name, and the `provides` is just `php-whatever`. This also takes advantage of the new `${{package.full-version}}` variable.

cc: @vaikas @mattmoor


vaikas (approved):
Woot Woot!!! Thanks!

PR #6409, opened 2023-10-06
lerna: new package

julienv3:
https://github.com/lerna/lerna

#### For new package PRs only

\* upstream support policy:
There is no apparent support policy on the project's [home page](https://lerna.js.org/) but the project hasn't been EOL'ed and still gets frequent [releases](https://github.com/lerna/lerna/releases).

### DIFF
--- a/lerna.yaml
+++ b/lerna.yaml
@@ -1,0 +1,39 @@
+package:
+  name: lerna
+  version: 7.3.0
+  epoch: 0
+  description: "Lerna is a fast, modern build system for managing and publishing multiple JavaScript/TypeScript packages from the same repository."
+  copyright:
+    - license: MIT
+  checks:
+    disabled:
+      - usrlocal
+  dependencies:
+    runtime:
+      - nodejs
+
+environment:
+  contents:
+    packages:
+      - bash
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - nodejs
+      - python3
+
+pipeline:
+  - name: npm install
+    runs: |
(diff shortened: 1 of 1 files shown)


lerna.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - nodejs
+      - python3
+
+pipeline:
+  - name: npm install
+    runs: |
+      npm install -g ${{package.name}}@${{package.version}} -prefix ${{targets.destdir}}/usr/local/
+
+      # https://github.com/browserify/resolve/issues/288
+      sed -i 's/monorepo-symlink-test/false-positive/g' ${{targets.destdir}}/usr/local/lib/node_modules/lerna/node_modules/resolve/test/resolver/multirepo/package.json
```
julienv3, line 31: Curious what people think of this, context:
The resolve npm package, with features involving parsing `package.json` files, has a test `package.json` file that ships with it and happens to have the same name as the now-known `monorepo-symlink-test` malware npm package.

Note that the [GHSA](https://github.com/advisories/GHSA-2jcg-qqmg-46q6) is very scary, so figured since this is just a test file with a predictable path I could simply replace the name to avoid the false positive in not only our scanner but other ones too and save everyone a potential future headache.

Alternative would be an advisory, but given the severity of the GHSA I figured avoiding any flagging at all in the first place would be better?  Thoughts?

julienv3: Sounds good, will send advisory separately!

imjasonh (approved):

PR #6410, opened 2023-10-06
cni-plugins: bump epoch

imjasonh:
Ironically, I forgot to bump the epoch in https://github.com/wolfi-dev/os/pull/6407/


joshrwolf (approved):
Ironically, I forgot to check that you bumped the epoch 🤦‍♂️

PR #6412, opened 2023-10-06
renovate

julienv3:
https://github.com/renovatebot/renovate

#### For new package PRs only

\* upstream support policy:
There is no apparent OSS-specific support policy on the project's [home page](https://www.mend.io/renovate/) but the project hasn't been EOL'ed and gets multiple daily [releases](https://github.com/renovatebot/renovate/releases).

### DIFF
--- a/renovate.yaml
+++ b/renovate.yaml
@@ -1,0 +1,37 @@
+package:
+  name: renovate
+  version: 37.6.0
+  epoch: 0
+  description: "Automated dependency updates. Multi-platform and multi-language."
+  copyright:
+    - license: AGPL-3.0
+  checks:
+    disabled:
+      - usrlocal
+  dependencies:
+    runtime:
+      - nodejs
+      - git
+
+environment:
+  contents:
+    packages:
+      - bash
+      - build-base
+      - busybox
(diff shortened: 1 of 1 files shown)


renovate.yaml
```diff
@@ -0,0 +1,37 @@
...
+      - busybox
+      - ca-certificates-bundle
+      - nodejs
+
+pipeline:
+  - name: npm install
+    runs: |
+      npm install -g ${{package.name}}@${{package.version}} -prefix ${{targets.destdir}}/usr/local/
+
+      # https://github.com/browserify/resolve/issues/288
```
julienv3, line 30: Same as 
https://github.com/wolfi-dev/os/pull/6409/files#r1349344680
Curious what people think?

julienv3: Thanks James, was chatting with Dan on the other thread and I agree creating the advisory is a good move too, will send one separately!
Sounds like we'd also (maybe temporarily?) like to do the `sed` too though from the other thread, please let me know if you think that's a bad idea and I'll adjust?

imjasonh (approved):

PR #6424, opened 2023-10-07
Revert batch 2 of the go update.

mattmoor:
ref: https://github.com/wolfi-dev/os/pull/6131

### DIFF
--- a/dataplaneapi.yaml
+++ b/dataplaneapi.yaml
@@ -1,7 +1,7 @@
 package:
   name: dataplaneapi
   version: 2.8.1
-  epoch: 3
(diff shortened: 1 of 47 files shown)


mattmoor:
As with the prior revert, we SHOULD NOT MERGE this if we can avoid it.  If https://github.com/wolfi-dev/os/actions/runs/6438771760 passes then we should close this, but I'm staging this to prepare for a scenario where that still fails.

imjasonh (approved):

dlorenc:
Do we need to merge this one? It looks like that other run Matt linked did pass?

PR #6454, opened 2023-10-08
always upload APKs we were able to build

imjasonh:
In cases where the build fails, we want to be able to at least upload APKs we were able to build, so we don't have to rebuild them next time, potentially never making progress.

If the build was not successful, for either leg, we shouldn't update and sign the APKINDEX, since this could lead to inconsistent APK versions between archs, which break our image builds.

We _could_ be smarter and add any consistent APKs between archs to the APKINDEX to make progress there too, but that's left for a future change.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -80,12 +80,14 @@ jobs:
             MELANGE_EXTRA_OPTS="--keyring-append=/gcsfuse/wolfi-registry/wolfi-signing.rsa.pub" \
             all -j1
 
-      - run: |
+      - if: always()
+        run: |
           # Clean up the symlinks and create an archive for uploading
           find ./packages/${{ matrix.arch }} -type l -exec rm -f {} \;
           tar -cvzf /tmp/packages-${{ matrix.arch }}.tar.gz ./packages/${{ matrix.arch }}
 
-      - name: 'Upload built packages archive to Github Artifacts'
+      - if: always()
+        name: 'Upload built packages archive to Github Artifacts'
         uses: actions/upload-artifact@v3
         with:
           name: packages-${{ matrix.arch }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -96,6 +96,7 @@ jobs:
   upload:
     runs-on: ubuntu-latest
     needs: build
+    if: always()
```
mattmoor, line 99: I think we need this on the last two steps of the workflow after `'Build Wolfi'` too

mattmoor: I think I actually wouldn't start by changing `upload` at all.

Let's ~fork it into a new second job that only runs `if: failure()` and dry runs 🤔

.github/workflows/build.yaml
```diff
@@ -137,13 +138,32 @@ jobs:
...
+        if: always()
+        run: |
+          for arch in "x86_64" "aarch64"; do
+          mkdir -p ./packages/${arch}
+
+          # Consolidate with the built artifacts
+          tar xvf /tmp/artifacts/packages-${arch}.tar.gz
+
+            # Upload any APKs we were able to build.
+            apks=$(ls ./packages/${arch}/*.apk 2>/dev/null)
```
mattmoor, line 151: The indentation here is strange

.github/workflows/build.yaml
```diff
@@ -137,13 +138,32 @@ jobs:
...
+            if [ -n "$apks" ]; then
```
mattmoor, line 152: I'm worried about uploading APKs that aren't present for all archs because I believe if we withdraw in that state the index will be recomputed and we'll end up in an inconsistent state?

mattmoor: I think we also need to deal with some cases where APKs only exist for a single arch in enterprise too, so IDK how we reconcile those 🙃

PR #6473, opened 2023-10-09
libmilter/1.0.3 package update

### LINKED ISSUE #6258: libmilter/8.17.2 new package update

cpanato:
- libmilter/1.0.3 package update

Fixes: https://github.com/wolfi-dev/os/issues/6258

Related:

#### For version bump PRs

### DIFF
--- a/libmilter.yaml
+++ b/libmilter.yaml
@@ -1,6 +1,6 @@
 package:
   name: libmilter
-  version: 1.0.2
+  version: 1.0.3
   epoch: 0
   description: Sendmail Mail Filter API (Milter)
   copyright:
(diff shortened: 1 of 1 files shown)


libmilter.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: libmilter
-  version: 1.0.2
+  version: 1.0.3
```
cpanato, line 3: did not find where this version come from, alpine does the same, so since this is a patch version of sendmail we do a patch bump here as well

imjasonh: I see 1.0.2 in Alpine edge: https://pkgs.alpinelinux.org/package/edge/main/x86_64/libmilter

Where do you see Alpine calling this 1.0.3 ?

cpanato: i mean they are using the same version.
but they did not update to `sendmail.8.17.2` it is still on `sendmail.8.17.1`

rawlingsj:
I'm not sure about the versions here, I wonder if libmilter should use the sendmail version (8.17.2) rather than 1.0.3?

rawlingsj (approved):

PR #6476, opened 2023-10-09
`gitlab-logger`: new package

mattmoor:
I went with `fetch` because that's what the upstream build process does, but if it makes adding automated updates easier we can switch this to use `git-checkout`: https://gitlab.com/gitlab-org/build/CNG/-/blob/master/gitlab-logger/Dockerfile#L10

#### For new package PRs only

### DIFF
--- a/gitlab-logger.yaml
+++ b/gitlab-logger.yaml
@@ -1,0 +1,29 @@
+package:
+  name: gitlab-logger
+  version: 3.0.0
+  epoch: 0
+  description: GitLab Logger provides a means of wrapping non-structured log files within structure JSON.
+  copyright:
+    - license: MIT
+
+environment:
+  environment:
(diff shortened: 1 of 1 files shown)


gitlab-logger.yaml
```diff
@@ -0,0 +1,33 @@
...
+  epoch: 0
+  description: GitLab Logger provides a means of wrapping non-structured log files within structure JSON.
+  copyright:
+    - license: MIT
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - go
```
imjasonh, line 13: (not blocking) Both of these should come in from depending on `go/build`, so you could just remove this.

imjasonh (dismissed):

imjasonh (approved):

PR #6477, opened 2023-10-09
Sprinkle `strip` across a bunch of Go packages.

mattmoor:
Also drop `-s` so that we can strip while preserving symbols.

This isn't all of the instances, but I'm staging a first shard.

### DIFF
--- a/aws-ebs-csi-driver.yaml
+++ b/aws-ebs-csi-driver.yaml
@@ -1,7 +1,7 @@
 package:
   name: aws-ebs-csi-driver
   version: 1.23.1
-  epoch: 1
+  epoch: 2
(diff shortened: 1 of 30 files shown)


imjasonh (approved):

jonjohnsonjr:
```
time="2023-10-09T16:54:12Z" level=fatal msg="error during command execution: 1 error occurred:\n\t* package kubeflow-pipelines: update found newer version 2.3.0 compared with package.version 2.1.3 in melange config\n\n"
```

mattmoor:
@jonjohnsonjr I know, bumping the package is out of scope for this change.  I'm leaving that to the automation.

PR #6483, opened 2023-10-09
always upload APKs we were able to build (take 2)

imjasonh:
Starting https://github.com/wolfi-dev/os/pull/6454 from scratch

This splits apk and APKINDEX uploading into separate steps, and copies the `upload` workflow into a separate run-on-failure workflow that just uploads apks.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -153,26 +153,26 @@ jobs:
             # Sign the APK index
             melange sign-index -f --signing-key ./wolfi-signing.rsa packages/${arch}/APKINDEX.tar.gz
 
-            # Only attempt to sign when *.apk's exist.
+            # Only attempt to sign and upload APKINDEX when *.apk's changed.
             apks=$(ls ./packages/${arch}/*.apk 2>/dev/null || true)
             if [ -n "$apks" ]; then
               melange sign --signing-key ./wolfi-signing.rsa ./packages/${arch}/*.apk
+
+              # Don't cache the APKINDEX.
+              gcloud --quiet storage cp \
+                  --cache-control=no-store \
+                  "./packages/${arch}/APKINDEX.tar.gz" "gs://wolfi-production-registry-destination/os/${arch}/"
+
+              gcloud --quiet storage cp \
+                  --cache-control=no-store \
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -153,26 +155,90 @@ jobs:
...
               melange sign --signing-key ./wolfi-signing.rsa ./packages/${arch}/*.apk
+
+              # Don't cache the APKINDEX.
+              gcloud --quiet storage cp \
+                  --cache-control=no-store \
+                  "./packages/${arch}/APKINDEX.tar.gz" "gs://wolfi-production-registry-destination/os/${arch}/"
+
+              gcloud --quiet storage cp \
+                  --cache-control=no-store \
+                  "./packages/${arch}/APKINDEX.json" "gs://wolfi-production-registry-destination/os/${arch}/"
```
mattmoor, line 170: I think this was intentionally outside of the loop to minimize the time the indices might be inconsistent

mattmoor: Do we upload the index before the APKs?  That doesn't seem like something you are changing, but it seems like a terrible idea 😅

.github/workflows/build.yaml
```diff
@@ -153,26 +155,90 @@ jobs:
...
+      - name: 'Download aarch64 package archives'
+        uses: actions/download-artifact@v3
+        with:
+          path: /tmp/artifacts/
+          name: packages-aarch64
+
+      - run: echo "${{ secrets.MELANGE_RSA }}" > ./wolfi-signing.rsa
+      - run: |
+          mkdir -p /etc/apk/keys
+          cp ./wolfi-signing.rsa.pub /etc/apk/keys/wolfi-signing.rsa.pub
```
mattmoor, line 236: You aren't actually signing with this, just uploading the APKs

.github/workflows/build.yaml
```diff
@@ -153,26 +155,90 @@ jobs:
...
+
+      - name: 'Upload the repository to the bucket'
+        run: |
+          for arch in "x86_64" "aarch64"; do
+            # Only attempt to upload when *.apk's changed.
```
mattmoor, line 241: I feel like here we should walk `x86_64` and only upload if `./packages-aarch64/$(basename $pkg)` exists, WDYT?

PR #6516, opened 2023-10-10
curl-rustls: disable empty linter

imjasonh:
The curl-rustls package is expected to be empty, it brings in a runtime dep on its subpackage, `libcurl-rustls4`.

This PR also adds a flag to `make` to re-enable lint checks, which were disabled by default before.


mamccorm (approved):
looks good

pdeslaur (approved):

PR #6559, opened 2023-10-10
python: make 3.12 the latest python-3

imjasonh:
A future change will rebuild pip/setuptools/installer with this latest 3.12 and make those the latest for py3-pip et al.

### DIFF
--- a/meson.yaml
+++ b/meson.yaml
@@ -1,13 +1,13 @@
 package:
   name: meson
   version: 1.2.2
-  epoch: 1
+  epoch: 0
   description: Fast and user friendly build system
   copyright:
     - license: Apache-2.0
   dependencies:
     runtime:
-      - python-3.11
+      - python3
       - samurai
 
 environment:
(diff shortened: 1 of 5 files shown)


py3.12-installer.yaml
```diff
@@ -6,9 +6,8 @@ package:
   copyright:
     - license: "MIT"
   dependencies:
-    # TODO(jason): Uncomment when py-3.12 is default everywhere.
-    #provides:
-    #  - py3-installer=${{package.version}}
+    provides:
+      - py3-installer=${{package.version}}
```
mattmoor, line 10: these all need epoch bumps

meson.yaml
```diff
@@ -7,17 +7,18 @@ package:
     - license: Apache-2.0
   dependencies:
     runtime:
-      - python3
+      - python-3.11
```
mattmoor, line 10: Should this bump epoch?

meson.yaml
```diff
@@ -7,17 +7,18 @@ package:
...
 
 environment:
   contents:
     packages:
       - busybox
       - ca-certificates-bundle
-      - py3-setuptools
-      - python3
+      - py3.11-setuptools
       - samurai
```
mattmoor, line 19: Is this not the standard way we bootstrap?

kaniini: I think this is perhaps a little suboptimal and we might want to look at making the solver in wolfictl a little more intelligent to deal with this edge case, but seems fine for now.

kaniini (approved):

PR #6593, opened 2023-10-11
add redis-bitnami-compat and reusable bitnami/compat pipeline

joshrwolf:
adds a bitnami compatibility package, alongside how we would use it for `redis`

The goal with the first instance of the compatibility package pipeline is to be as wide as possible, over time we can refine it to get more savings.

This redis instance includes additional symlinks that are not always present in the bitnami images, so in this case they're added after the `uses: bitnami/compat`.


pipelines/bitnami/compat.yaml
```diff
@@ -0,0 +1,28 @@
...
+  version-path:
+    description: |
+      The path to the root of the Dockerfile, relative to the image name.
+    required: true
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/bitnami/containers
+      branch: main
```
imjasonh, line 22: Well I don't love this. 😓 

Is this something we can automate bumping? I guess we'd also need to bump packages that use this pipeline to rebuild them with the new hotness.

joshrwolf: yeah, it's a mono with no tags/releases. The only signal we have for bumping is the commit message, which we can debate over how it gets into wolfictl (cc @rawlingsj)

imjasonh: Would it make sense to have an expected-commit on this, so we at least fail if it moves, and know to manually bump+rebuild things?

Let's unblock with something short-term and gross and hope we can come up with something better long term.

(Hello Jason from the future! If you find this, I hope that it brings you joy at least.)

joshrwolf: I started there, but in the time it took me to put together this PR was hit by 4-5 updates irrelevant to `redis` 😅

imjasonh (dismissed):

imjasonh (approved):

PR #6617, opened 2023-10-11
try to fix pipenv/poetry conflicting on filelock

imjasonh:
Both have pip dependencies on filelock, so both pull it in at `pip install` time and vendor it, at different versions.

This makes both packages depend on filelock at build time (so they don't try to install/vendor it) and at runtime (so they have it at runtime when they need it).

I'm a Python noob, I'd love someone who knows what they're talking about to tell me if this is wrong.

### DIFF
--- a/poetry.yaml
+++ b/poetry.yaml
@@ -1,13 +1,14 @@
 package:
   name: poetry
   version: 1.6.1
-  epoch: 1
+  epoch: 2
   description: "Python packaging and dependency management made easy"
   copyright:
     - license: MIT
   dependencies:
     runtime:
       - python3
+      - py3-filelock
 
 environment:
   contents:
@@ -18,6 +19,7 @@ environment:
       - python3
       - py3-pip
       - py3-setuptools
+      - py3-filelock
 
 pipeline:
   - uses: fetch
--- a/py3-pipenv.yaml
+++ b/py3-pipenv.yaml
@@ -1,13 +1,14 @@
 package:
   name: py3-pipenv
   version: 2023.10.3
-  epoch: 0
+  epoch: 1
   description: "Pipenv is a Python virtualenv management tool that supports a multitude of systems and nicely bridges the gaps between pip, python (using system python, pyenv or asdf) and virtualenv."
   copyright:
     - license: MIT
   dependencies:
     runtime:
       - python3
+      - py3-filelock
 
 environment:
   contents:
@@ -16,6 +17,7 @@ environment:
       - ca-certificates-bundle
       - build-base
       - py3-pip
+      - py3-filelock
 
 pipeline:
   - uses: git-checkout


poetry.yaml
```diff
@@ -1,13 +1,14 @@
...
   version: 1.6.1
-  epoch: 1
+  epoch: 2
   description: "Python packaging and dependency management made easy"
   copyright:
     - license: MIT
   dependencies:
     runtime:
       - python3
+      - py3-filelock
```
julienv3, line 11: So those two packages actually share at least 5 deps; diffing `apk manifest <apk> | grep -o site-packages/[^\/]* | uniq` yields
```
site-packages/certifi
site-packages/distlib
site-packages/filelock
site-packages/platformdirs
site-packages/virtualenv
```
IIUC, apko complained about `filelock` as it was installed as two different versions for [poetry](https://apk.dag.dev/https/packages.wolfi.dev/os/aarch64/poetry-1.6.1-r1.apk@sha1:a[%E2%80%A6]acb048830a/usr/lib/python3.11/site-packages/filelock/version.py) and [py3-pipenv](https://apk.dag.dev/https/packages.wolfi.dev/os/aarch64/py3-pipenv-2023.10.3-r0.apk@sha1:96d51a39b80b3a55912ad7d58476820e2244ac51/usr/lib/python3.11/site-packages/filelock/version.py).

From looking at `site-packages` for [poetry](https://apk.dag.dev/https/packages.wolfi.dev/os/aarch64/poetry-1.6.1-r1.apk@sha1:a%5B%E2%80%A6%5Dacb048830a/usr/lib/python3.11/site-packages/) and [pipenv](https://apk.dag.dev/https/packages.wolfi.dev/os/aarch64/py3-pipenv-2023.10.3-r0.apk@sha1:96d51a39b80b3a55912ad7d58476820e2244ac51/usr/lib/python3.11/site-packages/) it looks like this may be an issue for other deps, like platformdirs ([poetry](https://apk.dag.dev/https/packages.wolfi.dev/os/aarch64/poetry-1.6.1-r1.apk@sha1:a%5B%E2%80%A6%5Dacb048830a/usr/lib/python3.11/site-packages/platformdirs/version.py) vs [pipenv](https://apk.dag.dev/https/packages.wolfi.dev/os/aarch64/py3-pipenv-2023.10.3-r0.apk@sha1:96d51a39b80b3a55912ad7d58476820e2244ac51/usr/lib/python3.11/site-packages/platformdirs/version.py))

imjasonh: These seem to all come from https://github.com/wolfi-dev/os/blob/main/py3-virtualenv.yaml

julienv3: Alright so using `pip install poetry` then `pip show poetry`, the actual list of deps is shown to be (`py3-` prefixes added)
```
- py3-build
- py3-cachecontrol
- py3-cleo
- py3-crashtest
- py3-dulwich
- py3-installer
- py3-jsonschema
- py3-keyring
- py3-packaging
- py3-pexpect
- py3-pkginfo
- py3-platformdirs
- py3-poetry-core
- py3-poetry-plugin-export
- py3-pyproject-hooks
- py3-requests
- py3-requests-toolbelt
- py3-shellingham
- py3-tomlkit
- py3-trove-classifiers
- py3-virtualenv
```

That brings up two followup issues.
1, we don't have all these packages, but we can probably build them.
2, those dep requests are [versioned](https://pypi.org/pypi/poetry/1.6.1/json) ;
```
requires_dist: [
  "poetry-core (==1.7.0)",
  "poetry-plugin-export (>=1.5.0,<2.0.0)",
  "build (>=0.10.0,<0.11.0)",
  ...
]
```
and taking `build` == `py3-build` as an example we have version 1.0.3 which won't satisfy the request.

julienv3: Also note that adding `py3-virtualenv` causes a mix of vendor'ed and dependent packages to clash within poetry itself, causing on install errors like:
```
ERROR: poetry-1.6.1-r9: trying to overwrite usr/lib/python3.11/site-packages/importlib_metadata/_meta.py owned by py3-importlib-metadata-6.8.0-r0.
```

Unclear why.  At build time, both `py3-virtualenv` and `py3-importlib-metadata` are installed (because the former has a runtime dep on the latter):
```
❕ aarch64   | installing py3-importlib-metadata (6.8.0-r0)
❕ aarch64   | installing py3-virtualenv (20.24.5-r0)
```
And then `pip install` is happy with the installed `virtualenv`
```
ℹ️  aarch64   | Requirement already satisfied: virtualenv<21.0.0,>=20.22.0 in /usr/lib/python3.11/site-packages (from poetry==1.6.1) (20.24.5)
```
but not with the installed `importlib-metadata`
```
ℹ️  aarch64   | Collecting importlib-metadata>=4.11.4 (from keyring<25.0.0,>=24.0.0->poetry==1.6.1)
ℹ️  aarch64   |   Obtaining dependency information for importlib-metadata>=4.11.4 from https://files.pythonhosted.org/packages/cc/37/db7ba97e676af155f5fcb1a35466f446eadc9104e25b83366e8088c9c926/importlib_metadata-6.8.0-py3-none-any.whl.metadata
ℹ️  aarch64   |   Downloading importlib_metadata-6.8.0-py3-none-any.whl.metadata (5.1 kB)
```

imjasonh: Well this is fun. 🤣

Re: versioning: for ruby (where we also package language deps), there's a pipeline to relax the version constraints to accept our latest versions. There's problems with that too.

Another problem is we're flying blind wrt whether anything we're doing actually works. Can we install poetry and virtualenv today? We don't know until we build an image and it fails. We should have some kind of simple tests for this.

I think we might need to either make 1000% sure we don't vendor py packages, which has downsides, or do what pip wants and vendor, and figure out how to make that work too.

julienv3: Short term: #6658 causes “pip install” to reevaluate all dependency requests, which will match all the versions we got for pipenv dependencies because that was built yesterday, and so while apko will still have multiple vendored copies to install the hashes will match and it will be happy.  This also fixes 2 CVEs via the same side effect.

Long term: two possible[1] solutions (huge AFAIK/IIUC caveats)

A) We correctly mark all python package dependencies as build&runtime dependencies, so a py3 package never vendors anything.
That involves version relaxing, because otherwise poetry needs build<=0.11 but virtualenv needs build>=1.0 and so installing both has a conflict since “root” dependencies and “vendored” dependencies share the same installed location, /usr/lib/python3.11/site-packages/.
[1] Is that possible though? Problem with that is poetry might simply not work with build>=1.0 due to breaking changes since build<=0.11, so what do we do then.

B) We figure out a way to vendor everything all the time.  That’s the npm model, e.g. npm install --global lerna causes the addition of /usr/lib/node_modules/lerna/node_modules/all_the_vendored_dep ; since every dep is nested, module X will never conflict with module Y.

Looking around, pipx seems to be exactly B.
https://pypa.github.io/pipx/
https://python.land/virtual-environments/pipx

PR #6618, opened 2023-10-11
runc: fix GHSA-6xv5-86q9-7xr8

wlynch:
Fixes: https://github.com/advisories/GHSA-6xv5-86q9-7xr8

#### For security-related PRs

Notes: We've already [NACKed this](https://github.com/wolfi-dev/advisories/pull/232), but it's showing up in grype scans and it's easy to patch out 🤷

### DIFF
--- a/runc.yaml
+++ b/runc.yaml
@@ -1,7 +1,7 @@
 package:
   name: runc
   version: 1.1.9
-  epoch: 4
+  epoch: 5
   description: CLI tool for spawning and running containers according to the OCI specification
   copyright:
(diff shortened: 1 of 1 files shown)


ajayk:
https://github.com/advisories/GHSA-6xv5-86q9-7xr8 is applicable for windows  this can be NAC'ked

wlynch:
Follow up off-thread: we're okay with grype reporting CVEs from go-module if we've nacked it for the package. We're working with grype to try to improve this from the vuln scanner side.

PR #6646, opened 2023-10-12
gitlab-kas: bump its version

hectorj2f:
Fixes:

Related: When deploying the Gitlab helm chart, I found issues related to the current package version. The default deployed version uses the version `v16.4.0`

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/gitlab-kas.yaml
+++ b/gitlab-kas.yaml
@@ -1,6 +1,6 @@
 package:
   name: gitlab-kas
-  version: 14.5.0
+  version: 16.4.0
   epoch: 0
   description: GitLab Kas is a component installed together with GitLab. It is required to manage the GitLab agent for Kubernetes.
   copyright:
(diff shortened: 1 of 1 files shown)


gitlab-kas.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: gitlab-kas
-  version: 14.5.0
+  version: 16.4.0
```
tcnghia, line 3: This package has `update` disabled ... is it possible to enable update for this, or add a comment to explain why we can't?

hectorj2f: Let's enable it. Question: Which value do we set to the `release-monitor` ?

hectorj2f: Okay, I figured it out. I couldn't find a release-monitor for this project. What should we do under these situations ?

cpanato (approved):

PR #6680, opened 2023-10-12
remove some py3.11 requirements that should float to 3.12

imjasonh:


imjasonh:
bringing this into https://github.com/wolfi-dev/os/pull/6679

PR #6681, opened 2023-10-12
py3.12: bump epochs

imjasonh:


mattmoor (dismissed):

glibc.yaml
```diff
@@ -34,7 +34,7 @@ environment:
       - bison
       - flex
       - grep
-      - python3
+      - python-3.11 # undo this
```
imjasonh, line 37: This is just to trick `wolfictl text` into ignoring a cycle it thinks it has. We're not bumping the epoch, so the package won't get rebuilt with this dep, and we can undo this in a followup PR (also not bumping epoch) to avoid any confusion.

luhring (approved):

imjasonh:
skipping merge queue to save time

PR #6718, opened 2023-10-13
wolfi bot: schedule release monitor job to run hourly now we have imp…

rawlingsj:
…roved api limits using an api key

### DIFF
--- a/.github/workflows/wolfictl-update-rm.yaml
+++ b/.github/workflows/wolfictl-update-rm.yaml
@@ -2,9 +2,9 @@ name: Wolfictl Update From Release Monitor
 
 on:
   workflow_dispatch:
-  # Triggers the workflow daily at 4:30pm UTC
+  # Triggers the workflow every hour
   schedule:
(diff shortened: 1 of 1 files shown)


mattmoor (approved):
Is this configured to notify slack on failures?  Wondering how our monitoring is for this process 🤔

rawlingsj:
> Is this configured to notify slack on failures? Wondering how our monitoring is for this process

There are the occasional failures which tend to always be network related or 503 when hitting github or release monitor.  As they trigger often the next attempt seems to always work.  We get emails from GitHub when they fail, which is what I've been using till now.  There's also one intermittent git related failure which I've never been able to recreate out of GitHub Actions.

I mentioned to @imjasonh earlier this week I'd like to move these jobs to Cloud Run and have proper monitoring and alerting there.  I think I can get that done Monday / Tuesday next week.  Does that sound ok to you rather than adding the slack stuff here now?

mattmoor:
I was mostly curious, not a rush.

PR #6799, opened 2023-10-14
Remove some obsolete go module pinning

tcnghia:
This removes some obsolete patches, as @luhring [patched](https://github.com/wagoodman/dive/commit/6884c46d0fb89fc4cf07423c0743848ac13aa66f) upstream


k4leung4 (approved):
thanks

confirmed that deps are at equal or newer versions in the repo

PR #6805, opened 2023-10-15
new package: py3-2to3

imjasonh:


py3-2to3.yaml
```diff
@@ -0,0 +1,32 @@
...
+pipeline:
+  - uses: fetch
+    with:
+      expected-sha256: 958bc212c928bbdcbc778b72528e0a39ae4ee8040eda6af6c3b5dd640c98ce6d
+      uri: https://files.pythonhosted.org/packages/70/06/9bd891a16b16572e42d124627048a673f9e201c490ef1638b689a94cacd7/2to3-${{package.version}}.tar.gz
+
+  - name: Python Build
+    uses: python/build-wheel
+
+  - uses: strip
```
mattmoor, line 32: No update block?

imjasonh: Sounds like 2to3 is installed along with python itself, and this project just makes it available as a script?

2to3 is also deprecated and will be removed in 3.13, so I don't think this will get updates.

PR #6896, opened 2023-10-16
zookeper-3.8 and 3.9 with bitnami compliant subpackages

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/zookeeper-3.8.yaml
+++ b/zookeeper-3.8.yaml
@@ -1,17 +1,16 @@
 package:
-  name: zookeeper
-  version: 3.9.1.0
-  epoch: 1
+  name: zookeeper-3.8
+  version: 3.8.3.0
+  epoch: 0
   description: Distributed, highly available, robust, fault-tolerant system for distributed coordination
   copyright:
-    - paths:
-        - "*"
-      attestation:
-      license: Apache-2.0
+    - license: Apache-2.0
   dependencies:
     runtime:
       - bash # some helper scripts use bash
       - openjdk-17-default-jvm
+    provides:
+      - zookeeper=${{package.full-version}}
 
 var-transforms:
   - from: ${{package.version}}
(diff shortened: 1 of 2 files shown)


zookeeper-3.8.yaml
```diff
@@ -0,0 +1,80 @@
+package:
+  name: zookeeper-3.8
+  version: 3.8.3.0
+  epoch: 0
+  description: Distributed, highly available, robust, fault-tolerant system for distributed coordination
+  copyright:
+    - paths:
+        - "*"
+      attestation:
+      license: Apache-2.0
```
imjasonh, line 10:
```suggestion
    - license: Apache-2.0
```

zookeeper-3.8.yaml
```diff
@@ -0,0 +1,80 @@
...
+  dependencies:
+    runtime:
+      - bash # some helper scripts use bash
```
imjasonh, line 13: Does zookeeper itself need bash, or does bitnami bits need bash?

If it's only bitnami that needs it, we can add this as a runtime dep of the compat package below.

zookeeper-3.9.yaml
```diff
@@ -1,5 +1,5 @@
 package:
-  name: zookeeper
+  name: zookeeper-3.9
```
imjasonh, line 2: We should have this provide `zookeeper=${{package.full-version}}` so that folks doing `apk add zookeeper` keep getting updates to this version stream.

developer-guy: nice, good catch!

zookeeper-3.8.yaml
developer-guy: I didn't change anything about the old package config, I think its coming from the old config, so I believe its the package we need in zookeeper at runtime.

imjasonh (approved):

PR #6909, opened 2023-10-16
Add liblbfgs

Dentrax:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/liblbfgs.yaml
+++ b/liblbfgs.yaml
@@ -1,0 +1,58 @@
+package:
+  name: liblbfgs
+  version: 1.10
+  epoch: 0
+  description: A library of Limited-memory Broyden-Fletcher-Goldfarb-Shanno (L-BFGS)
+  copyright:
+    - license: MIT
+
+environment:
+  contents:
+    packages:
+      - wolfi-base
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+      - automake
+      - autoconf
+      - libtool
+      - m4
+
+pipeline:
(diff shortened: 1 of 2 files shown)


liblbfgs.yaml
```diff
@@ -0,0 +1,60 @@
...
+      - autoconf
+      - libtool
+      - m4
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/chokkan/liblbfgs
+      branch: master
+      expected-commit: 5ad02fbefefdeff339ab03635e673571055a0644
```
rawlingsj, line 27: I'm not sure this the best approach, if there's a new commit into the upstream main branch our builds will fail because of the clone of the main branch and expected commit will not match.

An alternative is to use a `fetch` for the exact commit.  i.e.
```
https://github.com/chokkan/liblbfgs/archive/5ad02fbefefdeff339ab03635e673571055a0644.tar.gz
```
we can make the commit a pipeline var too so it's easier to update, i.e. similar to this https://github.com/wolfi-dev/os/blob/02718ca4af87e0295b4e54080f6c62b8260eb699/ingress-nginx-controller.yaml#L201-L205

Dentrax: Thanks for the comment, that make sense! Ack.

rawlingsj (approved):

PR #6964, opened 2023-10-17
Add Falco

### LINKED ISSUE #4342: [Wolfi Package Request]: falcosecurity/falco

https://github.com/falcosecurity/falco

latest

```
grype falcosecurity/falco:latest
 ✔ Vulnerability DB                [updated]
 ✔ Parsed image                                                                                       sha256:928ed230b0b66cbc1cb79191438b71ab89cf7823609744f29afeb6c4ebe1af80
 ✔ Cataloged packages              [287 packages]
 ✔ Scanned for vulnerabilities     [753 vulnerabilities]
   ├── 0 critical, 110 high, 99 medium, 46 low, 488 negligible (10 unknown)
   └── 43 fixed
```

- [X] The package/versions proposed are actively maintained upstream
- [X] I am interested in adding this package to Wolfi OS myself
- [X] I am willing to help maintain this package

Dentrax:
> **Note**
> **Before:** `973MB, 366 vulnerability`
> **After:** `6.6MB, 0-CVE`

Fixes: #4342

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -1,0 +1,108 @@
+package:
+  name: falco
+  version: 0.36.1
+  epoch: 0
+  description: Cloud Native Runtime Security
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - falco-rules
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - clang-16
+      - gcc
+      - cmake
+      - make
+      - git
+      - bash
+      - perl
+      - linux-headers
+      - autoconf
+      - automake
+      - m4
+      - libtool
+      - elfutils-dev
+      - libelf
+      - libelf-static
+      - patch
+      - binutils
+      - libbpf
+      - openssl
+      - openssl-dev
+      - yaml-dev
+      - c-ares
+      - c-ares-dev
+      - protobuf
+      - protobuf-dev
+      - protobuf-c-dev
+      - re2
+      - re2-dev
+      - zlib-dev
+      - libcurl-openssl4
+      - llvm16
+      - abseil-cpp
+      - abseil-cpp-dev
+      - abseillib
+      - jq-dev
+      - curl-dev
+      - grpc
+      - grpc-dev
+      - icu
+      - icu-dev
+      - yaml-cpp
+      - yaml-cpp-dev
+      - systemd-dev
+      - libsystemd
+      - libzstd1
+      - zstd
+      - zstd-dev
+      - libtbb-dev
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/falcosecurity/falco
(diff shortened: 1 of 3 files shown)


falco.yaml
```diff
@@ -0,0 +1,121 @@
...
+      mkdir -p "${{targets.destdir}}"/etc/falco
+      install -Dm755 ./falco.yaml "${{targets.destdir}}"/etc/falco/falco.yaml
+      sed -e 's/time_format_iso_8601: false/time_format_iso_8601: true/' < "${{targets.destdir}}"/etc/falco/falco.yaml
+
+  - working-directory: /home/build/rules
+    pipeline:
+      - uses: git-checkout
+        with:
+          repository: https://github.com/falcosecurity/rules
+          tag: falco-rules-${{vars.RULES_VERSION}}
```
imjasonh, line 89: This is going to require manual updates since the wolfibot won't know how to update it.

Would it make sense to have `falco-rules.yaml` as a separate package, that only contained these rules definitions?

vaikas: Are the rules something that we expect users of the package to modify? In addition to what @imjasonh said, if we make this subpackage it's easier for folks to update it themselves.

falco.yaml
```diff
@@ -0,0 +1,121 @@
...
+      tag: ${{package.version}}
+      expected-commit: 9eb611609a2876a5f5a5378e0613f0ff767f0d42
+      recurse-submodules: true
+
+  - runs: |
+      # Replace the find_dependency with find_package macro for newer cmake, otherwise it will fail.
+      sed -i 's/find_dependency(Protobuf CONFIG)/find_package(Protobuf CONFIG)/' /usr/lib64/cmake/grpc/gRPCConfig.cmake
+
+  - runs: |
+      mkdir -p "${{targets.destdir}}"/etc/falco
```
vaikas, line 80: Same here as below, but is this something that we expect users to customize? Maybe turn this into subpackage also?

falco.yaml
Dentrax: > Are the rules something that we expect users of the package to modify?

Yep, correct. But Falco also provides default rules.

> Would it make sense to have falco-rules.yaml as a separate package

It does make sense.

I was thinking if we should provide `falco_rules` by default with the `falco` package itself. If so, we should make separate `falco-rules` package; otherwise a sub-pkg is way to go.

Should we provide this as a sub-pkg?

Am I allowed to use `patch` in sub-pkg?

imjasonh: You can use patch in a subpackage.

But since it comes from a separate repo, I think falco-rules belongs as a toplevel package, so they can be updated automatically

imjasonh (approved):

kaniini (approved):

PR #6968, opened 2023-10-17
gitlab: add gitlab-base

hectorj2f:
Fixes:

Related: https://github.com/chainguard-dev/image-requests/issues/457

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/gitlab-base.yaml
+++ b/gitlab-base.yaml
@@ -1,0 +1,25 @@
+package:
+  name: gitlab-base
+  # The release-monitor identified is for the gitlab-runner,
+  # but these seem to be co-versioned.
+  version: 16.4.1
+  epoch: 0
+  description: GitLab Base is the base image, which other GitLab containers can build on top of.
+  copyright:
+    - license: MIT
+  dependencies:
+    runtime:
+      - bash
+      - busybox
+      - ca-certificates-bundle
+      - curl
+      - gitlab-cng-base-scripts
+      - gitlab-logger
+      - gomplate
+      - procps
+      - xtail
+
+update:
+  enabled: true
+  release-monitor:
+    identifier: 11337


gitlab-base.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+pipeline:
+  - runs: |
+      mkdir -p ${{targets.destdir}}/usr/local/bin/ ${{targets.destdir}}/scripts
+      install -m755 -D "./exec-env" ${{targets.destdir}}/scripts
+      install -m755 -D "./set-config" ${{targets.destdir}}/scripts
+      install -m755 -D "./entrypoint.sh" ${{targets.destdir}}/scripts
+  - uses: strip
+
+update:
```
hectorj2f, line 31: Do we want to create a new project in release-monitoring ? Or follow other gitlab version updates ?

hectorj2f: I'll reuse the one from cng as it looks similar.

gitlab-base.yaml
```diff
@@ -0,0 +1,34 @@
+package:
+  name: gitlab-base
+  # The release-monitor identified is for the gitlab-runner,
+  # but these seem to be co-versioned.
```
imjasonh, line 4: Would it help to make this a subpackage of gitlab-runner then? That way the wolfibot updates always keep these in sync, instead of needing two PRs.

gitlab-base/entrypoint.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -e
+
+# If in RHEL / UBI images, set ENV.
+# NOTICE: ubi-micro does not have grep
+if grep -q 'ID="rhel"' /etc/os-release ; then
```
imjasonh, line 7: This will never be the case for our Wolfi-based images.

hectorj2f: True! I was copying their scripts literally. I wasn't sure if we are supposed to change them, but I know the answer now :D.

joshrwolf:
where do these scripts come from? does it make more sense to include these in a central `gitlab` package instead?

hectorj2f:
The scripts come from here https://gitlab.com/gitlab-org/build/CNG/-/tree/master/gitlab-base/scripts?ref_type=heads and they are used on several (not all the images).

gitlab-base.yaml
hectorj2f: Generally this is the part where I feel a bit more uncertain of what we should do. Either create a project for this and make subpackages of it, as there are some images using it as base image.

Happy to hear your feedback as a noob.

hectorj2f: @mattmoor have triaged these images. Perhaps he knows what we should better do for this one.

joshrwolf: posting this here for reference: https://github.com/wolfi-dev/os/blob/main/gitlab-cng.yaml

looks like this sets us up to reference the scripts as runtime deps, for example: `gitlab-cng-base-scripts` essentially replicates what this PR is doing

mattmoor: Yeah, I don't think base needs any packages we don't have.  It should just be assembling the image from assorted packages like the `gitlab-cng-base-scripts`, `xtail`, and `gitlab-logger` packages (this is off the top of my head, I'd consult the Dockerfile).

mattmoor: @imjasonh they are in different repos :-/

hectorj2f: Yes, I talked with Josh and I'll create the base image based on the CNG image with the required deps.

hectorj2f:
I've been talking with @rawlingsj this morning, and I think we could simply add `gitlab-base` as a subpackage of `gitlab-cng`. 

I'll open a PR with this proposed change and close this PR when we agree on a final solution.

hectorj2f:
Closing this PR in favor of #7036 !

PR #7007, opened 2023-10-18
Add bitnami-compat subpackage to prometheus

priyawadhwa:
https://github.com/chainguard-dev/image-requests/issues/340


rawlingsj (approved):

rawlingsj:
looks like this package now fails after the node 20 to 21 update

PR #7036, opened 2023-10-18
gitlab: add gitlab-cng-base as subpackage of gitlab-cng

hectorj2f:
Fixes:

We need to create a gitlab-base subpackage that contains few dependencies and some scripts from the original git checkout from cng.

Related: 
 https://github.com/chainguard-dev/image-requests/issues/457

#### For new version streams

### DIFF
--- a/gitlab-cng.yaml
+++ b/gitlab-cng.yaml
@@ -7,7 +7,7 @@ package:
   # The release-monitor identified is for the gitlab-runner,
   # but these seem to be co-versioned.
   version: 16.4.1
-  epoch: 0
+  epoch: 1
   description: Cloud Native container images per component of GitLab
   copyright:
     - license: MIT
@@ -28,7 +28,6 @@ data:
   # Used to create all of the *-scripts subpackages from the CNG repo.
   - name: scripts
     items:
-      base: ./gitlab-base
       cfssl-self-sign: ./cfssl-self-sign
       container-registry: ./gitlab-container-registry
       exporter: ./gitlab-exporter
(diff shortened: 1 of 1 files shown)


rawlingsj (approved):

gitlab-cng.yaml
```diff
@@ -54,6 +53,25 @@ subpackages:
...
 
+  - name: "${{package.name}}-base"
+    pipeline:
+      - runs: |
+          cd ./gitlab-base
+          for x in $(find scripts/ -type f); do
+            mkdir -p ${{targets.subpkgdir}}/$(dirname $x)
+            cp -r $x ${{targets.subpkgdir}}/$x
+          done
+    dependencies:
```
mattmoor, line 64: I’d do these at the image level.  Gitlab has several flavors of each of their images and it’s not clear to me that each of them is going to have this full set.  For example, I’m not sure if their hardened image has xtail or procps.

If there are things like perhaps the logger, gomplate, busybox that are truly common across all images I’d just add them to the range version.

fwiw ca-certs is a dependency of our baselayout, so it’s never needed explicitly.

hectorj2f: I think this would simplify the creation of the images, and it'll also "fix" the creation of a current package `gitlab-cng-base` which wasn't quite useful at this moment. 

It seems xtail and procps seems to be part of their gitlab-base image https://gitlab.com/gitlab-org/build/CNG/-/blob/master/gitlab-base/Dockerfile?ref_type=heads#L27. Also I can confirm procps is used in 8 images, xtail in 5 images, gitlab-logger in all of them, as well as gomplate.

> fwiw ca-certs is a dependency of our baselayout, so it’s never needed explicitly.

I wasn't aware of that detail, I can remove it.

PR #6616, opened 2023-10-11
rootlesskit: new package

wlynch:
https://github.com/rootless-containers/rootlesskit

Related: https://github.com/wolfi-dev/os/pull/1860 (rootlesskit required for rootless buildkit)

#### For new package PRs only

- [ ] This PR is marked as fixing a pre-existing package request bug
  - [x] Alternatively, the PR is marked as related to a pre-existing package request bug, such as a dependency
- [x] REQUIRED - The package is available under an OSI-approved or FSF-approved license
- [x] REQUIRED - The version of the package is still receiving security updates
- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`)

PR #7047, opened 2023-10-18
Add rootlesskit-config-nonroot package.

wlynch:
Rootlesskit requires configuration to carve out UID namespaces. This adds a configuration for the user "nonroot" which is a common user we use for images.

See https://rootlesscontaine.rs/getting-started/common/subuid/ for more details.

Related: https://github.com/wolfi-dev/os/pull/6616


vaikas:
I would think we would want this to be a subpackage of rootlesskit?
https://wiki.inky.wtf/docs/teams/engineering/images/package-and-image-guides/serviceable-packages/

And then folks could override it?
https://wiki.inky.wtf/docs/teams/engineering/images/package-and-image-guides/customizing-configurations/

wlynch:
oooo nice! Done.

(we should consider adding these docs to the wolfi repo)

imjasonh (dismissed):

rootlesskit.yaml
```diff
@@ -1,10 +1,15 @@
...
   name: rootlesskit
   version: 1.1.1
-  epoch: 0
+  epoch: 3
   description: RootlessKit is a Linux-native implementation of "fake root" using user_namespaces(7).
   copyright:
     - license: Apache-2.0
+  dependencies:
+    # Let's ensure that we have config. This gets the default, but can also
+    # be override by users.
```
vaikas, line 10: super nit my stupid typo s/override/overridden/

rootlesskit.yaml
```diff
@@ -1,10 +1,15 @@
 package:
   name: rootlesskit
   version: 1.1.1
-  epoch: 0
+  epoch: 3
```
vaikas, line 4: nit, maybe change this back to 1? I assume this is because of local devl / changes? 🤣

rootlesskit.yaml
wlynch: Done.

rootlesskit.yaml
wlynch: Done.

imjasonh (approved):

PR #7071, opened 2023-10-18
Remove unnecessary runtime dependencies from bitnami subpackage

priyawadhwa:
I think this is making the image larger than it needs to be.

### DIFF
--- a/prometheus.yaml
+++ b/prometheus.yaml
@@ -1,7 +1,7 @@
 package:
   name: prometheus
   version: 2.47.2
-  epoch: 1
+  epoch: 2
   description: The Prometheus monitoring system and time series database.
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


pdeslaur:
I'm curious about the rationale, wouldn't the `prometheus` package included in the image anyway?

priyawadhwa:
For the `prometheus-bitnami` image I'm only copying in the `prometheus-bitnami` package. The only different between `prometheus` and `prometheus-bitnami` packages is that the bitnami package moves the prometheus binaries to the expected location in the bitnami image (`/opt/bitnami/prometheus`). 

By pulling in the `prometheus` package at runtime, it means there's two of every prometheus binary in the image at different locations, almost doubling the size of the image.

imjasonh (approved):

PR #7088, opened 2023-10-19
cilium: update llvm deps + fix cni install script

tcnghia:
Current script assumes the location of the `cni-plugins` package's `loopback` binary to be at `/cni/loopback`. Fixing to `/usr/bin/loopback` to match the Wolfi package expectation.

Also, fix the runtime LLVM dep to be LLVM 15 to avoid having 2 LLVM versions in the runtime dependencies (since clang packs its own LLVM dep).

### DIFF
--- a/cilium.yaml
+++ b/cilium.yaml
@@ -42,6 +42,7 @@ pipeline:
   - uses: patch
     with:
       patches: loopback-location.patch
+
   - runs: |
       # Remove groupadd from Makefile: it's not doing anything useful in
       # a package build anyway, and it's not available in busybox.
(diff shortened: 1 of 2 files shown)


cilium.yaml
```diff
@@ -37,6 +39,9 @@ pipeline:
       tag: v${{package.version}}
       expected-commit: 252a99efbc5def73405f9382b343846edc11d1ca
 
+  - uses: patch
+    with:
+      patches: loopback-location.patch
```
tcnghia, line 44:
```suggestion
      patches: loopback-location.patch

```

jonjohnsonjr (approved):

PR #7110, opened 2023-10-19
gitlab: install deps to run the images

hectorj2f:
Fixes:

gitlab-logger is expected to be installed under `/usr/local/bin`

gitlab-shell missed some runtime dependencies to run the image such as openssl.

Related:

### DIFF
--- a/gitlab-logger.yaml
+++ b/gitlab-logger.yaml
@@ -23,11 +23,18 @@ pipeline:
       output: gitlab-logger
       ldflags: -w
 
-  - runs: |
-      mkdir -p "${{targets.destdir}}"/usr/local/bin
-      cp ${{targets.destdir}}/usr/bin/gitlab-logger "${{targets.destdir}}"/usr/local/bin/gitlab-logger
-
   - uses: strip
+
+subpackages:
+  - name: gitlab-logger-compat
+    dependencies:
+      runtime:
+        - gitlab-logger
+    pipeline:
+      - runs: |
+          # GitLab helm chart mostly hardcodes multiple executables path to /usr/local/bin/*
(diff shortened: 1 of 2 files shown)


gitlab-logger.yaml
```diff
@@ -23,6 +23,10 @@ pipeline:
       output: gitlab-logger
       ldflags: -w
 
+  - runs: |
+      mkdir -p "${{targets.destdir}}"/usr/local/bin
+      cp ${{targets.destdir}}/usr/bin/gitlab-logger "${{targets.destdir}}"/usr/local/bin/gitlab-logger
+
```
rawlingsj, line 29: I think we'll need a -compat subpackage for this, rather than using `/usr/local/` we can use a subpackage that creates a symlink to the binary `/usr/bin/gitlab-logger`  into `/usr/local/`

rawlingsj: i.e. like this https://github.com/wolfi-dev/os/blob/1c49ca89b30fb7f2c65553c1d4aaadd2d4417693/flux.yaml#L40-L47

the reason is in wolfi we avoid installing into /usr/local

gitlab-logger.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: gitlab-logger
   version: 3.0.0
-  epoch: 1
+  epoch: 5
```
rawlingsj, line 4: sorry should have seen this, think this should be 2?
```suggestion
  epoch: 2
```

gitlab-logger.yaml
hectorj2f: Awesome! Thanks for the explanation.

rawlingsj (approved):

hectorj2f:
@rawlingsj Thanks for the review!

PR #7199, opened 2023-10-20
Add py3-tfx-bsl

Dentrax:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/py3-tfx-bsl.yaml
+++ b/py3-tfx-bsl.yaml
@@ -1,0 +1,85 @@
+package:
+  name: py3-tfx-bsl
+  version: 1.14.0
+  epoch: 0
+  description: Common code for TFX
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - py3-absl-py
+      - py3-apache-beam
+      - py3-google-api-python-client
+      - numpy
+      - py3-pandas
+      - py3-protobuf
+      - pyarrow
+      - tensorflow-core
+      - py3-tensorflow-core
+      - py3-tensorflow-metadata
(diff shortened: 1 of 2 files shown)


alexeagle:
Hello! It looks like I'll come help out to green this one up as soon as we get some contract in place.

alexeagle:
FYI @jamonation I've been trying to help green this one up. Based on your work in https://github.com/wolfi-dev/os/pull/8326 I think this probably has similar problems.

jamonation:
Ah @alexeagle I think it might be an issue with upstream Tensorflow needing to support 3.12 and we're too ahead of things. I found https://github.com/tensorflow/tensorflow/issues/62003 that seems to indicate 3.12 is still a work in progress.

alexeagle:
Yes, totally agree we are trying with a newer python than they support. I assumed the goal with wolfi was to always be on the latest interpreter for security fixes, so it was a design goal to patch up all the packages to work with a newer version.

Dentrax:
Hey @alexeagle @jamonation, thank you for your thoughts!

Could you please check if my _patching_ approach is correct?

I have tried to follow similar approach as you already done here: https://github.com/wolfi-dev/os/pull/8326/files

alexeagle:
I'm not sure what you mean by "approach". Patching the dependencies is likely required unless we build it exactly like the project does on their own CI (ie with the same python version)

I found I had to edit your patch to remove the zeta->m4 portion as the version bump of m4 brought in those changes. It seems like you do have a working technique for creating the patches.

PR #7238, opened 2023-10-22
Add python astroid package

jspeed-meyers:
Add the Python `astroid` package, a key dependency for `pylint`, a popular Python static analysis tool.

source code URL: https://github.com/pylint-dev/astroid

#### For new package PRs only
- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`) [can't find one]


py3-astroid.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/pylint-dev/astroid
+      tag: v${{package.version}}
+      expected-commit: 0191f534213364a844e8138dd68bb71f3ba603e3
+
+  - name: Python Build
+    runs: python -m pip install --prefix=/usr --root="${{targets.destdir}}" .
```
rawlingsj, line 31: I don't think we have any firm docs on packaging python but from the recent python 3.12 upgrade I think there was a suggestion that `pip install` wasn't great as it can vendor other python dependencies.  We now have some reusable melange pipelines that might be better https://github.com/chainguard-dev/melange/tree/main/pkg/build/pipelines/python

jspeed-meyers: Ah, 🤦 ! I think I even knew that. I blindly followed the preferred build steps in the `astroid` README.

The new commit uses an approach other than `pip install`.

In your opinion, what docs should I update with this advice? The internal wiki? Somewhere else?

jspeed-meyers: THANK YOU, BTW.

rawlingsj (approved):

PR #7272, opened 2023-10-23
Add temporal-ui-server

jonjohnsonjr:
The temporal-ui-server OCI entrypoint uses dockerize, so we need https://github.com/wolfi-dev/os/pull/7275 first.

#### For new package PRs only

### DIFF
--- a/temporal-ui-server.yaml
+++ b/temporal-ui-server.yaml
@@ -1,0 +1,54 @@
+package:
+  name: temporal-ui-server
+  version: 2.19.0
+  epoch: 0
+  description: Golang Server for https://github.com/temporalio/ui
+  copyright:
+    - license: MIT License
+
+environment:
+  contents:
+    packages:
+      - ca-certificates-bundle
+      - busybox
+  environment:
+    CGO_ENABLED: "0"
+
+pipeline:
+  - uses: git-checkout
+    with:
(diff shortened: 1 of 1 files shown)


imjasonh (dismissed):

temporal-ui-server.yaml
```diff
@@ -0,0 +1,61 @@
...
+      go get -u golang.org/x/net@v0.17.0
+
+      go mod tidy -compat=1.18
+      go mod vendor
+
+  - uses: go/build
+    with:
+      packages: ./cmd/server
+      output: ui-server
+      ldflags: -s -w
```
pdeslaur, line 35: Can we use `deps` from `go/build` to fix the CVE, or does it not do vendoring?

temporal-ui-server.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+          mkdir -p ${{targets.subpkgdir}}/home/ui-server/config
+          cp docker/config-template.yaml ${{targets.subpkgdir}}/home/ui-server/
+
+update:
+  enabled: true
+  github:
+    identifier: temporalio/ui-server
+    strip-prefix: v
+    use-tag: true
```
pdeslaur, line 60: We can probably drop `use-tag` since the project has releases, I think!

pdeslaur (approved):

PR #7276, opened 2023-10-23
falco: add `falco-no-driver` subpackage

cmdpdx:
Add `falco-no-driver` subpackage to build the falco binary without the driver loader and with modern eBPF support.

```
 $ wolfictl scan ./packages/aarch64/falco-no-driver-0.36.1-r1.apk
Will process: falco-no-driver-0.36.1-r1.apk
✅ No vulnerabilities found
```

#### For new package PRs only

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -42,7 +42,6 @@ environment:
       - libbpf-dev
       - libcurl-openssl4
       - libelf
-      - libelf-static
       - libsystemd
       - libtbb-dev
       - libtool
@@ -131,11 +130,14 @@ subpackages:
               cmake \
                 -DCMAKE_INSTALL_PREFIX=/usr \
                 -DCMAKE_INSTALL_LIBDIR=/usr/lib \
+                -DBUILD_SHARED_LIBS=On \
                 -DFALCO_ETC_DIR=/etc/falco \
                 -DCMAKE_BUILD_TYPE=Release \
                 -DMINIMAL_BUILD=Off \
                 -DUSE_BUNDLED_DEPS=Off \
                 -DUSE_BUNDLED_LIBELF=Off \
+                -DUSE_BUNDLED_TBB=Off \
+                -DUSE_BUNDLED_RE2=Off \
                 -DBUILD_BPF=Off \
                 -DBUILD_LIBSCAP_MODERN_BPF=Off \
                 -DLIBSCAP_INCLUDE_DIRS=/usr/lib/falco-libscap-modern-ebpf \
@@ -143,7 +145,7 @@ subpackages:
                 -DBUILD_FALCO_MODERN_BPF=On \
               ..
           - runs: |
-              make falco
+              make -ltbb falco
               install -Dm755 ./userspace/falco/falco "${{targets.subpkgdir}}"/usr/bin/falco
       - uses: strip
 
--- a/falco.yaml
+++ b/falco.yaml
@@ -1,7 +1,7 @@
 package:
   name: falco
   version: 0.36.1
-  epoch: 1
+  epoch: 2
   description: Cloud Native Runtime Security
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


falco.yaml
```diff
@@ -98,8 +102,32 @@ pipeline:
...
+                -DCMAKE_INSTALL_LIBDIR=/usr/lib \
+                -DFALCO_ETC_DIR=/etc/falco \
+                -DCMAKE_BUILD_TYPE=Release \
+                -DUSE_BUNDLED_DEPS=Off \
+                -DMINIMAL_BUILD=On \
+                -DBUILD_DRIVER=Off \
+                -DBUILD_FALCO_MODERN_BPF=On \
+                -DBUILD_LIBSCAP_MODERN_BPF=Off \
+                -DLIBSCAP_INCLUDE_DIRS=/usr/lib/falco-libscap-modern-ebpf \
+                -DBUILD_BPF=Off \
```
Dentrax, line 125: How this is different than `BUILD_FALCO_MODERN_BPF`?

Dentrax (dismissed):

falco.yaml
cmdpdx: `BUILD_BPF` determines if the legacy (non-modern) eBPF driver should be built.

I'll put a link to the build docs for falco and the libs for reference, though not all of the flags are explicitly mentioned in the docs (e.g. `LIBSCAP_INCLUDE_DIRS`)

cmdpdx:
Converted to draft while I fix a missing config and slight refactor

falco.yaml
```diff
@@ -76,30 +77,68 @@ pipeline:
...
       sed -i 's/find_dependency(Protobuf CONFIG)/find_package(Protobuf CONFIG)/' /usr/lib64/cmake/grpc/gRPCConfig.cmake
 
-  - runs: |
-      mkdir -p "${{targets.destdir}}"/etc/falco
-      install -Dm755 ./falco.yaml "${{targets.destdir}}"/etc/falco/falco.yaml
-      sed -e 's/time_format_iso_8601: false/time_format_iso_8601: true/' < "${{targets.destdir}}"/etc/falco/falco.yaml
+data:
+  - name: builds
+    items:
+      falco-standard: Cloud Native Runtime Security; built with the driver loader
```
cmdpdx, line 83: I didn't see the `falco` packaged used in any images yet, and since you can't (?) have a subpackage the same name as the package I opted for `falco-standard` for the build with the driver loader.

falco.yaml
```diff
@@ -76,30 +77,68 @@ pipeline:
...
-            -DCMAKE_BUILD_TYPE=Release \
-            -DUSE_BUNDLED_DEPS=Off \
-            -DMINIMAL_BUILD=On \
-            -DBUILD_DRIVER=On \
-            -DBUILD_FALCO_MODERN_BPF=Off \
-            -DBUILD_BPF=Off \
-          ..
+          mkdir -p "${{targets.subpkgdir}}"/etc/falco
+          cp ./falco.yaml "${{targets.subpkgdir}}"/etc/falco/falco.yaml
+          sed -e 's/time_format_iso_8601: false/time_format_iso_8601: true/' < "${{targets.subpkgdir}}"/etc/falco/falco.yaml
```
cmdpdx, line 97: Rather than duplicate this block with `s/destdir/subpkgdir/` I made the standard falco a subpackage also

falco.yaml
```diff
@@ -76,30 +77,68 @@ pipeline:
...
+            cmake \
+              -DCMAKE_INSTALL_PREFIX=/usr \
+              -DCMAKE_INSTALL_LIBDIR=/usr/lib \
+              -DFALCO_ETC_DIR=/etc/falco \
+              -DCMAKE_BUILD_TYPE=Release \
+              -DUSE_BUNDLED_DEPS=Off \
+              -DMINIMAL_BUILD=On \
+              -DBUILD_DRIVER=On \
+              -DBUILD_FALCO_MODERN_BPF=Off \
+              -DBUILD_BPF=Off \
```
Dentrax, line 134: Instead of duplicating whole block, why not just set the changed flags?

falco.yaml
```diff
@@ -76,30 +77,68 @@ pipeline:
...
-            -DBUILD_BPF=Off \
-          ..
+          mkdir -p "${{targets.subpkgdir}}"/etc/falco
+          cp ./falco.yaml "${{targets.subpkgdir}}"/etc/falco/falco.yaml
+          sed -e 's/time_format_iso_8601: false/time_format_iso_8601: true/' < "${{targets.subpkgdir}}"/etc/falco/falco.yaml
       - runs: |
-          make falco
-          install -Dm755 ./userspace/falco/falco "${{targets.destdir}}"/usr/bin/falco
+          mkdir -p build
+          cd build
```
Dentrax, line 100: Let's use `- working-directory: build` here, so you don't need to `mkdir && cd`.

falco.yaml
Dentrax: I'm not sure if we able to use exact same package name in its one of the sub-packages. FWIW, I would expect consumers to install with the `falco` naming rather than `falco-standard`.

Dentrax (changes requested):

falco.yaml
cmdpdx: I tried that, but then those blocks were skipped (cmake and make didn't execute, no binary was built) 🤔

falco.yaml
cmdpdx: I ran into an issue when I reused the name for a subpackage, but that must have been a separate issue because I tried again and was successful with a subpackage named `falco`

cmdpdx: From the CI
```
unable to add edge for subpackage "falco-0.36.1-r2" from falco-0.36.1-r2: edge would create a cycle
```

Subpackage cannot be name of package. If we want the subpackage to stay `falco`, the main package will need to change.

cmdpdx: But if the main package name changes, how does that affect epoch and already built falco packages 🤔 

As nice as it is to reduce duplication by using subpackages, I think the easiest path forward here is to either
1. build `falco` as the main package in `falco.yaml`, and `falco-no-driver` as a subpackge, with most of the steps duplicated (replacing e.g. `targets.subpkgdir` for `targets.destdir`)
2. submit a separate `falco-no-driver.yaml` package

falco.yaml
cmdpdx: works now without the use of `range` for subpackage `falco-no-driver`

Dentrax (dismissed):

cmdpdx:
Building an image with the `falco-no-driver` package and running it per falco docs gives the following error:

```
 $ docker run --rm -i -t --privileged -v /var/run/docker.sock:/host/var/run/docker.sock -v /proc:/host/proc:ro ttl.sh/colin/falco-no-driver:latest falco --modern-bpf
Error: Option ‘modern-bpf’ does not exist
```

Srishti found this comment `# Modern BPF is not supported on not Linux systems and in MINIMAL_BUILD` that likely explains the failure. I'm attempting to rebuild this package with `MINIMAL_BUILD=Off`

PR #7280, opened 2023-10-23
package: libxss

lpcalisi:

### DIFF
--- a/libxss.yaml
+++ b/libxss.yaml
@@ -1,0 +1,44 @@
+package:
+  name: libxss
+  version: 1.2.4
+  epoch: 0
+  description: "Xlib-based X11 Screen Saver extension client library"
+  copyright:
+    - license: custom:XFREE86
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - build-base
+      - libxext-dev
+      - libx11-dev
+
+pipeline:
+  - uses: fetch
+    with:
+      uri: https://www.x.org/releases/individual/lib/libXScrnSaver-${{package.version}}.tar.xz
+      expected-sha256: 75cd2859f38e207a090cac980d76bc71e9da99d48d09703584e00585abc920fe
+
+  - uses: autoconf/configure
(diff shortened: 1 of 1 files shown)


libxss.yaml
```diff
@@ -0,0 +1,44 @@
+package:
+  name: libxss
+  version: 1.2.4
+  epoch: 0
+  description: "Xlib-based X11 Screen Saver extension client library"
+  copyright:
+    - license: custom:XFREE86
```
rawlingsj, line 7: I'm a bit confused on this license, I can see that alpine package this at MIT https://git.alpinelinux.org/aports/tree/community/libxscrnsaver/APKBUILD#n8 but the license in the gitlab repo is `X Consortium` while looking that up on spdx org it looks like it should be X11? https://spdx.org/licenses/X11.html - I also don't find a mention of `custom:XFREE86`

lpcalisi: I have not clear the right license too but I'm relying on the [X11 repository license](https://gitlab.freedesktop.org/xorg/lib/libxscrnsaver/-/blob/master/COPYING) that have two Copyright's

* *Copyright (c) 1992  X Consortium*
* *Copyright (C) 2003 The XFree86 Project, Inc.  All Rights Reserved.*

I assumed that `custom:XFREE86` might be correct because the package [libx11](https://github.com/wolfi-dev/os/blob/main/libx11.yaml) use it.

*XFree86* is referenced [here](https://spdx.org/licenses/XFree86-1.1.html)

Do you think about `X11 AND XFree86-1.1`? (both short identifiers from spdx)

lpcalisi: @rawlingsj is that ok?

joshrwolf: the justification wrt libx11 makes sense to me. going to approve this and we can address any necessary changes postsubmit.

thank you for your patience!

joshrwolf (approved):

PR #7295, opened 2023-10-24
`libretls/3.8.1` package update

### LINKED ISSUE #7025: Fix libretls/3.8.1 package update (6651)
The auto-update PR for `libretls/3.8.1` requires attention. Check PR #6651 for more details.

_Don't forget to leave a comment if you want to claim this issue for Hacktoberfest._

debasishbsws:
Fixes: #7025

Related: #6651

#### For version bump PRs

py3-xmlsce package is dependent on the `libtls.so` provided by libretls.
Updated the epoch for soname checks. which is failing on #6651


debasishbsws:
Hey @rawlingsj here I am facing the same issue with sonames, I found only one package `py3-xmlsec` that uses this libretls.

rawlingsj:
ok great - thanks @debasishbsws , @kaniini this seems ok to override the failed check and merge?  

FWIW I wonder if we should split the soname check into a second job, much like the new scan one.  That way we can easily tell if the build is failing because of the soname and also other jobs will continue to run rather than the while CI check fail.

debasishbsws:
Hm🤔 It sounds like a good idea to create a second job for soname checks.

rawlingsj:
once this merges we can close https://github.com/wolfi-dev/os/pull/7295

PR #7337, opened 2023-10-24
change all destdir and subpkgdir to contextdir, enforce in lint

imjasonh:
```
git grep -l destdir | grep .yaml$ | xargs sed -i.bak "s|destdir|contextdir|g"
git grep -l subpkgdir | grep .yaml$ | xargs sed -i.bak "s|subpkgdir|contextdir|g"
```


imjasonh:
There's more here we need to do, because we do this:

```yaml
subpackages:
  - name: "scanelf"
    pipeline:
      - runs: |
          mv "${{targets.destdir}}"/usr/bin/scanelf "${{targets.subpkgdir}}"/usr/bin/scanelf
```

This pulls a file out of the main package filesystem and puts it in the subpackage filesystem, and changing both of those to `${{targets.contextdir}}` doesn't work.

So at the very least we need to do something smarter than this, and at most it might not be possible to do what we want to do. Closing this for now, but it's still something we should investigate.

PR #7373, opened 2023-10-25
gitlab: add gitlab-exporter

hectorj2f:
Fixes:
https://github.com/chainguard-dev/image-requests/issues/459
Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/gitlab-exporter.yaml
+++ b/gitlab-exporter.yaml
@@ -1,0 +1,77 @@
+# source is gitlab so we can't use github updates to get expected commit
+# let's still auto create the PR, it will fail as expected commit will be wrong
+# however it will be easy to fix
+#nolint:git-checkout-must-use-github-updates
+package:
+  name: gitlab-exporter
+  version: 13.4.1
+  epoch: 0
+  description: GitLab Exporter is a Prometheus Web exporter.
+  copyright:
+    - license: MIT
+  dependencies:
+    runtime:
+      - jemalloc
+      - ruby-3.2
+      - libpq-15
+      - busybox
+      - gitlab-cng-exporter-scripts
+      - ruby3.2-webrick
+      - ruby3.2-faraday
+      - ruby3.2-faraday-excon
+      - ruby3.2-rack
+      - ruby3.2-bundler
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - libpq-15
+      - postgresql-15-dev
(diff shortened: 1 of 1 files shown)


gitlab-exporter.yaml
```diff
@@ -0,0 +1,77 @@
...
+      gem: ${{vars.gem}}
+
+  - uses: ruby/install
+    with:
+      gem: ${{vars.gem}}
+      version: ${{package.version}}
+
+  # GitLab-Exporter Runtime Dependencies
+  - runs: |
+      gem install puma -v 5.6.5 --install-dir ${{targets.contextdir}}$(ruby -e 'puts Gem.default_dir')/
```
rawlingsj, line 61: I wonder.. ;) Would it be too complicated to have a little bash script that parsed the upstream `gem.spec` to grab the runtime dependencies and then 
```
gem install $RUNTIME_DEP -v $VERSION --install-dir ${{targets.contextdir}}$(ruby -e 'puts Gem.default_dir')/
```
?

That way when upstream updates their runtime dependencies it wont break our package or require manual updates?

If that's viable we could add an extra reusable [pipeline](https://github.com/chainguard-dev/melange/tree/main/pkg/build/pipelines/ruby) to melange for installing runtime dependencies?

rawlingsj: i.e. something like

```
  runs: |

    # Replace this with the actual path to your gitlab-exporter.gemspec file
    GEMSPEC_FILE="./gitlab-exporter.gemspec"

    # This is your target directory for installing the gems
    # I'm assuming it's an environment variable, change as needed
    TARGET_DIR="${targets.contextdir}$(ruby -e 'puts Gem.default_dir')/"

    # Use grep and sed to extract the lines containing runtime dependencies
    grep 's.add_runtime_dependency' $GEMSPEC_FILE | sed -E 's/s.add_runtime_dependency[[:space:]]+"([^"]+)",[[:space:]]+"([^"]+)"/\1 \2/' | while read -r GEM_NAME GEM_VERSION; do
      # Run the gem install command
      gem install $GEM_NAME -v $GEM_VERSION --install-dir $TARGET_DIR
    done
```

hectorj2f: I created one issue here https://github.com/chainguard-dev/melange/issues/785

rawlingsj (approved):

PR #7378, opened 2023-10-25
Add Gitlab mailroom

hectorj2f:
Fixes:
https://github.com/chainguard-dev/image-requests/issues/455
Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/gitlab-mailroom.yaml
+++ b/gitlab-mailroom.yaml
@@ -4,8 +4,8 @@
 #nolint:git-checkout-must-use-github-updates
 package:
   name: gitlab-mailroom
-  version: 0.0.23
-  epoch: 2
+  version: 0.10.1
+  epoch: 0
   description: GitLab mail_room contains some merged functionality that GitLab requires, so this mirror fork is to help us release custom functionality.
   copyright:
     - license: MIT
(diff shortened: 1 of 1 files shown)


gitlab-mailroom.yaml
```diff
@@ -0,0 +1,76 @@
...
+
+  - runs: |
+      # This is your target directory for installing the gems
+      # I'm assuming it's an environment variable, change as needed
+      INSTALL_GEM_DIR=${{targets.destdir}}$(ruby -e 'puts Gem.default_dir')/
+
+      # GitLab-Mail_Room dependencies installed when running the package.
+      # These dependencies are registered as development dependencies in the repository
+      # but installed when using the binary.
+      gem install charlock_holmes:0.7.7 \
```
hectorj2f, line 54: This is a particular scenario here, they are installed in the Dockerfile https://github.com/chainguard-dev/image-requests/issues/455 but they are not marked as runtime deps in the gemfile. I am not sure what we want to take as the source of truth but considering this will end up in one of our images. I'd say we should install all the deps installed in the Dockerfile.

imjasonh (approved):

gitlab-mailroom.yaml
hectorj2f: Sorry, the link is this one https://gitlab.com/gitlab-org/build/CNG/-/blob/master/gitlab-mailroom/scripts/install-dependencies#L13.

PR #7379, opened 2023-10-25
Bump xmlsec in keycloak to address GHSA-xfrj-6vvc-3xm2.

dlorenc:
Fixes:

Related:

### DIFF
--- a/keycloak.yaml
+++ b/keycloak.yaml
@@ -1,7 +1,7 @@
 package:
   name: keycloak
   version: 22.0.5
-  epoch: 0
+  epoch: 1
   description: Open Source Identity and Access Management For Modern Applications and Services
   copyright:
     - paths:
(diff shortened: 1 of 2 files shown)


keycloak.yaml
```diff
@@ -38,6 +38,9 @@ pipeline:
       export LANG=en_US.UTF-8
       export JAVA_HOME=/usr/lib/jvm/java-17-openjdk
 
+      # Update xmlsec from 2.2.3 to 2.2.6 to fix GHSA-xfrj-6vvc-3xm2
+      sed -i 's/2.2.3/2.2.6/g' pom.xml
```
luhring, line 42: This seems a little brittle to me, idk — what do you think about creating a patch to achieve this? That seems to be the most common way we've been handling bumps in Java projects, including in `pom.xml` files

luhring (approved):

PR #7393, opened 2023-10-25
fix directories kubeflow-pipelines frontend package

developer-guy:
Fixes:

we noticed that we were using the wrong directories for the image. At the times we were working on that package we were also have lots of other packages to do, so, we missed that one and made mistakes about the directories, so, the directories should be and we should also include .dist folder:

- [kubeflow/pipelines@570e56d/frontend/Dockerfile#L30-L31](https://github.com/kubeflow/pipelines/blob/570e56dd09af32e173cf041eed7497e4533ec186/frontend/Dockerfile#L30-L31)
- 
- [kubeflow/pipelines@570e56d/frontend/Dockerfile#L39](https://github.com/kubeflow/pipelines/blob/570e56dd09af32e173cf041eed7497e4533ec186/frontend/Dockerfile#L39)

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/kubeflow-pipelines.yaml
+++ b/kubeflow-pipelines.yaml
@@ -1,7 +1,7 @@
 package:
   name: kubeflow-pipelines
-  version: 2.1.3
-  epoch: 7
+  version: 2.3.0
+  epoch: 0
   description: Machine Learning Pipelines for Kubeflow
   checks:
     disabled:
@@ -12,23 +12,22 @@ package:
 environment:
   contents:
     packages:
-      - wolfi-base
+      - argo-cd
+      - build-base
       - busybox
       - ca-certificates-bundle
-      - build-base
+      - eslint
+      - git
       - go
       - go-licenses
-      - git
-      - argo-cd
-      - python3-dev
-      - py3-setuptools
-      - py3-gpep517
-      - py3-wheel
-      - py3-pip
+      - jq
       - kubectl
       - nodejs
-      - jq
-      - eslint
+      - py3-gpep517
+      - py3-pip
+      - py3-setuptools
+      - py3-wheel
+      - python3-dev
 
 data:
   - name: backends-go-builds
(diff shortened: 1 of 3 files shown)


kubeflow-pipelines.yaml
```diff
@@ -12,23 +12,23 @@ package:
...
       - kubectl
       - nodejs
-      - jq
-      - eslint
+      - py3-gpep517
+      - py3-pip
+      - py3-setuptools
+      - py3-wheel
+      - python3-dev
+      - wolfi-base
```
imjasonh, line 31: Does this need to include apk? 🤔

imjasonh (approved):
Can you include a description of the issue being fixed (or link to issue), so it's clearer what the change is trying to accomplish?

developer-guy:
> Can you include a description of the issue being fixed (or link to issue), so it's clearer what the change is trying to accomplish?

we noticed that we were using the wrong directories for the image. At the times we were working on that package we were also have lots of other packages to do, so, we missed that one and made mistakes about the directories, so, the directories should be and we should also include .dist folder:
> https://github.com/kubeflow/pipelines/blob/570e56dd09af32e173cf041eed7497e4533ec186/frontend/Dockerfile#L30-L31

> https://github.com/kubeflow/pipelines/blob/570e56dd09af32e173cf041eed7497e4533ec186/frontend/Dockerfile#L39

imjasonh (dismissed):

kubeflow-pipelines.yaml
cpanato: @developer-guy need to check this comment here, if that is not needed we can resolve this for now

developer-guy: I don't think that we need this one, let me remove it.

imjasonh (approved):

PR #7409, opened 2023-10-25
Poll releases instead of tags for updates, plus misc changes

mamccorm:
Poll for updates using git releases instead of tags. Added filter to exclude the only non-conventional release name. Also applied some general updates to align with some recent changes made to our FIPS image variant.

### DIFF
--- a/keycloak.yaml
+++ b/keycloak.yaml
@@ -21,6 +21,7 @@ environment:
       - ca-certificates-bundle
       - nodejs-18
       - openjdk-17
+      - openjdk-17-default-jvm
       - wolfi-base
       - wolfi-baselayout
   environment:
@@ -40,6 +41,10 @@ pipeline:
       patches: bump-xmlsec.patch
 
   - runs: |
+      # Upstream removed `keycloak-js-adapter` as a build-time dependency (March 2023).
+      # The maven build now pulls this in as a pre-compiled jar. Filed upstream
+      # issue to query: https://github.com/keycloak/keycloak/issues/24320.
+
       # Keycloak installation. Note we use the maven wrapper as configured in
       # the source repo to build - ensures the correct maven version for
       # building the project, preventing issues such as CI hangs.
--- a/keycloak.yaml
+++ b/keycloak.yaml
@@ -1,7 +1,7 @@
 package:
   name: keycloak
   version: 22.0.5
-  epoch: 1
+  epoch: 2
   description: Open Source Identity and Access Management For Modern Applications and Services
   copyright:
     - paths:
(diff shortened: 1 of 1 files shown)


keycloak.yaml
```diff
@@ -10,22 +10,22 @@ package:
...
 environment:
   contents:
     packages:
+      - bash
       - busybox
       - ca-certificates-bundle
-      - curl
-      - maven
-      - openjdk-17
-      - openjdk-17-default-jvm
```
rawlingsj, line 24: Do we build keycloak from source?  I was a little surprised that we didn't need `openjdk-17-default-jvm` and when I checked the logs of the CI build it looks like we are fetching prebuilt jars rather than doing a maven build.  Just wanted to check if that is the case and expected.

keycloak.yaml
```diff
@@ -10,22 +10,22 @@ package:
...
-      - bash
       - nodejs-18
```
rawlingsj, line 22: Similar to the comment above, I couldn't see in the CI build logs where a node build was happening, so just checking that this is needed?

keycloak.yaml
mamccorm: Thanks for taking a look. With regards to maven, we need to invoke the maven wrapper and use the project-defined version of maven to compile keycloak, so i'm guessing thats the jars your referring to? This was the original PR: https://github.com/wolfi-dev/os/pull/6350. theres an internal slack thread I can link you on as well via 1-1 if interested

keycloak.yaml
mamccorm: Sure, so if node is missing the build will eventually with an error similar to the below:

```
The parameters 'nodeVersion' for goal com.github.eirslett:frontend-maven-plugin:1.12.1:install-node-and-pnpm are missing or invalid
```

keycloak.yaml
mamccorm: So it looks like upstream made a change several months ago, where they've removed a build-time dependency on `keycloak-js-adapter`, and instead consume it as a pre-compiled jar. That explains why we needed `openjdk-17-default-jvm` in our builds originally and why it looks to be no longer required.

This doesn't look like an ideal change or something that is properly documented / clarified on their side. I've dropped over an issue to upstream:

- https://github.com/keycloak/keycloak/issues/24320

I'll also add a note to melange clarifying this. Good catch spotting this one

mamccorm: Have updated with a comment in melange and also added back the jvm, in case keycloak decide to revert the change or we need to build some custom release from an older branch / release that requires it. @rawlingsj would you mind another peak?

rawlingsj (approved):

keycloak.yaml
rawlingsj: lgtm thanks!

mamccorm:
Going to fix this up proper and build keycloak-js-adapter-jar from source. Will take a bit of extra time to test so i'll re-submit again.

PR #7455, opened 2023-10-26
adds sqlpad package

mritunjaysharma394:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/sqlpad.yaml
+++ b/sqlpad.yaml
@@ -1,6 +1,6 @@
 package:
   name: sqlpad
-  version: 7.1.2
+  version: 7.1.3
   epoch: 0
   description: Web-based SQL editor. Legacy project in maintenance mode.
   copyright:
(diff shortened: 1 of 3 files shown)


sqlpad.yaml
```diff
@@ -0,0 +1,47 @@
+package:
+  name: sqlpad
+  version: 7.1.3
```
rawlingsj, line 3:
```suggestion
  version: 7.1.3 # when updating check the patch below as it contains dependency version updates which may downgrade if upstream upgrades them
```

sqlpad.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - nodejs-18
+      - yarn
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - build-base
+      - yarn
+      - nodejs-18
```
rawlingsj, line 19: just checking - does this have to be pinned to `nodejs-18`?  Am wondering if `nodejs` would be fine which today is nodejs-20.  Just means we don't have to manage this in the future.

mritunjaysharma394: As much I remember, when I tried last, it failed and had to check sqlpad and they supported version 18 that's why I pinned it, let me re-try

mritunjaysharma394: Here https://github.com/sqlpad/sqlpad/blob/master/.nvmrc

mritunjaysharma394: No worries at all, just done, thanks

rawlingsj (approved):

PR #7462, opened 2023-10-26
fix the metadata-writer.py path in compat package and fix CVEs

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches


kubeflow-pipelines.yaml
```diff
@@ -213,7 +213,7 @@ subpackages:
     pipeline:
       - runs: |
           mkdir -p ${{targets.subpkgdir}}/kfp/metadata_writer
-          ln -s /usr/lib/python3.11/site-packages/metadata_writer.py "${{targets.subpkgdir}}"/kfp/metadata_writer/
+          ln -s /usr/lib/python3.12/site-packages/metadata_writer.py "${{targets.subpkgdir}}"/kfp/metadata_writer/
```
mamccorm, line 216: Could we use a wildcard so when we've a minor python3.x bump again in future we'll not need to increment this?

```
ln -s /usr/lib/python3.*/site-packages/metadata_writer.py "${{targets.subpkgdir}}"/kfp/metadata_writer/
```

Haven't tested but assume that'd work?

developer-guy: let me quickly check that! great idea.

developer-guy: unfortunately, it didn't work.

developer-guy:
I have created an advisories for CVEs: https://github.com/wolfi-dev/advisories/blob/c23a36957638522dfed555fd2f2f8785d6a5ec3a/kubeflow-pipelines.advisories.yaml#L4

kubeflow-pipelines.yaml
```diff
@@ -157,13 +148,13 @@ subpackages:
...
-          npm ci
-          npm run postinstall
+          npm i
           npm run build
           cp -r ./build/* ${{targets.subpkgdir}}/client/
           cd ./server
-          npm ci
+          npm i
           npm run build
           mv ./* ${{targets.subpkgdir}}/server/
```
julienv3, line 160: The npm modules that have CVEs are all copied through the `/server` dir, because `/build` doesn't contain any.
The build steps (copied from their dockerfile) actually ship all dev dependencies with `/server`.
You can fix that by `npm ci` to build and then `rm` + `npm ci --production` to install only runtime deps.
Combine it with an extra `npm audit fix` before and you _should_ be dealing with a much more manageable list of CVEs.

```suggestion
          npm ci
          npm run postinstall
          npm run build
          cp -r ./build/* ${{targets.subpkgdir}}/client/
          cd ./server
          npm ci
          npm run build
          rm -rf ./node_modules
          npm audit fix --package-lock-only
          npm ci --production
          mv ./* ${{targets.subpkgdir}}/server/

```

Note: I was able to get the frontend running with `npm audit fix --package-lock-only --force`, but I have no idea what will break at runtime if you do this, so you'd want to do some thorough testing or resort to advisories for these.

developer-guy: thank you so much!

>but I have no idea what will break at runtime if you do this, so you'd want to do some thorough testing or resort to advisories for these.

yes definitely I hope the tests is going to tell us what went wrong.

rawlingsj: Is there still more to do with this review comment?  I re-ran the failing scan CI check and there's still some CVEs detected which could be patched or advisories created for.

developer-guy: I already created here: 

- https://github.com/wolfi-dev/advisories/pull/432

mamccorm (dismissed):
looks good. spotted one new CVE in the last 2 days related to kubernetes 1.11 version. project cannot be upgraded as does not compile with newer version. @developer-guy please submit one more advisory

kubeflow-pipelines.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: kubeflow-pipelines
-  version: 2.3.0
```
imjasonh, line 3: Why is this downgrading to 2.0.3?

Anything that depends on `kubeflow-pipelines` without selecting a version will pick up the latest package, which is 2.3.0 and won't include your changes here.

imjasonh: Oh I see 2.0.3 looks like the latest version, but KFP SDK is 2.4.0 with a previous 2.3.0: https://github.com/kubeflow/pipelines/releases

Do we need to withdraw some packages?

developer-guy: the latest version is https://github.com/kubeflow/pipelines/releases/tag/2.0.3. they also use sdk- version to keep track kfp sdk's only.

developer-guy: I noticed that when I looked at the tags of the images of the kubeflow-pipelines components. The latest is 2.0.3

developer-guy: > Do we need to withdraw some packages?
I think so yes!

developer-guy: added packages that needs to be withdrawn to withdrawn.txt.

developer-guy:
We recently discovered that we had been using the incorrect version stream, specifically the one labeled as 'sdk-'. Upon closer examination, I realized that I had been searching for the version I needed within the image tags to compare the count of Common Vulnerabilities and Exposures (CVEs) between two versions. It became apparent that the 'sdk-' version stream is primarily intended for releasing KFP SDKs, and not suitable for our purposes.

I conducted a comparison between the two version streams, and it appeared that there were no discernible differences. However, I believe it would be prudent to switch to the original version stream, which is '2.0.x'. This change is important because using the 'sdk-' version stream could potentially lead to the publication of incorrect image tags. These two version streams are distinct and may have different release frequencies, making it crucial for us to ensure that we're using the right one to avoid any unintended issues.

Switching to the correct version stream will necessitate the removal of the incorrectly versioned packages, which entails some operational effort.

WDYT?

kubeflow-pipelines.yaml
```diff
@@ -62,29 +59,32 @@ pipeline:
...
-
-          go get golang.org/x/mod@v0.12.0
-
-          # Mitigate GHSA-gwc9-m7rh-j2ww
-          # Mitigate GHSA-8c26-wmh5-6g9v
-          go get golang.org/x/crypto@1baeb1ce4c0b
-
           # GHSA-p782-xgp4-8hr8
           go get golang.org/x/sys@v0.8.0
+          go get golang.org/x/text@v0.9.0
```
mamccorm, line 83: Looks like this used to be commented with #GHSA-69ch-w2m2-3vjp, but now it's not. Any reason? It doesn't look to be part of GHSA-p782-xgp4-8hr8.

Likewise for the others here?

mamccorm:
Overall looks good, added a query about patch comments. I'm not as familiar with the process for withdrawing packages (yet), so i'll let someone else do the final review / sign-off

imjasonh (dismissed):

mamccorm (dismissed):

kubeflow-pipelines.yaml
developer-guy: I removed it by an accident, re-added, thank you!

mamccorm (approved):

kubeflow-pipelines.yaml
tuananh: or you could get version from python :)

```sh
$ python -c 'import sys; print("%s.%s" % (sys.version_info.major, sys.version_info.minor))'
3.11
```

PR #7478, opened 2023-10-27
Add logging operator compat subpackage for fluentd

found-it:
Fixes:

The[ logging operator](https://kube-logging.dev/) builds its own fluentd image which uses a custom entrypoint and healthcheck script. This just grabs those scripts and places them in a subpackage for fluentd 1.16

Related:

#### For new package PRs only

### DIFF
--- a/ruby3.2-fluentd-1.16.yaml
+++ b/ruby3.2-fluentd-1.16.yaml
@@ -2,7 +2,7 @@ package:
   # fluentd supported versions: https://github.com/fluent/fluentd/blob/master/SECURITY.md
   name: ruby3.2-fluentd-1.16
   version: 1.16.2
-  epoch: 2
+  epoch: 3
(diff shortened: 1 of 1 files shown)


ruby3.2-fluentd-1.16.yaml
```diff
@@ -64,6 +64,19 @@ pipeline:
...
 
+subpackages:
+  - name: ${{package.name}}-logging-operator-compat
+    description: Entrypoint used by the logging operator image
+    dependencies:
+      runtime:
+        - busybox
+    pipeline:
+      - runs: |
+          git clone https://github.com/kube-logging/fluentd-images.git
```
imjasonh, line 75: Can you fetch or checkout a commit, just so this doesn't move forward unexpectedly?

found-it: Good call, they don't have togs or releases so I just pinned to the current commit

imjasonh (approved):

PR #7659, opened 2023-10-30
Run CI using pull_request_target

imjasonh:
Opening this as draft to start discussion.

We have a CI check today that runs package diffs and posts PR comments, which are great. We'd love to have more of these kinds of automation.

However, because the workflow requires `pull_request: write` permissions, it doesn't have permission to write PR comments when the PR is coming from a fork repo (as most Wolfi developers use today). Wolfi-bot's PRs are opened from a branch on the wolfi-dev/os repo, so those get the package diff comments, but otherwise most PRs don't get those.

Switching to `pull_request_target` will mean PRs from forks will be able to post PR comments, _but_, it means PRs won't be able to update the GH workflow definition and have it apply in the PR itself. This means that changes to CI workflows will be a bit harder to make, since you'll have to wait until the following PR to see if they worked or not.


jonjohnsonjr (approved):

pdeslaur (approved):

jonjohnsonjr (approved):

pdeslaur:
I think we'll have to force merge this one!

.github/workflows/ci-build.yaml
```diff
@@ -1,7 +1,13 @@
...
 
 on:
-  pull_request:
+  # When a PR is opened from a fork repo, we want to be able to
+  # get permission to post comments on the PR. The workflow
+  # will be defined by the workflow definition *at HEAD*, not
+  # from the PR. When iterating on this workflow, you may want to
+  # change this to `pull_request` and create your PR from a
+  # branch on this repo, instead of from a fork.
+  pull_request_target:
```
mattmoor, line 10: `tl;dr` Before we do this, let's narrow the scope of where we need write permissions.

Right now, the vast majority of workflows in this file have:
```yaml
permissions:
  contents: read
```

Which is great. However, the `build` workflow, which runs untrusted code asks for:
```yaml
permissions:
  contents: read
  pull-requests: write # so we have permission to comment on pull requests
```

Let's split up `build` into a phase that only needs `contents: read`:
1. Checks out the PR
2. Builds the APK(s)
3. Uploads the APK(s) to the PR

Then have a follow-on workflow that elevates to `pull-requests: write` and:
1. Downloads the APKs
2. Diffs the APKs against the latest build
3. Posts the diff to the PR

The key feature of this second workflow is that it does not execute any code from the checkout, which is not the case with this change as-is.

Issue #6063, opened 2023-09-29
image-magick:7.1.1-18


ahmetak4n:
### What software would you like us to add to wolfi-os. Ideally include a URL to the project and its source.

[ImageMagick®](https://imagemagick.org/) is a free and [open-source](https://imagemagick.org/script/license.php) software suite for displaying, converting, and editing raster image and vector image files. It can read and write over 200 image file [formats](https://imagemagick.org/script/formats.php), and can support a wide range of image manipulation operations, such as resizing, cropping, and color correction.

### which versions of the software should we include?

7.1.1-18

### Add some justification for why this specific package and versions are important.

ImageMagick is so popular image editing libraries. Dev teams requires this libraries while resize, edit or converting images in run-time. 7.1.1-18 is latest version of image magick

### Please check all that apply

- [X] This package has an un-restrictive license
- [X] The package/versions proposed are actively maintained upstream
- [X] I am interested in adding this package to Wolfi OS myself
- [X] I am willing to help maintain this package

Also I prepared a pull request for the package. #6064

ajayk:
This was addressed by #7697

PR #7697, opened 2023-10-31
Add ImageMagick.

amouat:
This is in progress and still needs subpackages split out. We're also missing some libraries so some formats are not currently supported.

Will add update data shortly...

Fixes:
https://github.com/wolfi-dev/os/issues/6063
Related:
https://github.com/wolfi-dev/os/pull/6064

#### For new package PRs only

- [ x ] This PR is marked as fixing a pre-existing package request bug
- [ x ] REQUIRED - The package is available under an OSI-approved or FSF-approved license
It's a variant of Apache
- [ x ] REQUIRED - The version of the package is still receiving security updates

### DIFF
--- a/imagemagick.yaml
+++ b/imagemagick.yaml
@@ -1,35 +1,36 @@
 package:
-    name: imagemagick
-    version: 7.1.1.21
-    epoch: 0
-    description: Tools and libraries for manipulating common image formats
-    copyright:
-        - license: ImageMagick
+  name: imagemagick
+  version: 7.1.1.21
+  epoch: 0
+  description: Tools and libraries for manipulating common image formats
+  copyright:
+    - license: ImageMagick
+
 environment:
-    contents:
-        packages:
-            - busybox
-            - ca-certificates-bundle
-            - build-base
-            - automake
-            - autoconf
-            - chrpath
-            - fftw-dev
-            - fontconfig-dev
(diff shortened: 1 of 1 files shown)


amouat:
Oh and the fetch is wrong. On it...

cpanato (commented):
it lgtm

are we going to fix the cves before merging? or are you okay with that?

amouat:
@cpanato I don't know how we can do that? I think we have to merge then issue advisories, right?

This is the newest version of the package, it should contain all fixes.

cpanato:
@amouat, but maybe those cves are new and are not in the latest release :/

This is a new package, so I think we don't need to add advisories, only if we merge as is and fix it later

cc @luhring to correct me if i am wrong

amouat:
Not sure I follow, maybe we should have a call?

I think the issue is the ImageMagick CVE data doesn't specify versions, so it applies to all versions unless an advisory is issued.

rawlingsj (approved):
Approving as advisories will follow.

luhring:
>This is a new package, so I think we don't need to add advisories, only if we merge as is and fix it later
>
>cc @luhring to correct me if i am wrong

It's only "fixed" advisories that don't get added when a package is first introduced. Other advisories can and should be used as needed for new packages

PR #7861, opened 2023-11-01
py3-pandas: switch back to samurai

kaniini:
Add a patch to add the missing dependency relationships, so that Samurai gets the build order right.


vaikas (approved):
Thanks!!!

PR #7863, opened 2023-11-01
pyarrow: ignore dev tags for update

ajayk:
Related to https://github.com/wolfi-dev/os/pull/7857 

this tried to pull a dev tag from apache-arrow


pyarrow.yaml
```diff
@@ -45,9 +45,9 @@ update:
   # There are some other things like js-* that break our update block.
   ignore-regex-patterns:
     - 'js-'
+    - '.*\.dev$'
```
debasishbsws, line 48:
```suggestion
    - '.*\\.dev$'
```
I am not sure but I think in wolfictl if this is directly passed in `regexp.Compile()` it will give an error `unknown escape`

Because, in Go strings, the backslash itself is an escape character, so we need to escape the backslash with another backslash (`\\`). So `\\.` in a Go string becomes `\.` in the regular expression, which matches a literal dot.

---
This will give the error:
```go
func main() {
	str1 := "Hello.dev"
	regex, err := regexp.Compile(".*\.dev$")
	if err != nil {
		fmt.Printf("failed to compile regex: %v\n", err)
		return
	}
	if regex.MatchString(str1) {
		fmt.Println("str1 matched")
	}
}
```

debasishbsws: I am not sure what happens when we marshal data from json or yaml, will it take care of the escape character or not. Have to check

debasishbsws: Okay looks like with yaml Unmarshal it takes everything literally. So no issue.

ajayk: cc: @rawlingsj

ajayk: somehow the regex seems to be working for arrow and other few packages , I will check this out soon

debasishbsws: Yes, it should be. As we are passing the regex from this Yaml file and golang unmarshal it without treating '\' as an escape character.

rawlingsj: So just confirming this conversation / concern can be resolved?

debasishbsws: Yes, It is resolved. You can merge it.

luhring (approved):

PR #7899, opened 2023-11-02
Icu/74.1 package version update

debasishbsws:
Fixes: #7897

#### For version bump PRs


debasishbsws:
Update epoch of downstream packages as there is a so lib version update

debasishbsws:
Why in ci while building tensoreflow-core when we installing `py3-gpep517` it installing `python3.11` with it, Locallly in my PC it comming with `python3.12`. Kinada similar issue https://github.com/chainguard-dev/melange/issues/766

PR #7958, opened 2023-11-03
openjdk-11: track upstream GA releases and not pre-release builds

rawlingsj:
We recently discovered the OpenJDK tag versioning includes pre-release builds.  There's more details in the Timeline section of the [upstream docs](https://wiki.openjdk.org/display/JDKUpdates/JDK11u).

This PR changes the OpenJDK 11 version to be the GA release and switched the `update:` config to track only GA releases.

After reviewing the upstream tags, when a new patch release is created **AFTER** a GA release, the git commit sha is updated on the GA tag.  Meaning we will still receive automated updates when security patches are created for GA releases, the bot will update the `expected-commit` sha and bump the epoch.

_**Note:**_ We are not withdrawing the `11.0.22.1` version as users will have likely already been using it.


rawlingsj:
The wolfictl update check is expected to fail.  I have a fix for for the bot but just running an end to end test against my fork of wolfi.  We can ignore this failed check here in the meantime.

openjdk-11.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: openjdk-11
-  version: 11.0.22.1
-  epoch: 0
+  version: 11.0.21 # on version bump ensure to update downstream projects that are pinned to a specific version, i.e. https://github.com/chainguard-images/images/tree/main/images
```
imjasonh, line 3:
```suggestion
  version: 11.0.21 # TODO(jason): on version bump, ensure to update downstream projects that are pinned to a specific prerelease version
```

imjasonh (dismissed):

mattmoor (dismissed):
I think we will need a withdraw for this to be effective?

imjasonh:
> I think we will need a withdraw for this to be effective?

I'd like to avoid it if possible. We can pin to the same package as needed downstream, until this surpasses it and we can remove the pin.

stormqueen1990 (approved):

PR #7970, opened 2023-11-03
nvidia: fix GHSA-m425-mq94-257g

hectorj2f:
```
wolfictl scan packages/aarch64/nvidia-device-plugin-0.14.2-r2.apk
🔎 Scanning "packages/aarch64/nvidia-device-plugin-0.14.2-r2.apk"
✅ No vulnerabilities found
```

### DIFF
--- a/nvidia-device-plugin.yaml
+++ b/nvidia-device-plugin.yaml
@@ -1,7 +1,7 @@
 package:
   name: nvidia-device-plugin
   version: 0.14.2
-  epoch: 1
+  epoch: 2
   description: NVIDIA device plugin for Kubernetes
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


nvidia-device-plugin.yaml
```diff
@@ -38,6 +41,7 @@ pipeline:
 
 update:
   enabled: true
+  manual: false
```
luhring, line 44: Why this change? Just curious

hectorj2f: This is a default value, honestly I don't see a problem to keep it. But I could change my script.

hectorj2f: My code is also re-ordering the properties according to the initial yaml structure.

luhring: AFAIK we normally don't have `manual: false` in these files

hectorj2f: @luhring Almost all the py3- files set it to false. https://github.com/wolfi-dev/os/pull/6863/files#diff-5fbae266c308c5a4453523df15a67dabe1e6a8e8770be85ea47632df4534be66R57 We have many more files with this property and values true | false.

luhring (approved):

PR #7980, opened 2023-11-03
fix up update config for packages

rawlingsj:
In preparation for https://github.com/wolfi-dev/wolfictl/pull/458 let's add `ignore-regex-patterns:` for when `tag-filter` **contains** the value rather than **starts** with it.

Also favour using GitHub Releases when they are available.


rawlingsj:
The update check fails because this PR fixes up the update config for those packages.  The bot will update them after this merges.

The scan check fails possibly as some packages are old.  

Both are non required checks let's merge this and let the bot do it's job.

ajayk (approved):

PR #8027, opened 2023-11-06
withdraw kyverno 1.10.5 rc versions

ajayk:
with this commit https://github.com/wolfi-dev/os/commit/8ee67ec3cf7f0237d7ca157d34f1649dde3af1b7 we picked up an rc version of kyverno and released it as 1.10.5 due to a filtering issue in automation which was not ignoring the beta and rc tags 

This has been fixed in https://github.com/wolfi-dev/os/commit/85c4c5b32fccc69b98c62c96e88c251e5cb0b577

so we need to withdraw the 1.10.5 versions of kyverno apks as they are built from an rc version

### DIFF
--- a/withdrawn-packages.txt
+++ b/withdrawn-packages.txt
@@ -1,4 +1,6 @@
-py3-certifi-2023.7.22-r0.apk
-py3-certifi-2023.5.7-r0.apk
-nspr-4.35-r0.apk
-clang-16-16.0.6-r1.apk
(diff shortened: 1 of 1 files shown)


hectorj2f (approved):
@ajayk Could you add a comment to the description for everyone to understand why we withdraw these packages ?

ajayk:
> @ajayk Could you add a comment to the description for everyone to understand why we withdraw these packages ?

PR comment updated

PR #8031, opened 2023-11-06
actions-runner-controller/0.6.1 package update (Fix #7990)

debasishbsws:
Fixes: #7990

Related:

Removed the `make manager` command as it is unnecessary to build the same thing twice. and the codegen that was mentioned is just adding [hedders](https://github.com/actions/actions-runner-controller/blob/master/hack/boilerplate.go.txt) to the go files.
Without it, everything works fine.
I am not sure why I am getting the error using melange but not when I interactively do it inside a wolfi-base container.

#### For version bump PRs

### DIFF
--- a/actions-runner-controller.yaml
+++ b/actions-runner-controller.yaml
@@ -1,7 +1,7 @@
 package:
   name: actions-runner-controller
   version: 0.6.1
-  epoch: 3
+  epoch: 0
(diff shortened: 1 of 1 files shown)


debasishbsws:
cc @ajayk

ajayk (approved):

PR #8085, opened 2023-11-06
add bitnami-compat for postgres12-16

Dentrax:


pgaudit-11.yaml
```diff
@@ -0,0 +1,38 @@
+# pgAudit v1.3.X is intended to support PostgreSQL 11.
+# https://github.com/pgaudit/pgaudit/blob/master/README.md#postgresql-version-compatibility
+package:
+  name: pgaudit-11
+  version: 1.3.4
+  epoch: 0
+  description: PostgreSQL Audit Extension
+  copyright:
+    - license: BSD
```
mamccorm, line 9: License looks fine. is actually 'PostgreSQL License' but I see they note its to be interrupted similar to BSD, so this is okay.

pgaudit-11.yaml
```diff
@@ -0,0 +1,38 @@
+# pgAudit v1.3.X is intended to support PostgreSQL 11.
+# https://github.com/pgaudit/pgaudit/blob/master/README.md#postgresql-version-compatibility
+package:
+  name: pgaudit-11
```
mamccorm, line 4: this could get confusing having the package name as pgaudit-11 whenever the version of pgaudit is pgaudit-1.3.

That said if we change it, it'll be as equally confusing when the version doesn't match the postgres version. ACK their docs for version compatibilty: https://github.com/pgaudit/pgaudit/blob/master/README.md#postgresql-version-compatibility.

All that considered i think this is fine as is

mamccorm (dismissed):

ajayk:
the ftp sources align with the release notes from Postgres community , can we revert the change for now

ajayk (approved):
Looks Good , I need an another pair 👀 to look into this

imjasonh (approved):

PR #8172, opened 2023-11-08
Remove subpackages that use if

jonjohnsonjr:
This is only used in a single place to make openssl configurable in a way that is no longer used, so this just adds complexity. Drop it so we can drop the logic in melange for handling subpackage ifs.


kaniini:
I intentionally added `if` conditionals to subpackages, as a first-class feature of the Melange DSL.

What is the rationale for removing them?

PR #8396, opened 2023-11-12
spire: fix GHSA-vfp6-jrw2-99g9

hectorj2f:
```
wolfictl scan  packages/aarch64/spire-agent-1.8.3-r1.apk --govulncheck
🔎 Scanning "packages/aarch64/spire-agent-1.8.3-r1.apk"
✅ No vulnerabilities found
```

### DIFF
--- a/spire-server.yaml
+++ b/spire-server.yaml
@@ -1,7 +1,7 @@
 package:
   name: spire-server
   version: 1.8.4
-  epoch: 0
+  epoch: 1
   description: The SPIFFE Runtime Environment (SPIRE) server
   copyright:
     - license: Apache-2.0
@@ -46,6 +46,9 @@ pipeline:
       esac
       echo ${goversion} > .go-version
       cat .go-version
+
+      go get github.com/sigstore/cosign/v2@v2.2.1
+      go mod tidy
 
       make bin/spire-agent
       make bin/spire-server
--- a/spire-server.yaml
+++ b/spire-server.yaml
@@ -47,6 +47,7 @@ pipeline:
       echo ${goversion} > .go-version
       cat .go-version
 
+      # Mitigate CVE-2023-46737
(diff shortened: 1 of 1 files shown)


spire-server.yaml
```diff
@@ -47,6 +47,9 @@ pipeline:
       echo ${goversion} > .go-version
       cat .go-version
 
+      go get github.com/sigstore/cosign/v2@v2.2.1
```
pdeslaur, line 50: If possible, could you leave a comment with the CVE ID: CVE-2023-46737

pdeslaur:
> ```
> wolfictl scan  packages/aarch64/spire-agent-1.8.3-r1.apk --govulncheck
> 🔎 Scanning "packages/aarch64/spire-agent-1.8.3-r1.apk"
> ✅ No vulnerabilities found
> ```

The scan above is for `1.8.3-r1`, but this PR is `1.8.4-r1` I wonder if we need the fix in the first place? 🤔

hectorj2f:
>  The scan above is for 1.8.3-r1, but this PR is 1.8.4-r1 I wonder if we need the fix in the first place? 🤔

No, that was me wrongly pasting a scan result of an older version that I previously patched. Once I rebased the latest version 1.8.4, I found the same results as you can see in the pushed yaml.

```
wolfictl scan  packages/aarch64/spire-agent-1.8.4-r1.apk --govulncheck
🔎 Scanning "packages/aarch64/spire-agent-1.8.4-r1.apk"
✅ No vulnerabilities found
```

spire-server.yaml
hectorj2f: Added

pdeslaur (approved):

hectorj2f:
@pdeslaur Thanks for the review.

PR #8401, opened 2023-11-12
nodetaint: Bump k8s dependencies

pdeslaur:
Bump a number of Kubernetes packages to reduce the CVE count.

### DIFF
--- a/nodetaint.yaml
+++ b/nodetaint.yaml
@@ -1,7 +1,7 @@
 package:
   name: nodetaint
   version: 0.0.4
-  epoch: 6
+  epoch: 7
   description: Controller to manage taints for nodes in a k8s cluster.
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


pdeslaur:
I'm not sure anymore about the `-dropreplace` here... I may just force a specific version instead

nodetaint.yaml
```diff
@@ -22,18 +22,44 @@ pipeline:
       expected-commit: 9f648717bd31a782696335450ee05a5ed79cd870
 
   - runs: |
-      # Mitigate  GHSA-3vm4-22fp-5rfm, GHSA-gwc9-m7rh-j2ww and GHSA-8c26-wmh5-6g9v
+      # Mitigate GHSA-3vm4-22fp-5rfm, GHSA-gwc9-m7rh-j2ww and GHSA-8c26-wmh5-6g9v
```
vaikas, line 25: Can't comment on L29, but that's woefully out of date.
https://pkg.go.dev/golang.org/x/text?tab=versions

pdeslaur: Yeah... The upstream repo has an old version, so believe it or not, but v0.3.8 is still an update: https://github.com/wish/nodetaint/blob/master/go.sum#L655

I prefer to not always "upgrade to latest" unless we need to - but happy to bump it!

vaikas (approved):
Thank you!!!

nodetaint.yaml
vaikas: Yeah, makes sense.

PR #8475, opened 2023-11-14
fix recovery.conf.sample not found for postgresql-11

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/postgresql-11.yaml
+++ b/postgresql-11.yaml
@@ -1,7 +1,7 @@
 package:
   name: postgresql-11
   version: "11.22"
-  epoch: 2
+  epoch: 3
   description: A sophisticated object-relational DBMS
   copyright:
     - license: BSD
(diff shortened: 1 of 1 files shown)


cpanato (dismissed):
thanks

developer-guy:
> It looks like this affects all postgres bitnami images

they said they dropped the support of that file, here, https://www.postgresql.org/docs/current/recovery-config.html#RECOVERY-CONFIG. We might not need that for other postgres besides 11.

developer-guy:
> Can we symlink the entire directory instead of the single file as well?

Yeah for sure but we already created bunch of symlinks for conf files by changing their names from `.sample` to `.conf`, so, I think I should remain them as is and create symlink for others.

found-it (approved):

found-it:
Looks good, thank you @developer-guy

postgresql-11.yaml
```diff
@@ -115,8 +115,32 @@ subpackages:
           mkdir -p ${{targets.subpkgdir}}/opt/bitnami/postgresql
           mkdir -p ${{targets.subpkgdir}}/opt/bitnami/postgresql/conf
           mkdir -p ${{targets.subpkgdir}}/opt/bitnami/postgresql/bin
+          mkdir -p ${{targets.subpkgdir}}/opt/bitnami/postgresql/share
+
           ln -sf /usr/share/postgresql/pg_hba.conf.sample ${{targets.subpkgdir}}/opt/bitnami/postgresql/conf/pg_hba.conf
           ln -sf /usr/share/postgresql/postgresql.conf.sample ${{targets.subpkgdir}}/opt/bitnami/postgresql/conf/postgresql.conf
+
+          # Create symbolic links individually for each file
```
imjasonh, line 123: Does it make sense to symlink wildcards? Or the whole directory?

```
ln -sf /usr/share/postgresql/ ${{targets.contextdir}}/opt/bitnami/postgresql/share
```

developer-guy: I did that way because there are files which we created symlink by changing the names like:

```sh
          ln -sf /usr/share/postgresql/pg_hba.conf.sample ${{targets.subpkgdir}}/opt/bitnami/postgresql/conf/pg_hba.conf
          ln -sf /usr/share/postgresql/postgresql.conf.sample ${{targets.subpkgdir}}/opt/bitnami/postgresql/conf/postgresql.conf
```

imjasonh (approved):

kaniini (approved):

PR #8513, opened 2023-11-14
Add tinydir v1.2.5, valijson v1.0.1 and uthash v2.3.0 for building sinsp

srishtih:
…lco-sinsp

Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/tinydir.yaml
+++ b/tinydir.yaml
@@ -1,0 +1,32 @@
+package:
+  name: tinydir
+  version: "1.2.5"
+  epoch: 0
+  description: "Lightweight, portable C library that simplifies interaction with directories and files in a filesystem"
+  copyright:
+    - license: "BSD-2-Clause"
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - wolfi-baselayout
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/cxong/tinydir.git
+      tag: ${{package.version}}
+      expected-commit: 4acafe7916720645bf65e19be8603acfe7acdba1
+
+  - runs: mkdir -p ${{targets.destdir}}/usr/include
+
+  - runs: cp tinydir.h ${{targets.destdir}}/usr/include/
+
+  - uses: strip
+
+update:
+  enabled: true
+  github:
+    identifier: cxong/tinydir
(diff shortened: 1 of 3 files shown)


tinydir.yaml
```diff
@@ -0,0 +1,32 @@
+package:
```
mamccorm, line 1: This file looks good, the only concern I have though is it doesn't look to be actively maintained (last commit ~2 years ago). Great there are no current vulnerabilities however.

Id establish if these additional packages resolve the falco build issues - if so then we'll decide whether this is okay for wolfi or needs to be private (due to not being actively maintained)

uthash.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+  - runs: cp src/uthash.h ${{targets.destdir}}/usr/include/
+
+  - uses: strip
+
+update:
+  enabled: true
+  github:
+    identifier: troydhanson/uthash
+    strip-prefix: v
```
mamccorm, line 33: The upstream repo doesn't publish releases, only tags - so we need to specify to poll tags here:

```suggestion
    strip-prefix: v
    use-tag: true
```

uthash.yaml
```diff
@@ -0,0 +1,33 @@
+package:
```
mamccorm, line 1: Similar to above, doesn't look to be actively maintained (~1 year). We might want to discuss whether moving this to private packages is a better option given this

valijson.yaml
```diff
@@ -0,0 +1,47 @@
...
+  - name: valijson-dev
+    description: "C++ library for validating JSON - development headers"
+    pipeline:
+      - uses: split/dev
+
+update:
+  enabled: true
+  github:
+    identifier: tristanpenman/valijson
+    strip-prefix: v
```
mamccorm, line 47: This will require some additional logic, as the release names are in the format: 'Valijson v1.0.1'. i.e need to strip off the preceding 'Valijson '. I don't have an example to hand but this is implemented for other packages so we should be able to find something to reference

tinydir.yaml
kaniini: I think it is fine.  It's just a small header library with very simple functionality.  The risk is low for having it.

uthash.yaml
kaniini: This is also a simple header-only data structure library.  Its low risk to have.

rawlingsj (approved):

PR #8670, opened 2023-11-16
`dav1d/1.3.0` was failing soname checks, So I updated the epoch of downstream packages

### LINKED ISSUE #7031: Fix dav1d/1.3.0 package update package update (6331)

debasishbsws:
Fixes: #7031 #6331 

#### For version bump PRs


rawlingsj:
Great thanks - @kaniini the soname check is failing but the bumped epoch for the downstream dep's build passed.

If looks good to you we could do with an override merge here.

kaniini (approved):

PR #8675, opened 2023-11-16
`aerospike/7.0.0.0` package update

debasishbsws:
Fixes: #8536 

#### For version bump PRs


debasishbsws:
cc @rawlingsj @ajayk

rawlingsj:
Thanks for this @debasishbsws!  

I noticed that this is a major version update and that there are notes on how to upgrade https://docs.aerospike.com/server/operations/upgrade/special_upgrades/70_upgrade

I'm just looking for the aerospike supported versions as we may need to convert this to a version stream, i.e. a new package called `aerospike-7` with the config in this PR.

Just going to pop this on hold while we check.

rawlingsj:
Found the support version docs https://docs.aerospike.com/reference/platform-support#aerospike-database-release-ga-and-support-schedules

And indeed it looks like aerospike-server should be a version stream, i.e. the current version 6.4 that's in wolfi is still being maintained and supported until August 3, 2025.

So I think what we need to do is 
- create a new package `aerospike-7` which contains the config from this PR
- in the update config block we can then tell the bot to only track the `7` version stream using `tag-filter-prefix: 7.`
- also update the existing aerospike package update config to track the 6.4 stream, i.e. `tag-filter-prefix: 6.4.`

@debasishbsws I know you've already done some good work on this, do you fancy having a go at the steps above?

rawlingsj (changes requested):
See thread above, looks like aerospike needs to be a version stream

debasishbsws:
@rawlingsj Working on it.

aerospike-6.4.yaml
rawlingsj, line 1: for compatibility we don't need to rename this file, else we'd also need to rename the `package.name` but that would mean any downstream usages of aerospike would stop receiving updates.

rawlingsj (changes requested):
All looks great apart from that one renamed file comment above, if you can rename that back to aerospike.yaml then I think we can merge

aerospike-6.4.yaml
debasishbsws: Okay, I understand, I was a mistake from my side.

rawlingsj (approved):

rawlingsj:
Amazing - thanks @debasishbsws

PR #8688, opened 2023-11-16
envoy: build envoys using libc++ as intended by upstream

### LINKED ISSUE #8486: envoy: build using libc++
Envoy defaults builds to using `libc++`, it supposedly has some performance benefits. https://github.com/envoyproxy/envoy/issues/4251
Is there a reason for wolfi not to build against `libc++`?

This would mean:
- Adding a package `llvm-libcxx-15` because envoy builds using llvm-15
- Fix builds of `llvm-libcxx`, currently envoy fails to build with current `llvm-libcxx` packages in wolfi, this is apparently because of this: https://github.com/wolfi-dev/os/blob/main/llvm-libcxx-16.yaml#L42-L44

I have successfully built `llvm-libcxx` without those options and managed to successfully build envoy using that `llvm-libcxx`, I can open PRs for this.

PD: this also impacts istio-envoy which should build with `libcxx`

sschepens:
Make envoy builds use `libc++` as intended by upstreams.

Fixes: #8486

Related:

#### For PRs that add patches

### DIFF
--- a/envoy.yaml
+++ b/envoy.yaml
@@ -1,7 +1,7 @@
 package:
   name: envoy
   version: 1.28.0
-  epoch: 0
+  epoch: 1
   description: Cloud-native high-performance edge/middle/service proxy
   copyright:
     - license: Apache-2.0
@@ -28,10 +28,11 @@ environment:
       - llvm-lld-15
       - llvm15-tools
       - llvm15-cmake-default
+      - llvm-libcxx-15
+      - llvm-libcxx-15-dev
+      - llvm-libcxxabi-15
       - coreutils
       - patch
-      # We need to stick to gcc 12 for now, envoy doesn't build with gcc >= 13
-      - gcc-12-default
 
 pipeline:
   - uses: git-checkout
@@ -42,10 +43,6 @@ pipeline:
       destination: envoy
 
   - runs: |
-      # Apply https://github.com/envoyproxy/envoy/pull/29261/files so build
-      # can work with make > 4.3
-      cp -f luajit.patch envoy/bazel/foreign_cc
-
       export JAVA_HOME=/usr/lib/jvm/java-11-openjdk
       mkdir -p .cache/bazel/_bazel_root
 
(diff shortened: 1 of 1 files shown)


envoy.yaml
```diff
@@ -42,10 +43,6 @@ pipeline:
       destination: envoy
 
   - runs: |
-      # Apply https://github.com/envoyproxy/envoy/pull/29261/files so build
```
sschepens, line 45: this patch is no longer necessary as of envoy `1.28.0`

istio-envoy-1.20.yaml
```diff
@@ -52,8 +50,6 @@ pipeline:
       export JAVA_HOME=/usr/lib/jvm/java-11-openjdk
       mkdir -p .cache/bazel/_bazel_root
 
-      echo "build --config=clang" >> user.bazelrc
```
sschepens, line 55: istio's `.bazelrc` already specifies `--config=libc++20`, stop forcing `clang`

sschepens:
@ajayk you recently pushed `istio-envoy-1.20` and correctly configured dependencies for `libcxx`, but you forgot to remove the forced `--config=clang`

rawlingsj (dismissed):

rawlingsj:
This is awesome - thanks @sschepens!

sschepens:
hmm @rawlingsj seems like CI was cancelled, maybe it was taking too long? might be worth splitting into multiple PRs?

ajayk:
> hmm @rawlingsj seems like CI was cancelled, maybe it was taking too long? might be worth splitting into multiple PRs?

@sschepens yeah CI times out at 6 hours , can we split it up so its easier to get to main

sschepens:
@rawlingsj on it

sschepens:
@rawlingsj modified this PR to only build envoy and created #8724 #8726 #8727 for each istio release

ajayk (approved):

PR #8715, opened 2023-11-16
Support incremental package uploads

### LINKED ISSUE #8698: Feature: Incremental Package Uploads
A couple existing attempts at this:
* https://github.com/wolfi-dev/os/pull/6454
* https://github.com/wolfi-dev/os/pull/6483

I'll try my hand at a third with some new context.

We don't want to have each `APKINDEX` become out of sync across architectures (for very long) because it means our downstream builds will fail to lock consistent versions.

This made incremental package uploads ~impossible to do safely because we used to withdraw packages by deleting certain APKs and re-indexing the entire set of APKs:
https://github.com/wolfi-dev/os/pull/6454/files#r1349767488

If we uploaded a package for one arch and not the other, running a withdraw in the meantime would add that package to only one APKINDEX.

With the [new](https://github.com/wolfi-dev/os/pull/8090) `wolfictl withdraw` based workflow, we just remove certain packages explicitly rather than looking at what APKs exist. This means we can partially uploaded packages without worrying about prematurely adding them to the index.

So as a first pass at this, we can upload any APK that we successfully built, but we will upload a new APKINDEX if and only if we successfully build _every_ APK. This means that we will make some progress on the slow parts (building the APKs) if any of them succeed.

On the other hand, this also means that packages will only be available (in the APKINDEX) once we have fixed all the broken builds. That is no worse than today, but attempting to partially update the APKINDEX is complicated and requires tracking a lot more state than I'm comfortable doing in yamlbash.

Something worth mentioning is that this only works because the APKINDEX generation is ~level-triggered in that it looks at every APK in the repo, so the next successful build will index any uploaded-but-not-indexed packages. If we ever switch to a different workflow that only considers the APKINDEX and any newly built packages, we could accidentally orphan packages.

jonjohnsonjr:
Fixes https://github.com/wolfi-dev/os/issues/8698

Currently, we build any out of date packages as part of this workflow. If the build step of the workflow failsf or any package, we don't proceed to the "upload" job, which means we don't upload any of the packages that were successfully built.

This change splits the "upload" job into "upload-packages" and "upload-index".

The "upload-packages" job will always run. It signs the packages and indexes, but only uploads the packages to GCS; the indexes get uploaded to GHA Artifacts.

The "upload-index" job only runs if the "build" job succeeded. It downloads the indexes from GHA Artifacts and uploads them to GCS.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -79,12 +79,17 @@ jobs:
             MELANGE_EXTRA_OPTS="--keyring-append=/gcsfuse/wolfi-registry/wolfi-signing.rsa.pub" \
             all -j1
 
-      - run: |
+      # Always run this step for https://github.com/wolfi-dev/os/issues/8698
+      - if: ${{ always() }}
+        name: 'Create artifacts tarball'
+        run: |
           # Clean up the symlinks and create an archive for uploading
           find ./packages/${{ matrix.arch }} -type l -exec rm -f {} \;
           tar -cvzf /tmp/packages-${{ matrix.arch }}.tar.gz ./packages/${{ matrix.arch }}
 
-      - name: 'Upload built packages archive to Github Artifacts'
+      # Always run this step for https://github.com/wolfi-dev/os/issues/8698
+      - if: ${{ always() }}
+        name: 'Upload built packages archive to GitHub Artifacts'
         uses: actions/upload-artifact@v3
         with:
           name: packages-${{ matrix.arch }}
@@ -92,9 +97,12 @@ jobs:
           retention-days: 1 # Low ttl since this is just an intermediary used once
           if-no-files-found: warn
 
-  upload:
+  upload-packages:
     runs-on: ubuntu-latest
     needs: build
+
+    # Always run this job for https://github.com/wolfi-dev/os/issues/8698
+    if: ${{ always() }}
 
     permissions:
       id-token: write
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -159,18 +167,9 @@ jobs:
...
-            # Don't cache the APKINDEX.
-            gcloud --quiet storage cp \
-                --cache-control=no-store \
-                "./packages/${arch}/APKINDEX.tar.gz" "gs://wolfi-production-registry-destination/os/${arch}/"
-
-            gcloud --quiet storage cp \
-                --cache-control=no-store \
-                "./packages/${arch}/APKINDEX.json" "gs://wolfi-production-registry-destination/os/${arch}/"
-
             # Only attempt to sign when *.apk's exist
```
imjasonh, line 173: This comment seems to be out of place now? Was it always?

imjasonh (dismissed):

.github/workflows/build.yaml
jonjohnsonjr: Updated s/sign/upload/

jonjohnsonjr:
My confidence in this is medium-low because I don't have a way to really test the whole thing e2e, so if this breaks anything we should not hesitate to revert it.

jdolitsky (approved):

imjasonh (approved):

PR #7401, opened 2023-10-25
fixing oci entrypoints for using temporal-server

mritunjaysharma394:
Fixes:

Related:

#### For new package PRs only

- [ ] This PR is marked as fixing a pre-existing package request bug
  - [ ] Alternatively, the PR is marked as related to a pre-existing package request bug, such as a dependency
- [x] REQUIRED - The package is available under an OSI-approved or FSF-approved license
- [x] REQUIRED - The version of the package is still receiving security updates
- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`)

#### For new version streams

- [ ] The upstream project actually supports multiple concurrent versions.
- [ ] Any subpackages include the version string in their package name (e.g. `name: ${{package.name}}-compat`)
- [ ] The package (and subpackages) `provides:` logical unversioned forms of the package (e.g. `nodejs`, `nodejs-lts`)

#### For security-related PRs

- [ ] The security fix is recorded in the [advisories](https://github.com/wolfi-dev/advisories) repo

#### For version bump PRs

- [ ] The `epoch` field is reset to 0

#### For PRs that add patches

- [ ] Patch source is documented

PR #8716, opened 2023-11-16
adds temporal-docker-builds package

mritunjaysharma394:
Needed for fixing entrypoints in https://github.com/wolfi-dev/os/pull/7401/files#diff-a96a0e1f38ff3cd582e50a8b59770f2cc04245eacdecea305ebad441a16d39feR71

Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches


temporal-docker-builds.yaml
```diff
@@ -0,0 +1,49 @@
...
+      extract: "false"
+      uri: https://github.com/temporalio/docker-builds/archive/${{vars.commit}}.zip
+
+  - runs: |
+      unzip -q ${{vars.commit}}.zip
+      mv docker-builds-${{vars.commit}} docker-builds
+      rm *.zip
+
+  - runs: |
+      mkdir -p "${{targets.destdir}}"/temporal-docker-builds
```
jonjohnsonjr, line 34: Maybe put these files under `/etc/temporal` since that's where they end up in the dockerfiles?

jonjohnsonjr: E.g. https://github.com/temporalio/docker-builds/blob/347568012222589dbe2577dcdc7f12b6855a72c2/server.Dockerfile#L64-L65

temporal-docker-builds.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - busybox
+      - ca-certificates-bundle
+      - build-base
+
+pipeline:
+  - uses: fetch
+    with:
+      expected-sha256: 2eb296dc8493320261c762377356dedc23a4b3e4040502c21618e61b407a82a5
+      extract: "false"
+      uri: https://github.com/temporalio/docker-builds/archive/${{vars.commit}}.zip
```
ajayk, line 26: - uses: fetch
```suggestion
      expected-sha256: 5cea800b194c364ce182a2f917f8736476ef60a6e1ef8a830665a1a7bdec2cd6 
      uri: https://github.com/temporalio/docker-builds/archive/${{vars.commit}}.tar.gz
```

temporal-docker-builds.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+  - runs: |
+      unzip -q ${{vars.commit}}.zip
+      mv docker-builds-${{vars.commit}} docker-builds
+      rm *.zip
```
ajayk, line 31:
```suggestion
```

temporal-docker-builds.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+  - runs: |
+      mkdir -p "${{targets.destdir}}"/temporal-docker-builds
+
+      cp docker-builds/docker/entrypoint.sh "${{targets.destdir}}"/temporal-docker-builds
```
ajayk, line 36:
```suggestion
      cp docker/entrypoint.sh "${{targets.destdir}}"/temporal-docker-builds
```

ajayk: similar to other copies as above

temporal-docker-builds.yaml
mritunjaysharma394: Thanks @jonjohnsonjr, I was thinking of doing it in the main temporal-server repo? What will be better? https://github.com/wolfi-dev/os/pull/7401/files#diff-a96a0e1f38ff3cd582e50a8b59770f2cc04245eacdecea305ebad441a16d39feR73

mritunjaysharma394: If we can do it here itself, while others there, I think I am okay with that, eventually I hope it won't conflict?

temporal-docker-builds.yaml
mritunjaysharma394: Thanks @ajayk, just cleaning up all that!

temporal-docker-builds.yaml
mritunjaysharma394: And can we even do that in main temporal-server repo by including this package as runtime dep? I think this should be an environment dep for our cause?

jonjohnsonjr: Having them under `/temporal-docker-builds/` doesn't give us anything, right? It's just a temporary location that could be anything?

If that's true, I think it makes sense to just put them under `/etc/temporal` so we can skip the copies in `temporal-server-oci-entrypoint`.

> And can we even do that in main temporal-server repo by including this package as runtime dep?

I think that would work, yeah.

mritunjaysharma394: Yes, I guess, thanks for this suggestion, just fixed it @jonjohnsonjr !

mritunjaysharma394: Just fixed this too https://github.com/wolfi-dev/os/pull/7401/commits/b27d619732be5fc789fe7b5727e77f70c4083662

jonjohnsonjr (approved):

PR #8771, opened 2023-11-17
nginx-mainline: add brotli module

vr:
copying what was done in nginx-ingress

### DIFF
--- a/nginx-mainline.yaml
+++ b/nginx-mainline.yaml
@@ -38,6 +38,8 @@ environment:
 data:
   - name: modules
     items:
+      http_brotli_filter: "10"
+      http_brotli_static: "10"
       http_geoip: "10"
       http_image_filter: "10"
       http_perl: "10"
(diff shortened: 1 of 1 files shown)


nginx-mainline.yaml
```diff
@@ -53,6 +55,17 @@ pipeline:
     with:
       expected-sha256: 64c5b975ca287939e828303fa857d22f142b251f17808dfe41733512d9cded86
       uri: https://nginx.org/download/nginx-${{package.version}}.tar.gz
+  - uses: git-checkout
+    with:
+      repository: https://github.com/google/ngx_brotli.git
+      branch: master
+      expected-commit: a71f9312c2deb28875acc7bacfdd5695a111aa53
+      destination: ngx_brotli
```
rawlingsj, line 63: I think this should probably change to be 
```
   - runs: |
        git clone https://github.com/google/ngx_brotli.git
        git checkout a71f9312c2deb28875acc7bacfdd5695a111aa53
```
Otherwise the next time this build runs the expected commit sha will not match and the build will fail.

Unfortunately we don't have an automated way to handle keeping multiple git-checkout.

rawlingsj: Thinking about it a bit more, a better solution would probably be to create a new `ngx_brotli` package and add a `environment.contents.packages` dependency for the new `ngx_brotli` which can then be used for `--add-dynamic-module=./ngx_brotli`

rawlingsj (changes requested):

nginx-mainline.yaml
rawlingsj: @vr did you want to try that?  I'm happy to if you prefer, up to you.  This way both nginx-mainline and nginx-ingress can share the same ngx_brotli package.

vr: i'm not sure you can build once for both nginx version 🤔

vr: and i was not able to directly use `git` without having `uses: git-checkout` before because `git` command was not available before

PR #8814, opened 2023-11-18
Add minizip subpackage to zlib.

dlorenc:
This utility is part of zlib but not built by default. It's commonly packaged into other distros.

Fixes:

Related:

### DIFF
--- a/zlib.yaml
+++ b/zlib.yaml
@@ -1,7 +1,7 @@
 package:
   name: zlib
   version: "1.3"
-  epoch: 1
+  epoch: 2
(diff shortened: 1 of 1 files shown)


rawlingsj (approved):

mswilson:
I wouldn't recommend doing this. minizip is "contrib" software in the zlib source release. Other Linux distros are moving away from packaging minizip in favor of https://github.com/zlib-ng/minizip-ng. See for example  https://fedoraproject.org/wiki/Changes/MinizipNGTransition

PR #8825, opened 2023-11-19
Bump opentofu to alpha4.

dlorenc:
There's also an alpha5 but for completeness we should get them all and bump to that one next.


dlorenc:
The update check is expected to fail here until we get the real release.

developer-guy (approved):

PR #8880, opened 2023-11-21
openmp-17: new version stream for openmp

ajayk:


openmp-17.yaml
```diff
@@ -0,0 +1,71 @@
...
+      - cmake
+      - samurai
+      - binutils-dev
+      - libffi-dev
+      - zlib-dev
+      - libxml2-dev
+      - pkgconf
+      - python3
+      - help2man
+      - llvm17
```
hectorj2f, line 25: Do we need both llvm17 and llvm17-dev?

openmp-17.yaml
```diff
@@ -0,0 +1,71 @@
...
+  version: 17.0.5
+  epoch: 0
+  description: "LLVM OpenMP library"
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - wolfi-base
```
hectorj2f, line 12: nit: Wolfi-base already includes the `ca-certificates-bundle` in case you want to simplify it.

openmp-17.yaml
```diff
@@ -0,0 +1,71 @@
...
+      - pkgconf
+      - python3
+      - help2man
+      - llvm17
+      - llvm-cmake-17
+      - llvm17-dev
+      - libLLVM-17
+
+pipeline:
+  - uses: fetch
```
hectorj2f, line 31: suggestion: we could use git-checkout here, which sounds like the ideal choice when possible.

openmp-17.yaml
ajayk: yes we need both 

llvm17 provides `usr/lib/llvm17/bin/llvm-tblgen` which is needed here

openmp-17.yaml
ajayk: wolf-base doesn't include  `ca-certificates-bundle` , its build-base which provides  `ca-certificates-bundle` 
 build-base -> wolfi-baselyout -> ca-certificates-bundle

hectorj2f: question: do we care about adding a package if another package already includes it ?

hectorj2f (approved):

openmp-17.yaml
ajayk: I think we should remove them , its spread across various packages , there might be some reasoning behind it , i am not completely sure

openmp-17.yaml
ajayk: will address this in the upcoming pr

PR #8897, opened 2023-11-21
Fix binary locations for temporal package and it's sub packages

shivanshuraj1333:
Fix binary locations for the temporal package and it's sub-packages

### DIFF
--- a/tctl.yaml
+++ b/tctl.yaml
@@ -1,7 +1,7 @@
 package:
   name: tctl
   version: 1.18.0
-  epoch: 1
+  epoch: 2
   description: Temporal CLI
   copyright:
     - license: MIT License
@@ -37,10 +37,19 @@ pipeline:
       go mod tidy
 
       make build
-      install -Dm755 tctl ${{targets.destdir}}/usr/bin/tctl
-      install -Dm755 tctl-authorization-plugin ${{targets.destdir}}/usr/bin/tctl-authorization-plugin
+      install -Dm755 tctl ${{targets.destdir}}/usr/local/bin/tctl
+      install -Dm755 tctl-authorization-plugin ${{targets.destdir}}/usr/local/bin/tctl-authorization-plugin
 
   - uses: strip
+
+subpackages:
+  - name: ttctl-authorization-plugin
+    description: "tctl-authorization-plugin"
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/usr/local/bin
+          install -Dm755 tctl-authorization-plugin ${{targets.destdir}}/usr/local/bin/tctl-authorization-plugin
+      - uses: strip
 
 update:
   enabled: true
--- a/tctl.yaml
+++ b/tctl.yaml
@@ -42,7 +42,6 @@ pipeline:
   - uses: strip
 
 subpackages:
-
   - name: tctl-compat
     description: Compat package for tctl
     pipeline:
--- a/tctl.yaml
+++ b/tctl.yaml
@@ -37,17 +37,33 @@ pipeline:
       go mod tidy
 
       make build
-      install -Dm755 tctl ${{targets.destdir}}/usr/local/bin/tctl
+      install -Dm755 tctl ${{targets.destdir}}/usr/bin/tctl
 
   - uses: strip
 
 subpackages:
-  - name: ttctl-authorization-plugin
-    description: "tctl-authorization-plugin"
+
+  - name: tctl-compat
+    description: Compat package for tctl
     pipeline:
       - runs: |
           mkdir -p "${{targets.subpkgdir}}"/usr/local/bin
-          install -Dm755 tctl-authorization-plugin ${{targets.destdir}}/usr/local/bin/tctl-authorization-plugin
+          ln -s /usr/bin/tctl "${{targets.subpkgdir}}"/usr/local/bin
+
+  - name: tctl-authorization-plugin
+    description: "tctl-authorization-plugin"
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/usr/bin
+          install -Dm755 tctl-authorization-plugin ${{targets.subpkgdir}}/usr/bin/tctl-authorization-plugin
+      - uses: strip
+
+  - name: tctl-authorization-plugin-compat
+    description: "Compat package for tctl-authorization-plugin"
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/usr/local/bin
+          ln -s /usr/bin/tctl-authorization-plugin "${{targets.subpkgdir}}"/usr/local/bin
       - uses: strip
 
 update:
--- a/tctl.yaml
+++ b/tctl.yaml
@@ -38,7 +38,6 @@ pipeline:
 
       make build
       install -Dm755 tctl ${{targets.destdir}}/usr/local/bin/tctl
-      install -Dm755 tctl-authorization-plugin ${{targets.destdir}}/usr/local/bin/tctl-authorization-plugin
 
   - uses: strip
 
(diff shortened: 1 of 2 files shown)


mritunjaysharma394 (commented):
We will also need epoch bump for both since we are making changes them

tctl.yaml
```diff
@@ -37,11 +37,20 @@ pipeline:
       go mod tidy
 
       make build
-      install -Dm755 tctl ${{targets.destdir}}/usr/bin/tctl
-      install -Dm755 tctl-authorization-plugin ${{targets.destdir}}/usr/bin/tctl-authorization-plugin
+      install -Dm755 tctl ${{targets.destdir}}/usr/local/bin/tctl
+      install -Dm755 tctl-authorization-plugin ${{targets.destdir}}/usr/local/bin/tctl-authorization-plugin
```
mritunjaysharma394, line 41: I think we no longer need this since we are already adding a sub package and installing it there

mritunjaysharma394 (changes requested):

tctl.yaml
shivanshu1333: smh, true

tctl.yaml
```diff
@@ -37,11 +37,19 @@ pipeline:
       go mod tidy
 
       make build
-      install -Dm755 tctl ${{targets.destdir}}/usr/bin/tctl
-      install -Dm755 tctl-authorization-plugin ${{targets.destdir}}/usr/bin/tctl-authorization-plugin
+      install -Dm755 tctl ${{targets.destdir}}/usr/local/bin/tctl
```
rawlingsj, line 40: rather than writing to `${{targets.destdir}}/usr/local` we tent to create a "compat"ability  subpackage with a symlink in the desired location to the actual real location, i.e like this https://github.com/wolfi-dev/os/blob/487aec1ebf9ba475af0b25e5d060e9b084180e5d/kots.yaml#L56-L64

temporal-server.yaml
```diff
@@ -43,30 +43,62 @@ pipeline:
...
   - name: tdbg
     description: "tdbg"
     pipeline:
       - runs: |
-          mkdir -p "${{targets.subpkgdir}}"/usr/bin
+          mkdir -p "${{targets.subpkgdir}}"/usr/local/bin
           install -Dm755 tdbg "${{targets.subpkgdir}}"/usr/bin/tdbg
       - uses: strip
 
+  - name: tdbg
```
mritunjaysharma394, line 78: I personally think, having two sub packages of same names is not feasible, we can add something like compat to them while doing symlinks, @rawlingsj may have more inputs on them? Similar for others

shivanshu1333: having compact makes sense, fixed

mritunjaysharma394: A minor nit, I think it should be compat (short for compatibility). Also I see the scanning of CVE fails, will be good to see if it’s fixable or we have to need an advisory for that, thanks for doing it!

tctl.yaml
```diff
@@ -34,14 +34,33 @@ pipeline:
...
 
+      # CVE-2023-47108
+      # go get go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc@v0.46.0
+
       go mod tidy
 
       make build
       install -Dm755 tctl ${{targets.destdir}}/usr/bin/tctl
-      install -Dm755 tctl-authorization-plugin ${{targets.destdir}}/usr/bin/tctl-authorization-plugin
+      install -Dm755 tctl-authorization-plugin ${{targets.subpkgdir}}/usr/bin/tctl-authorization-plugin
```
rawlingsj, line 44:
```suggestion
      install -Dm755 tctl-authorization-plugin ${{targets.contextdir}}/usr/bin/tctl-authorization-plugin
```
the main package pipeline needs to use `${{targets.destdir}}` or we've recently added `${{targets.contextdir}}` which can be used in both the main package and sub package pipelines.

temporal-server.yaml
```diff
@@ -43,30 +43,62 @@ pipeline:
   - uses: strip
 
 subpackages:
+  - name: temporal-server-compat
+    description: "Compat package for temporal-server"
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/usr/local/bin
+          ln -s /usr/bin/temporal-server "${{targets.subpkgdir}}"/usr/local/bin
```
rawlingsj, line 51:
```suggestion
          ln -s /usr/bin/temporal-server "${{targets. contextdir}}"/usr/local/bin
```

temporal-server.yaml
```diff
@@ -43,30 +43,62 @@ pipeline:
...
+      - uses: strip
+
   - name: temporal-cassandra-tool
     description: "temporal-cassandra-tool"
     pipeline:
       - runs: |
-          mkdir -p "${{targets.subpkgdir}}"/usr/bin
+          mkdir -p "${{targets.subpkgdir}}"/usr/local/bin
```
rawlingsj, line 58:
```suggestion
          mkdir -p "${{targets.subpkgdir}}"/usr/bin
```

rawlingsj (changes requested):

rawlingsj:
Thanks - it's close, just a cpl of small suggested changes

shivanshu1333:
https://github.com/wolfi-dev/advisories/pull/572 is merged
this is unblocked now, ready for review/merge

tctl.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: tctl
   version: 1.18.0
-  epoch: 1
```
developer-guy, line 4: shouldn't be epoch is 2? why we bumped it to `7` directly?

tctl.yaml
```diff
@@ -38,10 +38,26 @@ pipeline:
 
       make build
       install -Dm755 tctl ${{targets.destdir}}/usr/bin/tctl
-      install -Dm755 tctl-authorization-plugin ${{targets.destdir}}/usr/bin/tctl-authorization-plugin
```
developer-guy, line 41: contextdir is a kind of special variable and we use that in melange pipelines because you can't know where the pipeline is going to be run so if it is used in the main package pipelines it would be `targets.destdir` if it is sub package it would be `targets.subpkgdir`, what I'm trying to say is that we mostly avoid using targets.contexdir direclty in the packages. but PTAL @mamccorm

temporal-server.yaml
```diff
@@ -51,22 +59,46 @@ subpackages:
...
+          ln -s /usr/bin/temporal-cassandra-tool "${{targets.subpkgdir}}"/usr/local/bin
+      - uses: strip
+
   - name: tdbg
     description: "tdbg"
     pipeline:
       - runs: |
-          mkdir -p "${{targets.subpkgdir}}"/usr/bin
+          mkdir -p "${{targets.subpkgdir}}"/usr/local/bin
           install -Dm755 tdbg "${{targets.subpkgdir}}"/usr/bin/tdbg
```
developer-guy, line 75: you created a /usr/local/bin in the previous step but you install the package into /usr/bin, this doesn't sound correct, am I right?

tctl.yaml
shivanshu1333: I was testing something locally, fixed it

tctl.yaml
shivanshu1333: @rawlingsj  requested this change here https://github.com/wolfi-dev/os/pull/8897#discussion_r1407972076

rawlingsj: `contextdir` is expected to be used instead of `targets.destdir` and `targets.subpkgdir` - unless there is a specific reason that a pipeline needs to write to another packages output directory.

There's a bit more info https://github.com/chainguard-dev/melange/pull/622

rawlingsj (approved):

tctl.yaml
jonjohnsonjr: I don't think you pushed an update for this?

PR #5191, opened 2023-09-06
php: Add version stream to php 8.2, rename to php-8.2, also now provides php

vaikas:
Signed-off-by: Ville Aikas <vaikas@chainguard.dev>

As I noticed in the #5165 the release-monitor lumps all the releases into the same thing. It's also my understanding from talking with @rawlingsj that the github is the preferred method. Lastly, there's a new 8.3.RC train rolling in, so figured to convert the 8.2 to use github, and provide a version stream while at it.

Tested with (installing both 'existing' php as well as php-8.2 to verify we get the same stuff. Also test with php-8.1 to make sure folks can get that to install as well.

```
vaikas@vaikas-MBP os % make local-wolfi
<SNIP>
1984fcb2cd0d:/work/packages# apk add php
fetch https://packages.wolfi.dev/os/aarch64/APKINDEX.tar.gz
(1/7) Installing xz (5.4.4-r0)
(2/7) Installing libxml2 (2.11.5-r0)
(3/7) Installing ncurses-terminfo-base (6.4_p20230722-r1)
(4/7) Installing ncurses (6.4_p20230722-r1)
(5/7) Installing readline (8.2-r2)
(6/7) Installing sqlite (3.40.0-r1)
(7/7) Installing php-8.2 (8.2.10-r1)
OK: 31 MiB in 21 packages
1984fcb2cd0d:/work/packages# exit
vaikas@vaikas-MBP os % make local-wolfi
<SNIP>
102d527a14af:/work/packages# apk add php-8.2
fetch https://packages.wolfi.dev/os/aarch64/APKINDEX.tar.gz
(1/7) Installing xz (5.4.4-r0)
(2/7) Installing libxml2 (2.11.5-r0)
(3/7) Installing ncurses-terminfo-base (6.4_p20230722-r1)
(4/7) Installing ncurses (6.4_p20230722-r1)
(5/7) Installing readline (8.2-r2)
(6/7) Installing sqlite (3.40.0-r1)
(7/7) Installing php-8.2 (8.2.10-r1)
OK: 31 MiB in 21 packages
102d527a14af:/work/packages# exit
vaikas@vaikas-MBP os % make local-wolfi
<SNIP>
5b08df9863ad:/work/packages# apk add php-8.1
fetch https://packages.wolfi.dev/os/aarch64/APKINDEX.tar.gz
(1/7) Installing xz (5.4.4-r0)
(2/7) Installing libxml2 (2.11.5-r0)
[...]

PR #8984, opened 2023-11-23
php-8.3.0 release version. Do not promote yet though.

vaikas:
PHP-8.3.0 was released, so move from RC to release version. Do not make it the default yet, so one has to explicitly ask for php-8.3

Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/php-8.3.yaml
+++ b/php-8.3.yaml
@@ -1,6 +1,6 @@
 package:
   name: php-8.3
-  version: 8.3_rc5
+  version: 8.3.0
   epoch: 0
   description: "the PHP programming language"
   copyright:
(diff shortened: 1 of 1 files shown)


php-8.3.yaml
```diff
@@ -315,7 +306,10 @@ subpackages:
             echo 'decorate_workers_output = no'; \
           } | tee ${{targets.subpkgdir}}/etc/php/php-fpm.d/zz-apko.conf
 
-# The release candidates are in a different location, so we can't really
-# monitor the github tags for these since they differ.
 update:
-  enabled: false
+  enabled: true
+  github:
```
mamccorm, line 311: Do we not need release monitor here? as we're not pulling from github?

vaikas: Good question, we had set up the other php this way, and I also checked with the wolfictl update. Maybe we should instead move them all to use github to checkout instead of the distributions? 🤔 
I think we want to use github for update over release monitoring, so I think the right thing is to see about pulling the source from github, yeah. Let me see...

vaikas: Interesting, looks like it now wants re2c which we do not currently seem to have? 🤔 
```
⚠️  aarch64   | configure: error: re2c 1.0.3 is required to generate PHP lexers.
```

mamccorm: I guess we meant to switch over to git checkout in this previous commit? https://github.com/wolfi-dev/os/pull/5191/files when we setup git polling?

mamccorm (approved):
Added a comment / thread, but not specific just to this change. I think we've some examples where we should consider migrating to git checkout, especially when we poll github for new release / tags.

I'll let you decide to merge in when ready, in case you are planning on pushing up a change to flip over to git clone

vaikas:
> Added a comment / thread, but not specific just to this change. I think we've some examples where we should consider migrating to git checkout, especially when we poll github for new release / tags.
> 
> I'll let you decide to merge in when ready, in case you are planning on pushing up a change to flip over to git clone

Since the other ones are using this pattern, I'm going to merge this and then I'll tackle the git-checkout in a followup PR.

PR #9267, opened 2023-11-30
Add minio docker entrypoint

denhamparry:
* `docker-entrypoint.sh` is used within minio
* add subpackage to install entrypoint

### DIFF
--- a/minio/docker-entrypoint.sh
+++ b/minio/docker-entrypoint.sh
@@ -1,0 +1,26 @@
+#!/bin/sh
+#
+
+# If command starts with an option, prepend minio.
+if [ "${1}" != "minio" ]; then
+	if [ -n "${1}" ]; then
+		set -- minio "$@"
+	fi
+fi
+
+docker_switch_user() {
+	if [ -n "${MINIO_USERNAME}" ] && [ -n "${MINIO_GROUPNAME}" ]; then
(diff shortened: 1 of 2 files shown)


minio/docker-entrypoint.sh
```diff
@@ -0,0 +1,26 @@
...
+if [ "${1}" != "minio" ]; then
+	if [ -n "${1}" ]; then
+		set -- minio "$@"
+	fi
+fi
+
+docker_switch_user() {
+	if [ -n "${MINIO_USERNAME}" ] && [ -n "${MINIO_GROUPNAME}" ]; then
+		if [ -n "${MINIO_UID}" ] && [ -n "${MINIO_GID}" ]; then
+			chroot --userspec=${MINIO_UID}:${MINIO_GID} / "$@"
```
joshrwolf, line 14: I'm not sure I love the use of `chroot` here, can we not just rely on the container config or the runtime to set the user?

joshrwolf: similarly, does upstream have precedence for how this is done?

denhamparry: We can do, I didn't know what was best practice and thought I'd just drive and see how far I could get 😅 
So instead, I'll focus on implementing this within the image build using the package.
Would that be preferred?

PR #9301, opened 2023-12-01
use SPDX-compliant license for Public Domain

imjasonh:
SPDX doesn't recognize the license identifier `Public Domain` or `Public-Domain`, so SBOMs produced by such packages are considered invalid according to the NTIA conformance checker.

This moves these licenses to https://spdx.org/licenses/CC-PDDC.html instead, which SPDX recognizes.

@kaniini please tell me if this is wrong. 🙏

### DIFF
--- a/dns-root-hints.yaml
+++ b/dns-root-hints.yaml
@@ -4,7 +4,7 @@ package:
   epoch: 0
   description: The DNS root hint(s)
   copyright:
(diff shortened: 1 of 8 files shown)


kaniini:
This seems fine to me.  Let's do it.

kaniini (approved):

PR #9399, opened 2023-12-04
Add py3-docker-squash

tcnghia:
We need to use packaging.version instead of distutils.StrictVersion because the latter is not available on Python 3.12. I am working on upstreaming this change as well, but using `sed` here in the meantime.

* "py3-docker-squash-1.1.0: new package"

Fixes:

Related:

#### For new package PRs only

- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`): No, but this is the latest version.

```
❯ wolfictl scan packages/aarch64/*.apk
Will process: py3-docker-squash-1.1.0-r0.apk

✅ No vulnerabilities found

```

### DIFF
--- a/py3-docker-squash.yaml
+++ b/py3-docker-squash.yaml
@@ -1,0 +1,46 @@
+# Generated from https://pypi.org/project/docker-squash/
+package:
+  name: py3-docker-squash
+  version: 1.1.0
+  epoch: 0
+  description: Docker layer squashing tool
+  copyright:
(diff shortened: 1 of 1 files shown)


priyawadhwa (dismissed):

priyawadhwa (dismissed):

joshrwolf:
needs more yam!

priyawadhwa (approved):

PR #9409, opened 2023-12-05
feat: new package kubeflow-metadata-envoy

stormqueen1990:
Create new package for the envoy configuration from metadata-envoy as specified in https://github.com/kubeflow/pipelines/blob/master/third_party/metadata_envoy/envoy.yaml

#### For new package PRs only


kubeflow-metadata-envoy-config.yaml
```diff
@@ -0,0 +1,24 @@
+package:
+  name: kubeflow-metadata-envoy-config
+  version: 2.0.4 # check if this should be the version
```
rawlingsj, line 3: what would 2.0.4 map to?  I'm wondering if we should start this at 1.0.0 so it has it's own version lifecycle?

kubeflow-metadata-envoy-config.yaml
```diff
@@ -0,0 +1,24 @@
+package:
+  name: kubeflow-metadata-envoy-config
```
rawlingsj, line 2: Just to double check, if this is related to another package in wolfi could we instead add this config as a subpackage of that other wolfi package?  I.e. rather than having a new top level package?

kubeflow-metadata-envoy-config.yaml
stormqueen1990: That's the version of the tag where the file came from in kubeflow/pipelines, but I think it'd be fair to version from 1.0.0.

kubeflow-metadata-envoy-config.yaml
stormqueen1990: We could move it to a subpackage, yes. What did you have in mind?

stormqueen1990: Moved it into the main kubeflow as a subpackage.

kubeflow-metadata-envoy-config.yaml
stormqueen1990: This is addressed by the latest changes.

dlorenc (approved):

PR #9411, opened 2023-12-05
Move 9 more packages to gobump.

dlorenc:


cloud-sql-proxy.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: cloud-sql-proxy
   version: 2.8.0
-  epoch: 0
+  epoch: 1
```
ajayk, line 4: No change to this package , probably not needed to bump

dlorenc:
We can skip the update check here.

ajayk (approved):

PR #9428, opened 2023-12-05
Bump to opentofu beta2.

dlorenc:


rawlingsj:
Looks like this is missing the actual bump to `1.6.0-beta2` and epoch reset

rawlingsj:
ah sorry - just saw @ajayk PR for the same bump https://github.com/wolfi-dev/os/pull/9418, think we can close this one.

PR #9432, opened 2023-12-05
Convert 12 more packages with embedded go gets into go/bump.

dlorenc:
This also adds a -go=1.20 to the pipeline itself. We had a few packages that were pinned to go 1.16 in their go.mods, which isn't updated by default to higher go versions.

### DIFF
--- a/cilium.yaml
+++ b/cilium.yaml
@@ -1,7 +1,7 @@
 package:
   name: cilium
   version: 1.14.4
-  epoch: 1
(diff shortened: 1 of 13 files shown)


dlorenc:
There will be a few CVE failures left in here, but they're not new so I don't think they should block this. I did patch some, but I can't them all in this PR.

joshrwolf (approved):

PR #9460, opened 2023-12-05
Added package brew

dustinkirkland:
Added a brew package.  Notes: This is the linux version of the Homebrew package manager for MacOS.  Not ideal, but upstream brew insists on installing everything in the /home/linuxbrew/.linuxbrew directory.  Everything it needs is kept nice and tidy in there.  FWIW, apk audit --system is clean.


brew.yaml
```diff
@@ -0,0 +1,67 @@
...
+      destination: ./brew
+  - runs: |
+     set -x
+     mkdir -p ${{targets.destdir}}/home/linuxbrew/.linuxbrew/
+     rm -rf ./brew/.git/
+     mv ./brew/ ${{targets.destdir}}/home/linuxbrew/.linuxbrew/
+     cd ${{targets.destdir}}/home/linuxbrew/.linuxbrew/brew/bin/
+     mkdir -p ${{targets.destdir}}/etc/profile.d/
+     echo "export PATH=\$PATH:/home/linuxbrew/.linuxbrew/brew/bin" > ${{targets.destdir}}/etc/profile.d/brew.sh
+     echo "export HOMEBREW_NO_AUTO_UPDATE=1" >> ${{targets.destdir}}/etc/profile.d/brew.sh
```
kaniini, line 54: How will users consume updates to the Homebrew recipes?  Will they be distributed via the APK or...?

dustinkirkland:
Updates to brew will be distributed by Wolfi/apk as brew self updates have
been disabled by environment variable configuration in
/etc/profile.d/brew.sh

All of Brew's work (including formula, packages, builds, cache, etc) lands
under /home/linuxbrew/.linuxbrew/.  I don't love that it's not FHS, but it
is super well contained and tidy.  And that does include updates to formula
within that dir too.

DK

kaniini:
It might be useful to note that in the post-install message so users don't run "brew update" to fetch the updates.  Or perhaps we can patch "brew update" to display a message?

dustinkirkland:
Hmm, so the HOMEBREW_NO_AUTO_UPDATE=1 environment variable just short-circuits and avoids brew's automatic update mechanism every time you run the command.

Running "brew update", though, does go out to github.com/Homebrew/brew and does a complete checkout of the code and overwrites itself in /home/linuxbrew/.linuxbrew/, if someone goes out of their way to run "brew update".  Presumably, if the user is running "brew update", they want brew to update itself.

Alternatively, if the user runs "apk upgrade", then they would get a newer brew package from apk.

kaniini:
I guess what I am coming from is that if a user does "brew update" and then they do "apk upgrade" there is a risk that the user will wind up unable to do "apk upgrade" successfully because it errors out due to modified files that are supposed to be owned by the brew package.

So we should try to at least document that running "brew update" outside the package manager is done at the user's own risk.

dustinkirkland:
Okay, so pushed a couple more changes...Bumped the version to match latest upstream.  Fixed a missing shebang in the post-deinstall.  I've tested an 'apk del brew', which now completely removes all of /home/linuxbrew.  I've added a WARNING message that gets printed in the post-install, that running 'brew update' will probably not play nicely with 'apk upgrade brew'.  I've actually tested that, and here's what happens...

brew.yaml
dustinkirkland: I've added a WARNING message that gets printed in the post-install, that running 'brew update' will probably not play nicely with 'apk upgrade brew'.
I've actually tested that (apk install brew ; brew update ; apk upgrade brew), and here's what happens... The brew upgrade goes out to github, and git pulls the latest code into /home/linuxbrew/.linuxbrew/brew. Which runs fine. Then, apk upgrade overwrites the code in the git repo there (which, conveniently shows the changes with a git diff). brew -v actually notices this, and adds a "-dirty" on the end of the version which is kind of nice. If the user wants to get back to the brew version from git, it's just git rebase commands. Otherwise, the user is running the new apk version.

vaikas:
I think we need to do some trimming of the files included in the package overall, and maybe split into multiple subpackages. I don't know enough of the brew (except just install/uninstall, so some of these may not make sense?):

For example, looks like the `.github` dir gets added into the package, and I'm pretty sure these don't belong in the package:
```
a1b8bd91f879:/work/packages# apk info -L brew | grep \.github
home/linuxbrew/.linuxbrew/brew/.github/ISSUE_TEMPLATE.md
home/linuxbrew/.linuxbrew/brew/.github/PULL_REQUEST_TEMPLATE.md
home/linuxbrew/.linuxbrew/brew/.github/actionlint-matcher.json
home/linuxbrew/.linuxbrew/brew/.github/actionlint.yaml
home/linuxbrew/.linuxbrew/brew/.github/codecov.yml
home/linuxbrew/.linuxbrew/brew/.github/dependabot.yml
<SNIPPITY SNIP SNIP>
```

I think there might also be a case for having a subpackage brew-dev, at least looking at the files/quick read through the files on what they do, seems like they are not necessary for the `brew` user.
Some examples:
```
<SNIP>
home/linuxbrew/.linuxbrew/brew/Library/Homebrew/dev-cmd/update-license-data.rb
home/linuxbrew/.linuxbrew/brew/Library/Homebrew/dev-cmd/update-maintainers.rb
<SNIP>
```

Also, not sure if the tests should be added to the package, unless they are used to test incoming packages?
```
<SNIP>
home/linuxbrew/.linuxbrew/brew/Library/Homebrew/test/unpack_strategy/cvs_spec.rb
home/linuxbrew/.linuxbrew/brew/Library/Homebrew/test/unpack_strategy/directory_spec.rb
home/linuxbrew/.linuxbrew/brew/Library/Homebrew/test/unpack_strategy/dmg_spec.rb
home/linuxbrew/.linuxbrew/brew/Library/Homebrew/test/unpack_strategy/git_spec.rb
home/linuxbrew/.linuxbrew/brew/Library/Homebrew/test/unpack_strategy/gzip_spec.rb
<SNIP>
```

Maybe the docs should be removed, some examples?
```
home/linuxbrew/.linuxbrew/brew/docs/.mdl_ruleset.rb
home/linuxbrew/.linuxbrew/brew/docs/.mdl_style.rb
home/linuxbrew/.linuxbrew/brew/docs/.mdlrc
home/linuxbrew/.linuxbrew/brew/docs/Acceptable-Casks.md
home/linuxbrew/.linuxbrew/brew/docs/Acceptable-Formulae.md
home/linuxbrew/.linuxbrew/brew/docs/Adding-Software-to-Homebrew.md
home/linuxbrew/.linuxbrew/brew/docs/Analytics.md
home/linuxbrew/.linuxbrew/brew/docs/Bottles.md
home/linuxbrew/.linuxbrew/brew/docs/governance/2019-membership.md
home/linuxbrew/.linuxbrew/brew/docs/governance/2019-moss-track-iii-grant-nomination.md
home/linuxbrew/.linuxbrew/brew/docs/governance/2019-plc-minutes.md
home/linuxbrew/.linuxbrew/brew/docs/governance/2020-membership.md
```

brew.yaml
```diff
@@ -0,0 +1,66 @@
...
+    with:
+      repository: https://github.com/Homebrew/brew
+      tag: ${{package.version}}
+      destination: ./brew
+      expected-commit: c32bd1c7cc0c3b1b914845bddfeda53f4d877a3f
+
+  - runs: |
+      set -x
+      mkdir -p ${{targets.destdir}}/home/linuxbrew/.linuxbrew/
+      rm -rf ./brew/.git/
```
vaikas, line 56: Yeah, so maybe remove a few more directories here as per the comment on the PR.

dustinkirkland:
Happy to remove the .github.  That's easy, and would match removing the
.git.  Happy to remove the docs and the test too, in the interest of
keeping it minimal.

Note, though, as explained above (and in the message echo'd to screen) --
if the brew user ever runs "brew update", brew is going to reach out to git
and repopulate ALL of that by cloning the upstream git repo.  Perhaps
that's fine, if that's what the user asks brew for...

dustinkirkland:
Okay, @vaikas -- I've minimized the brew package down to what I think it's the minimum set of files to still be functional -- basically just the bin/ Library/ and package/ directories are now installed in /home/linuxbrew/.linuxbrew/brew.

I also split out a separate package for brew-doc, with most of the other stuff.  Note that I couldn't find a manpage reader in wolfi, but I verify that the manpage is installed in the right location.

How's this looking?

lephyrus:
I've found this through reading up on uBlue/bluefin, and wanting to base a devcontainer on Wolfi and brew. Unfortunately, Homebrew ends up in slightly the wrong place: the "root" should be `/home/linuxbrew/.linuxbrew`, not ` /home/linuxbrew/.linuxbrew/brew`. [Anything but that default prefix is not supported](https://docs.brew.sh/Homebrew-on-Linux#install). While I could install some formulas without a problem, I found the issue when the `openssl@3` install complained, did a fall back to building from source and eventually failed.

dustinkirkland:
Aha!  Great catch!  I pushed up a fix to my branch.  Try that out and see if it works for you...

dlorenc (approved):

PR #9544, opened 2023-12-06
Full source bootstrap for rust 1.29 🎉 

Elizafox:
This is the first step in a full bootstrap chain for Rust 1.29 through 1.74. This uses mrustc to do the initial bootstrap from pure source. Additional bootstrapping work will be followed up in future PR's.

#### For new package PRs only
- [ ] This PR is marked as fixing a pre-existing package request bug (part of chainguard-dev/internal#3355)
- [ ] REQUIRED - The version of the package is still receiving security updates (**EXCEPTION**: bootstrapping)


Elizafox:
Here is a screenshot of it working:

Screenshot 2023-12-06 at 4 09 09 PM

kaniini:
It seems that the job was reaped on both hosts, which indicates that perhaps the runner's resources exceeded what they were allowed.  Maybe we can use a smaller parallelism than nproc?

rust-1.29.yaml
```diff
@@ -0,0 +1,265 @@
...
+      export CPPFLAGS="${{vars.CPPFLAGS}}"
+      export CXXFLAGS="${{vars.CXXFLAGS}}"
+
+      export OPENSSL_NO_VENDOR=1
+      export RUST_BACKTRACE=1
+      DESTDIR=${{targets.destdir}} python3 ./x.py install --jobs $(nproc)
+
+  - uses: strip
+
+  # delete uneeded files eg uninstalltion
```
kaniini, line 226:
```suggestion
  # delete uneeded files such as uninstallation scripts
```

rust-1.29.yaml
```diff
@@ -0,0 +1,265 @@
...
+      # Remove vendored LLVM stored in /home/build, not needed as rustc is statically linked
+      rm -r ${{targets.destdir}}/home
+
+      rm ${{targets.destdir}}/usr/lib/rustlib/components
+      rm ${{targets.destdir}}/usr/lib/rustlib/install.log
+      rm ${{targets.destdir}}/usr/lib/rustlib/rust-installer-version
+      rm ${{targets.destdir}}/usr/lib/rustlib/uninstall.sh
+      rm ${{targets.destdir}}/usr/lib/rustlib/manifest-*
+
+  # rustbuild always installs copies of the shared libraries to /usr/lib,
```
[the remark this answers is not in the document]
Elizafox: These files are actually generated by Rust itself and this rm does not fail on ARM either. These should probably be stripped out. I took this from the existing Rust build file, fwiw.

rust-1.29.yaml
Elizafox: This is also from the original Rust build. Looks good to me.

Elizafox:
Okay I did an update and then it failed. Maybe it's when the runners are busy?

Elizafox:
The CI passes now! This can be merged now. /CC @kaniini

kaniini (dismissed):

Elizafox:
This will be superseded.

PR #9475, opened 2023-12-05
batch 1: go epoch bump

ajayk:

PR #9696, opened 2023-12-09
DO NOT MERGE: this is to test cache performance.

dlorenc:
This is a commit that replicates https://github.com/wolfi-dev/os/pull/9475

That build took 1hr and 15m on amd64 and 1hr and 12m on arm.

### DIFF
--- a/aactl.yaml
+++ b/aactl.yaml
@@ -1,7 +1,7 @@
 package:
   name: aactl
   version: 0.4.12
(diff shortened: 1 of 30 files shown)


dlorenc:
BOOM - 40m and 9s on amd64! that's 50% faster!

dlorenc:
And 45m on arm!

PR #9794, opened 2023-12-12
fix epoch in tctl

shivanshuraj1333:
fixes epoch number

### DIFF
--- a/tctl.yaml
+++ b/tctl.yaml
@@ -1,7 +1,7 @@
 package:
   name: tctl
   version: 1.18.0
-  epoch: 9
+  epoch: 1
   description: Temporal CLI
   copyright:
     - license: MIT License


tctl.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: tctl
   version: 1.18.0
-  epoch: 9
+  epoch: 1
```
kaniini, line 4: epoch cannot go backwards, e.g. 9 to 1

kaniini (changes requested):

tctl.yaml
shivanshu1333: yeah, but somebody merged the PR with epoch 9 by mistake before 9 could be reverted back to 1, should we leave it at 9 now?

mritunjaysharma394: Yeah, I think it went unnoticed here, I think reviewers had left thought but somehow it got merged. https://github.com/wolfi-dev/os/pull/8897/files#r1410816853

ajayk: this is not needed anymore lets close this

mamccorm:
[ajayk](https://github.com/ajayk) [4 days ago](https://github.com/wolfi-dev/os/pull/9794#discussion_r1427524687)
this is not needed anymore lets close this

PR #9872, opened 2023-12-13
Fix incorrect 'compat' package name

mamccorm:
Recent commit mis-spelt 'compat' as 'compact' for two new packages: https://github.com/wolfi-dev/os/pull/8897. These are not yet used elsewhere. This PR is to correct that


shivanshu1333:
created this https://github.com/wolfi-dev/os/pull/9877 and increased epoch, this would not break anything and fix the typo.

PR #10129, opened 2023-12-19
mitigate CVE-2023-48795 for cloud-sql-proxy

cpanato:
- mitigate CVE-2023-48795 for cloud-sql-proxy

Closes: https://github.com/wolfi-dev/os/pull/9901

#### For security-related PRs

- [x] The security fix is recorded in the [advisories](https://github.com/wolfi-dev/advisories) repo: https://github.com/wolfi-dev/advisories/pull/633

### DIFF
--- a/cloud-sql-proxy.yaml
+++ b/cloud-sql-proxy.yaml
@@ -1,7 +1,7 @@
 package:
   name: cloud-sql-proxy
   version: 2.8.1
-  epoch: 1
+  epoch: 2
   description: The Cloud SQL Auth Proxy is a utility for ensuring secure connections to your Cloud SQL instances
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


hectorj2f (approved):

cloud-sql-proxy.yaml
```diff
@@ -23,6 +23,11 @@ pipeline:
       tag: v${{package.version}}
       expected-commit: 7cc6c8ebbc96de76e9962ba2a77aaf6534122774
 
+  - uses: go/bump
+    with:
+      deps: golang.org/x/crypto@v0.17.0
+      go-version: "1.21"
```
hectorj2f, line 29: is it just me or having to specify the go-version here is not ideal ? One more think to update in the future, how do we know the version to go here ?

hectorj2f: Just a general comment, it is not related to this PR.

pdeslaur: I agree - it's not ideal - I wonder if we could default to the version in the `go.mod` file instead?

pdeslaur (approved):

cloud-sql-proxy.yaml
hectorj2f: ` I wonder if we could default to the version in the go.mod file instead?`

Indeed 🤩, that is how it should work. I wasn't sure why we have to set this property value 🤔.

cpanato: the go/bump defaults to 1.20 but when we use the apk go package I think that is the latest in this case 1.21, so we need to make explicit here that is using go1.21

i can be totally wrong as well

PR #10142, opened 2023-12-19
Create a chainguard-keys package

dustinkirkland:
Let's offer a "chainguard-keys" package, which adds Chainguard's public signing keys to /etc/apk/keys.  "apk add chainguard-keys" is way better than sloppy, possibly error prone wgets of .rsa.pub files from URLs (which can be typo'd or suffer lookalike domain attacks).  It also gives us a clean way to update those keys, or ship additional keys over time (via apk update / apk upgrade).

Note that this does NOT add the repos to /etc/apk/repositories.  Just puts the keys conveniently in the keyring.


chainguard-keys.yaml
```diff
@@ -0,0 +1,28 @@
+package:
+  name: chainguard-keys
+  version: 1
+  epoch: 5
```
joshrwolf, line 4:
```suggestion
  epoch: 0
```

chainguard-keys.yaml
```diff
@@ -0,0 +1,28 @@
...
+    runs: |
+      mkdir -p "${{targets.destdir}}"/etc/apk/keys
+      install -m644 chainguard-extras.rsa.pub "${{targets.destdir}}"/etc/apk/keys
+      install -m644 chainguard-enterprise.rsa.pub "${{targets.destdir}}"/etc/apk/keys
+
+      for i in x86_64 aarch64; do
+        mkdir -p "${{targets.destdir}}"/usr/share/apk/keys/${i}
+        install -m644 chainguard-extras.rsa.pub "${{targets.destdir}}"/usr/share/apk/keys/${i}/chainguard-extras.rsa.pub
+        install -m644 chainguard-enterprise.rsa.pub "${{targets.destdir}}"/usr/share/apk/keys/${i}/chainguard-enterprise.rsa.pub
+      done
```
mattmoor, line 25: I believe apko also sets up `/etc/apk/keys`, so I would be curious which one "wins" in these cases.  cc @jonjohnsonjr

mattmoor: Ack this is based on `wolfi-keys`, which would already have this problem (if it's a problems).

dustinkirkland: Yeah, I literally cloned wolfi-keys.yaml, and assumed that it does the right thing...

chainguard-keys.yaml
dustinkirkland: Doh.  Pushed fixes.

chainguard-keys.yaml
jonjohnsonjr: I _think_ this should be fine but I can dig in if something goes wonky.

tuananh:
is there a mechanism to prevent maintainer from forgetting to update this pkg when rorating chainguard keys?

rawlingsj:
> is there a mechanism to prevent maintainer from forgetting to update this pkg when rorating chainguard keys?

There's not an automated way ATM but would be something we'd need to do.  I think we can proceed with this PR.

rawlingsj (approved):

PR #10184, opened 2023-12-20
go-1.22: add go 1.22 rc1

ajayk:
Adds go-1.22 rc1 version stream

### DIFF
--- a/go-1.22.yaml
+++ b/go-1.22.yaml
@@ -1,0 +1,94 @@
+package:
+  name: go-1.22
+  version: 1.22_rc1
+  epoch: 0
+  description: "the Go programming language"
+  copyright:
+    - license: BSD-3-Clause
+  dependencies:
+    #provides:
+    # - go=${{package.full-version}}
+    # - go-bootstrap=1.22
+    runtime:
+      - bash
+      - binutils-gold # Needed for cgo linking due to upstream issue #15696 which forces use of the gold linker.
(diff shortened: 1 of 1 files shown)


go-1.22.yaml
```diff
@@ -0,0 +1,94 @@
...
+subpackages:
+  - name: "go-1.22-doc"
+    description: "go documentation"
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/usr/share
+          mv "${{targets.destdir}}"/usr/share/doc "${{targets.subpkgdir}}"/usr/share/
+
+update:
+  enabled: true
```
rawlingsj, line 88: release candidates (versions with with rc) are automatically removed from github results so the automated wolfi bot check fails.  We may want to disable auto updates for this RC to avoid having an automated bot issue created.

imjasonh: Disabling sgtm, they announce them and we don't need these out super fast.

go-1.22.yaml
```diff
@@ -0,0 +1,94 @@
...
+  shared: true
+  github:
+    identifier: golang/go
+    strip-prefix: go
+    tag-filter: go1.22
+    use-tag: true
```
imjasonh, line 94:
```suggestion
update:
  enabled: false
```

imjasonh (approved):

PR #10426, opened 2023-12-27
Add fluxbox package

mamccorm:
Adds a 'fluxbox' package which is required by some other images, including docker-selenium.

Note, fluxbox does not actively have any new releases or tags cut, but still has active contributions, hence the comments added in the melange file

### DIFF
--- a/fluxbox.yaml
+++ b/fluxbox.yaml
@@ -14,11 +14,18 @@ environment:
       - build-base
       - busybox
       - ca-certificates-bundle
+      - freetype-dev
+      - fribidi-dev
       - gettext-dev
       - imlib2-dev
       - libtool
       - libx11-dev
       - libxext-dev
+      - libxft-dev
+      - libxinerama-dev
+      - libxpm-dev
+      - libxrandr-dev
+      - libxrender-dev
       - pkgconf-dev
 
 pipeline:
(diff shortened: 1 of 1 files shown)


fluxbox.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - automake
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - gettext-dev
+      - libtool
+      - libx11-dev
+      - libxext-dev
+      - pkgconf-dev
+
```
Dentrax, line 22: https://git.alpinelinux.org/aports/tree/community/fluxbox/APKBUILD#n11

AFAICS, `imlib2` also required which we need to build `libid3tag` first. I was about to get into it 🙈

mamccorm: Sure, feel free to use this as reference material in that case (marked as draft for now). there where no complication errors, is imlib2 needed as a runtime dependency?

Dentrax: I think its required as build-dependency but since pipeline is green, we can say its not a hard-requirement. But it would be difficult to predict the runtime behaviour without having this dep, but i will package `imlib2` anyway since will be needed by other deps. I got this 👍

dlorenc (approved):

PR #10484, opened 2023-12-28
php-8.3: enable provides

ajayk:


php-8.3.yaml
```diff
@@ -1,17 +1,16 @@
...
+  epoch: 1
   description: "the PHP programming language"
   copyright:
     - license: PHP-3.01
   dependencies:
-    # DO NOT USE PROVIDES UNTIL A RELEASE HAS BEEN MADE (not a release candidate)
-    #provides:
-    #  - php=${{package.full-version}}
+    provides:
+      - php=${{package.full-version}}
```
mattmoor, line 10: Are we sure downstream packages won't be broken by this?  Do we need to do any rebuilds?

ajayk: currently there are no packages depending on php-8.3 yet

vaikas:
So, this will mean that anybody asking for php will get it, is that something that we want to do right now, or put another way, will it break something?
Also, There are other `provides` in this file (config, other subpackages) that will also need to be uncommented for consistency. I sure hope I remembered to label them all 🤔

php-8.3.yaml
mattmoor: As Ville says below.  It doesn't matter if they are depending on `php-8.3`, if they depend on `php` this will mean they get it.

ajayk: Thanks @mattmoor the only package using `php` virtual  is composer , and  I checked the [official composer](https://hub.docker.com/_/composer) docker image   it seems to be bundled with 8.3.1 

```docker run --rm -it \
  composer:lts "-vvv about"
Failed to initialize global composer: Composer could not find the config file: /tmp/composer.json
To initialize a project, please create a composer.json file. See https://getcomposer.org/basic-usage
Running 2.2.22 (2023-09-29 10:53:45) with PHP 8.3.1 on Linux / 6.5.11-linuxkit
```

ajayk: moving this to draft for now

PR #10490, opened 2023-12-29
elixir-1.16: new version stream

ajayk:


elixir-1.16.yaml
```diff
@@ -0,0 +1,42 @@
...
+  version: 1.16.0
+  epoch: 0
+  description: General-purpose programming language and runtime environment
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - erlang
+    provides:
+      - elixir=${{package.full-version}}
```
rawlingsj, line 12: I wonder if we should have an extra 
```suggestion
    provides:
      - elixir=${{package.full-version}}
      - elixir
```
So that we don't break existing consumers of elixir?  Then when a new version stream is created we'd need to move it from here and onto the new malange yaml?

rawlingsj (approved):

PR #10495, opened 2023-12-29
GHSA-45x7-px36-x8w8: fix CVE for Wolfi package crossplane

hectorj2f:
Fix crossplane CVE!


hectorj2f:
This is a good example of why scanning using the remote package might omit cve fixes. I'll re-create the PR using `wolfi-bump --build-package=true` so we re-build it from the sources and generate the respective PR.

PR #10608, opened 2024-01-03
fix: bunx broken symlink in Bun Package

debasishbsws:
there is a typo.
Ref: https://github.com/oven-sh/bun/blob/main/dockerhub/distroless/Dockerfile#L70


rawlingsj (approved):

rawlingsj (approved):
Great catch!

PR #10715, opened 2024-01-04
Remove stale workflows, drop privileges.

mattmoor:
This removes two workflows that we no longer use.

This drops the privileges on three workflows that have switched to using PATs.


imjasonh (approved):

.github/workflows/digestabot.yaml
```diff
@@ -11,12 +11,8 @@ jobs:
     runs-on: ubuntu-latest
     if: github.repository == 'wolfi-dev/os'
 
-
     permissions:
-      contents: write
-      pull-requests: write
-      id-token: write
```
mattmoor, line 18: we need this

PR #10772, opened 2024-01-05
gh: fix path issues for gh and add tests

ajayk:
gh package had a wrong prefix in the install 

```
# apk info -L gh 
WARNING: opening ./../../packages: No such file or directory
gh-2.40.1-r0 contains:
usr/bin/bin/gh
usr/bin/share/bash-completion/completions/gh
usr/bin/share/fish/vendor_completions.d/gh.fish
usr/bin/share/man/man1/gh-alias-delete.1

```

### DIFF
--- a/gh.yaml
+++ b/gh.yaml
@@ -1,7 +1,7 @@
 package:
   name: gh
   version: 2.40.1
-  epoch: 0
+  epoch: 1
   description: GitHub's official command line tool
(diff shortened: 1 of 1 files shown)


k4leung4 (approved):

imjasonh (commented):
Thank you!

PR #10803, opened 2024-01-07
imagemagick: fix update config block as new releases are being ignore…

rawlingsj:
…d by the bot


rawlingsj:
No epoch bump is required as it's just the update config that's changing.

The wolfii update check is expected to fail as there is a new version available.  This change should mean the bot finds the latest version and create a followup automated PR.

imjasonh (approved):

PR #10804, opened 2024-01-07
soname CI check: split into it's own job so it's easier to see when t…

rawlingsj:
…he check fails + we continue to run other jobs like scan.

__Once we're ready to merge we should make the soname check a GitHub required check__ so we don't accidentally merge PRs.  The failed check should still require admin override due to the importance of ABI compatability.

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -101,11 +101,6 @@ jobs:
             echo "exists=false" >> $GITHUB_OUTPUT
           fi
 
-      - name: Check sonames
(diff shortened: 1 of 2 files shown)


rawlingsj:
Screenshot 2024-01-07 at 13 41 37

Using the hello-wolfi package as a canary you can see the check has been split out into its own job and runs corerctly.

rawlingsj:
The new check is now a required check - thanks @imjasonh.  This is ready for review / approve / merge whenever now.

imjasonh (approved):

PR #10821, opened 2024-01-08
buck2: update buck2 and enable updates

ajayk:


buck2.yaml
```diff
@@ -1,8 +1,8 @@
 package:
   name: buck2
   # When bumping, bump the tag below
-  version: 0.0_git20230815
```
rawlingsj, line 4: I'm not sure of the reasons so just checking we want to drop the `0.0_git` version prefix here?

rawlingsj: It might be fine but it could have an impact on the solver getting the correct latest version?

ajayk: solver should be fine it should probably remove to the highest version

rawlingsj (approved):

PR #10854, opened 2024-01-08
mitigate CVE `GHSA-7ww5-4wqc-m92c` `GHSA-jq35-85cj-fj4p` `GHSA-2c7c-3mj9-8fqh`

debasishbsws:
& also `GHSA-8pgv-569h-w5rw`  `GHSA-45x7-px36-x8w8` `GHSA-m425-mq94-257g` in kubescape

#### For PRs that add patches

### DIFF
--- a/kubescape.yaml
+++ b/kubescape.yaml
@@ -15,7 +15,6 @@ environment:
       - cmake
       - git
       - go
-      - gobump
       - python3
 
 pipeline:
--- a/kubescape.yaml
+++ b/kubescape.yaml
@@ -1,7 +1,7 @@
 package:
   name: kubescape
   version: 3.0.3
-  epoch: 0
+  epoch: 1
   description: "Kubescape is an open-source Kubernetes security platform for your IDE, CI/CD pipelines, and clusters. It includes risk analysis, security, compliance, and misconfiguration scanning, saving Kubernetes users and administrators precious time, effort, and resources."
   copyright:
     - license: Apache-2.0 AND MIT
@@ -15,6 +15,7 @@ environment:
       - cmake
       - git
       - go
+      - gobump
       - python3
 
 pipeline:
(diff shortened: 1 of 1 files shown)


gobump/temp-multiline.patch
rawlingsj, line 1: It's probably best to send a PR for this directly to the gobump repo https://github.com/chainguard-dev/gobump if you don't mind.  Also @hectorj2f FYI as I think this kind of thing might be on your radar in case you have any extra feedback before the PR is created on gobump?

rawlingsj:
@debasishbsws it looks like this will fail the yam formatting.  We do plan to have better support for multiline support directly in melange but wont land for a few more weeks.  For now it's probably best to pop them on one line until we have that.  I know it's not great to read but it's definitely something that will be fixed soon + then the formatting check will pass then too.

gobump/temp-multiline.patch
rawlingsj: On second thoughts, as the formatting check failed it might be best to wait for the proper multi line support in melange to land.

debasishbsws: Actually, I am trying it out right now, and look if anything breaks, like `yam` doesn't allow multiline string.
also I will make a conversation with @hectorj2f regarding this.

kubescape.yaml
```diff
@@ -15,6 +15,7 @@ environment:
       - cmake
       - git
       - go
+      - gobump
```
rawlingsj, line 18: fwiw we don't need to explicitly add gobump here because the bobump pipeline referenced below will include it https://github.com/wolfi-dev/os/blob/cafd7ffd4451b7f0f6f7e0450538691fdbd38813/pipelines/go/bump.yaml#L7

Not a big issue, just figured it best to mention.

debasishbsws: Sorry, my bad forgot to remove it.

rawlingsj (approved):

PR #10876, opened 2024-01-09
icu: bump downstream deps batch 2

rawlingsj:


rawlingsj:
Build too 3 hours 11 mins, when we're ready let's skip the merge queue

PR #10893, opened 2024-01-09
cassandra-reaper

developer-guy:
Fixes:

Related: https://github.com/wolfi-dev/advisories/pull/807

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/cassandra-reaper/fix.patch
+++ b/cassandra-reaper/fix.patch
@@ -1,0 +1,89 @@
+# disables some css which sb-admin-2 and sb-admin-2.css and execution of npm run minimize:
+# details are included in: https://github.com/BlackrockDigital/startbootstrap/issues/434
+
+# The frontend build here includes a dependency that doesn't exist anymore on the internet, 
+# Thankfully it's just some js and css, this step disables that part. We should check to see if this is safe to remove after the next release, 
+# the project will have to change the library here significantly.
+
+diff --git a/src/server/pom.xml b/src/server/pom.xml
+index 2962017e..2f0e80eb 100644
+--- a/src/server/pom.xml
++++ b/src/server/pom.xml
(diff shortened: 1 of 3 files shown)


cassandra-reaper/fix.patch
```diff
@@ -0,0 +1,82 @@
+diff --git a/src/server/pom.xml b/src/server/pom.xml
```
imjasonh, line 1: Can you explain these diffs so we can tell when/if they're not necessary anymore?

When there's a new upstream release of this we'll likely have to reapply these diffs and it'll be helpful to understand how and why.

luhring: Sorry if this is me being dumb, but I'm still not sure I understand what's going on with this patch. I see that it's removing dependencies and a build step, and I'm guessing this is to get the build to exit 0.

But what effect does this have on the package? And to Jason's point, how will we know to remove this patch?

developer-guy: @dlorenc can help me to explain it better why we need that

dlorenc: The frontend build here includes a dependency that doesn't exist anymore on the internet, Thankfully it's just some js and css, this step disables that part.

We should check to see if this is safe to remove after the next release, the project will have to change the library here significantly.

luhring: That makes sense, thanks.

@developer-guy can we say that in the patch file? I think the goal here is to leave good enough notes behind that the next person who comes along and looks at this knows what to do. If we can say that this might not be needed after version X, that'd help, so we don't require ourselves to dig up this PR again.

luhring (approved):

PR #11028, opened 2024-01-11
0.14 version stream for memcached-exporter

joshrwolf:
convert memcached-exporter to a `0.14` version stream (the current latest)


memcached-exporter.yaml
rawlingsj, line 1: I think this is missing the renamed memcached-exporter-0.14.yaml?

PR #11037, opened 2024-01-11
new package: scap-security-guide

joshrwolf:
new package, used for references for scanning with `openscap`


joshrwolf:
linter is WARNING on documentation existing in the main package, but I'm leaving this as is since this package is strictly `/usr/man` material used by tools like `openscap`.

this matches other distributions as well.

stormqueen1990 (approved):

PR #11079, opened 2024-01-12
GHSA-8pgv-569h-w5rw: fix CVE for Wolfi package temporal-server

chainguardian:
GHSA-8pgv-569h-w5rw: fix CVE for Wolfi package temporal-server

### DIFF
--- a/temporal-server.yaml
+++ b/temporal-server.yaml
@@ -30,7 +30,7 @@ pipeline:
 
   - uses: go/bump
     with:
-      deps: golang.org/x/net@v0.17.0 google.golang.org/grpc@v1.58.3 golang.org/x/crypto@v0.17.0
+      deps: google.golang.org/grpc@v1.58.3 golang.org/x/crypto@v0.17.0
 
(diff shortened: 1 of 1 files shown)


hectorj2f:
It looks like one issue with the Otel dependency.

hectorj2f:
We cannot bump otel metrics here because there is a strict dependency of the source code to a specific version, as seen below:

```
common/metrics/opentelemetry_provider.go:78:18: cannot use aggregation.ExplicitBucketHistogram{…} (value of type aggregation.ExplicitBucketHistogram) as "go.opentelemetry.io/otel/sdk/metric".Aggregation value in struct literal: aggregation.ExplicitBucketHistogram does not implement "go.opentelemetry.io/otel/sdk/metric".Aggregation (missing method copy)
		have Copy() aggregation.Aggregation
		want copy() "go.opentelemetry.io/otel/sdk/metric".Aggregation
```

rawlingsj (approved):

PR #11205, opened 2024-01-16
Add more comprehensive recursive support

vaikas:
Add support for the planned recursive repo. Wires through source-dir so that things like patches
get properly processed.

This is broken into two commits, one to demonstrate changes to the Makefile, and the second that shows how
moving a trino to subdirectory will be processed correctly with the patches applied.

Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -24,7 +24,10 @@ jobs:
         id: changes
         uses: tj-actions/changed-files@716b1e13042866565e00e85fd4ec490e186c4a2f #tj-actions/changed-files@v41.0.1
         with:
-          files: ./*.yaml
+          files_yaml: |
+            melange:
+              - ./*.yaml # Only top level files without structure
+              - ./*/*/*.melange.yaml # Support recursive melange files with the new naming convention.
 
       - name: "Install wolfictl onto PATH"
         run: |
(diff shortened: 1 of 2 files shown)


rawlingsj (dismissed):

vaikas:
Hm, I need to see why the package was not triggered for rebuild. It builds fine locally, so for some reason it's not being picked up as a diff.
/hold

rawlingsj:
I think the https://github.com/wolfi-dev/os/blob/ef625450791667f327469b9333c7a34fbaf942dc/.github/workflows/ci-build.yaml#L27

i.e. wonder if something like
```
**/*.yaml
````
works?  

I'm not sure if the other actions that are in this repo also need to be updated to handle yaml files in subfolders, my guess is they wont work OOTB yet.  

Happy to help here BTW.

fwiw I always wondered if melange files should be foo.melange.yaml so that we could use
```
**/*melange.yaml 
```
then we’d ignore any yaml files that get included in the source dirs

though renaming files might be a PITA to do at the same time as all this

vaikas:
So, it now works, there were couple of other things that needed changing, but great pointer @rawlingsj thank you 🙇 

So, you can see where I moved the trino to db/trino, and it gets successfully built, as well as appropriate patches applied from the root directory:
https://github.com/wolfi-dev/os/actions/runs/7548716090/job/20551352338?pr=11205#step:6:792

And successful build here:
https://github.com/wolfi-dev/os/actions/runs/7548716090/job/20551352338?pr=11205#step:6:5377

However, the CVE scan is failing, which I do not want to block the PR on, and that will be done in a separate PR.
https://github.com/wolfi-dev/os/actions/runs/7548716090/job/20551352338?pr=11205#step:6:5377

dlorenc (approved):

PR #11233, opened 2024-01-16
apache-arrow/15.0.0 package update

debasishbsws:
Fixes: #11154 and #11152

Apparently `grpc=1.60.0-r2` needs the older version of `protoc` and `protobuf`.
After installing both protoc, when we try to install grpc it is just downgrading the library versions of protoc and protobuf and that's why we are getting the error from protoc
```
/usr/bin/protoc: error while loading shared libraries: libprotoc.so.25.2.0: cannot open shared object file: No such file or directory
```

- Bump up the `grpc` epoch to build it again with the newer version of `protoc` and `protobuf` libs
- Remove init submodule section as we are not using any third-party modules to build it.

#### For version bump PRs


ajayk (dismissed):

debasishbsws:
The issue should have been fixed by #11325.

PR #11296, opened 2024-01-17
ci: Always log packages.log and diff.log

jonjohnsonjr:
These only show up in GitHub PR comments when opened by the robots, but I always want to see this information. I don't know how to fix the permission situation, so as a compromise, let's at least make them visible in the CI logs.


imjasonh:
> I don't know how to fix the permission situation

https://github.com/wolfi-dev/os/pull/7659

imjasonh (approved):

PR #11373, opened 2024-01-18
ruby-3.2: update to ruby 3.2.3 and remove fixed patches

ajayk:


cpanato (approved):

imjasonh (commented):
Should we switch to GH releases, which we can filter? https://github.com/ruby/ruby/tags

PR #11382, opened 2024-01-18
new package: statsd

imjasonh:
Based on https://github.com/statsd/statsd/blob/master/Dockerfile and browsing around the [statsd docker image](https://oci.dag.dev/size/statsd/statsd@sha256:4d4bfc2925dff838caa998d7fa764135c2322237671ee1793ca08b89942408da?mt=application/vnd.oci.image.layer.v1.tar+gzip&size=99830)

The upstream seems to intend to exclude some unnecessary files from the final image using `.dockerignore`, but this doesn't seem to work, or the image wasn't built from that source. https://github.com/statsd/statsd/issues/710 also mentions this.

### DIFF
--- a/statsd.yaml
+++ b/statsd.yaml
@@ -35,4 +35,5 @@ update:
   enabled: true
   github:
     identifier: statsd/statsd
+    use-tag: true
     strip-prefix: v
--- a/statsd.yaml
+++ b/statsd.yaml
@@ -1,0 +1,38 @@
+package:
+  name: statsd
+  version: 0.10.2
+  epoch: 0
+  description: Daemon for easy but powerful stats aggregation
+  copyright:
+    - license: MIT
+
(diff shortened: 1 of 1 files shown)


statsd.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+  - runs: |
+      # Set graphite hostname to "graphite"
+      sed -i 's/graphite.example.com/graphite/' exampleConfig.js
+      mv exampleConfig.js ${{targets.destdir}}/usr/src/app/config.js
+
+update:
+  enabled: true
+  github:
+    identifier: statsd/statsd
```
stormqueen1990, line 37:
```suggestion
    identifier: statsd/statsd
    use-tag: true
```

luhring (approved):
👏

PR #11385, opened 2024-01-18
run package tests in CI

imjasonh:


joshrwolf:
are we ready to block on this in presubmit?

imjasonh:
> are we ready to block on this in presubmit?

I'm ready if @vaikas is. Is there something you're worried about?

vaikas (approved):
Thanks!

pdeslaur (approved):

PR #11426, opened 2024-01-19
Add nemo v1.22.0

srishtih:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches


nemo.yaml
```diff
@@ -0,0 +1,75 @@
...
+      # allow site-packages
+      sed -i "s|include-system-site-packages = false|include-system-site-packages = true|g" ${{targets.destdir}}/usr/share/nemo/.venv/pyvenv.cfg
+
+  - uses: strip
+
+update:
+  enabled: true
+  github:
+    identifier: NVIDIA/NeMo
+    strip-prefix: v
```
mamccorm, line 75: I think you need to also filter by 'v' here as there are other version tags such as 'stable' which may get built, like this one: https://github.com/NVIDIA/NeMo/releases/tag/stable

```
tag-filter: v
```

nemo.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+  - runs: |
+      # edit the venv paths
+      rm -rf ${{targets.destdir}}/usr/share/nemo/.venv/bin/__pycache*
+      sed -i "s|/home/build|/usr/share/nemo|g" ${{targets.destdir}}/usr/share/nemo/.venv/bin/*
+
+  - runs: |
+      # allow site-packages
+      sed -i "s|include-system-site-packages = false|include-system-site-packages = true|g" ${{targets.destdir}}/usr/share/nemo/.venv/pyvenv.cfg
+
```
mamccorm, line 70: Do we need an extra step here similar to this package?
- https://github.com/wolfi-dev/os/blob/9e21b9522815e253e6fe89102a999186d1e06c49/az.yaml#L51

i.e something like:

```
  - runs: |
      ln -s /usr/share/nemo/.venv/bin/nemo ${{targets.destdir}}/usr/bin/nemo
```

However if i build nemo and check in the `/usr/share/nemo/.venv/bin` directory, i don't see a 'nemo' executable. if I do the same with the AZ package I do.

nemo.yaml
```diff
@@ -0,0 +1,78 @@
...
+      repository: https://github.com/NVIDIA/NeMo.git
+      tag: v${{package.version}}
+      expected-commit: 0b7467ebd158ef6876dd147189517b25ef626982
+
+  - runs: git submodule update --init --recursive
+
+  - runs: |
+      sed -i '/fasttext/d' requirements/requirements_nlp.txt
+      sed -i 's/megatron_core==0.4.0/megatron_core==0.3.0/' requirements/requirements_nlp.txt
+
```
mamccorm, line 45: I know this is draft btw, so feel free to ignore this comment if not relevant, but looks like we're missing the build step?
 
i.e we're just setting up the virtualEnv and installing the dependencies, but not building nemo? Maybe this what your testing.

I guess that explains why there is no nemo package inside the virtualEnv (my comment below).

[See here](https://github.com/wolfi-dev/os/blob/9e21b9522815e253e6fe89102a999186d1e06c49/az.yaml#L26) for another example.

nemo.yaml
```diff
@@ -0,0 +1,130 @@
...
+      # Install transformer engine
+          # same as Apex
+
+      # pip3 uninstall -y sacrebleu torchtext most likely not needed as these came from nvidia-pytorch image
+
+      # export LHOTSE_REQUIRE_TORCHAUDIO=0
+
+      # install requirements from all requirements_xyz.txt
+      for f in ./requirements/requirements*.txt;
+        do pip3 install --disable-pip-version-check --no-cache-dir -r $f
```
smoser, line 86: this is actually done by setup.py below when '.[all]' is used.  the setup builds its requirements from "all" of the requirements/*.txt

PR #11484, opened 2024-01-20
google-cloud-sdk: update and build with python 3.12

imjasonh:
https://issuetracker.google.com/issues/303280713 and https://cloud.google.com/sdk/docs/install#supported_python_versions indicate that latest version of gcloud should be compatible with Python 3.12

Let's find out!

Editor's note: It turns out gcloud updates are basically manual, since we require a human to update the expected SHA each time. I did this, and to make this easier for auto-update triagers in the future, I added a pre-step that prints the observed SHA.

### DIFF
--- a/google-cloud-sdk.yaml
+++ b/google-cloud-sdk.yaml
@@ -1,7 +1,7 @@
 package:
   name: google-cloud-sdk
   version: 427.0.0
-  epoch: 3
+  epoch: 4
   description: "Google Cloud Command Line Interface"
   copyright:
     - license: Apache-2.0
@@ -9,7 +9,7 @@ package:
     runtime:
       # Required for cyclic redunancy check (gsutil help crcmod)
       - py3-crcmod
-      - python-3.11
+      - python-3.12
 
 environment:
   contents:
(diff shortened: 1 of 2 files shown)


google-cloud-sdk.yaml
```diff
@@ -1,15 +1,15 @@
 package:
   name: google-cloud-sdk
-  version: 427.0.0
-  epoch: 3
+  version: 452.0.0
+  epoch: 4
```
jonjohnsonjr, line 4: 0 I think?

imjasonh: Right you are!

google-cloud-sdk.yaml
```diff
@@ -70,6 +78,14 @@ pipeline:
 
   - uses: strip
 
+test:
+  environment:
+    contents:
+      packages:
+        - wolfi-base
+  pipeline:
+    - runs: gcloud --version
```
jonjohnsonjr, line 87: I'm in love with you.

ajayk (approved):

pdeslaur (approved):

PR #11491, opened 2024-01-20
dockerd: remove redundant runtime dep on ctr

imjasonh:
`ctr` is a subpackage of `containerd` that just splits out the `ctr` binary, which is also provided by `containerd`.

Depending on both `ctr` and `containerd` seems to be confusing our solver, and apk's. At least that's the theory. In any case it's a redundant dependency, since depending on `containerd` will bring in `ctr` regardless.

https://github.com/wolfi-dev/os/blob/ac9f4551e540ae6a198aea0414b6ba1cc521736f/containerd.yaml#L48

### DIFF
--- a/docker.yaml
+++ b/docker.yaml
@@ -1,7 +1,7 @@
 package:
   name: docker
   version: 25.0.0
-  epoch: 0
+  epoch: 1
   description: A meta package for Docker Engine and Docker CLI
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


imjasonh:
```
$ docker run --rm -it cgr.dev/chainguard/wolfi-base sh -c "apk update && apk add dockerd"
...
(12/37) Installing containerd (1.7.3-r0)
(13/37) Installing ctr (1.7.12-r2)
...
```

That's weird right?

PR #11495, opened 2024-01-20
keda-2.11/2.11.2-r11: cve remediation

chainguardian:
keda-2.11/2.11.2-r11: fix GHSA-9763-4f94-gfch/GHSA-2c7c-3mj9-8fqh/

### DIFF
--- a/keda-2.11.yaml
+++ b/keda-2.11.yaml
@@ -3,7 +3,7 @@
 package:
   name: keda-2.11
   version: 2.11.2
-  epoch: 12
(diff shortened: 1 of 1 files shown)


hectorj2f:
This bug is fixed in main but we didn't deploy it on production. The latest package version is `keda-2.11/2.11.2-r12` , as you see the title of the PR refers to `keda-2.11/2.11.2-r11` which is related to the index package code in our services returning the wrong version as latest.

PR #11545, opened 2024-01-22
Bump grpc to fix depends

jonjohnsonjr:


jonjohnsonjr:
Looks like this works:

```
tar -Oxf packages/aarch64/grpc-1.60.0-r3.apk .PKGINFO | grep proto
depend = so:libprotobuf.so.25.2.0
depend = so:libprotoc.so.25.2.0
```

ajayk (approved):

PR #11568, opened 2024-01-22
rebuild packages with valid SBOMs

imjasonh:
These packages haven't been rebuilt in months, sometimes a year+, and their SBOMs are invalid:

```
│ spdx_id must only contain letters, numbers, "." and "-" and must begin with "SPDXRef-", but is: SPDXRef-File-/usr/share/doc/libunistring/libunistring_abt.html
│ spdx_id must only contain letters, numbers, "." and "-" and must begin with "SPDXRef-", but is: SPDXRef-File-/usr/share/doc/libunistring/libunistring_toc.html
│ spdx_id must only contain letters, numbers, "." and "-" and must begin with "SPDXRef-", but is: SPDXRef-File-/usr/share/info/libunistring.info
│ Components missing an supplier: libunistring,libidn2
```

Rebuilding should give them new valid SBOMs.

Also adding a check to CI that the SBOMs we produce are valid. We also [run this check in melange's CI](https://github.com/chainguard-dev/melange/blob/eb44fc3229f8047d17171cf70ed727b13a9a7642/.github/workflows/e2e.yaml#L52), but there are a lot more exotic packages in this repo than there, so we may find more edge cases here.

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -123,9 +123,10 @@ jobs:
         env:
           NTIA_IMAGE: cgr.dev/chainguard/ntia-conformance-checker:latest@sha256:bde670cb65228e873ca2a2d674d407f310b11e608465454c668a5285fbaba3ee
         run: |
+          apk add py3-ntia-conformance-checker
           for f in $(find packages -name '*.apk'); do
               tar -tvf $f var/lib/db/sbom/ > sbom.json
-              docker run --rm --user 0 -v $(pwd)/sbom.json:/sbom.json:ro ${NTIA_IMAGE} -v --file /sbom.json
+              ntia-checker -v --file sbom.json
           done
 
       - name: Check for file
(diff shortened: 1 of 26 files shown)


rawlingsj (dismissed):

luhring (dismissed):

.github/workflows/ci-build.yaml
```diff
@@ -119,6 +119,20 @@ jobs:
               echo "No .apk files found."
           fi
 
+      - name: Check SBOMs
+        env:
+          NTIA_IMAGE: cgr.dev/chainguard/ntia-conformance-checker:latest@sha256:bde670cb65228e873ca2a2d674d407f310b11e608465454c668a5285fbaba3ee
```
mattmoor, line 124: What’s the image for?

mattmoor (dismissed):

.github/workflows/ci-build.yaml
imjasonh: A previous iteration of this did `docker run <image>`, so this isn't needed anymore. Good catch, it's gone.

Approval again pls 🙏

mattmoor (approved):

PR #11584, opened 2024-01-22
rebuild packages with valid licenses

imjasonh:


kaniini:
Most of these (especially GNU packages) should be GPL-3.0-or-later not GPL-3.0-only.

kaniini:
I think it is ok to assume that but you should probably spot check.  The difference is in the copyright text, one says “version X or later” and the other says “version X”.

For example Linux is GPL-2.0-only.

imjasonh:
> I think it is ok to assume that but you should probably spot check. The difference is in the copyright text, one says “version X or later” and the other says “version X”.
> 
> For example Linux (kernel) is GPL-2.0-only.

- neo4j just says "GPL v3" https://neo4j.com/licensing
- minio says "AGPL v3" https://min.io/compliance
- dependency-track was actually Apache 😬  (fixed)
- pygithub seems to be GPL-3.0-or-later (doesn't mention any specific version) https://github.com/pygithub/pygithub?tab=GPL-3.0-2-ov-file
- proxysql seems to be GPL-3.0-or-later https://github.com/sysown/proxysql/blob/v2.x/LICENSE

renovate.yaml
```diff
@@ -1,10 +1,10 @@
 package:
   name: renovate
   version: 37.144.0
-  epoch: 0
+  epoch: 1
   description: "Automated dependency updates. Multi-platform and multi-language."
   copyright:
-    - license: AGPL-3.0
+    - license: AGPL-3.0-or-later
```
ajayk, line 7: renovate has  it as AGPL-3.0-only on the README https://github.com/renovatebot/renovate

ajayk (approved):

PR #11658, opened 2024-01-23
Adding Collectd

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/collectd.yaml
+++ b/collectd.yaml
@@ -6,18 +6,6 @@ package:
   copyright:
     - license: MIT
 
-# NOTE: Commented out packages are ones we don't have today in wolfi
-#
-# Thus list also uses the alpine package for comparison. HOWEVER there are a lot
-# of 'optional' packages listed in the pre-reqisites section of the git repo
-# docs which alpine has choosen not to include.
-#
-# QUESTION: Assume we do the same here and mirror what alpine has done?
-#
-# References:
-# - https://git.alpinelinux.org/aports/tree/community/collectd/APKBUILD#n13
-# - https://github.com/collectd/collectd/blob/main/README.md
-# - https://www.collectd.org/download.html#source
 environment:
   contents:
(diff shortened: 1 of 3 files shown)


collectd.yaml
```diff
@@ -0,0 +1,179 @@
...
+      - openldap-dev
+      - openssl-dev>3
+      - owfs-dev
+      - pango-dev
+      - patchelf
+      - perl-dev
+      - pkgconf
+      - pkgconf-dev
+      - postgresql-16-dev
+      - python3-dev
```
jamonation, line 59:
```suggestion
```

Since the `--disable-python` flag is included

collectd.yaml
```diff
@@ -0,0 +1,179 @@
...
+      expected-commit: 3f935d017e81b7eaf84c9df421490230845dc5c0
+
+  - runs: |
+      ./build.sh
+
+      if "${{build.arch}}" == "x86_64"; then
+        _flags="--disable-turbostat"
+      elif "${{build.arch}}" == "aarch64"; then
+        _flags="--disable-java"
+      fi
```
jamonation, line 82:
```suggestion
```

Does this build without all these? I see a `--disable-turbostat` flag further down. I'd try without and drop the `$_flags` bit at the end of configure as a test

imjasonh: I'd like to remove these too if we can. I don't like the idea of having different features enabled per-arch, since we basically only test things on x86_64. If someone used the aarch64 image expecting Java to work, they would be surprised.

developer-guy: removed, yes.

jamonation (commented):
Looks good to me here, passes CI on both architectures.

dlorenc (approved):

PR #11769, opened 2024-01-25
kots/1.106.0-r0: cve remediation

chainguardian:
kots/1.106.0-r0: fix GHSA-vh7g-p26c-j2cw


hectorj2f:
We can safely close this PR, this CVE has been marked as false-positive in our advisory data. Once we filter the scan results using the `concluded` filter, we won't open these PRs.

PR #11770, opened 2024-01-25
kubeflow-pipelines/2.0.5-r2: cve remediation

chainguardian:
kubeflow-pipelines/2.0.5-r2: fix GHSA-f9jg-8p32-2f55

### DIFF
--- a/kubeflow-pipelines.yaml
+++ b/kubeflow-pipelines.yaml
@@ -1,7 +1,7 @@
 package:
   name: kubeflow-pipelines
   version: 2.0.5
-  epoch: 2
+  epoch: 3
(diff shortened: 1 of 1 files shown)


hectorj2f:
This PR change might require upstream change to support such a version, they are setting the api to v1.11 here https://github.com/kubeflow/pipelines/blob/2.0.5/go.mod#L64. I'll update the advisory data.

hectorj2f:
Closing this PR! It is marked as `false-positive` https://github.com/hectorj2f/advisories/blob/c96e3ff4b64bb141f3464d420222ab744c394633/kubeflow-pipelines.advisories.yaml#L179.

PR #11848, opened 2024-01-26
coredns/1.11.1-r9: cve remediation

chainguardian:
coredns/1.11.1-r9: fix CVE-2023-49295


hectorj2f:
We aren't going to remediate coredns with a newer version of coredns, but we should update it as soon as there is a more stable version than a rc.

PR #11795, opened 2024-01-25
thanos-0.33 package update

stgarf:
* "thanos": update to 0.33

#### For security-related PRs

- [x] The security fix is recorded in the [advisories](https://github.com/wolfi-dev/advisories) repo
	- [x] [thanos-0.33 advisories PR](https://github.com/wolfi-dev/advisories/pull/985) 

#### For version bump PRs

- [x] The `epoch` field is reset to 0

PR #11862, opened 2024-01-26
removing version streams which are not supported

mamccorm:
Removing unsupported version streams (application maintainer only supports a single version stream), and enabling single version stream, which will now also produce the latest (which we where missing before).


ajayk:
Closing this https://github.com/wolfi-dev/os/pull/11873 and  the @stgarf had a previous PR #11795

PR #11873, opened 2024-01-26
Update Thanos package to latest version

stgarf:
* "thanos": update to 0.33
* This PR also moves Thanos away from version streams since only the latest is released upstream.

#### For version bump PRs


thanos.yaml
```diff
@@ -1,13 +1,13 @@
...
+  name: thanos
+  version: 0.33.0
+  epoch: 0
   description: Highly available Prometheus setup with long term storage capabilities.
   copyright:
     - license: Apache-2.0
   dependencies:
     provides:
-      - thanos=0.32.999
+      - thanos=0.33.999
```
ajayk, line 10:
```suggestion
```

stgarf: Ack, ty -- removed section

ajayk (approved):

PR #11964, opened 2024-01-29
Switch to the dedicated action

mattmoor:
This version is written in javascript and revokes the token when the workflow completes.

### DIFF
--- a/.github/workflows/digestabot.yaml
+++ b/.github/workflows/digestabot.yaml
@@ -18,7 +18,7 @@ jobs:
     steps:
     - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
 
-    - uses: chainguard-dev/actions/octo-sts@main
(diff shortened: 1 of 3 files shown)


.github/workflows/digestabot.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
     steps:
     - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
 
-    - uses: chainguard-dev/actions/octo-sts@main
+    - uses: chainguard-dev/octo-sts-action@6177b4481c00308b3839969c3eca88c96a91775f # v1.0.0
```
imjasonh, line 21: We should do this closer to where we might use the secret, since it will expire after 1h.

imjasonh: nvm this shouldn't take that long.

imjasonh (approved):

.github/workflows/digestabot.yaml
mattmoor: I think these are all used immediately, but totally agree (this is one of the practices in a doc I'm working on)

PR #12207, opened 2024-01-31
src-fingerprint: fix go-git cve

ajayk:


src-fingerprint.yaml
```diff
@@ -13,6 +13,7 @@ environment:
       - busybox
       - ca-certificates-bundle
       - go
+      - patch
```
luhring, line 16: I don't think you need this since we're using a pipeline below

pdeslaur (approved):

PR #12223, opened 2024-02-01
fix(font-wqy-zenhei): Don't overwrite font config, patch instead

EyeCantCU:
Fixes: When the font is retrieved from sourceforge, it overwrites the file that is supposed to stand in its place. This patches the config file instead to prevent the replacement font config from being overwritten

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches


font-wqy-zenhei.yaml
```diff
@@ -24,6 +24,10 @@ pipeline:
       expected-sha256: e4b7e306475bf9427d1757578f0e4528930c84c44eaa3f167d4c42f110ee75d6
       uri: https://downloads.sourceforge.net/wqy/wqy-zenhei-${{package.version}}.tar.gz
 
+  - uses: patch
```
mamccorm, line 27: Great work figuring this one out - can we add a couple of lines code comment explaining the purpose of the patch and what it's doing?

Reason - if / when upstream makes code changes in this area, the patch will break and we'll need to re-generate and apply it. So any useful info we can add here will help whoever is triaging in future

mamccorm:
```suggestion
# When this font is retrieved from sourceforge, it overwrites the file that is
# supposed to stand in its place. This patches the config file instead to
# prevent the replacement font config from being overwritten.
  - uses: patch
```

EyeCantCU: We definitely can! Thank you for the suggested change

mamccorm (dismissed):

mamccorm (approved):

PR #12367, opened 2024-02-02
Fix jruby and add concurrent-ruby java extensions/tests

found-it:
We weren't actually building the concurrent-ruby java extensions to work with jruby. I went to add tests to concurrent-ruby and realized we needed to make jruby installable alongside ruby to get all the functionality of concurrent-ruby. The logstash-core gem also makes use of the java functionality of concurrent-ruby, so this is required for that gem to work.

Ideally we use rbenv or rvm to make virtual machines rather than installing a system ruby alongside system jruby, but this is a workaround for now.

### DIFF
--- a/jruby-9.4.yaml
+++ b/jruby-9.4.yaml
@@ -1,7 +1,7 @@
 package:
   name: jruby-9.4
   version: 9.4.5.0
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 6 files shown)


rawlingsj (dismissed):

stormqueen1990 (dismissed):

stormqueen1990 (approved):
Thank you for this 🎉

PR #12409, opened 2024-02-03
logstash-exporter

developer-guy:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/logstash-exporter.yaml
+++ b/logstash-exporter.yaml
@@ -1,0 +1,50 @@
+package:
+  name: logstash-exporter
+  version: 1.6.1
+  epoch: 0
+  description: Prometheus exporter for Logstash written in Go
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - go
+  environment:
+    GITHUB_REPO: "github.com/kuskoman/logstash-exporter"
+
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: 387fb36f814e74e81dd14d82a948e61cefa76f5c
+      repository: https://github.com/kuskoman/logstash-exporter
+      tag: v${{package.version}}
+
+  - uses: go/build
+    with:
+      packages: ./cmd/exporter/main.go
+      ldflags: -s -w -X ${GITHUB_REPO}/config.Version="${{package.version}}" -X ${GITHUB_REPO}/config.GitCommit="$(git rev-parse --short HEAD)" -X ${GITHUB_REPO}/config.BuildDate="$(date -u +%Y-%m-%dT%H:%M:%SZ)"
+      repository: https://github.com/kuskoman/logstash-exporter
+      output: logstash-exporter
+
+  - uses: strip
+
+update:
+  enabled: true
+  ignore-regex-patterns:
+    - "-pre*"
+  github:
+    identifier: kuskoman/logstash-exporter
+    strip-prefix: v
+    use-tag: true
+
+test:
+  environment:
+    contents:
+      packages:
+        - wolfi-base
+  pipeline:
+    - runs: |
+        logstash-exporter -version


logstash-exporter.yaml
```diff
@@ -0,0 +1,47 @@
...
+      expected-commit: 387fb36f814e74e81dd14d82a948e61cefa76f5c
+      repository: https://github.com/kuskoman/logstash-exporter
+      tag: v${{package.version}}
+
+  - uses: go/build
+    with:
+      packages: ./cmd/exporter/main.go
+      ldflags: -s -w -X ${GITHUB_REPO}/config.Version="${{package.version}}" -X ${GITHUB_REPO}/config.GitCommit="$(git rev-parse --short HEAD)" -X ${GITHUB_REPO}/config.BuildDate="$(date -u +%Y-%m-%dT%H:%M:%SZ)"
+      repository: https://github.com/kuskoman/logstash-exporter
+      output: main
```
mamccorm, line 30: Lets give the binary a more suitable name:

```suggestion
      output: logstash-exporter
```

logstash-exporter.yaml
```diff
@@ -0,0 +1,47 @@
...
+    strip-prefix: v
+
+test:
+  environment:
+    contents:
+      packages:
+        - wolfi-base
+  pipeline:
+    - runs: |
+        main -version
```
mamccorm, line 47:
```suggestion
        logstash-exporter -version
```

logstash-exporter.yaml
```diff
@@ -0,0 +1,47 @@
...
+  - uses: go/build
+    with:
+      packages: ./cmd/exporter/main.go
+      ldflags: -s -w -X ${GITHUB_REPO}/config.Version="${{package.version}}" -X ${GITHUB_REPO}/config.GitCommit="$(git rev-parse --short HEAD)" -X ${GITHUB_REPO}/config.BuildDate="$(date -u +%Y-%m-%dT%H:%M:%SZ)"
+      repository: https://github.com/kuskoman/logstash-exporter
+      output: main
+
+  - uses: strip
+
+update:
```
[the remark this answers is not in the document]
developer-guy: `wolfictl check update` looks happy, so, it shouldn't be any problem I guess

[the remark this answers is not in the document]
mamccorm: There are pre-releases here:
v2.0.0-pre2 (tag name). Release v2.0.0-pre2 (release name)

Depending which one your checking (i.e tags or releases, then these will need to be filtered out, otherwise it'll build for pre-releases.

There are some examples in wolfi using `ignore-regex-patterns:

logstash-exporter.yaml
```diff
@@ -0,0 +1,47 @@
...
+      repository: https://github.com/kuskoman/logstash-exporter
+      output: logstash-exporter
+
+  - uses: strip
+
+update:
+  enabled: true
+  github:
+    identifier: kuskoman/logstash-exporter
+    strip-prefix: v
```
mamccorm, line 38: I think this might be what we need:

```suggestion
  enabled: true
  ignore-regex-patterns:
    - 'pre*'
  github:
    identifier: kuskoman/logstash-exporter
    strip-prefix: v
    use-tag: true
```

logstash-exporter.yaml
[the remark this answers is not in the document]
mamccorm: It'll still include pre-releases though so we'll need to filter those out using ignore-regex. I'll make a suggestion inline below but check if the syntax is correct :)

logstash-exporter.yaml
developer-guy: iinm, prereleases are not the real releases so they can't be listed unless you specifically ask for them, also, I remember that we use `use-tag: true` when there is no actual releases for the tags, so, in this case we might don't need this one.

developer-guy: I'm mentioning @rawlingsj for additional help since he is the master of these updates 🙈

mamccorm: I took this from our existing implementations in wolfi:
- https://github.com/search?q=repo%3Awolfi-dev%2Fos%20ignore-regex-patterns%3A&type=code

Another example, filters out 'test':
- https://github.com/wolfi-dev/os/blob/4a1394715ffe432b7f4f17f11e116ddca69f7199/py3-xattr.yaml#L39

Because of:
- https://github.com/xattr/xattr/releases/tag/test-v0.9.8

Here is another example:
 - https://github.com/wolfi-dev/os/blob/4a1394715ffe432b7f4f17f11e116ddca69f7199/istio-operator-1.20.yaml#L45
 
 And looking at the git repo:
 - https://github.com/istio/istio/tags
 
 Also you can see when we need to do these filtering we typically flip over to tags. the istio repo is similar to yours where it has releases and tags, but filtering needs to be applied:
 - https://github.com/istio/istio/releases
 - https://github.com/istio/istio/tags

developer-guy: ah great! yeah, okay, thanks Mark!

mamccorm: You stil need 'use-tag' (see thread above). If you want to base it on releases your going to have to strip out the prepending 'Release ' (inc space), from the version. I don't think thats possible, which is why the other examples we have in wolfi revert to 'use_tag', whenever the upstream repository publishes funny tags.

Example which is very similar to your scenario, where we apply filtering and also use-tags:
https://github.com/wolfi-dev/os/blob/4a1394715ffe432b7f4f17f11e116ddca69f7199/istio-operator-1.20.yaml#L47

https://github.com/wolfi-dev/os/blob/4a1394715ffe432b7f4f17f11e116ddca69f7199/istio-operator-1.20.yaml#L51

If you look at the upstream repo you will see they do publish both releases and tags. The release name is in a similar format to yours here:

- https://github.com/istio/istio/releases
- https://github.com/istio/istio/tags

But we have to add use-tags and apply the filtering here. If there was no issue with having to filter out non-release tags / names, then we'd not need any of this

mamccorm (approved):

PR #12487, opened 2024-02-05
fix(volumes-webapp): Remove urllib3 as a buildtime dependency

EyeCantCU:
Fixes: When installed as a buildtime dependency, urllib3 doesn't get installed to the virtual environment, leaving the volumes webapp unable to find the urllib3 module

#### For new package PRs only

### DIFF
--- a/kubeflow-volumes-web-app.yaml
+++ b/kubeflow-volumes-web-app.yaml
@@ -1,7 +1,7 @@
 package:
   name: kubeflow-volumes-web-app
   version: 1.8.0
-  epoch: 2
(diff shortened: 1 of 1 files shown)


imjasonh (commented):
Is there a package test we could add to this to ensure this change fixes the issue? Something that says "before this change <this test> doesn't work, and after this change it does"?

EyeCantCU:
Implemented a test for this. All packages should be able to use something like this. With a virtual environment, you need to activate it first however

dlorenc (approved):

PR #12503, opened 2024-02-05
chore: change ruby3.2-jrjackson to build using rake

stormqueen1990:
* Change the package to build using `rake` instead of `mvn`, which is the recommended way by upstream, and update the dependency on JUnit to mitigate a CVE.

* Add a basic test to ensure the gem works.

### DIFF
--- a/ruby3.2-jrjackson.yaml
+++ b/ruby3.2-jrjackson.yaml
@@ -1,30 +1,20 @@
 package:
   name: ruby3.2-jrjackson
   version: 0.4.18
-  epoch: 2
+  epoch: 3
   description: A mostly native JRuby wrapper for the java jackson json processor jar
   copyright:
     - license: Apache-2.0
-  dependencies:
-    runtime:
-      - openjdk-11-default-jvm
 
 environment:
   contents:
     packages:
-      - bash
       - build-base
-      - busybox
-      - ca-certificates-bundle
-      - git
       - jruby-9.4
       - jruby-9.4-default-ruby
-      - maven
-      - openjdk-11
-      - openjdk-11-default-jvm
-      - openjdk-11-jre
+      - openjdk-8-default-jdk # build with JDK 8 since this is the declared version in the upstream Mavenfile
   environment:
-    JAVA_HOME: /usr/lib/jvm/java-11-openjdk
+    JAVA_HOME: /usr/lib/jvm/java-1.8-openjdk
 
 pipeline:
   - uses: git-checkout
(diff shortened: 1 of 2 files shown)


ruby3.2-jrjackson.yaml
```diff
@@ -80,3 +52,27 @@ update:
...
     strip-prefix: v
     use-tag: true
+
+test:
+  environment:
+    contents:
+      packages:
+        - busybox
+        - jruby-9.4
+        - jruby-9.4-default-ruby
```
found-it, line 62:
```suggestion
```

found-it: I don't think you need default ruby here right since you're using jruby for the test

ruby3.2-jrjackson.yaml
```diff
@@ -1,30 +1,20 @@
 package:
   name: ruby3.2-jrjackson
   version: 0.4.18
-  epoch: 2
+  epoch: 3
   description: A mostly native JRuby wrapper for the java jackson json processor jar
   copyright:
     - license: Apache-2.0
-  dependencies:
```
found-it, line 8: Good call not pinning this

ruby3.2-jrjackson.yaml
```diff
@@ -33,32 +23,14 @@ pipeline:
...
-      mkdir -p ./lib/com/fasterxml/jackson/core/jackson-databind/${JACKSON_VERSION}
-      mkdir -p ./lib/com/fasterxml/jackson/module/jackson-module-afterburner/${JACKSON_VERSION}
-      mkdir -p ./lib/jrjackson/jars
-
-      mvn dependency:copy -Dartifact=com.fasterxml.jackson.core:jackson-annotations:${JACKSON_VERSION} -DoutputDirectory=lib/com/fasterxml/jackson/core/jackson-annotations/${JACKSON_VERSION}
-      mvn dependency:copy -Dartifact=com.fasterxml.jackson.core:jackson-core:${JACKSON_VERSION} -DoutputDirectory=lib/com/fasterxml/jackson/core/jackson-core/${JACKSON_VERSION}
-      mvn dependency:copy -Dartifact=com.fasterxml.jackson.core:jackson-databind:${JACKSON_VERSION} -DoutputDirectory=lib/com/fasterxml/jackson/core/jackson-databind/${JACKSON_VERSION}
-      mvn dependency:copy -Dartifact=com.fasterxml.jackson.module:jackson-module-afterburner:${JACKSON_VERSION} -DoutputDirectory=lib/com/fasterxml/jackson/module/jackson-module-afterburner/${JACKSON_VERSION}
+      bundle install
+      rake
```
found-it, line 32: Just a sanity check, I saw a `vendor_jars` task in the Rakefile. Does that vendor happen in the `lib` directory so it gets packaged and installed with the gem or does it drop it in a different directory?

stormqueen1990: The JAR files are placed in the `vendor` directory at the end but I'm not entirely sure about the mechanism. I found this in the jrjackson build files: https://github.com/guyboertje/jrjackson/blob/86e21566467349cb92ee5687f2933335ee227c93/lib/jrjackson/build_info.rb#L37-L45

found-it: Do they get bundled in the gem itself so they come with the APK?

found-it: Looks like it does!

found-it (dismissed):

ruby3.2-jrjackson.yaml
found-it, line 32:
```suggestion
      bundle exec rake
```

found-it: Technically this should be executed in the context of the bundle. Practically in this case it won't matter, but it's a case of good practice

found-it (approved):

PR #12519, opened 2024-02-06
Bump epoch

mamccorm:
Bumping epoch so we get a fresh CI build of this package.

### DIFF
--- a/spark.yaml
+++ b/spark.yaml
@@ -1,7 +1,7 @@
 package:
   name: spark
   version: 3.5.0
-  epoch: 0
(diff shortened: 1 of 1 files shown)


spark.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: spark
   version: 3.5.0
-  epoch: 0
+  epoch: 1
```
pdeslaur, line 4:
```suggestion
  epoch: 2
```

pdeslaur (approved):

PR #12555, opened 2024-02-06
added a basic test for w3m command

dustinkirkland:
Use w3m to fetch https://example.com and dump to stdout.  Note that this will require internet access.  We could nerf that, or use it to render a local html page instead?


w3m.yaml
```diff
@@ -2,7 +2,7 @@
 package:
   name: w3m
   version: 0.5.3.20230718
-  epoch: 0
+  epoch: 1
```
imjasonh, line 5: Technically I don't think we need to bump the epoch to add a test.

The resulting package should be identical from before the test, and CI will rebuild the package and run the test on it to ensure your test is right.

It's not a huge deal if we do bump the epoch, it's just not necessary.

dustinkirkland: Ah, gotcha.  I wasn't sure about that.  Want me to move it back to 0?

rawlingsj (approved):

w3m.yaml
rawlingsj: for this one we may as well leave it

PR #12570, opened 2024-02-06
Update temporal-server.yaml

cjohannsen81:
Removing tags as releases are available and version tags are not shown in the image


ajayk:
```
    use-tag: true
    tag-filter: v
```

these are  referring to the GitHub tags whether to use the tags for fetching in the git-checkout

cjohannsen81:
Thanks @ajayk

Issue #9187, opened 2023-11-29
busybox wget has different behaviour between arm and amd


rawlingsj:
To reproduce the error, we can run two containers, one arm, the second amd
```
docker run --privileged --platform linux/aarch64 --rm -it -v "${PWD}:/work" -w /work cgr.dev/chainguard/wolfi-base:latest
```
```
docker run --privileged --platform linux/amd64 --rm -it -v "${PWD}:/work" -w /work cgr.dev/chainguard/wolfi-base:latest
```

In each of them run `wget pear.php.net`

The results are different

arm:
```
wget pear.php.net
Connecting to pear.php.net (109.203.101.62:80)
Connecting to pear.php.net (109.203.101.62:443)
wget: note: TLS certificate validation not implemented
saving to 'index.html'
```

amd:
```
Connecting to pear.php.net (109.203.101.62:80)
Connecting to pear.php.net (109.203.101.62:443)
wget: note: TLS certificate validation not implemented
wget: TLS error from peer (alert code 40): handshake failure
wget: error getting response: Connection reset by peer
```

rawlingsj:
This doesn't explain the difference in behaviours on both platforms but did see this https://github.com/wolfi-dev/os/blob/d2607fb882034652448e1adf1df785bb76f2f8da/busybox/busyboxconfig#L998

Wonder if this would help?

kaniini:
busybox cannot depend on openssl, as openssl needs busybox to build.  solution is actually to provide an ssl_client binary which proxies to openssl.

rawlingsj:
TBH, I'm not sure how to do that is this something you could help with?

kaniini:
Yeah I can do it!

xnox:
> busybox cannot depend on openssl, as openssl needs busybox to build. solution is actually to provide an ssl_client binary which proxies to openssl.

this is factually incorrect. There is no build-dependency or runtime-dependency loops between openssl and busybox in any form. Busybox openssl feature doesn't link against libssl, it simply has code to exec `/usr/bin/openssl`. Most distributions that enable this feature, do not even add any runtime depends or recommends on openssl client utility.
And separately openssl can be built without busybox (non-full) or coreutils. Thus enabling openssl feature in busybox doesn't create any bootstrapping concerns.

PR #12631, opened 2024-02-07
Remove wget from default busybox config.

smoser:
Wget is a http/https client.  We have decided to remove it from the default busybox build as simply good hygiene.

Images that need a command line http client have the following options:
 * add curl package
 * add wget package
 * use busybox-full package instead of busybox.

Also note that ssl config option is dropped.  This is because wget is the only thing that was pulling it in.

See also https://github.com/wolfi-dev/os/pull/12583


imjasonh:
I checked to be sure, and Melange's `fetch` pipeline uses the `wget` package itself, instead of depending on it being provided by `busybox`: https://github.com/chainguard-dev/melange/blob/a298935850e80ecdb2e05c094ac507939b2bca47/pkg/build/pipelines/fetch.yaml#L5

It looks like our Alpine-based `busybox` image relies on wget being provided by busybox, but that's unrelated to this change, since it's the Alpine-based image: https://github.com/chainguard-images/images/blob/f13ca95620c6dd251f421426b96edb87054824cf/images/busybox/config/alpine/main.tf#L14

dlorenc:
I'm sure this will break something but it makes sense anyway.

imjasonh:
We may want to make our images' dev variants include `wget` explicitly after this, and/or make it pull in `busybox-full` instead of just `busybox`

https://github.com/chainguard-images/images-private/blob/450a1910ec6436b9f69f2e607f52e2a9b9a3b589/tflib/publisher/main.tf#L101

imjasonh (approved):

vaikas:
https://github.com/wolfi-dev/os/issues/9187

PR #12640, opened 2024-02-07
Install logstash plugins at build-time

found-it:
Logstash likes to install plugins itself rather than finding them on disk. We have a few plugins that need patching so this is just composing everything so logstash can make use of them.


found-it:
Advisories https://github.com/wolfi-dev/advisories/pull/1195

found-it:
One more advisory https://github.com/wolfi-dev/advisories/pull/1196

rawlingsj (dismissed):

rawlingsj (approved):

PR #12676, opened 2024-02-08
opensearch-dashboards-2: New package

pdeslaur:
There are 3 known CVEs on the package - we'll follow up with advisories.

Version's policy: https://endoflife.date/opensearch

### DIFF
--- a/opensearch-dashboards-2.yaml
+++ b/opensearch-dashboards-2.yaml
@@ -15,6 +15,7 @@ environment:
       - ca-certificates-bundle
       - gcc-12
       - gcc-12-default
+      - git
       - node-gyp
       - nodejs-18
       - py3-setuptools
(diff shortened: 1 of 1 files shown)


opensearch-dashboards-2.yaml
```diff
@@ -0,0 +1,83 @@
...
+      sed -i 's/\("osd:bootstrap": "scripts\/use_node scripts\/build_ts_refs\)\( && scripts\/use_node scripts\/register_git_hook\)/\1 --allow-root\2 --allow-root/' package.json
+
+  - runs: |
+      set -x
+
+      # Our commond LDFLAGS cause some issues when building for aarch64. We
+      # unset our global flags to allow the build to succeed.
+      unset LDFLAGS
+      yarn osd bootstrap --allow-root
+      yarn build-platform --linux-arm --skip-os-packages --release --allow-root
```
rawlingsj, line 51: I'm wondering if these flags need to change based on which arch is building. I.e x86 or arm. 

I can take a look shortly

rawlingsj (commented):
Just the one comment which I can check if need

opensearch-dashboards-2.yaml
rawlingsj, line 51: I also noticed that the plugins folder is empty..
```
ls build/opensearch-dashboards/plugins/
```

I'm guessing we'll need to build them too

rawlingsj: I'm not sure yet if we should expect plugins built into the ./plugins folder, let's come back to this.

dlorenc (approved):

PR #12710, opened 2024-02-08
add gitsign config subpackage

cpanato:
- add gitsign config subpackage

### DIFF
--- a/gitsign.yaml
+++ b/gitsign.yaml
@@ -1,7 +1,7 @@
 package:
   name: gitsign
   version: 0.8.0
-  epoch: 1
+  epoch: 2
   description: Keyless Git signing with Sigstore!
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


imjasonh (changes requested):
I'd like @wlynch's ✅ to make sure this is right.

gitsign.yaml
```diff
@@ -13,6 +13,18 @@ pipeline:
...
+subpackages:
+  - name: "${{package.name}}-config"
+    description: "GitSign config"
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/etc/
+          git config --global commit.gpgsign true       # Sign all commits
+          git config --global tag.gpgsign true          # Sign all tags
+          git config --global gpg.x509.program gitsign  # Use gitsign for signing
+          git config --global gpg.format x509           # gitsign expects x509 args
```
wlynch, line 25:
```suggestion
          git config --system commit.gpgsign true       # Sign all commits
          git config --system tag.gpgsign true          # Sign all tags
          git config --system gpg.x509.program gitsign  # Use gitsign for signing
          git config --system gpg.format x509           # gitsign expects x509 args
```

I think we probably want `--system` here so it configures for any user. (`--global` is global for the current user)

cpanato: done

wlynch (commented):
LGTM (can't approve since I'm not a wolfi-dev member 😢 )

imjasonh (approved):

PR #12735, opened 2024-02-08
fix(avahi): update to rc tag to fix vulns

luhring:
This updates `avahi` to a new `v0.9-rc1` tag. We had been using `v0.8` previously, which is very old and is lacking several vulnerability fixes. There's an upstream issue about this, and it looks like there could be a full-fledged release out soon: https://github.com/avahi/avahi/issues/503

This also fixes some aspects of the Wolfi packaging: we had been relying on built artifacts in the project's release assets, as opposed to building everything from source ourselves. This PR switches to building everything ourselves.

We're missing one thing in this PR, which is the manpage generation, which looks like it requires `xmltoman` to be packaged. I'll come back to this soon, but in the meantime this gives us fixes to a handful of vulnerabilities.


jonjohnsonjr (approved):

evverx:
I came from https://github.com/avahi/avahi/issues/503 and since it came with no release notes I thought I could mention a few changes that could affect Wolfi.

As far as I understand `systemd` isn't used so it should be possible to pass `--disable-libsystemd`. That part of `avahi` is used to send notifications to systemd only and can be safely turned off.

The hard-coded runtime "/run" directory was replaced with `runstatedir` to make it easier for various distros to decide what they need. systemd-based distros set it to `/run` and distros without systemd set it to `/var/run` or `/run` depending on what their use case is. By default it should be `/var/run` (`$(localstatedir)/run`)

The D-Bus policies are no longer installed to `${datadir}` according to https://gitlab.freedesktop.org/dbus/dbus/-/blob/dbus-1.14.0/NEWS#L43-54. I'm not sure how relevant it is though.

> it requires xmltoman to be packaged

The OpenEmbedded folks aren't happy about introducing it into their build pipelines: https://github.com/avahi/avahi/issues/527 and Debian Fedora/RH are OK with moving away from it too. It isn't a priority though.

PR #12837, opened 2024-02-11
Fix openldap patch and bump to 2.6.7

dlorenc:
Something changed in the file near the patch so I had to rewrite it.


dlorenc:
Ref: https://github.com/wolfi-dev/os/pull/12038

cpanato (approved):

PR #12843, opened 2024-02-11
fix(linux-pam): Relocate PAM configuration files from util-linux to linux-pam

EyeCantCU:
Fixes: This moves PAM configuration files out of util-linux and into pam-linux, properly installs the configuration files, and creates a PAM configuration file in util-linux for sulogin


imjasonh (commented):
Is there a package test we could add for this to give us more confidence this is the correct way to install this?

Ideally something that would have failed prior to this change, but passes now?

EyeCantCU:
> Is there a package test we could add for this to give us more confidence this is the correct way to install this?
> 
> Ideally something that would have failed prior to this change, but passes now?

Yup :). Should be as easy as verifying `su` works as currently it fails. I will write one up

EyeCantCU:
A test has been added :)

imjasonh (approved):
Thanks!

PR #12844, opened 2024-02-11
feat(packages): Add shadow

EyeCantCU:
Fixes: Adds a package for shadow, providing PAM login and password utilities

#### For new package PRs only

### DIFF
--- a/shadow/useradd-defaults.patch
+++ b/shadow/useradd-defaults.patch
@@ -1,0 +1,24 @@
+From: Jakub Jirutka <jakub@jirutka.cz>
+Date: Sun, 19 Dec 2021 21:50:00 +0100
+Subject: [PATCH] Change some useradd defaults
+
+--- a/src/useradd.c
++++ b/src/useradd.c
+@@ -87,14 +87,14 @@
+ /*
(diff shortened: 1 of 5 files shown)


dlorenc (approved):

shadow/useradd-defaults.patch
```diff
@@ -0,0 +1,24 @@
...
+  * These defaults are used if there is no defaults file.
+  */
+-static gid_t def_group = 1000;
++static gid_t def_group = 100;
+ static const char *def_groups = "";
+ static const char *def_gname = "other";
+ static const char *def_home = "/home";
+-static const char *def_shell = "/bin/bash";
++static const char *def_shell = "/bin/ash";
+ static const char *def_template = SKEL_DIR;
```
xnox, line 18: Above two hunks are highly unusual.

kaniini: This hunk is to set the default shell to busybox ash, which is the one usually available on alpine.  Not sure about the GID change.  That one seems wrong to me, as Alpine starts at GID 1000 for new groups.

PR #12881, opened 2024-02-12
opensearch: package internal and external plugins as subpackages (generate 41 packages)

rawlingsj:
The main motivation for putting all the plugins in one package is they all share the same tag version, even though 19 of the plugins are in different repos, but the same org.

If we had separate YAMLs for them then we’d update then automatically at different times, which would then break any downstream consumers of these packages.

So having them all in a single yaml means we build and publish opensearch + the main 41 related plugins together

### DIFF
--- a/opensearch-2.yaml
+++ b/opensearch-2.yaml
@@ -24,6 +24,7 @@ environment:
       - gradle
       - openjdk-11
       - openjdk-11-default-jvm
+      - patch
       - unzip
   environment:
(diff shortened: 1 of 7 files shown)


rawlingsj:
re vulns: I'll start with 

```
└── 📄 /usr/share/opensearch/plugins/opensearch-repository-hdfs/htrace-core4-4.2.0-incubating.jar
        📦 jackson-databind 2.4.0 (java-archive)
```

found-it (approved):

PR #12909, opened 2024-02-13
rust-1.75: continue the rust bootstrap chain

imjasonh:
rust 1.76 came out and the `rust` package got updated in https://github.com/wolfi-dev/os/commit/150950c56615c9e52c94dae4a422f5e201e0cd16

This continues our bootstrap chain so that when bootstrapping 1.76 will be built from 1.75, which is built from 1.74.

I think I did this right.

### DIFF
--- a/rust-1.75.yaml
+++ b/rust-1.75.yaml
@@ -1,0 +1,121 @@
+package:
+  name: rust-1.75
+  version: 1.75.0
+  epoch: 0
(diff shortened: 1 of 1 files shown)


ajayk:
I think we can convert this to manual version stream and let the updates continue with the same existing format

stormqueen1990 (approved):

PR #12914, opened 2024-02-13
use vendored-openssl feature with lychee package

kranurag7:
_vendored-openssl compiles and statically links a copy of OpenSSL._ (from lychee docs)
Ref: https://github.com/lycheeverse/lychee/blob/13f4339710d76831d9daf961584d796cee4847d2/.github/workflows/release.yml#L70 

For static linking. I copied the lychee binary in a multistage pipeline and got the following error. 
```bash
lychee: error while loading shared libraries: libssl.so.3: cannot open shared object file: No such file or directory
```
this commit updates the package so that we get a statically linked binary. 

Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/lychee.yaml
+++ b/lychee.yaml
@@ -1,7 +1,7 @@
 package:
   name: lychee
   version: 0.14.3
-  epoch: 0
+  epoch: 1
   description: "Fast, async, stream-based link checker written in Rust"
   copyright:
     - license: Apache-2.0 AND MIT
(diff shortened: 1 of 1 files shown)


rawlingsj:
If I understand this correctly this would mean lychee would use it's own copy of openssl?  If there is a vulnerability in openssl then in wolfi we'd ensure that package is updated as fast as a remediation is available, though lychee would then not benefit from that and still be vulnerable.  

Can you install openssl in your multi-stage Dockerfile to resolve the original issue you are seeing?

kranurag7:
James, thanks for the review. I think it's not possible to have a static executable for lychee as of now. 

I tried adding openssl but that didn't work. 
I falled back to using curl for fetching the binary inside the image. 

A more detailed overview is here https://github.com/lycheeverse/lychee/discussions/1376 

closing this one and I'll open a PR in the future probably when it's possible.

PR #12992, opened 2024-02-14
Split off libecpg subpackages from postgres.

dlorenc:
This is split off in alpine, and some apps depend on it directly: https://github.com/apache/superset/blob/master/Dockerfile#L73C9-L73C16

### DIFF
--- a/postgresql-12.yaml
+++ b/postgresql-12.yaml
@@ -1,7 +1,7 @@
 package:
   name: postgresql-12
   version: "12.18"
-  epoch: 0
+  epoch: 1
   description: A sophisticated object-relational DBMS
   copyright:
     - license: BSD-3-Clause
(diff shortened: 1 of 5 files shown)


postgresql-12.yaml
```diff
@@ -53,6 +53,40 @@ pipeline:
...
 
 subpackages:
+  - name: libecpg-12
+    description: PostgreSQL libraries
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/usr/lib
+          mv ${{targets.destdir}}/usr/lib/libecpg.so* ${{targets.subpkgdir}}/usr/lib/
+    dependencies:
+      provider-priority: 12
```
imjasonh, line 63: This feels like it's doing something clever, can you explain it in a comment?

dlorenc: I copied it from a few other places in here: https://github.com/wolfi-dev/os/blob/main/postgresql-12.yaml#L81

PR #13026, opened 2024-02-14
Adding OpenJDK-11 to DataDog Agent to enable JMX

afeddersen:


joshrwolf:
should this be a runtime dependency? it seems like too much magic to me that this would work as a build dependency without any change to the build process

afeddersen:
> should this be a runtime dependency? it seems like too much magic to me that this would work as a build dependency without any change to the build process

Yes you are correct.  My GCP workstation is on the fritz so I'm going to close this PR and I have a new one opened here:  https://github.com/wolfi-dev/os/pull/13215.  Thanks!

PR #13095, opened 2024-02-15
new package: speedtest-go

imjasonh:


speedtest-go.yaml
```diff
@@ -0,0 +1,28 @@
...
+  epoch: 0
+  description: CLI and Go API to Test Internet Speed using speedtest.net
+  copyright:
+    - license: MIT
+
+pipeline:
+  - uses: go/install
+    with:
+      package: github.com/showwin/speedtest-go
+      version: v${{package.version}}
```
luhring, line 13: Can we do `git-checkout` and `go-build` instead to allow for gobumps when needed? 🤔

luhring (approved):

PR #13096, opened 2024-02-15
Add INTERACTIVE variable to Makefile for build debugging

jamonation:
Adds an `INTERACTIVE` variable to the Makefile, which gets added to `MELANGE_OPTS` - handy for interactive debugging while using `make package/thing` without a full melange invocation.


pdeslaur:
Using something like this also works nicely:

```
MELANGE_OPTS="--interactive" make package/kubeflow-jupyter-web-app
```

jamonation:
Ah I didn't make the connection between MELANGE_OPTS in the Makefile and evironment variables, thanks for the pointer, will start using that.

PR #13215, opened 2024-02-17
Adding openjdk-11 as a runtime dependency to support datadog-agent java jmx metrics

afeddersen:
- Bumping the version.
- Adding openjdk-11 as a runtime dependency to support java jmx metrics collection. 
- Adding a basic test.

More testing details in the support issue here: https://github.com/chainguard-dev/customer-issues/issues/1059#issuecomment-1949943943


datadog-agent.yaml
```diff
@@ -64,6 +65,7 @@ pipeline:
       go-version: "1.21.0"
 
   - runs: |
+      export PATH=$PATH:/usr/lib/jvm/java-11-openjdk/bin
```
dlorenc, line 68: I'm not sure this will do much... If we're only adding the jdk as a runtime dependency, it won't be present in the build environment so adding it to the path here isn't going to change anything in the build itself.

Do you intend for this to be set in the final image?

afeddersen: > I'm not sure this will do much... If we're only adding the jdk as a runtime dependency, it won't be present in the build environment so adding it to the path here isn't going to change anything in the build itself.
> 
> Do you intend for this to be set in the final image?

I'll remove it later today.  The history / testing details are over in customer issues:  https://github.com/chainguard-dev/customer-issues/issues/1059#issuecomment-1949943943

afeddersen: Removed it from here and I'll add it to the image.

```
environment:
  JAVA_HOME: /usr/lib/jvm/java-11-openjdk
```

mamccorm (approved):

PR #13305, opened 2024-02-19
chore(packages): Split libssp from gcc

EyeCantCU:
Fixes: This library functions standalone

#### For new package PRs only

- [ ] This PR is marked as fixing a pre-existing package request bug
  - [ ] Alternatively, the PR is marked as related to a pre-existing package request bug, such as a dependency
- [x] REQUIRED - The package is available under an OSI-approved or FSF-approved license
- [x] REQUIRED - The version of the package is still receiving security updates
- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`)

PR #13306, opened 2024-02-19
chore(avahi): Rebuild with libssp

EyeCantCU:
Fixes: Previously avahi brought in all of gcc when it only needed libssp


EyeCantCU:
Merge after libssp has been merged and built: https://github.com/wolfi-dev/os/pull/13305

EyeCantCU:
I just realized we don't need this :)

PR #13350, opened 2024-02-20
Use the wolfi signing key from the new location.

mattmoor:
ref: https://github.com/chainguard-dev/secrets/blob/main/wolfi-dev.tf


mattmoor:
switched to draft, it looks like we may still need the old GSA for the GCS interactions

imjasonh (approved):

PR #13395, opened 2024-02-21
bun/1.0.28 package version update

debasishbsws:
Fixes: #13319 

This issue was fixed in bun version `1.0.23` [upstream info](https://github.com/oven-sh/bun/issues/7617#issuecomment-1900013158). Here in the bun-bootstrap we are using the old version.

#### For version bump PRs

### DIFF
--- a/bun.yaml
+++ b/bun.yaml
@@ -1,6 +1,6 @@
 package:
   name: bun
-  version: 1.0.27
+  version: 1.0.28
   epoch: 0
   description: "Incredibly fast JavaScript runtime, bundler, test runner, and package manager - all in one"
   copyright:
@@ -40,7 +40,7 @@ pipeline:
   - uses: git-checkout
     with:
       repository: https://github.com/oven-sh/bun
-      expected-commit: c34bbb2e3fcaaadd8dfb5e75d545acee1c3cfd1b
+      expected-commit: 7056384702231f634e4e5d0f7c65965041314fc6
       tag: bun-v${{package.version}}
 
   # bun requires a specific zig to build
--- a/bun-bootstrap.yaml
+++ b/bun-bootstrap.yaml
@@ -1,6 +1,6 @@
 package:
   name: bun-bootstrap
-  version: 1.0.20
+  version: 1.0.23
   epoch: 0
   description: "Bun requires itself to bootstrap."
   copyright:
--- a/bun-bootstrap.yaml
+++ b/bun-bootstrap.yaml
@@ -17,7 +17,7 @@ environment:
 
 pipeline:
   - runs: |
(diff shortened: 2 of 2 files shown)


bun.yaml
```diff
@@ -12,7 +12,7 @@ environment:
       - autoconf
       - automake
       - bash
-      - bun-bootstrap # bun requires itself to build. this can get moved to our real build in the future
+      - bun # bun requires itself to build. this can get moved to our real build in the future
```
rawlingsj, line 15: Ah, CI is failing because we cant have a dependency on itself.  We need to be able to bootstrap wolfi so to do that we can't have this kind of cyclic dependency.

Looking at the CI failure on the original update PR I can see the error
```
2024/02/21 03:13:08 WARN bun install v1.0.20 (09d51486)
2024/02/21 03:13:30 WARN error: install script from "@napi-rs/canvas-linux-x64-musl" exited with 1
2024/02/21 03:13:30 WARN error: script "setup" exited with code 1
```
So it might be that for some reason the build for x86 thinks we use musl rather than glibc.  

I wonder if unsetting the LDFLAGS helps?  I.e. like this https://github.com/wolfi-dev/os/blob/4216178c11a50264d61da5930d4f5d26e9bf5b96/opensearch-dashboards-2.yaml#L68

If not we might need to see if we can create a patch to avoid the musl script being used.

Let me know if that makes sense.

FWIW there's a new interactive mode for melange which can help debug these kind of failures.  i.e.

```
ARCH=x86_64 MELANGE_EXTRA_OPTS="-i --rm --create-build-log" make package/bun
```
when the build fails you will be popped into a shell in the build container where you can play around running the build until it works.

debasishbsws: Thanks for the information, I am not sure but I may have seen long ago that rust depends on itself I am not sure, so I thought it would also work here 🤔. 
`unset LDFLAGS` doesn't work,
The problem is in the `bun`(version `1.0.20`) we are using to build the package in its [`default-trusted-dependencies`](https://github.com/oven-sh/bun/blob/09d51486e9ae219c261b3491776ea03fd6274dfc/src/install/default-trusted-dependencies.txt#L80-L85) it has some `musl` libraries that are not supported yet. those have been removed in version `1.0.23`. https://github.com/oven-sh/bun/commit/0bfe3760f6dd542603aab239bb31034171bb7191 

So to resolve this can we bump up the bun-bootstrap to version `1.0.23`?

debasishbsws: I understand now, it is just silly from my side to assume that the bun-bootstrap is fetching the specific version.
I thought as it is working it should be fetching that specific version.

rawlingsj (approved):

PR #13424, opened 2024-02-21
defualt to debug mode for tests

joshrwolf:
unlike for builds, running with `-x` is almost always desirable to see whats going on and what steps fail, this will also make debugging any CI failures easier too


imjasonh (approved):

vaikas (commented):
This seems good, I was wondering about if we should do this for the builds as well. However, I was curious if it leaves the CI hanging forever in case there are failures.

Could we maybe set up a 'CI' flag and then not set this in that case, or something like that (for both build / test)?

vaikas (approved):
Read wrong :) Was thinking, --interactive, as that's what I'm working on right now. Ignore me :)

PR #13586, opened 2024-02-23
python-3.11 - cherry pick upstream patch for CVE-2023-27043

smoser:
The change here patches addresses CVE-2023-27043 by backporting a 'strict' keyword argument to the email.getaddresses and email.parseaddr function.  The change may break some consumers as it does change the behavior.

After this change is landed, python 3.11 (and 3.12 via #13847) will behave like python master with regard to these two functions.  Existing code for 3.11 or 3.12 will not pass the 'strict' argument and the default is None.   The functions are modified to check for a None value and consult an environment variable for the desired behavior.

The environment variable PYTHON_EMAIL_STRICT_PARSING_DEFAULT will default will considered 'true' by default (unset or empty string). If you wish to enable the previous behavior, the the environment variable should be set to 'false'.

Please see comment below for background.

### DIFF
--- a/python-3.11.yaml
+++ b/python-3.11.yaml
@@ -1,7 +1,7 @@
 package:
   name: python-3.11
   version: 3.11.8
-  epoch: 0
+  epoch: 1
   description: "the Python programming language"
   copyright:
     - license: PSF-2.0
@@ -41,6 +41,14 @@ pipeline:
         Modules/_ctypes/darwin* \
         Modules/_ctypes/libffi*
 
+  - uses: patch
+    with:
+      patches: CVE-2023-27043.patch
+
+  - uses: patch
+    with:
+      patches: CVE-2023-27043-enable-disable.patch
+
   - name: Configure
     runs: |
       ./configure \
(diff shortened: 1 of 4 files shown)


dlorenc:
> c. Follow RHEL change, allowing environment variable or config file to change the default behavior.

This one gets my vote (just consider it one vote though)

ware:
> b. cherry-pick the upstream fix but turn 'strict' off by default. Users won't be broken, but will have to 'opt-in' through code change to the remediation.
> c. Follow RHEL change, allowing environment variable or config file to change the default behavior.

I think either one of these two solutions would be acceptable.

xnox:
Debian codesearch for getaddress in python filetypes

https://codesearch.debian.net/search?q=getaddress+filetype%3Apython&literal=1&perpkg=1

```
$ curl -s https://codesearch.debian.net/results/8ee7eac86202d906/packages.txt | wc
     45
```

Gives me 45 packages. I can cross check by hand, how many of these are in any of our packages or images. Assuming the scope is small, and because our goal is to provide security to people, my preference is to

- ship updated change
- default to secure
- provide opt-out for 3-rd party customer code compat

As the point of chainguard is to ship security updates without friction, meaning no opt-in needed.

smoser:
Thanks @xnox.  Your prompting made me read a bit more. the RHEL change opts into the new behavior by default and allows the user a code-change-free option to opt out.

I like this approach, and think we should go with that.

dlorenc:
+1 to this, hardened by default.

smoser:
I think this is ready to land, would appreciate a review. thanks.

xnox (approved):

PR #13605, opened 2024-02-23
Use pombump to handle spark dep bumps.

vaikas:
Switch to using pombump. Has less CVEs, but they can not be currently removed. 
Trying to bump `libthrift-0.12.0.jar` to 13 or later does not compile because API has changed.
`jackson-mapper-asl-1.9.13` has not been updated since 2013.
`hadoop-client-runtime-3.3.6` is the latest
Others are shaded jars.

Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches


spark/bump-packages.patch
jamonation, line 1: I think this patch file can disappear - I don't see it used anywhere now 🎉

spark.yaml
```diff
@@ -48,7 +49,13 @@ pipeline:
     with:
       patches: make-distribution.patch bump-packages.patch
```
jamonation, line 50:
```suggestion
      patches: make-distribution.patch
```

Same as comment below, I don't think? the bump-packages.patch gets used anywhere now with pombump doing the driving

jamonation (approved):

spark.yaml
vaikas: It does get used, there are some other kind of patching it does. For example:

```
+              <excludes>
+                <exclude>com.google.common.util.concurrent.internal.**</exclude>
+              </excludes>
```

vaikas: We can add more capabilities into pombump as we go along.
But the above file also patches a `.scala` file.

vaikas (commented):
There are still some other patches that need to get applied.

PR #13616, opened 2024-02-23
generate a few package tests

joshrwolf:
all of these were mostly (99%) generated with `gptscript` and the following:

https://gist.github.com/joshrwolf/a197145cfdccb73d2bbe4a780484a9a2

### DIFF
--- a/go-1.22.yaml
+++ b/go-1.22.yaml
@@ -27,7 +27,7 @@ environment:
 var-transforms:
   - from: ${{package.version}}
     match: \_
-    replace: ''
+    replace: ""
     to: mangled-package-version
 
 pipeline:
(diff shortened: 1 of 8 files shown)


go-1.22.yaml
```diff
@@ -92,3 +92,19 @@ update:
     strip-prefix: go
     tag-filter: go1.22
     use-tag: true
+
+test:
+  environment:
+    contents:
+      packages:
+        - busybox
+        - go-1.22
```
imjasonh, line 101: We don't need this explicitly

postgresql-16.yaml
```diff
@@ -143,3 +143,33 @@ update:
     strip-prefix: REL_
     tag-filter: REL_16
     use-tag: true
+
+test:
+  environment:
+    contents:
+      packages:
+        - wolfi-base
+        - postgresql-16
```
imjasonh, line 152: Same

imjasonh (dismissed):
I think we need some lint checks, but the tests lgtm

go-1.22.yaml
joshrwolf: remnants courtesy of 🤖

go-1.22.yaml
```diff
@@ -92,3 +92,18 @@ update:
...
+
+test:
+  environment:
+    contents:
+      packages:
+        - busybox
+  pipeline:
+    - name: Verify Go version
+      runs: |
+        go version | grep "go1.22"
```
ajayk, line 104:
```suggestion
        go version | grep "go${{package.version}}"
```

ajayk: Most of these are on auto version streams these will cause breakages in the test when we move to 1.23 automatically

PR #13645, opened 2024-02-24
Add 7 Python package tests - #1

afeddersen:
Automated Python package tests created by [auto-package-test](https://github.com/chainguard-dev/quality-squad/tree/main/scripts/auto-package-test).

Note:

1. These will eventually be cleaned up and moved to a dedicated pipeline once bubblewrap and the pipeline is reliable.
2. We also looked at creating these as a script some packages don't have a directory.

The top priority is to improve our coverage and we can come back through with automation to clean these up.

### DIFF
--- a/py3-build.yaml
+++ b/py3-build.yaml
@@ -44,3 +44,22 @@ update:
     identifier: pypa/build
     use-tag: true
     strip-suffix: .post1
+
+test:
+  environment:
+    contents:
+      packages:
+        - busybox
+  pipeline:
+    - runs: |
+        LIBRARY="build"
+        IMPORT_STATEMENT="import build"
+
+        if ! python -c "$IMPORT_STATEMENT"; then
+            echo "Failed to import library '$LIBRARY'."
+            python -c "$IMPORT_STATEMENT" 2>&1
+            exit 1
+        else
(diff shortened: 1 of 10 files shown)


py3-build.yaml
```diff
@@ -44,3 +44,22 @@ update:
...
+        IMPORT_STATEMENT="import build"
+
+        if ! python -c "$IMPORT_STATEMENT"; then
+            echo "Failed to import library '$LIBRARY'."
+            python -c "$IMPORT_STATEMENT" 2>&1
+            exit 1
+        else
+            echo "Library '$LIBRARY' is installed and can be imported successfully."
+            exit 0
+        fi
```
xnox, line 65: Can we factor this out to be similar to "git-checkout" named pipeline?

something like
```yaml
- uses: py-import
   imports: [....] # list of imports implied ${{package.name}}, without the py3 prefix?
```

Cause Ideally, we would want to factor out the common code like this. Such that we can improve and extend it everywhere at once. Similarly for the steps to `build` python package too.

xnox: but i see this is mentioned already in the pull request description. Never mind.

ajayk (approved):

PR #13696, opened 2024-02-26
lint: default wolfi-base in tests

imjasonh:
Also add a lint check that removes it from configs and removes newly empty blocks.

Supercedes https://github.com/wolfi-dev/os/pull/13330

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,7 @@ MELANGE_TEST_OPTS += --arch ${ARCH}
 MELANGE_TEST_OPTS += --pipeline-dirs ./pipelines/
 MELANGE_TEST_OPTS += --repository-append https://packages.wolfi.dev/os
 MELANGE_TEST_OPTS += --keyring-append https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
+MELANGE_TEST_OPTS += --test-package-append wolfi-base
 MELANGE_TEST_OPTS += --debug
 MELANGE_TEST_OPTS += ${MELANGE_EXTRA_OPTS}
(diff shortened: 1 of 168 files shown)


rawlingsj:
Just a thought on this and there maybe other motivations which make my thought irrelevant but...  sometimes for downstream consumers of packages it can be hard to know what the minimal set of dependencies are. I.e folks should use a `depends` block to include things like bash, busybox etc so if we always install wolfi-base in the test environment, it means we're not testing the depends section if a package requires a shell. So we could end up testing a package with extra default packages that are not installed by default when adding a package.

imjasonh:
> Just a thought on this and there maybe other motivations which make my thought irrelevant but... sometimes for downstream consumers of packages it can be hard to know what the minimal set of dependencies are. I.e folks should use a `depends` block to include things like bash, busybox etc so if we always install wolfi-base in the test environment, it means we're not testing the depends section if a package requires a shell. So we could end up testing a package with extra default packages that are not installed by default when adding a package.

The problem is that _all_ packages _must_ have a shell in order to work at all, because of how melange runs steps. Even if you're just doing `tool command`, melange runs it as `sh -c "tool command"`, so you'll never be able to _not_ have a shell dependency.

rawlingsj (approved):

rawlingsj:
Ah gotcha - SGTM!

mattmoor (approved):

imjasonh:
Merging past the tests because this is a purely mechanical cleanup change, and unrelated to pre-existing test failures.

PR #13765, opened 2024-02-27
vault-1.14: version update and add back the now missing `go mod tidy`

xnox:
Previously vault-1.14 had a go/bump pipeline, which has a side effect of calling `go mod tidy`. This step appears to be required, as the build fails with go toolchain requesting to perform `go mod tidy` prior to being able to complete the `go generate` steps.

Add back `go mod tidy` as the first pipeline action. It's either harmless, or unbreaks the build.

Fixes: #12121


xnox:
hm, maybe i should just propose this into main branch, because otherwise no actions are executed.

rawlingsj (approved):

PR #13794, opened 2024-02-27
Add version test to grafana-agent-operator

jamonation:
Related: #13440

Adds a test to check `grafana-agent-operator` version info.

### DIFF
--- a/grafana-agent-operator.yaml
+++ b/grafana-agent-operator.yaml
@@ -1,7 +1,7 @@
 package:
   name: grafana-agent-operator
   version: 0.40.0
-  epoch: 1
+  epoch: 0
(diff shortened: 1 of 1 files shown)


grafana-agent-operator.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: grafana-agent-operator
   version: 0.40.0
-  epoch: 0
+  epoch: 1
```
ajayk, line 4:
```suggestion
  epoch: 0
```

ajayk: Test changes don't need epoch bumps

ajayk (approved):

grafana-agent-operator.yaml
jamonation: gotcha, thanks for pointing that out, I was just cargo-cult bumping to be sure I got things to build 👍

PR #13817, opened 2024-02-28
chore(dask-gateway): Install inside of a virtual environment

EyeCantCU:
Installs dask-gateway and dask-gateway-server into a venv

### DIFF
--- a/dask-gateway.yaml
+++ b/dask-gateway.yaml
@@ -100,9 +100,43 @@ subpackages:
           mkdir -p ${{targets.subpkgdir}}/usr/bin/
           ln -s /usr/share/${{package.name}}-server/.venv/bin/${{package.name}}-jobqueue-launcher ${{targets.subpkgdir}}/usr/bin/${{package.name}}-jobqueue-launcher
           ln -s /usr/share/${{package.name}}-server/.venv/bin/${{package.name}}-server ${{targets.subpkgdir}}/usr/bin/${{package.name}}-server
+    test:
+      environment:
+        contents:
+          packages:
+            - wolfi-base
+      pipeline:
+        - runs: |
+            # Version check, also ensures executable paths are correct
+            dask-gateway-server --version
+
+            # Test imports in virtual environment
+            source /usr/share/${{package.name}}-server/.venv/bin/activate
+            python -c "import dask_gateway_server"
+            python -c "import kubernetes_asyncio"
+            python -c "import sqlalchemy"
+            python -c "import typing_extensions"
 
 update:
   enabled: true
   github:
     identifier: dask/dask-gateway
     use-tag: true
+
+test:
+  environment:
+    contents:
+      packages:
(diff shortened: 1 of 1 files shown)


dask-gateway.yaml
```diff
@@ -1,13 +1,19 @@
...
+  options:
+    # We create a dependency on libpython even though we provide
+    # libpython in the virtual environment. This prevents python
+    # versions on the host from being swapped out. Enabling no-
+    # depends works around this
+    no-depends: true
   dependencies:
     runtime:
-      - python3
+      - python-3
```
mamccorm, line 16: [non-blocking] i'd generally align to using 'python3' here, which is more consistent with the other package names. Both are an option and reference the same package

dask-gateway.yaml
```diff
@@ -38,31 +39,67 @@ pipeline:
...
+      ln -s /usr/share/${{package.name}}/.venv/bin/dask-worker ${{targets.destdir}}/usr/bin/dask-worker
 
 subpackages:
   - name: dask-gateway-server
     description: A multi-tenant server for securely deploying and managing Dask clusters
+    options:
+      no-depends: true
+    dependencies:
+      runtime:
+        - python-3
```
mamccorm, line 74: [non-blocking] Same as above

dask-gateway.yaml
```diff
@@ -38,31 +39,67 @@ pipeline:
...
+          # Fix venv paths
+          sed -i "s|/home/build|/usr/share|g" ${{targets.subpkgdir}}/usr/share/${{package.name}}-server/.venv/bin/*
+
+          # Include system site-packages
+          sed -i "s|include-system-site-packages = false|include-system-site-packages = true|g" ${{targets.subpkgdir}}/usr/share/${{package.name}}-server/.venv/pyvenv.cfg
+
+          # Symlink scripts to PATH
+          mkdir -p ${{targets.subpkgdir}}/usr/bin/
+          ln -s /usr/share/${{package.name}}-server/.venv/bin/${{package.name}}-jobqueue-launcher ${{targets.subpkgdir}}/usr/bin/${{package.name}}-jobqueue-launcher
+          ln -s /usr/share/${{package.name}}-server/.venv/bin/${{package.name}}-server ${{targets.subpkgdir}}/usr/bin/${{package.name}}-server
```
mamccorm, line 102: Could we look at adding some melange tests? especially given we've changed the packaging format

dask-gateway.yaml
EyeCantCU: Thank you. I'll get that switched up :)

dask-gateway.yaml
EyeCantCU: Yeah. I can do that

EyeCantCU: Added :)

mamccorm (approved):

PR #13931, opened 2024-03-01
update the bitnami paths to debian 12

found-it:
[Last week the bitnami repo dropped all references to debian-11](https://github.com/bitnami/containers/commit/d624c10ea7e23d8059dc62adb11e1414d8e39123) in favor of debian-12. This PR points the bitnami/compat pipeline to the correct directory for all packages pointed towards debian-11.

It also adds a check to the bitnami/compat pipeline to check if the directory exists so this doesn't silently fail again.

PR #13932, opened 2024-03-01
Update redis-sentinel compat bitnami path

found-it:
Last package missed in this PR https://github.com/wolfi-dev/os/pull/13931


redis-sentinel-6.2-compat.yaml
ajayk, line 1: needs epoch bump

found-it: You're too quick Ajay 😁

found-it: Done

ajayk (approved):

PR #13961, opened 2024-03-01
optimising hugo-extended to use already available hugo

debasishbsws:
Previously in hugo-extended, we were installing hugo directly from GitHub.
As we are already building and have hugo in the repository it was repetitive.

### DIFF
--- a/hugo-extended.yaml
+++ b/hugo-extended.yaml
@@ -1,32 +1,25 @@
 package:
   name: hugo-extended
   version: 0.123.7
-  epoch: 0
-  description: The world's fastest framework for building websites.
+  epoch: 1
+  description: The world's fastest framework for building websites (with extended SASS libraries).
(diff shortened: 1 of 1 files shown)


hugo-extended.yaml
```diff
@@ -1,32 +1,25 @@
...
-      - ca-certificates-bundle
-      - git
+      - wolfi-baselayout
 
 pipeline:
-  - uses: go/install
-    with:
-      package: github.com/gohugoio/hugo
-      version: v${{package.version}}
-      tags: extended
```
ajayk, line 24: we are building with  tags: extended , in the `hugo` package we just build the standard versions 

go install -tags extended github.com/gohugoio/hugo@latest

https://github.com/gohugoio/hugo/blob/master/README.md#build-from-source

PR #13971, opened 2024-03-01
update thollander/actions-comment-pull-request to version using node 20

imjasonh:


ajayk (approved):

xnox:
how come ABI compat check is both required, and skipped? can it be regressing semaphore - as in skipped counts as pass for required?

rawlingsj:
> how come ABI compat check is both required, and skipped? can it be regressing semaphore - as in skipped counts as pass for required?

they're skipped because there are no package changes in this PR which means no apks were built so the ABI and Scan checks are skipped.

rawlingsj (approved):

PR #14041, opened 2024-03-04
adds a patch to use sha256 instead of md5

mritunjaysharma394:
The existing upstream of opensearch-dashboard uses md5, this PR adds a patch to use more secure sha256 instead

### DIFF
--- a/opensearch-dashboards-2.yaml
+++ b/opensearch-dashboards-2.yaml
@@ -55,6 +55,10 @@ pipeline:
       tag: ${{package.version}}
       expected-commit: 989d8f41f37cca3275bf3fedc5c2057a717d1d64
 
+  - uses: patch
+    with:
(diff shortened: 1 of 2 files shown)


rawlingsj (changes requested):
This will change the expected default behaviour of opensearch dashboards, let's pause on merging this for now.

xnox:
@mritunjaysharma394 This looks great. Can you please submit this upstream? Or do you want me to do this? Cause if it is upstream, we don't need to maintain it ourselves downstream.

mritunjaysharma394:
Created PR and an issue in upstream for this 

https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6032
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6031

xnox:
See review & feedback upstream.

PR #14133, opened 2024-03-05
Libxcrypt transition

xnox:
glibc,libxcrypt: move libcrypt1 package from glibc to libxcrypt

This will be required with next glibc release, but was possible to do
for a while now. All other distros have transitioned to using
libxcrypt already. I have double checked the abi compatibility with
abi-compliance-checker. Separately, once this lands, we might want to
bump epoch on all the reverse depends for them to gain stronger crypt
support.

Also note previous yaml for libxcrypt had a mistake and didn't use any custom provided options.

(See
https://apk.dag.dev/https/packages.wolfi.dev/os/aarch64/APKINDEX.tar.gz@etag:2edafbb750de4d6487204c1cd6110c7b/APKINDEX?&depend=so%3Alibcrypt.so.1
for packages to bump).

Fixes:

Unblocks: #12226

### DIFF
--- a/glibc.yaml
+++ b/glibc.yaml
@@ -1,7 +1,7 @@
 package:
   name: glibc
   version: 2.38
-  epoch: 11
+  epoch: 12
   description: "the GNU C library"
   copyright:
     - license: GPL-3.0-or-later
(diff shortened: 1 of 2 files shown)


ajayk (approved):

xnox:
waiting for things to settle before merging this.

xnox:
here be dragons - if this is all good, we should consider to do epoch bump on all libcrypt1 deps.

PR #14209, opened 2024-03-06
coreutils: use openssl & split legacy tools

xnox:
md5sum & sha1sum are obsolete and should not be used anymore. Split them into separate subpackage.

Enable OpenSSL backend for sha256sum and friends. This gains hardware acceleration, reduces coreutils binary size, and opens the path to FIPS compliant coreutils - when paired with OpenSSl in FIPS mode.

### DIFF
--- a/coreutils.yaml
+++ b/coreutils.yaml
@@ -1,7 +1,7 @@
 package:
   name: coreutils
   version: "9.4"
-  epoch: 4
+  epoch: 5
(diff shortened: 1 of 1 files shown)


imjasonh:
LGTM but I think I'm lacking context, and I'd like @smoser's 👍

smoser (approved):
:+1:

PR #14250, opened 2024-03-07
pgvectors for each postgres version

developer-guy:
I noticed that according to [here](https://github.com/pgvector/pgvector#postgres-location) we should compile pgvector for each postgres version, also I confirmed that during the tests of the pgvector extension with postgres and saw the following errors:

```bash
Server is version 15, library is version 16.
```

This PR aims to fix this issue.

I couldn't test it in the image side because docker gave the this error:

``` 
 docker: failed to register layer: lsetxattr com.docker.grpcfuse.ownership
│ /usr/include/postgresql/server/extension/vector: operation not supported.
```

Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/pgvector-14.yaml
+++ b/pgvector-14.yaml
@@ -1,5 +1,7 @@
+# pgvector-14 is intended to support PostgreSQL 14.
+# https://github.com/pgvector/pgvector#installation-notes
 package:
-  name: pgvector
+  name: pgvector-14
   version: 0.6.1
   epoch: 0
   description: Open-source vector similarity search for PostgreSQL
@@ -13,7 +15,7 @@ environment:
       - automake
       - build-base
       - busybox
-      - postgresql-dev
+      - postgresql-14-dev
 
 pipeline:
   - uses: git-checkout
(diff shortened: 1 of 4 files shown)


EyeCantCU (dismissed):

pgvector-14.yaml
```diff
@@ -1,5 +1,5 @@
 package:
-  name: pgvector
+  name: pgvector-14
```
mamccorm, line 2: Any more info here? I see in the description we talk about needing to compile against different postgres versions but i'm not sure I understand the change.

looks like we'll break any dependent images / packages that used 'pgvector' as the package name, and don't see any package withdraws or provides?

pgvector-14.yaml
```diff
@@ -1,5 +1,5 @@
 package:
```
mamccorm, line 1: Similar to pgaudit, can we add a comment to each of these melanges summarising? i.e from pgaudit:

```
# pgAudit v1.6.X is intended to support PostgreSQL 14.
# https://github.com/pgaudit/pgaudit/blob/master/README.md#postgresql-version-compatibility
```

developer-guy: added.

mamccorm (approved):

PR #14268, opened 2024-03-07
Do not remove pyc files from python-3.12 and python-3.11

smoser:
This will absolutely increase size of the package, but will improve performance.

Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/python-3.11.yaml
+++ b/python-3.11.yaml
@@ -1,7 +1,7 @@
 package:
   name: python-3.11
   version: 3.11.8
-  epoch: 2
+  epoch: 3
   description: "the Python programming language"
   copyright:
     - license: PSF-2.0
@@ -70,8 +70,6 @@ pipeline:
   - uses: autoconf/make-install
 
   - runs: |
-      find ${{targets.destdir}}/usr/lib -type f -name '*.pyc' -exec rm -rf '{}' +
-      find ${{targets.destdir}}/usr/lib -type f -name '*.pyo' -exec rm -rf '{}' +
       find ${{targets.destdir}}/usr/lib -type f -name 'libpython*.a' -exec rm -rf '{}' +
       find ${{targets.destdir}}/usr/lib -type d -name 'test' -exec rm -rf '{}' +
       find ${{targets.destdir}}/usr/lib -type d -name 'tests' -exec rm -rf '{}' +
@@ -80,6 +78,14 @@ pipeline:
       cd ${{targets.destdir}}/usr/bin
       rm -f python3 pydoc3 idle3* 2to3*
       rm ${{targets.destdir}}/usr/lib/libpython3.so
+
+  - runs: |
+      find ${{targets.destdir}}/usr/lib -type f -name '*.pyc' -exec rm -rf '{}' +
+      find ${{targets.destdir}}/usr/lib -type f -name '*.pyo' -exec rm -rf '{}' +
+
+      export LD_LIBRARY_PATH="${{targets.destdir}}/usr/lib${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}"
+      ${{targets.destdir}}/usr/bin/python3.11 -m compileall --invalidation-mode=unchecked-hash \
+          -r100 ${{targets.destdir}}/usr/lib
 
   - uses: strip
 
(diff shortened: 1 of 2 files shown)


imjasonh:
> This will absolutely increase size of the package, but will improve performance.

Can you quantify either or both of the size increase and the performance improvement?

I believe that this is worth doing, but it'd be nice to have numbers to reference.

smoser:
For information, This is the growth in python3.12 due to .pyc files:

```
# find /usr/lib/python3.12/ -name "*.py" | xargs ls -l | awk '{tot += $5}; END { print tot/1024 }'
11324.7

# find /usr/lib/python3.12/ -name "*.pyc" | xargs ls -l | awk '{tot += $5}; END { print tot/1024 }'
11943.8

# find /usr/lib/python3.12/ -name "*.pyc" | wc -l
686

# find /usr/lib/python3.12/ -name "*.py" | wc -l
686
```

So cost of .pyc files includes is roughly the size of the size of .py files
(.pyc files in general are close to the same size as .py file they came from).

smoser:
fwiw, i hadn't seen @imjasonh 's question before i went to put more info here. so I don't consider previous comment on the size increase to answer the performance question.  Google seems not to know much about this, and I don't have a good way to get realistic performance numbers.

There is some info [here](https://peps.python.org/pep-3147/)

> CPython compiles its source code into “byte code”, and for performance reasons, it caches this byte code on the file system whenever the source file has changes. This makes loading of Python modules much faster because the compilation phase can be bypassed. When your source file is foo.py, CPython caches the byte code in a foo.pyc file right next to the source.

It is important to point out that Debian/Ubuntu/Fedora/RHEL all *do* run with pyc files, they just do not ship them in their .deb or .rpm files.  The .pyc files are compiled at install time.

We could potentially do that (change to install-time generation of .pyc files), but it would not have any difference in the end -- the image would still be the same.

The path we're in right now, the image does not have the .pyc files and the filesystem is not writable by the python process, so the python process incurs the cost at start up every time.

smoser:
So here is a bit of info, just to provide something. 

this is in a local-wolfi with python3.12 installed.  As root, I removed all .pyc files:

```
## remove all .pyc files
## run a simple python program (json.tool with -h), which should show --help output
% find /usr/lib/python3.12/ -name "__pycache__" | xargs rm -Rf
% time python3.12 -m json.tool -h >/dev/null
real    0m 0.14s
user    0m 0.13s
sys     0m 0.01s

% find /usr/lib/python3.12/ -name "__pycache__" | xargs rm -Rf 
% time python3.12 -m json.tool -h >/dev/null
real    0m 0.14s
user    0m 0.13s
sys     0m 0.01s

% find /usr/lib/python3.12/ -name "__pycache__" | xargs rm -Rf 
% time python3.12 -m json.tool -h >/dev/null
real    0m 0.14s
user    0m 0.13s
sys     0m 0.00s

% find /usr/lib/python3.12/ -name "__pycache__" | xargs rm -Rf
% time python3.12 -m json.tool -h >/dev/null
real    0m 0.14s
user    0m 0.13s
sys     0m 0.01s

## now we have some .pyc files
% find /usr/lib/python3.12/ -name "*.pyc" | wc -l
37

## run it again without removing the .pyc files
## what changes - its faster now (consistently)
% time python3.12 -m json.tool -h >/dev/null
real    0m 0.03s
user    0m 0.02s
sys     0m 0.00s
% time python3.12 -m json.tool -h >/dev/null
real    0m 0.03s
user    0m 0.02s
sys     0m 0.00s
% time python3.12 -m json.tool -h >/dev/null
real    0m 0.03s
user    0m 0.02s
sys     0m 0.00s


## Now as non-root, so the patsh are not writable
% find /usr/lib/python3.12/ -name "__pycache__" | xargs rm -Rf
% sudo -u nobody /bin/sh
$ time python3.12 -m json.tool -h >/dev/null
real    0m 0.14s
user    0m 0.13s
sys     0m 0.00s
$ time python3.12 -m json.tool -h >/dev/null
real    0m 0.14s
user    0m 0.13s
sys     0m 0.00s
$ time python3.12 -m json.tool -h >/dev/null
real    0m 0.14s
user    0m 0.13s
sys     0m 0.01s

```

That _is_ startup cost, that is incurred just once, so longer running programs would not be negatively affected.  I'm not sure if there are memory footprint changes due to not reading the cached file.

imjasonh (approved):

PR #14283, opened 2024-03-07
Use wolfictl build for postsubmit builds

jonjohnsonjr:
This gives us concurrent package builds, which speeds things up a bit.

We can revert this if things don't work as expected, but I've also moved the current definition to build-old.yaml in case we want to manually invoke it down the line. We may want to remove that once we've decided this will stick.

### DIFF
--- a/.github/workflows/build-old.yaml
+++ b/.github/workflows/build-old.yaml
@@ -1,4 +1,4 @@
-name: Build Wolfi OS
+name: Build Wolfi OS with make all
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-old.yaml
```diff
@@ -0,0 +1,323 @@
+name: Build Wolfi OS with make all
+
+on:
+  push:
+    branches: ['main']
```
imjasonh, line 5: I think we should comment this out, so we don't build everything twice, right?

jonjohnsonjr: Indeed!

imjasonh (approved):

PR #14443, opened 2024-03-08
fix(openblas): fix update block

stormqueen1990:
This package was missing an `enabled: true` and a `strip-prefix: v` to be able to correctly auto-update.


xnox (approved):
LGTM

xnox:
Well that works but I guess we should update URLs to new ones, and actually bump the upgrade.

I'll look into that after coffee.

PR #14456, opened 2024-03-08
trino: Patch CVEs

pdeslaur:
Transfers old CVE fixes to `pombump` and introduce new ones.


xnox:
I like the pombump syntax, but i with there would be a notes field as to why a given thing is being bumped.
I.e. postgresql pom bump is to address [CVE-2024-1597](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2024-1597)

Cause it is not clear if the pombump is to resolve a failure to build from source; compat with wolfi / our java / etc; or if it is indeed a high CVE fix.

But yeap, this looks great.

xnox (dismissed):

pdeslaur:
Agreed! I added comments

cpanato (approved):

Issue #13623, opened 2024-02-24
Expand package testing coverage


joshrwolf:
With the introduction of `melange test` fairly recently, we have a backlog of packages that we would gain to backfill at least _some_ tests.

This issue serves as a tracking issue for attempting to keep a list of an effort to accomplish backfilling this will 🤖s.

mamccorm:
https://github.com/wolfi-dev/os/pull/25715#discussion_r1706151034

jonjohnsonjr:
We did it?

PR #14496, opened 2024-03-09
add postgresql-{} tests

joshrwolf:
Add a couple postgresql tests for all the versions

ref: #13623


joshrwolf:
tests appear to be failing in CI but not remotely. I have a sneaky suspicion this is a bwrap vs docker thing 🤔 

```
sudo-rs: cannot execute '/usr/bin/initdb': Permission denied (os error 13)
```

joshrwolf:
superceded by #14707

PR #14537, opened 2024-03-10
feat(packages): Produce auditable rust binaries

EyeCantCU:
Leverages cargo-auditable to produce rust binaries with an embedded dependency tree in a dedicated linker section of the produced binaries containing crate versions in JSON. Useful for tracking bugs and vulnerabilities impacting crates

NOTE: Increases binary size by less than 4kB

### DIFF
--- a/atuin.yaml
+++ b/atuin.yaml
@@ -1,7 +1,7 @@
 package:
   name: atuin
   version: 18.0.2
-  epoch: 0
+  epoch: 1
   description: Magical shell history
   copyright:
     - license: MIT
@@ -12,6 +12,7 @@ environment:
       - build-base
       - busybox
       - ca-certificates-bundle
+      - cargo-auditable
       - rust
 
 pipeline:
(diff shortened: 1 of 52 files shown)


atuin.yaml
```diff
@@ -22,7 +23,7 @@ pipeline:
       expected-commit: a78aaa78e487b2499ffd7eed86bac15aa3df0960
 
   - runs: |
-      cargo build --locked --release
+      cargo auditable build --locked --release
```
imjasonh, line 26: What can we do to encourage new rust builds to use this instead of `cargo build`? This is great but keeping it going will depend on folks knowing this is an option.

EyeCantCU: I think there may be room here to standardize our rust package builds in a melange pipeline. I can look into that shortly. Otherwise, a nudge somewhere like Chainguard Academy with code guidelines may also be helpful

PR #14655, opened 2024-03-10
php-8.3-msgpack: init at 2.2.0

shyim:
It's just php-8.2-msgpack which is already packaged for PHP 8.3

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/php-8.3-msgpack.yaml
+++ b/php-8.3-msgpack.yaml
@@ -1,0 +1,67 @@
+package:
+  name: php-8.3-msgpack
+  version: 2.2.0
+  epoch: 0
+  description: "A PHP extension for msgpack"
+  copyright:
+    - license: BSD-3-Clause
+  dependencies:
+    runtime:
+      - ${{package.name}}-config
+      - php-8.3
+    provides:
+      - php-msgpack=${{package.full-version}}
+
+environment:
+  contents:
+    packages:
+      - autoconf
+      - build-base
+      - busybox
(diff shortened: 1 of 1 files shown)


shyim:
@ajayk is here anything I can do to make the reviews easier? kinda like merge all the php related PRs into one 🤔 

Or let me know if you don‘t want to have them here. I would be also fine, no problem :)

ajayk:
@shyim still the [php-8.3](https://github.com/wolfi-dev/os/blob/main/php-8.3.yaml#L9-L11) is not the default version of php its set to [php-8.2](https://github.com/wolfi-dev/os/blob/main/php-8.2.yaml#L9-L10)

so anybody using php-msgpack would get the latest version of php-8.3-msgpack  instead of getting php-8.2-msgpack , once we switch the default provides this can all be merged 

@vaikas : can you please take a look

shyim:
Maybe we should go rather and remove the provides of all php packages? The users should explicitly choose always their PHP versions.  The exact same is also already with Redis why is that there not a problem 🤔 

- https://github.com/wolfi-dev/os/blob/main/php-8.1-redis.yaml
- https://github.com/wolfi-dev/os/blob/main/php-8.2-redis.yaml
- https://github.com/wolfi-dev/os/blob/main/php-8.3-redis.yaml

ajayk (approved):

PR #14661, opened 2024-03-10
Move runtime deps from image to package, use git checkout and add tests

mamccorm:
- Moved some runtime dependencies defined at image level, to package level (No new dependencies where introduced)
- Switch to using git checkout
- Added tests
- Added new package to fetch entrypoint script, and compat package
- Use JDK11 as buildtime dep, lining up with upstream(compatible with JDK11, 17 or 21 at runtime)


jenkins.yaml
```diff
@@ -1,13 +1,21 @@
...
     - license: MIT
   dependencies:
     runtime:
+      # Jenkins requires both bash and busybox.
+      - bash
+      - busybox
+      - coreutils
+      - glibc-locale-en
+      - openjdk-17-default-jvm
+      - openjdk-17-jre
```
ajayk, line 16: This prevents us from running in higher JRE environments

ajayk: I would wait for @stormqueen1990 and @rawlingsj on this

mamccorm: Jenkins supports multiple Java versions - I can leave these specific settings in as image runtime dependencies, but there might be tweaks needed to the melange for the particular java version and they'd essentially be different version streams?

mamccorm: Jenkins supports multiple Java versions at any given time. I can leave these specific settings in as image runtime dependencies, but there might be tweaks needed to the melange for the particular java version builds, and they'd essentially be different version streams.

Also passing them in at the image level would still mean it'd be compiled with a different JDK version. i.e at the minute JDK17, and run with whatever is passed in at the image level.

With that in mind it seems cleaner to define them in here and split to separate melange if we need to support concurrent JDK version builds / runtimes per release - was my thinking here

ajayk: Jenkins is compiled with Java 11 and run with different JRE Versions 11,17,21, please see https://hub.docker.com/r/jenkins/jenkins/tags

ajayk: Splitting of melange files is not needed here , instead compelling with JDK11 would be good instead

mamccorm: Thanks @ajayk - in that case i'll flip over to jdk11 for buildtime and drop the runtimes from melange

mamccorm: Applied the suggested change

developer-guy (dismissed):
lgtm

jenkins.yaml
```diff
@@ -22,26 +28,27 @@ environment:
       - git
       - glibc-locale-en
       - maven
-      - openjdk-17
-      - openjdk-17-default-jvm
+      - openjdk-11
```
mamccorm, line 31: Intentional change - mirrors upstream, compiles with JDK11 but supports runtime of JDK11, JDK17, or JDK21.

rawlingsj (approved):

PR #14679, opened 2024-03-11
add timeout for scan ci step

cpanato:
the job should run pretty fast even for bigger batches

adding a timeout of 1 hour in case the job get stuck

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -236,6 +236,8 @@ jobs:
     needs: build
     if: needs.build.outputs.packages_were_built == 'true'
 
+    timeout-minutes: 30
+
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-build.yaml
```diff
@@ -236,6 +236,8 @@ jobs:
     needs: build
     if: needs.build.outputs.packages_were_built == 'true'
 
+    timeout-minutes: 60
```
mattmoor, line 239: This seems high?

mattmoor (dismissed):

.github/workflows/ci-build.yaml
cpanato: yes, I was wondering that when I open the PR, maybe 30 min max?

mattmoor (approved):

PR #14691, opened 2024-03-11
withdraw-packages: Withdraw obsolete python3

xnox:
Foundations is working on providing co-installable multiple versions of python (i.e. python-3.11 and python-3.12), such that one can generate wheels / venv of the desired major python version. And we are also working on providing a way to choose which python is default (python-3.11-default, which makes python3 point at python3.11, ditto python-3.12-default, etc).

I have noticed that in the past there was just the one and only python3, and it is still published as installable.

Upon package renames and splits, they should really be unpublished, as otherwise they remain installable and vulnerable.

Fixes: https://github.com/wolfi-dev/os/commit/635a3c353cbfb672bb23ba123ee79bc4dea06bb7

CC: @rawlingsj @smoser @ajayk


smoser (dismissed):

withdrawn-packages.txt
```diff
@@ -1 +1,4 @@
-pgvector-0.6.1-r0.apk
\ No newline at end of file
+pgvector-0.6.1-r0.apk
+python3-3.11.1-r3
```
ajayk, line 2:
```suggestion
python3-3.11.1-r3.apk
```

withdrawn-packages.txt
```diff
@@ -1 +1,4 @@
...
+python3-dev-3.11.1-r3
+python3-doc-3.11.1-r3
```
ajayk, line 4:
```suggestion
python3-dev-3.11.1-r3.apk
python3-doc-3.11.1-r3.apk
```

withdrawn-packages.txt
smoser: this does look correct, given pgvector having the .apk on the end.
@xnox was there a reason you did not add .apk to the new entriews?

xnox:
@smoser @ajayk was paying too much attention to copy exact package name/version/epoch from apk output and totally forgot about adding .apk extension. Thanks for catching this.

ajayk (approved):

PR #14701, opened 2024-03-11
build confluent-common-docker and fix permission issues

Dentrax:
Build: https://github.com/confluentinc/common-docker

Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/confluent-common-docker.yaml
+++ b/confluent-common-docker.yaml
@@ -1,8 +1,7 @@
-#nolint:git-checkout-must-use-github-updates
 package:
   name: confluent-common-docker
-  version: 7.7.0.16
-  epoch: 0
+  version: 7.6.0
+  epoch: 1
   description: Confluent Commons with support for building and testing Docker images.
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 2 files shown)


EyeCantCU (approved):

confluent-common-docker.yaml
```diff
@@ -50,12 +52,7 @@ subpackages:
       - uses: strip
 
 update:
-  enabled: true
-  version-transform:
-    - match: "-"
-      replace: "."
-  release-monitor:
-    identifier: 371725
+  enabled: false # Upstream repo doesn't provide any jars/poms past 7.6.0: https://packages.confluent.io/maven/io/confluent/common/
```
jamonation, line 55: I'm thinking this should be enabled and use the git update check. A tag filter that looks for 7.6 only, with a note to remove the filter once 7.7 jars are public would ensure we get updates while that branch is still being maintained by confluent.

jamonation: Right, since we're blocked from checking for releases, this won't work.

jamonation (approved):

PR #14723, opened 2024-03-11
Sort out py3-pyyaml/py3-yaml - use pyyaml as canonical.

smoser:
There were 2 "yaml" packages both coming from
  https://github.com/yaml/pyyaml/archive/
Lets reduce that to one of them.

Do that by removing py3-pyyaml and then moving 'py3-yaml' to 'py3-pyyaml'.

I chose 'pyyaml' even though it is redundant as the name of the package. It is the canonical pypi named and also that as found in the url above.

Last, drop a comment 'without --with-libyaml for now'. The build is correctly enabling libyaml, the comment was wrong.

Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches


xnox (approved):
Yeap, wanted to do this in Atlanta but got distracted. Same resolution as I wanted to do.

I believe pyyaml was created precisely because it was not called under that name already.

PR #14747, opened 2024-03-12
opam: add 'support' for wolfi

xnox:
opam: add 'support' for wolfi

This gets rid of red-herring warnings that opam doesn't know how to query system packages on wolfi, by quering things using apk just like it does for Alpine.

This removes a warning from build logs. And this seems to be about it.

Related: https://github.com/ocaml/opam/pull/5878

### DIFF
--- a/opam.yaml
+++ b/opam.yaml
@@ -1,7 +1,7 @@
 package:
   name: opam
   version: 2.1.5
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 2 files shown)


xnox:
Meh, will come when it will come via new upstream release of opam https://github.com/ocaml/opam/releases/tag/2.2.0-beta2

PR #14755, opened 2024-03-12
add surrealdb

joshrwolf:
add surrealdb package

### DIFF
--- a/surrealdb.yaml
+++ b/surrealdb.yaml
@@ -1,0 +1,68 @@
+package:
+  name: surrealdb
+  version: 1.2.2
+  epoch: 0
+  description: "A scalable, distributed, collaborative, document-graph database, for the realtime web"
+  copyright:
+    - license: BUSL-1.1
+
+environment:
+  contents:
+    packages:
+      - bash
+      - build-base
+      - ca-certificates-bundle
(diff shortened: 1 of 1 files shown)


surrealdb.yaml
```diff
@@ -0,0 +1,68 @@
...
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/surrealdb/surrealdb
+      tag: v${{package.version}}
+      expected-commit: 0975683d170b61fcba1861da671b5bf775da0e1d
+
+  - name: Configure and build
+    runs: |
+      rustup install stable
```
dlorenc, line 30: Do we really need this if it only relies on stable? In the past I've only had to use this hack if they needed nightly.

joshrwolf: the build was failing with ours wit

```
error: linker `aarch64-linux-gnu-gcc` not found
```

so I assumed I needed to add the other toolchains off stable? if there's a better way I'm all 👂!

dlorenc: Oof, no that makes sense, seems like it's trying to do weird cross compilation. We can probably fix this in rust, but don't block here.

dlorenc (approved):

PR #13170, opened 2024-02-16
add csi-driver-hostpath

joshrwolf:
new package, useful for testing other csi extensions

also adds a new pipeline for simplifying the creation of `kwok` clusters.

advisory for false positive is here: https://github.com/wolfi-dev/os/pull/13170

PR #14754, opened 2024-03-12
Fix trivy version, add test to prevent

jdolitsky:
Fixes: https://github.com/wolfi-dev/os/issues/14753

Related:

#### For new package PRs only

- [ ] This PR is marked as fixing a pre-existing package request bug
  - [ ] Alternatively, the PR is marked as related to a pre-existing package request bug, such as a dependency
- [ ] REQUIRED - The package is available under an OSI-approved or FSF-approved license
- [ ] REQUIRED - The version of the package is still receiving security updates
- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`)

#### For new version streams

- [ ] The upstream project actually supports multiple concurrent versions.
- [ ] Any subpackages include the version string in their package name (e.g. `name: ${{package.name}}-compat`)
- [ ] The package (and subpackages) `provides:` logical unversioned forms of the package (e.g. `nodejs`, `nodejs-lts`)

#### For security-related PRs

- [ ] The security fix is recorded in the [advisories](https://github.com/wolfi-dev/advisories) repo

#### For version bump PRs

- [ ] The `epoch` field is reset to 0

#### For PRs that add patches

- [ ] Patch source is documented

PR #14782, opened 2024-03-12
add a smoke test for helm-operator

joshrwolf:
also re-introduces a `kwok` testing pipeline when mock clusters are useful

#13170


jamonation (approved):
LGTM! The CVE is an old/reopened one from a few years ago. https://github.com/wolfi-dev/os/pull/14754#issuecomment-1990110671 has some more context

PR #14783, opened 2024-03-12
withdrawn-packages: remove more old python3s

xnox:
pgvector & python3 are now gone

see:
https://github.com/wolfi-dev/os/actions/runs/8214813297/job/22467629459
https://github.com/wolfi-dev/os/actions/runs/8251913018/job/22570143014

However older revisions of unstreamed python are now published as latest. Withdraw all of them as well.

Fixes: https://github.com/wolfi-dev/os/pull/14691


withdrawn-packages.txt
```diff
@@ -1,4 +1,15 @@
-pgvector-0.6.1-r0.apk
-python3-3.11.1-r3.apk
-python3-dev-3.11.1-r3.apk
-python3-doc-3.11.1-r3.apk
+python3-3.10.7-r0.apk
```
smoser, line 1: is this meant to be a historical/complete list of all things withdrawn? or is it only input to a "run" of a tool that withdraws things.

ie, are you supposed to remove 'pgvector' from this list?

xnox: I believe that after one runs the withdraw packages workflow, these become historic and just "do nothing each time, cannot remove thing" and hence the workflow has `|| true`. I wish the withdraw packages would autogenerate a commit that drops removed packages from this file.

The full list of all withdrawn packages is in the bucket and published at https://packages.wolfi.dev/os/withdrawn-packages.txt see how it has all old versions.

smoser (approved):

PR #14796, opened 2024-03-12
Template python3

xnox:
Use template variable substitution as much as possible, to ease maintaining multiple streamed pythons.
Also this corrects typoed -doc package name.
Finally drop site-packages/README.txt and thus do not ship usr/lib/pytothon3.M/site-packages which then drops SCA generated python3~3.M dependency from the package. As python itself ships no site-packages.

Fixes: https://github.com/chainguard-dev/melange/pull/1079 https://github.com/wolfi-dev/os/commit/a17b2e5bca1434b7f953e24df6b5753bb760b66b


python-3.12.yaml
```diff
@@ -25,6 +25,17 @@ environment:
...
+# creates helpfull python3.M and 3.M variables
+var-transforms:
+  - from: ${{package.name}}
+    match: '-'
+    replace: ''
+    to: python
+  - from: ${{package.version}}
+    match: (\d).(\d+).(\d+)
+    replace: '$1.$2'
+    to: pyversion
```
smoser, line 37: see https://github.com/chainguard-dev/melange/pull/1072

smoser:
> i like it, and it does seem like a common enough pattern and a comon enough thing. It's either useful or wrong. For customization we could totaly add var-transforms dict key majorminorRegexp to override it with something else for the "weird" non-semver projects.
> maybe call it SemverMajorMinor? because that's the intent here. Or if it fails to parse set it to empty to avoid confusions.
> also doing it in vars-transforms today; means no need to wait on melage here. and it adds weight that we can refactor all of these things out of .yaml into melange default down the road.

@xnox said the above in slack wrt my "see chainguard-dev/melange#1072" comment.

I'm fine with this as it is now.

smoser (dismissed):

xnox:
> @xnox said the above in slack wrt my "see [chainguard-dev/melange#1072](https://github.com/chainguard-dev/melange/pull/1072)" comment.
> 
> I'm fine with this as it is now.

Resolved conflict, and resolved that comment, which means re-approval is needed.
Please re-approve.

smoser (approved):

PR #14841, opened 2024-03-12
chore(chromium): Add busybox and coreutils as runtime deps

EyeCantCU:
Needed by the wrapper

### DIFF
--- a/chromium.yaml
+++ b/chromium.yaml
@@ -2,7 +2,7 @@
 package:
   name: chromium
   version: 122.0.6261.99
-  epoch: 4
+  epoch: 5
   description: "Open souce version of Google's chrome web browser"
   copyright:
     - license: BSD-3-Clause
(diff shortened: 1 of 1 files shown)


mamccorm (dismissed):

EyeCantCU:
I had to bump the epoch which I did several minutes ago, but GitHub is slow right now. Not sure if the approval will be dismissed as stale when the PR finally catches up or not unfortunately

ajayk:
is it possible to update to this version mentioned in the update  https://github.com/wolfi-dev/os/pull/14840

EyeCantCU:
> is it possible to update to this version mentioned in the update #14840

It is. Unfortunately it looks like it isn't updating the expected commit. I'll manually update Chromium to match the version in the PR with the expected commit bumped

EyeCantCU:
Added the update to this PR so we get everything in one go rather than waiting on two builds back to back

mamccorm (approved):

EyeCantCU:
We're running out of space on the update check 😂

PR #14852, opened 2024-03-13
setuptools: all in one

xnox:
Remove `py3.*-setuptools.yaml` packages.
Create pre-installed `py3.*-setuptools` packages.

Todo:
- [ ] verify that this way of building `py3.*-setuptools` is equivalent to the old ones
- [ ] ensure provides/depends/content is correct
- [ ] diff apks


setuptools.yaml
```diff
@@ -0,0 +1,60 @@
...
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - py3-pip
+      - py3-wheel
+      - python-3.10-base
+      - python-3.11-base
+      - python-3.12-base
```
smoser, line 22: i was thinking to add a 'pythons-meta.yaml' package (or similar name) with a 'pythons-base' and 'pythons-base-dev'.

setuptools.yaml
```diff
@@ -0,0 +1,60 @@
...
+      uri: https://files.pythonhosted.org/packages/source/s/setuptools/setuptools-${{package.version}}.tar.gz
+
+  - name: Build Python Wheels
+    runs: |
+      pip wheel --no-deps .
+      mkdir -p ${{targets.destdir}}/usr/share/python-wheel/
+      mv *.whl ${{targets.destdir}}/usr/share/python-wheel/
+
+subpackages:
+  - range: py-versions
```
smoser, line 45: i dont see how, but I'd really like to define the range based on the 'pythons-meta' or something. but that feels ilke too much magic at this point.

xnox: Ideally we would do `include yaml`

pypi-setuptools.yaml
```diff
@@ -0,0 +1,56 @@
...
+  pypi-package: setuptools
+
+data:
+  - name: py-versions
+    items:
+      3.12: ""
+      3.11: ""
+      3.10: ""
+
+pipeline:
```
imjasonh, line 33: This means the package will be empty, which might be surprising to folks. We have a linter that might complain about it too, which is not blocking.

Is it possible this package being empty is causing the problems you're seeing in CI? Does adding something to `${{targets.destdir}}/blah` cause anything different to happen?

xnox: let me try that, i do see that "empty" build base does do `mkdir ${{targets.destdir}}` and that it ends up containing an sbom......

.... maybe our sbom generator fails without even a destdir?

pypi-setuptools.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+subpackages:
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}
+    description: Easily download, build, install, upgrade, and uninstall Python packages
+    pipeline:
+      - runs: |
+          python${{range.key}} setup.py install --no-compile --root="${{targets.subpkgdir}}" --prefix=/usr
+          rm -rf build/
+          python${{range.key}} -m compileall --invalidation-mode=unchecked-hash -r100 ${{targets.subpkgdir}}/usr/lib/
```
smoser, line 49: fwiw, i am not convinced i like 'unchecked-hash', i ran into an issue with it today ... admittedly debugging, but if you go and change a .py file it will _not_ change code path. the .pyc will be used and run and a stack trace will show the actual line number (from the .pyc file) but the content of the  .py file. 

lets just say its confusing.

pypi-setuptools.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+vars:
+  pypi-package: setuptools
+
+data:
+  - name: py-versions
+    items:
+      3.12: ""
+      3.11: ""
+      3.10: ""
```
mesaglio, line 31:
```suggestion
      3.12:
      3.11:
      3.10:
```
If you don't use the value, you can omitted.

pypi-setuptools.yaml
```diff
@@ -0,0 +1,60 @@
...
+    pipeline:
+      - runs: |
+          python${{range.key}} setup.py install --no-compile --root="${{targets.subpkgdir}}" --prefix=/usr
+          rm -rf build/
+          python${{range.key}} -m compileall --invalidation-mode=unchecked-hash -r100 ${{targets.subpkgdir}}/usr/lib/
+
+      - uses: strip
+    dependencies:
+      provides:
+        - py3-setuptools=${{package.full-version}}-${{range.key}}
```
mesaglio, line 54: Is this version apk version compliance?
Can no be
```suggestion
        - py${{range.key}}-setuptools=${{package.full-version}}
```

smoser: @mesaglio ,

I didn't follow the question "Is this version apk compliance?"

I think the intent of 'py3-setuptools'  provides is to allow something to depend on "latest setuptools", similar to the 'python3' provides that we've had before. The newest available version would be the highest.

I think your suggested provides is (py3.12-setuptools) ends up being the same as the package name, right?

mesaglio: The variabel `package.full-version` have, for example the follow format `1.36.0-r6` (**version**-**epoch**) value. We need to check if we define, a version formated **version**-**epoch**-**py_version**, dont break another thing. For example, https://github.com/chainguard-dev/melange/blob/165db39be30913d0a1c98258ba4e1451d3002d45/pkg/config/config.go#L1020

pypi-setuptools.yaml
xnox: @smoser the other two options are force hash check (where on startup all .py files are read and hashed), or switch to timestamp (where on startup all .py files are stat'ed and timestamps are compared). In practice, most other distributions do timestamp check (as they generate pyc files in maintainer scripts).

pypi-setuptools.yaml
xnox: @mesaglio i see what you mean, and I do agree mostly. We don't need the `py${{range.key}}-setuptools=${{package.full-version}}` as that's recursive and is already the name of this subpackage. We do need to provide `py3-setuptools` somewhere, as all previous builds did that I think (for backwards compat). Or maybe we need to give up on that, if we want to allow all setuptools to be co-installable for multiple python versions. As provides conflict with each other right?

pypi-setuptools.yaml
xnox: i see curl rustlts does something similar and explicitely marks the parent package as empty.

pypi-setuptools.yaml
xnox: Ideally i wish them to have provides at different priorities.

But somehow failing to make that work.

xnox: making py3-setuptools an explicit meta-subpackage.

PR #14941, opened 2024-03-14
python3.12: Remove -default package from python-3.12, use -base instead

ajayk:


smoser:
I put this up at https://github.com/wolfi-dev/os/pull/14943 . it is slightly different than here.
There will be a slight difference in the output of the two python-3.12.yaml files , with respect to usr/lib/pytyon3.X/config-3.X-x86_64-linux-gnu . mine has some better comments.

I'll pull your epoch bumps in also.

smoser (changes requested):
Can we use https://github.com/wolfi-dev/os/pull/14943 instead?

i re-worked that config- dir a bit and i think the result is nicer.

ajayk:
yep we can close this in favor of https://github.com/wolfi-dev/os/pull/14943

PR #15068, opened 2024-03-14
libxcrypt: prepare for transition

xnox:
apko prefers same-origin libcrypt1 from glibc. Even if it is no longer built, and only obsolete versions available.

Until libcrypt1 are withdrawn from glibc, keep so:libcrypt.so.1 provides in glibc.

This means libxcrypt will not provide so:libcrypt.so.1, and glibc will choose to depend on it for the actual implementation.

Update configuration options, to match against libcrypt1 - specifically keep static library build, and do not provide OpenSUSE / xcrypt.h symlinks which were never used in wolfi.

Update packaging metadata to ensure that libxcrypt-dev and libxcrypt can be unpacked by apk-tools with documented file replacements.

Once this lands, glibc can be upgraded to stop building crypt library and instead depend on libxcrypt from libcrypt1 package.

Glibc is blocked until this lands first - see https://github.com/wolfi-dev/os/pull/15069

### DIFF
--- a/libxcrypt.yaml
+++ b/libxcrypt.yaml
@@ -1,10 +1,17 @@
 package:
   name: libxcrypt
   version: 4.4.36
-  epoch: 3
+  epoch: 4
   description: "Modern library for one-way hashing of passwords"
   copyright:
     - license: GPL-2.0-or-later AND LGPL-2.1-or-later
+  options:
+    # For apko same-origin resolution
+    no-provides: true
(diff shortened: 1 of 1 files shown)


smoser:
Lets do this the week of 3/25.

xnox:
@smoser @ajayk it is time; can i please get an approval for this pull request?

smoser (approved):

PR #15107, opened 2024-03-14
confluent-common-docker: use maven package

Dentrax:


confluent-common-docker/pombump-deps.yaml
```diff
@@ -0,0 +1,13 @@
+patches:
+  # Fixes CVE-2023-51775
+  - groupId:  org.bitbucket.b_c
+    artifactId: jose4j
+    version: 0.9.6
+    scope: import
+    type: pom
+  # Fixes CVE-2024-23944
+  - groupId: org.apache
```
jamonation, line 9:
```suggestion
  - groupId: org.apache.zookeeper
```

mamccorm (approved):
Looks good. Separate advisory to be filed for couple CVEs but talked to @Dentrax and he is taking care of that in parallel.

PR #15117, opened 2024-03-14
coredns/1.11.2 revert as release dropped upstream

rawlingsj:


debasishbsws:
last pipeline :`setcap cap_net_bind_service=+ep "${{targets.destdir}}/usr/bin/coredns"` will also not work. Not sure why. 
And we can drop the go version it will build with the latest go-1.22

coredns.yaml
```diff
@@ -25,17 +25,21 @@ environment:
...
       make check
 
+  - uses: go/bump
+    with:
+      deps: golang.org/x/net@v0.17.0 google.golang.org/grpc@v1.58.3 golang.org/x/crypto@v0.17.0 github.com/quic-go/quic-go@v0.37.7
+
   - uses: go/build
     with:
-      go-package: go-1.20
+      go-package: 1.20
```
debasishbsws, line 42:
```suggestion
```

debasishbsws: Hey If you are trying to use the --go-version=1.20 of go/bump. It might be a mistake but Its always looks for the platform go version no matter what we specify

mattmoor (approved):

mattmoor:
merging so this gets a go/bump

mattmoor:
@rawlingsj we need to run withdraw

PR #15318, opened 2024-03-16
fix(R): Add needed system dependencies

EyeCantCU:
Adds dependencies needed at runtime: https://solutions.posit.co/envs-pkgs/environments/docker/#system-dependencies

Also fixes the R package manager

### DIFF
--- a/R.yaml
+++ b/R.yaml
@@ -2,14 +2,40 @@
 package:
   name: R
   version: 4.3.1
-  epoch: 2
+  epoch: 3
   description: Language and environment for statistical computing
   copyright:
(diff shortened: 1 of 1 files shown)


EyeCantCU:
The only one MIA is `pandoc`. Packaging that right now and this will be good to go

EyeCantCU:
Referenced the current container based on Ubuntu Jammy (realized the template Dockerfile covering system dependencies was based on Xenial) to ensure we are providing everything we need to be and found that pandoc was not installed nor is a package maintained in Ubuntu for it anymore. Marking as ready for review :)

dlorenc (approved):
Amazing!

PR #15397, opened 2024-03-18
fix(cassandra-medusa): Add python-3.11-base as a runtime dep

EyeCantCU:
Fixes: no-depends was blocking python-3.11-base from being installed as an implicit dependency

### DIFF
--- a/py3-cassandra-medusa.yaml
+++ b/py3-cassandra-medusa.yaml
@@ -108,8 +108,5 @@ test:
         fail() { echo "$@" 1>&2; exit 1; }
         out=$(/usr/share/medusa/.venv/bin/python3 -m medusa.service.grpc.server 2>&1)
         status=$?
-        if [ $status -ne 0 ]; then
-          fail "medusa.service.grpc.server exited $status: $out"
-        fi
-        echo "$out" | grep -q '/etc/medusa/medusa.ini' || fail "medusa.service.grpc.server output did not contain expected 'medusa.ini' message!"
-        echo "medusa.service.grpc.server started successfully."
+        echo "$out" | grep -q '/etc/medusa/medusa.ini' || fail "medusa.service.grpc.server output did not contain expected 'medusa.ini' message. Exit status $status: $out"
+        echo "medusa.service.grpc.server exited with expected error message"
--- a/py3-cassandra-medusa.yaml
+++ b/py3-cassandra-medusa.yaml
@@ -104,4 +104,8 @@ test:
         - python-3.11-dev
   pipeline:
     - runs: medusa --version
-    - runs: /usr/share/medusa/.venv/bin/python3 -m medusa.service.grpc.server server.py 2>&1 | grep "/etc/medusa/medusa.ini"
+    - runs: |
+        fail() { echo "$@" 1>&2; exit 1; }
+        out=$(/usr/share/medusa/.venv/bin/python3 -m medusa.service.grpc.server 2>&1) ||
+            fail "medusa.service.grpc.server exited $?: $out"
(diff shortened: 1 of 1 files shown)


py3-cassandra-medusa.yaml
```diff
@@ -96,5 +103,5 @@ test:
       packages:
         - python-3.11-dev
   pipeline:
-    - runs: |
-        medusa --version
+    - runs: medusa --version
+    - runs: /usr/share/medusa/.venv/bin/python3 -m medusa.service.grpc.server server.py 2>&1 | grep "/etc/medusa/medusa.ini"
```
smoser, line 107: this fails to capture error if stdout happened to have `/etc/medusa/medusa.ini` in it.

for example, the left side of the pipe could be this program and you'd pass:

```
#!/bin/sh
echo PANIC PANIC PANIC no /etc/medusa.ini 
exit 1
```

you can avoid ignoring 'python3' exiting non-zero like this:

```yaml
runs: |
    fail() { echo "$@" 1>&2; exit 1; }
    out=$(/usr/share/medusa/.venv/bin/python3 -m medusa.service.grpc.server 2>&1) ||
        fail "medusa.service.grpc.server exited $?: $out"
    echo "$out" | grep -q /etc/medusa/medusa.ini || fail "medusa.service.grpc.server output did not contain 'medusa.ini'"
```

These are just shell footguns that we need to get rid of.

we need a library of test tools to work with so that you could write something like this:

```yaml:
 - runs: |
      assert --exit=0 --output-contains=/etc/medusa/medusa.ini \
           "medusa.service.grpc.server output contains medusa.ini" -- \
           /usr/share/medusa/.venv/bin/python3 -m medusa.service.grpc.server server.py
```

smoser (commented):
I'm just marking this as a comment rather than 'request changes', because the right fix is more involved and requires more infrastructure.

py3-cassandra-medusa.yaml
```diff
@@ -2,13 +2,16 @@
...
+  epoch: 2
   description: Apache Cassandra backup and restore tool
   copyright:
     - license: Apache-2.0
   options:
     no-provides: true
     no-depends: true
+  dependencies:
+    runtime:
+      - python-3.11-base
```
rawlingsj, line 14: does this need to be tied to python 3.11?

py3-cassandra-medusa.yaml
Dentrax: Thanks! Revisited your test snippet to following:
```yaml
- runs: |
         set +e
         fail() { echo "$@" 1>&2; exit 1; }
         out=$(/usr/share/medusa/.venv/bin/python3 -m medusa.service.grpc.server 2>&1)
         status=$?
         echo "$out" | grep -q '/etc/medusa/medusa.ini' || fail "medusa.service.grpc.server output did not contain expected 'medusa.ini' message. Exit status $status: $out"
         echo "medusa.service.grpc.server exited with expected error message"
```

py3-cassandra-medusa.yaml
EyeCantCU: Oh nice. Will close this one then. Thank you so much for your help with this!

EyeCantCU: Oh wait. Not this one. The other one. My had

smoser (approved):
there are still shell foot-guns here (example: the program can still exit 127 with a message saying '/etc/medusa/medusa.ini has invalid syntax' and the test will pass).

Fixing that completely isn't something that should be expected of your package fix.

developer-guy (approved):
it looks like all the comments are addressed so far, so lgtm

PR #15442, opened 2024-03-19
patch the default supported ciphers

joshrwolf:
upstream ref: https://github.com/argoproj/argo-cd/pull/17569


joshrwolf:
closed in favor of #15456

PR #15455, opened 2024-03-19
Add step-kms-plugin

jamonation:
#### For new package PRs only

### DIFF
--- a/step-kms-plugin.yaml
+++ b/step-kms-plugin.yaml
@@ -1,0 +1,34 @@
+package:
+  name: step-kms-plugin
+  version: 0.10.0
+  epoch: 0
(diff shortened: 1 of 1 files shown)


step-kms-plugin.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/smallstep/step-kms-plugin
+      tag: v${{package.version}}
+      expected-commit: fcb6083d8d929de845b7c13bafcdb85c5287855d
+
+  - runs: |
+      CGO_ENABLED=1 go build -v \
```
ajayk, line 24: can we use go/build  here ?

ajayk (approved):

PR #15492, opened 2024-03-20
feat(packages): Add harbor registryctl, jobservice, and portal

EyeCantCU:
#### For new package PRs only

### DIFF
--- a/harbor-core.yaml
+++ b/harbor-core.yaml
@@ -62,7 +62,7 @@ subpackages:
           modroot: ./src
       - runs: |
           mkdir -p ${{targets.subpkgdir}}/harbor
-          ln -sf /usr/bin/harbor-jobservice ${{targets.destdir}}/harbor/harbor_jobservice
+          ln -sf /usr/bin/harbor-jobservice ${{targets.subpkgdir}}/harbor/harbor_jobservice
 
   - name: harbor-portal
     description: harbor portal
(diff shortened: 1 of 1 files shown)


harbor-core.yaml
```diff
@@ -47,6 +51,74 @@ pipeline:
...
+    description: harbor jobservice
+    pipeline:
+      - uses: go/build
+        with:
+          packages: ./jobservice
+          output: harbor-jobservice
+          modroot: ./src
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/harbor
+          ln -sf /usr/bin/harbor-jobservice ${{targets.destdir}}/harbor/harbor_jobservice
```
dlorenc, line 65: This means that the subpackage has to depend on the full package right? Is that intentional?

harbor-registry.yaml
```diff
@@ -0,0 +1,62 @@
...
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/distribution/distribution
+      tag: v${{package.version}}
+      expected-commit: 4772604ae973031ab32dd9805a4bccf61d94909f
+
+  - runs: |
+      # Fetch goharbor's redis patch
+      curl https://raw.githubusercontent.com/goharbor/harbor/main/make/photon/registry/redis.patch -o redis.patch
```
dlorenc, line 27: Should we either vendor this patch in this repo or pin it by digest?

EyeCantCU: Vendoring it here is probably way to go. Then we can just drop curl as an env dep

harbor-core.yaml
EyeCantCU: No, it doesn't have to depend on the full package. There isn't an implicit dependency created here that brings in `harbor-core` so we're good

EyeCantCU:
CI attempts to build core and associated subpackages before the registry causing registryctl's build to fail. I'll split this into another PR

EyeCantCU:
This PR will build after this has been merged: https://github.com/wolfi-dev/os/pull/15496

harbor-core.yaml
EyeCantCU: OH WAIT. I see what you mean. Fix incoming

imjasonh (approved):

PR #15539, opened 2024-03-20
fix: remove py3-tensorflow-core provides tensorflow shared libraries

lpcalisi:
* Remove python cache files (.pyc )
* `/usr/bin/` files are Python scripts that should be provided by py3-tensorflow-core
* Remove tensor shared  libraries from py3-tensorflow-core because it's already been provided by tensorflow-core package.

### DIFF
--- a/tensorflow-core.yaml
+++ b/tensorflow-core.yaml
@@ -2,7 +2,7 @@ package:
   name: tensorflow-core
   description: Framework for data-graph oriented computing (core libraries, oneDNN build)
   version: 2.13.0
-  epoch: 1
+  epoch: 2
   copyright:
     - license: Apache-2.0
 
@@ -144,7 +144,10 @@ subpackages:
           _pyver=$(python -c 'import sys; print(str(sys.version_info[0]) + "." + str(sys.version_info[1]))')
           _pypath="/usr/lib/python${_pyver}"
           mkdir -p "${{targets.contextdir}}"/usr/lib
+          mkdir -p "${{targets.contextdir}}"/usr/bin
           mv "${{targets.destdir}}"${_pypath} "${{targets.contextdir}}"${_pypath}
+          mv "${{targets.destdir}}"/usr/bin "${{targets.contextdir}}"/usr/bin
+          find "${{targets.contextdir}}"${_pypath} -name "libtensorflow*.so*" -exec rm -rf '{}' +
 
   - name: tensorflow-dev
     description: Tensorflow development headers


tensorflow-core.yaml
```diff
@@ -144,7 +144,11 @@ subpackages:
           _pyver=$(python -c 'import sys; print(str(sys.version_info[0]) + "." + str(sys.version_info[1]))')
           _pypath="/usr/lib/python${_pyver}"
           mkdir -p "${{targets.contextdir}}"/usr/lib
+          mkdir -p "${{targets.contextdir}}"/usr/bin
           mv "${{targets.destdir}}"${_pypath} "${{targets.contextdir}}"${_pypath}
+          mv "${{targets.destdir}}"/usr/bin "${{targets.contextdir}}"/usr/bin
+          find "${{targets.contextdir}}"${_pypath} -name "*.pyc" -exec rm -rf '{}' +
+          find "${{targets.contextdir}}"${_pypath} -name "libtensorflow*.so*" -exec rm -rf '{}' +
```
smoser, line 151: Can you verify that py3-tensorflow-core correctly gets the dependency on libtensorflow ? I expect it will.

tensorflow-core.yaml
```diff
@@ -144,7 +144,11 @@ subpackages:
           _pyver=$(python -c 'import sys; print(str(sys.version_info[0]) + "." + str(sys.version_info[1]))')
           _pypath="/usr/lib/python${_pyver}"
           mkdir -p "${{targets.contextdir}}"/usr/lib
+          mkdir -p "${{targets.contextdir}}"/usr/bin
           mv "${{targets.destdir}}"${_pypath} "${{targets.contextdir}}"${_pypath}
+          mv "${{targets.destdir}}"/usr/bin "${{targets.contextdir}}"/usr/bin
+          find "${{targets.contextdir}}"${_pypath} -name "*.pyc" -exec rm -rf '{}' +
```
smoser, line 150: We switched in python base packages to _not_ remove these files. lack of .pyc files means that every python startup includes the cost of creating them (even if there is no place to write them).

smoser (changes requested):

tensorflow-core.yaml
lpcalisi: When libtensorflow.so not exists on site-packages, python search it on system paths. We could set tensorflow-core as dependency of this subpackage or install it only if you need, because not always tensorflow loads the shared libaries

smoser: I think  you're saying that python will load the shared library from the site-packages dir and if not present then use one from /usr/lib.  That is correct.

If you remove the library from site-packages, then there will be a dependency on the /usr/lib version to work correctly, so we need to verify that there is a dependency added.  It _should_ get automatically added, but I want to verify.  

(you can look at log output or the apk itself to check).

lpcalisi: Yes, now py3-tensorflow-core depends on `so:libtensorflow_cc.so.2` and `so:libtensorflow_framework.so.2`


[logs here](https://github.com/wolfi-dev/os/actions/runs/8364395164/job/22906710667?pr=15539#step:6:15330)

tensorflow-core.yaml
lpcalisi: rollbacked

smoser (approved):

lpcalisi:
@ajayk

tensorflow-core.yaml
mesaglio: @smoser we need to update [some](https://github.com/search?q=repo%3Awolfi-dev%2Fos+*.pyc&type=code) melange definitions.

PR #15645, opened 2024-03-21
add parallel postsubmit on GH hosted runners

imjasonh:
tl; dr: It works! 🎉 

It doesn't use gcsfuse yet, but that's okay since it consults what's already available in the existing index to decide what to build.


joshrwolf (approved):
🚀

PR #15749, opened 2024-03-23
confluent-kafka: add java and busybox deps

Dentrax:


rawlingsj:
Is this PR needed? It's still in draft so not sure if expecting a review.

confluent-kafka.yaml
```diff
@@ -10,18 +10,21 @@ package:
...
   dependencies:
     runtime:
       - bash # Required by kafka - uses shebangs to launch the image.
+      - busybox
       - confluent-common-docker
       - confluent-common-docker-ub
       - confluent-common-docker-base
       - confluent-docker-utils
       - confluent-kafka-images-kafka
+      - openjdk-17-jre
```
mamccorm, line 26: These are currently set in the image as runtime deps:
- https://github.com/chainguard-images/images/blob/ff7a75578060194b99465fb2b1bcf03414b64170/images/confluent-kafka/config/main.tf#L10

So unless your thinking of re-factoring then we probably don't need this one? with Java, keeping the JRE as runtime dep means we can switch out the runtime (i.e jdk11, jdk17, jdk21) as needed. But i'd not be too pushed if we included them here instead. At the minute they are in the image

mamccorm: Given the approach we're taking means we don't need to invoke this module and pass in an alternative JRE, think this is actually cleaner when we've other jobs invoking this, so LGTM

mamccorm (approved):

confluent-kafka.yaml
Dentrax: Actually I was thinking to re-factor the image. Package should be installed with all of the required runtime deps.

PR #15775, opened 2024-03-24
DNM: add confluent-common

Dentrax:


confluent-common.yaml
```diff
@@ -0,0 +1,42 @@
+package:
+  name: confluent-common
+  version: 7.6.0
```
mamccorm, line 3: what was the issue with the later versions? does it look like something we can fix? This would still have vulns so if we can't move forward we'd have to patch the two we're trying to resolve?

Dentrax: This was the compilation issue:

```
[ERROR] Failed to execute goal on project common-config: Could not resolve dependencies for project io.confluent:common-config:jar:7.8.0-1: The following artifacts could not be resolved: io.confluent:common-utils:jar:7.8.0-1 (absent), io.confluent:assembly-plugin-boilerplate:zip:resources:7.8.0-1 (absent): Could not find artifact io.confluent:common-utils:jar:7.8.0-1 in confluent (https://packages.confluent.io/maven/) -> [Help 1]
```

PR #11127, opened 2024-01-13
new pkg: aws-cli (v2), rename aws-cli to aws-cli-v1

tuananh:
Tested by creating an apko image

image

I tried moving documentation to sub-package but apparently, they use it for `aws help` command. so we will need to ignore melange documentation linter as well.

Fixes:

Related: #1848 #2272 #3884

#### For new package PRs only

- [x] This PR is marked as fixing a pre-existing package request bug
  - [ ] Alternatively, the PR is marked as related to a pre-existing package request bug, such as a dependency
- [x] REQUIRED - The package is available under an OSI-approved or FSF-approved license
- [x] REQUIRED - The version of the package is still receiving security updates
- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`)

#### For new version streams

- [x] The upstream project actually supports multiple concurrent versions.
- [ ] Any subpackages include the version string in their package name (e.g. `name: ${{package.name}}-compat`)
- [ ] The package (and subpackages) `provides:` logical unversioned forms of the package (e.g. `nodejs`, `nodejs-lts`)

#### For security-related PRs

- [ ] The security fix is recorded in the [advisories](https://github.com/wolfi-dev/advisories) repo

#### For version bump PRs

- [ ] The `epoch` field is reset to 0

#### For PRs that add patches

- [ ] Patch source is documented

PR #15808, opened 2024-03-25
Revert "new pkg: aws-cli (v2), rename aws-cli to aws-cli-v1"

jamonation:
Reverts wolfi-dev/os#11127


xnox (changes requested):
NAK, we better fix it up.

PR #15901, opened 2024-03-27
fix(shadowsocks-rust): recude package size

Dentrax:
Fixes an issue where all sub-packages were containing the all other binaries.

### DIFF
--- a/shadowsocks-rust.yaml
+++ b/shadowsocks-rust.yaml
@@ -1,7 +1,7 @@
 package:
   name: shadowsocks-rust
   version: 1.18.2
-  epoch: 2
+  epoch: 3
   description: A Rust port of shadowsocks
   copyright:
     - license: MIT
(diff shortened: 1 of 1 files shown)


shadowsocks-rust.yaml
```diff
@@ -53,8 +43,12 @@ subpackages:
     name: "${{package.name}}-${{range.key}}"
     pipeline:
       - runs: |
-          mkdir -p "${{targets.subpkgdir}}"/usr/bin/
-          install -Dm755 "${{targets.destdir}}"/target/release/${{range.key}} "${{targets.subpkgdir}}"/usr/bin/
+          rustup override set stable
+          rustup target add "${{host.triplet.rust}}"
+      - runs: |
```
vaikas, line 48: Do the tests below need everything to work?  To get tighter tests, we could add subpackage tests that only test with that package installed 🤷 

Since I don't know, consider this uninformed drive by commentary 🤣 

If they don't we could add the test pipeline here and run the:
  ${{package.name}}-${{range.key}} --version
  
  You'd have to have an if/else cuase for the sslocal/ssserver to nohup bin them.

Dentrax: Yeah actually packages need everything to work. I didn't aware we can add test pipelines outside of `tests:` section. Is there any example?

vaikas: There's a simple one here:
https://github.com/chainguard-dev/melange/blob/main/docs/TESTING.md#execution-environment-guest

vaikas (dismissed):

imjasonh (approved):

PR #15931, opened 2024-03-27
Python dev - Build python packages for each of supported python versions.

smoser:
I need to write some doc for this PR.


smoser:
Things to point out that a reviewer would want to look at:
1. use of 'py3-supported-python' - this is a way to get rid of a hard coded list of build time dependencies.
2. use of py3-supported-<thing> (like 'py3-supported-installer' or 'py3-supported-pip') which allows a package to build-depend on a single package rather than 3.
3. building with other pythons outside wolfi is not great here. 

Things that would make life easier:
* ability to loop or some other way avoid the '-supported'  packages above.
* dynamnically building subpackages (so we don't have to have 3 lines like below.  I'd rather have it "just work" than require the yaml file to declare that there is a package).

      - name: pypi-wheels-${{vars.pypi-package}}
        pipeline:
          - uses: py/collect-wheels

* not declare 'py-versions' everywhere.

dlorenc:
I like this approach! it seems like it won't confuse our scanner/advisory feed because it's still one version of each python library just "linked" against different python versions.

smoser:
I don't think the build failure right now is due to my changes, but rather the dependency resolver.  I do not think it is a circular dependency (which did cause issues earlier)

    2024-03-29T16:56:46.4906470Z 2024/03/29 16:56:46 INFO error during command execution: 
        failed to build package: unable to build guest: 
           unable to generate image: installing apk packages: 
             error getting package dependencies: 
             solving "py3-supported-gpep517" constraint: 
               resolving "py3-supported-gpep517-15-r3.apk" deps:
    2024-03-29T16:56:46.4908250Z could not find package either named 
        py3.10-tomli or that provides py3.10-tomli for py3.10-gpep517

py3-tomli.yaml does build a py3.10-tomli-2.0.1-r5.apk.


It does build locally given this order:

    py3-supported-python
    py3-flit-core
    py3-setuptools
    py3-pip
    py3-installer
    py3-tomli
    py3-gpep517
    py3-flit-core
    py3-wheel
    py3-cython
    py3-numpy
    py3-pathspec
    py3-pyyaml
    py3-yamllint

xnox:
This is really good.

1. This will enable us to trivially build numpy image latest-3.12 latest-3.11 latest-3.10 images, and then choose whichever one we need to be shipped to the user.

2. pybin thing is a necessary evil and that's fine to get there. However, I somehow feel we might want to auto-generate conflicting packages too. Or ask apko to move them around. Such that numpy image latest-3.12 tag has /usr/bin/numpy with the valid python.

3. To land this safely I do wonder if we need to generate default symlinks in the top level package for now.

xnox:
> ```
> 
> ```
> 
> 
>     
> 
>   
> 
> py3-tomli.yaml does build a py3.10-tomli-2.0.1-r5.apk.

we may need to land it first / separately, as that's install check rather than build check. And install checks are done after each build, rather than after all builds are completed first.

pnasrat:
Close in favor of #20318

PR #15983, opened 2024-03-28
Add flyte package.

dlorenc:


dlorenc:
There are a bunch of other binaries in this repo, but we can add them as subpackages later - this one is named `flyte` so it seems fine to add as the top level binary in this package.

ajayk (approved):

PR #16021, opened 2024-03-29
feat: get hints for failed builds from Open AI on presubmits that are created from a branch, not a fork

rawlingsj:
The bulk of this work comes from @hectorj2f 🙏

We get a fairly large number of Pull Requests that fail presubmit checks.  It can be a little hard to parse logs with human eyes from every build system and language ecosystem, let's try using machines 🤖!  

This feature is an initial pass at getting automated hints from Open AI based on failed build logs, these suggested hints are posted back to the PR as a comment.  The aim is to reduce the initial time it takes to understand what the error may be and consider possible fixes.

E.g.
 
Screenshot 2024-03-29 at 10 33 45

As a secret is required to interact withe the Open AI API, this feature is only available for PRs created from a branch on the main repo, not forks.  This means it will work with the automation that we have in Wolfi for package updates and CVE remediation but not outside contributions.  For now at least, options to move out of GitHub Actions in the future if this feature proves useful.

One slightly annoying this is I had to move the `make foo` command that performs the build into a script so we can both log to standard out AND log to a file that we can trim and send to the Open AI endpoint.

**Note**, these presubmit jobs are likely to change to using a parallel build system in the near future so we will need to change how we collect failed build logs.  I'm hoping that until then, we can merge this and get real world feedback on how affective this approach of automated hints can be.

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -81,7 +81,7 @@ jobs:
             runner: ubuntu-arm-16-cores
             oci: arm64
       fail-fast: false
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runner }}
     needs: changes
     outputs:
       packages_were_built: ${{ steps.file_check.outputs.exists }}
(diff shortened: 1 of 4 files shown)


hectorj2f (approved):
LGTM

.github/workflows/ci-build.yaml
```diff
@@ -149,10 +148,85 @@ jobs:
...
+        if: failure() && matrix.arch == 'x86_64' steps.secrets.outputs.token != '' 
+        run: |
+          if test -f "wolfi-packages.log"; then
+            echo "exists=true" >> $GITHUB_OUTPUT
+          else
+            echo "exists=false" >> $GITHUB_OUTPUT
+          fi
+
+      - name: GPTScript PR comment diff
+        if: failure() && matrix.arch == 'x86_64'&& steps.secrets.outputs.token != '' && steps.file_check_gpt_log.outputs.exists == 'true'
```
hectorj2f, line 217: Here I only post the hints if `arch==x86_64` and the build failed. But if a different architecture failed and x86_64 passed, then we don't call OpenAI to get any hints. I initially did this to avoid calling OpenAI twice assuming we got the same error on both architectures.

hectorj2f: Perhaps we might want to have some smart logic here, I went with a simple approach in my initial PR.

PR #16038, opened 2024-03-29
Switch to github-generated source tarballs for xz package.

dlorenc:


xz.yaml
```diff
@@ -9,19 +9,26 @@ package:
...
+      - m4
       - wolfi-baselayout
 
 pipeline:
   - uses: fetch # use fetch instead of git-checkout else there is a circular dependency with git
     with:
-      uri: https://github.com/tukaani-project/xz/releases/download/v${{package.version}}/xz-${{package.version}}.tar.gz
-      expected-sha256: aeba3e03bf8140ddedf62a0a367158340520f6b384f75ca6045ccc6c0d43fd5c
+      uri: https://github.com/tukaani-project/xz/archive/refs/tags/v${{package.version}}.tar.gz
+      expected-sha256: 60831005fddb270824fa9f7cdd28a59da8757fe95466ed5b10bcfe23379f17d9
```
mattmoor, line 27: bump epoch?

dlorenc:
I think I can break this dependency cycle (libxml2 is only a recommended dependency of gettext), but there might be more.

It might be worth pulling a few of these up to bootstrap repos to make cycle breaking easier.

PR #16057, opened 2024-03-30
changes user and adds compat package for squid

mritunjaysharma394:
Since, the source code doesn't have the `entrypoint.sh`, I exec'ed in the` ubuntu/squid image` and got that too which looks like the script added in the PR which I think is important to build and run the image. 

Also, In order to run this as non-root, this PR adds opts to build with default user as `squid`


squid/entrypoint.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
imjasonh, line 1: Where does this file come from? Ideally we'd get it from some upstream git repo where we can also get updates

mritunjaysharma394: Got this by execing into the only trustworthy image as the upstream: https://hub.docker.com/r/ubuntu/squid/tags

Seems not to have any Dockerfile https://github.com/chainguard-dev/image-requests/issues/1556#issuecomment-2020435125

mamccorm: Re: this part:

```bash
# re-create snakeoil self-signed certificate removed in the build process
if [ ! -f /etc/ssl/private/ssl-cert-snakeoil.key ]; then
    /usr/sbin/make-ssl-cert generate-default-snakeoil --force-overwrite > /dev/null 2>&1
fi
```

snakeoil is what debian uses as the name for it's default self-signed cert I believe. So not sure this transitions well here to wolfi. We might not even need this if we already have one or take care of it in existing packaging? This part probably needs a bit more digging into

mritunjaysharma394: Yes, even I am not sure about it, I feel our `openssl` package could be sufficient but I will dig more in this

squid/entrypoint.sh
```diff
@@ -0,0 +1,30 @@
...
+# systemd unit.
+
+# [1] The default configuration is changed in the Dockerfile to allow local
+# network connections. See the Dockerfile for further information.
+
+# re-create snakeoil self-signed certificate removed in the build process
+if [ ! -f /etc/ssl/private/ssl-cert-snakeoil.key ]; then
+    /usr/sbin/make-ssl-cert generate-default-snakeoil --force-overwrite > /dev/null 2>&1
+fi
+
```
mamccorm, line 23: Re: previous comment, if turns out we can remove the default cert step, then all that would be in this file is printing the logs to stdout and launching squid.

If that was the case i'd delete everything else, just have those lines left, and add a note about squid not logging anything by default to stdout, and it being a launch script.

Dependent on previous comment of course :)

mritunjaysharma394: And also that it requires root to run :)

jamonation: @mritunjaysharma394 - squid does drops uid/gid to `nobody` if it detects it is running as root. There are two settings it uses `cache_effective_user` and `cache_effective_group` in `squid.conf` to configure a different user.

mritunjaysharma394: Got it @jamonation that means we can still solve the problem of running it as non-root but I am still confused about snakeoil self-signed certificate thing, maybe I should give it a try without it and see

jamonation: @mritunjaysharma394 - yes squid's config that's being shipped here will run by default with no TLS. So If you're adding tests, a separate pipeline step to generate a key/cert and run with a squid.conf that uses them would be a substitute for the Debian/Ubuntu snakeoil cert.

squid/entrypoint.sh
mritunjaysharma394: I have removed this snakeoil thing and would like to test this by building an image so would love to know if this can be approved and we can see what works or is needed more?

mritunjaysharma394: Having discussed it with Jamon, we both feel that the current state of entrypoint.sh looks good with removal of cert generation from the script. Additionally, in order to run this as non-root, this PR adds opts to build with default user as squid

jamonation (approved):

squid/entrypoint.sh
jamonation: I'm ok with this file as it is - there aren't likely to be any changes since it is just tailing logs and running squid with argv passed through.

PR #16125, opened 2024-04-01
cleanup php packages

shyim:
- Setup release monitoring for all packages which don't have it
- Remove pecl suffix from packages to have similar naming
- Updated package if available

Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/php-8.1-http.yaml
+++ b/php-8.1-http.yaml
@@ -1,13 +1,14 @@
 package:
-  name: php-8.1-pecl-http
-  version: 4.2.3
-  epoch: 3
+  name: php-8.1-http
+  version: 4.2.4
+  epoch: 0
   description: "Provides PHP 8.1 HTTP module for PHP Extended HTTP Support- PECL"
   copyright:
     - license: BSD-2-Clause
   dependencies:
     provides:
       - php-pecl-http=${{package.full-version}}
+      - php-8.1-pecl-http=${{package.full-version}}
     runtime:
       - php-8.1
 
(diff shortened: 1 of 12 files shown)


php-8.1-http.yaml
```diff
@@ -1,13 +1,14 @@
...
+  name: php-8.1-http
+  version: 4.2.4
+  epoch: 0
   description: "Provides PHP 8.1 HTTP module for PHP Extended HTTP Support- PECL"
   copyright:
     - license: BSD-2-Clause
   dependencies:
     provides:
       - php-pecl-http=${{package.full-version}}
+      - php-8.1-pecl-http=${{package.full-version}}
```
shyim, line 11: Is that the correct way to rename a package?

shyim: or do I create empty packages which have a runtime dependency on the new stuff?

rawlingsj:
thanks for the PR, I think we need to keep the package names as they are else this will break downstream consumers of these apks.  The additions of the release monitoring to the config blocks is great though, I wonder if you would mind creating a new PR with just those changes so we get automated updates for these packages?

rawlingsj (changes requested):
thanks for the PR, I think we need to keep the package names as they are else this will break downstream consumers of these apks. The additions of the release monitoring to the config blocks is great though, I wonder if you would mind creating a new PR with just those changes so we get automated updates for these packages?

mamccorm:
Closing PR as theres been no initial loop backs since the first review, and all files have merge conflicts at this stage

PR #16165, opened 2024-04-02
Update otel-collector-contrib package build to use go/build pipeline

stgarf:
Update package build to use go/build pipeline

### DIFF
--- a/opentelemetry-collector-contrib.yaml
+++ b/opentelemetry-collector-contrib.yaml
@@ -12,8 +12,6 @@ environment:
       - busybox
       - ca-certificates-bundle
       - go
-  environment:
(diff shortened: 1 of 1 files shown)


opentelemetry-collector-contrib.yaml
```diff
@@ -12,8 +12,6 @@ environment:
       - busybox
       - ca-certificates-bundle
       - go
```
xnox, line 14: The whole environment paragraph can be deleted. All of the needed things here are pulled in by go/build pipeline already.

stgarf: Thank you for your thorough reviews @xnox -- fixed!!

xnox (approved):

PR #16267, opened 2024-04-03
Add conda-base that creates base env at the expected /root/conda

srishtih:
Thanks @EyeCantCU for putting this together!


jamonation (approved):
Looks good - hope we can find a way to run as non-root sometime.

PR #16387, opened 2024-04-05
gops: assert no crypto usage

xnox:
During build, verify that golang crypto module is not used. This ensures that no cryptography is performed by this software, and thus it does not need a FIPS variant.

Also use git-checkout and asserts expected tag hash.


tcnghia (approved):

imjasonh (approved):
Thank you!

PR #16443, opened 2024-04-06
chore(conda-init): Create base environment if it doesn't exist

EyeCantCU:
Creates a base environment after initializing conda if one can't be found. This is user agnostic however, the conda-base package should still be maintained to cut down on time to entry - this is a bandaid for nonroot users

Also redirect stdout to /dev/null during initialization to avoid throwing it in the user's face :)

### DIFF
--- a/conda.yaml
+++ b/conda.yaml
@@ -3,7 +3,7 @@
 package:
   name: conda
   version: 24.3.0
-  epoch: 2
+  epoch: 3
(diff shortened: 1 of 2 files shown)


dlorenc:
We need an epoch bump

EyeCantCU:
On it

EyeCantCU:
All good now :)

developer-guy (approved):

PR #16576, opened 2024-04-10
add ack tests

joshrwolf:
#13623 

Signed-off-by: Josh Wolf <josh@wolfs.io>

### DIFF
--- a/ack.yaml
+++ b/ack.yaml
@@ -1,4 +1,3 @@
-# Generated from https://git.alpinelinux.org/aports/plain/main/ack/APKBUILD
 package:
   name: ack
   version: 3.7.0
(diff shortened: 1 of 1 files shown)


ack.yaml
```diff
@@ -1,4 +1,3 @@
-# Generated from https://git.alpinelinux.org/aports/plain/main/ack/APKBUILD
```
ajayk, line 1: Comment removed

ajayk (dismissed):

ack.yaml
rawlingsj: I think it's ok, it was originally added by melange convert and most of the time we removed these

rawlingsj (approved):

PR #16603, opened 2024-04-10
add rust tests

joshrwolf:
#13623 
Signed-off-by: Josh Wolf <josh@wolfs.io>

### DIFF
--- a/rust.yaml
+++ b/rust.yaml
@@ -117,3 +117,31 @@ update:
   github:
     identifier: rust-lang/rust
     use-tag: true
+
+test:
+  environment:
+    contents:
+      packages:
+        - gcc
+        - glibc-dev
+  pipeline:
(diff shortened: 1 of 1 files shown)


rust.yaml
```diff
@@ -117,3 +117,31 @@ update:
   github:
     identifier: rust-lang/rust
     use-tag: true
+
+test:
+  environment:
+    contents:
+      packages:
+        - gcc
+        - glibc-dev
```
joshrwolf, line 126: should these be added as runtime dependencies?

ajayk: rustc needs any c compiler to be present 
https://users.rust-lang.org/t/why-rust-language-required-to-install-g-visual-c-gcc-additionally-for-compiling-rust-file/46123/15

ajayk (approved):

PR #16623, opened 2024-04-10
Slimmer OpenSSL

xnox:
Disable obsolete and deprecated build options:

 * disable DES and disable DSA. Both DES and DSA are obsoleted by NIST see:
    https://csrc.nist.gov/news/2023/nist-to-withdraw-sp-800-67-rev-2
    https://csrc.nist.gov/news/2023/nist-releases-fips-186-5-and-sp-800-186

 * disable engines support. Deprecated in 3.0 and no longer contain any interesting hardware acceleration. See: https://github.com/openssl/openssl/blob/master/README-ENGINES.md

 * drop fips packaging as it has moved elsewhere

 * dump config data after the build for ease of inspection


openssl.yaml
```diff
@@ -60,7 +52,10 @@ pipeline:
          no-ssl3 \
          no-seed \
          no-weak-ssl-ciphers \
+         no-des \
+         no-dsa \
          -Wa,--noexecstack
+      perl configdata.pm --dump
```
vaikas, line 58: Just checking that this is not debug stuff left around since it wasn't there before.

vaikas (approved):

openssl.yaml
```diff
@@ -12,14 +12,6 @@ package:
     runtime:
       - openssl-provider-legacy
 
-vars:
-  fips-enable: 'no-fips'
-
-options:
-  fips:
-    vars:
-      fips-enable: 'enable-fips'
```
imjasonh, line 21: This might be the last remaining usage of `options` in our repos, which means we might be able to remove this from Melange 👀

dlorenc: yay!

openssl.yaml
xnox: Intentional commit and mentioned in the pull request description.

The output this produces allows inspection of all the things enabled and disabled. After configure is executed it even suggests to run that to see the configuration state.
Many keys are non declarative, and cascade enable/disable things in unexpected ways.

xnox:
Well that didn't land well. https://github.com/wolfi-dev/os/issues/16698 on steps on how to land it better.

PR #16634, opened 2024-04-10
run test-the-world on the host instead of in a container to avoid dind

joshrwolf:


hectorj2f (approved):

imjasonh (approved):

.github/workflows/test-world.yaml
```diff
@@ -59,23 +59,24 @@ jobs:
           sudo apt-get install acl
           sudo setfacl --modify user:$USER:rw /var/run/docker.sock
 
+      # this need to point to main to always get the latest action
```
imjasonh, line 62: This should get updates with PRs, using dependabot. Not blocking.

PR #16693, opened 2024-04-11
vertical-pod-autoscaler: add tests

Dentrax:
Also `golang.org/x/net@v0.17.0` is no longer needed.


vaikas (dismissed):
Since you removed the deps, should the epoch be bumped also?

Dentrax:
> Since you removed the deps, should the epoch be bumped also?

Good catch! Done.

hectorj2f (approved):

PR #16702, opened 2024-04-11
node-feature-discovery-0.15: new package

Pyronewbic:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/node-feature-discovery-0.15.yaml
+++ b/node-feature-discovery-0.15.yaml
@@ -1,0 +1,94 @@
+package:
+  name: node-feature-discovery-0.15
+  version: 0.15.4
+  epoch: 0
+  description: Node feature discovery for Kubernetes
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - protoc
+    provides:
+      - node-feature-discovery=${{package.full-version}}
+
+environment:
+  contents:
+    packages:
+      - protoc
+  environment:
+    GRPC_GO_LOG_SEVERITY_LEVEL: "INFO"
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/kubernetes-sigs/node-feature-discovery
+      tag: v${{package.version}}
+      expected-commit: 08c731f67fddff47b13aad83fafc2d4e745a72cd
+
+  - name: copy conf
+    runs: |
+      mkdir -p ${{targets.contextdir}}/etc/kubernetes/node-feature-discovery/
+      cp ./deployment/components/worker-config/nfd-worker.conf.example ${{targets.contextdir}}/etc/kubernetes/node-feature-discovery/nfd-worker.conf
+
+  - uses: go/build
+    with:
+      modroot: .
+      packages: ./cmd/kubectl-nfd
+      output: kubectl-nfd
+      tags: osusergo,netgo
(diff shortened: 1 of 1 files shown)


node-feature-discovery-0.15.yaml
```diff
@@ -0,0 +1,94 @@
+package:
+  name: node-feature-discovery-0.15
+  version: 0.15.4
+  epoch: 0
+  description: Node feature discovery for Kubernetes
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - protoc
```
jamonation, line 10: Do these binaries need protoc at runtime?

node-feature-discovery-0.15.yaml
```diff
@@ -0,0 +1,94 @@
...
+    provides:
+      - node-feature-discovery=${{package.full-version}}
+
+environment:
+  contents:
+    packages:
+      - protoc
+  environment:
+    GRPC_GO_LOG_SEVERITY_LEVEL: "INFO"
```
EyeCantCU, line 19: This should be removed. It's set for the environment in the image. Setting this in the build environment won't do anything

node-feature-discovery-0.15.yaml
EyeCantCU, line 10: This was only used at buildtime. This should be able to be removed

node-feature-discovery-0.15.yaml
```diff
@@ -0,0 +1,94 @@
...
+      repository: https://github.com/kubernetes-sigs/node-feature-discovery
+      tag: v${{package.version}}
+      expected-commit: 08c731f67fddff47b13aad83fafc2d4e745a72cd
+
+  - name: copy conf
+    runs: |
+      mkdir -p ${{targets.contextdir}}/etc/kubernetes/node-feature-discovery/
+      cp ./deployment/components/worker-config/nfd-worker.conf.example ${{targets.contextdir}}/etc/kubernetes/node-feature-discovery/nfd-worker.conf
+
+  - uses: go/build
```
EyeCantCU, line 33: The image just throws all of these into /usr/bin so we should be good without splitting these up into subpackages

node-feature-discovery-0.15.yaml
```diff
@@ -0,0 +1,94 @@
...
+update:
+  enabled: true
+  github:
+    identifier: kubernetes-sigs/node-feature-discovery
+    strip-prefix: v
+    use-tag: true
+    tag-filter: v0.15.
+
+test:
+  pipeline:
```
EyeCantCU, line 89: Is there a functional test we can implement outside of retrieving the binary versions/help info?

EyeCantCU (changes requested):

node-feature-discovery-0.15.yaml
Pyronewbic: Unfortunately, no, since they all require a kubernetes cluster to be running with env values. Just checked.

node-feature-discovery-0.15.yaml
Pyronewbic: Removed!

node-feature-discovery-0.15.yaml
Pyronewbic: Removed it!

node-feature-discovery-0.15.yaml
Pyronewbic: Nope, it's been removed.

EyeCantCU (approved):

PR #16704, opened 2024-04-11
openssl: Add smoke test cases to prevent openssl missbuilt breaking Wolfi

xnox:
Test that wget, curl, and git still work after borked openssl updates.

This is in response to failures triggered by #16623

### DIFF
--- a/openssl.yaml
+++ b/openssl.yaml
@@ -1,7 +1,7 @@
 package:
   name: openssl
   version: 3.3.0
-  epoch: 4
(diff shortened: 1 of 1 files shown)


xnox:
Recreating original bad landing in https://github.com/wolfi-dev/os/pull/16705 on top of these new tests added here shows that it catches the ABI breaking change (even when there are no other packages pending build).

ajayk (approved):

PR #16707, opened 2024-04-11
Bump pkgconf and switch to a git clone instead of fetch.

dlorenc:
~This requires using the bootstrap repo to avoid a circular dependency, because pkgconf depends on itself.~

### DIFF
--- a/pkgconf.yaml
+++ b/pkgconf.yaml
@@ -1,6 +1,6 @@
 package:
   name: pkgconf
-  version: 2.1.1
+  version: 2.2.0
   epoch: 0
   description: "An implementation of pkg-config"
   copyright:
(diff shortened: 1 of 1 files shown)


kaniini:
pkgconf shouldn’t depend on itself (we go
to some level of effort to ensure it is bootstrappable), but you will need autotools to build the configure script.

dlorenc:
> pkgconf shouldn’t depend on itself (we go to some level of effort to ensure it is bootstrappable), but you will need autotools to build the configure script.

I couldn't get it to build without pkgconf-dev, but maybe it's something else in there that's getting pulled in.

kaniini:
weird, let me check the build log

dlorenc:
> weird, let me check the build log

I got it! Thanks!

imjasonh (approved):

kaniini:
Another option worth looking into would be to package muon, then you wouldn’t need autoconf at all.  Meson itself is too dependency heavy on its own for bootstrapping pkgconf.

Issue #16698, opened 2024-04-11
Dropping OpenSSL engines support, DSA, DES


xnox:
- [ ] Rebuild packages with 
```
CFLAGS='-DOPENSSL_NO_ENGINE -DOPENSSL_NO_DSA -DOPENSSL_NO_DES'
```
- [ ] Verify no reverse deps use them.

- [ ] Then rebuild openssl with those things turned off.

xnox:
Binaries using DES

```
./unpacked-cmake-3.29.0-r0.apk/usr/bin/ccmake
./unpacked-cmake-3.29.0-r0.apk/usr/bin/cmake
./unpacked-cmake-3.29.0-r0.apk/usr/bin/cpack
./unpacked-cmake-3.29.0-r0.apk/usr/bin/ctest
./unpacked-wget-1.24.5-r0.apk/usr/bin/wget
./unpacked-stunnel-5.72-r0.apk/usr/bin/stunnel
./unpacked-mariadb-backup-10.6.12-r1.apk/usr/bin/mariadb-backup
./unpacked-mariadb-11.2-embedded-11.2.3-r2.apk/usr/bin/mariadb-embedded
./unpacked-openssl-provider-legacy-3.3.0-r5.apk/usr/lib/ossl-modules/legacy.so
./unpacked-mariadb-10.11-10.11.7-r1.apk/usr/bin/test-connect-t
./unpacked-mariadb-10.11-10.11.7-r1.apk/usr/bin/mariadbd
./unpacked-mariadb-10.11-10.11.7-r1.apk/usr/lib/libmariadbd.so.19
./unpacked-mariadb-11.2-backup-11.2.3-r2.apk/usr/bin/mariadb-backup
./unpacked-mariadb-10.6-10.6.17-r1.apk/usr/bin/test-connect-t
./unpacked-mariadb-10.6-10.6.17-r1.apk/usr/bin/mariadbd
./unpacked-mariadb-10.6-10.6.17-r1.apk/usr/lib/libmariadbd.so.19
./unpacked-mariadb-test-10.11.7-r1.apk/usr/bin/mariadb-client-test-embedded
./unpacked-mariadb-test-10.11.7-r1.apk/usr/bin/mariadb-test-embedded
./unpacked-net-snmp-5.9.4-r0.apk/usr/lib/libsnmp.so.40.2.1
./unpacked-mariadb-10.11-embedded-10.11.7-r1.apk/usr/bin/mariadb-embedded
[...]

PR #16713, opened 2024-04-11
disable NTLM in curl & wget

xnox:
NTLM authentication relies on broken DES cryptography. Disable ntlm authentication in curl and wget, such that DES support can be removed from our builds of openssl.

Also build curl without deprecated engine support, such that engines can be turned off in openssl as well.

Note NTLM authentication already is not possible under FIPS.

Also see:

https://techcommunity.microsoft.com/t5/windows-it-pro-blog/the-evolution-of-windows-authentication/ba-p/3926848

Related: #16698

### DIFF
--- a/curl.yaml
+++ b/curl.yaml
@@ -1,7 +1,7 @@
 package:
   name: curl
   version: 8.9.1
-  epoch: 0
+  epoch: 1
   description: "URL retrieval utility and library"
   copyright:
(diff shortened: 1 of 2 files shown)


mamccorm:
Discussed with @xnox via 1-1, also comment is updated

mamccorm (approved):

PR #16722, opened 2024-04-11
nodejs-21: Add nodejs provide for nodejs

pdeslaur:
`nodejs-21` should provide `nodejs` (but not `nodejs-lts`!)


amdawson:
note that we should plan to release the nodejs-22 version stream (not LTS) on April 23 https://github.com/nodejs/release#release-schedule

smoser (approved):

PR #16763, opened 2024-04-12
openssl: disable des dsa engine support for newly built packages only

xnox:
Keep DES DSA ENGINE support in the library runtime ABI for existing packages.

Disable DES DSA ENGINE support, in openssl-dev headers only. Such that newly built software stops using deprecated features at runtime.

This should be a safer way to land #16623

The package tests for curl / wget should verify the above runtime guarantee. Follow up wolfictl bump of curl / wget should automatically drop DES DSA ENGINE support (will submit as a separate PR).

### DIFF
--- a/cmake.yaml
+++ b/cmake.yaml
@@ -1,7 +1,7 @@
 package:
   name: cmake
   version: 3.29.0
-  epoch: 3
+  epoch: 4
   description: "CMake is an open-source, cross-platform family of tools designed to build, test and package software"
   dependencies:
     provider-priority: 10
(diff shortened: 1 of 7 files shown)


mepi262:
@xnox 
I could understand current situation.
Thank you for your reply!

xnox:
> @xnox I could understand current situation. Thank you for your reply!

reviving this a bit; some package updates have landed; and we can land a few more stand alone.

xnox:
a hardened version of openssl-dev has now landed.

PR #16982, opened 2024-04-16
Rebuild previous version

mamccorm:
Once-off change to re-build previous revision

### DIFF
--- a/conda-base.yaml
+++ b/conda-base.yaml
@@ -1,6 +1,6 @@
 package:
   name: conda-base
-  version: 24.3.0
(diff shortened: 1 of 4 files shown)


conda-base.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: conda-base
-  version: 24.3.0
+  version: 24.1.1
```
mamccorm, line 3: there was no v24 of this package, so nothing to add to withdrawn packages btw

EyeCantCU (approved):

PR #17421, opened 2024-04-22
iptables: split xtables multi-call binaries into standalone subpackage

xnox:
This will allow to add a new iptables-wrapper package, which also installs `/sbin/iptables` symlink and depends on xtables to only provide the multi-call binaries.


dannf:
superseded by https://github.com/wolfi-dev/os/pull/58325

PR #17120, opened 2024-04-18
add iptables-wrapper

Dentrax:
Fixes:

Related:

#### For new package PRs only

- [ ] This PR is marked as fixing a pre-existing package request bug
  - [ ] Alternatively, the PR is marked as related to a pre-existing package request bug, such as a dependency
- [X] REQUIRED - The package is available under an OSI-approved or FSF-approved license
- [X] REQUIRED - The version of the package is still receiving security updates
- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`)

#### For new version streams

- [ ] The upstream project actually supports multiple concurrent versions.
- [ ] Any subpackages include the version string in their package name (e.g. `name: ${{package.name}}-compat`)
- [ ] The package (and subpackages) `provides:` logical unversioned forms of the package (e.g. `nodejs`, `nodejs-lts`)

#### For security-related PRs

- [ ] The security fix is recorded in the [advisories](https://github.com/wolfi-dev/advisories) repo

#### For version bump PRs

- [ ] The `epoch` field is reset to 0

#### For PRs that add patches

- [ ] Patch source is documented

PR #17491, opened 2024-04-23
iptables-wrapper: new package that uses split-xtables

xnox:
This package provides iptables-wrapper built from master branch as static go binary. As upstream hasn't tagged any releases for over two years now.

iptables-wrapper has symlinks to wrap regular iptables commands to point at the wrapper, which in turn calls appropriate xtables-legacy|nft-multi binaries to actually implement appropriate chains on a given host.

```
f621db237e45:/work/packages# ls -latr /sbin/ip*tables*
-rwxrwxr-x    1 root     root       2068181 Apr 23 09:26 /sbin/iptables-wrapper
lrwxrwxrwx    1 root     root            22 Apr 23 09:29 /sbin/iptables-save -> /sbin/iptables-wrapper
lrwxrwxrwx    1 root     root            22 Apr 23 09:29 /sbin/iptables-restore -> /sbin/iptables-wrapper
lrwxrwxrwx    1 root     root            22 Apr 23 09:29 /sbin/iptables -> /sbin/iptables-wrapper
lrwxrwxrwx    1 root     root            22 Apr 23 09:29 /sbin/ip6tables-save -> /sbin/iptables-wrapper
lrwxrwxrwx    1 root     root            22 Apr 23 09:29 /sbin/ip6tables-restore -> /sbin/iptables-wrapper
lrwxrwxrwx    1 root     root            22 Apr 23 09:29 /sbin/ip6tables -> /sbin/iptables-wrapper
```

Fixes: #17120

Related: #17421


iptables-wrapper.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - iptables
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/kubernetes-sigs/iptables-wrappers
+      branch: master
+      expected-commit: f6ef44b2c449cca8f005b32dea9a4b497202dbef
+
+  - uses: go/build
```
Dentrax, line 27: shouldn't we run `./iptables-wrapper-installer.sh` here?

iptables-wrapper.yaml
```diff
@@ -0,0 +1,45 @@
...
+      packages: .
+      prefix: /
+      install-dir: sbin
+      output: iptables-wrapper
+
+  - runs: |
+      for binary in iptables iptables-restore iptables-save ip6tables ip6tables-restore ip6tables-save
+      do
+        ln -sf /sbin/iptables-wrapper ${{targets.contextdir}}/sbin/$binary
+      done
```
xnox, line 38: @Dentrax this bit of symlinks is the equivalent of what iptables-wrapper-installer.sh would do.

Note, we need the symlinks to be in `${{targets.contextdir}}/sbin`, not in '/sbin' on the builder host.

dannf:
superseded by https://github.com/wolfi-dev/os/pull/17120

PR #17512, opened 2024-04-23
zfp package created

klrmngr:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/zfp.yaml
+++ b/zfp.yaml
@@ -1,0 +1,38 @@
+package:
+  name: zfp
+  version: 1.0.1
+  epoch: 0
+  description: zfp is a compressed format for representing multidimensional floating-point and integer arrays.
+  copyright:
+    - license: BSD-3-Clause
+
+environment:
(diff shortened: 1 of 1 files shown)


zfp.yaml
```diff
@@ -0,0 +1,37 @@
...
+  - uses: cmake/build
+
+  - uses: cmake/install
+
+  - uses: strip
+
+update:
+  enabled: true
+  github:
+    identifier: LLNL/zfp
```
jamonation, line 37: I'd like to have a test here showing the library/binary in use. I think the cmake command creates a test script - perhaps you can add it in the test pipeline and check the results? There are a decent number of other packages that use the test pipeline to copy examples from

kz3640: Added this

jamonation (approved):

PR #17537, opened 2024-04-23
Add a bash-binsh package to have bash provide /bin/sh

smoser:
If you have bash installed, then you can have it be /bin/sh also.  Previously if you had bash in an image and wanted a /bin/sh you would need to install another shell.

This adds bash-binsh package to provide the /bin/sh symlink.

Fixes:

Related:

#### For new package PRs only

- [ ] This PR is marked as fixing a pre-existing package request bug
  - [ ] Alternatively, the PR is marked as related to a pre-existing package request bug, such as a dependency
- [ ] REQUIRED - The package is available under an OSI-approved or FSF-approved license
- [ ] REQUIRED - The version of the package is still receiving security updates
- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`)

#### For new version streams

- [ ] The upstream project actually supports multiple concurrent versions.
- [ ] Any subpackages include the version string in their package name (e.g. `name: ${{package.name}}-compat`)
- [ ] The package (and subpackages) `provides:` logical unversioned forms of the package (e.g. `nodejs`, `nodejs-lts`)

#### For security-related PRs

- [ ] The security fix is recorded in the [advisories](https://github.com/wolfi-dev/advisories) repo

#### For version bump PRs

- [ ] The `epoch` field is reset to 0

#### For PRs that add patches

- [ ] Patch source is documented

PR #17554, opened 2024-04-23
Bumping bash to see if it fails

jonjohnsonjr:


smoser:
The issue that was being reproduced here resulted in a change in melange at https://github.com/chainguard-dev/melange/pull/1173 .
I'm going to close _this_ issue, re-open if its needed. we have #17537 also to reproduce that.

PR #17557, opened 2024-04-23
openssl: make openssl-config pure virtual package

xnox:
Make openssl-config a pure virtual package. Set provider-priority on the openssl-config-default. This allows to build
openssl-config-fipshardened separately, with a matching provides that will always work.

Fixes: inability to seamlessly install openssl-config-fipshardened when provides versions drift.

### DIFF
--- a/openssl.yaml
+++ b/openssl.yaml
@@ -1,7 +1,7 @@
 package:
   name: openssl
   version: 3.3.0
-  epoch: 5
+  epoch: 6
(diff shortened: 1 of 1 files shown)


xnox:
Hm, need to test that image building works correctly.

xnox:
no idea why this was fixing things; and the reproducer just works now. so who knows what was broken.

PR #17570, opened 2024-04-24
python: fix MD5 nonsecurity usage with OpenSSL base+fips providers

### LINKED ISSUE #118224: Fallback md5 is used, when trying hard to only use OpenSSL
# Bug report

### Bug description:

When
* OpenSSL is configured in FIPS mode
* recommended config is used to only load "base + fips" providers
* without the default provider
* CPython is compiled with `--with-builtin-hashlib-hashes=blake2` to exclude fallback implementation of MD5

upon importing hashlib fails to create MD5 construct.

```python
# python3.10 -c 'import hashlib'
ERROR:root:code for hash md5 was not found.
  File "/usr/lib/python3.10/hashlib.py", line 137, in __get_openssl_constructor
    f(usedforsecurity=False)
ValueError: [digital envelope routines] unsupported

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.10/hashlib.py", line 261, in <module>
    globals()[__func_name] = __get_hash(__func_name)
  File "/usr/lib/python3.10/hashlib.py", line 141, in __get_openssl_constructor
    return __get_builtin_constructor(name)
  File "/usr/lib/python3.10/hashlib.py", line 123, in __get_builtin_constructor
    raise ValueError('unsupported hash type ' + name)
ValueError: unsupported hash type md5
```

Reference implementation is upstream openssl 3.3.0, with enable-fips, fipsinstall completed and openssl.cnf set to
```
# cat /etc/ssl/openssl.cnf 
config_diagnostics = 1
openssl_conf = openssl_init

.include /etc/ssl/fipsmodule.cnf

[openssl_init]
providers = provider_sect
alg_section = algorithm_sect

[provider_sect]
fips = fips_sect
base = base_sect

[base_sect]
activate = 1

[algorithm_sect]
default_properties = fips=yes
```

In essence, things work well only when "default + fips" providers are loaded, as then MD5 functions in OpenSSL are detected as available and are used at runtime and correctly get blocked.

When only "base + fips" providers are loaded, ValueError is raised by OpenSSL constructor, and instead fallback implementation used from _md5 module if it was compiled in.

It seems like the above configuration was not tested, however it can be made to work. CPython should try to load  the "default" OpenSSL provider, to guarantee access to non-fips hashes.

### Security concerns

This is FedRAMP/FIPS compliance by-pass. This issue may allow using md5 without specifying "usedforsecurity=False" on systems otherwise configured to be in FIPS-mode only. And is the primary reason why documentation mentions that certain distributors of python remove md5 module altogether.

### CPython versions tested on:

3.10, 3.11, 3.12

### Operating systems tested on:

Linux

### Linked PRs
* gh-118236
* gh-118238
* gh-118239
* gh-118240
* gh-118264
* gh-127300

xnox:
Compile python without hashlib fallbacks for NIST algorithms, as they are slower and allow to bypass OpenSSL implementations in certain conditions.

Fix OpenSSL base and fips provider usage for unapproved algorithms, such that one can use MD5 for non-security. And vice versa.

Fixes: https://github.com/python/cpython/issues/118224

Patches are backports linked in the above upstream issue.

With this change python-fips releases 3.10..3.12 now work.

CVE checks is failing on 3.10 as there is a medium CVE without an upstream patch that we self-fixed in 3.11+

This removes unused builtin non-hw accelerated cryptographic implementations, to instead always use OpenSSL as was always the case for vanilla builds. FIPS builds, depending on OpenSSL configuration, had opportunity to missuse the fallback in unapproved ways:

- 28.0K	_md5.cpython-312-x86_64-linux-gnu.so
- 28.0K	_sha1.cpython-312-x86_64-linux-gnu.so
- 64.0K	_sha2.cpython-312-x86_64-linux-gnu.so
- 40.0K	_sha3.cpython-312-x86_64-linux-gnu.so

### DIFF
--- a/python-3.10.yaml
+++ b/python-3.10.yaml
@@ -1,7 +1,7 @@
 package:
   name: python-3.10
   version: 3.10.14
-  epoch: 0
+  epoch: 1
   description: "the Python programming language"
   copyright:
     - license: PSF-2.0
@@ -62,6 +62,10 @@ pipeline:
     with:
       patches: CVE-2023-27043-enable-disable.patch
 
+  - uses: patch
+    with:
+      patches: gh-118224.patch
+
   - name: Configure
     runs: |
       ./configure \
@@ -80,6 +84,7 @@ pipeline:
          --with-system-ffi \
          --with-system-libmpdec \
          --without-ensurepip \
+         --with-builtin-hashlib-hashes=blake2 \
          --with-lto
 
   - uses: autoconf/make
--- a/python-3.10/gh-118224.patch
+++ b/python-3.10/gh-118224.patch
@@ -1,0 +1,80 @@
+From 1c1e963118bf28c86c72c6bb36fe500a372b4d0b Mon Sep 17 00:00:00 2001
+From: Dimitri John Ledkov <dimitri.ledkov@surgut.co.uk>
+Date: Wed, 24 Apr 2024 14:47:10 +0100
+Subject: [PATCH 1/2] [3.10] gh-118224: Load default OpenSSL provider for
+ nonsecurity algorithms
+
+When OpenSSL is configured to only load "base+fips" providers into the
+Null library context, md5 might not be available at all. In such cases
+currently CPython fallsback to internal hashlib implementation is
+there is one - as there might not be if one compiles python with
+--with-builtin-hashlib-hashes=blake2. With this change "default"
+provider is attempted to be loaded to access nonsecurity hashes.
+---
+ Modules/_hashopenssl.c | 14 ++++++++++++++
+ 1 file changed, 14 insertions(+)
+
+diff --git a/Modules/_hashopenssl.c b/Modules/_hashopenssl.c
+index 35addf49e96c63..6418e4295babf2 100644
+--- a/Modules/_hashopenssl.c
++++ b/Modules/_hashopenssl.c
+@@ -51,6 +51,7 @@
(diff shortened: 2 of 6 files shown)


xnox:
This is also super ugly, because it results in tracebacks on every hashlib usage:

```
# python3.10 -c 'import hashlib'
ERROR:root:code for hash md5 was not found.
  File "/usr/lib/python3.10/hashlib.py", line 137, in __get_openssl_constructor
    f(usedforsecurity=True)
ValueError: [digital envelope routines] unsupported

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.10/hashlib.py", line 261, in <module>
    globals()[__func_name] = __get_hash(__func_name)
  File "/usr/lib/python3.10/hashlib.py", line 141, in __get_openssl_constructor
    return __get_builtin_constructor(name)
  File "/usr/lib/python3.10/hashlib.py", line 123, in __get_builtin_constructor
    raise ValueError('unsupported hash type ' + name)
ValueError: unsupported hash type md5
```

smoser (approved):
This looks really good, thank you for following up upstream as well as you did.
We'll just have to track those upstream PRs and adjust this to the feedback that is provided there.

@xnox answered one question I had out of band about "base+fips" and "default" with a link to his description https://github.com/chainguard-dev/enterprise-packages/pull/3254 .

Thanks.

PR #17582, opened 2024-04-24
ruby: Update Ruby versions to the latest releases (3.3.1, 3.2.4, 3.1.5, 3.0.7)

dbussink:
These also include the previously manually applied patches (and more additional security fixes as well).

See also:

- https://www.ruby-lang.org/en/news/2024/04/23/ruby-3-3-1-released/
- https://www.ruby-lang.org/en/news/2024/04/23/ruby-3-2-4-released/
- https://www.ruby-lang.org/en/news/2024/04/23/ruby-3-1-5-released/
- https://www.ruby-lang.org/en/news/2024/04/23/ruby-3-0-7-released/

The additional CVE fixed that wasn't addressed yet by custom patches (which are cleaned up here) is:

https://www.ruby-lang.org/en/news/2024/04/23/arbitrary-memory-address-read-regexp-cve-2024-27282/

#### For security-related PRs

Is this something I should also try to send a pull request for? 

#### For version bump PRs


ruby-3.0.yaml
```diff
@@ -18,6 +18,7 @@ environment:
       - gdbm-dev
       - git
       - libffi-dev
+      - libxcrypt-dev
```
dbussink, line 21: Added this, since I saw this failure somewhere else in our usage:

```
 /usr/bin/ruby: /lib/libcrypt.so.1: version `XCRYPT_2.0' not found (required by /usr/lib/libruby.so.3.2)
```

ajayk: Can we remove this and test if it is really needed here

dbussink: Sure, but how would that other error be happening then? We had to manually add libxcrypt to make it work so an explicit dependency seems better?

dbussink: Removed it for now if you think it's not needed.

dbussink: Fwiw, you can see Ruby use it here: https://github.com/ruby/ruby/blob/master/string.c#L48-L55, but it will fall back to a bundled wrapper if it's not found.

So it's not really I think a question of if it works (it probably will), but what is desired here? 

> can you send us a reproducible test where you are seeing this error , even if we add `libxcrypt-dev` this should be a runtime dependency

This is for the existing packages btw. If you manually install the existing Ruby packages on a newer base, it will break if you don't install `libxcrypt` since when the existing Ruby packages were built, `glibc` still provided `libcrypt` while now it's in `libxcrypt`.

ajayk: @dbussink thanks for looking into this glibc still provides libxcrypt through libcrypt1 check commit [here](https://github.com/wolfi-dev/os/commit/abdeb31b9520abcefdbd3e980b108e8777ec9629)

@xnox for further clarification incase if I am wrong

ajayk: from the build logs  for the PR 

```
2024-04-24T13:18:19.7705477Z 2024/04/24 13:18:19 INFO checking for dynamic size alloca... ok
2024-04-24T13:18:19.8351479Z 2024/04/24 13:18:19 INFO checking for working memcmp... yes
2024-04-24T13:18:19.8742227Z 2024/04/24 13:18:19 INFO checking for dup... yes
2024-04-24T13:18:19.9132244Z 2024/04/24 13:18:19 INFO checking for dup2... yes
2024-04-24T13:18:19.9578986Z 2024/04/24 13:18:19 INFO checking for acosh... yes
2024-04-24T13:18:20.0026931Z 2024/04/24 13:18:20 INFO checking for cbrt... yes
2024-04-24T13:18:20.0416904Z 2024/04/24 13:18:20 INFO checking for crypt... yes

```

xnox: > Added this, since I saw this failure somewhere else in our usage:
> 
> ```
>  /usr/bin/ruby: /lib/libcrypt.so.1: version `XCRYPT_2.0' not found (required by /usr/lib/libruby.so.3.2)
> ```

If you saw this message, it means (accidentally or on purpose) and obsolete package version was installed possibly in the obsolete container. As in libcrypt1 not up to date at runtime (nothing to do with build-time, build-time was up to date).

Where did you see it? what was the contents of `/etc/apk/world`? was the container up to date (i.e. did you pull latest sdk; wolfi; base container).

dbussink: @xnox The Ruby package might have been older, I'll check. 

But if this works without here for building, I think that's separate from that issue then. Rather would move forward then with the fixed stuff and the updates like in this PR :smile:.

xnox: yeah it was entirely redundant as it is already present there anyway

```
2024-04-24T13:17:57.1973200Z 2024/04/24 13:17:57 INFO installing libxcrypt (4.4.36-r4)
2024-04-24T13:17:57.1978636Z 2024/04/24 13:17:57 INFO installing libxcrypt-dev (4.4.36-r4)
```

and the error you showed was a runtime one (not build environment one) and could only have happened on out of date containers/hosts. so all is good.

xnox (approved):

xnox:
> 2024/04/24 13:16:20 INFO error during command execution: package ruby-3.2: update found newer version 3.3.1 compared with package.version 3.2.4 in melange config

Some streams updater things are not setup correctly it seems. Maybe something to fix later.

PR #17679, opened 2024-04-25
font-liberation: use release download rather than file link

tstromberg:
I'm adding a wolfictl rule to forbid /file/ links, as they include comment attachments. With /file/ links, it's impossible to tell between an officially generated tarball and one that a random person on the Internet attached to an issue.

For more info, see https://www.mcafee.com/blogs/other-blogs/mcafee-labs/redline-stealer-a-novel-approach/


mepi262:
@tstromberg 
It seems that some check are not successful.
Would you check this pull request?

PR #17902, opened 2024-04-26
fix(py3-itsdangerous): Fixup packaging, make it not so dangerous

EyeCantCU:
This was previously installing to the root of the build env and not to the package itself


smythp:
Fast move on this, thanks RJ! 🚀

joshrwolf (approved):

PR #17964, opened 2024-04-28
fix(rust): Build jemalloc under aarch64 with 64k page support

EyeCantCU:
While running an image I built under Asahi Remix on an MBP with an M2 Max, I ecountered this error:

`<jemalloc>: Unsupported system page size`

Support 8k, 16k, and 64k aligned processors under aarch64 by setting the page size to 64k. By default, jemalloc pulls the page size from the builder, but this results in binaries that can't be ran universally

### DIFF
--- a/build-aarch64.env
+++ b/build-aarch64.env
@@ -5,3 +5,5 @@ export CXXFLAGS="$CFLAGS"
 export LDFLAGS="-Wl,--as-needed,-O1,--sort-common -Wl,-z,relro,-z,now,-z,noexecstack"
 export GOFLAGS=""
 export GOTOOLCHAIN=local
+# Build jemalloc with 64k page support
+export JEMALLOC_SYS_WITH_LG_PAGE=16


rust.yaml
```diff
@@ -44,6 +44,11 @@ pipeline:
       rm rustc-${{package.version}}-src.tar.xz
 
   - runs: |
+      # Build aarch64 with 64k page support
+      if [[ "${{build.arch}}" == "aarch64" ]]; then
+        export JEMALLOC_SYS_WITH_LG_PAGE=64
+      fi
```
xnox, line 50: does this make binaries universal? and they still run correctly on all lower page sizes too?

EyeCantCU: Anything using 4k (MacOS), 8k, 16k, and 64k page sizes will work. I believe it currently defaults to 4k. This covers everything

We'll need to rebuild all our binaries too (I want to move everything to our cargo/build pipeline today so that will handle that). A small number of projects override this and set their own value so we'll also need to export it from the cargo/build pipeline too just in case

xnox: Linux typically ships kernels in two configurations - so my question was specifically about Linux. I.e. Ubuntu Rhel SUSE all offer 4k and 64k kernels, and support running systems with either. Thus I hope with this option set our binaries/containers will work with either kernels.

Have you tested this change on both generic & generic64k kernels on any of the Linux distros in GCP or AWS? If not, do you want to do this yourself, or shall I?

xnox:
```
* `JEMALLOC_SYS_WITH_LG_PAGE=<lg-page>`: Specify the base 2 log of the allocator
  page size, which must in turn be at least as large as the system page size. By
  default the configure script determines the host's page size and sets the
  allocator page size equal to the system page size, so this option need not be
  specified unless the system page size may change between configuration and
  execution, e.g. when cross compiling.
```

Are you sure you want this in rust.yaml only, and not in `build-aarch64.env` such that it is active whenever any package calls cargo and builds jemallocator create?

EyeCantCU: I've tested with 4k and 8k. Happy to cover all grounds here though and throw 64k into the mix

I'm very relieved you shared this. I didn't realize from what I was reading earlier that it's log2(allocator size)... We need to set this to 16, not 64 (though it worked 🤷‍♂️). I think that's a much better idea than placing it here

EyeCantCU: Tested this with `fd`, and I can confirm building with this set to 16 produces a binary that runs across systems that use 4k, 8k, 16k, and 64k page sizes set in the kernel

I guess the larger question here is... should we just outright set this to 16 for both x64 and aarch64?

EyeCantCU: PR updated to move this over to build-aarch64.env as that's much better than handling this in the rust package and in our cargo/build pipeline

xnox (approved):

PR #18021, opened 2024-04-29
Make Jenkins agent its own subpackage

johnfosborneiii:
This PR allows users to leverage the Jenkins Agent embedded in the Jenkins.war file as opposed to pulling down the prebuilt binary (as in the official image):
https://github.com/jenkinsci/docker-agent/tree/master

Jenkins Agent and Slave are the same executables so this allows for the .jar to be used both ways (similar to upstream):
[Agent executables (remoting.jar is named agent.jar or slave.jar in Jenkins)](https://www.jenkins.io/projects/remoting/)
https://github.com/jenkinsci/docker-agent/blob/master/alpine/Dockerfile#L77

### DIFF
--- a/jenkins.yaml
+++ b/jenkins.yaml
@@ -59,9 +59,38 @@ pipeline:
       mvn install -DskipTests=true
 
       mkdir -p ${{targets.destdir}}/usr/share/java/jenkins
-      mv war/target/jenkins.war ${{targets.destdir}}/usr/share/java/jenkins/
+      cp -p war/target/jenkins.war ${{targets.destdir}}/usr/share/java/jenkins/
 
 subpackages:
+  #
+  # Agent executables (remoting.jar is named agent.jar or slave.jar in Jenkins)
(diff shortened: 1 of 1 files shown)


jamonation:
@johnfosborneiii is this different from the https://github.com/jenkinsci/remoting project? Ideally we'd be able to package that using `mvn` as opposed to exploding the .war file.

johnfosborneiii:
> is this different from the https://github.com/jenkinsci/remoting project? Ideally we'd be able to package that using mvn as opposed to exploding the .war file.

My understanding is that it's exactly the same. Users can also download the agent from Jenkins Server which is why there is one embedded in the war lib folder. 

My rationale was that this way prevents duplicative builds and ensures compatibility with our Jenkins image but I'm not opposed to completely separate the build. Let me know which approach you prefer.

johnfosborneiii:
Closing in favor of https://github.com/wolfi-dev/os/pull/18064

PR #18115, opened 2024-05-01
opentelemetry-collector: update build config

cmdpdx:
Use `manifest.yaml` from `open-telemetry/opentelemetry-collector-releases` repo for the `build-config.yaml` provider to the builder so the resulting `otelcol` binary matches what is found in `open-telemetry/opentelemetry-collector-releases`

### DIFF
--- a/opentelemetry-collector.yaml
+++ b/opentelemetry-collector.yaml
@@ -1,7 +1,7 @@
 package:
   name: opentelemetry-collector
   version: 0.99.0
-  epoch: 1
+  epoch: 2
   description: OpenTelemetry Collector
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


opentelemetry-collector.yaml
```diff
@@ -1,7 +1,7 @@
 package:
```
jamonation, line 1:
```suggestion
#nolint:valid-pipeline-git-checkout-commit
package:
```

jamonation: I think this will get the checkout pipeline to pass

cmdpdx: Is `nolint` preferable over `curl`ing the file? I made that change locally to test. This manifest file gets updated during release so it should match the latest version from the main repo.

jamonation: Oh I guess curling would make sense - in either case if something breaks we'll know about it, and curl saves on the whole repo clone

jamonation (approved):

PR #18122, opened 2024-05-01
add libnvidia-container and lsb-release-minimal

Dentrax:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/libnvidia-container.yaml
+++ b/libnvidia-container.yaml
@@ -1,0 +1,83 @@
+package:
+  name: libnvidia-container
+  version: 1.15.0
+  epoch: 0
+  description: NVIDIA container runtime library
+  copyright:
+    - license: Apache-2.0
+    - license: GPL-3.0-only
+    - license: LGPL-3.0-or-later
+
+environment:
+  contents:
+    packages:
+      - automake
+      - bmake
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - coreutils # for GNU date
+      - curl
+      - glibc-dev
+      - go
+      - libcap-dev # for sys/capability.h
+      - libseccomp
+      - libseccomp-dev
+      - libtirpc-dev # for rpc/rpc.h
+      - libtool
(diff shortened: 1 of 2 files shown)


libnvidia-container.yaml
```diff
@@ -0,0 +1,83 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/NVIDIA/libnvidia-container
+      tag: v${{package.version}}
+      expected-commit: 6c8f1df7fd32cea3280cf2a2c6e931c9b3132465
+
+  # Fix prefix to /usr instead of /usr/local
+  - runs: sed -i 's|export prefix *= */usr/local|export prefix = /usr|' Makefile
```
rawlingsj, line 45: just checking, does `- uses: autoconf/configure` work here instead of using `sed`?

Dentrax: Unfortunately, repo doesn't provide a `./configure` file in root dir.

libnvidia-container.yaml
```diff
@@ -0,0 +1,83 @@
...
+      runs: |
+        ldd /usr/lib/libnvidia-container.so
+        ldd /usr/lib/libnvidia-container-go.so
+    - name: Test nvidia-container-cli
+      runs: nvidia-container-cli --version
+    - name: Test nvidia-container-cli with privileged mode
+      runs: |
+        set +e
+        useradd nvidia
+        # This is the expected error since `libnvidia-ml.so` must be available on the host system. Library can be found in the CUDA.
```
jamonation, line 82: Could the package that provides this be added? I think there's value in an affirmative test that can prove this package is able to load its dependencies.

Dentrax: We're going to handle this case in the image. 👍

jamonation (approved):

PR #18132, opened 2024-05-01
Improve license info on glibc & gcc

xnox:
Ideally need ability to inject license files into spdx. Ideally need ability to set a different license on per-subpackage basis.

Related: https://github.com/chainguard-dev/melange/issues/1007


smoser (approved):
This is fine. generally improvement.

PR #18133, opened 2024-05-01
Correct linux-headers license

xnox:
it's gpl-v2 only, with syscall exception making it available to all

note this package only ships files under the syscall exception.


smoser (approved):
thanks. lgtm.

PR #18158, opened 2024-05-01
readline: don't use -Wl,--as-needed

pnasrat:
Fixes explicit ncursesw linking

### DIFF
--- a/readline.yaml
+++ b/readline.yaml
@@ -1,7 +1,7 @@
 package:
   name: readline
   version: "8.2"
-  epoch: 3
+  epoch: 4
   description: "GNU readline library"
   copyright:
     - license: GPL-3.0-or-later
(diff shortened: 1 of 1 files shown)


pnasrat:
cc @imjasonh

pnasrat:
You can see the linking worked when it checks it can be installed and pulls in ncurses

https://github.com/wolfi-dev/os/actions/runs/8913595150/job/24479504501?pr=18158#step:9:116

imjasonh (approved):

readline.yaml
imjasonh, line 1: Need to increase the epoch to 4 to rebuild with this change

imjasonh (changes requested):

readline.yaml
pnasrat: Force pushed with epoch bump

imjasonh (approved):

PR #18160, opened 2024-05-01
zig.yaml: convert from fetch to git-checkout

xnox:
Convert zig.yaml from fetch to git-checkout


xnox:
version upgrade is being worked on in https://github.com/wolfi-dev/os/pull/17217

joshrwolf (approved):

joshrwolf (dismissed):

PR #18194, opened 2024-05-01
add datadog-agent-s6-overlay subpackage

joshrwolf:
I don't love this approach, and am open to suggestions. Before relying on this, I first tried with no success:

- using the embedded `s6` in apko via `service-bundle`
- using the recent 3.x `s6-overlay`

The level of indirections baked into the `datadog-agent` startup scripts is just too error prone to be comfortable with anything other than using the same s6 setup. This is the only thing I've got successfully working in the image.

I chose to just fetch the release directly from `s6-overlay` rather than build a `s6-overlay-2` package because the [build process](https://github.com/just-containers/s6-overlay/blob/v2.2.0.3/builder/build-latest) looks like it simply just curls from other sources, which is a lot of onion peeling for questionable gains.

### DIFF
--- a/datadog-agent.yaml
+++ b/datadog-agent.yaml
@@ -124,6 +124,7 @@ subpackages:
       runtime:
         - bash
         - coreutils
+        - datadog-agent-s6-overlay
     pipeline:
       - runs: |
           mkdir -p ${{targets.subpkgdir}}/opt/datadog-agent/embedded/bin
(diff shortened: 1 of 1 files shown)


datadog-agent.yaml
```diff
@@ -218,6 +218,22 @@ subpackages:
...
+  # datadog-agent subpackage, only to be used by image builders.
+  - name: datadog-agent-s6-overlay
+    description: "Deprecated s6-overlay for datadog-agent"
+    options:
+      no-provides: true
+    pipeline:
+      - runs: |
+          S6_VERSION=v2.2.0.3
+          JUST_CONTAINERS_DOWNLOAD_LOCATION=https://github.com/just-containers/s6-overlay/releases/download
+          S6ARCH=$([ "$TARGETARCH" = "amd64" ] && echo "amd64" || echo "aarch64") && curl -L ${JUST_CONTAINERS_DOWNLOAD_LOCATION}/${S6_VERSION}/s6-overlay-${S6ARCH}.tar.gz -o s6.tgz
```
dustinkirkland, line 233: Uh, this is kind of yucky...  At the very least, I think we need to verify a checksum of this download against a known good checksum included here in this packaging...

joshrwolf: good call, added a checksum validation step 🙏

jamonation (approved):

PR #18321, opened 2024-05-02
ruby3.2-sinatra.yaml: convert from fetch to git-checkout

xnox:
Convert ruby3.2-sinatra.yaml from fetch to git-checkout

### DIFF
--- a/ruby3.2-sinatra.yaml
+++ b/ruby3.2-sinatra.yaml
@@ -2,7 +2,7 @@
 package:
   name: ruby3.2-sinatra
   version: 3.1.0
-  epoch: 1
+  epoch: 2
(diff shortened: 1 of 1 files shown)


xnox:
> 2024/05/08 14:40:39 INFO error during command execution: package ruby3.2-sinatra: update found newer version 4.0.0 compared with package.version 3.1.0 in melange config

but i would rather land packaging changes separately from new major version upgrade. I would rather bot open that separately please.

xnox:
2024/05/08 14:40:39 INFO error during command execution: package ruby3.2-sinatra: update found newer version 4.0.0 compared with package.version 3.1.0 in melange config

would rather have that submitted separately.

imjasonh (approved):

PR #18374, opened 2024-05-02
Add test/daemon-check-output pipeline and use it.

smoser:
I recently had added a test to keycloak and guacamole-server which would run a the daemon and then look in output for some given strings.

This adds a reusable piepline to do that which is implemented in shell. Ultimately it'd be nicer to have this in melange proper without any shell dependencies.

Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/couchdb.yaml
+++ b/couchdb.yaml
@@ -77,58 +77,39 @@ test:
         - jq
   pipeline:
     - name: CouchDB Start Server
-      runs: |
-        #!/bin/bash
-
-        mkdir /var/log/couchdb
-
-        # Configure CouchDB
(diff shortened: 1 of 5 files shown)


smoser:
The thing this really needs is some "post" check.. to let the caller verify the daemon looks good.

smoser:
> The thing this really needs is some "post" check.. to let the caller verify the daemon looks good.
done.
probably could move move of the tests to this.

joshrwolf (approved):

PR #18492, opened 2024-05-03
test couch test - does not seem to run locally

smoser:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/couchdb.yaml
+++ b/couchdb.yaml
@@ -1,7 +1,7 @@
 package:
   name: couchdb
   version: 3.3.3
-  epoch: 1
+  epoch: 2
   description: Seamless multi-master syncing database with an intuitive HTTP/JSON API, designed for reliability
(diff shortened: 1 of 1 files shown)


smoser:
Just going to summarize something from slack and then close this.
https://chainguard-dev.slack.com/archives/C05CM0DM7RV/p1714749789595289

things to note:
* c-i runs with docker
* mac runs with docker
* linux currently runs with bwrap. but bwrap behaves slightly differently than docker
* cleaner environment of each indivivudal test section might be better but would be breaking to change that this point. would want to do that in a new 'tests' block or some backwards compat way.
* you can use docker locally with `MELANGE_EXTRA_OPTS="--runner docker" make ...`
*

PR #18494, opened 2024-05-03
neuvector-controller-5.3: Added Scripts, Template, Certs

Pyronewbic:
Fixes: 

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches


Pyronewbic:
Controller has no tests - it needs to be in a cluster with the rest of neuvector for it to work

neuvector-controller-5.3.yaml
```diff
@@ -38,9 +39,40 @@ pipeline:
...
+      mkdir -p ${{targets.contextdir}}/etc/tmp
+
+      # Scripts
+      cp scripts/sysctl.conf ${{targets.contextdir}}/etc/
+      cp scripts/teardown.sh ${{targets.contextdir}}/usr/local/bin/
+      cp scripts/runtime-gdb.py ${{targets.contextdir}}/usr/local/bin/
+
+      # Template
+      cp templates/podTemplate.json ${{targets.contextdir}}/etc/neuvector/templates/podTemplate.json
+      wget https://raw.githubusercontent.com/neuvector/manifests/main/build/share/.nvcontainer -P ${{targets.contextdir}}/usr/local/bin/.nvcontainer
```
jamonation, line 56: What is this for? It seems like an empty file?

EyeCantCU: They use it as a way of determining whether or not it's running in a container: https://github.com/neuvector/neuvector/blob/490eda9564b36c78f918ab44db21b6de8eba1d0d/share/system/cgroup_linux.go#L112

I would just use `touch` to create the file

jamonation, line 56:
```suggestion
      touch ${{targets.contextdir}}/usr/local/bin/.nvcontainer
```

jamonation: Thanks for clarifying @EyeCantCU - made a suggestion for you @Pyronewbic

neuvector-controller-5.3.yaml
```diff
@@ -17,6 +17,7 @@ environment:
       - go
       - pcre-dev
       - pcre2-dev
+      - wget
```
jamonation, line 20:
```suggestion
```

jamonation: Hmm wait I see this is used elsewhere, checking

neuvector-controller-5.3.yaml
```diff
@@ -38,9 +39,40 @@ pipeline:
...
+
+      # symlink
+      ln -sf /usr/bin/controller ${{targets.contextdir}}/usr/local/bin/controller
+      ln -sf /usr/bin/opa ${{targets.contextdir}}/usr/local/bin/opa
+      ln -sf /bin/consul ${{targets.contextdir}}/usr/local/bin/consul
+
+      # Certs
+      mkdir -p ${{targets.contextdir}}/etc/neuvector/certs
+      wget https://raw.githubusercontent.com/neuvector/manifests/main/build/share/etc/neuvector/certs/ssl-cert.key -P ${{targets.contextdir}}/etc/neuvector/certs/
+      wget https://raw.githubusercontent.com/neuvector/manifests/main/build/share/etc/neuvector/certs/ssl-cert.pem -P ${{targets.contextdir}}/etc/neuvector/certs/
```
jamonation, line 66: I'd like these to be openssl generated instead of downloaded, even if they're just placeholders. Does the controller 
 need them to exist to run, is there anything specific to the key/cert that requires using their files?

neuvector-controller-5.3.yaml
```diff
@@ -38,9 +39,40 @@ pipeline:
...
   enabled: true
   github:
     identifier: neuvector/neuvector
     strip-prefix: v
     tag-filter: v5.3.
+
+test:
+  pipeline:
+    - runs: |
+        controller --help || true
```
jamonation, line 78:
```suggestion
        controller --help 2>&1 |pcregrep -M 'START.*\nUsage'
```

This will capture the output from the controller and exit 1 if it doesn't find lines like this:
```
2024-05-03T18:26:23.611|INFO|CTL|main.main: START - version=interim/master.xxxx
Usage of controller:
```

Pyronewbic:
`wolfictl scan packages/aarch64/neuvector-controller-5.3-5.3.2-r1.apk
🔎 Scanning "packages/aarch64/neuvector-controller-5.3-5.3.2-r1.apk"
└── 📄 /usr/bin/controller
        📦 github.com/neuvector/neuvector 3.39.4 (go-module)
            Critical GHSA-622h-h2p8-743x fixed in 5.2.2`
            
 is a false positive as I am building 5.3.2. There is no way to change the versioning https://chainguard-dev.slack.com/archives/C062D52J3T9/p1715063965898279

neuvector-controller-5.3.yaml
```diff
@@ -31,16 +33,50 @@ pipeline:
...
+
+      # symlink
+      ln -sf /usr/bin/controller ${{targets.contextdir}}/usr/local/bin/controller
+      ln -sf /usr/bin/opa ${{targets.contextdir}}/usr/local/bin/opa
+      ln -sf /bin/consul ${{targets.contextdir}}/usr/local/bin/consul
+
+      # Retrieve Certs - with changes to openssl I can't generate them anymore.
+      mkdir -p ${{targets.contextdir}}/etc/neuvector/certs
+      wget https://raw.githubusercontent.com/neuvector/manifests/main/build/share/etc/neuvector/certs/ssl-cert.key -P ${{targets.contextdir}}/etc/neuvector/certs/
+      wget https://raw.githubusercontent.com/neuvector/manifests/main/build/share/etc/neuvector/certs/ssl-cert.pem -P ${{targets.contextdir}}/etc/neuvector/certs/
```
xnox, line 70: None of our packages may ship private keys, or certs like these. Because that's insecure, and trips up security scanners.

Please explain the purpose of this key and cert, and why would a private key be the same for all installs and usage of this package? (meaning any install on any system can impersonate any other instance.....)

Shouldn't each deployment of this software deploy with its own self-managed certs?

xnox (changes requested):
It is unclear why shipping a shared private key is required in a package.

neuvector-controller-5.3.yaml
xnox: The openssl.cnf with stock CA settings was removed in https://github.com/wolfi-dev/os/pull/18053 precisely to prevent people from generating unmanaged certificates (i.e. certs without private key management, revocation, transparency logs, revocation urls, etc)

xnox: I would expect a helm chart or vault to provide certificates; or use something like snakeoil; or generate certificates during image or package testing (see for example docker).

xnox: sorry not docker, but mongodb tests.

PR #18518, opened 2024-05-03
rabbitmq-default-user-credential-updater: new package

cmdpdx:
```
 $ wolfictl scan packages/aarch64/rabbitmq-default-user-credential-updater-1.0.4-r0.apk
🔎 Scanning "packages/aarch64/rabbitmq-default-user-credential-updater-1.0.4-r0.apk"
✅ No vulnerabilities found
```

#### For new package PRs only

### DIFF
--- a/rabbitmq-default-user-credential-updater.yaml
+++ b/rabbitmq-default-user-credential-updater.yaml
@@ -25,3 +25,8 @@ update:
   github:
     identifier: rabbitmq/default-user-credential-updater
     strip-prefix: v
+
+test:
+  pipeline:
+    - runs: |
+        default-user-credential-updater -h  2>&1 | grep "Usage of default-user-credential-updater:"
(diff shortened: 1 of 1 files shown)


imjasonh (dismissed):

rabbitmq-default-user-credential-updater.yaml
```diff
@@ -0,0 +1,27 @@
...
+      packages: .
+      output: default-user-credential-updater
+
+  - uses: strip
+
+update:
+  enabled: true
+  github:
+    identifier: rabbitmq/default-user-credential-updater
+    strip-prefix: v
```
jamonation, line 27:
```suggestion
    strip-prefix: v
    
test:
  pipeline:
    - runs: |
        default-user-credential-updater -h  2>&1 | grep "Usage of default-user-credential-updater:"
```

Adding this will run the binary and check for the help string. It isn't much functionally, but it at least shows that the binary runs

cmdpdx: Good call! Added.

jamonation (approved):

PR #18521, opened 2024-05-03
add libva-glx

lpcalisi:
`libva-glx` can not be compiled inside `libva` because a circular dependency with `mesa` (mesa needs libva for compilation and libva with glx needs mesa)

libva-glx and libva have the same source.

Do you think there is a better way to do this?


lpcalisi:
@xnox @imjasonh

lpcalisi:
@ajayk 🏓

xnox:
> `libva-glx` can not be compiled inside `libva` because a circular dependency with `mesa` (mesa needs libva for compilation and libva with glx needs mesa)
> 
> libva-glx and libva have the same source.
> 
> Do you think there is a better way to do this?

the breaking the bootstrap cycle like this is acceptable.
Note that `update:` section should use `github:` update tracker ideally (the lint check will point this out as well).

mesaglio:
can you help me with this too? @ajayk is another dependecy of opencv

ajayk (changes requested):
Libva needs an epoch bump

ajayk (dismissed):
Looks good to me 
@xnox can you please take a look

lpcalisi:
@xnox shall we merge it?

lpcalisi:
@ajax could you approve workflows? i updated with the latest main changes

lpcalisi:
@ajayk could you re-run actions? scan cve was canceled

mepi262:
@lpcalisi 
It seems that this pull request has conflicts.
Would you resolve it?

PR #18570, opened 2024-05-05
use git-checkout libucontext

mesaglio:


mesaglio:
@hectorj2f

libucontext.yaml
```diff
@@ -44,10 +45,15 @@ subpackages:
...
+      - runs: |
+          make docs install_docs DESTDIR="${{targets.contextdir}}"
       - uses: split/manpages
     description: libucontext manpages
 
 update:
   enabled: true
-  release-monitor:
-    identifier: 17085
+  github:
```
hectorj2f, line 55: Could we also add a test here?

hectorj2f (commented):
LGTM

libucontext.yaml
mesaglio: a test to validate if `libucontext.so` is in the apk?

hectorj2f (approved):

ajayk (approved):

PR #18749, opened 2024-05-07
Add rabbitmq-server-compat package

jamonation:
#### For new package PRs only

#### For new version streams

- [X] Any subpackages include the version string in their package name (e.g. `name: ${{package.name}}-compat`)
- [X] The package (and subpackages) `provides:` logical unversioned forms of the package (e.g. `nodejs`, `nodejs-lts`)


rawlingsj (changes requested):
Hey @jamonation - could this be a subpackage of https://github.com/wolfi-dev/os/blob/main/rabbitmq-server.yaml?

jamonation:
> Hey @jamonation - could this be a subpackage of https://github.com/wolfi-dev/os/blob/main/rabbitmq-server.yaml?

Ideally yeah, this would be a subpackage @rawlingsj. I've been trying to iterate on this quickly so went with this approach to get something working. Otherwise as a subpackage it means rebuilding rabbitmq for each change, which takes too long for small changes like symlinks etc.

I think this can get turned into a subpackage once I have a working image built on it.

rawlingsj:
AH I see, I guess the issue there is we then need to remember to withdraw packages.  If we can avoid adding packages that we know will be removed then it saves on work later which may be forgotten.  

Is this for local development?  One option is to iterate locally, once you have the main package built, you can comment that bit out in the melange yaml and just iterate on the subpackage symlinks until you're happy + tested.  Then PR that?

jamonation:
Somewhat locally @rawlingsj - this is for an image that's got a really tight deadline so I've been head down trying to get it building, and having code in a place where I can hand off to someone else if needed. The other issue is it needs the binaries from the main rabbitmq package copied into the various /opt/bitnami locations, and I can't figure out an elegant way to do that (symlinks don't seem to work)

rawlingsj:
Damn ok - what issues are you running into with the symlinks?  Maybe we can look at that?

imjasonh (dismissed):

rabbitmq-server.yaml
```diff
@@ -78,6 +78,51 @@ subpackages:
...
+    pipeline:
+      - uses: bitnami/compat
+        with:
+          image: rabbitmq
+          version-path: 3.13/debian-12
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/opt/bitnami/rabbitmq
+          ln -sf /usr/lib/rabbitmq/lib/rabbitmq_server-3.13.2/plugins ${{targets.contextdir}}/opt/bitnami/rabbitmq/
+          ln -sf /usr/lib/rabbitmq/lib/rabbitmq_server-3.13.2/escript ${{targets.contextdir}}/opt/bitnami/rabbitmq/
+          ln -sf /usr/lib/rabbitmq/lib/rabbitmq_server-3.13.2/sbin ${{targets.contextdir}}/opt/bitnami/rabbitmq/
```
rawlingsj, line 92:
```suggestion
          ln -sf /usr/lib/rabbitmq/lib/rabbitmq_server-${{package.version}}/plugins ${{targets.contextdir}}/opt/bitnami/rabbitmq/
          ln -sf /usr/lib/rabbitmq/lib/rabbitmq_server-${{package.version}}/escript ${{targets.contextdir}}/opt/bitnami/rabbitmq/
          ln -sf /usr/lib/rabbitmq/lib/rabbitmq_server-${{package.version}}/sbin ${{targets.contextdir}}/opt/bitnami/rabbitmq/
```

rawlingsj (changes requested):
YAY! This is awesome @jamonation!  Just the one nit in the code suggestion above which will help when there's a new automated package update - other than that LGTM :)

rabbitmq-server.yaml
jamonation: You're my hero @rawlingsj! Thanks for spotting that, been staring at these for so long

rawlingsj (approved):

PR #18801, opened 2024-05-08
openrc: new package

jonjohnsonjr:
The subpackages for openrc mirror the alpine subpackages, but I'm open to collapsing them to a single package if we'd prefer that.

In order to test this, I've added an nginx-mainline-openrc package, which needed an init.d script, which I've shamelessly yanked from alpine.


xnox:
i cry as an upstart & sysv-init & systemd committer.

do we really need to?!

imjasonh (approved):

PR #18821, opened 2024-05-08
use intermediate environment variable

cpanato:
similar to https://github.com/chainguard-dev/extra-packages/pull/248

related to https://github.com/chainguard-dev/internal/issues/3686


cpanato:
cc @tstromberg

cpanato:
looks like it is not needed

EyeCantCU (approved):

PR #18845, opened 2024-05-08
readline: fix missing readline update patches

ajayk:


readline.yaml
```diff
@@ -30,6 +30,56 @@ pipeline:
       patches: fix-ncurses-underlinking.patch
       strip-components: 0
 
+  - uses: patch
+    with:
+      patches: readline82-001.patch
+      strip-components: 0
```
pdeslaur, line 36: Can we just use a single `patch` with multiple patches?

pdeslaur (approved):

PR #18915, opened 2024-05-09
postgresql-16.yaml: convert from fetch to git-checkout

xnox:


xnox:
> Wolfictl Check Updates / Wolfictl Check Updates (pull_request) 

will be fixed by the upgrade bot. the tag is there but release tarballs are not out yet.

rawlingsj (approved):

PR #18917, opened 2024-05-09
git bootstrap

xnox:
- **git-bootstrap: add git with vendored curl**
    Ideally zlib-dev openssl-dev curl-dev and maybe even git-bootstrap
    should be in bootstrap-stage2/bootstrap-stage3. Until such time, add
    git-bootstrap in wolfi-dev/os which statically links against
    self-built dependencies. This enables using git-checkout to build any
    package in wolfi, including git itself. As git-bootstrap provides git
    that is installable and buildable using bootstrap-stage3 alone.
  

- **lz4.yaml: convert from fetch to git-checkout**
  

- **lz4.yaml: convert relese-monitor to github**

Previously submissions of lz4.yaml would error out, enable to resolve build cycle. See https://github.com/wolfi-dev/os/pull/18260

### DIFF
--- a/git-bootstrap.yaml
+++ b/git-bootstrap.yaml
@@ -1,0 +1,162 @@
+# Ideally zlib-dev openssl-dev curl-dev and maybe even git-bootstrap
+# should be in bootstrap-stage2/bootstrap-stage3. Until such time, add
+# git-bootstrap in wolfi-dev/os which statically links against
+# self-built dependencies. This enables using git-checkout to build
+# any package in wolfi, including git itself. As git-bootstrap
+# provides git that is installable and buildable using
+# bootstrap-stage3 alone.
+package:
+  name: git-bootstrap
+  version: 2.45.2
+  epoch: 0
+  description: "distributed version control system"
+  copyright:
+    - license: GPL-2.0-or-later
+  dependencies:
+    provider-priority: 5
+    provides:
+      - git
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - perl
+
+vars:
+  zlib-version: 1.3.1
+  openssl-version: 3.3.0
+  curl-version: 8.7.1
+
+pipeline:
+  - working-directory: zlib
+    pipeline:
+      - uses: fetch
+        with:
+          uri: https://zlib.net/zlib-${{vars.zlib-version}}.tar.gz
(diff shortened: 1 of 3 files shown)


xnox:
> unresolvable cycle: zlib:1.3.1-r1@local -> libtool:2.4.7-r3@local, caused by: libtool:2.4.7-r3@local -> build-base:1-r7@local -> glibc-dev:2.39-r3@local -> glibc:2.39-r3@local -> zlib:1.3.1-r1@local

this is because "zlib-dev" is needed, whereas bootstrap only has "zlib" sigh.

and also likely bootstrap needs to add libtool, because otherwise it is pain.

xnox:
@smoser this enables to build key packages with git, including git itself.

smoser:
> @smoser this enables to build key packages with git, including git itself.

but how are you planning on using it?  I kind of feel like why would I not just use it for all the things (updating the pipeline to use git-bootstrap).

xnox:
> > @smoser this enables to build key packages with git, including git itself.
> 
> but how are you planning on using it? I kind of feel like why would I not just use it for all the things (updating the pipeline to use git-bootstrap).

because it is slower and doesn't support http2 and a few other things. and by default it provides git; but a lower priority. Thus it will be used if git doesn't exist; but stops being used once full git is built.

git-bootstrap.yaml
```diff
@@ -0,0 +1,159 @@
...
+        git clone --depth 1 https://github.com/wolfi-dev/os.git test_clone
+        [ -d "test_clone/.git" ] || exit 1
+    - name: Fetch repository updates
+      runs: |
+        cd test_clone
+        git fetch || exit 1
+    - name: Check git configuration
+      runs: |
+        git config --list | grep 'user.name=Wolfi' || exit 1config --list | grep 'user.name=Test User' || exit 1
+        git config --list | grep 'user.email=wolf@wolfi.dev' || exit 1ist | grep 'user.email=test@example.com' || exit 1
```
smoser, line 159: `exit 1ist` and `exit 1config --list` ?

I see that you copied that, but it looks just like nonsense to me.

And , i see that @imjasonh fixed that test yesterday e1de9032301294948e48cd1897fe8b23778d83a4

So, please update to grab that change.

smoser (changes requested):
one change but other than that looks great. thank you.

smoser (approved):

PR #19000, opened 2024-05-10
Add explicit lib dependencies for openjdk-jre

pnasrat:
Adds required X11, sound, graphics, and a truetype font required by the JRE.

melange SCA so deps not found as requiring libs 644

Fixes: Issues discovered during Java compliance testing


pnasrat:
cc @vaikas

kaniini:
Please consider adding -jre-full variants instead of doing this as most Java applications do not actually require these features and they will just bloat the images.

pnasrat:
@kaniini that's definitely a future change I want to look at in terms of how the JRE is packaged to correctly , . I am aware that note Linux distros that provide a headless jre such as [debian](https://packages.debian.org/bookworm/java/openjdk-17-jre-headless)  

Currently the jre image is broken as it specifically includes the libs (eg `libawt_xawt.so`) https://github.com/wolfi-dev/os/blob/50c7526c99618ffdedc6aec6c297fd91de2d856a/openjdk-21.yaml#L133 

Before restructuring the package and subsequent images I would  want to ensure it is done in a way that doesn't break things for users but I'll file a separate issue for that and I've converted this to draft for now.

openjdk-21.yaml
```diff
@@ -125,6 +125,15 @@ subpackages:
         - freetype
         - libfontconfig1
         - openjdk-21-jre-base
+        - alsa-lib
```
vaikas, line 128: super nit, could you alphabetize these?

vaikas (commented):
We should probably open an issue for tracking the appropriate packages that we want to have (headless, etc.).

openjdk-21.yaml
pnasrat: @vaikas done via a forced push

vaikas (approved):

PR #19252, opened 2024-05-14
Add firefox-esr package

mamccorm:
Adding firefox-esr package. epoch set to 1 is an intentional change here.

### DIFF
--- a/firefox-esr.yaml
+++ b/firefox-esr.yaml
@@ -45,7 +45,7 @@ environment:
       - llvm-lld-17
       - llvm17
       - m4
-      - mercurial=6.4.5-r0
+      - mercurial
       - nasm
       - nodejs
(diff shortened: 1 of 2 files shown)


firefox-esr.yaml
```diff
@@ -0,0 +1,157 @@
...
+      - libxfixes-dev
+      - libxft-dev
+      - libxi-dev
+      - libxrandr-dev
+      - libxrender-dev
+      - libxtst-dev
+      - llvm-lld-17
+      - llvm17
+      - m4
+      - mercurial=6.4.5-r0
```
jamonation, line 48:
```suggestion
      - mercurial
```

jamonation: Let's see if this works to get the build going

PR #19253, opened 2024-05-14
Be safer with here docs in docker-run and lint-world actions.

smoser:
This allows the user of docker-run action to not have to worry that their document will be parsed by 2 shells.

from https://manpages.ubuntu.com/manpages/noble/en/man1/bash.1.html

> If any part of word is quoted, the delimiter  is the  result of quote
> removal on word, and the lines in the here-document are not expanded.
> If word is unquoted, all lines of the here-document are subjected to
> parameter  expansion, command  substitution,  and  arithmetic
> expansion,  the  character sequence \<newline> is ignored, and \ must be
> used to quote the characters \, $, and `.

What that sums up to is that there is less interpretation on the here document when using `<<"EOF"` than when using `<<EOF`.

### DIFF
--- a/.github/actions/docker-run/action.yaml
+++ b/.github/actions/docker-run/action.yaml
@@ -39,7 +39,7 @@ runs:
           -e GOOGLE_GHA_CREDS_PATH=${GOOGLE_GHA_CREDS_PATH} \
           -i \
           ${{ inputs.opts }} \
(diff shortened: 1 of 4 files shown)


joshrwolf:
TIL! Thank you!

joshrwolf (approved):

PR #19254, opened 2024-05-14
fix: valkey provides redis which seems to be incorrect as redis packa…

ajayk:


valkey.yaml
```diff
@@ -1,13 +1,13 @@
...
   version: 7.2.5
-  epoch: 0
+  epoch: 1
   description: Advanced key-value store
   copyright:
     - license: BSD-3-Clause
   dependencies:
     provides:
-      - redis=${{package.full-version}}
+      - valkey=${{package.full-version}}
```
imjasonh, line 10: We don't need this line at all, since it already provides valkey=$version

imjasonh (approved):

PR #19266, opened 2024-05-14
adds some needed scripts for keycloak-operator

mritunjaysharma394:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/keycloak.yaml
+++ b/keycloak.yaml
@@ -1,13 +1,14 @@
 package:
   name: keycloak
   version: 24.0.4
-  epoch: 1
+  epoch: 2
   description: Open Source Identity and Access Management For Modern Applications and Services
   copyright:
     - license: Apache-2.0
   dependencies:
     runtime:
       - bash # Keycloak helper scripts require bash, aren't compatible with busybox.
+      - keycloak-scripts # Separate package for these scripts as they might be reused by other packages like keycloak-operator
       - openjdk-17-default-jvm
 
 environment:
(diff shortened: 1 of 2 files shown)


keycloak.yaml
```diff
@@ -62,6 +58,13 @@ subpackages:
       - runs: |
           mkdir -p "${{targets.subpkgdir}}"/opt
           ln -s /usr/share/java/keycloak "${{targets.subpkgdir}}"/opt/
+  - name: keycloak-scripts
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/usr/bin
+          for i in kc.sh kcadm.sh kcreg.sh; do
+            ln -sf /usr/share/java/keycloak/bin/$i ${{targets.subpkgdir}}/usr/bin/$i
```
jonjohnsonjr, line 66: I think we want `keycloak-scripts` to contain the actual files, not just symlinks, right?

mritunjaysharma394: yes, but I think since in operator manifest we are going to use our keycloak image as well so probably that can handle that?:

```yaml
    spec:
      containers:
        - env:
            - name: RELATED_IMAGE_KEYCLOAK
              value: quay.io/keycloak/keycloak:24.0.3
          image: quay.io/keycloak/keycloak-operator:24.0.3
  ```

mritunjaysharma394: the other option is to add scripts it in `keycloak-operator` itself probably and symlink or cp it there itself without calling `keycloak-scripts`?

jonjohnsonjr: > yes, but I think since in operator manifest we are going to use our keycloak image as well so probably that can handle that?

I would avoid touching `keycloak` at all and just copy these scripts into `keycloak-scripts` so that `keycloak-operator` can depend on them.

mritunjaysharma394:
Closing this one as we figured out and think `keycloak-compat` in apko files of keycloak image is enough to fix it

PR #19272, opened 2024-05-14
Add LDFLAGS to a list of Go packages

hectorj2f:
We converted some go/install pipelines to git-checkout + go/bump and forgot to set these flags to get the version info.


ajayk (approved):

xnox (commented):
`-w` are redundant and already added by the default pipeline.

i was too slow to review this.

PR #19455, opened 2024-05-14
rust: add rustmfmt and clippy

lpcalisi:
* add rustmfmt and clippy tools for lintering (might be good to use in pipelines)

### DIFF
--- a/rust-1.78.yaml
+++ b/rust-1.78.yaml
@@ -1,7 +1,7 @@
 package:
   name: rust-1.78
   version: 1.78.0
-  epoch: 0
+  epoch: 1
   description: "Empowering everyone to build reliable and efficient software."
   copyright:
     - license: Apache-2.0 AND MIT
@@ -64,7 +64,7 @@ pipeline:
         --llvm-config="/usr/lib/llvm17/bin/llvm-config" \
         --disable-docs \
         --enable-extended \
-        --tools="cargo,src" \
+        --tools="cargo,src,clippy,rustfmt" \
         --enable-llvm-link-shared \
         --enable-option-checking \
         --enable-locked-deps \


lpcalisi:
@ajayk  @xnox @smoser what do you think about it? should we do it for all rust versions too?

this make each package more atomic but after this, installs rust will  not have the entire content, how we can be handle it? Maybe we can add all these packages to a rust runtime dependencies until 1.79 versions will release

rust-1.78.yaml
```diff
@@ -114,6 +114,83 @@ pipeline:
...
+      runtime:
+        - rust=${{package.full-version}}
+        - gdb
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}/usr/bin"
+          mkdir -p "${{targets.subpkgdir}}/usr/share/rust/etc"
+          mv "${{targets.destdir}}/usr/bin/rust-gdb" "${{targets.subpkgdir}}/usr/bin"
+          mv "${{targets.destdir}}/usr/bin/rust-gdbgui" "${{targets.subpkgdir}}/usr/bin"
+          mv "${{targets.destdir}}/usr/lib/rustlib/etc/gdb_*.py" "${{targets.subpkgdir}}/usr/share/rust/etc"
```
smoser, line 179: /usr/lib/rustlib/etc -> /usr/share/rust/etc ?

xnox:
Adding tools is good.

It seems to me that splitting packages adds no value here? rust is a development tool only, only used as build-dependency, without any runtime dependencies (i.e. there is no librust.so similar to libgcc_s1.so) and one typically needs all of that installed always. And developers that need fmt or cargo, without rust itself is a very small target audience.

See how many packages declare "rust" in the environment and then call cargo.

lpcalisi:
> Adding tools is good.
> 
> It seems to me that splitting packages adds no value here? rust is a development tool only, only used as build-dependency, without any runtime dependencies (i.e. there is no librust.so similar to libgcc_s1.so) and one typically needs all of that installed always. And developers that need fmt or cargo, without rust itself is a very small target audience.
> 
> See how many packages declare "rust" in the environment and then call cargo.

@xnox i was thinking from the perspective of a build system or pipelines, when probably we dont need all the tools on it and we would are increasing size of images with not used tools, but.. i finished to compile all subpackages and there is not big saving making subpackages

```
-rw-r--r--    1 root     root        6.7K May 15 13:10 APKINDEX.json
-rw-r--r--    1 root     root        1.5K May 15 13:10 APKINDEX.tar.gz
-rw-r--r--    1 root     root        9.6M May 15 13:10 cargo-1.78.0-r0.apk
-rw-r--r--    1 root     root       57.9M May 15 13:10 rust-1.78-1.78.0-r0.apk
-rw-r--r--    1 root     root        3.7M May 15 13:10 rust-clippy-1.78.0-r0.apk
-rw-r--r--    1 root     root       94.6K May 15 13:10 rust-doc-1.78.0-r0.apk
-rw-r--r--    1 root     root        8.1K May 15 13:10 rust-gdb-1.78.0-r0.apk
-rw-r--r--    1 root     root        8.1K May 15 13:10 rust-lldb-1.78.0-r0.apk
-rw-r--r--    1 root     root        4.2M May 15 13:10 rust-src-1.78.0-r0.apk
-rw-r--r--    1 root     root        2.6M May 15 13:10 rustfmt-1.78.0-r0.apk
```

xnox (approved):

lpcalisi:
@ajayk shall we merge it? thanks

ajayk (approved):

PR #19477, opened 2024-05-15
chore: remove redundant ldflags

hectorj2f:
The go package comes with a default `ldflag: -w` so we can safely remove it from the list of set ldflags in the package definitions.

### DIFF
--- a/conftest.yaml
+++ b/conftest.yaml
@@ -1,7 +1,7 @@
 package:
   name: conftest
   version: 0.52.0
-  epoch: 1
+  epoch: 2
   description: Write tests against structured configuration data using the Open Policy Agent Rego query language
   copyright:
     - license: Apache-2.0
@@ -18,7 +18,7 @@ pipeline:
 
   - uses: go/build
     with:
-      ldflags: -w -s -X github.com/open-policy-agent/conftest/internal/commands.version=${{package.version}}
+      ldflags: -X github.com/open-policy-agent/conftest/internal/commands.version=${{package.version}}
       modroot: .
       output: conftest
       packages: .
--- a/crossplane.yaml
+++ b/crossplane.yaml
@@ -1,7 +1,7 @@
 package:
   name: crossplane
   version: 1.15.2
-  epoch: 1
+  epoch: 2
   description: Cloud Native Control Planes
   copyright:
     - license: Apache-2.0
@@ -30,7 +30,6 @@ pipeline:
 
   - uses: go/build
     with:
-      ldflags: -w
       output: crossplane
       packages: ./cmd/crossplane
 
@@ -44,7 +43,6 @@ subpackages:
           deps: github.com/go-git/go-git/v5@v5.11.0
       - uses: go/build
         with:
-          ldflags: -w
           output: crank
           packages: ./cmd/crank
           subpackage: "true"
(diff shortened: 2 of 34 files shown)


conftest.yaml
```diff
@@ -18,7 +18,7 @@ pipeline:
 
   - uses: go/build
     with:
-      ldflags: -w -s -X github.com/open-policy-agent/conftest/internal/commands.version=${{package.version}}
+      ldflags: -s -X github.com/open-policy-agent/conftest/internal/commands.version=${{package.version}}
```
xnox, line 21: hmmmm i think we agree to not strip symbols, that's why we used to have '-s' before in default flags and then dropped. please drop '-s' too as we want symbols tables for govulnchecker to work.

crossplane.yaml
```diff
@@ -30,7 +30,7 @@ pipeline:
 
   - uses: go/build
     with:
-      ldflags: -w
+      ldflags:
       output: crossplane
```
xnox, line 34: empty ldflags key should be removed.

crossplane.yaml
```diff
@@ -44,7 +44,7 @@ subpackages:
           deps: github.com/go-git/go-git/v5@v5.11.0
       - uses: go/build
         with:
-          ldflags: -w
+          ldflags:
           output: crank
```
xnox, line 48: empty ldflags key should be removed.

cue.yaml
```diff
@@ -34,7 +34,7 @@ pipeline:
       modroot: .
       packages: ./cmd/cue
       output: cue
-      ldflags: -w -buildid= -X cuelang.org/go/cmd/cue/cmd.version=${{package.version}}
+      ldflags: -buildid= -X cuelang.org/go/cmd/cue/cmd.version=${{package.version}}
```
xnox, line 37: i thought passing in empty buildid= destroys information no? i would rather not pass one in, or have something meaningful like sha1 commit. but i can't remember the expected buildid format.

gitlab-logger.yaml
```diff
@@ -21,7 +21,7 @@ pipeline:
     with:
       packages: ./cmd/gitlab-logger
       output: gitlab-logger
-      ldflags: -w
+      ldflags:
```
xnox, line 24: empty ldflags key should be dropped

go-bindata.yaml
```diff
@@ -24,7 +24,7 @@ pipeline:
     with:
       packages: ./go-bindata
       output: go-bindata
-      ldflags: -w
+      ldflags:
```
xnox, line 27: drop.

crossplane.yaml
hectorj2f: oh! This wasn't intended.

trillian.yaml
```diff
@@ -35,7 +35,7 @@ subpackages:
           subpackage: "true"
           packages: ./cmd/trillian_log_server
           output: trillian_log_server
-          ldflags: -w
+          ldflags:
```
xnox, line 38: drop

trillian.yaml
```diff
@@ -46,7 +46,7 @@ subpackages:
           subpackage: "true"
           packages: ./cmd/trillian_log_signer
           output: trillian_log_signer
-          ldflags: -w
+          ldflags:
```
xnox, line 49: drop

xnox (changes requested):
please tweak commented things.

xnox (approved):

PR #19499, opened 2024-05-15
lint: fix all remaining copyright tags

xnox:
With upgraded wolfictl that allows custom license tag, and updated
spdx license list, all packages pass:

    $ wolfictl lint
    found 2387 packages

Fixing failure to build:

* **ncurses:** bump epoch to 3, as 2 was previously builtncurses: bump epoch to 3, as 2 was previously built
* **ncurses:** ensure dev libraries link with tinfo too


ajayk (dismissed):

imjasonh (dismissed):

openipmi.yaml
```diff
@@ -1,10 +1,10 @@
 package:
   name: openipmi
   version: 2.0.34
-  epoch: 0
+  epoch: 1
```
imjasonh, line 4: It seems this package doesn't build anymore?

```
2024/05/15 16:21:41 WARN configure: error: libtermcap, libcurses or libncurses are required!
```

xnox: @imjasonh ncurses build failure is resolved by making sure dev libraries of curses link with tinfo.

xnox:
@imjasonh 

> Unrecognized license reference: custom. license_expression must only use IDs from the license list or extracted licensing info, but is: custom

What do we do here? given i'm correcting invalid tags for tag "custom" how can this be merged? Also it seems like we are enforcing only valid licenses via the sbom check, meaning wolfictl lint is redundant - or rather is just an early signal.

Do we really need to implement "find custom license and generate spdx reference for it" right now?

imjasonh:
> ...meaning wolfictl lint is redundant - or rather is just an early signal.
> 
> Do we really need to implement "find custom license and generate spdx reference for it" right now?

Yeah, we wanted it as an early signal before finding out at image-build time that the license identifier chosen in the package was invalid. I still think it's valuable for that.

If `custom` is a valid license identifier for our use case, I think it makes sense to make `wolfictl lint` accept it. This means we'll have to disable SBOM validity checks on any image that package ends up in, unless we also relax that test (which I'd prefer not to do, since it involves external tools).

xnox:
> > ...meaning wolfictl lint is redundant - or rather is just an early signal.
> > Do we really need to implement "find custom license and generate spdx reference for it" right now?
> 
> Yeah, we wanted it as an early signal before finding out at image-build time that the license identifier chosen in the package was invalid. I still think it's valuable for that.
> 
> If `custom` is a valid license identifier for our use case, I think it makes sense to make `wolfictl lint` accept it. This means we'll have to disable SBOM validity checks on any image that package ends up in, unless we also relax that test (which I'd prefer not to do, since it involves external tools).

Can we force merge this PR as it doesn't introduce any new bad license tags; and fixes many of them.

SPDX allows to include custom license text - designing how to achieve it over here https://github.com/chainguard-dev/melange/issues/1212

xnox:
@imjasonh updating this PR to not touch any packages which are custom. However, when "check SBOMs" check was added, it didn't do anything to existing invalid licenses. Meaning, we no longer can build any of those packages. Hopefully those packages will not change any time soon.

imjasonh (approved):

imjasonh:
> @imjasonh updating this PR to not touch any packages which are custom. However, when "check SBOMs" check was added, it didn't do anything to existing invalid licenses. Meaning, we no longer can build any of those packages. Hopefully those packages will not change any time soon.

If we need to force merge changes to those packages to get past `check SBOMs` we can, it just means it'll hit us later when those packages get built into images.

PR #19515, opened 2024-05-15
add datadog-integrations-core

joshrwolf:
the integrations components needed at runtime by `datadog-agent`

### DIFF
--- a/datadog-integrations-core.yaml
+++ b/datadog-integrations-core.yaml
@@ -1,0 +1,40 @@
+package:
+  name: datadog-integrations-core
+  version: 7.53.0
+  epoch: 0
+  description: "Core integrations of the Datadog Agent"
+  copyright:
+    - license: BSD-3-Clause
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - krb5-dev
+      - linux-headers
+      - py3-pip
+      - python3-dev
+  environment:
+    # This is needed to work around the error "ValueError: ZIP does not support timestamps before 1980"
+    SOURCE_DATE_EPOCH: 315532800
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/DataDog/integrations-core
+      tag: ${{package.version}}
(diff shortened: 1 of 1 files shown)


datadog-integrations-core.yaml
```diff
@@ -0,0 +1,39 @@
...
+  - runs: |
+      python3 -m pip install "./datadog_checks_base[deps, http]" --prefix=/usr --root=${{targets.destdir}}
+
+  - uses: strip
+
+update:
+  enabled: true
+  github:
+    identifier: DataDog/integrations-core
+    use-tag: true
```
stormqueen1990, line 38: https://github.com/datadog/integrations-core uses releases. Should we omit `use-tag` so it pulls from the release instead?
```suggestion
```

joshrwolf: from what I can tell, the `ddev` releases are unrelated, what we actually want are the `x.y.z` tags, which map to the upstream agent version

smoser (changes requested):
```
$ tar tvf packages/x86_64/datadog-integrations-core-7.53.0-r0.apk
drwxrwxr-x root/root         0 1969-12-31 19:00 usr/bin/__pycache__
-rw-rw-r-- root/root      3204 1969-12-31 19:00 usr/bin/__pycache__/jp.cpython-312.pyc
-rwxrwxr-x root/root       227 1969-12-31 19:00 usr/bin/ddtrace-run
-rwxrwxr-x root/root      1695 1969-12-31 19:00 usr/bin/jp.py
-rwxrwxr-x root/root       213 1969-12-31 19:00 usr/bin/jsonschema
-rwxrwxr-x root/root       233 1969-12-31 19:00 usr/bin/normalizer
-rwxrwxr-x root/root       214 1969-12-31 19:00 usr/bin/pyspnego-parse
```

There are differing levels of yuck there. The biggest is __pycache__ dir in bin/.

Where we use the `pip install` approach, we really need to also use a virtual env. 
You can see az.yaml for an example of what we're doing there.

joshrwolf:
closing this since I now realize this needs to be part of the main `datadog-agent` package (for datadog reasons)

PR #19608, opened 2024-05-16
include system-probe in the build

joshrwolf:
this turned into a large PR, in general I assumed the existing package is wrong/broken, so I took quite a bit of liberty in restructuring things. the highlights of this change are:

- bundles the `system-probe` component in the multicall `agent`
  - this uses kernel headers from debian stable `5.10.216-1~deb10u1` (to match what upstream is doing)
- fixes the CGO impl by using the correct include paths 
- fixes the `rtloader` integration by using the correct `--embedded-path`
  - fixes the `RPATH` used by the `agent` to call the appropriate `rtloader` headers
- creates a new `datadog-agent-core-integrations` subpackage that includes all the python plugin check integrations
  - installed using the py3.11 venv approach
  - pulls in a [patch](https://github.com/DataDog/datadog-agent/pull/25348) from upstream to simplify the discovery of checks
- better matches the directory structure to upstream, including a few key hardcoded paths
  - place multicall symlinks to `/opt/datadog-agent/embedded` where the entrypoints are expecting
- embed the `jmx` integration in the multicall, but separate the runtime dependency on `openjdk` as a separate package, included at image build time
- downgrade go (1.21) and python (3.11) to match upstreams dependency locks

there's a lot of changes at once here, for some additional confidence, this has been tested (in its image form) against the standard [docker](https://docs.datadoghq.com/containers/docker/?tab=standard) and [fargate](https://docs.datadoghq.com/integrations/ecs_fargate/?tab=webui) setup

### DIFF
--- a/datadog-agent.yaml
+++ b/datadog-agent.yaml
@@ -1,16 +1,17 @@
 package:
   name: datadog-agent
   version: 7.53.0
-  epoch: 4
+  epoch: 7
   description: "Collect events and metrics from your hosts that send data to Datadog."
   copyright:
     - license: Apache-2.0
   dependencies:
     runtime:
+      # - datadog-integrations-corentegrations-core
       - findutils
       - grep
       - libseccomp
-      - openjdk-11
+      # - openjdk-11
       - python3
       - shadow
 
@@ -20,11 +21,19 @@ environment:
       - build-base
       - busybox
       - ca-certificates-bundle
+      - clang-16
+      - clang-16-dev
       - cmake
       - coreutils
       - curl
+      # https://github.com/DataDog/datadog-agent/blob/95f4d66f5fb0c10940e0f9b8fc2ab33f0c6099f0/omnibus/config/software/datadog-agent-integrations-py3-dependencies.rb
+      - freetds-dev
+      # - gstatus # TODO: Required by gluster
       - gcc-12
-      - go
+      - go~1.21
+      - llvm16
+      - ninja
+      - procps-dev
       - py3-boto3
       - py3-codeowners
       - py3-docker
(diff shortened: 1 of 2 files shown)


datadog-agent.yaml
```diff
@@ -58,115 +71,246 @@ pipeline:
...
     with:
       deps: github.com/moby/buildkit@v0.13.1 golang.org/x/net@v0.23.0
       replaces: github.com/mholt/archiver/v3=github.com/anchore/archiver/v3@v3.5.2
 
-  # This step essentially replicates https://github.com/DataDog/datadog-agent/blob/main/omnibus/config/software/datadog-agent.rb
+  - if: ${{build.arch}} == 'aarch64'
+    runs: |
+      mkdir -p /usr/src
+      wget -O arch.deb http://deb.debian.org/debian-security/pool/updates/main/l/linux-5.10/linux-headers-${LINUX_HEADERS_VERSION}-arm64_${LINUX_KERNEL_VERSION}_arm64.deb
+      echo "a9f5b8614dc1dc123b13de540bb98283cd927a29438e0fe010c9e477b5448124 arch.deb" | sha256sum -c
```
xnox, line 96: if you want you can use:

```
      - uses: fetch
        with:
          uri: URI
          expected-sha256: SHA256
          extract: false
```

This should simplify downloads a little bit. Note the `extract: false` key to basically fetch blobs which are not .tar.gz tarballs.

datadog-agent.yaml
```diff
@@ -58,115 +71,246 @@ pipeline:
...
+      wget -O common.deb http://deb.debian.org/debian-security/pool/updates/main/l/linux-5.10/linux-headers-${LINUX_HEADERS_VERSION}-common_${LINUX_KERNEL_VERSION}_all.deb
+      dpkg -x common.deb /tmp/common
+      ln -s /tmp/common/usr/src/linux-headers-${LINUX_HEADERS_VERSION}-common /usr/src/linux-headers-${LINUX_HEADERS_VERSION}-common
+
+      mkdir -p /opt/clang
+      tar xf /tmp/clang.tar.xz --no-same-owner -C /opt/clang --strip-components=1
+      PATH="/opt/clang/bin:${PATH}"
+
+      # HOW DOES THIS WORK: https://stackoverflow.com/a/60357720
+      ln -s /usr/lib/libncursesw.so.6 /usr/lib/libtinfo.so.5
```
xnox, line 127: hahahhahah magic. normally we link against ncurses which is actually a libtool script of two things new style wide curses and libtinfo. Note this was fixed up recently in our headers, but you should be able to say and link against libtinfo everywhere (aka extra libs `-ltinfo`). But this is ok as well.

xnox (approved):
Looks good, minor comments to simplify things potentially.

datadog-agent.yaml
```diff
@@ -270,10 +414,15 @@ test:
     contents:
       packages:
         - datadog-agent-fakeintake
+        - datadog-agent-core-integrations
+    environment:
+      PYTHONPATH: "/usr/share/datadog-agent/lib/python3.11/site-packages"
```
smoser, line 419: did the previous use of PATH not work? i think i kind of prefer it to this.

i'm not fussed about it though.

we have lots of this stuff to "best practicify"

smoser (commented):
lgtm.

datadog-agent.yaml
xnox: https://github.com/wolfi-dev/os/commit/9a41b88dcb707beb09bcdf282631c3eb410c657a#diff-46e9271bf720bb36250e2cb96afc5004fcad2abd1dda375215cab63918e49a7d if really interested

datadog-agent.yaml
joshrwolf: the current package didn't build the core-integrations correctly so it was never loading to begin with.

I kept things namespaced in `/usr/share` because directly installing was conflicting with some system build libraries. I imagine I was just doing things wrong, but iiuc the end result is mostly the same. if that's correct I'll just leave it as is for now until the best practices are more solidified 🙏

joshrwolf:
going to land this as is to kick off a HEAD build, will address the nits in a follow up

PR #19681, opened 2024-05-17
spdx-tools-java: test all our own SBOMs

imjasonh:
ntia-conformance-checker too

### DIFF
--- a/py3-ntia-conformance-checker.yaml
+++ b/py3-ntia-conformance-checker.yaml
@@ -2,7 +2,7 @@
 package:
   name: py3-ntia-conformance-checker
   version: 2.0.0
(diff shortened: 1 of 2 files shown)


py3-ntia-conformance-checker.yaml
```diff
@@ -2,7 +2,7 @@
 package:
   name: py3-ntia-conformance-checker
   version: 2.0.0
-  epoch: 0
+  epoch: 1
```
imjasonh, line 5: Bumping this and not the other one because the current SBOM isn't actually valid until it's rebuilt.

luhring (approved):

Issue #19701, opened 2024-05-17
[Wolfi Package Request]: Flexible I/O Tester (fio)


lyoung-confluent:
### What software would you like us to add to wolfi-os. Ideally include a URL to the project and its source.

https://github.com/axboe/fio

### which versions of the software should we include?

Latest (3.37)

### Add some justification for why this specific package and versions are important.

This is a useful tool for testing I/O performance and is included in various other package repositories already (ubuntu, alpine, redhat, etc) 

### Please check all that apply

- [X] This package has an un-restrictive license
- [X] The package/versions proposed are actively maintained upstream
- [X] I am interested in adding this package to Wolfi OS myself
- [ ] I am willing to help maintain this package

krisharyan117:
fio is already added. @debasishbsws  please close this issue

PR #19702, opened 2024-05-17
Add fio package

lyoung-confluent:
Fixes:
* #19701

#### For new package PRs only

### DIFF
--- a/fio.yaml
+++ b/fio.yaml
@@ -1,0 +1,41 @@
+package:
+  name: fio
+  version: "3.37"
+  epoch: 0
+  description: Flexible I/O Tester
+  copyright:
+    - license: GPL-2.0-or-later
+
+environment:
+  contents:
+    packages:
+      - bash
+      - build-base
+      - busybox
+      - libaio-dev
+      - zlib-dev
+      - coreutils
+      - ca-certificates-bundle
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/axboe/fio
+      tag: fio-${{package.version}}
+      expected-commit: 028fcc7db00155fe329497d7501184e8fe49342f
+
+  - runs: |
+      ./configure --prefix=/usr
+      make
+      make DESTDIR="${{targets.destdir}}" install
+
+subpackages:
+  - name: fio-docs
+    pipeline:
+      - uses: split/manpages
+
+update:
+  enabled: true
+  github:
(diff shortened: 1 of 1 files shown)


fio.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - uses: git-checkout
+    with:
+      repository: https://github.com/axboe/fio
+      tag: fio-${{package.version}}
+      expected-commit: 028fcc7db00155fe329497d7501184e8fe49342f
+
+  - runs: |
+      ./configure --prefix=/usr
+      make
+      make DESTDIR="${{targets.destdir}}" install
```
ajayk, line 30:
```suggestion
      make DESTDIR="${{targets.contextdir}}" install
```

fio.yaml
```diff
@@ -0,0 +1,41 @@
...
+subpackages:
+  - name: fio-docs
+    pipeline:
+      - uses: split/manpages
+
+update:
+  enabled: true
+  github:
+    identifier: axboe/fio
+    strip-prefix: fio-
```
ajayk, line 41: Lets add a basic test 
```suggestion
    strip-prefix: fio-
 
test:
    pipeline:
        - name: Verify fio installation
          runs: |
            fio --version || exit 1   
```

fio.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/axboe/fio
+      tag: fio-${{package.version}}
+      expected-commit: 028fcc7db00155fe329497d7501184e8fe49342f
+
+  - runs: |
+      ./configure --prefix=/usr
```
tuananh-vpbank, line 28: could use builtin melange pipelines for consistency?

```yaml
- uses: autoconf/configure
- uses: autoconf/make
- uses: autoconf/make-install
```

lyoung-confluent: I tried that initially, but it's actually not an autoconf configure script, it's a hand-crafted one that doesn't support some of the configuration options used by that pipeline: https://github.com/axboe/fio/blob/3ed8eea0ee9b5d7de603e9b128e6c05941b99938/configure#L3-L4

It seems like I could still use `autoconf/make` and `autoconf/make-install` but I'm not sure if that's a good pattern to follow given it's not actually autoconf...

ajayk (approved):

PR #19788, opened 2024-05-20
Node fix

sschubertchainguard:
Add `usr/local/lib` to address failure which can occur with `npm ls -g` if no packages are installed


ericsmalling:
I'm curious, would it make more sense to make the change in the npm package instead of all of the node ones?

xnox:
Please do not change node packages for this.

if we want these directories to be real and present we should simply implement that in baselayout packages.

PR #19796, opened 2024-05-20
add libreoffice to wolfi

mamccorm:
Adds latest libreoffice to wolfi


libreoffice-24.yaml
```diff
@@ -0,0 +1,171 @@
+package:
+  name: libreoffice-24
+  version: 24.2.3.2
+  epoch: 1
```
mamccorm, line 4: Setting this to 1 is intentional here

imjasonh (approved):

PR #19827, opened 2024-05-21
cilium 1.15.5 upgrade

xnox:
- **Upgrade to a git-checkout**

- **Upgrade proxy repo commit** 

- **Upgrade to llvm-17**

- **Add no-lint tags due to nested checkout**

- **go/bump top level too**

Currently still fails to build from source, will check if our include paths are correct.


xnox:
Should fix https://github.com/wolfi-dev/os/issues/19505

xnox:
hm our llvm packages are weird, maybe i need add clang-17-dev too..... which is odd.

tcnghia (approved):

PR #19839, opened 2024-05-21
Add /usr/local/lib to baselayout

xnox:
- **wolfi-baselayout: make usr/local/lib real**
  Right now we ship usr/local/lib64 as a dangling symlink to
  usr/local/lib. This confuses some software that gets excited to scan
  usr/local/lib64 and then fails. Add usr/local/lib as a real empty
  directory to the baselayout.
  
  This fixes, among many other things, warnings and errors from node in
  all node images.

### DIFF
--- a/wolfi-baselayout.yaml
+++ b/wolfi-baselayout.yaml
@@ -1,7 +1,7 @@
 package:
   name: wolfi-baselayout
   version: 20230201
-  epoch: 9
+  epoch: 10
   description: "baselayout data for Wolfi"
   copyright:
     - license: MIT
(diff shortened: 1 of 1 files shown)


pnasrat (approved):
LGTM and makes sense!

wolfi-baselayout.yaml
```diff
@@ -37,7 +37,7 @@ pipeline:
 
   - name: Install
     runs: |
-      for i in bin etc etc/profile.d etc/secfixes.d home lib root var/log usr/bin usr/sbin usr/local tmp var/spool/cron opt run usr/lib; do
+      for i in bin etc etc/profile.d etc/secfixes.d home lib root var/log usr/bin usr/sbin usr/local/lib tmp var/spool/cron opt run usr/lib; do
```
smoser, line 40: this kind of raises the question if /usr/local/lib should be in /etc/ld.so.conf.  
thoughts?

xnox: it already is....

```console
# cat /etc/ld.so.conf 
/usr/local/lib
/usr/local/lib64
/lib
/lib64
/usr/lib
/usr/lib64
```

xnox: given lib64 are symlinks, it is pointless for them to be there but hey ho =)

smoser: oh yeah, i had forgotten that they were in there. duh. thanks.

PR #19847, opened 2024-05-21
add osixia-openldap

Dentrax:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches


xnox (commented):
all of this feels like super weird stuff, and redundant.

This stuff is all in place only to shoehorn classic linux installs into a container.

daemon UIDs we normally specify in apko template or helm chart.

this all feels like a tonne of unnecessary work re-implementing what apko can do natively.

Dentrax:
> all of this feels like super weird stuff, and redundant.
> 
> This stuff is all in place only to shoehorn classic linux installs into a container.
> 
> daemon UIDs we normally specify in apko template or helm chart.
> 
> this all feels like a tonne of unnecessary work re-implementing what apko can do natively.

Thanks for the feedback. I had to `sed` those UIDs to use `adduser` but it's a no-op. Otherwise their startup script was throwing `no such file` due to non-existent `usermod` and `groupmod`. Normally we will have set those in the Apko for sure.

Any thoughts how should we move forward on this?

Upstream startup file: https://github.com/osixia/docker-openldap/blob/master/image/service/slapd/startup.sh
Run file: (in different repo): https://github.com/osixia/container-baseimage/blob/main/image/tool/run

xnox:
> Any thoughts how should we move forward on this?

what is even more concerning is that all those environment variables leak secrets; and then others wrap that into helm chart matching those variables (as in see https://artifacthub.io/packages/helm/helm-openldap/openldap ).

it feels like docker container was started in pre-helm days, and hence generates runtime config files on the fly from environment variables, instead of accepting config from helm directly.

PR #19929, opened 2024-05-22
github: add dependabot to update github actions

xnox:
Not sure if any other settings need to be adjusted on the repo. It
seems like some of our actions are not being upgraded. At least one
action is out of date which dependabot would upgrade.


k4leung4 (approved):

xnox:
Hooray, it is useful, already it opened 5 prs to upgrade things.

PR #20051, opened 2024-05-23
Add staging and prod lofo to trust policy

found-it:
Adding this trust policy to start collecting logs and artifacts


.github/chainguard/lofo.sts.yaml
```diff
@@ -2,7 +2,11 @@
 
 issuer: https://accounts.google.com
 
-subject: "107094302893801739313"
+
+# 107094302893801739313 -
```
k4leung4, line 6: whats this one for?

found-it: Not sure, it existed previously. The only other person using this service is @joshrwolf (that I know of) but I looked in his project and didn't see it. I'm waiting until he's back online to address it

k4leung4 (approved):

PR #20138, opened 2024-05-24
give rust builds more resources

imjasonh:
These are requesting 2 CPUs currently, using 4 CPUs and taking an hour. Bumping the request to 16 CPUs to hopefully see them finish faster and use all the resources given. We can bump the request higher after that to see if that's useful.


jonjohnsonjr (approved):

imjasonh:
Merging this past presubmit since it doesn't change the epoch and rebuilding is just wasting time (and likely to time out)

PR #20234, opened 2024-05-26
build gitsign with ldflags configured

kranurag7:
Ref: https://github.com/sigstore/gitsign/blob/7c3d86db51bce4cff1fc481b8733615b9830b9d5/.goreleaser.yaml#L21-L25

### DIFF
--- a/gitsign.yaml
+++ b/gitsign.yaml
@@ -1,10 +1,14 @@
 package:
   name: gitsign
   version: 0.10.2
-  epoch: 0
+  epoch: 1
   description: Keyless Git signing with Sigstore!
   copyright:
     - license: Apache-2.0
+
+environment:
+  environment:
+    CGO_ENABLED: "0"
 
 pipeline:
   - uses: git-checkout
@@ -17,6 +21,11 @@ pipeline:
     with:
       packages: .
       output: gitsign
+      ldflags: |
+        -s -w
+        -extldflags=-znow
+        -extldflags=-zrelro
+        -X github.com/sigstore/gitsign/pkg/version.gitVersion=${{package.version}}
 
   - uses: strip
 


ajayk (approved):

gitsign.yaml
```diff
@@ -17,6 +21,11 @@ pipeline:
     with:
       packages: .
       output: gitsign
+      ldflags: |
+        -s -w
```
xnox, line 25: This strips symbols table from the binary, thus making it not work with [go-vulnce](https://go.dev/blog/vuln)

kranurag7: @xnox In all of my go related patches, I've omitted the symbols table. This is what I do in all the kubernetes controllers as well and I think I mostly got this in habit from looking a lot of controllers build. I was happy because the binary in the end with getting reduced in size meaning less overhead over the wire. 

From the reference above, I was not able to find that by stripping off the symbols table, we are violating something that govulncheck recommends. I had this in mind that govulncheck mostly scans _(go.mod, go.sum)_ deps rather than doing a runtime binary analysis. 

Let me know and I'm happy to revert and send a fix bumping the epoch.

kranurag7: @xnox thanks again, I was wrong about govulncheck parsing go.mod and go.sum files. I quickly skimmed over https://pkg.go.dev/golang.org/x/vuln/internal/vulncheck/internal/gosym#section-readme and now it's clear that it's getting this information from symbols table. 

I think as per the blog https://go.dev/blog/govulncheck this says that you can run it inside any project and that made me think that it's parsing the dependencies (go.mod and go.sum) somehow but I think it's first building the binary and then getting information from the symbols table. 

```bash
$ docker run --rm -it cgr.dev/chainguard/wolfi-base sh
/ # apk add go kubectl
(1/30) Installing ncurses-terminfo-base (6.5_p20240629-r0)
# truncated
(29/30) Installing kubectl-1.30 (1.30.3-r0)
(30/30) Installing kubectl-1.30-default (1.30.3-r0)
OK: 677 MiB in 44 packages
/ # go version -m $(command -v kubectl)
# truncated 
        build   -ldflags="all=-X 'k8s.io/client-go/pkg/version.buildDate=2024-07-17T13:48:48Z' -X 'k8s.io/component-base/version.buildDate=2024-07-17T13:48:48Z' -X 'k8s.io/client-go/pkg/version.gitCommit=6fc0a69044f1ac4c13841ec4391224a2df241460' -X 'k8s.io/component-base/version.gitCommit=6fc0a69044f1ac4c13841ec4391224a2df241460' -X 'k8s.io/client-go/pkg/version.gitTreeState=clean' -X 'k8s.io/component-base/version.gitTreeState=clean' -X 'k8s.io/client-go/pkg/version.gitVersion=v1.30.3' -X 'k8s.io/component-base/version.gitVersion=v1.30.3' -X 'k8s.io/client-go/pkg/version.gitMajor=1' -X 'k8s.io/component-base/version.gitMajor=1' -X 'k8s.io/client-go/pkg/version.gitMinor=30' -X 'k8s.io/component-base/version.gitMinor=30'  -s -w"
```

As of now, there a lot of packages including kubectl which I tested that's stripping off symbols table. Should I create a patch making sure that symbols table is included in the go binaries? Also, We don't have a CI check for this as of now.

PR #20284, opened 2024-05-28
elastic: mass resource tweak; lowers resource to allow better binpacking

k4leung4:


imjasonh:
Bypassing presubmit because this is just a resource change and doesn't need to be rebuilt in CI.

imjasonh (approved):

PR #20312, opened 2024-05-28
require at least one primary or subpackage test to merge

joshrwolf:
This lint check is satisfied with either:

- a primary package test (`.test`)
- at least one subpackage test (`.subpackage[].test`)

Absent is the check that the test is of a sufficient quality, but we won't gate on that (yet!).

Notably, this hooks into our `make list`, which requires an epoch bump, so this doesn't prevent a package without an epoch bump from sneaking in without a test, but we have other checks for that

### DIFF
--- a/R-wk.yaml
+++ b/R-wk.yaml
@@ -1,7 +1,7 @@
 package:
   name: R-wk
   version: 0.9.1
-  epoch: 0
+  epoch: 2
   description: Lightweight Well-Known Geometry Parsing
   copyright:
     - license: MIT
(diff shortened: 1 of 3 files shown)


dlorenc:
This is going to be really tricky for library-only packages, is this check going to start in blocking mode? Do we have a way to override it during bumps/CVE fixes for existing packages that don't have tests?

rawlingsj:
Yeah the lint.sh is a required check so that would cause a few issues for overriding it with the volume of automated PRs we are getting.

joshrwolf:
Based on feedback, I took a different approach and plumbed the logic in `wolfictl lint`, the PR is here: https://github.com/wolfi-dev/wolfictl/pull/920

PR #20341, opened 2024-05-28
chore(neuvector): Drop streaming, fix updates, and remove certs

EyeCantCU:
- Removes certs from manager and scanner, these should be generated
- Filters out nonstandard releases
- Use new versioning scheme for nv promotheus exporter
- Drop 5.3 from package name, NeuVector doesn't appear to support prior versions

### DIFF
--- a/neuvector-controller.yaml
+++ b/neuvector-controller.yaml
@@ -62,8 +62,7 @@ pipeline:
 update:
   enabled: true
   ignore-regex-patterns:
-    - '.*beta.*'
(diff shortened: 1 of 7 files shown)


jamonation (approved):
Nice, thanks for tidying this up and sorting out

rawlingsj:
We should also rename the advisory https://github.com/wolfi-dev/advisories/blob/main/neuvector-controller-5.3.advisories.yaml

PR #20428, opened 2024-05-29
use tags for identifying releases

joshrwolf:


docker-cli.yaml
```diff
@@ -67,3 +67,4 @@ update:
     identifier: docker/cli
     strip-prefix: v
     tag-filter: v
+    use-tag: true
```
ajayk, line 70: can we please ignore the rc versions in the same PR :-)

joshrwolf: added via `ignore-regex-pattern`, ptal and lmk if that was wrong/dumb

ajayk (approved):

PR #20469, opened 2024-05-30
fix(neuvector): Add missing runtime dependencies and correct binary paths

EyeCantCU:
Adds missing runtime dependencies for NeuVector manager, controller, and scanner. Removes unnecessary runtime dependency from NeuVector monitor. Also corrects binary names and paths (these are hardcoded throughout NeuVector)

### DIFF
--- a/neuvector-monitor.yaml
+++ b/neuvector-monitor.yaml
@@ -1,7 +1,7 @@
 package:
   name: neuvector-monitor
   version: 5.3.2
-  epoch: 1
+  epoch: 3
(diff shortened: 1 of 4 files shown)


neuvector-monitor.yaml
```diff
@@ -1,13 +1,10 @@
 package:
   name: neuvector-monitor
   version: 5.3.2
-  epoch: 2
+  epoch: 1
```
developer-guy, line 4: epoch should be bump to 3

developer-guy (approved):

PR #20470, opened 2024-05-30
feat(packages): Add NeuVector database and db generator

EyeCantCU:
Needed to generate database for NeuVector scanner

### DIFF
--- a/neuvector-dbgen.yaml
+++ b/neuvector-dbgen.yaml
@@ -17,6 +17,11 @@ pipeline:
       expected-sha256: 8f334ddf3b99aefeef7884e4f126c03e7f5a44fa288b9aaba902b410150d27c9
       uri: https://github.com/neuvector/vul-dbgen/archive/23c9cea6be0762c825673803624300cc4a720d19.tar.gz
 
+  # CVE-2022-29526
+  - uses: go/bump
+    with:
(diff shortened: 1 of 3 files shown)


jamonation (dismissed):

neuvector-dbgen.yaml
```diff
@@ -0,0 +1,54 @@
...
+          output: updater
+          prefix: usr/local
+          subpackage: true
+          vendor: true
+      - uses: strip
+
+test:
+  pipeline:
+    - runs: |
+        dbgen --help
```
jamonation, line 51: Looks like this --help is exiting non-zero :(

EyeCantCU: How on earth....

EyeCantCU: Updated the test to just check for the executable. While --help is built in, it still returns nonzero as if it wasn't 🤦. This is great

jamonation (dismissed):

jonjohnsonjr (approved):

xnox:
This seems to take over 3 hours to build in post-submit and is killing all builds.

I am proposing to revert this temporarily, check why it takes so long to build, and possibly resubmit it by itself - and/or split (it seems to try to download all CVE data of all linux distributions ever and parse it all and stalls).

PR #20486, opened 2024-05-30
kustomize: build with default settings

xnox:
Do not strip symbols table (for determination if fips is needed, and
vulnerability scanning). Do not set CGO_ENABLED - as it is
automatically used as needed.


pnasrat (approved):
LGTM

PR #20726, opened 2024-05-31
xz: enable update and update to 5.6.2

ajayk:


xnox:
is this one without backdoors? :smile:

rawlingsj (approved):

PR #21254, opened 2024-06-04
envoy-ratelimit: include commit data in package

imjasonh:
This project doesn't do upstream releases, and we'd like a way to be able to determine the version being used from image contents. This is a cheap and cheerful way to accomplish that.

### DIFF
--- a/envoy-ratelimit.yaml
+++ b/envoy-ratelimit.yaml
@@ -34,6 +34,8 @@ pipeline:
       output: ratelimit
       packages: ./src/service_cmd
 
+  - runs: echo "${var.commit}" > /var/envoy-ratelimit-commit
+
 subpackages:
   - name: envoy-ratelimit-compat
     pipeline:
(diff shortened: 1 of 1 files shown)


xnox (commented):
> uri: https://github.com/envoyproxy/ratelimit/archive/${{vars.commit}}.zip

i think this is automatic export, which we could convert to a git checkout with a rewind to a desired commit and then go/build would include vcs info in the build.

Separately, i wonder if we can set -vcs.commit directly to ensure it ends up in the binarie's `go version -m ratelimit` output maybe?!

likely very wishlist desires.

xnox:
@imjasonh i saw date tags option in chainctl - but none of the images have them really, are we able to publish date tags?

imjasonh:
> @imjasonh i saw date tags option in chainctl - but none of the images have them really, are we able to publish date tags?

Chainguard Images customers can enable unique tags, which produce tags like `:latest-202406041118`, but we're trying to get envoy-ratelimit tagged with the commit, for compatibility with the upstream image.

imjasonh:
> i think this is automatic export, which we could convert to a git checkout with a rewind to a desired commit and then go/build would include vcs info in the build.

I like this, this is now doing that.

xnox (approved):

PR #21342, opened 2024-06-04
fix(rqlite): update entrypoint

stormqueen1990:
The entrypoint for rqlite has moved into the main repository. Replace the hardcoded entrypoint we previously had for a patch to use `/usr/bin` instead.


stormqueen1990:
Bumped epoch to 2 due to https://github.com/wolfi-dev/os/pull/21294

ajayk (dismissed):

ajayk (approved):

PR #21410, opened 2024-06-05
feat: new package docker-nginx

stormqueen1990:
Create new package for docker-nginx, so the Docker entrypoints can be consumed.


pnasrat (approved):
LGTM

PR #21423, opened 2024-06-05
Add jre runtime packages

mamccorm:
Previously we where defining the JRE runtime packages at the image level - but for consistency with other packages, moving to runtime dependencies in the package instead. We originally believed we may need to switch out the JDKs, but this hasn't been the case.


k4leung4 (commented):
please bump epoch

EyeCantCU (approved):

PR #21269, opened 2024-06-04
Rename rabbitmq-server to rabbitmq-server-3.13

jdolitsky:
replaces https://github.com/wolfi-dev/os/pull/16029 and https://github.com/wolfi-dev/os/pull/20756

PR #21445, opened 2024-06-06
Fix provides for rabbitmq-server-compat-3.13 subpackage

jdolitsky:
In https://github.com/wolfi-dev/os/pull/21269 we modified the package name `rabbitmq-server-compat` to be `rabbitmq-server-compat-3.13`. Now when we try to install `rabbitmq-server-compat`, there was nothing that provides it because it was always incorrectly providing `rabbitmq-bitnami-compat=${{package.full-version}}`. It does not appear that the `rabbitmq-bitnami-compat` package (virtual package) is used anywhere.

### DIFF
--- a/rabbitmq-server-3.13.yaml
+++ b/rabbitmq-server-3.13.yaml
@@ -1,7 +1,7 @@
 package:
   name: rabbitmq-server-3.13
   version: 3.13.3
-  epoch: 0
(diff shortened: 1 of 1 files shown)


jdolitsky:
closing in favor of https://github.com/wolfi-dev/os/pull/21448

PR #21464, opened 2024-06-06
esbuild: convert to go/build

xnox:
This introduces x86-64-v2 optimisation, and adds symbols table, and
makes it easier to create a fips build.


pnasrat (approved):
LGTM

PR #21470, opened 2024-06-06
vertical-pod-autoscaler: go cleanups

xnox:
Remove duplicate go/bump.

Remove duplicate copies of dependencies, which pipelines self-declare.

Move go/build pipelines from main package to their respective
subpackages, instead of moving binaries to subpackages by hand.

Drop strip pipeline, as go/build builds go binaries stripped already.

This will make it easier to maintain fips version of the package.


pnasrat (approved):
LGTM

PR #21523, opened 2024-06-06
Updates 9 python packages to use git-checkout instead of fetch. Upstr…

dakaneye:
…eam repos sourced from pypi homepages

Fixes: 9 python packages move to gitcheckout instead of fetch

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/py3-attrs.yaml
+++ b/py3-attrs.yaml
@@ -35,10 +35,9 @@ pipeline:
   - name: Fetch History (required by pip install)
     runs: |
       git fetch --unshallow
-      export SOURCE_DATE_EPOCH=$(date +%s)
 
   - name: Python Install
-    runs: pip3 install . --root=${{targets.destdir}}
+    runs: export SOURCE_DATE_EPOCH=$(date +%s) && pip3 install . --root=${{targets.destdir}}
 
   - uses: strip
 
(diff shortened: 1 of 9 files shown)


py3-attrs.yaml
```diff
@@ -21,21 +21,35 @@ environment:
...
+      tag: ${{package.version}}
+      expected-commit: 9e443b18527dc96b194e92805fa751cbf8434ba9
+
+  - name: Fetch History (required by pip install)
+    runs: |
+      git fetch --unshallow
 
   - name: Python Install
-    runs: pip3 install . --root=${{targets.destdir}}
+    runs: export SOURCE_DATE_EPOCH=$(date +%s) && pip3 install . --root=${{targets.destdir}}
```
ajayk, line 40:
```suggestion
    runs: export SOURCE_DATE_EPOCH=315532800 && pip3 install . --root=${{targets.destdir}}
```

dakaneye: was talking w/ @xnox who suggested that initially but we pivoted to trying to set it to the commit time: https://chainguard-dev.slack.com/archives/C063AT3FVFA/p1717712979583829, falling back to the minimum allowed time like you suggested.

xnox (approved):

PR #21565, opened 2024-06-07
new package: sbom-convert

dlg1206:
Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/sbom-convert.yaml
+++ b/sbom-convert.yaml
@@ -11,6 +11,7 @@ pipeline:
     with:
       repository: https://github.com/protobom/sbom-convert
       tag: v${{package.version}}
+      expected-commit: 0e9c07a02126ec383ed1618aea7386098fdfc542
 
   - uses: go/bump
     with:
--- a/sbom-convert.yaml
+++ b/sbom-convert.yaml
@@ -1,0 +1,24 @@
+package:
+  name: sbom-convert
+  version: 0.0.4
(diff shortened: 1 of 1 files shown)


sbom-convert.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - uses: go/bump
+    with:
+      deps: google.golang.org/protobuf@v1.33.0
+
+  - uses: go/build
+    with:
+      output: sbom-convert
+      packages: .
+
+  - uses: strip
```
imjasonh, line 24: We should also add an `update:` block here, which will tell Wolfi's automation when/how to update the package when 0.0.5+ comes out. You can find some examples throughout the repo, but it should be pretty simple since it's just on GitHub. Look for

```
update:
  enabled: true
  github:
    ... 
```

dlg1206: Added the `update` block

sbom-convert.yaml
```diff
@@ -0,0 +1,31 @@
...
+
```
ajayk, line 26: Can you please add a basic test  , here is a simple example of a [test](https://github.com/wolfi-dev/os/blob/main/helm.yaml#L34-L39)

dlg1206: Added a simple test that checks the tool version

ajayk (approved):

PR #21581, opened 2024-06-07
Update py3-rfc3339-validator.yaml

imjasonh:
The previous change to this updated its license to be a valid identifier, but didn't bump the epoch to rebuild.


pnasrat (approved):
LGTM

PR #21670, opened 2024-06-10
bats: New package Bash Automated Testing System

pnasrat:
Source: https://github.com/bats-core/bats-core/tree/master

Docs: https://bats-core.readthedocs.io/en/stable/

Security policy: https://github.com/bats-core/bats-core/security
Security advisories: https://github.com/bats-core/bats-core/security/advisories

#### For new package PRs only

### DIFF
--- a/bats.yaml
+++ b/bats.yaml
@@ -41,4 +41,3 @@ update:
   github:
     identifier: bats-core/bats-core
     strip-prefix: v
-    use-tag: true
--- a/bats.yaml
+++ b/bats.yaml
@@ -1,0 +1,44 @@
+package:
+  name: bats
+  version: 1.11.0
+  epoch: 0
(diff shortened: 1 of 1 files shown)


bats.yaml
```diff
@@ -0,0 +1,44 @@
...
+    pipeline:
+      - uses: split/manpages
+    description: bats manpages
+
+update:
+  enabled: true
+  github:
+    identifier: bats-core/bats-core
+    strip-prefix: v
+    use-tag: true
```
stormqueen1990: Looks like BATS uses releases, should we prefer those over the tags?

```suggestion
```

pnasrat: @stormqueen1990 already pushed that!

ajayk (approved):

PR #21772, opened 2024-06-11
Bump epoch to get latest go changes for newrelic-fluent-bit-output

hectorj2f:


hectorj2f:
Closing the PR! We need to file one entry in the advisory. This cve fix may not get backported to go-1.20.

PR #21943, opened 2024-06-13
Remove bun

xnox:
Embeds copies of zig, webkit. Requires old clang. Fails to build
and/or execute. Very long build times. Unused and unsupportable.

Resolves:
- https://github.com/wolfi-dev/os/pulls?q=is%3Apr+bun


imjasonh:
+1 on removing and not building in the future without strong signal that the time is worth it. 

Do we need to withdraw existing packages?

cpanato (approved):

xnox:
> Do we need to withdraw existing packages?

yes. they will accumulate binaries, not supportable and cannot be upgraded. Imho we should delete all renamed & removed origs, to prevent them from being installable.

Multiple times removed/renamed packages break image building - i.e. old real python3 getting preferred over up to date python-3.10 and so on.

imjasonh:
I'm sold.

PR #21944, opened 2024-06-13
fix: replace busybox with more targeted package deps

stormqueen1990:
Busybox isn't required here, use more targeted package dependencies instead.


smoser:
I'm not really sure what I think here.
Pros:
 * more targetted deps
Cons:
 * probably a bigger image in size and code
 * harder to maintain

PR #22005, opened 2024-06-13
fix(nginx): more fixes for config

stormqueen1990:


jamonation:
Looks good, @EyeCantCU and I both built these packages locally, then made images that use the `-r4` versions without any issues.

jamonation (approved):

PR #22027, opened 2024-06-14
git checkouts - tiff tzdata zlib zstd

smoser:
convert these packages to git checkout.

### DIFF
--- a/tiff.yaml
+++ b/tiff.yaml
@@ -1,7 +1,7 @@
 package:
   name: tiff
   version: 4.6.0 # Remove CVE patches at the next release
-  epoch: 3
+  epoch: 4
   description: Provides support for the Tag Image File Format or TIFF
   copyright:
     - license: libtiff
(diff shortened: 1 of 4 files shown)


tiff.yaml
```diff
@@ -16,24 +16,28 @@ environment:
       - ca-certificates-bundle
```
dakaneye, line 16: we should iterate the epoch

tzdata.yaml
```diff
@@ -16,17 +16,11 @@ environment:
       - wolfi-baselayout
 
 pipeline:
-  - uses: fetch
+  - uses: git-checkout
     with:
-      uri: https://www.iana.org/time-zones/repository/releases/tzcode${{package.version}}.tar.gz
```
dakaneye, line 21: wow, two fetches for stuff in one repo before?

dakaneye: we should iterate the epoch on tzdata too

dakaneye (changes requested):
LGTM besides epoch updates.

tiff.yaml
smoser: will fix epoch.

smoser: fixed.

dakaneye (approved):

PR #22102, opened 2024-06-16
sshd won't start without a /var/empty directory

dustinkirkland:
Try to start /usr/sbin/sshd in wolfi and you'll get the following error, solved by a mkdir /var/empty:

```
wolfi:/etc/ssh# /usr/sbin/sshd            
Missing privilege separation directory: /var/empty
wolfi:/etc/ssh# mkdir /var/empty
wolfi:/etc/ssh# /usr/sbin/sshd 
wolfi:/etc/ssh# ps -ef | grep sshd
  357 root      0:00 sshd: /usr/sbin/sshd [listener] 0 of 10-100 startups
```


kaniini:
In my opinion it would be better to add this directory to wolfi-baselayout.

openssh.yaml
```diff
@@ -122,11 +122,16 @@ subpackages:
...
+          chmod 755 "${{targets.subpkgdir}}"/var/empty
           mv "${{targets.destdir}}"/usr/sbin/sshd "${{targets.subpkgdir}}"/usr/sbin/
     dependencies:
       runtime:
         - openssh-keygen
         - openssh-server-config
+    scriptlets:
+      post-install: |
+        #!/bin/sh
+        ssh-keygen -q -N "" -t rsa -b 4096 -f /etc/ssh/ssh_host_rsa_key
```
kaniini, line 134: Why not ssh-keygen -A to generate all of the host keys at once?

dustinkirkland: Good call!  I just  pushed that change.

dustinkirkland:
Happy to move the mkdir /var/empty to wolfi-baselayout, if it belongs there.  I don't know that it's caused any other problem besides making sshd fail, so I put it here.  But I'll move it there if you think that's better....

openssh.yaml
```diff
@@ -121,12 +121,17 @@ subpackages:
...
+          chmod 755 "${{targets.subpkgdir}}"/var/empty
           mv "${{targets.destdir}}"/usr/sbin/sshd "${{targets.subpkgdir}}"/usr/sbin/
     dependencies:
       runtime:
         - openssh-keygen
         - openssh-server-config
+    scriptlets:
+      post-install: |
+        #!/bin/sh
+        ssh-keygen -A -q -N ""
```
imjasonh, line 134: I'm skeptical of scriptlets in general, since they're not executed by apko, only on `apk add`. Is there a way to get this side effect for apko-built users too?

openssh.yaml
```diff
@@ -121,12 +121,17 @@ subpackages:
     description: "OpenSSH server"
     pipeline:
       - runs: |
-          mkdir -p "${{targets.subpkgdir}}"/usr/sbin
+          mkdir -p "${{targets.subpkgdir}}"/usr/sbin "${{targets.subpkgdir}}"/var/empty
+          chmod 755 "${{targets.subpkgdir}}"/var/empty
```
imjasonh, line 125: Could you add a test in this PR as well? You said the change fixes a behavior where the daemon doesn't come up, it would be great to reflect that in a test to safeguard against regression.

dustinkirkland: Sure, good idea.  I added a test.

openssh.yaml
dustinkirkland: Yep.  Teach apko to run scriptlets.

joshrwolf (dismissed):

openssh.yaml
joshrwolf: looks like this is failing in presubmit tests:

```
2024/06/17 15:04:57 INFO running the main test pipeline
2024/06/17 15:04:57 WARN + export 'PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin'
2024/06/17 15:04:57 WARN + '[' -d /home/build ]
2024/06/17 15:04:57 WARN + cd /home/build
2024/06/17 15:04:57 WARN + /usr/sbin/sshd
2024/06/17 15:04:57 WARN sshd: no hostkeys available -- exiting.
```

Likely becauase we use apko to build the test workspace, which doesn't yet know about scriptlets

dustinkirkland: Well....should I force the keygen in the test itself to make it pass?

dustinkirkland: I just pushed a change that forces the ssh key gen in the test itself, with a comment

joshrwolf (approved):

openssh.yaml
xnox: > Yep. Teach apko to run scriptlets.

That makes no sense here =) because we do not want to bake the same host key into the golden image at apko execution time.

We would only want to do this at image startup / runtime, and ideally provision onto a persistent volume.

xnox (approved):

PR #22109, opened 2024-06-16
local-path-provisioner: add test

Dentrax:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/local-path-provisioner.yaml
+++ b/local-path-provisioner.yaml
@@ -42,3 +42,11 @@ update:
   github:
     identifier: rancher/local-path-provisioner
     strip-prefix: v
+
+test:
+  pipeline:
+    - runs: |
+        set +e
+        output=$(local-path-provisioner --debug start 2>&1)
+        set -e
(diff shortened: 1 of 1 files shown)


local-path-provisioner.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: local-path-provisioner
   version: 0.0.27
-  epoch: 1
+  epoch: 2
```
EyeCantCU, line 4: When we're only modifying/adding tests, this is one of few cases where we don't need to increment the epoch

EyeCantCU (changes requested):

local-path-provisioner.yaml
cpanato: +1

EyeCantCU (approved):

local-path-provisioner.yaml
Dentrax: TIL that thank you!

PR #22150, opened 2024-06-17
adds contour package

mritunjaysharma394:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/contour.yaml
+++ b/contour.yaml
@@ -1,10 +1,13 @@
 package:
-  name: contour
-  version: 1.29.1
+  name: contour-1.27
+  version: 1.27.4
   epoch: 0
   description: Contour is a Kubernetes ingress controller using Envoy proxy.
   copyright:
     - license: Apache-2.0
+  dependencies:
+    provides:
+      - contour=${{package.full-version}}
 
 environment:
   contents:
@@ -21,7 +24,7 @@ pipeline:
     with:
       repository: https://github.com/projectcontour/contour
       tag: v${{package.version}}
-      expected-commit: be445838b789500b1ef4241b61d96afc3316989e
+      expected-commit: 64dc935bffbfe30ba42b409362770d142d8a4948
 
   - uses: go/build
     with:
--- a/contour.yaml
+++ b/contour.yaml
@@ -1,0 +1,57 @@
+package:
+  name: contour
+  version: 1.29.1
+  epoch: 0
+  description: Contour is a Kubernetes ingress controller using Envoy proxy.
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - go
+  environment:
+    CGO_ENABLED: "0"
+    GO111MODULE: "on"
+
+pipeline:
+  - uses: git-checkout
+    with:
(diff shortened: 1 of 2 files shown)


contour.yaml
```diff
@@ -0,0 +1,48 @@
+package:
+  name: contour
```
rawlingsj, line 2: should this be a version stream?  I "think" if you create the lowest in support version stream i.e 
```
package:
  name: contour-1.27
  version: 1.27.4
```

To match [Contour v1.27.4](https://github.com/projectcontour/contour/releases/tag/v1.27.4) then automation should auto create PRs for the .28 and .29 streams.

developer-guy: the lowest version stream would be 1.27.x and the latest patch version of that stream is 4 so the version of that package should be 1.27.4, I agree with @rawlingsj here.

developer-guy: we should also create a [package-version-metadata](https://github.com/chainguard-dev/package-version-metadata/) for contour

mritunjaysharma394: Done, also created the same for that @developer-guy, will refactor it once we have contour in eol https://github.com/chainguard-dev/package-version-metadata/pull/551

contour.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+  - uses: strip
+
+update:
+  enabled: true
+  github:
+    identifier: projectcontour/contour
+    strip-prefix: v
+
+# requires k8s and kwok alone won't help so going with a simple test in melange
```
mamccorm, line 47: nit - but i'd personally just drop this comment, otherwise we'd be adding it for most melange files

mritunjaysharma394: true lol, I will remove it, thanks Mark for the review

contour.yaml
```diff
@@ -0,0 +1,51 @@
...
+        -X github.com/projectcontour/contour/internal/build.Sha=$(git rev-parse HEAD)
+        -X github.com/projectcontour/contour/internal/build.Branch=$(git rev-parse --abbrev-ref HEAD)
+
+  - uses: strip
+
+update:
+  enabled: true
+  github:
+    identifier: projectcontour/contour
+    strip-prefix: v
```
mamccorm, line 45: This needs to filter to the specific version stream / tag in the repo, otherwise it'll get upgraded but the package name remains the old version

mamccorm: i.e tag-filter

contour.yaml
```diff
@@ -0,0 +1,51 @@
+package:
```
EyeCantCU, line 1: The file name should match the package name: `contour-1.27.yaml`

EyeCantCU (changes requested):

contour.yaml
mritunjaysharma394: oh yes, I forgot to change it while switching to v1.27, thanks RJ

contour.yaml
mritunjaysharma394: Got it, on my way

contour.yaml
EyeCantCU: No problem!

EyeCantCU (approved):

PR #20318, opened 2024-05-28
python dev

xnox:
- **Make several core python packages build for mulitple python versions.**
  At this point most things should use the py/pip-build-install pipeline
  and generally look like py3-pyyaml
  
  Also:
  * rename cython.yaml, numpy.yaml to py3-<name>
  
    These are python packages and present in pypi, for
    consistency, name them like other python packages.
  
  Co-authored-by: Dimitri John Ledkov <dimitri.ledkov@chainguard.dev>
  

- **Resolve the dependency loop.**
  flit-core has bootstrap_install.py to handle this, so we use it.
  
  pip doesn't stricktly depend on wheel, although buidling many projects
  assumes that it is there.
  

- **Fun stuff**
  Signed-off-by: Dimitri John Ledkov <dimitri.ledkov@chainguard.dev>
  

- **Revert "Fun stuff"**
  hahhah just kidding
  
  This reverts commit 3a964132ebe26225b8510ec2875b506a8e617633.
  

- **Drop pybins and allow to install them all**
  Requires https://github.com/chainguard-dev/melange/pull/1166
  
  Drop pybins, instead build all packages with commands in /usr/bin.
  Set replaces, replaces-priority to ensure they are apk installable.
  Also set provider-priority such that `apk add py3-pip` transparently
  installs py3.12-pip alone.
  
  This is all great, however not at all supported by apko. But imho
  should be supported. As one simply has to sort packages in replaces
  priority order prior to streaming. And allow overriding files between
  packages that have replaces declared.
  
  ```console
  ...
  (13/25) Installing python-3.10-base (3.10.14-r1)
  (14/25) Installing py3.10-flit-core (3.9.0-r2)
  (15/25) Installing py3.10-setuptools (69.5.1-r1)
  (16/25) Installing py3.10-pip (24.0-r2)
  (17/25) Installing python-3.11-base (3.11.9-r1)
  (18/25) Installing py3.11-flit-core (3.9.0-r2)
  (19/25) Installing py3.11-setuptools (69.5.1-r1)
  (20/25) Installing py3.11-pip (24.0-r2)
  (21/25) Installing python-3.12-base (3.12.3-r1)
  (22/25) Installing py3.12-flit-core (3.9.0-r2)
  (23/25) Installing py3.12-setuptools (69.5.1-r1)
  (24/25) Installing py3.12-pip (24.0-r2)
  (25/25) Installing py3-supported-pip (24.0-r2)
  OK: 197 MiB in 40 packages
  
[...]

PR #22162, opened 2024-06-17
pip-zipapp: New package for bootstrapping python pipeline

pnasrat:
See #20318 needed to bootstrap python pipeline

#### For new package PRs only

- [ ] This PR is part of python toolchain improvements
- [ x REQUIRED - The version of the package is still receiving security updates
- [ n/a ] This PR links to the upstream project's support policy (e.g. `endoflife.date`) (see main pip package)

### DIFF
--- a/pip-zipapp.yaml
+++ b/pip-zipapp.yaml
@@ -63,8 +63,9 @@ pipeline:
       cp *.whl "$destd/$sdir/wheels/"
 
 update:
-  ignore-regex-patterns:
-    - b
   enabled: true
+  ignore-regex-patterns:
+    - 'b'
   github:
     identifier: pypa/pip
+    use-tag: true
--- a/pip-zipapp.yaml
+++ b/pip-zipapp.yaml
@@ -26,37 +26,31 @@ pipeline:
       uri: https://bootstrap.pypa.io/pip/zipapp/pip-${{package.version}}.pyz
       expected-sha256: 1af141650f2b1e71bf9d856db74324042b4bfe3a36c37f8c2b3e2d8d6b3da411
       extract: false
-
   - uses: fetch
     with:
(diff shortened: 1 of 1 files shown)


smoser (dismissed):

pip-zipapp.yaml
```diff
@@ -0,0 +1,65 @@
...
+  description: "Pip as a python zipapp"
+  copyright:
+    - license: MIT
+
+environment:
+  contents:
+    packages:
+      - busybox
+
+pipeline:
```
vaikas, line 14: Could we add some comments for where the various sources below are fetched from, and why. Not being a python expert, it's a bit tricky to reason about it. Things like if there's an update to pypa/pip, how do these get updated.

pip-zipapp.yaml
```diff
@@ -0,0 +1,65 @@
...
+      cp "pip-${{package.version}}.pyz" "$destd/$sdir/$fname"
+      chmod 0644 "$destd/$sdir/$fname"
+
+      mkdir -p "$destd/$sdir/wheels"
+      cp *.whl "$destd/$sdir/wheels/"
+
+update:
+  enabled: true
+  github:
+    identifier: pypa/pip
```
vaikas, line 65: Looks like there are beta releases, so wonder if they should be excluded too?
https://github.com/pypa/pip/tags

pnasrat: @vaikas done

pip-zipapp.yaml
pnasrat: Done, I'll add an issue once merged to see if we can make this clearer

pnasrat:
Also filed https://github.com/wolfi-dev/wolfictl/issues/960 to make wolfictl support full semver pre-releases

smoser (approved):
looks good.

PR #22174, opened 2024-06-17
add: Cloudnative-pg(CloudNativePG is designed to seamlessly manage PostgreSQL databases)

Aditevil:
…ge PostgreSQL databases

Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/cloudnative-pg.yaml
+++ b/cloudnative-pg.yaml
@@ -35,6 +35,7 @@ subpackages:
           packages: ./cmd/kubectl-cnpg
           output: plugins
           subpackage: true
+
   - name: ${{package.name}}-compat
     pipeline:
       - runs: |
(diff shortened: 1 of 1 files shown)


cloudnative-pg.yaml
```diff
@@ -0,0 +1,62 @@
+package:
+  name: cloudnative-pg
+  version: 1.23.2
+  epoch: 0
+  description: CloudNativePG is a comprehensive platform designed to seamlessly manage PostgreSQL databases
+  copyright:
+    - license: Apache-2.0
+
+environment:
```
EyeCantCU, line 9: Environment can be dropped since we're using `go/build`

cloudnative-pg.yaml
```diff
@@ -0,0 +1,62 @@
...
+subpackages:
+  - name: ${{package.name}}-plugins
+    pipeline:
+      - uses: go/build
+        with:
+          packages: ./cmd/kubectl-cnpg
+          output: plugins
+          subpackage: true
+
+  - name: ${{package.name}}-compat
```
EyeCantCU, line 39: Hmm... Is the `/manager` path actually needed? Is setting the entrypoint to `/usr/bin/manager` sufficient? We should only create compat packages when we're sure we'll need them :)

Aditvil-Dev: Sure. I'll just remove it for now.

EyeCantCU (approved):

PR #22192, opened 2024-06-18
Adds 5 more packages that now use git-checkout instead of fetch

dakaneye:
Updates 5 packages to use git-checkout over fetch:
* gnutls
* gzip
* haproxy-2.9
* iperf
* iproute2

Note haproxy we disable update for.

### DIFF
--- a/haproxy-2.9.yaml
+++ b/haproxy-2.9.yaml
@@ -80,5 +80,17 @@ subpackages:
 
 update:
   enabled: true
+  # This is a hack to ignore all valid semantic versions greater than version 2.9
+  #   Explanation
+  #     \b: Word boundary to ensure we are matching whole semantic versions, not parts of longer strings.
+  #     (3\.\d+\.\d+|[4-9]\.\d+\.\d+|[1-9]\d+\.\d+\.\d+):
+  #       3\.\d+\.\d+: Matches versions starting with 3 and any minor/patch version.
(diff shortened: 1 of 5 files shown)


haproxy-2.9.yaml
```diff
@@ -32,10 +24,12 @@ environment:
       - pcre2-dev
 
 pipeline:
-  - uses: fetch
+  - uses: git-checkout
     with:
-      uri: https://www.haproxy.org/download/${{vars.mangled-package-version}}/src/haproxy-${{package.version}}.tar.gz
-      expected-sha256: d1a0a56f008a8d2f007bc0c37df6b2952520d1f4dde33b8d3802710e5158c131
+      repository: https://git.haproxy.org/git/haproxy-2.9.git/
```
ajayk, line 29: Officially listed mirror of ha-proxy on GitHub 
```suggestion
      repository: https://github.com/haproxy/haproxy
```

dakaneye: I noticed that! But if you look at the tags in that github repo, they don't seem to follow the same version scheme as the repo above. Although, it does appear to be the official development repository and gets updates more often.

xnox: github only has develpment tags (rc and final release) but it doesn't have stable maintainance branches or tags which are only on haproxy.org. I.e. the point releases we want to package.

Also this packages 2.9 without a package version subsitituion. I was hoping for a var transform to be included in the uri such that 2.9 is automatically updated.

ajayk (approved):

PR #22209, opened 2024-06-18
Bump py3-installer to higher than py3.12-installer

xnox:
Bump py3-installer to 7, such that it is higher than py3.12-installer
which it supersedes.


imjasonh (approved):

imjasonh:
Trying to get builds unblocked, don't want to wait for a bunch of unrelated builds to finish.

PR #22229, opened 2024-06-18
PHP Extensions Load Order

erikaheidi:
This PR should work as a long-term fix for the issue with the order in which PHP extensions are loaded within images.

I am currently working on a WordPress image and it is not possible to use some of the extensions as they are loaded in alphabetical order, even though they have dependencies between each other. So anyone trying to build a PHP image today with the `mysqli` extension will get errors because `mysqli` depends on `mysqlnd`, which is loaded later because of the extension name. The issue is specifically with extension configuration files (.ini), they work like Nginx configuration files, can be renamed to force a specific load order.

This PR does **not** change any extension or package names. Only the configuration files are affected, which should not cause any disruption.

The change uses the file `config.w32` which is present in most extensions and defines other extensions in which this depends on. The script then creates a prefix for the `.ini` files based on the number of deps they have (more deps, higher number, loads later). This is what Alpine does to fix the same problem in the [Alpine PHP build](https://git.alpinelinux.org/aports/tree/community/php82/APKBUILD) (see the [resolve_extension_deps()](https://git.alpinelinux.org/aports/tree/community/php82/APKBUILD#n596) function).

With the change, this is how the configuration files will look like:

```
552b5d2d572f:/work/packages# ls /etc/php/conf.d/
10-mbstring.ini   10-mysqlnd.ini    20-xmlwriter.ini  30-mysqli.ini     30-xmlreader.ini
```

Related: https://github.com/chainguard-dev/internal/issues/3826

### DIFF
--- a/php-8.1.yaml
+++ b/php-8.1.yaml
@@ -1,7 +1,7 @@
 package:
   name: php-8.1
   version: 8.1.29
-  epoch: 0
+  epoch: 1
   description: "the PHP programming language"
   copyright:
     - license: PHP-3.01
@@ -218,9 +218,15 @@ subpackages:
       - runs: |
           export CONF_DIR="${{targets.subpkgdir}}/etc/php/conf.d"
           mkdir -p $CONF_DIR
+          order=10
           prefix=
           [ "${{range.key}}" != "opcache" ] || prefix="zend_"
-          echo "${prefix}extension=${{range.key}}.so" > $CONF_DIR/"${{range.key}}.ini"
+          # Determine load order based on number of dependencies found in config.w32
+          [ -f "ext/${{range.key}}/config.w32" ] && \
+            deps=`sed -En "s/.*ADD_EXTENSION_DEP\('${{range.key}}', ([^)]+)\).*/\1/p" "ext/${{range.key}}/config.w32" \
+                | tr -d "'," | tr ' ' '\n' \
+                | sort -u | wc -w`
+          echo "${prefix}extension=${{range.key}}.so" > $CONF_DIR/"$((order+order*deps))-${{range.key}}.ini"
 
   - name: ${{package.name}}-dev
     description: PHP 8.1 development headers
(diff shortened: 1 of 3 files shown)


erikaheidi:
The build error happens because the `pgsql` extension has a dependency on some extension that is being loaded only afterwards - need to figure out which one. where are these tests defined? I could not find the source for the tests.

xnox (approved):
superb stuff ! also the pgsql "wrong version" is/was blocking icu transition, so very much appreciate this.

erikaheidi:
I just tested an image build here and there is an issue with the `php-pdo` extension, the config is coming out with prefix `30`, which I don't quite understand why since it has only 1 dependency 😓 I am investigating now since this can break things.

erikaheidi:
I think I identified the issue, working on a fix now. The result from the one-liner to check for num of deps is not super accurate, example:
```
echo `sed -En "s/.*ADD_EXTENSION_DEP\('phar', ([^)]+)\).*/\1/p" "phar/config.w32" \
                | tr -d "'," | tr ' ' '\n' \
                | sort -u`
```

return:

```
spl true
```

In this output it's clear that "true" is not a dep, that's why the Alpine Makefile checks the return strings against the list of extensions (**now** I get it!)

erikaheidi:
I am having an issue now that I don't understand. This should work:

```
          [ -f "ext/${{range.key}}/config.w32" ] && \
            deps=`sed -En "s/.*ADD_EXTENSION_DEP\('${{range.key}}', ([^)]+)\).*/\1/p" "ext/${{range.key}}/config.w32" \
                | tr -d "'," | tr ' ' '\n' \
                | sort -u \
                | while read -r dep; do
                  [ "$dep" = "true" ] || echo $dep
                done | wc -w`
```

But at build time the path `ext/${{range.key}}/config.w32` is not being found. This is weird since it was working before! These files are located in the build dir, so I'm wondering if there's an ENV var or something that I could use to access the build dir without relying on relative paths.

PR #22300, opened 2024-06-19
clang-18: add more runtime dependencies

xnox:
It is a very confusing experience when installing clang-18 doesn't
result in ability to compile trivial c++ applications. clang needs a
gcc installation and libstdc++-dev and glibc-dev to ensure it can
dynamically locate internal gcc headers, and thus discover all the
proper include paths to find headers like stddef.h, features.h,
iostream and so on. Without all of these things installed debugging
those failures is quite non-trivial.

Also not sure if our package splits are correct, or worth the effort,
given one typically needs all the things at runtime.

Also not sure why llvm stack is split into so many packages, given
that LLVM supports checking out all the source code side by side, and
build all the things in one go.

### DIFF
--- a/clang-18.yaml
+++ b/clang-18.yaml
@@ -1,7 +1,7 @@
 package:
   name: clang-18
   version: 18.1.8
-  epoch: 0
+  epoch: 1
   description: "C language family frontend for LLVM"
   copyright:
     - license: Apache-2.0
@@ -10,7 +10,13 @@ package:
     memory: 64Gi
   dependencies:
     runtime:
+      - gcc
+      - glibc-dev
       - libLLVM-18
+      - libclang-cpp-18
+      - llvm-18
+    provides:
+      - clang-18-dev=${{package.full-version}}
 
 environment:
   contents:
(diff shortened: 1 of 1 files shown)


smoser:
> It is a very confusing experience when installing clang-18 doesn't
> result in ability to compile trivial c++ applications.

I agree. Thank you!

smoser (commented):
OK, so if the main package (clang) depends on the -dev is there any use for the `-dev` at all?

should we just drop the -dev ? Is there a reason anyone would want -dev without the clang?

smoser:
I made https://github.com/xnox/os/pull/5/files to show what I meant. 
Should we do that to the other clang-* packages? Is there something specific about 18 ?

xnox:
> I made https://github.com/xnox/os/pull/5/files to show what I meant. Should we do that to the other clang-* packages? Is there something specific about 18 ?

I don't think there is anything specific about 18 here.

Separately I have no idea what the difference is between LLVM-18 and llvm-18 and why some runtime libraries are in one and another package...... and if it matters.

I am sort of inclined to say if one needs llvm, one needs all of it. As I have not seen in practice ability to use libLLVM.so in isolation without the rest of the things matching the same major version.

pnasrat (approved):
LGTM

PR #22361, opened 2024-06-20
Git checkout pkgs 6

dakaneye:
Updates 5 packages to use git-checkout instead of fetch:
* ipset
* iptables
* ipvsadm
* isl
* jansson

### DIFF
--- a/ipset.yaml
+++ b/ipset.yaml
@@ -1,7 +1,7 @@
 package:
   name: ipset
   version: "7.22"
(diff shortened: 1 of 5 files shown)


pnasrat:
@dakaneye non-blocking nit for future ones of these can you edit the PR description once it's filed and trim unnecessary template and provide a PR description (eg including the list of package names)

pnasrat (approved):
LGTM

PR #22447, opened 2024-06-21
Git checkout pkgs 10

dakaneye:
Updates packages to use git-checkout:
* libsm
* libssh
* libssh2
* libsodium: Note this one causes an ABI compatibility check failure, but I've bumped the epochs on it's dependent packages and nothing appears broken.

### DIFF
--- a/libsodium.yaml
+++ b/libsodium.yaml
@@ -1,7 +1,7 @@
 package:
   name: libsodium
-  version: 1.0.18
-  epoch: 5
+  version: 1.0.20
+  epoch: 0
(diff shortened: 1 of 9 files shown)


libsodium.yaml
```diff
@@ -45,4 +46,7 @@ subpackages:
 
 update:
   enabled: true
-  manual: true # need to add a melange config to strip the suffix -RELEASE
```
smoser, line 48: nice!

smoser (approved):

PR #22463, opened 2024-06-22
Fix description for iniparser dev

fwilhe2:
It seems like a copy/paste mistake


rawlingsj (changes requested):
Thanks! Would you mind bumping the epoch field by one in this PR? That is required by the build system to build and publish a new revision with the corrected description.

fwilhe2:
Thanks for the hint @rawlingsj 

Updated the epoch field.

rawlingsj (approved):

PR #22520, opened 2024-06-24
Updating Wordpress package to include entrypoint script

erikaheidi:
This PR updates the WordPress package to include an oci entrypoint as subpackage (wordpress-oci-entrypoint), reproducing the behavior of the default WordPress FPM/Alpine image [on Docker Hub](https://github.com/docker-library/wordpress/blob/c37f27433bb26ea3ec3154fcb1f546d855afcff6/latest/php8.3/fpm-alpine/Dockerfile) and following the example of our MariaDB image which also creates a subpackage to share the entrypoint script.

I also added a couple tests to check that the wp-config file is successfully created when database ENV vars are provided.

### DIFF
--- a/wordpress.yaml
+++ b/wordpress.yaml
@@ -1,7 +1,7 @@
 package:
   name: wordpress
   version: 6.5.4
-  epoch: 0
+  epoch: 1
   description: "The Open Source Publishing Platform"
   copyright:
     - license: GPL-2.0-only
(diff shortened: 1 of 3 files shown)


erikaheidi:
I see there is a CI error with the CVE scan. I had investigated these, and they are very old CVEs related to plugins, not the WP core. I don't know why they are being detected since these files are definitely not in the package.

xnox:
@erikaheidi once this lands, I do wonder if respective wordpress image readme.md needs a changelog entry to indicate that "wordpress installation moved from /usr/this/path to /usr/this/path".

xnox (approved):

erikaheidi:
> @erikaheidi once this lands, I do wonder if respective wordpress image readme.md needs a changelog entry to indicate that "wordpress installation moved from /usr/this/path to /usr/this/path".

Luckily the image is not there yet! I am still working on it, should be making a PR soon 🙏🏼

PR #22534, opened 2024-06-24
git checkouts 3 - tcpdump, swaks, strace, squid, skalibs

smoser:
move to git checkout

### DIFF
--- a/squid.yaml
+++ b/squid.yaml
@@ -2,7 +2,7 @@
 package:
   name: squid
   version: "6.10"
-  epoch: 0
+  epoch: 1
   description: Full-featured Web proxy cache server
   copyright:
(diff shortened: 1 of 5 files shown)


squid.yaml
```diff
@@ -15,19 +15,30 @@ environment:
       - build-base
       - busybox
       - ca-certificates-bundle
+      - ed # busybox ed is not sufficient for boostrap.sh use
```
dakaneye, line 18: incomplete comment?

dakaneye (dismissed):
LGTM

squid.yaml
smoser: updated.

krishjainx (approved):
LGTM as well!

PR #22539, opened 2024-06-24
Updates packages to use git-checkout (batch 14)

dakaneye:
Updates packages to use git-checkout over fetch:
* lvm2
* mariadb-10.11
* mariadb-11.2


smoser (approved):
lgtm.

PR #22542, opened 2024-06-24
Cleanup/git checkouts 4 - s6, redis-7.2, readlin, py3-setuptools, python-3.12, python-3.11, python-3.10, protobuf-c

smoser:
change these to git checkout.


dakaneye (dismissed):

smoser:
gnu.org is down, so these are dead for the moment.

luhring (approved):

PR #22557, opened 2024-06-24
Updates some packages to use git-checkout (batch 16)

dakaneye:
Packages:
* net-tools
* nghttp2
* nginx-mainline
* nginx-stable
* npth

### DIFF
--- a/nginx-mainline.yaml
+++ b/nginx-mainline.yaml
@@ -1,7 +1,7 @@
 package:
   name: nginx-mainline
   version: 1.27.0
-  epoch: 4
+  epoch: 5
   description: HTTP and reverse proxy server (mainline version)
   copyright:
     - license: BSD-2-Clause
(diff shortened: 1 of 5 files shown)


netcat-openbsd.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: netcat-openbsd
-  version: "1.226"
```
dakaneye, line 3: Are there any other consequences from changing this versioning scheme? release-monitoring.org doesn't have the -1.1 suffix but the repo does, wasn't sure which to use.

nginx-mainline.yaml
```diff
@@ -108,7 +111,8 @@ pipeline:
           --with-stream_realip_module \
           --with-stream_geoip_module=dynamic \
           --with-stream_ssl_preread_module
-  - runs: |
+  - working-directory: nginx-mainline
```
smoser, line 114: What does the word 'mainline' mean to you?
I would have interpreted this as upstream/main .

maybe `nginx-git` or `nginx-checkout` or `git-checkout` ?

smoser (changes requested):
for nginx, its not useful to the reader to say 'mainline' or 'stable'. I'd suggest making them the same 'nginx-checkout' to reduce differences between the two files.

Feel free to disagree and i can probably be swayed. 

thank you.

nginx-mainline.yaml
dakaneye: AFAICT it refers to the version stream of the package. I was just trying to be consistent with the package naming scheme.

smoser: i hadn't even noticed this was `nginx-mainline.yaml`.

smoser (approved):

PR #22655, opened 2024-06-25
add a dosfstools package for mkfs.fat filesystem formatting

dustinkirkland:
mkfs.fat and friends come from the dosfstools package


dosfstools.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - busybox
+      - ca-certificates-bundle
+      - libtool
+      - linux-headers
+
+pipeline:
+  - uses: fetch
+    with:
+      expected-sha256: 64926eebf90092dca21b14259a5301b7b98e7b1943e8a201c7d726084809b527
+      uri: https://github.com/dosfstools/dosfstools/releases/download/v${{package.version}}/dosfstools-${{package.version}}.tar.gz
```
xnox, line 25: best practice is to use git-checkout, especially when upstream comes from github already. In which case autoreconf may be needed to build the package.

xnox (approved):

dosfstools.yaml
xnox: @dustinkirkland

PR #22693, opened 2024-06-25
Updates perl to use git-checkout

dakaneye:
Everything else is a epoch bump and an execution of yam for formatting, to make sure I didn't break perl packages

Notes
* perl-template-toolkit: failing update check because it says there's a newer version, but that newer version is not available at the fetch URL for download...So, I've updated it to pull from the official github repo instead of a cpan tar archive fetch download


smoser (approved):
thanks @dakaneye

PR #22732, opened 2024-06-26
pytorch: withdraw package

pnasrat:
Withdraw pytorch package from wolfi

current package pulls in non-free nvidia deps from pypi


smoser:
@pnasrat , are you going to replace the existing pytorch(.yaml) with something else? or drop it?

pnasrat:
@smoser - good point I've git rm'd the yaml

smoser (approved):

PR #22743, opened 2024-06-26
teleport: use rustup stable triplet

pnasrat:
The melange triple does not match rustup stable channel so the path is incorrect

Discovered in issue with wasm rustup packages in #22375

### DIFF
--- a/teleport.yaml
+++ b/teleport.yaml
@@ -1,7 +1,7 @@
 package:
   name: teleport
   version: 16.0.1
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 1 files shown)


jonjohnsonjr:
This was me breaking more things (sorry), fixed in https://github.com/chainguard-dev/melange/pull/1331

Should be fixed now (so closing this PR), but FWIW `${{cross.triplet.rust.gnu}}` is probably what we'd want here if we do end up droping the `${{host.*}}` flavored things (https://github.com/chainguard-dev/melange/issues/1056 for context as to why).

PR #22754, opened 2024-06-26
Move conntrack-tools to git-checkout

krishjainx:
cc/ @smoser


smoser (changes requested):
need to bump the epoch (`wolfictl bump epoch conntract-tools.yaml`).
need to run `yam`

i hit 'rerun' on the check-update task, hopefully that goes this time. but maybe the netfilter git repo is not great.

smoser (approved):

PR #22821, opened 2024-06-27
fix(sonarqube): mitigate CVEs

stormqueen1990:
Mitigate a few CVEs present in the SonarQube package.


stormqueen1990:
This change cannot be applied.

PR #22837, opened 2024-06-28
openssl: add patches for CVE-2024-5535

ajayk:
https://www.openssl.org/news/secadv/20240627.txt

Due to the low severity of this issue we are not issuing new releases of
OpenSSL at this time. The fix will be included in the next releases when they
become available. The fix is also available in commit e86ac436f0 (for 3.3),
commit 99fb785a5f (for 3.2), commit 4ada436a19 (for 3.1) and commit cf6f91f612
(for 3.0) in the OpenSSL git repository. It is available to premium support
customers in commit b78ec0824d (for 1.1.1) and commit 99472514130 for (1.0.2).

### DIFF
--- a/openssl.yaml
+++ b/openssl.yaml
@@ -1,7 +1,7 @@
 package:
   name: openssl
   version: 3.3.1
-  epoch: 2
+  epoch: 3
   description: "the OpenSSL cryptography suite"
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


smoser (approved):

xnox:
Pondering if this should be included in SBOM as an external ref, or not.

PR #22840, opened 2024-06-28
fix: jenkins cve

hectorj2f:
This PR attempts to fix a CVE in jenkins with id: GHSA-xfx3-cr74-x3cv,

### DIFF
--- a/jenkins.yaml
+++ b/jenkins.yaml
@@ -1,7 +1,7 @@
 package:
   name: jenkins
   version: "2.464"
-  epoch: 1
+  epoch: 2
   description: Open-source CI/CD application.
   copyright:
     - license: MIT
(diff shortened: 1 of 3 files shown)


jenkins.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: jenkins
-  version: "2.463"
```
hectorj2f, line 3: I don't understand this https://github.com/wolfi-dev/os/blob/main/jenkins.yaml#L3. Jenkins is already at `2.464` 👻

cpanato: conflict

hectorj2f: nvm

cpanato (approved):

PR #22846, opened 2024-06-28
py3-build dependencies versioned python support

pnasrat:
Adds versioned python packages for py3-build and deps


py3-packaging.yaml
```diff
@@ -1,23 +1,31 @@
...
-  epoch: 0
+  epoch: 1
   description: "core utilities for python3 packaging"
   copyright:
     - license: Apache-2.0 AND BSD-2-Clause
   dependencies:
-    runtime:
-      - py3-parsing
-      - python3
+    provider-priority: 0
```
dakaneye, line 9: What does this mean? In Melange it says it will determines the priority of provides against other equal package provides

py3-packaging.yaml
```diff
@@ -27,11 +35,36 @@ pipeline:
...
 
 update:
   enabled: true
+  ignore-regex-patterns:
+    - a
+    - b
+    - rc
+    - post
   github:
     identifier: pypa/packaging
```
dakaneye, line 70: should we add a test pipeline for this too?

py3-pyproject-metadata.yaml
```diff
@@ -1,40 +1,76 @@
...
-      - py3-setuptools
-      - py3-wheel
-      - python-3
+      - py3-supported-pip
+      - py3-supported-python
       - wolfi-base
 
 pipeline:
-  - uses: fetch
+  - uses: git-checkout
```
dakaneye, line 33: +1

dakaneye (approved):
lgtm, all my comments are minor.

py3-packaging.yaml
pnasrat: This is part of the multi versioned python support, which allows each versioned py3.MN-foo to provide py3-foo and default to the highest priority 3.12 - this probably needs documentation.

py3-packaging.yaml
pnasrat: once melange supports versioned python I'll fix the tests so it runs on all versions

PR #22878, opened 2024-06-28
`vscode-skip-server-requirements-check` create tmp

found-it:
The file this package touched wasn't extracted into the final apk


rawlingsj:
looks like this change has already been added https://github.com/wolfi-dev/os/pull/22928, going to close this PR

PR #22899, opened 2024-06-30
kyverno: version stream

Dentrax:
VersionStream PR: https://github.com/chainguard-dev/package-version-metadata/pull/573

### DIFF
--- a/kyverno-1.12.yaml
+++ b/kyverno-1.12.yaml
@@ -1,5 +1,5 @@
 package:
-  name: kyverno
+  name: kyverno-1.12
   version: 1.12.4
   epoch: 2
   description: Kubernetes Native Policy Management
@@ -8,6 +8,8 @@ package:
   dependencies:
     runtime:
       - ca-certificates-bundle
+    provides:
+      - kyverno=${{package.full-version}}
 
 environment:
   contents:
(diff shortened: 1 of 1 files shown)


kyverno-1.12.yaml
```diff
@@ -1,13 +1,15 @@
 package:
-  name: kyverno
+  name: kyverno-1.12
```
hectorj2f, line 2: Could we check if we need to update any images using this package ?

hectorj2f: Also we might want to copy the advisory of this new package rename too.

Dentrax: > Could we check if we need to update any images using this package ?

I thought passing `provides:` would handle that:
```
provides:
       - kyverno=${{package.full-version}}
```

🤔

hectorj2f: ops! yeah! never mind. But we need to copy the advisory.

Dentrax: I may a bit confused here, should I `cp kyverno.advisories.yaml kyverno-1.12.advisories.yaml` (or `mv`?) and rename `package.name` to `kyverno-1.12`?

Dentrax: https://github.com/wolfi-dev/advisories/pull/6585

imjasonh (dismissed):

hectorj2f (approved):

PR #22922, opened 2024-07-01
glibc: enable hardening options

xnox:
Enable multiple hardening options at configure time, to be sure (some of these are toolchain built-in defaults anyway).
Embed the full package version in the build.

With these changes, `hardening-check` reports that "Control flow integrity" is now enabled on `libc.so.6` like it is on other distributions.

However, "Stack clash protection" is still reported as unknown/no, instead of "yes" like it is on other distributions. Will investigate that separately.

### DIFF
--- a/glibc.yaml
+++ b/glibc.yaml
@@ -1,7 +1,7 @@
 package:
   name: glibc
   version: 2.39
-  epoch: 6
+  epoch: 7
   description: "the GNU C library"
   copyright:
     - license: LGPL-2.1-or-later
@@ -47,6 +47,9 @@ environment:
       - texinfo
       - wolfi-baselayout
       - zlib
+  # glibc manages FORTIFY_SOURCE on per-file basis
+  environment:
+    CPPFLAGS: ""
 
 pipeline:
   - uses: fetch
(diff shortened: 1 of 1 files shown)


xnox:
> Would it be worth it to add a test that runs hardening-check to the package?

Yes, especially once we have a few more of the hardening things enabled in the distro. Probably will want to package that script as stand alone; as it is written in perl; and is part of a kitchen sink collection of things - which might have too many dependencies if attempted to be packaged in full. And most other scripts are too Debian specific.

xnox:
hardening-check package:
- https://github.com/wolfi-dev/os/pull/23321

dakaneye (approved):
LGTM. Did some research on potentially other hardening options and was curious if we should also add the following:
* --enable-static-pie: Enables position-independent executables (PIE) for static linking, improving address space layout randomization (not sure if this might affect how other packages link though)

xnox:
> LGTM. Did some research on potentially other hardening options and was curious if we should also add the following:
> 
> * --enable-static-pie: Enables position-independent executables (PIE) for static linking, improving address space layout randomization (not sure if this might affect how other packages link though)

we could enable that. Note that it is sort of odd one - for statically linked binaries that gains one nothing, but it does gain one something for "mixed" binaries - i.e. shared linked to something else, but statically linking in glibc for some reason. As that's when this will matter (the overall binary will not be pie unless we have static-pie). So yeah, can add this too.

PR #22933, opened 2024-07-01
use var transform to enable automatically updates for openssh

cpanato:
- use var transform to enable automatically updates for openssh

### DIFF
--- a/openssh.yaml
+++ b/openssh.yaml
@@ -23,10 +23,16 @@ environment:
       - wolfi-baselayout
       - zlib-dev
 
+var-transforms:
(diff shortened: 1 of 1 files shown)


vaikas (approved):

xnox (approved):

mattmoor:
Should we have bumped the epoch to test this?

xnox:
> Should we have bumped the epoch to test this?

presubmit builds are based on git changes, https://github.com/wolfi-dev/os/actions/runs/9745234844/job/26892728322?pr=22933#step:8:190 and this was built so it does build.

PR #22942, opened 2024-07-01
extract the sshd-session binary in the server subpackage

found-it:
The `sshd` binary in the `openssh-server` package was broken because it was missing the `sshd-session` binary. This PR just extracts the `sshd-session binary as part of the `openssh-server` subpackage so it is individually installable.

Release Notes: https://www.openssh.com/releasenotes.html
```
* [sshd(8)](https://man.openbsd.org/sshd.8): the server has been split into a listener binary, [sshd(8)](https://man.openbsd.org/sshd.8),
   and a per-session binary "sshd-session". This allows for a much
   smaller listener binary, as it no longer needs to support the SSH
   protocol. As part of this work, support for disabling privilege
   separation (which previously required code changes to disable) and
   disabling re-execution of [sshd(8)](https://man.openbsd.org/sshd.8) has been removed. Further
   separation of sshd-session into additional, minimal binaries is
   planned for the future.
```

### DIFF
--- a/openssh.yaml
+++ b/openssh.yaml
@@ -1,7 +1,7 @@
 package:
   name: openssh
   version: 9.8_p1
-  epoch: 0
+  epoch: 1
   description: "the OpenBSD SSH implementation"
   copyright:
     - license: ISC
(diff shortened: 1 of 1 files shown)


k4leung4 (commented):
should we add a test for this regression?

mattmoor (approved):
We should figure out how we get test coverage, but approving to unbreak things.

hectorj2f (approved):

xnox (approved):

PR #22944, opened 2024-07-01
py3-build versioned python

pnasrat:
Move py3-build to use versioned python

### DIFF
--- a/py3-build.yaml
+++ b/py3-build.yaml
@@ -1,27 +1,30 @@
 package:
   name: py3-build
   version: 1.2.1
-  epoch: 1
+  epoch: 2
   description: A simple, correct Python build frontend
   copyright:
     - license: MIT
   dependencies:
-    runtime:
-      - py3-packaging
-      - py3-pyproject-hooks
-      - python3
+    provider-priority: 0
+
+vars:
+  pypi-package: build
+
+data:
+  - name: py-versions
+    items:
+      3.10: "310"
+      3.11: "311"
+      3.12: "312"
 
 environment:
   contents:
     packages:
       - busybox
       - ca-certificates-bundle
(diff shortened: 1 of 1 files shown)


py3-build.yaml
```diff
@@ -31,31 +34,53 @@ pipeline:
...
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/usr
+          mv ${{targets.contextdir}}/../py${{range.key}}-${{vars.pypi-package}}/usr/bin ${{targets.contextdir}}/usr
 
-  - uses: strip
+  - name: py3-supported-${{vars.pypi-package}}
+    description: meta package providing ${{vars.pypi-package}} for supported python versions.
+    dependencies:
+      runtime:
+        - py3.10-${{vars.pypi-package}}
```
vaikas, line 71: Should this be depending on the -bin version which pulls in the binaries, and then transivitely the libs?

vaikas: Asking mainly because I'm trying to document the new way of doing things 🤣

py3-build.yaml
```diff
@@ -31,31 +34,53 @@ pipeline:
...
+        - py3.11-${{vars.pypi-package}}
+        - py3.12-${{vars.pypi-package}}
+
+test:
+  pipeline:
+    - runs: python3.12 -c "import ${{vars.pypi-package}}"
```
vaikas, line 77: ❤️

py3-build.yaml
pnasrat: Supported is really for us to build packages and we can just use  `python -m build` vs the binary build (which I don't think is heavily used)

vaikas (approved):

PR #22951, opened 2024-07-01
Add openssh-server test

found-it:
This is a follow-up to https://github.com/wolfi-dev/os/pull/22942

The old test _looked_ like it was testing `openssh-server` and it technically was. The test pipeline was tied to the `openssh` package though which means the `openssh` package was implicitly installed in the environment so the `sshd-session` binary was picked up. I took the same test and added it under the `openssh-server` subpackage so it gets tested without the extra packages.

I confirmed this test catches the breakage for the 9.8 r0 package

### DIFF
--- a/openssh.yaml
+++ b/openssh.yaml
@@ -138,6 +138,11 @@ subpackages:
       post-install: |
         #!/bin/sh
         ssh-keygen -A -q -N ""
+    test:
+      pipeline:
+        - runs: |
+            ssh-keygen -A -q -N ""    # This is unfortunate, but scriplets don't run in test environments, so force this here
+            /usr/sbin/sshd
 
   - name: "openssh-server-config"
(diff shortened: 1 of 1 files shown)


vaikas (approved):

hectorj2f (approved):

xnox (changes requested):
The test is now executed twice. And once with incorrect dependencies.

If we want to move sshd test to openssh-server subpackage test, i would expect to remove the top level test.

found-it:
@xnox I don't mind if it's executed twice. The top level `openssh` package includes `openssh-server` so keeping a test at the top level acts a sanity check that the `openssh` package also contains a working `sshd`. The cost of the test is cheap so I thought it more valuable to leave it

xnox (approved):

PR #23078, opened 2024-07-03
grpc-health-probe: tag 0.4.29 was withdrawn from upstream

ajayk:
This was reverted here https://github.com/wolfi-dev/os/commit/3c5063d18ce8b1647eb4c89b62716298de1e2a74


grpc-health-probe.yaml
```diff
@@ -1,6 +1,7 @@
 package:
   name: grpc-health-probe
   version: 0.4.28
+  # bump to epoch 1 when 0.4.29 is released
```
cpanato, line 4: why that?

cpanato (approved):

grpc-health-probe.yaml
ajayk: incase when somebody runs a withdraw action again this version will be removed again

PR #23187, opened 2024-07-04
calico: fix FTBFS and rebuild with new golang

xnox:
Recently melange has updated how it calculates required steps, and it
seems that number of required pipelines to execute has always been two
yet incorrectly asserted as one. Update the assertion to the new
currently expected value.

Also this rebuilds calico with new golang for CVE remediation.


cpanato (approved):
thank you!

Issue #23202, opened 2024-07-05
[Wolfi Package Request]: async-profiler (Java)


lyoung-confluent:
### What software would you like us to add to wolfi-os. Ideally include a URL to the project and its source.

https://github.com/async-profiler/async-profiler

### which versions of the software should we include?

v3.0 (latest)

### Add some justification for why this specific package and versions are important.

async-profiler is a popular (~7.3k stars) tool for profiling Java processes at runtime. While it is used to profile Java applications, the tool itself is written primarily in CPP, it would be convenient to be able to install the tool via `apk add` to profile Java applications running on Wolfi

### Please check all that apply

- [X] This package has an un-restrictive license
- [X] The package/versions proposed are actively maintained upstream
- [X] I am interested in adding this package to Wolfi OS myself
- [ ] I am willing to help maintain this package

ajayk:
thanks @lyoung-confluent 🎉  , this has been closed by #23203

PR #23203, opened 2024-07-05
async-profiler: new package

lyoung-confluent:
Fixes:
- #23202

### Pre-review Checklist

### DIFF
--- a/async-profiler.yaml
+++ b/async-profiler.yaml
@@ -1,0 +1,44 @@
+package:
+  name: async-profiler
+  version: 3.0
+  epoch: 0
+  description: A low overhead sampling profiler for Java
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - bash
+      - build-base
+      - openjdk-11
+      - openjdk-11-default-jvm
+      - wolfi-base
+  environment:
+    JAVA_HOME: /usr/lib/jvm/java-11-openjdk
+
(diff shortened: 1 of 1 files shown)


async-profiler.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/async-profiler/async-profiler
+      tag: v${{package.version}}
+      expected-commit: 4e441b4024a5873a5764a1e8b9f0bb25ad997fbf
+
+  - runs: |
+      make all
```
lyoung-confluent, line 28: It looks like the upstream `Makefile` does _not_ use `CFLAGS` and `CXXFLAGS` from the environment which is not ideal, unsure if we should patch that out: https://github.com/async-profiler/async-profiler/blob/4e441b4024a5873a5764a1e8b9f0bb25ad997fbf/Makefile#L11-L12

async-profiler.yaml
```diff
@@ -0,0 +1,44 @@
...
+test:
+  pipeline:
+    - runs: |
+        /usr/bin/asprof --help
+
+update:
+  enabled: true
+  github:
+    identifier: async-profiler/async-profiler
+    strip-suffix: v
```
ajayk, line 44:
```suggestion
    strip-prefix: v
```

ajayk (approved):

PR #23220, opened 2024-07-05
add package: fluent-bit-docker-image

ritster:
New package: fluent-bit-docker-image

Attempting to migrate https://github.com/chainguard-dev/enterprise-packages/blob/main/aws-for-fluent-bit-1.9.yaml#L58,  https://github.com/chainguard-dev/enterprise-packages/blob/main/aws-for-fluent-bit-fips.yaml#L58, and https://github.com/wolfi-dev/os/blob/a887e75fbd958082ae9075a662a478ed8cef87b8/aws-for-fluent-bit.yaml#L58 away from using those second git-checkouts steps, which will enable auto-updates to work properly.

Though this repo is not being maintained upstream, there is no code (so security updates shouldn't be an issue) and this is needed for a wolfi-os package, so being placed here.

### DIFF
--- a/fluent-bit-docker-image.yaml
+++ b/fluent-bit-docker-image.yaml
@@ -1,0 +1,28 @@
+package:
+  name: fluent-bit-docker-image
+  version: 20220322
+  epoch: 0
+  description: Docker image for Fluent Bit
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - busybox
+
+vars:
+  FLB_DOCKER_BRANCH: "1.8"
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/fluent/fluent-bit-docker-image
+      destination: ${{targets.destdir}}/fluent-bit-docker-image
+      branch: ${{vars.FLB_DOCKER_BRANCH}}
+      expected-commit: 98ce3316ea93751ddd33bc319a8f9d177493155b
+
+  - runs: rm -rf ${{targets.destdir}}/fluent-bit-docker-image/.git ${{targets.destdir}}/fluent-bit-docker-image/.github
+
+update:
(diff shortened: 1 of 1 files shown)


ritster:
tag is missing as upstream uses branch HEAD, which is unchanging as the repository is archived, but also not a tagged version.

fluent-bit-docker-image.yaml
```diff
@@ -0,0 +1,28 @@
...
+    packages:
+      - busybox
+
+vars:
+  FLB_DOCKER_BRANCH: "1.8"
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/fluent/fluent-bit-docker-image
```
hectorj2f, line 20: what do we get when we install this package ? I don't understand the purpose of this package more than just adding a git repo.

ritster: Yep it's strange for sure. If you look at https://github.com/wolfi-dev/os/blob/main/aws-for-fluent-bit.yaml#L58, I think this repo is just used for copying some configuration files to the working directory.

The issue with the current setup is that the auto-updates don't quite work, as the sha of the second git-checkout step in https://github.com/wolfi-dev/os/blob/main/aws-for-fluent-bit.yaml (and for two other packages in enterprise-packages) is overwritten on every package update. @ajayk suggested we decouple the two repos and make this one its own package, however minimal it is.

hectorj2f: Alright! I wasn't sure this was a good practice. That is why I asked about it.

fluent-bit-docker-image.yaml
```diff
@@ -0,0 +1,28 @@
+package:
```
ajayk, line 1:
```suggestion
#nolint:valid-pipeline-git-checkout-tag
package:
```

ajayk (approved):

PR #23225, opened 2024-07-05
fix: cloudnative-pg use build.goarch

imjasonh:
This didn't work the way it was intended before, and newer versions of `melange` detect and fail in this case.

As it was, only the first one was used, so it worked as expected on `x86_64`, producing `manager_amd64`. `aarch64` was unaffected, and produced `manager_aarch64`. Because the var followed that behavior, the test didn't detect it.

It turns out we have `${{build.goarch}}` for this, which we can use directly instead of mangling vars.

### DIFF
--- a/cloudnative-pg.yaml
+++ b/cloudnative-pg.yaml
@@ -56,6 +56,6 @@ test:
     - name: "Verify Installation"
       runs: |
         cd /
-        manager_${{vars.build.goarch}} version | grep "${{package.version}}"
+        manager_${{build.goarch}} version | grep "${{package.version}}"
         /manager version | grep "${{package.version}}"
         /manager debug show-architectures | grep ${{build.goarch}}
--- a/cloudnative-pg.yaml
+++ b/cloudnative-pg.yaml
@@ -1,21 +1,10 @@
 package:
   name: cloudnative-pg
   version: 1.23.2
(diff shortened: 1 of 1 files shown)


hectorj2f (dismissed):

cpanato (dismissed):

cloudnative-pg.yaml
```diff
@@ -31,15 +20,15 @@ pipeline:
   - uses: go/build
     with:
       modroot: ./cmd/manager
-      output: manager_${{vars.mangled-arch}}
+      output: manager_${{build.goarch}}
```
hectorj2f, line 23: @imjasonh But this won't made the replacement from 'aarch64' to 'arm64', afaik.

imjasonh: I believe it will -- we can see it's named correctly in the package:

```
$ make package/cloudnative-pg
$ tar -tvf packages/aarch64/cloudnative-pg-1.23.2-r6.apk
...
-rwxr-xr-x  0 jason  dialout 58386910 Jul  5 10:44 bin/manager_arm64
lrwxr-xr-x  0 jason  dialout        0 Jul  5 10:44 manager -> /bin/manager_arm64
```

hectorj2f (approved):

PR #23227, opened 2024-07-05
mockgen: new package

macabu:
Adds https://github.com/uber-go/mock as a new package: `mockgen`

Not sure if I got all the requirements right, let me know 🙂 

#### For new package PRs only


rawlingsj:
Thanks @macabu.  The PR mostly looks great, just one small thing which we probably haven't documented or made obvious, we are moving away from `go/install` pipelines and favouring `git-checkout` + `go/build` e.g. https://github.com/wolfi-dev/os/blob/29dabcaa9232d7814e05f5e311ca6562ab42efda/tekton-chains.yaml#L24-L27

This makes it easier for us to automate the bumping (and removal) of any go dependencies that may contain vulnerabilities

macabu:
Thanks for the feedback @rawlingsj , I've followed the template you linked, and also removed the `environment.packages` since I've noticed the `pipelines` can require packages that get automatically pulled in (pretty cool!)

rawlingsj (approved):

SamyDgz (approved):

PR #23244, opened 2024-07-05
Add symlink for entrypoint binary, for compat with upstream image and charts

mamccorm:
Add missing symlinks for compatibility with upstream image, as well as any charts depending on these files being present.

### DIFF
--- a/multus-cni.yaml
+++ b/multus-cni.yaml
@@ -85,3 +85,5 @@ test:
       runs: |
         stat /usr/src/multus-cni/bin/multus-shim
         stat /usr/src/multus-cni/bin/multus-daemon
+        /usr/bin/thin_entrypoint --help
+        /thin_entrypoint --help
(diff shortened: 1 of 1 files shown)


multus-cni.yaml
```diff
@@ -61,8 +61,10 @@ subpackages:
     pipeline:
       - runs: |
           mkdir -p ${{targets.contextdir}}/usr/src/multus-cni/bin
+          ln -sf /usr/bin/multus ${{targets.contextdir}}/usr/src/multus-cni/bin/multus
```
dakaneye, line 64: I think we need an install_multus binary too for the whereabouts chart. https://raw.githubusercontent.com/k8snetworkplumbingwg/multus-cni/master/deployments/multus-daemonset.yml

developer-guy: done.

imjasonh (approved):

PR #23316, opened 2024-07-08
gcc: upgrade current gcc streams to latest point release

xnox:
Use sha512 checksums as published by upstream. Use smaller xz tarball
in both streams. Drop CVE patch, applied in latest point releases.


pnasrat (approved):
LGTM

PR #23321, opened 2024-07-08
New package hardening-check

xnox:
A useful script to determine hardening features enabled in a given ELF
binary by parsing and creating summary out of readelf/objdump output.

This is packaged and widely used in Debian, Ubuntu, SUSE, Fedora.

### DIFF
--- a/hardening-check.yaml
+++ b/hardening-check.yaml
@@ -1,0 +1,31 @@
+package:
+  name: hardening-check
+  version: 2.23.7
+  epoch: 0
+  description: "Debian devscripts hardening-check"
+  copyright:
+    - license: GPL-2.0-or-later
+  dependencies:
+    runtime:
+      - binutils
+
+environment:
+  contents:
+    packages:
+      - busybox
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://salsa.debian.org/debian/devscripts.git
+      tag: v${{package.version}}
+      expected-commit: eb7dc80e72486f7955ec6793a8067fe69fda7883
+
+  - runs: |
+      mkdir -p "${{targets.destdir}}"/usr/bin
+      mv scripts/hardening-check.pl "${{targets.destdir}}"/usr/bin/hardening-check
+
+update:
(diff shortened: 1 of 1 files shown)


hardening-check.yaml
```diff
@@ -0,0 +1,31 @@
...
+      expected-commit: eb7dc80e72486f7955ec6793a8067fe69fda7883
+
+  - runs: |
+      mkdir -p "${{targets.destdir}}"/usr/bin
+      mv scripts/hardening-check.pl "${{targets.destdir}}"/usr/bin/hardening-check
+
+update:
+  # Last change 2 years ago. It only makes sense to update when there
+  # is a manjor toolchain evolution.
+  enabled: false
```
rawlingsj, line 31: even if it's not updated upstream very often it's still probably worth tracking the version. It's on release monitoring https://release-monitoring.org/project/5419/

hardening-check.yaml
```diff
@@ -0,0 +1,31 @@
+package:
+  name: hardening-check
```
rawlingsj, line 2: is the name specific enough?  i.e the repo is called devscripts, so wonder if `hardening-check-devscripts` or something else would be better?

xnox: Other distros package this script stand alone in that package name. I.e. https://packagehub.suse.com/packages/hardening-check/ https://packages.fedoraproject.org/pkgs/hardening-wrapper/hardening-check/

Majority of other devscripts, are highly debian/ubuntu/.deb specific and have additional runtime dependencies on debian parsing libraries. Thus even if there is ever a need to package debian devscripts, this script likely will remain as a standalone package (or subpackage).

pnasrat (commented):
LGTM

As @rawlingsj  is reviewing I'll leave it to him to approve but the package name makes sense to me per @xnox's comment

hardening-check.yaml
rawlingsj: got it - thanks

rawlingsj (approved):

PR #23338, opened 2024-07-08
kubeflow-katib: fix GHSA-248v-346w-9cwc

rawlingsj:


kubeflow-katib.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: kubeflow-katib
-  epoch: 15
+  epoch: 16
```
hectorj2f, line 3: Why do we need to bump the epoch here? It has been done in a previous PR here: https://github.com/wolfi-dev/os/commit/712af6897dfbbe2bf9a0069b5fe9c70be60e96ce ?

rawlingsj: I was working off the list of issues to triage which was still open, so just hadn't noticed you'd already bumped it.  I'll close this and update the linked issue so nobody else thinks it's needed

PR #23346, opened 2024-07-08
hunspell: add missing dictionaries

89luca89:
Related: https://github.com/chainguard-dev/image-requests/issues/3637

Fixing subpackages creation for dictinaries, needed for spellchecking to actually work

### DIFF
--- a/hunspell-dict.yaml
+++ b/hunspell-dict.yaml
@@ -1,0 +1,163 @@
+package:
+  name: hunspell-dictionaries-all
+  version: 0.1_git20240709
+  epoch: 0
+  description: Dictionaries for spell checker and morphological analyzer library and program
+  copyright:
+    - license: GPL-2.0-or-later OR LGPL-2.1-or-later OR MPL-1.1
+  dependencies:
+    runtime:
+      - hunspell-dictionary-af
+      - hunspell-dictionary-an
+      - hunspell-dictionary-ar
+      - hunspell-dictionary-bg
+      - hunspell-dictionary-bn
+      - hunspell-dictionary-bo
+      - hunspell-dictionary-br
+      - hunspell-dictionary-bs
+      - hunspell-dictionary-ca
+      - hunspell-dictionary-ckb
+      - hunspell-dictionary-cs
+      - hunspell-dictionary-da
+      - hunspell-dictionary-de
+      - hunspell-dictionary-el
+      - hunspell-dictionary-en
+      - hunspell-dictionary-eo
+      - hunspell-dictionary-es
+      - hunspell-dictionary-et
+      - hunspell-dictionary-fr
+      - hunspell-dictionary-gd
+      - hunspell-dictionary-gl
+      - hunspell-dictionary-gu
+      - hunspell-dictionary-gug
+      - hunspell-dictionary-he
+      - hunspell-dictionary-hi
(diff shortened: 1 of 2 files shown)


hunspell.yaml
```diff
@@ -32,6 +33,26 @@ pipeline:
 
   - uses: strip
 
+  - runs: |
+      mkdir -p ${{targets.destdir}}/usr/share/hunspell
+      cd ${{targets.destdir}}/usr/share/hunspell
+      wget -c https://downloads.sourceforge.net/project/wordlist/speller/2020.12.07/hunspell-en_US-2020.12.07.zip
+      wget -c https://downloads.sourceforge.net/project/wordlist/speller/2020.12.07/hunspell-en_CA-2020.12.07.zip
+      wget -c https://downloads.sourceforge.net/project/wordlist/speller/2020.12.07/hunspell-en_AU-2020.12.07.zip
+      wget -c https://downloads.sourceforge.net/project/wordlist/speller/2020.12.07/hunspell-en_GB-large-2020.12.07.zip
```
rawlingsj, line 42: I think we'd want to find a better way to fetch these, are these available via a git checkout?  We'd maybe want to split these into subpackages using a data / range feature of melange similar to https://github.com/wolfi-dev/os/blob/120fd8a3abda7a5c729907410d683481505cf2d3/glibc.yaml#L126-L131 and https://github.com/wolfi-dev/os/blob/120fd8a3abda7a5c729907410d683481505cf2d3/glibc.yaml#L383-L389

also this change on the package would require an epoch bump for new apks to include the changes

89luca89: Thanks for the feedback @rawlingsj  :)
Found a better source of dicts from libreoffice, so I'm now using git for that

Split package is now done with data/range over `_dictionaries` :+1:

hunspell.yaml
```diff
@@ -49,11 +49,6 @@ subpackages:
       - uses: split/manpages
     description: hunspell manpages
 
-  - name: hunspell-lang
```
dlorenc, line 52: Should we withdraw this package too?

hunspell-dict.yaml
```diff
@@ -0,0 +1,162 @@
...
+      sw:
+      te:
+      th:
+      tr:
+      uk:
+      vi:
+
+subpackages:
+  - range: _dictionaries
+    name: hunspell-dictionary-${{range.key}}
```
dlorenc, line 154: Just curious - some other distros seem to call these packages "hunspell-$lang" without the -dictionary- part, did you copy that convention from somewhere?

dlorenc:
Looks good, just two comments.

hunspell.yaml
89luca89: It was empty anyway, hunspell doesn't provide locales

hunspell-dict.yaml
89luca89: usually `-lang` is for locales (like `glibc-lang-foo`) these are dictionaries, so I didn't want to make it confusing

on alpine they just use `hunspell-en-...` without `-dictionaries` but I thought having a separate package like this enables us also to have an `-all` package that depends on the subpackages

This enables apk to properly handle adding `-all` after a subpackage without overwriting files

dlorenc (approved):

PR #23348, opened 2024-07-08
upgrade node 16 dependents to latest node

imjasonh:
Checking to see if these build.

Node 16 has been EOL for 10 months. These are the only packages dependent on it in Wolfi.

Node 18 is in security-only support for another 9 months, at which point these should upgrade to the next Node.

### DIFF
--- a/neuvector-manager.yaml
+++ b/neuvector-manager.yaml
@@ -1,7 +1,7 @@
 package:
   name: neuvector-manager
   version: 5.3.3
-  epoch: 1
+  epoch: 2
   description: NeuVector Security Center Admin Console.
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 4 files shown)


rawlingsj (dismissed):

py3-jupyterhub.yaml
```diff
@@ -35,7 +35,7 @@ environment:
       - build-base
       - busybox
       - ca-certificates-bundle
-      - nodejs-16
+      - nodejs-18
```
xnox, line 38: Note this is basically a build-time only dependency of npm. Because we only call "npm" and it needs "a nodejs" but doesn't care which one. So I think this can be actually updated to just generic `nodejs`.

neuvector-manager.yaml
```diff
@@ -21,7 +21,7 @@ environment:
     packages:
       - build-base
       - busybox
-      - nodejs-16
+      - nodejs-18
       - npm
```
xnox, line 25: Same here.

xnox (approved):

PR #23350, opened 2024-07-08
py3.NN-setuptools: depend on python-3.NN

imjasonh:


imjasonh:
Sounds like this isn't the fix, since it means py3.12-setuptools couldn't be installed alongside py3.11-setuptools.

PR #23371, opened 2024-07-08
Make ruby version streams provide ruby

jonjohnsonjr:
This gives us a "ruby" virtual that allows wolfictl to properly construct a graph and make ruby-3.2 and ruby-3.3 conflict with each other more explicitly to help the apko solver when it struggles.

### DIFF
--- a/ruby-3.0.yaml
+++ b/ruby-3.0.yaml
@@ -1,11 +1,14 @@
 package:
   name: ruby-3.0
   version: 3.0.7
-  epoch: 1
+  epoch: 2
   description: "the Ruby programming language"
   copyright:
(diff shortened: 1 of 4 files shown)


rawlingsj:
I know this is still in draft, but just in case I noticed there's not an epoch bump on these packages

dakaneye (dismissed):
👍 LGTM

smoser:
This seems "right" :+1: .
I'm not using the review due to needing epoch bump as James said.

jonjohnsonjr:
Bumped epochs and un-drafted.

smoser (approved):

PR #23381, opened 2024-07-08
Fix/python missing deps

smoser:
this fixes some packages that used /usr/bin/python3 in their build, but did not directly depend on anything that would provide /usr/bin/python3.

Some changes to the python base packages meant that they would fail to build.


smoser:
> [drop py3-urllib3-1.yaml - this is not used anywhere](https://github.com/wolfi-dev/os/pull/23381/commits/65de202ebbba63dd8a6fe9d14bc60e844b314421)

I'm sort of sure that statement is true.

rawlingsj (changes requested):
A few of the packages changed in this PR are missing an epoch bump

smoser:
This is replaced by much less invasive https://github.com/wolfi-dev/os/pull/23627 .

rawlingsj:
> This is replaced by much less invasive https://github.com/wolfi-dev/os/pull/23627 .

sounds good, going to close this for now then, we can reopen if still needed

Issue #23352, opened 2024-07-08
FTBFS: tigerbeetle


imjasonh:
```
/home/build/build.zig:1014:22: error: expected 2 arguments, found 3
/home/build/build.zig:1039:22: error: expected 2 arguments, found 3
/home/build/build.zig:1091:22: error: expected 2 arguments, found 3
```

smoser:
fixed [#23605](https://github.com/wolfi-dev/os/pull/23605)

ritster:
Issue wasn't fixed in previous PR (not actually merged) - now fixed in https://github.com/wolfi-dev/os/pull/25089. The upstream build process changed, now our package config reflects that.

PR #23605, opened 2024-07-09
remove some broken packages: buck2, tigerbeetle

### LINKED ISSUE #23351: FTBFS: buck2
```
error: package `triomphe v0.1.13` cannot be built because it requires rustc 1.76 or newer, while the currently active rustc version is 1.74.0-nightly
Either upgrade to rustc 1.76 or newer, or use
cargo update triomphe@0.1.13 --precise ver
where `ver` is the latest version of `triomphe` supporting rustc 1.74.0-nightly
```

imjasonh:
Fixes #23351 #23352 

These packages don't currently build.

- buck2 was last updated 6 months ago, and is only built so it can be included in a public developer tier image. After this change we can withdraw that image.
- tigerbeetle was last updated 4 months ago, but isn't used as a dependency of any other package, and isn't included in an image.


dakaneye:
In general would it make more sense to withdraw the image for `buck2` before removing the package? Or does it not matter at all since we're just gonna remove both anyways.

imjasonh:
Closing while we decide.

PR #23633, opened 2024-07-09
fix(metric-collector-for-apache-cassandra): Stream for Cassandra 4.1

EyeCantCU:
By default, the metric collector is built against a significantly older version of Cassandra for backwards compatibility, in turn making it incompatible with Cassandra 4.1

This streams the metric collector and bumps the Cassandra version to 4.1


mamccorm (changes requested):
Looks great - we do have some dependent images using this package name and expecting v4.0 compat, so we'll need to merge the v4.0 one first, make the image updates, then merge this one.

Marking as request changes, in case someone else accidently merges before we get to do this in the AM

tstromberg (approved):

mamccorm (approved):

PR #23709, opened 2024-07-10
gcc: stop building go and objc

xnox:
gccgo only supports EOL golang 1.18 without generics, it is thus not
useful to build up to date golang code. Unused by any of the packages
in Wolfi and not shipped in any Chainguard Images. It is holding up
shipping EOL go-1.18 in Wolfi.

objc compiler typically requires gnustep libraries to be usable, which
Wolfi does not package. Unused by any of the packages in Wolfi and not
shipped in any Chainguard Images.

These compilers can be reintroduced if popular software starts to
require these toolchains, or if there is demand to ship them as
Chainguard Images. In the meantime, building these toolchains is a
waste of resources.

### DIFF
--- a/gcc.yaml
+++ b/gcc.yaml
@@ -1,7 +1,7 @@
 package:
   name: gcc
   version: 13.3.0
-  epoch: 0
+  epoch: 1
   description: "the GNU compiler collection"
   copyright:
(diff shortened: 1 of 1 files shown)


xnox:
and we use this for bootstrap, so hey ho.

PR #23719, opened 2024-07-10
gcc-12 explicitly add deps for libexec binaries

pnasrat:
tested by apk add gcc-12 and ldd

Fix compiler issues after 12.4.0 release 

```
/usr/libexec/gcc/x86_64-pc-linux-gnu/12.4.0 # ldd cc1
        linux-vdso.so.1 (0x00007ffcca3f5000)
        libisl.so.23 => not found
        libmpc.so.3 => not found
        libmpfr.so.6 => not found
        libgmp.so.10 => not found
        libz.so.1 => /lib/libz.so.1 (0x00007f7ac1109000)
        libm.so.6 => /lib/libm.so.6 (0x00007f7ac1028000)
        libc.so.6 => /lib/libc.so.6 (0x00007f7ac0e3f000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f7ac1126000)
```

### DIFF
--- a/gcc-12.yaml
+++ b/gcc-12.yaml
@@ -1,7 +1,7 @@
 package:
   name: gcc-12
   version: 12.4.0
-  epoch: 0
+  epoch: 2
   description: "the GNU compiler collection - version 12"
   resources:
     cpu: 16
--- a/gcc-12.yaml
+++ b/gcc-12.yaml
@@ -1,7 +1,7 @@
 package:
   name: gcc-12
   version: 12.4.0
-  epoch: 2
+  epoch: 1
   description: "the GNU compiler collection - version 12"
   resources:
     cpu: 16
--- a/gcc-12.yaml
+++ b/gcc-12.yaml
@@ -15,6 +15,10 @@ package:
       - binutils
       - libstdc++-12-dev
       - posix-cc-wrappers
+      - gmp
+      - isl
+      - mpc
+      - mpfr
 
 environment:
   contents:
--- a/gcc-12.yaml
+++ b/gcc-12.yaml
@@ -13,12 +13,12 @@ package:
   dependencies:
     runtime:
       - binutils
-      - libstdc++-12-dev
-      - posix-cc-wrappers
       - gmp
       - isl
+      - libstdc++-12-dev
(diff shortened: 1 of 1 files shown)


imjasonh (dismissed):
Need to bump the epoch?

pnasrat:
@imjasonh done

xnox (dismissed):

xnox:
Indeed this looks broken.

And I wonder if this is happening because it is like first time gcc-12 is being rebuilt in a long time, with new melange, and new gcc.

Will also check if previously libexec was used.

xnox (approved):

smoser:
> And I wonder if this is happening because it is like first time gcc-12 is being rebuilt in a long time, with new melange, and new gcc.
> Will also check if previously libexec was used.

@xnox , did you sort out why this failed?

I'd like to fix/update melange SCA to DTRT.

xnox:
> > And I wonder if this is happening because it is like first time gcc-12 is being rebuilt in a long time, with new melange, and new gcc.
> > Will also check if previously libexec was used.
> 
> @xnox , did you sort out why this failed?
> 
> I'd like to fix/update melange SCA to DTRT.

TBD - i do think we have recently changed which binaries and in which locations we SCA, so i need to go dig back to see which melange change behaviour. Separately working on adding tests that locally do pass with new build and would have failed the previous update - see https://github.com/wolfi-dev/os/pull/23720

Note fixing this universally might be hard - as the binaries in question are not in $PATH, and we cannot tell easily if they need runpath provided libraries or not.

smoser:
> Note fixing this universally might be hard - as the binaries in question are not in $PATH, and we cannot tell easily if they need runpath provided libraries or not.

Is that true? If we're shipping a binary or shared library, we should ship its runtime dependencies.  Am I missing something?

xnox:
> > Note fixing this universally might be hard - as the binaries in question are not in $PATH, and we cannot tell easily if they need runpath provided libraries or not.
> 
> Is that true? If we're shipping a binary or shared library, we should ship its runtime dependencies. Am I missing something?

So something really weird is happening:

PR #23720, opened 2024-07-10
gcc: add C and C++ tests

xnox:
A recent update of gcc-12 resulted in incorrect runtime SCA
dependencies which caused compilers to not work at all. Add tests that
exercise both frontend and underlying compiler binaries to ensure they
are operation and are able to compile C and C++ code.

This is a smoketest, but would have caught the breakage in gcc-12
update.

Explicitely not doing epoch bump, as tests are executed anyway against
existing builds, and epoch bumps of old gcc toolchains likely result
in SCA runtime library breaks. Thus these tests should be landed prior
to future gcc epoch bumps.


xnox:
Too big to merge at onces, splitting.

PR #23730, opened 2024-07-10
add chainguard security guide

joshrwolf:
add chainguard-security-guide (the STIG SRGs content).

the naming of this package is meant to mimic the complianceascode project (`scap-security-guide`), I understand its a little confusing in isolation.

### DIFF
--- a/chainguard-security-guide.yaml
+++ b/chainguard-security-guide.yaml
@@ -1,0 +1,41 @@
+package:
+  name: chainguard-security-guide
+  version: 0.1.1
+  epoch: 0
+  description: Security automation content for Chainguard Images
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - busybox
+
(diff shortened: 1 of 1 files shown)


chainguard-security-guide.yaml
```diff
@@ -0,0 +1,41 @@
+package:
+  name: chainguard-security-guide
+  version: 0.1.0
+  epoch: 0
+  description: Security automation content for Chainguard Images
+  copyright:
+    - license: Apache-2.0
```
xnox, line 7: Please resolve https://github.com/chainguard-dev/stigs/issues/2

xnox (changes requested):

chainguard-security-guide.yaml
joshrwolf: 🤦 its unchecked locally, good catch

xnox (approved):

PR #23868, opened 2024-07-12
fix: add default config file

89luca89:
Fixes:

Add missing config file for grafana-alloy


grafana-alloy.yaml
```diff
@@ -34,6 +34,11 @@ pipeline:
 
   - uses: strip
 
+  - name: Setup default config files
```
imjasonh, line 37: You'll need to update the `epoch` so this gets rebuilt with this change.

89luca89: Done thanks! :+1:

ajayk (approved):

rawlingsj (approved):

PR #23899, opened 2024-07-12
chore(gitlab): Migrate packages with nonstandard releases to gitlab-cng

EyeCantCU:
This migrates the GitLab container registry, ElasticSearch indexer, exporter, logger, mailroom, and shell to GitLab CNG so that we can build packages that align with the current GitLab release. This ensures that GitLab components are not mismatched

Unfortunately, this is the best solution we have for tracking the in-use versions of all of these different components, and will require manual updates

GitLab KAS, runner, and pages all follow the release structure of GitLab so those have been left intact as individual packages


EyeCantCU:
Drafting for now. We should either fetch and validate the source archives via checksum or use our git checkout and set an expected commit. Unfortunately, both solutions require manual intervention

EyeCantCU:
Proper checkouts are in place now. This will require manual updates every time there's a new stream. I've placed the expected commit along with the tag for each component in variables at the top of the package file with some docs that hopefully make maintaining this easier

mamccorm (approved):

hectorj2f (approved):
LGTM, it looks like a False positive. It is common to use that url on health check scripts.

hectorj2f:
It has been detected here https://gitlab.com/gitlab-org/build/CNG/-/blob/master/gitlab-webservice/scripts/healthcheck

PR #23946, opened 2024-07-14
Recheck that go bootstrap still works

xnox:


xnox:
still works!

PR #23949, opened 2024-07-14
gcc cleanups

xnox:
- **gcc: provide runtime libquadmath.so in a subpackage**
  Currently libquadmath subpackage is empty, but should contain runtime
  libquadmath library, as used by boost and hdf5 on x86 only.
  
  aarch64 supports 128 precision without need for libquadmath runtime
  library.

```console
# apk info -L libquadmath
libquadmath-13.3.0-r1 contains:
usr/lib64/libquadmath.so.0
usr/lib64/libquadmath.so.0.0.0
var/lib/db/sbom/libquadmath-13.3.0-r1.spdx.json
```

- **gcc: stop building objc**
  objc compiler typically requires gnustep libraries to be usable, which
  Wolfi does not package. Unused by any of the packages in Wolfi and not
  shipped in any Chainguard Images.
  

- **gcc: set full pkgversion like clang**

```console
# gcc --version
gcc (Wolfi 13.3.0-r1) 13.3.0
Copyright (C) 2023 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

- **gcc: set with-arch & with-mtune**  
  Use current settings of march from build-flags. Bump mtune to
  current/latest generation of silicon available in the public clouds.

```console
# gcc -Q --help=target | grep -e march= -e mtune= | head -n2
  -march=                     		x86-64-v2
  -mtune=                     		sapphirerapids

# gcc -Q --help=target | grep -e march= -e mtune=
  -march=                     		armv8-a+crc+crypto
  -mtune=                     		neoverse-v2
```

- **gcc: (build) avoid linking compilers at the same time**
  
- **gcc: bump epoch**

### DIFF
--- a/gcc.yaml
+++ b/gcc.yaml
@@ -61,7 +61,7 @@ pipeline:
         --enable-default-pie \
         --enable-default-ssp \
         --with-system-zlib \
-        --enable-languages=c,c++,objc,fortran,jit,go \
+        --enable-languages=c,c++,fortran,jit,go \
         --enable-bootstrap \
         --enable-gnu-indirect-function \
         --enable-gnu-unique-object \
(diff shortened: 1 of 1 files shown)


gcc.yaml
```diff
@@ -45,11 +45,22 @@ pipeline:
 
   - name: 'Configure GCC'
     runs: |
+      case "${{build.arch}}" in
```
pnasrat, line 48: For future updates, can you add a comment here on how to identify the current march/mtune from cloud providers

gcc.yaml
```diff
@@ -113,9 +127,10 @@ subpackages:
   - name: "libquadmath"
     description: "128-bit math library provided by GCC"
     pipeline:
-      - runs: |
+      - if: ${{build.arch}} == 'x86_64'
```
pnasrat, line 130: Could the entire subpackage be arch conditional rather than an empty package on aarch64? It doesn't look like melange subpackage supports target-architecture perhaps it should?

https://github.com/chainguard-dev/melange/blob/e855ba4a69a3256aa04754dd96906f222d88cbbd/pkg/config/config.go#L358

pnasrat (dismissed):

smoser (dismissed):

gcc.yaml
xnox:
```
2024-07-15T12:46:39.8424616Z 2024/07/15 12:46:39 INFO skipping subpackage libquadmath because ${{build.arch}} == 'x86_64' == false
```

Looks like substitutions are not done correctly.

xnox: https://github.com/chainguard-dev/melange/issues/1377

smoser: :cry:  You want to just leave it as it was and put in a link?

pnasrat (approved):
LGTM

smoser (approved):
thank you.

PR #23972, opened 2024-07-15
superset: fix GHSA-jfmj-5v4g-7637

hectorj2f:
fix GHSA-jfmj-5v4g-7637


hectorj2f:
This is a False Positive from bincapz, we are safe to merge it.

cpanato (approved):

PR #23973, opened 2024-07-15
Delete EOL gcc

xnox:
gcc 9,10,11 are all EOL, and unused/unneeded.

This is in preparation to add 14 to supported gcc versions 12 and 13.


pnasrat (commented):
Will you be withdrawing these too or just deleting for going ahead?

xnox:
> Will you be withdrawing these too or just deleting for going ahead?

Just deleting for going ahead. As I'm planning to do lots of configuration changes in supported compilers and `build-*.env` files, and don't want to implement support for these EOL compilers.

pnasrat (approved):
LGTM

PR #23974, opened 2024-07-15
fix version streams for loki and pass dependencies.provides to other packages

Dentrax:
* rename `loki-3.0` to `loki-3`
* add missing `dependencies.provides` for some packages
* fixes `loki-3` build due to missing config file

### DIFF
--- a/jruby-9.4.yaml
+++ b/jruby-9.4.yaml
@@ -1,7 +1,7 @@
 package:
   name: jruby-9.4
   version: 9.4.8.0
-  epoch: 0
+  epoch: 1
   description: JRuby, an implementation of Ruby on the JVM
   copyright:
     - license: Apache-2.0
--- a/jruby-9.4.yaml
+++ b/jruby-9.4.yaml
@@ -5,6 +5,9 @@ package:
   description: JRuby, an implementation of Ruby on the JVM
   copyright:
     - license: Apache-2.0
+  dependencies:
+    provides:
+      - jruby=${{package.full-version}}
 
 environment:
   contents:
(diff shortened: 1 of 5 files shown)


gtk-2.yaml
```diff
@@ -1,5 +1,5 @@
 package:
-  name: gtk-2.0
+  name: gtk-2
```
xnox, line 2: @Dentrax gtk-2.0 is not a version stream, but an obsolete fork of the project. All other distributions use "2.0" for the ancient fork, and "3" for the currentish fork. Thus my preference would be to leave this historical name as is.

Note, all subpackages are still called "2.0" - did you want to change all of those too? and thus all reverse depends?

Dentrax: Ah TIL that, thank you! FWIW better keep this as `-2.0`.

rawlingsj (changes requested):
we'll need to bump the epochs so we get the updated package metadata for the provides

Dentrax:
> we'll need to bump the epochs so we get the updated package metadata for the provides

In this case pipeline would take ~1-2 hours. Actually this is not a blocker thing but bot eventually bump the versions so that we get the updated package metadata right? @rawlingsj 

But if you say we should do it in this PR, happy to bump epochs!

rawlingsj:
Yeah I think it's best in case anything breaks we find out now rather than later down the line.  Maybe bump the epochs and I can merge when I wake while it is quiet?

Dentrax:
@rawlingsj this is now ready to go!

rawlingsj (approved):

PR #23989, opened 2024-07-15
java-cacerts: enable automatic updates

xnox:
Enable update tracker of ca-certificates bundle. Ensure that matching
package-version is used for the build-time and runtime-time
dependencies. This way automatic update will pass CI, once
ca-certificates package update lands and a retry is performed.


xnox:
fail

PR #24006, opened 2024-07-15
py3-build add compat tomli dep

pnasrat:
Older pythons may need tomli


pnasrat:
Lint failures new update check by the looks not related

joshrwolf (approved):

PR #24020, opened 2024-07-15
chore(gitlab-cng): Unify package naming

EyeCantCU:
Remove gitlab-cng from package names and append <major>.<minor> version to the end

### DIFF
--- a/gitlab-cng-17.1.yaml
+++ b/gitlab-cng-17.1.yaml
@@ -150,7 +150,7 @@ subpackages:
               fi
             done
 
-  - name: ${{package.name}}-gitlab-container-registry
+  - name: ${{package.name}}-container-registry
     description: The GitLab Container Registry originated as a fork of the Docker Distribution Registry, now CNCF Distribution, both distributed under Apache License Version 2.0.
     dependencies:
       provides:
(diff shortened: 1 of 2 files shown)


ajayk (dismissed):
Needs epoch bump

developer-guy:
I fixed the cves

gitlab-cng-17.1.yaml
```diff
@@ -233,7 +237,7 @@ subpackages:
             # Stop the registry
             kill $PID
 
-  - name: ${{package.name}}-gitlab-container-registry-compat
+  - name: ${{package.name}}-container-registry-compat
```
mamccorm, line 240: Should we use the var-transforms approach here and append only the version number at the end? 

i.e this change would produce:
gitlab-cng-17.1-container-registry-compat

whereas we ideally want
gitlab-cng-container-registry-compat-17.1

EyeCantCU: I agree, that's a lot better. I'll get a commit up

EyeCantCU: Realized in the process that if we're going to track the version that way, we can eliminate `gitlab-cng` from the package name entirely too

rawlingsj (dismissed):

hectorj2f (approved):
LGTM, we marked this CRITICAL as a False Positive here: https://github.com/chainguard-dev/bincapz/pull/364

hectorj2f:
We are safe to merge

rawlingsj (approved):

PR #24051, opened 2024-07-16
nss: enable library self-tests and checksums

xnox:
Enable library self-tests and checksums. Note there is no
certification for this build.

This is on parity with what other community distributions do without
any certification (Debian, Fedora, devel releases).


pnasrat (approved):
LGTM

PR #24113, opened 2024-07-17
fix(gitlab-exporter-17.1): Unlock gemspec

EyeCantCU:
Needed to substitute expected gems with ours


egibs:
The bincapz finding is a false positive that was addressed yesterday.

jamonation (approved):

PR #24126, opened 2024-07-17
keycloak: add bitnami-compat

Dentrax:
This PR adds a new `keycloak-bitnami-compat` subpackage to `keycloak`.

### DIFF
--- a/keycloak.yaml
+++ b/keycloak.yaml
@@ -135,6 +135,7 @@ subpackages:
       environment:
         contents:
           packages:
+            - curl
             - postgresql
             - postgresql-client
             - shadow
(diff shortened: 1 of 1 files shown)


keycloak.yaml
```diff
@@ -62,6 +62,144 @@ subpackages:
...
+              gid: 65532
+              uid: 65532
+          run-as: 0
+        environment:
+          PGDATA: /tmp/test_db
+          PGUSER: bn_keycloak
+          PGDB: bitnami_keycloak
+          ALLOW_EMPTY_PASSWORD: yes
+      pipeline:
+        - working-directory: /home/nonroot
```
stormqueen1990, line 156: Your working directory might need to be outside of the bind mount.

```suggestion
        - working-directory: /tmp
```

keycloak.yaml
```diff
@@ -62,6 +62,147 @@ subpackages:
...
+    dependencies:
+      runtime:
+        - krb5
+        - libaio
+        - procps
+        - zlib
+        - wait-for-port
+        - net-tools
+        - posix-libc-utils
+        - su-exec
```
smoser, line 77: keycloak-compat shoudl depend on keycloak?

keycloak.yaml
```diff
@@ -62,6 +62,147 @@ subpackages:
       - runs: |
           mkdir -p "${{targets.subpkgdir}}"/opt
           ln -s /usr/share/java/keycloak "${{targets.subpkgdir}}"/opt/
+  # https://github.com/bitnami/containers/tree/main/bitnami/keycloak/24/debian-12
+  - name: ${{package.name}}-bitnami-compat
```
smoser, line 66: as it is right now (in local testing) this package contains only a single symlink

```
# apk info -L keycloak-compat
keycloak-compat-25.0.2-r0 contains:
opt/keycloak
var/lib/db/sbom/keycloak-compat-25.0.2-r0.spdx.json
```

was that intended?

smoser: my bad. that is keycloak-compat, not bitnami-compat. ignore.

keycloak.yaml
Dentrax: No actually, for the version-streaming, it should be added in the image manifest

smoser: I admit to not fully understanding version-streaming. but why would you not add the dependency here? you can't use keycloak-bitnami-compat without keycloak.

Dentrax: > you can't use keycloak-bitnami-compat without keycloak.

Correct, so we need to install both of at the same time.

> but why would you not add the dependency here?

If we have that dep here, we would copy the entire bitnami-compat for other versions by including version suffix, `keycloak-bitnami-compat-X.Y`. We also wouldn't consume this subpkg for FIPS package. By not including the `keycloak` package in the runtime deps, we would only have 1 compat package, without needing duplicating.

keycloak.yaml
```diff
@@ -62,6 +62,147 @@ subpackages:
...
+                timeout: 60
+                expected_output: |
+                  Welcome to the Bitnami keycloak container
+                  ** keycloak setup finished! **
+                  server listening on
+                  Keycloak ${{package.version}}
+                  Profile dev activated.
+                post: |
+                  #!/bin/sh -e
+                  apk add curl
```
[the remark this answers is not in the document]
Dentrax: thanks, added curl to packages!

smoser (approved):

mamccorm (approved):
Nice work :)

PR #24130, opened 2024-07-17
gcc-12: openssf integration

xnox:
Bring in relevant improvements from #23949 and #24194:

- set default march & mtune settings
- set package version information
- drop manual runtime dependencies, as `no-provides` issues were all fixed
- enable Intel CET automatically
- do not link mutliple compilers simultaneously (should reduce peak RAM usage)
- install internal binaries into major number subpaths, thus making spec files locations predictable

Bumped reverse-build dependencies to test rebuilds of them.

### DIFF
--- a/datadog-agent.yaml
+++ b/datadog-agent.yaml
@@ -1,7 +1,7 @@
 package:
   name: datadog-agent
   version: 7.55.2
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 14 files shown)


xnox:
In addition to the above provided rebuilds; I have also locally successfully rebuilt tensorflow and cilium.

pnasrat (approved):
LGTM

PR #24134, opened 2024-07-17
revert symlink change

mamccorm:
https://github.com/wolfi-dev/os/commit/3fac15ee0dac0f6c02e4ff16fc869b53ce81a662 was recently committed, as part of a batch of other changes to try and make our image behave closer to the upstream image. Unfortunately, it has caused some unexpected issues, so we're rolling back this change.

### DIFF
--- a/nginx-mainline.yaml
+++ b/nginx-mainline.yaml
@@ -157,7 +157,6 @@ subpackages:
           mkdir -p ${{targets.subpkgdir}}/etc/nginx/conf.d
           install -m644 -D nginx.conf ${{targets.subpkgdir}}/etc/nginx/
 
-          # Should be exported as conf.d/default.conf as per https://hg.nginx.org/pkg-oss/file/tip/alpine/alpine/APKBUILD-base.in#l87
           install -m644 -D nginx.default.conf ${{targets.subpkgdir}}/etc/nginx/conf.d/default.conf
           install -m644 -D nginx.default.conf ${{targets.subpkgdir}}/etc/nginx/conf.d
 
--- a/nginx-mainline.yaml
+++ b/nginx-mainline.yaml
@@ -157,7 +157,6 @@ subpackages:
           mkdir -p ${{targets.subpkgdir}}/etc/nginx/conf.d
           install -m644 -D nginx.conf ${{targets.subpkgdir}}/etc/nginx/
 
-          install -m644 -D nginx.default.conf ${{targets.subpkgdir}}/etc/nginx/conf.d/default.conf
           install -m644 -D nginx.default.conf ${{targets.subpkgdir}}/etc/nginx/conf.d
 
           # Follow Docker Image and link logs to to stdout/stderr
--- a/nginx-mainline.yaml
+++ b/nginx-mainline.yaml
@@ -222,6 +222,30 @@ subpackages:
           cp -r ./nginx-mainline/* ${{targets.contextdir}}/usr/src/nginx
           ls -latr ${{targets.contextdir}}/usr/src/nginx
 
+  - name: ${{package.name}}-config-compat
+    description: Nginx config compatibility with upstream image
+    dependencies:
+      runtime:
+        - nginx-mainline
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/etc/nginx/conf.d
+          install -m644 -D nginx.default.conf ${{targets.subpkgdir}}/etc/nginx/conf.d/
+          ln -s /etc/nginx/conf.d/nginx.default.conf ${{targets.subpkgdir}}/etc/nginx/conf.d/default.conf
+
+test:
+  environment:
+    contents:
+      packages:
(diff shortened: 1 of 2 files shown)


nginx-mainline.yaml
```diff
@@ -156,10 +156,10 @@ subpackages:
       - runs: |
           mkdir -p ${{targets.subpkgdir}}/etc/nginx/conf.d
           install -m644 -D nginx.conf ${{targets.subpkgdir}}/etc/nginx/
-          install -m644 -D nginx.default.conf ${{targets.subpkgdir}}/etc/nginx/conf.d/
 
           # Should be exported as conf.d/default.conf as per https://hg.nginx.org/pkg-oss/file/tip/alpine/alpine/APKBUILD-base.in#l87
```
stormqueen1990, line 160: This comment doesn't apply anymore.

```suggestion
```

nginx-mainline.yaml
```diff
@@ -156,10 +156,9 @@ subpackages:
       - runs: |
           mkdir -p ${{targets.subpkgdir}}/etc/nginx/conf.d
           install -m644 -D nginx.conf ${{targets.subpkgdir}}/etc/nginx/
-          install -m644 -D nginx.default.conf ${{targets.subpkgdir}}/etc/nginx/conf.d/
 
-          # Should be exported as conf.d/default.conf as per https://hg.nginx.org/pkg-oss/file/tip/alpine/alpine/APKBUILD-base.in#l87
-          ln -s /etc/nginx/conf.d/nginx.default.conf ${{targets.subpkgdir}}/etc/nginx/conf.d/default.conf
+          install -m644 -D nginx.default.conf ${{targets.subpkgdir}}/etc/nginx/conf.d/default.conf
+          install -m644 -D nginx.default.conf ${{targets.subpkgdir}}/etc/nginx/conf.d
```
stormqueen1990, line 161: I think this isn't quite right 🤔 
To go back to the original state, we'd have to fully remove line 160:
https://github.com/wolfi-dev/os/blob/7c9113b966549b10db93bd568b852d9a3d43561b/nginx-mainline.yaml#L150-L154
```suggestion
          install -m644 -D nginx.default.conf ${{targets.subpkgdir}}/etc/nginx/conf.d
```

nginx-mainline.yaml
```diff
@@ -223,6 +222,30 @@ subpackages:
...
 
+  - name: ${{package.name}}-config-compat
+    description: Nginx config compatibility with upstream image
+    dependencies:
+      runtime:
+        - nginx-mainline
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/etc/nginx/conf.d
+          install -m644 -D nginx.default.conf ${{targets.subpkgdir}}/etc/nginx/conf.d/
```
stormqueen1990, line 233: I'm wondering if we can get away with just:
```suggestion
```

nginx-stable.yaml
```diff
@@ -156,10 +156,9 @@ subpackages:
       - runs: |
           mkdir -p ${{targets.subpkgdir}}/etc/nginx/conf.d
           install -m644 -D nginx.conf ${{targets.subpkgdir}}/etc/nginx/
-          install -m644 -D nginx.default.conf ${{targets.subpkgdir}}/etc/nginx/conf.d/
 
-          # Should be exported as conf.d/default.conf as per https://hg.nginx.org/pkg-oss/file/tip/alpine/alpine/APKBUILD-base.in#l87
-          ln -s /etc/nginx/conf.d/nginx.default.conf ${{targets.subpkgdir}}/etc/nginx/conf.d/default.conf
+          install -m644 -D nginx.default.conf ${{targets.subpkgdir}}/etc/nginx/conf.d/default.conf
```
stormqueen1990, line 160: Same here
```suggestion
```

nginx-stable.yaml
```diff
@@ -216,6 +215,30 @@ subpackages:
...
 
+  - name: ${{package.name}}-config-compat
+    description: Nginx config compatibility with upstream image
+    dependencies:
+      runtime:
+        - nginx-stable
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/etc/nginx/conf.d
+          install -m644 -D nginx.default.conf ${{targets.subpkgdir}}/etc/nginx/conf.d/
```
stormqueen1990, line 226: Same as the previous one
```suggestion
```

nginx-mainline.yaml
mamccorm: It was failing without line 231 "no file / directory", even with nginx-mainline as a runtime dep, so thats why I added in line 231. For line 232,  you don't think we meed nginx.default.conf copied in now?

stormqueen1990: Ah interesting, please disregard this comment then. It makes sense since each subpackage will have its own target directory.

stormqueen1990 (approved):
LGTM!

PR #24158, opened 2024-07-17
openjdk: add tests for openjdk17 and openjdk 21

ajayk:
- Adds tests for openjdk 17 and openjdk 21  
- No epoch needed as these are tests only


pnasrat (approved):
LGTM

PR #24174, opened 2024-07-17
Disable the dual upload for now.

mattmoor:
This is in case the red X is annoying people, but the current upload failures shouldn't be blocking anything.


k4leung4 (approved):

mattmoor:
We need to add this identity to our allow-list 🤦

PR #24192, opened 2024-07-18
New package openssf-compiler-options

xnox:
This package aims to provide compiler configuration files that set flags as per https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.html

It also enables optimized builds by default, for Wolfi baseline ABI.

Two test pipelines introduced. One to ensure ELF binaries in hardened packages remain hardened. The other one to test integration of openssf-compiler-flags and compilers - such that one can execute the same test whenever compiler is updated or openssf-compiler-flags are updated.

For clang toolchains the config options are prepended to any user supplied options.

For gcc toolchains the spec file options are appended to any user supplied options, hence negative tests for already supplied options when overriding explicit upstream/user choices is not desired.

Compiler check verifies that with openssf-compiler-options installed, a given compiler produces hardened binaries without any environmental variables are command line arguments:

```console
2024/07/31 16:25:46 WARN + hardening-check --color hello
2024/07/31 16:25:46 INFO hello:
2024/07/31 16:25:46 INFO  Position Independent Executable: yes
2024/07/31 16:25:46 INFO  Stack protected: yes
2024/07/31 16:25:46 INFO  Fortify Source functions: yes (some protected functions found)
2024/07/31 16:25:46 INFO  Read-only relocations: yes
2024/07/31 16:25:46 INFO  Immediate binding: yes
2024/07/31 16:25:46 INFO  Stack clash protection: yes
2024/07/31 16:25:46 INFO  Control flow integrity: yes
```

Reviews:
- This is mostly about infrastructure to support passing lots of compiler options via config files. Note we can re-arrange how they are shipped, add additional file includes and redirects.
- The options added are a combination of
-- https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.html
-- https://github.com/wolfi-dev/os/blob/main/build-x86_64.env
-- https://github.com/wolfi-dev/os/blob/main/build-aarch64.env
-- defaults passed to gcc, which at compiler configuration time not available for clang thus
-- https://github.com/wolfi-dev/os/blob/main/gcc.yaml#L52
- The intention is that switching to this package in our build environment can allow us to drop/remove C/C++ variables from the build-*.env files
- Test pipelines are added to recheck openssf-compiler-flags package; compilers; and C/C++ binaries to ensure these flags/integration does not regress, and that binaries produced remain hardened
- hardening-check script https://github.com/wolfi-dev/os/blob/main/hardening-check.yaml used as an approximation test for most important hardening checks

Pre-depends (done):
- gcc https://github.com/wolfi-dev/os/pull/25169
- gcc-12 https://github.com/wolfi-dev/os/pull/24130

Follow-ups:
- https://github.com/wolfi-dev/os/pull/25356
- Opt-in individual packages to use openssf-compiler-options
- Switch to have openssf-compiler-options preinstalled during builds
- Add c++ compiler-hardening-check


pnasrat (approved):
LGTM

PR #24194, opened 2024-07-18
New version stream gcc-13

xnox:
This ideally should land simultaneously with gcc-14 upgrade, or just before it:
- #24198


xnox:
Merged in https://github.com/wolfi-dev/os/pull/24198

PR #24234, opened 2024-07-18
Add sniff test to tomcat-10.1.

smoser:
The test is similar to that found in the current image test for tomcat. It verifies tomcat starts and can run the sample webapp.


pnasrat (dismissed):
LGTM

pnasrat (approved):
LGTM

PR #24352, opened 2024-07-20
py3-pytest: new version and py-versioned

pnasrat:


pnasrat:
Upstream released 8.3.0 then .1 shortly after today this probably needs more testing

rawlingsj (approved):

PR #24445, opened 2024-07-22
Drop flit-core from pip deps.

smoser:
* Add py/pip-check pipeline for running 'pip check'
*     py3-pip - drop flit-core as a runtime dependency.
    
    Pip itself does not have any imports of flit or flit_core.
    This was added in the pyversions refactor probably because
    of a misunderstood dependency.  Several packages have
    a pyproject entry like wheel does:
    
       pyproject.toml:requires = ["flit_core >=3.8,<4"]
       pyproject.toml:build-backend = "flit_core.buildapi"
    
    And a simple 'pip build' will show a stack trace that doesn't
    actually even leave pip so it appears/appeared that pip
    had a flit_core dep.
    
    Reality is that flit-core should be in the build environment
    of those packages.

 *  bump python projects to rebuild due to flit-core move


smoser:
I left this as draft for now. I'm locally re-building all things that have 'py3-pip' or 'py3-supported-pip' in their yaml.

pnasrat (changes requested):
Only change runtime dep in py3-pip in this PR

pnasrat (dismissed):
LGTM

Disregard previous comment

pnasrat (dismissed):
LGTM

pnasrat (approved):
LGTM

PR #24448, opened 2024-07-22
add packer v1.9.5: new package

willswire:
Fixes:

Related:

#### For new package PRs only

#### For version bump PRs


mamccorm:
Thanks for the contribution. We only add latest package versions to wolfi, and in this instance cannot add newer versions due to licensing changes.

PR #24449, opened 2024-07-22
fix: drop custom licenses

imjasonh:


wlynch (approved):
qq: is this okay to do? My initial impression of this field was to call out the OSI license that this software is released under (e.g. MIT, Apache, etc), but here we seem to be doing `LicenseRef-<package>` which feels the same as saying "the project's license is the project's license".

imjasonh:
cc @xnox probably has a better answer

The goal is to declare the license being used. If there's an SPDX license identifier, we use that. If the license isn't on SPDX's list, we can declare any other license this way. If that license is OSI approved (e.g., berkeley DB's) then I think we're good. If it's not, we can also consider removing it from the repo.

I don't have all the context about why SPDX doesn't include berkeley DB's license, or the xorg font licenses. This is just to be able to keep rebuilding these packages. If we decide their license is incompatible with this repo, that's a separate conversation IMO.

ajayk (approved):

xnox:
If I am not mistaken all of these license texts are unique and say "you must include this text along with your binary distribution" (sort of advertising clause).

And doing this via licenseref ensures that said text is included in the spdx alongside the package and the image.

@wlynch @imjasonh

wlynch:
berkeley db, font-util, xproto look like multi-licenses - I'm not sure whether this is because of the project being re-licensed over time or they are including licenses for vendored dependencies within the project. Most of these look like OSI licenses BSD, Sleepycat, MIT, etc.

The rest of the fonts are non-OSI but permissive licenses, with some only saying "Public domain".

I'm not sure I fully understand how license refs are handled - what tells melange to copy the license to the package? Do we also do this for other [notice-style licenses](https://opensource.google/documentation/reference/thirdparty/licenses#notice) that use SPDX IDs rather than license-ref? Any documentation you can point me to? I couldn't find any mention of this in the [docs](https://github.com/chainguard-dev/melange/blob/main/docs/BUILD-FILE.md#license) 🙏

PR #24460, opened 2024-07-23
fix(neuvector-db): Update daily

EyeCantCU:
I set up a repository here, intended to automatically cut releases over intervals of time: https://github.com/EyeCantCU/timed-releases

Use this to pull the latest CVE DB for NeuVector every day


rawlingsj (changes requested):
I think we may need to think a bit more about this one. 

I'm not sure that packaging the CVE DB daily is the right way as it will mean Wolfi becomes responsible for insuring the DB is up to date. 

@EyeCantCU has shared some docs so will take a look this morning and get a bit more familiar with it.

PR #24470, opened 2024-07-23
Fix some python FTBFS , add tests, bump conda/conda-build

smoser:
* melange SCA sees the python3 dep, so it is not needed
 * python3 is needed in build env after python3 base restructure
 * add test


pnasrat (approved):
Lgtm

PR #24471, opened 2024-07-23
font-xproto: MIT license

imjasonh:
https://gitlab.freedesktop.org/xorg/proto/xproto


wlynch (approved):

wlynch:
For future reference - all licenses in here appear to be MIT variants (some being [old style](https://fedoraproject.org/wiki/Licensing:MIT#Old_Style))

mamccorm (approved):

PR #24499, opened 2024-07-23
fix py3-gcovr

smoser:
*  py3-gcovr - fix missing dep, add tests
 *  Add py/pip-check pipeline for running 'pip check'
    
    This is most useful for packages that we have manually
    encoded runtime package dependencies.  Running 'pip check'
    will verify that the versions we have match expectation.

Then... drop py/pip-check.


pipelines/py/pip-check.yaml
```diff
@@ -0,0 +1,42 @@
+name: Use pip from a zipapp for bootstrap. Only for use with bootstrapping.
```
pnasrat, line 1: please make the name and comments meaningful for what this pipeline does.

pnasrat (changes requested):

pipelines/py/pip-check.yaml
smoser: done.

pnasrat (approved):
LGTM

PR #24775, opened 2024-07-23
Systemd init

dustinkirkland:
Provide a slightly smarter /init in Wolfi's systemd-init (than just symlinking /init to /usr/lib/systemd/systemd).  Ultimately, this script will do that (see the last line).  But it'll also handle a switch_root persistent disk boot as well.

This version supports a couple of different modes, including initializing a system within the initramfs (ephemeral), as well as finding and mounting a root partition (stateful).  I've tested this as widely as I could, and it works in all of the scenarios I could think of.

Considering we don't really yet fully support booting, I'd recommend that we start here and iterate on it.

Also, there's another change in here, that adds an opportunistic wired network startup.

Thanks!
DK


systemd/init
```diff
@@ -0,0 +1,52 @@
...
+findmnt /sys >/dev/null  2>&1 || mount -t sysfs none /sys
+findmnt /dev >/dev/null  2>&1 || mount -t devtmpfs none /dev
+
+# Allow override of path to init on kernel command line
+if (grep -qs "init=" $CMDLINE); then
+	init=$(sed -e "s:.*\binit=::" -e "s: .*::" $CMDLINE)
+	if [ ! -z "$init" ] && [ -x "$init" ]; then
+		# Exec init as overriden on the kernel command line
+		info "Running [$init] as init"
+		exec $init
```
smoser, line 32: you're not honoring init= in if there is a 'root='

systemd/init
```diff
@@ -0,0 +1,52 @@
...
+
+# Override /proc/cmdline path as environment variable, useful for testing
+[ -z "$CMDLINE" ] && CMDLINE="/proc/cmdline"
+info "cmdline = [$(cat $CMDLINE)]"
+
+# Setup our system mounts if needed
+mkdir -p /proc /sys /dev
+findmnt /proc >/dev/null 2>&1 || mount -t proc none /proc
+findmnt /sys >/dev/null  2>&1 || mount -t sysfs none /sys
+findmnt /dev >/dev/null  2>&1 || mount -t devtmpfs none /dev
```
smoser, line 24: why would /proc or /sys or /dev ever be mounted?

systemd/init
```diff
@@ -0,0 +1,52 @@
...
+
+# Allow override of path to init on kernel command line
+if (grep -qs "init=" $CMDLINE); then
```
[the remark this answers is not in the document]
dustinkirkland: Indeed.  That was a relic of an earlier implementation where there was some additional logic.  I'll get rid of it.

systemd/init
dustinkirkland: Great catch.  You're absolutely right.

dustinkirkland: Okay, fixed.

systemd/init
dustinkirkland: During a real, actual, fresh boot, they should not.  But I was also doing some testing in a container, just trying to get the shell logic right.  And in the container, they are already mounted.

systemd/init
```diff
@@ -0,0 +1,49 @@
...
+if grep -qs "init=" $CMDLINE 2>/dev/null; then
+	init=$(sed -e "s:.*\binit=::" -e "s: .*::" $CMDLINE)
+fi
+info "Running [$init] as init"
+
+# Find and mount our root partition by uuid or block device, if specified
+if grep -qs "\broot=" $CMDLINE 2>/dev/null; then
+	mkdir -p /root_dir
+	root_dev=$(sed -e "s:.*\broot=::" -e "s: .*::" $CMDLINE)
+	if (echo "$root_dev" | grep -qs "^[0-9a-f\-]\+$"); then
```
smoser, line 37: here is another subshell.

systemd/init
```diff
@@ -0,0 +1,49 @@
...
+#     VERBOSE=1 for debugging
+#     CMDLINE=/path/to/file to override /proc/cmdline, useful for testing
+# - Mount /proc, /sys, /dev if not already mounted
+# - Allow for override of path to /init with init=/path/to/init on kernel /proc/cmdline
+# - Allow for uuid or device path for root filesystem with root= on kernel /proc/cmdline
+# - Mount root, if specified and found, switch_root to it, and run systemd as init
+# - Otherwise, try to run systemd as init in the initrd
+
+# Verbose debugging
+info() { [ "$VERBOSE" = "1" ] && printf "INFO: $@\n" 1>&2; }
```
smoser, line 14: you probably mean
```
printf "INFO: %s\n" "$*" 1>&1
```

at least i think thats how you're using it.  ie, you aren't intending to expose printf's formatting to callers.

systemd/init
dustinkirkland: Thanks, fixed.

systemd/init
dustinkirkland: Sure, that works.

systemd/init
```diff
@@ -0,0 +1,49 @@
+#!/bin/busybox sh
```
mattmoor, line 1: Should we have a `set -e` in here? 🤔

dustinkirkland: Sure, I went ahead an added it.  Initially I was thinking to leave it out, and let the script get as far as it can, but I agree with you -- better to be a little more deterministic here and fail on any error.

systemd/init
```diff
@@ -0,0 +1,51 @@
...
+
+# Verbose debugging
+info() { [ "$VERBOSE" = "1" ] && printf "INFO: %s\n" "$*" 1>&2; }
+
+# Override /proc/cmdline path as environment variable, useful for testing
+[ -z "$CMDLINE" ] && CMDLINE="/proc/cmdline"
+info "cmdline = [$(cat $CMDLINE)]"
+
+# Setup our system mounts if needed
+mkdir -p /proc /sys /dev
```
89luca89, line 23: should we move the creation of the basic rootfs tree to the package itself? like in this commit: https://github.com/wolfi-dev/os/pull/24903/commits/9c383d2f35d5e1d9634d14003f051004cbd22609

systemd/init
```diff
@@ -0,0 +1,51 @@
...
+findmnt /proc >/dev/null 2>&1 || mount -t proc none /proc
```
89luca89, line 24: we could add `nodev,nosuid,hidepid=2` to harden the mountpoint

systemd/init
```diff
@@ -0,0 +1,51 @@
...
+findmnt /sys >/dev/null  2>&1 || mount -t sysfs none /sys
+findmnt /dev >/dev/null  2>&1 || mount -t devtmpfs none /dev
```
89luca89, line 26: Also here we can add `nosuid,noexec`

systemd/init
```diff
@@ -0,0 +1,51 @@
...
+info() { [ "$VERBOSE" = "1" ] && printf "INFO: %s\n" "$*" 1>&2; }
+
+# Override /proc/cmdline path as environment variable, useful for testing
+[ -z "$CMDLINE" ] && CMDLINE="/proc/cmdline"
+info "cmdline = [$(cat $CMDLINE)]"
+
+# Setup our system mounts if needed
+mkdir -p /proc /sys /dev
+findmnt /proc >/dev/null 2>&1 || mount -t proc none /proc
+findmnt /sys >/dev/null  2>&1 || mount -t sysfs none /sys
```
89luca89, line 25: Also here we can add `nodev,nosuid,noexec`

systemd/init
```diff
@@ -0,0 +1,51 @@
...
+findmnt /dev >/dev/null  2>&1 || mount -t devtmpfs none /dev
+
+# Allow override of path to init on kernel command line
+init="/usr/lib/systemd/systemd"
+if grep -qs "init=" $CMDLINE 2>/dev/null; then
+	init=$(sed -e "s:.*\binit=::" -e "s: .*::" $CMDLINE)
```
89luca89, line 31: what about having `init=` and `args=` so we can have a full entrypoint override?

systemd/init
dustinkirkland: Hmm...maybe.  I like it here, especially with the "-p" option to make absolutely sure that these mount points exist, just before using them.  This feels very safe, and defensive, and also ensures that this script can run under other circumstances (perhaps in a debug mode, outside of this package.).

systemd/init
dustinkirkland: Sure, I'm good with that.

systemd/init
dustinkirkland: Sure, that's a good idea.

systemd/init
dustinkirkland: Yep, good idea.

systemd/init
dustinkirkland: Great idea, I agree.

dustinkirkland: Can I leave you to add that one, as a followup?  (args= that is)  I suspect there's going to be some careful work needed to handle quotes and escaping.

PR #24781, opened 2024-07-23
py3-aiohttp: switch to git-checkout, github updates; update patch

ritster:


ritster:
Context: https://github.com/wolfi-dev/os/pull/24443 - py3-aiohttp package is mistakenly picking up pre-release versions, so switch to git-checkout and GitHub update pipelines.

hectorj2f (approved):

PR #24903, opened 2024-07-25
feat(melange): add microvm-init subpackage for qemu runners

89luca89:
Add a melange subpackage to install a minimal busybox-based init for microvm/qemu runners.

### DIFF
--- a/melange/init
+++ b/melange/init
@@ -57,9 +57,9 @@ fi
 # If we have an external disk, we want to perform builds in that
 if [ -e /dev/vda ]; then
 	/usr/sbin/modprobe ext4
-	/sbin/mkfs.ext4 /dev/vda
+	/sbin/mkfs.ext4 -O ^has_journal -E nodiscard /dev/vda
 	/bin/mkdir -p /mount
-	/bin/mount -t ext4 /dev/vda /home/build
+	/bin/mount -o nobarrier,noatime,nodiratime,nobh -t ext4 /dev/vda /home/build
 fi
 
 # Setup SSH keys for external access
--- a/melange/init
+++ b/melange/init
@@ -17,6 +17,16 @@ if [ -x /sbin/ldconfig ]; then
 	/sbin/ldconfig /lib /usr/lib /usr/lib64 || :
 fi
 
+# Fetch settings from cmdline
+DNS="$(tr ' ' '\n' < /proc/cmdline | /bin/grep "dns=" | /usr/bin/cut -d'=' -f2-)"
+HOSTNAME="$(tr ' ' '\n' < /proc/cmdline | /bin/grep "hostname=" | /usr/bin/cut -d'=' -f2-)"
+SSHKEY="$(tr ' ' '\n' < /proc/cmdline | /bin/grep "sshkey=" | /usr/bin/cut -d'=' -f2- | /bin/base64 -d)"
+
+if [ -z "${SSHKEY}" ]; then
+	/bin/echo "Missing default mount for ssh keys"
+	exit 1
+fi
+
 # We can ignore this fail, if we use a kernel with kvm_guest.config, we won't need this
 # and network will work anyway
 # If this fails and we won't have network, the ifconfig command will fail anyway.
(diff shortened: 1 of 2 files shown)


melange/init
```diff
@@ -0,0 +1,76 @@
+#!/bin/busybox sh
+set -e
+set -x
+
+/bin/mount -t proc proc -o nodev,nosuid,hidepid=2 /proc
+/bin/mount -t devtmpfs -o nosuid,noexec devtmpfs /dev
```
smoser, line 6: just set `PATH=/usr/bin:/usr/sbin:/sbin:/bin`
and export that and don't call things by their full path.

melange/init
```diff
@@ -0,0 +1,76 @@
...
+/bin/mount -t tmpfs -o nodev,nosuid,noexec tmpfs /tmp
+
+# Setup tty/pty
+/bin/mkdir /dev/pts
+/bin/mount -t devpts devpts -o noexec,nosuid,newinstance,ptmxmode=0666,mode=0620,gid=tty /dev/pts/
+/bin/mount --bind /dev/pts/ptmx /dev/ptmx
+
+# ldconfig is run to prime ld.so.cache for glibc packages which require it.
+if [ -x /sbin/ldconfig ]; then
+	/sbin/ldconfig /lib /usr/lib /usr/lib64 || :
```
smoser, line 17: Why `|| :` ? do you have a case where this fails and thats ok?

As it is, you're just swallowing an error that is going to pop up later and be harder to debug later.

I wonder why we should have to run /sbin/ldconfig. Shouldn't our built images work optimally on read-only filesystem?

melange/init
```diff
@@ -0,0 +1,76 @@
...
+# Also we load cpu accelleration drivers in case those are needed.
+/usr/sbin/depmod -a || :
+/usr/bin/sort -u \
+	/sys/devices/system/cpu/modalias  \
+	/sys/devices/pci*/*/virtio*/modalias | /usr/bin/xargs -n1 /usr/sbin/modprobe 2>/dev/null || :
+# modprobe 9p if absent
+/bin/grep -q 9p /proc/filesystems || /usr/sbin/modprobe 9p
+
+# Setup default mountpoint for 9p shared dir
+/bin/mount -t 9p -o trans=virtio defaultshare /mnt/ -oversion=9p2000.L
```
smoser, line 42: this is a weird combination of '-o key=value' and '-okey=value'.

```
mount -t 9p -otrans=virtio,version=9p2000.L defaultshare /mnt
```

or

```
mount -t 9p -otrans=virtio -oversion=9p2000.L defaultshare /mnt
```

Otherwise I'm going to read this one day and assume that '-oversion' is somehow _not_ the same as '-o version'

melange/init
```diff
@@ -0,0 +1,76 @@
...
+
+# If we have an external disk, we want to perform builds in that
+if [ -e /dev/vda ]; then
+	/usr/sbin/modprobe ext4
+	/sbin/mkfs.ext4 -O ^has_journal -E lazy_itable_init=0,lazy_journal_init=0 /dev/vda
+	/bin/mkdir -p /mount
```
smoser, line 48: this is vestigial ?

melange/init
```diff
@@ -0,0 +1,76 @@
...
+	/bin/mount -o nobarrier,noatime,nodiratime,nobh -t ext4 /dev/vda /home/build
+fi
+
+# Fix home permissions from cpio command
+/bin/chown -R build:build /home/build
+
+# Setup default network
+/bin/ifconfig lo up
+/bin/ifconfig eth0 10.0.2.15 netmask 255.255.255.0
```
smoser, line 57: ifconfig is deprecated for over a decade.

```
ip link set up dev lo
ip addr add 10.0.2.15/8 dev eth0
ip route add default via  10.31.21.1 dev eth0
```
I have questions on the stability of the name eth0, but ignoring that for now.

melange/init
89luca89: Thanks, fixed this, I think it was a leftover from older debugging

melange/init
89luca89: Yea, it's easier, implemented, thanks!

melange/init
89luca89: Yes the flags were indeed a leftover, thanks!

melange/init
89luca89: Moved to iproute2 and detecting interface name now :+1:

melange/init
89luca89: Fixed,thanks!

mattmoor (approved):
@smoser did you have anything else on this?

PR #24973, opened 2024-07-26
Revert "neon/6142 package update"

rawlingsj:
Reverts wolfi-dev/os#24935

### DIFF
--- a/neon.yaml
+++ b/neon.yaml
@@ -1,6 +1,6 @@
 package:
   name: neon
-  version: "6142"
+  version: "6138"
   epoch: 0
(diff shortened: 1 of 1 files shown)


imjasonh (approved):

imjasonh:
The CI fails because glibc has been updated and old neon no longer builds with it (neither does new neon)

Force merging this since in post-submit it should not try to rebuild it, and we can unblock the rest of the build.

We'll have to figure out how to get neon (old or new) to build with latest glibc.

SamyDgz (approved):

PR #24999, opened 2024-07-26
withdraw glibc 2.40 which causes failure to link binaries

xnox:
Generated using:

    $ apkrane ls https://packages.wolfi.dev/os/x86_64/APKINDEX.tar.gz --latest | sort | grep 2.40-r0 >> withdrawn-packages.txt


xnox:
see #24998

PR #25087, opened 2024-07-29
Upgrade transitive yarn devDependencies to resolve CVEs

xnox:
Whilst they are all listed as devDependencies, I cannot tell if there is any impact on the production web-packed js ui application.


xnox:
```console
# grype /usr/lib/argo-workflows-ui
 ✔ Vulnerability DB                [updated]  
 ✔ Indexed file system                                                                                                          /usr/lib/argo-workflows-ui
 ✔ Cataloged contents                                                                     187cf4ec2fe147d1c031adfe02b22996f7e2a47aa098bf80c61b9c763ae44f6d
   ├── ✔ Packages                        [1,250 packages]  
   └── ✔ Executables                     [0 executables]  
 ✔ Scanned for vulnerabilities     [0 vulnerability matches]  
   ├── by severity: 0 critical, 0 high, 0 medium, 0 low, 0 negligible
   └── by status:   0 fixed, 0 not-fixed, 0 ignored 
[0000]  WARN no explicit name and version provided for directory source, deriving artifact ID from the given path (which is not ideal)
No vulnerabilities found
```

Latest grype 0.79.3 reports nothing after these changes, however `wolfictl scan` with same grype in the dependencies reports a false positive:

```
$ wolfictl scan packages/x86_64/argo-workflows-ui-3.5.8-r3.apk 
🔎 Scanning "packages/x86_64/argo-workflows-ui-3.5.8-r3.apk"
└── 📄 /usr/lib/argo-workflows-ui/ui/node_modules/resolve/test/resolver/multirepo/package.json
        📦 monorepo-symlink-test 0.0.0 (npm)
            Critical GHSA-2jcg-qqmg-46q6
```

That match is bogus. https://github.com/browserify/resolve/issues/303

xnox:
@luhring any idea about the clean result from grype, yet a false positive from wolfictl scan?

luhring:
> @luhring any idea about the clean result from grype, yet a false positive from wolfictl scan?

`wolfictl scan` is using Grype, but the `grype` you're running is invoking a directory scan, which is looking for software differently. More info on that here: https://github.com/anchore/syft/wiki/package-cataloger-selection#concepts. If you were to run Grype on an image with this package, my expectation is you'd find the same false positive. And `wolfictl scan` is designed to anticipate results from Grype scanning an image.

We can submit a PR upstream to Syft/Grype if we think their logic should change. But we should also mark this as a false positive in our advisory data for now.

imjasonh (dismissed):

imjasonh (approved):

PR #25161, opened 2024-07-30
Add renovate test.

smoser:
renovate execution will fail if nodejs version differs from its expected/supported version.
The added test would fail if nodejs were set to 'nodejs'.

The supported node version value comes (I think) from package.json https://github.com/renovatebot/renovate/blob/8b88d9790888c60c4029249e1f554f45f6d44bb2/package.json#L136

It ends up failing with an error like:

    Unsupported node environment detected. Please update your node version.

### DIFF
--- a/renovate.yaml
+++ b/renovate.yaml
@@ -1,6 +1,6 @@
 package:
   name: renovate
-  version: 38.10.0
+  version: 38.11.1
   epoch: 0
   description: "Automated dependency updates. Multi-platform and multi-language."
   copyright:
--- a/renovate.yaml
+++ b/renovate.yaml
@@ -1,7 +1,7 @@
 package:
   name: renovate
   version: 38.10.0
-  epoch: 0
+  epoch: 1
   description: "Automated dependency updates. Multi-platform and multi-language."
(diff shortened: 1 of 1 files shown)


philroche:
@smoser How did you notice such a failure/error.?Since renovate wolfi package 38.10.0 there is a dependency on nodejs-20 which would prevent a scenario of anything other than node 20 being present on the system?

+1 on the test though as it will catch any future error when upstream change their dependencies.

smoser:
> @smoser How did you notice such a failure/error.?Since renovate wolfi package 38.10.0 there is a dependency on nodejs-20 which would prevent a scenario of anything other than node 20 being present on the system?
> 
> +1 on the test though as it will catch any future error when upstream change their dependencies.

The image build was failing it's test ([here](https://github.com/chainguard-dev/image-release-stats/issues/259)) before this had the nodejs-20 dep (which @ajayk added to fix https://github.com/wolfi-dev/os/pull/25035).

philroche (approved):

ajayk (approved):

PR #25169, opened 2024-07-30
gcc 13 openssf integration

xnox:
- **gcc: use major version only in filesystem paths**
  Only major versions are co-installable, and this make specs file
  location predictable
  

- **gcc: fixup major-version only libdir**
  

- **Bump gcc**

This changes paths like these:

    usr/libexec/gcc/x86_64-pc-linux-gnu/13.3.0/cc1

to

    usr/libexec/gcc/x86_64-pc-linux-gnu/13/cc1
  
Like it is in all other distributions, this makes location of specs files predictable within a given gcc release series.

### DIFF
--- a/gcc.yaml
+++ b/gcc.yaml
@@ -1,7 +1,7 @@
 package:
   name: gcc
   version: 13.3.0
-  epoch: 1
+  epoch: 2
   description: "the GNU compiler collection"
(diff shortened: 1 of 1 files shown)


pnasrat (approved):
LGTM

PR #25180, opened 2024-07-30
Additional gems required for Custom fluentd-fips image

Aditevil:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches


ruby3.2-ffi.yaml
```diff
@@ -0,0 +1,55 @@
+package:
+  name: ruby3.2-ffi
+  version: 1.17.0
+  epoch: 0
+  description: Ruby FFI
+  copyright:
+    - license: BSD-3-Clause And MIT
+  dependencies:
+    runtime:
```
EyeCantCU, line 9: These runtime dependencies don't look correct to me. They should just be `ruby-3.2` and `libffi`, both should be added via SCA. We could remove the entire `runtime` section here

ruby3.2-ffi.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - libffi-dev
+      - libtool
+      - ruby-3.2
+      - ruby3.2-bundler
+
+environment:
+  contents:
+    packages:
+      - autoconf
+      - automake
```
EyeCantCU, line 21: `automake` and `autoconf` are only needed if used in the build process. We should be able to get rid of these

ruby3.2-ffi.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - git
```
EyeCantCU, line 25: `git` is added via our `git-checkout` pipeline

ruby3.2-ffi.yaml
```diff
@@ -0,0 +1,55 @@
...
+    with:
+      expected-commit: ecfb225096ae76ba2a5e8115f046bd0ac23095e6
+      repository: https://github.com/ffi/ffi.git
+      tag: v${{package.version}}
+
+  - runs: |
+      git submodule update --init --recursive
+      bundle install
+      rake compile -- --enable-system-libffi
+      gem install ffi
```
EyeCantCU, line 46: This will install the gem from the repository and not the one built locally. You'd pass something like `./ffi-${{package.version}}`

fluent-plugin-systemd.yaml
```diff
@@ -0,0 +1,50 @@
...
+  name: fluent-plugin-systemd
+  version: 1.0.5
+  epoch: 0
+  description: This is a fluentd input plugin. It reads logs from the systemd journal.
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - ruby-3.2
+      - ruby3.2-fluentd
```
EyeCantCU, line 11: I see we're adding the fluentd gem to a few packages at runtime. Are we should it's needed by these individual gems?

fluent-plugin-systemd.yaml
```diff
@@ -0,0 +1,50 @@
+package:
+  name: fluent-plugin-systemd
+  version: 1.0.5
+  epoch: 0
+  description: This is a fluentd input plugin. It reads logs from the systemd journal.
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - ruby-3.2
```
EyeCantCU, line 10: `ruby-3.2` will be picked up by SCA

fluent-plugin-systemd.yaml
```diff
@@ -0,0 +1,50 @@
...
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - git
+      - ruby-3.2
+      - ruby-3.2-dev
+      - ruby3.2-fluentd
```
EyeCantCU, line 22: Are we sure this at buildtime?

fluent-plugin-concat.yaml
```diff
@@ -0,0 +1,50 @@
...
+  name: fluent-plugin-concat
+  version: 2.5.0
+  epoch: 0
+  description: Fluentd Filter plugin to concatenate multiline log separated in multiple events.
+  copyright:
+    - license: MIT
+  dependencies:
+    runtime:
+      - ruby-3.2
+      - ruby3.2-fluentd
```
EyeCantCU, line 11: See comments above

fluent-plugin-concat.yaml
```diff
@@ -0,0 +1,50 @@
...
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - git
+      - ruby-3.2
+      - ruby-3.2-dev
+      - ruby3.2-fluentd
```
EyeCantCU, line 22: See comments above

fluent-plugin-concat.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - git
+      - ruby-3.2
```
EyeCantCU, line 20: `ruby-3.2` is a transitive dependency of `ruby-3.2-dev`, we can remove it here

fluent-plugin-concat.yaml
```diff
@@ -0,0 +1,50 @@
...
+    - license: MIT
+  dependencies:
+    runtime:
+      - ruby-3.2
+      - ruby3.2-fluentd
+
+environment:
+  contents:
+    packages:
+      - build-base
```
EyeCantCU, line 16: We may or may not need `build-base`

EyeCantCU (dismissed):

fluent-plugin-systemd.yaml
Aditvil-Dev: Yes

fluent-plugin-systemd.yaml
Aditvil-Dev: Yes

EyeCantCU:
Also, is there a reason we're opting to use Ruby 3.2 here when 3.3. is latest?

ruby3.2-ffi.yaml
Aditvil-Dev: Turns out the gemspec is fetching version dynamically. So passing `version-tag` returns a `Not found` error.

[sdk] ❯ gem install ffi
Fetching ffi-1.17.0-aarch64-linux-gnu.gem
Successfully installed ffi-1.17.0-aarch64-linux-gnu
Ignoring debug-1.7.1 because its extensions are not built. Try: gem pristine debug --version 1.7.1
Parsing documentation for ffi-1.17.0-aarch64-linux-gnu
Installing ri documentation for ffi-1.17.0-aarch64-linux-gnu
Done installing documentation for ffi after 0 seconds
1 gem installed

EyeCantCU: I think we need to pass `--local` here. From the log output, it appears to be pulling it down and not using the one we've built

Aditvil-Dev: That worked.

[sdk] ❯ gem install ffi -- --local
Successfully installed ffi-1.17.0-aarch64-linux-gnu
Ignoring debug-1.7.1 because its extensions are not built. Try: gem pristine debug --version 1.7.1
Parsing documentation for ffi-1.17.0-aarch64-linux-gnu
Done installing documentation for ffi after 0 seconds
1 gem installed

jamonation (changes requested):
I'd like to see some tests here that validate the gems can be imported, similar to what we do with our python packages.

For example, this would work in a test pipeline to import the `fluent-plugin-prometheus` gem if it is installed:
 
```
ruby -W0 -e "require 'fluent/plugin/prometheus'"
```

Could you add similar tests for these as well?

jamonation (approved):
Tests look good, thanks for adding those!

PR #25197, opened 2024-07-30
xorg-xserver: switch from gcrypt to openssl

xnox:
This will reduce number of cryptographic libraries in some images
which already have openssl, as most software uses openssl.


mamccorm (approved):

mamccorm:
Checked with @xnox via 1-1, this one is ready to review

PR #25243, opened 2024-07-31
Removing wasi-sdk from wolfi

debasishbsws:


debasishbsws:
Okay my mistake, looks like I missed a transitive dependency here.

we have image Haproxy depend on firefox-esr depends on wasi-sysroot(subpackage of wasi-sdk)

PR #25356, opened 2024-07-31
Add compiler-hardening-checks to compilers

xnox:
Can only be done once openssf-compiler-options exists in the archive.


pnasrat (dismissed):
LGTM

Feel free to merge once the prior PR is in

xnox:
@pnasrat rebuilds triggerred linking failures as gcc-13 and gcc-12 libstd++ now do not have compatible headers. Added runtime dep on libstdc++-12-dev to pull in libstdc++-12 such that one can correctly link -12 headers with -12 .so.

It still has runtime forward compatibility.

imjasonh (approved):

PR #25373, opened 2024-07-31
skopeo: use go pipeline and cgr registry

Dentrax:


skopeo.yaml
```diff
@@ -23,11 +23,13 @@ pipeline:
...
-  - runs: |
-      CGO_ENABLED=0 go build -mod=vendor \
-        -tags "exclude_graphdriver_devicemapper exclude_graphdriver_btrfs containers_image_openpgp" \
-        -trimpath -ldflags "-s -w -extldflags -static -buildid=" \
-        -o "${{targets.destdir}}/usr/bin/skopeo" ./cmd/skopeo
+  - uses: go/build
+    with:
+      packages: ./cmd/skopeo
+      output: skopeo
+      ldflags: "-s -w -extldflags -static -buildid="
```
ajayk, line 30: we don't need to strip the symbols now we can remove the -s flag

ajayk (approved):

PR #25420, opened 2024-08-01
disable the webdav module in nginx per hardening guidelines

joshrwolf:
Per various hardening guides, building `nginx` without `webdav` is recommended: https://nginx.org/en/docs/http/ngx_http_dav_module.html

- https://github.com/kubernetes/ingress-nginx/blob/c6e86c86dc0a007e3bda34ee6eee9cabed34eb2c/docs/deploy/hardening-guide.md?plain=1#L39
- https://www.stigviewer.com/stig/web_server_security_requirements_guide/2023-09-13/finding/V-206383

### DIFF
--- a/nginx-mainline.yaml
+++ b/nginx-mainline.yaml
@@ -1,7 +1,7 @@
 package:
   name: nginx-mainline
   version: 1.27.0
-  epoch: 6
+  epoch: 7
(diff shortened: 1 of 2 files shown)


pnasrat (approved):
LGTM

PR #25519, opened 2024-08-02
spark: add tests and missing symlinks

Dentrax:
This PR expands test cases for Apache Spark. And also adds missing symlinks for core binaries.


mamccorm (approved):
thanks for expanding the tests!

PR #25531, opened 2024-08-02
bash: revert git-checkout and improve version test

ajayk:


mattmoor (approved):
Any sense for why this didn't work?

ajayk:
> Any sense for why this didn't work?

This was not considering the patch versions , the git branch doesn't seem to have any patches applied

PR #25604, opened 2024-08-05
spark: omit openjdk from runtime deps

Dentrax:
This PR omits `openjdk-17` dependency on the runtime, to allow to use with both 11 and 17 versions.

Upstream image built with JDK8.

And since they [support](https://github.com/apache/spark/blob/6bf6088e7719eccf050910d6ae495b369afb25be/python/docs/source/getting_started/install.rst#python-versions-supported) Python versions starting from > 3.9, the related Python version will be added in the image.

### DIFF
--- a/spark-3.5.yaml
+++ b/spark-3.5.yaml
@@ -1,18 +1,20 @@
 package:
   name: spark-3.5
   version: 3.5.1
-  epoch: 7
+  epoch: 8
   description: Unified engine for large-scale data analytics
   copyright:
     - license: Apache-2.0
   dependencies:
     provides:
       - spark=${{package.full-version}}
+    # Both `python` and `openjdk-{11,17}-default-jvm` are required for running Spark,
+    # but will be added in the image, since upstream supports different Python versions
+    # and Java versions. So we don't need to specify them here to avoid conflicts.
     runtime:
+      - R
       - bash
       - busybox
-      - openjdk-17
-      - openjdk-17-default-jvm
       - procps
 
 environment:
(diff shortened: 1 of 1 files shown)


spark-3.5.yaml
```diff
@@ -128,8 +130,9 @@ test:
   environment:
     contents:
       packages:
-        - R
         - python3
+        - openjdk-17
```
EyeCantCU, line 134: We don't need the full JDK for the test. Just our JVM package

spark-3.5.yaml
```diff
@@ -1,18 +1,20 @@
...
     provides:
       - spark=${{package.full-version}}
     runtime:
+      - R
+      # - python3: Python dependency will be added in the image, since upstream supports different Python versions starting from 3.9.
       - bash
       - busybox
-      - openjdk-17
-      - openjdk-17-default-jvm
+      # - openjdk-17: OpenJDK is required for running Spark, but will be added in the image, since upstream supports different Java versions.
```
EyeCantCU, line 16: nit: we should remove these versus commenting them out

EyeCantCU (changes requested):

spark-3.5.yaml
Dentrax: I was going to remove it but then thought that someone may bring those back in the future, so wanted to drop as a warning actually. Should we remove those anyway?

EyeCantCU: We still wouldn't want to add back the full JDK though. I'd :axe: it

Dentrax: Done. Removed all commented out packages but feel a bit uncomfortable actually, i'd be nice if we should drop a comment regarding to why we shouldn't have `python` and `openjdk-` package in the runtime deps? WDYT?

EyeCantCU: That could certainly be helpful for anyone looking at this in the future. I'd suspect we'll have more java packages where this is the case

mritunjaysharma394 (approved):
lgtm

EyeCantCU (approved):

PR #25640, opened 2024-08-05
fix: add nanohttp suppport

mesaglio:
libxml2 make http support disable by default, but some other libs not make this change yet.
For example, `libgsf` need `xmlNanoHTTPOpen` symbol on `libxml2.so`
```
symbol lookup error: /usr/lib/libgsf-1.so.114: undefined symbol: xmlNanoHTTPOpen, version LIBXML2_2.4.30
```
With this flag we generate this simbols
```
# old compilation
[sdk] ❯ nm -gDC usr/lib/libxml2.so.2.13.3 | grep xmlNanoHTTPOpenen
[sdk] ❯
# new compilation
[sdk] ❯ nm -gDC usr/lib/libxml2.so.2.13.3-1 | grep xmlNanoHTTPOpen
0000000000095260 T xmlNanoHTTPOpen@@LIBXML2_2.4.30
0000000000095200 T xmlNanoHTTPOpenRedir@@LIBXML2_2.4.30
[sdk] ❯
```

Ref: https://github.com/GNOME/libxml2/commit/3018842c07e9b88d6bb0257f5644e7695cdeb2e2

cc @lpcalisi @icasarino

PR #25652, opened 2024-08-06
libgsf: fix FTBFS against new libxml2

xnox:
Fix failure to build from source against new libxml2 without nanohttp support.

This is alternative to:
- #25640

### DIFF
--- a/libgsf.yaml
+++ b/libgsf.yaml
@@ -46,8 +46,9 @@ pipeline:
   - runs: |
       git fetch origin master:master
       # Fix failure to build from source against new libxml2, without
-      # nanohttp support. Ignore merge conflict in NEWS.
-      git cherry-pick -x 5d4bb55095d3d6ef793c1908a88504183e28644c || true
+      # nanohttp support.
+      git show 5d4bb55095d3d6ef793c1908a88504183e28644c -- configure.ac gsf/gsf-input-http.c > partial-cherry-pick.patch
+      git apply partial-cherry-pick.patch
 
   - runs: |
       sed -i -e 's/gsf-gnome//' -e 's/thumbnailer//' Makefile.*
--- a/libgsf.yaml
+++ b/libgsf.yaml
@@ -2,7 +2,7 @@
 package:
   name: libgsf
   version: 1.14.52
-  epoch: 1
+  epoch: 2
   description: Utility library for reading and writing structured file formats
   copyright:
     - license: LGPL-2.1-only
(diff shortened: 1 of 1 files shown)


libgsf.yaml
```diff
@@ -43,6 +43,12 @@ pipeline:
       expected-commit: ea9d8cd1369661bf62d0476474700a9e0887c812
       tag: LIBGSF_${{vars.mangled-package-version}}
 
+  - runs: |
+      git fetch origin master:master
+      # Fix failure to build from source against new libxml2, without
+      # nanohttp support. Ignore merge conflict in NEWS.
+      git cherry-pick -x 5d4bb55095d3d6ef793c1908a88504183e28644c || true
```
mamccorm, line 50: Would we be better carrying this as a custom patch? otherwise when upstream do cut a new release with the fix, we're ignoring any errors so it may be harder to debug later failures

libgsf.yaml
```diff
@@ -43,6 +43,12 @@ pipeline:
...
+
```
smoser, line 51: the `|| true` could fail in any way.  Unlikely but worst case is leaving buggy code that compiles.

I think I'd prefer a hand edited patch to remove the NEWS.  

Another option is https://github.com/wolfi-dev/os/pull/25657 paired with something like:
```
    git format-patch --stdout -1 5d4bb55095d3d6ef793c1908a88504183e28644c > cp.diff
    filterdiff -p1 --exclude=NEWS < cp.diff > filtered.diff
    git apply filtered.diff
    rm -f cp.diff filtered.diff
```

I don't love it. there is 
https://stackoverflow.com/questions/5717026/how-to-cherry-pick-only-changes-to-certain-files
which mentions `git show SHA -- file1 file2`, but that is include rather than exclude. :cry:


another option that I think I like the best

```
# Fix failure to build from source against new libxml2, without
# nanohttp support. Ignore merge conflict in NEWS.
git cherry-pick -x 5d4bb55095d3d6ef793c1908a88504183e28644c || true
git checkout HEAD NEWS
git cherry-pick --continue --no-edit
```

The improvement here is that the cherry-pick `--continue` will fail if there were other changes that were not cleanly applied.

If the NEWs issue is a repeating pattern, we could add support in some way for that into checkout's cherry-picks.

smoser (changes requested):

smoser (approved):
I wish there was a better way to do this, but this is good enough.

The key point being
1. we're not ignoring any errors
2. The 'apply' will (most likely) fail when upstream has a release with that commit in it 

(there are some paths where `apply` could pass, if upstream backed out the change and somehow avoided conflicting with this patch in their fix, but... hopefully it would fail to compile in that case).

PR #25657, opened 2024-08-06
Add patchutils package.

smoser:
My interest in patchutils is currently for 'filterdiff', which would be useful for cherry-picking patches that edit a file that many changes edit (ie, a NEWs file or changelog).

I'm suggesting the addition of patchutils for a need seen in https://github.com/wolfi-dev/os/pull/25652/files


smoser:
Closing this. Maybe someday we can revive if it is needed.

PR #25702, opened 2024-08-06
set exclude-reasons for update-disabled packages

imjasonh:


rawlingsj (approved):

imjasonh:
I'm going to force-merge this since it doesn't actually affect any package contents, and it'll take forever to build all these rust packages.

jonjohnsonjr (approved):

PR #25715, opened 2024-08-06
add ggshield tests

joshrwolf:
#13623

### DIFF
--- a/ggshield.yaml
+++ b/ggshield.yaml
@@ -1,12 +1,14 @@
 package:
   name: ggshield
   version: 1.30.2
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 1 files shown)


ggshield.yaml
```diff
@@ -1,12 +1,14 @@
...
   name: ggshield
   version: 1.30.2
-  epoch: 0
+  epoch: 1
   description: Find and fix 360+ types of hardcoded secrets and 70+ types of infrastructure-as-code misconfigurations.
   copyright:
     - license: MIT
   dependencies:
     runtime:
+      - git
```
mamccorm, line 10: @joshrwolf looks like new runtime deps were introduced here - wad this intentional? Any context?

joshrwolf: the tests I added failed unless `git` was installed, so I added it to the runtime

mamccorm (approved):

PR #25873, opened 2024-08-08
zlib openssf

xnox:
- **OpenSSF hardening of zlib**
  Add test case to check hardening, add openssf-compiler-options as a
  build dependency. Fix configure script, to use native gcc for build.
  
  This enables Control Flow integrity for zlib.

### DIFF
--- a/zlib.yaml
+++ b/zlib.yaml
@@ -1,7 +1,7 @@
 package:
   name: zlib
   version: 1.3.1
-  epoch: 3
+  epoch: 4
   description: "a library implementing the zlib compression algorithms"
   copyright:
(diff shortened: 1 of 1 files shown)


zlib.yaml
```diff
@@ -93,6 +100,16 @@ subpackages:
...
 
+test:
+  environment:
+    contents:
+      packages:
+        - minizip
+  pipeline:
+    - uses: test/hardening-check
+      with:
+        package-match: "zlib\\|minizip"
```
smoser, line 111: I really hate this. \\

smoser (approved):
Looks good. some day i will fix the regex escape thing.

Issue #23356, opened 2024-07-08
FTBFS: mariadb-connector-c


imjasonh:
```
running step "Split development files"
mv: can't rename '/home/build/melange-out/mariadb-connector-c/usr/bin': No such file or directory
```

https://cloudlogging.app.goo.gl/DejLp4yvTH89c85j8

OddBloke:
This was fixed in #25919

PR #25919, opened 2024-08-08
mariadb-connector-c: Fix build issues (wolfi-dev#23356)

justinvreeland:
The run stage for mariadb-connector-c-dev tries to copy files that have already been copied via the split/dev pipline.

Fixes:

FTBFS: mariadb-connector-c
[#23356](https://github.com/wolfi-dev/os/issues/23356)

Output once built shows the bin contents in the dev package as expected.

```
tar tvf packages/aarch64/mariadb-connector-c-dev-3.3.8-r0.apk | grep bin
drwxr-xr-x  0 justin dialout     0 Aug  8 14:47 usr/bin
-rwxr-xr-x  0 justin dialout 71576 Aug  8 14:47 usr/bin/mariadb_config
lrwxr-xr-x  0 justin dialout     0 Aug  8 14:47 usr/bin/mysql_config -> mariadb_config
```

### DIFF
--- a/mariadb-connector-c.yaml
+++ b/mariadb-connector-c.yaml
@@ -1,7 +1,7 @@
 package:
   name: mariadb-connector-c
   version: 3.3.8
-  epoch: 0
+  epoch: 1
   description: The MariaDB Native Client library (C driver)
   copyright:
     - license: LGPL-2.1-or-later
@@ -61,9 +61,6 @@ subpackages:
   - name: mariadb-connector-c-dev
     pipeline:
       - uses: split/dev
-      - runs: |
-          mkdir -p "${{targets.subpkgdir}}"/usr/
-          mv "${{targets.destdir}}"/usr/bin "${{targets.subpkgdir}}"/usr/
     dependencies:
       replaces:
         - mariadb-dev


xnox (changes requested):
For this change to actually ship, epoch bump is needed to trigger package rebuild in production.

You can use `wolfictl bump mariadb-connector-c` to achieve that, or with .yaml suffix, or edit epoch: N by hand.

dakaneye:
the check update appears to have failed (which seems common based on the message):
> 2024/08/08 22:01:43 ERRO package mariadb-connector-c: update found newer version 3.4.0 compared with package.version 3.3.8 in melange config

Go ahead and just iterate the version and reset the epoch and it should pass.

xnox:
> the check update appears to have failed (which seems common based on the message):
> 
> > 2024/08/08 22:01:43 ERRO package mariadb-connector-c: update found newer version 3.4.0 compared with package.version 3.3.8 in melange config
> 
> Go ahead and just iterate the version and reset the epoch and it should pass.

But one can also skip the version update, as it is not required to land. And once this pr is fixed the version update PR will be rebased and likely auto-approved. As currently it is failing to build from source because of this bug.

xnox (approved):

PR #26005, opened 2024-08-10
fix(kuma-2.8): Remove Envoy version requirement

EyeCantCU:
This check validates down to the patch level that the version of Envoy provided at buildtime is inline with the version at runtime. This would mean rebuilding Kuma every time Envoy updates. Given there really isn't a reason to rebuild Kuma otherwise, just patch out the check for now. An issue has been filed upstream suggesting to only validate the major and minor version of Envoy.

https://github.com/kumahq/kuma/issues/11086

### DIFF
--- a/kuma-2.8.yaml
+++ b/kuma-2.8.yaml
@@ -1,7 +1,7 @@
 package:
   name: kuma-2.8
   version: 2.8.2
-  epoch: 2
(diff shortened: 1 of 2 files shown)


mamccorm (approved):
Linked the upstream issue in the description

PR #26092, opened 2024-08-10
keycloak: bitnami: upgrade to 25 and fix data path

Dentrax:
* Dynamically parse major version to pass `bitnami/compat` pipeline
* Create a missing `data` dir
* Add `coreutils`

### DIFF
--- a/keycloak.yaml
+++ b/keycloak.yaml
@@ -1,7 +1,7 @@
 package:
   name: keycloak
   version: 25.0.2
-  epoch: 2
+  epoch: 3
   description: Open Source Identity and Access Management For Modern Applications and Services
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


keycloak.yaml
```diff
@@ -1,15 +1,25 @@
...
   version: 25.0.2
-  epoch: 2
+  epoch: 3
   description: Open Source Identity and Access Management For Modern Applications and Services
   copyright:
     - license: Apache-2.0
   dependencies:
     runtime:
       - bash # Keycloak helper scripts require bash, aren't compatible with busybox.
+      - coreutils # Keycloak Helm Chart scripts require coreutils, aren't compatible with busybox. (i.e., `cp` with `--preserve` option)
```
mamccorm, line 11: Its the bitnami helm charts that need coreutils if i'm following? Can we add this into the bitnami compat package as a runtime dependency if its not needed for non-bitnami keycloak?

Dentrax: Good catch, thanks. It seems upstream have made some changes to their Helm Chart a few days ago, so starting from `v22.0.0`, will require `coreutils`: https://github.com/bitnami/charts/commit/80bc148b1af4e75ff8619c68f20ec2be90015a37

Runtime dependency moved under bitnami compat package.

mamccorm (approved):

PR #26100, opened 2024-08-11
c ares cmake

xnox:
- **c-ares-bootstrap: provide c-ares autotools based build as bootstrap**
  c-ares is in the bootstrap cycle, thus provide a zero priority
  autotools build of c-ares without using cmake.
  
  The fixes wolfictl text build-dependency cycle:
  
      nghttp2:1.62.1-r1@local -> c-ares-dev:1.33.0-r1@local, caused by: c-ares-dev:1.33.0-r1@local -> c-ares:1.33.0-r1@local -> cmake:3.29.0-r0@local -> nghttp2-dev:1.62.1-r1@local -> nghttp2:1.62.1-r1@local
  

- **c-ares: switch from autotools to cmake**
  Build c-ares with cmake instead of autotools. This installs additional
  .cmake files in the -dev package, as needed to build other cmake-based
  projects that depend on c-ares (e.g. bun).

With this change, both apk add c-ares-dev and via apko/melange, results in

```console
# ls /usr/lib/cmake/c-ares/c-ares-*
/usr/lib/cmake/c-ares/c-ares-config-version.cmake      /usr/lib/cmake/c-ares/c-ares-targets-minsizerel.cmake
/usr/lib/cmake/c-ares/c-ares-config.cmake              /usr/lib/cmake/c-ares/c-ares-targets.cmake
```

being present.

### DIFF
--- a/c-ares-bootstrap.yaml
+++ b/c-ares-bootstrap.yaml
@@ -1,0 +1,53 @@
+package:
+  name: c-ares-bootstrap
+  version: 1.33.0
+  epoch: 1
+  description: "an asynchronous DNS resolution library"
+  copyright:
+    - license: MIT
+  dependencies:
+    runtime:
+      - libev
+    provider-priority: 0
+    provides:
+      - c-ares=${{package.full-version}}
+      - c-ares-dev=${{package.full-version}}
+
+environment:
+  contents:
+    packages:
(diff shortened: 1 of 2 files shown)


joshrwolf (approved):
til!

Issue #25933, opened 2024-08-09
Package Update: freerdp-3


debasishbsws:
Latest version of freerdp-3 is `3.9.0` and last version we build was `3.5.1`.

There are a lot new package required to build new versions.
creating an Issue to better organise the process.
Current active automated PR: https://github.com/wolfi-dev/os/pull/31386
Previous automated PR: https://github.com/wolfi-dev/os/pull/23318

debasishbsws:
https://github.com/wolfi-dev/os/pull/26122 Need review adding libsdl

debasishbsws:
It seems with cmake options `-DWITH_NEON=OFF -DWITH_WEBVIEW=OFF` we do not need the webkit2gkt package anymore as we are not building the webview.

debasishbsws:
### Background:
The FreeRDP-3 package update has been stalled for some time, with this issue initially created to track progress and manage superseding PRs.

### Previous Escalation:
This was previously escalated to the foundation in [PR #27343](https://github.com/wolfi-dev/os/pull/27343), which has since been superseded by [PR #31386](https://github.com/wolfi-dev/os/pull/31386).

### Progress & Dependencies:
During this update, we've noted that multiple new packages are required. Some dependencies added include:
- [libsdl2](https://github.com/wolfi-dev/os/pull/26122)
- [libsdl3](https://github.com/wolfi-dev/os/pull/30481)
- [gtk-4](https://github.com/wolfi-dev/os/pull/26125)

Additionally, there was an attempt to add [webkit2gtk](https://github.com/wolfi-dev/os/pull/26129), but it was excluded after deciding not to build the webview, by adding `-DWITH_NEON=OFF -DWITH_WEBVIEW=OFF` flags.

### Current Challenge:
The latest [comment from @xnox](https://github.com/wolfi-dev/os/pull/27343#issuecomment-2406138736) points out a key next step:
> Next todo - figure out how our libsdl2 and libsdl3 builds are incomplete and what are the _ttf (TrueType font) variant builds that FreeRDP requires and are currently missing or packaged as symlinks.

debasishbsws:
This is complete by the latest update https://github.com/wolfi-dev/os/pull/31386/files

PR #26125, opened 2024-08-12
gtk-4: new package

debasishbsws:
This package is require to build the freerdp-3 package.

Related: https://github.com/wolfi-dev/os/issues/25933

#### For new package PRs only

### DIFF
--- a/gtk-4.yaml
+++ b/gtk-4.yaml
@@ -1,0 +1,149 @@
+package:
+  name: gtk-4
+  version: 4.15.5
+  epoch: 0
+  description: The GTK+ Toolkit (v4)
+  copyright:
+    - license: LGPL-2.1-or-later
+
+environment:
+  contents:
+    packages:
+      - at-spi2-core-dev
+      - autoconf
+      - automake
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - cairo-dev
+      - cmake
+      - cups-dev
+      - dbus-dev
+      - expat-dev
+      - fontconfig-dev
+      - fribidi-dev
+      - gdk-pixbuf-dev
+      - gettext-dev
+      - gi-docgen
+      - git
+      - glib-dev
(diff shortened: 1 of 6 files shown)


gtk-4.yaml
```diff
@@ -0,0 +1,149 @@
...
+      - pango-dev
+      - py3-docutils
+      - py3-gi-docgen
+      - py3-packaging
+      - sassc
+      - shaderc
+      - shared-mime-info
+      - smartypants
+      - spirv-headers
+      # TODO: right now shaderc also inclused spirv-tools, causin a conflict need to fix that,
```
mamccorm, line 72: Not sure i understand this comment - mentions we use our own spirv-tools, but then we've it commented out below, is this still valid?

gst-plugins-base.yaml
```diff
@@ -1,7 +1,7 @@
 package:
```
mamccorm, line 1: Any context for the changes to this existing package? was there build issues before or were we skipping some components?

gtk-4.yaml
debasishbsws: Yes, I have created a new package called shaderc which now also include the spirv-tools within it. I will create a Issue to track this and make sure we can use shaderc with our spirv-tools.

gst-plugins-base.yaml
debasishbsws: Yes, the gl plugins are needed for the gtk-4. Previously we were not including it but now we are.

gtk-4.yaml
debasishbsws: https://github.com/wolfi-dev/os/pull/26125/files#diff-95912073ffd0aac4ceb05bf2c6ee2eb41fa14676a714207ad5f9970ccb2ac557R25 It needs to be done.

mamccorm (approved):

PR #26173, opened 2024-08-12
fix(neuvector-manager): Disable version mismatch notification

EyeCantCU:
NeuVector appears to chcek a few places to verify the version of the component thats running. One of those places being image labels. As we do not set image labels, disable the version mismatch notification as it is misleading when labels are not set


jamonation:
Just disabling the UI notification for a version mismatch (that will never match with wolfi packages anyways) so +1 here.

jamonation (approved):

Issue #23359, opened 2024-07-08
FTBFS: riemann-c-client


imjasonh:
```
running step "Run autoconf make"
make: Entering directory '/home/build'
/bin/sh /home/build/install-sh -d ./lib/riemann/proto
/bin/sh /home/build/install-sh -d ./lib/riemann/proto
protoc-c lib/riemann/proto/riemann.proto -I./lib/riemann/proto --c_out=./lib/riemann/proto
protoc-c lib/riemann/proto/riemann.proto -I./lib/riemann/proto --c_out=./lib/riemann/proto
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
E0000 00:00:1720297553.245814       7 command_line_interface.cc:1552] Built-in generator --c_out specifies a maximum edition PROTO3 which is not the protoc maximum 2023.
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
E0000 00:00:1720297553.245915       8 command_line_interface.cc:1552] Built-in generator --c_out specifies a maximum edition PROTO3 which is not the protoc maximum 2023.
make: *** [Makefile:1623: lib/riemann/proto/riemann.pb-c.c] Error 1
make: *** Waiting for unfinished jobs....
make: *** [Makefile:1623: lib/riemann/proto/riemann.pb-c.h] Error 1
make: Leaving directory '/home/build'
ERROR: failed to build package. the build environment has been preserved:
```

https://cloudlogging.app.goo.gl/obpWxKr3T8L96ivX6

justinvreeland:
I have a branch with a fix that involves removing one of 3 upstream patches we've added. I believe Archlinux also isn't using that patch in their patchiest for the problem. I don't know if not having that patch will cause other problems with projects that need protobuf-c so we'll want to do a build with those to verify things.

PR #26208, opened 2024-08-12
protobuf-c: Fix riemann-c-client build (wolfi-dev/#23359)

justinvreeland:
It looks like we pulled in an unmerged patches [1] from which is now causing build issues with riemann-c-client. Once the patch causing issues is removed riemann-c-client builds cleanly again.

https://github.com/protobuf-c/protobuf-c/pull/711

Fixes:
https://github.com/wolfi-dev/os/issues/23359

PR:

I just downloaded the patch from the GitHub PR request instead of applying it and using format patch so some of the top matter is gone now. I can restore it if that's an issue.

Notes:

We should validate this doesn't break existing builds before merging.

### DIFF
--- a/protobuf-c.yaml
+++ b/protobuf-c.yaml
@@ -1,7 +1,7 @@
 package:
   name: protobuf-c
   version: 1.5.0
-  epoch: 7
+  epoch: 8
   description: Protocol Buffers implementation in C
   copyright:
     - license: BSD-2-Clause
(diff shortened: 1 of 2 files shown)


justinvreeland:
Rebuilt the packages that depend on protobuf-c and don't see any issues.

mamccorm (approved):

PR #26220, opened 2024-08-13
Bump vendored dependency(urllib3) for pip 3.10, 3.11 and 3.12

srishtih:
Our pip packages (versions 3.10, 3.11, and 3.12) include several vendored dependencies, one of which is urllib3, currently pinned to version 1.26.18. We are now seeing vuln stemming from here:
[sdk] ❯ wolfictl scan packages/aarch64/sample-0.5.4-r2.apk
🔍 Scanning "packages/aarch64/sample-0.5.4-r2.apk"
└── 📄 /usr/share/vllm/.venv/lib/python3.10/site-packages/pip/_vendor/vendor.txt
    📦 urllib3 1.26.18 (python)
      Medium CVE-2024-37891 GHSA-34jh-p97f-mpxf fixed in 1.26.19 

#### For PRs that add patches

Added for CVE remdiation


py3-pip.yaml
```diff
@@ -32,6 +32,17 @@ pipeline:
...
+    with:
+      patches: bump-urllib3.patch
+
+  - runs: |
+      rm -rf ./src/pip/_vendor/urllib3
+      git clone https://github.com/urllib3/urllib3.git /home/urllib3
+      cd /home/urllib3
+      git checkout d9d85c88aa644af56d5e129634e750ce76e1a765
+      mv  ./src/urllib3 ../build/src/pip/_vendor/
+
```
pnasrat, line 45: you'll need to rerun the vendoring. I suggest you do that and use the patch from that here as we don't have nox installed

pnasrat (changes requested):

pnasrat (approved):
LGTM

PR #26255, opened 2024-08-13
airflow: bump epoch to see if it remediates GHSA-62qf-qm3g-fvcw

rawlingsj:


hectorj2f:
This won't fix the CVE, it needs a new release `v2.10.0`.

PR #26270, opened 2024-08-13
py3-parallel-ssh: remove package & deps (wolfi-dev#23756)

justinvreeland:
In b7de8cbd0d5c10a730d97f82e3301df74e936a90 py3-ssh-python py3-ssh2-python and py3-parallel-ssh were imported for use in py3-cassandra-medusa.

Sice then py3-cassandra-medusa switched to a venv
49c972383c24ffdef5cc4dddf9de10d8d78c478a. These packages are not used or useable in their current state and upstream is not in a position to be releasing python3.12 compatible versions right now.

Fixes:

Related:

### Pre-review Checklist

There's no Pre-review checklist for removing things but I see no packages or images that specify they need these contents and the original need for them is gone.

@mritunjaysharma394 can you confirm that we don't need these anymore?


mritunjaysharma394:
Thanks @justinvreeland , i checked and yes since we switched to venv method, I don't think we need them anymore, I also checked if any images uses them and seems like none use it directly so we should be safe to remove it

smoser (approved):
thanks.

PR #26282, opened 2024-08-14
gtk-4: enable git updates

rawlingsj:


rawlingsj:
Let's change from using object commit to tag commit as that is what we're using got the github impl

rawlingsj:
The failing `Wolfictl Check Updates` check is expected as this PR uses the new `git:` update config which is being validated by the replacement check `Package Update Config Check`

PR #26284, opened 2024-08-14
fix(jsoncpp): Build with cmake instead of meson

EyeCantCU:
Our dev package doesn't currently contain the cmake headers we need to build other projects. Switch from meson to cmake so that these are generated


jamie-albert:
Any reason why this is still a draft PR?

srishtih (approved):

PR #26370, opened 2024-08-15
Revert beego bump in harbor-2.11

priyawadhwa:
@EyeCantCU reverted this in https://github.com/wolfi-dev/os/pull/26137 but then our automation added it back in (https://github.com/wolfi-dev/os/pull/26157). I think it's breaking the harbor helm test which is causing https://github.com/chainguard-dev/image-release-stats/issues/280 (cc @smoser)

I added `manual: true` for now but please let me know if there's a better way to stop automation from adding this back in

### DIFF
--- a/harbor-2.11.yaml
+++ b/harbor-2.11.yaml
@@ -1,7 +1,7 @@
 package:
   name: harbor-2.11
   version: 2.11.0
-  epoch: 6
+  epoch: 7
   description: An open source trusted cloud native registry project that stores, signs, and scans content
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


EyeCantCU (dismissed):
Thank you for fixing this!

ajayk:
You need to bump the epoch as well

EyeCantCU:
Just to add, unfortunately I don't believe there's a way. I filed advisories but automation still tries to remediate the CVE

ajayk:
and `manual: true` is for updates not cve fixes , we can put the CVE fix in the draft whenever we get so it will not update it

priyawadhwa:
Sounds good, will keep an eye out for the automation PR and put it in draft. Thank you @ajayk & @EyeCantCU!

ajayk (approved):

PR #26396, opened 2024-08-15
renovate: add a daily update schedule

rawlingsj:


rawlingsj:
The failing update check is ok as this PR aims to avoid multiple renovate PRs per day, we've had three today already :)

ajayk (approved):

PR #26422, opened 2024-08-15
FTBFS libffi

xnox:
Workaround broken ax_builddir by disabling automatic build
subdirectory creation. Fixes the build.

Also switch to autoconf/configure pipeline that automatically does
autoreconf.


jamie-albert:
Linting needs to be addressed

jamie-albert (approved):

PR #26453, opened 2024-08-16
Allow installing gcc-14 by common prefix

xnox:


mamccorm (changes requested):
Noticed this is still draft - so  I dismissed this review, and added to comments instead

mamccorm:
Couple of issues here - it's leveraging release-monitor: https://release-monitoring.org/project/6502, which will move onto v15 as it's not filtering / limiting updates to v14. This would cause this package to jump to v15 but the provides would still be for v14.

We could use var-transforms to extract '14' from the `version`, then have the provides do something like: `gcc-${{var.major-version}}=${{package.full-version}}`, but not sure if that will work at this stage.

Other than that, we could:
rename this package to gcc-14, and create a new project on release-monitor, and limit it to v14 updates only

xnox:
> Couple of issues here - it's leveraging release-monitor: https://release-monitoring.org/project/6502, which will move onto v15 as it's not filtering / limiting updates to v14. This would cause this package to jump to v15 but the provides would still be for v14.
> 
> We could use var-transforms to extract '14' from the `version`, then have the provides do something like: `gcc-${{var.major-version}}=${{package.full-version}}`, but not sure if that will work at this stage.
> 
> Other than that, we could:
> rename this package to gcc-14, and create a new project on release-monitor, and limit it to v14 updates only

The other issue is that gcc-N.yaml and gcc.yaml are not symmetrical packages. As gcc.yaml contains compilers and runtime libraries that are intentionally not version streamed. Maybe all those should be split too.

Also I am trying to add these provides to generate a better gcc image for c/c++ compilers only.

PR #26468, opened 2024-08-16
gitaly-17.2: switch to git updates and revert to 17.2.2 as it was inc…

rawlingsj:
…orrectly updated to 17.3 stream, withdraw bad apk


rawlingsj:
I think the test failure is because the incorrect 17.3 which this PR withdraws is being added to the test env instead of the version we are reverting to, which fails the grep version check.

hectorj2f (approved):

PR #26471, opened 2024-08-16
tiff: switch to git updates

rawlingsj:


hectorj2f (approved):

tiff.yaml
```diff
@@ -66,5 +66,6 @@ subpackages:
 
 update:
   enabled: true
-  release-monitor:
-    identifier: 1738
+  github:
```
debasishbsws, line 69: Shouldn't it be `git`.
Also github without identifier will this work?

PR #26487, opened 2024-08-16
backfill: use gsutil

imjasonh:
It doesn't seem to be actually backfilling, and only seems to try for the first package in the list. This works in the other repo, so maybe `gcloud alpha storage cp` has different behavior we were relying on.

If this doesn't work I can disable some of the log-quieting stuff and figure out why it's broken.


k4leung4 (dismissed):

k4leung4:
rebase please

k4leung4 (approved):

PR #26490, opened 2024-08-16
statsd - Add modules that statsd depends on

smoser:
The request that triggered this was for proxy.js function which would throw an error :

    Error: Cannot find module 'hashring'
    code: 'MODULE_NOT_FOUND',

We were not installing any of its dependencies, so this adds both the production dependencies (only "generic-pool") and the optional (modern-syslog, hashring, winser).

### DIFF
--- a/statsd.yaml
+++ b/statsd.yaml
@@ -1,7 +1,7 @@
 package:
   name: statsd
   version: 0.10.2
-  epoch: 0
+  epoch: 1
   description: Daemon for easy but powerful stats aggregation
   copyright:
     - license: MIT
@@ -10,6 +10,8 @@ environment:
   contents:
     packages:
       - busybox
+      - nodejs
+      - npm
 
 pipeline:
   - uses: git-checkout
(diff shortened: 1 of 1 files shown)


statsd.yaml
```diff
@@ -26,11 +28,49 @@ pipeline:
...
   - runs: |
       # Set graphite hostname to "graphite"
       sed -i 's/graphite.example.com/graphite/' exampleConfig.js
       mv exampleConfig.js ${{targets.destdir}}/usr/src/app/config.js
 
+test:
+  environment:
+    contents:
+      packages:
+        - nodejs-22
```
EyeCantCU, line 45: Looks great. Just one minor nitpick - in our tests here, we pin to nodejs 22, but above we install hashring with the latest version of nodejs. The two should align

smoser: thanks for the review.  I dropped the '22' here.

EyeCantCU (approved):
LGTM!

PR #26492, opened 2024-08-16
cmake: add cmake-bootstrap to build more deps with cmake

xnox:
Many packages can be built with autotools/make or cmake. At the moment
we build those that are build-dependencies of cmake with
autotools/make to prevent a bootstrap cycle. However this causes many
of the core libraries to lack cmake files, preventing other higher up
software from using cmake and detecting zstd / lz4 and so on.

Resolve this as follows.

Create cmake-bootstrap that uses vendored copies of dependencies. Set
it to provide cmake, but at a low priority.

Delete previous duplicated packages such as c-ares-bootstrap,
zstd-cmake, that exist because they are part of cmake build dependency
chain.

Build c-ares / lz4 / zstd with cmake (either cmake-bootstrap or
cmake), such that cmake config files are generated in their -dev
packages. Build cmake itself with system libraries only, which incldue
all of the previously mentioned ones.

This resolves apache-arrow failing to build from source because
lz4-dev does not currently provide cmake configury files. But it also
makes it easier to maintain everything, as more packages can use cmake
earlier in the bootstrap.


mamccorm:
I'd not be experienced enough in this area to review this one. @smoser would you be able to take a look?

pnasrat (approved):
LGTM

PR #26505, opened 2024-08-17
go: trim 4 EOL go toolchains

xnox:
Follow https://go.dev/doc/install/source#go14 gcc-go provides 1.18,
attempt to use that to bootstrap go-1.20, and then use that to
bootstrap 1.22 and 1.23.

Thus removing need to keep go-1.19 and go-1.21.
Also drop EOL go-fips toolchains.

In the future we may want to create go-bootstrap package that uses
https://pkg.go.dev/golang.org/x/build/cmd/gorebuild#section-sourcefiles
to bootstrap our toolchain in a reproducible and verifyable manner.


xnox:
Shakes fist at datadog agent.

ajayk (approved):

PR #26526, opened 2024-08-18
nss_wrapper: new package

f-bn:
This PR adds the nss_wrapper library for supporting execution of the container with arbitrary user ID for some applications that my not supporting it correctly.

It is used in the entrypoint script of some official OCI images such as the `postgres` image (and also used as part of the packaged `postgresql-*-oci-entrypoint` Wolfi packages):

https://github.com/docker-library/postgres/blob/master/16/bookworm/docker-entrypoint.sh#L72

Can be useful in some OpenShift environments too.

#### For new package PRs only


mepi262:
@f-bn 
It seems that some checks are not successful.
Would you check it?

mamccorm:
Stale PR, has failing CI checks and no updated responses since last posts

PR #26565, opened 2024-08-19
fix: update git settings are wrong

hectorj2f:
We were getting errors when checking the github identifier for this package when it should be using `git`.


hectorj2f:
These settings are throwing errors in production

debasishbsws (approved):

PR #26607, opened 2024-08-19
opus: fix package to reach new tags instead of releases

hectorj2f:


mamccorm:
Looks like these changes already made it into the latest package

PR #26609, opened 2024-08-19
FTBFS zsh

xnox:
Cherry-pick upstream changes to configure.ac to use valid C and thus
correctly detect terminfo libraries with gcc-14.


pnasrat (approved):
LGTM

PR #26674, opened 2024-08-19
Multi-versionize py3-rouge-score

justinvreeland:
Fixes:

Related:

### DIFF
--- a/py3-rouge-score.yaml
+++ b/py3-rouge-score.yaml
@@ -13,6 +13,16 @@ package:
       - py3-six
       - python-3
 
+vars:
+  pypi-package: rouge-score
+
+data:
+  - name: py-versions
+    items:
+      3.10: "310"
+      3.11: "311"
+      3.12: "312"
+
 environment:
   contents:
     packages:
@@ -20,7 +30,9 @@ environment:
       - busybox
       - ca-certificates-bundle
       - py3-setuptools
-      - python-3
+      - py3-supported-build
+      - py3-supported-installer
+      - py3-supported-pip
       - wolfi-base
 
 pipeline:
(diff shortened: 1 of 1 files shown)


justinvreeland:
@pnasrat  Still can't assign unfortunately.

py3-rouge-score.yaml
```diff
@@ -29,10 +41,22 @@ pipeline:
...
+    name: py${{range.key}}-${{vars.pypi-package}}
+    pipeline:
+      - uses: py/pip-build-install
+        with:
+          python: python${{range.key}}
+      - uses: strip
 
-  - uses: strip
+  - name: py3-supported-${{vars.pypi-package}}
+    description: meta package providing ${{vars.pypi-package}} for supported python versions.
```
pnasrat, line 54: remove py3-supported block in general do not add this subpackage.

py3-rouge-score.yaml
```diff
@@ -29,10 +41,22 @@ pipeline:
...
-  - name: Python Build
-    uses: python/build-wheel
+subpackages:
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}
+    pipeline:
+      - uses: py/pip-build-install
+        with:
+          python: python${{range.key}}
+      - uses: strip
```
pnasrat, line 51: add a test block wit a test import (probably `rouge_score`)

pnasrat (changes requested):

py3-rouge-score.yaml
justinvreeland: Don't we need to to provide for packages that will depend on this package once they're also building for multiple versions?

pnasrat (approved):
LGTM

PR #26675, opened 2024-08-19
Add sentencepiece package

justinvreeland:
Builds 4 packages 1 for each python3 version and the libraries.

New sentencepiece package

Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/py3-sentencepiece.yaml
+++ b/py3-sentencepiece.yaml
@@ -1,0 +1,74 @@
+package:
+  name: py3-sentencepiece
+  version: 0.2.0
+  epoch: 0
+  description: "SentencePiece is an unsupervised text tokenizer and detokenizer"
+  copyright:
+    - license: Apache-2.0
+
+vars:
+  pypi-package: sentencepiece
+
+data:
+  - name: py-versions
+    items:
+      3.10: "310"
+      3.11: "311"
+      3.12: "312"
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - cmake
+      - py3-supported-pip
+      - py3-supported-python-dev
+
+pipeline:
+  - uses: git-checkout
+    with:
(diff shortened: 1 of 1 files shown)


justinvreeland:
@pnasrat

sentencepiece.yaml
```diff
@@ -0,0 +1,55 @@
...
+subpackages:
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}
+    description: ${{vars.pypi-package]} installed for python${{range.key}}
+    pipeline:
+      - runs: python${{range.key}} -m pip install ./python "--root=${{targets.subpkgdir}}" --prefix=/usr
+      - uses: strip
+    dependencies:
+      runtime:
+        - py3-${{vars.pypi-package}}
```
pnasrat, line 52: the `py3-${{vars.pypi-package}}` dependency looks incorrect

The py3-sentencepiece virtual should be handled by the provides below. You allso need to use provider-priority set to `${{range.value}}` for these subpackages.

pnasrat (dismissed):

smoser (changes requested):
My only question here is why you named this 'sentancepiece' rather than 'py3-sentancepiece'.  setuptools is named 'py3-setuptools' but the "upstream name" (and name in pypi) is "setuptools".

Why differ here?

I'm only discussing consistency, not which is right or wrong.  we have ~ 525 py-* packages.

justinvreeland:
> My only question here is why you named this 'sentancepiece' rather than 'py3-sentancepiece'. setuptools is named 'py3-setuptools' but the "upstream name" (and name in pypi) is "setuptools".
> 
> Why differ here?

This was the first one I put up and I didn't really know how we were naming things. It should be named to match the others.

smoser (approved):

PR #26685, opened 2024-08-19
fix(tomcat-10.1): Build packages for supported JDK versions

EyeCantCU:
Allows us to bundle the correct JDK version with the package and run tests for each supported JDK version at the package level

### DIFF
--- a/tomcat-10.1.yaml
+++ b/tomcat-10.1.yaml
@@ -10,6 +10,13 @@ package:
       - tomcat=${{package.full-version}}
       - tomcat-10=${{package.full-version}}
 
+data:
+  - name: openjdk-versions
+    items:
+      21: "openjdk-21"
+      17: "openjdk-17"
+      11: "openjdk-11"
+
 environment:
   contents:
     packages:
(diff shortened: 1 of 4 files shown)


tomcat-10.1.yaml
```diff
@@ -35,78 +40,77 @@ pipeline:
...
+subpackages:
+  - range: openjdk-versions
+    name: ${{package.name}}-${{range.value}}
+    dependencies:
+      runtime:
+        - ${{range.value}}-default-jvm
+        - tomcat-native
+    pipeline:
+      - runs: |
+          export JAVA_HOME="/usr/lib/jvm/java-17-openjdk"
```
pnasrat, line 52: java-${{range.key}}

EyeCantCU: Tomcat builds with OpenJDK 17, but supports 11-21 at runtime. We benefit from tests against all JVM versions supported at runtime, but we do end up building duplicate packages

pnasrat: ah ok can you add a comment to that effect!

EyeCantCU: Yes, I can definitely do that :slightly_smiling_face:

pnasrat (approved):
LGTM

PR #26694, opened 2024-08-20
supercronic: remove redundant dependency

jowiho:
The `supercronic` application doesn't need the `go` runtime (it only needs `go` at build time). Removing this unnecessary dependency saves roughly 700 MB of dependency packages from getting installed together with `supercronic`.

### DIFF
--- a/supercronic.yaml
+++ b/supercronic.yaml
@@ -1,7 +1,7 @@
 package:
   name: supercronic
   version: 0.2.30
-  epoch: 1
+  epoch: 2
   description: Cron for containers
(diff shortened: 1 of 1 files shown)


ajayk:
@jowiho Thanks for your contribution this needs  epoch bump for the rebuild `  epoch: 1` changes to   `epoch: 2`

jowiho:
Thanks for the help, @ajayk. I've pushed a commit to bump the epoch.

ajayk (approved):

PR #26761, opened 2024-08-20
Add python versioned httpcore and hatch-fancy-pypi-readme

srishtih:
Adds hatch-fancy-pypi-readme with -supported and -bin subpackages needed as build time tools for httpcore

### DIFF
--- a/py3-hatch-fancy-pypi-readme.yaml
+++ b/py3-hatch-fancy-pypi-readme.yaml
@@ -12,6 +12,7 @@ package:
 vars:
   pypi-package: hatch-fancy-pypi-readme
   module-name: hatch_fancy_pypi_readme
+
 data:
   - name: py-versions
     items:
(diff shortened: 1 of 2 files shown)


pnasrat (dismissed):
Drop the supported subpackages

srishtih:
I think its needed to build the httpcore package

pnasrat:
We use the following pipeline so build dep shouldn't matter unless it is a
build backend

pipelines/py/pip-build-install.yaml

I believe it should just work if you list as runtime dep

Pris

pnasrat:
If it's an actual build dep then LGTM

srishtih:
It is an actual build time dependency for bth httpx and httpcore, the builds fail without it.

mamccorm (approved):

EyeCantCU (approved):

PR #26785, opened 2024-08-20
mariadb-11.5: update docker-entrypoint

ajayk:


xnox (approved):
we are supposed to be compatible, so not going to question this entrypoint script.

PR #26802, opened 2024-08-21
Rebuild gcc against new glibc

xnox:
pthread mutex appears to use private symbols, and thus needs rebuild against new glibc.

Ideally need test for this, but nothing small is identified yet.

Typical symptoms are:

```
2024/08/21 00:42:33 WARN /usr/lib/gcc/x86_64-pc-linux-gnu/12/../../../../x86_64-pc-linux-gnu/bin/ld: somefile:(.text+0x5aa6): more undefined references to `__pthread_mutex_timedlock64' follow
```

### DIFF
--- a/gcc-12.yaml
+++ b/gcc-12.yaml
@@ -1,7 +1,7 @@
 package:
   name: gcc-12
   version: 12.4.0
-  epoch: 4
+  epoch: 5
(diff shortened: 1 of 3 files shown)


pnasrat (approved):
LGTM

imjasonh (approved):

PR #26807, opened 2024-08-21
keycloak: properly handle providers dir

Dentrax:


keycloak.yaml
```diff
@@ -109,6 +109,10 @@ subpackages:
           # an initContainer
           cp -r ${{targets.destdir}}/usr/share/java/keycloak/* ${{targets.contextdir}}/opt/bitnami/keycloak
 
+          # Properly symlink the providers directory after installation
+          rm -rf ${{targets.contextdir}}/opt/bitnami/keycloak/providers/
+          ln -sf /usr/share/java/keycloak/providers ${{targets.contextdir}}/opt/bitnami/keycloak
```
shimmerjs, line 114: I suspect this will re-introduce the same issue since everything under `/opt/bitnami/keycloak` is copied from the image into the volume (https://github.com/bitnami/charts/blob/bb05f9f24ef535ec72ac760529b15f2b7bd67254/bitnami/keycloak/templates/statefulset.yaml#L108)

Dentrax: Ah, it could be. So we should make a new subpackage for `cp` it?

EyeCantCU (approved):

Issue #26818, opened 2024-08-21
testing python is too hard / cannot use 'python3' / python transition will cause problems


smoser:
Because we have multiple pythons and we want them to be co-installable, we have packages like:
 1. python-3.12-base
 2. python-3.12
 3. python-3.11-base
 4. python-3.11

The -base package provides basically everything and the python-3.11 package provides a symlink making 'python3' default to python 3.11 (` /usr/bin/python3 -> python3.11`)

When we build packages, melange correctly identifies that a package that installs python files into `/usr/lib/python3.12/site-packages/` depends on `python-3.12-base`.  It does _not_ depend on `python-3.12` (it never uses that symlink. even the installed scripts have /usr/bin/python3.12).

The pipeline 'python/imports' [pipeline](https://github.com/chainguard-dev/melange/blob/98b252c0aaf289ae690e17256cec04e3438c4777/pkg/build/pipelines/python/import.yaml) takes a 'python' input, but if it is not provided it will use a shell snippet to find the right one (if there is only one installed).

That works well for the imports pipeline.

The problem is that anything that wants to run 'python3' in a test has to have a similar snippet of code.

Consider [py3-async-generator.yaml](https://github.com/wolfi-dev/os/blob/1016a42e71ebf11748aae1846cc0dd77c69edab2/py3-async-generator.yaml).

It looks like this:
```
package:
  name: py3-async-generator
  version: 1.10
  epoch: 1
  description: Async generators and context managers for Python 3.5+
  copyright:
    - license: MIT OR Apache-2.0
  dependencies:
    runtime:
      - python-3
...
test:
  environment:
    contents:
      packages:
        - python-3
  pipeline:
    - runs: |
        python -c "import async_generator; print(async_generator.__version__)"
```

Two things to call out there
1. It has 'python-3' as a runtime dep.   The  package built today will have the python-3.12 dep written by melange but will also have `python-3` dep.  In the future, when python-3.13 exists, python-3.13 will also provide python-3 and will be preferred to python-3.12.   So installation of this package would install 2 python versions.  Yikes.
2. It includes `python-3` in the test environment in order to get a `python` program to execute.  But similarly to 1, that will install the _wrong_ python in the future, and the test will begin to fail (it works today, but as soon as there is a python-3 package it will fail).

So, how should that test look?  Do we have to copy a shell snippet like this:
```
python=$(m=/usr/bin/python3.*; set +x; set -- $m
[...]

PR #26822, opened 2024-08-21
drop python req from pip-build-install, move to pip-build-install, add tests

smoser:
There are lots of changes here, centered around removing all occurences of SETUPTOOLS_SCM_PRETEND_VERSION.  That seems to be not necessary now, at least with the current pip-build-install workflow.

 * do not require 'python' setting in pip-build-install.  Default to using the target of the 'python3' link. If not present, user has to supply it.

   Also use contextdir rather than subpkgdir so it can be used from "main" packages also.

 * replace any packages that had SETUPTOOLS_SCM_PRETEND_VERSION with py/pip-build-install.

 * Add tests to touched packages

 * Drop 'python3' runtime dep from touched packages.

 * Fix uninstallable py3-gcloud-aio-storage which conflicted with its dependency py3-gcloud-aio-auth as they both provided usr/lib/python3.12/site-packages/gcloud/py.typed

 * Fix the provides in py3-maturin-bin packages to provide py3-maturin-bin rather than just py3-maturin.  That also fixes py3-pendulum which was failing to build because it actually needs 'maturin' executable but that was not being provided by the 'py3-maturin' dep.

 * Fix pendulum's dependency that had py3-pytzdata, it's deps list 'tzdata'

 * Drop the now-unused py3-pytzdata

Note the 'python3' shell snippet in py3-rich and others is painful. See https://github.com/wolfi-dev/os/issues/26818 for more context.


smoser:
closing this for now in favor of 2 steps.
https://github.com/wolfi-dev/os/pull/26914  is the first step, just to improve the pipelines and do one package.

PR #26827, opened 2024-08-21
py3-rouge-score: Fix missing dependencies

justinvreeland:


py3-rouge-score.yaml
pnasrat, line 1: you also need the provider-priority: 0 on the top level package

pnasrat (changes requested):

smoser (dismissed):

pnasrat (approved):
LGTM

PR #26841, opened 2024-08-21
Multi-versionize py3-multidict

justinvreeland:
Build py3-multidict for multiple python versions.

### DIFF
--- a/py3-multidict.yaml
+++ b/py3-multidict.yaml
@@ -2,13 +2,22 @@
 package:
   name: py3-multidict
   version: 6.0.5
-  epoch: 1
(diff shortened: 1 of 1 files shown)


py3-multidict.yaml
```diff
@@ -7,17 +7,26 @@ package:
   copyright:
     - license: Apache-2.0
   dependencies:
-    runtime:
-      - python3
+    provider-priority: 0
+
+vars:
```
imjasonh, line 12: You'll want to increment the `epoch` value, so this builds a new package.

mamccorm (approved):

PR #26842, opened 2024-08-21
Adds victoriametrics package with vmagent subpackage

dakaneye:
Related to https://github.com/chainguard-dev/image-requests/issues/1955

### DIFF
--- a/victoriametrics.yaml
+++ b/victoriametrics.yaml
@@ -19,8 +19,7 @@ subpackages:
     pipeline:
       - uses: go/build
         with:
-          ldflags: -extldflags "-static"
-          toolchaintags: netgo,osusergo,nethttpomithttp2,musl
+          tags: nethttpomithttp2
           packages: github.com/VictoriaMetrics/VictoriaMetrics/app/vmagent
           output: vmagent
     test:
--- a/victoriametrics.yaml
+++ b/victoriametrics.yaml
@@ -20,7 +20,7 @@ subpackages:
       - uses: go/build
         with:
           tags: nethttpomithttp2
-          packages: github.com/VictoriaMetrics/VictoriaMetrics/app/vmagent
+          packages: ./app/vmagent
           output: vmagent
     test:
       environment:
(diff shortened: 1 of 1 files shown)


victoriametrics.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+subpackages:
+  - name: ${{package.name}}-vmagent
+    description: vmagent is a tiny agent which helps you collect metrics from various sources, relabel and filter the collected metrics and store them in VictoriaMetrics or any other storage systems via Prometheus remote_write protocol or via VictoriaMetrics remote_write protocol.
+    pipeline:
+      - uses: go/build
+        with:
+          ldflags: -extldflags "-static"
+          toolchaintags: netgo,osusergo,nethttpomithttp2,musl
+          packages: github.com/VictoriaMetrics/VictoriaMetrics/app/vmagent
```
xnox, line 24: static linking is either redundant (as binary is already static anyway) or does not achieve what you think it does (statically links glibc, and shared libraries creating dark matter in our containers unaccounted for with sbom).

some of the tags are redundant as already specified by our default pipeline.

Some look sus - what does musl tag do?

victoriametrics.yaml
```diff
@@ -0,0 +1,40 @@
+package:
+  name: victoriametrics
```
xnox, line 2: If the package only ships a subpackages and nothing else, can you call this package vicrotiametrics-vmagent directly?

xnox (changes requested):

victoriametrics.yaml
dakaneye: Ah that's probably a relic of the fact they're own docker image is alpine, i'll try removing. nethttp one seems like it's just a space saver for them. I can remove static linking as I was just following their build: https://github.com/VictoriaMetrics/VictoriaMetrics/blob/22975d28ae4871a4e6ede1e482b373e13523870b/deployment/docker/Makefile#L43 and it worked without it anyways for the fips side.

victoriametrics.yaml
dakaneye: @developer-guy plans to add several more subpackages in the near future (that are separate apps in the same monorepo)

xnox (approved):

PR #26843, opened 2024-08-21
Py3 markdown it py

justinvreeland:
Build py3-markdown-it-py for multiple py3 versions

Also bump py3-mdurl because I didn't in a previous commit.


justinvreeland:
The build failure is because it can't find the updated mdurl which is included in this patch.

pnasrat (approved):
LGTM

PR #26844, opened 2024-08-21
Multi-versionize py3-yarl

justinvreeland:
Build py3-yarld and deps against multi versions of python3

Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/py3-cython.yaml
+++ b/py3-cython.yaml
@@ -1,7 +1,7 @@
 package:
   name: py3-cython
   version: 3.0.11
-  epoch: 0
+  epoch: 1
   description: Cython is an optimising static compiler for both the Python & the extended Cython programming languages.
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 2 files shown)


justinvreeland:
@smoser  Any changes to cython are likely to happen here or be split out into it's own PR since it's a deep for a few things and the other PR is not going to be ready for a while.

justinvreeland:
All of the deps should be available now we'll see what the CI comes back with though.

py3-yarl.yaml
```diff
@@ -29,10 +37,28 @@ pipeline:
...
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}
+    description: ${{vars.pypi-package}} installed for python${{range.key}}
+    dependencies:
+      provider-priority: ${{range.value}}
+      provides:
+        - py${{range.key}}-${{vars.pypi-package}}
+      runtime:
+        - py${{range.key}}-idna
+        - py${{range.key}}-multidict
```
smoser, line 50: I noticed you dropped typing-extensions.
```

$ curl https://pypi.org/pypi/yarl/json | jq '.info | .version, .requires_dist' < out
"1.9.4"
[
  "idna >=2.0",
  "multidict >=4.0",
  "typing-extensions >=3.7.4 ; python_version < \"3.8\""
]
```
that says its ok to do that.
It'd be good to mention it in the commit message for others.

smoser (dismissed):

py3-cython.yaml
```diff
@@ -40,17 +40,42 @@ subpackages:
...
+    dependencies:
+      runtime:
+        - py${{range.key}}-${{vars.pypi-package}}
+      provides:
+        - cython
+      provider-priority: ${{range.value}}
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/usr
+          mv ${{targets.contextdir}}/../py${{range.key}}-${{vars.pypi-package}}/usr/bin ${{targets.contextdir}}/usr
```
smoser, line 61: This is kind of yucky.  Depends on a specific value of targets.subpkgdir for another package.
another option would be for the non -bin version to 

```
  mkdir -p ./cleanup/${{range.key}/cython
  mv ${{targets.contextdir}}/usr/bin/cython ./cleanup/${{range.key}/
```

And then here you would:

```
  mv ./cleanup/${{range.key}}/cython ${{targets.contextdir}}/usr/bin
```

justinvreeland:
Testing? I have pending comments but don't see a submit button?

py3-cython.yaml
```diff
@@ -38,15 +38,47 @@ subpackages:
     name: py${{range.key}}-${{vars.pypi-package}}
     dependencies:
       provider-priority: ${{range.value}}
-      provides:
```
smoser, line 41: don't we want to
```yaml
dependencies:
  provides:
    - py3-${{vars.pypi-package}}
```

smoser (changes requested):

py3-cython.yaml
justinvreeland: I do!

justinvreeland: Do people expect `py3-cython` to provide `cython`?

py3-yarl.yaml
```diff
@@ -1,40 +1,74 @@
...
+    - uses: python/import
+      with:
+        import: ${{vars.pypi-package}}
 
 update:
   enabled: true
-  release-monitor:
-    identifier: 12895
+  github:
+    identifier: aio-libs/yarl
```
smoser, line 72: this is wrong.

smoser (changes requested):

py3-yarl.yaml
justinvreeland: Could you clarify how it's wrong?  That does seem to be the correct repo.

smoser: :cry: 
its not. i saw 'aio-libs' and was thinking that was the project rather than the org. like you pointed the updates at a different project.

smoser (dismissed):

smoser (approved):

PR #26855, opened 2024-08-21
py3-h5py: Multi-versionize py3-h5py and deps

justinvreeland:
Build py3-h5py and deps for multiple versions of python3

### DIFF
--- a/pipelines/py/build-installer.yaml
+++ b/pipelines/py/build-installer.yaml
@@ -1,0 +1,52 @@
+name: Build and install python packages with build and installer
+
+inputs:
+  python:
+    description: which python to use
+    default: DEFAULT
+  dest:
+    description: the destination
+    default: ${{targets.contextdir}}
+
+pipeline:
+  - name: "Build installer"
+    runs: |
+      py=${{inputs.python}}
+      if [ "$py" = "DEFAULT" ]; then
+        if p=$(command -v python3); then
+          py=$p
+          if [ -L "$p" ]; then
+            py=$(readlink -f "$p") ||
+              { echo "ERROR: failed 'readlink -f $p'" 1>&2; exit 1; }
+          fi
+        else
+          glob="/usr/bin/python3.[0-9][0-9] /usr/bin/python3.[789]"
+          n=0
+          for p in $glob; do
+            [ -x "$p" ] && n=$((n+1)) && py=$p && found="$found $p"
+          done
+          if [ "$n" -ne 1 ]; then
+            echo "ERROR: must set inputs.python: " \
+                "found $n executables matching $glob" 1>&2
+            [ "$n" -eq 0 ] || echo "  found: $found" 1>&2
+            exit 1
+          fi
+        fi
+      fi
+      pyver=$("$py" -c 'import sys; print("%s.%s" % (sys.version_info.major, sys.version_info.minor))')
+      sitepkgd=$("$py" -c 'import site; print(site.getsitepackages()[0])')
+      sitepkgd=${sitepkgd#/}
+      root=${{inputs.dest}}
+      wd="./.wheels/$pyver"
+      vr() { echo "execute:" "$@"; "$@"; }
+
+      tmpd=$(mktemp -d)
+      trap "rm -Rf '$tmpd'" EXIT
+
+      # --find-links to an empty dir and --no-index makes pip fully "offline"
+      distwheelsd="$tmpd/dist-wheels"
+      mkdir -p "$distwheelsd"
+      echo "$py is $pyver with site_packages dir '$sitepkgd'"
+      vr $py -m build --wheel --verbose --no-isolation
+      vr $py -m installer -d $root dist/*.whl
+      vr find $root -name "*.pyc" -exec rm -rf '{}' +
--- a/py3-meson-python.yaml
+++ b/py3-meson-python.yaml
@@ -47,11 +47,12 @@ subpackages:
         - py3-${{vars.pypi-package}}
       provider-priority: ${{range.value}}
       runtime:
+        - meson
         - py${{range.key}}-tomli
+        - py${{range.key}}-setuptools
         - py${{range.key}}-packaging
         - py${{range.key}}-pyproject-metadata
         - py${{range.key}}-colorama
-        - py${{range.key}}-packaging
     pipeline:
       - uses: py/pip-build-install
         with:
@@ -71,6 +72,18 @@ subpackages:
         - py3.10-${{vars.pypi-package}}
         - py3.11-${{vars.pypi-package}}
         - py3.12-${{vars.pypi-package}}
+    test:
+      pipeline:
+        - uses: python/import
+          with:
+            python: python3.10
+            import: mesonpy
+
+test:
+  pipeline:
+    - uses: python/import
+      with:
+        import: mesonpy
 
 update:
   enabled: true
--- a/py3-shellingham.yaml
+++ b/py3-shellingham.yaml
@@ -52,6 +52,7 @@ subpackages:
           with:
             python: python${{range.key}}
             import: ${{vars.pypi-package}}
+
 test:
   pipeline:
     - uses: python/import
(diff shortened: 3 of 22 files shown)


py3-cython.yaml
```diff
@@ -40,17 +40,39 @@ subpackages:
...
 
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}-bin
+    description: Executable binaries for ${{vars.pypi-package}} installed for python${{range.key}}
+    dependencies:
+      runtime:
+        - py${{range.key}}-${{vars.pypi-package}}-bin
+      provides:
+        - py3-${{vars.pypi-package}}
+        - cython
```
justinvreeland, line 57: Tagged this on the other PR but this shouldn't provide the models just the binaries.

py3-meson-python.yaml
```diff
@@ -10,10 +10,21 @@ package:
       - meson
       - py3-colorama
       - py3-packaging
-      - py3-pyproject-metadata
+      - py3-pyproject-metadat
```
justinvreeland, line 13: spelling error? ahh need to clean up the runtime deps.

justinvreeland (commented):
Draft of py3-p5py for context for python and dumpy changes. Still has some work updating the various blocks to match best practices.

py3-cython.yaml
pnasrat: `py3-build.yaml` has the provides here to be `py3-${{vars.pypi-package}}` for the bin subpackage. I don't know if we might want this to be

`cython=${{package.full-version}}" as we have a `cython-0` version stream also

py3-cython.yaml
```diff
@@ -40,17 +40,38 @@ subpackages:
...
+    name: py${{range.key}}-${{vars.pypi-package}}-bin
+    description: Executable binaries for ${{vars.pypi-package}} installed for python${{range.key}}
+    dependencies:
+      provides:
+        - cython
+      provider-priority: ${{range.value}}
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/usr
+          mv ${{targets.contextdir}}/../py${{range.key}}-${{vars.pypi-package}}/usr/bin ${{targets.contextdir}}/usr
```
pnasrat, line 61: Add a -bin subpackage inline test perhaps just `cython --version`

pnasrat (changes requested):

py3-cython.yaml
justinvreeland: I'm going to implement these changes over on https://github.com/wolfi-dev/os/pull/26844 because h5py still has a fair amount of deps to be fixed up and yarl can reasonably land soon. 

Alternatively I can split off cython and numpy and pickup pandas as well and do those in one PR which tackles some similar packages on our list.

py3-cython.yaml
justinvreeland: I think we want `cython-bin` subpackages to be dependent on the corresponding non-bin subpackages.   And we want the `-bin` packages to have the provides for `cython` that way installing the modules doesn't necessarily install the binaries but installing a specific binary version gets you the corresponding modules.

I don't think we want to provide a new stream though.  No-one should be targeting a specific stream other than the 0 stream and it's not being used by anything anymore. The package it was added for doesn't have problems building with newer versions of cython.

smoser: How would a multi-versioned py3-foo.yaml use cython?

As I understand it, something using cython is going to need a py3.XX-cython and a /usr/bin/cython.
I ran into this just recently in py3-maturin also.

How about this:
1. make py3-cython package build a non-versioned '/usr/bin/cython' with the package name 'cython'
2. remove the cython-0 package
3. make py3.XX-cython depend on 'cython'

Then, to use cython in a build the package just needs to add `py3-cython` or `py3-supported-cython`.

smoser (changes requested):

py3-cython.yaml
justinvreeland: > How would a multi-versioned py3-foo.yaml use cython?

The same way they do now by having a dep on cython? This only changes if their package imports cython from the building interpreter not if they use a compiler. 

> remove the cython-0 package

I missed 2 packages that are using cython~0 so getting rid of it might not be possible right now. I'll check if they can still need that version.

> Then, to use cython in a build the package just needs to add py3-cython or py3-supported-cython

We already have people depending more people depending on `cython` than `py3-cython` and I don't see an advantage to having a real `cython` package rather than bin packages that provide `cython` and letting apk sort out which one to use. As I understand it the dependency solver will choose based on package priority so it's already going to end up with the newest interpreter. I do see potential advantage to being able to create an environment with only a single python interpreter and all the expected pieces available.

pipelines/py/build-installer.yaml
```diff
@@ -0,0 +1,52 @@
...
+      tmpd=$(mktemp -d)
+      trap "rm -Rf '$tmpd'" EXIT
+
+      # --find-links to an empty dir and --no-index makes pip fully "offline"
+      distwheelsd="$tmpd/dist-wheels"
+      mkdir -p "$distwheelsd"
+      echo "$py is $pyver with site_packages dir '$sitepkgd'"
+      vr $py -m build --wheel --verbose --no-isolation
+      vr $py -m installer -d $root dist/*.whl
+      vr find $root -name "*.pyc" -exec rm -rf '{}' +
```
smoser, line 52: this is wrong. we do not want to remove pyc files.

py3-distlib.yaml
```diff
@@ -28,31 +34,34 @@ pipeline:
...
+subpackages:
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}
+    description: python${{range.key}} version of ${{vars.pypi-package}}
+    dependencies:
+      provider-priority: ${{range.value}}
+      provides:
+        - py3-${{vars.pypi-package}}
+    pipeline:
+      - uses: py/build-installer
```
smoser, line 46: took this file as it is right now, and added py3-supported-pip to the envionment.contents.packages list and then it works with 'pip-build-install'.

I suspect the same is true for many of the other ones.

Its annoying that you have to get the dependencies rigth for the pipeline to build.  That is fallout of the pipeline being "multi-version".  It _could_ say it depends on `py3-supported-build`.  maybe that is the rigth way to do things, but that would cause pain for the non-versioned path.... they'd get all the python3-base things installed for a build of just the latest python3.

py3-shellingham.yaml
```diff
@@ -28,13 +33,31 @@ pipeline:
...
+subpackages:
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}
+    description: python${{range.key}} version of ${{vars.pypi-package}}
+    dependencies:
+      provider-priority: ${{range.value}}
+      provides:
+        - py3-${{vars.pypi-package}}
+    pipeline:
+      - uses: py/build-installer
```
smoser, line 45: this is really wierd. you dropped py3-build and installer from the environment.contents.packages but then use build-installer ? how does that even work?

smoser (changes requested):
Probably best to split this up.  Its just really big at this point.  Maybe 3 or 4 at a time?

justinvreeland:
> Probably best to split this up. Its just really big at this point. Maybe 3 or 4 at a time?

Yeah I do intend of splitting this apart it much, much smaller pieces. But this is roughly the stack I had once I got all of the packages building and started cleaning things up.

py3-shellingham.yaml
justinvreeland: Probably a mistake, I added `build-installer` pretty late in going through these deps when once of my packages didn't install correctly with the pip pipeline and started retroactively converting the old ones back.

pipelines/py/build-installer.yaml
justinvreeland: Ahh OK I can remove that, It does seem to be fairly common in our files though

py3-distlib.yaml
justinvreeland: > I suspect the same is true for many of the other ones.
Yeah, I've only encountered one that it didn't work for but encountering that one made me think I was better off using build-installer for things that already specified it and thus I made the pipeline. I'm happy to only change it when something doesn't seem to work with the typical one we use.

PR #26859, opened 2024-08-21
Multi-versionize py3-humanfriendly

justinvreeland:
Build human friendly with multiple python3 versions


justinvreeland:
Each of these package is installing into /bin which seems to be causing problems with the tests. I'm not sure whether we'll want a versioned binaries or not for these.

smoser (approved):

PR #26860, opened 2024-08-21
Multi-versionize py3-distlib

justinvreeland:
Build py3-distlib for multiple python3 versions.


py3-distlib.yaml
```diff
@@ -7,18 +7,24 @@ package:
   copyright:
     - license: PSF-2.0
   dependencies:
-    runtime:
-      - python3
+    provider-priority: 0
```
imjasonh, line 10: ditto, bump the epoch

smoser (approved):

PR #26861, opened 2024-08-21
Multi-versionize py3-dill

justinvreeland:
Build py3-dill for multiple python versions.

### DIFF
--- a/py3-dill.yaml
+++ b/py3-dill.yaml
@@ -2,13 +2,22 @@
 package:
   name: py3-dill
   version: 0.3.8
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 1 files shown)


smoser (dismissed):

py3-dill.yaml
```diff
@@ -7,17 +7,25 @@ package:
   copyright:
     - license: BSD-3-Clause
   dependencies:
-    runtime:
-      - python-3
+    provider-priority: 0
```
imjasonh, line 10: You'll need to bump the epoch value to rebuild this.

smoser (approved):

PR #26909, opened 2024-08-22
Show stderr in wolfictl scan action nif it does not look right.

smoser:
A previous comment says that stderr output is confusing to viewing the github action so it was sent to /dev/null.

That makes debugging harder. So if it doens't look like a "there were vulnerabilities found" error, then show it.

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -335,10 +335,21 @@ jobs:
 
       - name: Scan for CVEs
         run: |
+          stderrf="/tmp/scan.stderr"
           wolfictl scan \
             --build-log \
             --advisories-repo-dir 'data/wolfi-advisories' \
             --advisory-filter 'resolved' \
             --require-zero \
             /tmp/artifacts-1 \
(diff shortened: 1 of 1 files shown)


smoser:
It would be nice if "there were scan errors" exited a different value than '1' (like every other error).
It looks like that is possible https://github.com/spf13/cobra/issues/221 .

If "there were errors" exited 42, then we could just show errors on anything other than 0 or 42.

smoser:
OK, the test run of this showed its value, since the scanner is failing right now.
See [here](https://github.com/wolfi-dev/os/actions/runs/10511419863/job/29122530866?pr=26909), which shows:

> Unexpected error content
> | 2024/08/22 15:54:20 ERRO failed to create scanner: failed to load vulnerability database: unable to update vulnerability database: bad db checksum (/tmp/grype-scratch4276522467/vulnerability.db): "sha256:a11915b8368897dd446ad5cbb855414870297841f8bccc3b466c5c5e9ba5539b" vs "sha256:9aba357712f1f68620ccd19349038e59f17a337189a075f9abea884591925f9b"
> Error: Process completed with exit code 1.

xnox (approved):

imjasonh (approved):

PR #26914, opened 2024-08-22
do not require 'python' setting in pip-build-install

smoser:
This allows non-"multi-versioned" py3 packages to use this pipeline without having to find which python is current themselves.

Update one such package py3-antlr4-python3-runtime.yaml to show the usage.

Default to using python3 in PATH.
If that is not present and there is only one /usr/bin/python3.XX then use it.
If you can't find a single obvious python3, then fail.


smoser:
This is just the pipeline update portion of https://github.com/wolfi-dev/os/pull/26822 .

pnasrat (approved):
LGTM

PR #26917, opened 2024-08-22
Move several py3- packages to pip-build-install, add tests.

smoser:
I set out to get rid of SETUPTOOLS_SCM_PRETEND_VERSION, and found some issues along the way.

 * replace any packages that had SETUPTOOLS_SCM_PRETEND_VERSION with py/pip-build-install.

 * Add tests to touched packages

 * Drop 'python3' runtime dep from touched packages - melange sca does the right thing.

 * py3-cppy - add setuptools dep, found when adding a test.

 * py3-gcloud-aio-storage - Fix uninstallable paackage which conflicted with its dependency py3-gcloud-aio-auth. they both provided usr/lib/python3.12/site-packages/gcloud/py.typed

 * py3-maturin - Fix the provides in py3-maturin-bin packages to provide py3-maturin-bin rather than just py3-maturin. That fixes py3-pendulum which was failing to build because it actually needs 'maturin' executable but that was not being provided by the 'py3-maturin' dep.

 * py3-pendulum - fix dependency it should be tzdata not pytzdata

 * py3-pytzdata - Drop this, it is not used anywhere.

Note the 'python3' shell snippet in py3-rich and others is painful. See https://github.com/wolfi-dev/os/issues/26818 for more context.


pnasrat (approved):
LGTM

PR #26924, opened 2024-08-22
Multi-versionize py3-regex

justinvreeland:
Build py3-regex for multiple python3 versions.


justinvreeland:
@mamccorm   That use tag doesn't seem to be correct, at least locally I'm getting:

```
line 55: field use-tag not found in type config.Update
``

pnasrat (approved):
LGTM

PR #26998, opened 2024-08-23
Multi-versionize py3-aio deps

justinvreeland:
This depends on the work done in https://github.com/wolfi-dev/os/pull/26844

### DIFF
--- a/py3-aiohappyeyeballs.yaml
+++ b/py3-aiohappyeyeballs.yaml
@@ -69,3 +69,5 @@ update:
   github:
     identifier: aio-libs/aiohappyeyeballs
     strip-prefix: v
+    tag-filter: v
+    use-tag: true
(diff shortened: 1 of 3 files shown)


py3-aiosignal.yaml
```diff
@@ -7,18 +7,25 @@ package:
   copyright:
     - license: Apache-2.0
   dependencies:
-    runtime:
-      - py3-frozenlist
-      - python3
+    provider-priority: 0
```
smoser, line 10: need to bump epoch

smoser (changes requested):
one nit.

py3-aiosignal.yaml
justinvreeland: Yeah I should've put this in draft mode a while ago since it can't find happy eyes for the build.

smoser (approved):

PR #27005, opened 2024-08-23
Updates victoriametrics to be two separate packages for cluster vs standalone mode

dakaneye:


victoriametrics-cluster.yaml
```diff
@@ -0,0 +1,59 @@
...
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: 115a76d28c3bb0e63685db1342f80df6af3033dd
+      repository: https://github.com/VictoriaMetrics/VictoriaMetrics
+      tag: v${{package.version}}-cluster
+
+subpackages:
+  - range: victoria-packages
+    name: ${{package.name}}-${{range.key}}
```
EyeCantCU, line 27: Can we do `victoriametrics-${{range.key}}-cluster`?

victoriametrics-single.yaml
```diff
@@ -0,0 +1,56 @@
+package:
+  name: victoriametrics-single
```
EyeCantCU, line 2: We could leave packages for single node configs just `victoriametrics` unless there are deviations in how these are named

victoriametrics-single.yaml
```diff
@@ -0,0 +1,56 @@
...
+        with:
+          tags: nethttpomithttp2
+          packages: ./app/${{range.key}}
+          output: ${{range.key}}
+          ldflags: -X github.com/VictoriaMetrics/VictoriaMetrics/lib/buildinfo.Version=${{range.key}}-$(date -u -d "@${SOURCE_DATE_EPOCH:-$(date +%s)}" "+%Y-%m-%dT%H:%M:%SZ")-$(git rev-parse HEAD)
+    test:
+      environment:
+        contents:
+          packages:
+            - busybox
```
EyeCantCU, line 38: Busybox's inclusion is implicit. We don't need to add it

victoriametrics-single.yaml
dakaneye: There are two differing naming conventions in their repo and docker image tags:
* In the [repo:](https://github.com/VictoriaMetrics/VictoriaMetrics/tags) tags are v1.102.1 and v1.102.1-cluster
* In [dockerhub](https://hub.docker.com/r/victoriametrics/vmagent/tags?page=&page_size=&ordering=&name=): they use `cluster` and `public-single-node`

dakaneye: removed the `single` suffix

victoriametrics.yaml
```diff
@@ -6,6 +6,13 @@ package:
   copyright:
     - license: Apache-2.0
 
+data:
+  - name: victoria-packages
+    items:
+      victoria-metrics: VictoriaMetrics is a fast, cost-effective and scalable monitoring solution and time series database.
```
EyeCantCU, line 12: I think we should retain this in the main package so that we don't end up with victoriametrics-victoria-metrics

EyeCantCU (approved):

PR #27036, opened 2024-08-24
add ruby3.2-llhttp with ffi and mri

Dentrax:
Adds `ruby3.2-llhttp` library with FFI and MRI implementations.

### DIFF
--- a/ruby3.2-llhttp.yaml
+++ b/ruby3.2-llhttp.yaml
@@ -111,6 +111,7 @@ subpackages:
 
 update:
   enabled: true
+  manual: true
   # The original release tag is in the format: 2023-03-29, but that is not a
   # valid apk package version, so we need to convert it to 2023.03.29.
   version-transform:
--- a/ruby3.2-llhttp.yaml
+++ b/ruby3.2-llhttp.yaml
@@ -1,0 +1,157 @@
+package:
+  name: ruby3.2-llhttp
+  version: 2023.03.29
+  epoch: 0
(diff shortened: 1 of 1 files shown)


ruby3.2-llhttp.yaml
```diff
@@ -0,0 +1,157 @@
+package:
+  name: ruby3.2-llhttp
+  version: 2023.03.29
+  epoch: 0
```
mamccorm, line 4: In sustaining, we'll get automated PRs opened to review, and often only see the version and epoch in the diff, so i've been starting to add some notes here to make it more obvious:


```suggestion
  # NOTE: There are also environment variables in this melange which may have to be bumped between upgrades.
  epoch: 0
```

ruby3.2-llhttp.yaml
```diff
@@ -0,0 +1,157 @@
...
+              fi
+              mv ${{vars.gem-ffi}}-*.gem ${{vars.gem-ffi}}.gem
+          - uses: ruby/install
+            with:
+              gem-file: ${{vars.gem-ffi}}.gem
+              version: ${LLHTTP_FFI_VERSION}
+          - uses: ruby/clean
+
+update:
+  enabled: true
```
mamccorm, line 112: Explicitly adding 'manual:true' here means the automation won't update and auto-merge the changes here. I'd it'll go onto the sustaining list for package updates and approvals

```suggestion
  enabled: true
  manual: true
```

mamccorm (approved):
thanks for this

PR #27111, opened 2024-08-26
Adds cxxopts v3.2.1

srishtih:
Added to provided runtime dependency for falco
https://github.com/wolfi-dev/os/pull/26622

### DIFF
--- a/cxxopts.yaml
+++ b/cxxopts.yaml
@@ -24,10 +24,11 @@ pipeline:
       expected-commit: 4bf61f08697b110d9e3991864650a405b3dd515d
       recurse-submodules: true
 
-  - runs: |
-      mkdir build && cd build
-      cmake -DCMAKE_INSTALL_PREFIX=/usr ..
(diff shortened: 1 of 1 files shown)


cxxopts.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/jarro2783/cxxopts.git
+      tag: v${{package.version}}
+      expected-commit: 4bf61f08697b110d9e3991864650a405b3dd515d
+      recurse-submodules: true
+
+  - runs: |
```
EyeCantCU, line 27: We should be able to do this here instead:

```
- uses: cmake/configure

- uses: cmake/build

- uses: cmake/install
```

srishtih: This worked. melange convert suggests autoconf pipelines and I ran into issues there. I'll try to check out what the issue with that is.

EyeCantCU (approved):

PR #27112, opened 2024-08-26
Adds cpp-httplib v0.16.3

srishtih:
Added to provided runtime dependency for falco
https://github.com/wolfi-dev/os/pull/26622

### DIFF
--- a/cpp-httplib.yaml
+++ b/cpp-httplib.yaml
@@ -24,10 +24,11 @@ pipeline:
       expected-commit: af56b7ec0ba9d18b788ed064f76d936eb52c5db5
       recurse-submodules: true
 
-  - runs: |
-      mkdir build && cd build
-      cmake -DCMAKE_INSTALL_PREFIX=/usr ..
-      make install DESTDIR="${{targets.destdir}}"
+  - uses: cmake/configure
+
+  - uses: cmake/build
+
+  - uses: cmake/install
 
   - uses: strip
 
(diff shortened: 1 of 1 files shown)


cpp-httplib.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/yhirose/cpp-httplib.git
+      tag: v${{package.version}}
+      expected-commit: af56b7ec0ba9d18b788ed064f76d936eb52c5db5
+      recurse-submodules: true
+
+  - runs: |
```
EyeCantCU, line 27:
```
- uses: cmake/configure
- uses: cmake/build
- uses: cmake/install
```

Should work

srishtih: This worked. `melange convert` suggests autoconf pipelines and I ran into issues there. I'll try to check out what the issue with that is.

cpp-httplib.yaml
```diff
@@ -0,0 +1,40 @@
...
+      recurse-submodules: true
+
+  - uses: cmake/configure
+
+  - uses: cmake/build
+
+  - uses: cmake/install
+
+  - uses: strip
+
```
EyeCantCU, line 34: We should create a dev subpackage here

EyeCantCU (approved):

PR #27183, opened 2024-08-27
Add python versioned prometheus-client

srishtih:


py3-prometheus-client.yaml
```diff
@@ -17,7 +27,7 @@ environment:
       - busybox
       - ca-certificates-bundle
       - py3-setuptools
```
EyeCantCU, line 29: Believe this should be `py3-supported-setuptools`

srishtih: Ah, yes. I missed it. I think the py/pip-build-install pipeline actually provides it, that's probably why the build went through. Good catch!

EyeCantCU (approved):

PR #27188, opened 2024-08-27
Adds supported- subpackage for numpy, pybind11 and pythran 

srishtih:
These are needed for scipy build.


justinvreeland:
I have most of these already and scikit build they're all stuck in a huge PR right now that I need to split up though.

https://github.com/wolfi-dev/os/pull/26855

PR #27194, opened 2024-08-27
py3-meson-python: Multiversionize

justinvreeland:


py3-meson-python.yaml
```diff
@@ -22,27 +33,51 @@ environment:
...
     with:
       repository: https://github.com/mesonbuild/meson-python
       tag: ${{package.version}}
       expected-commit: 4f50bf27ef5282f8147c1df073d27859bda5e890
 
-  - runs: |
-      python3 -m gpep517 build-wheel --wheel-dir dist --output-fd 3 3>&1 >&2
-      python3 -m installer -d "${{targets.destdir}}" dist/*.whl
-
   - uses: strip
```
smoser, line 82: Can you
1. move pipeline above subpackages (wierd that 'yam' didn't do that)
2. remove 'uses strip' since... this doesn't do aything at all.

smoser (changes requested):

srishtih:
Epoch needs to be bumped

srishtih:
Do we need to retain runtime packages for the main package?

justinvreeland:
You right I need to do both of those.

srishtih:
Epoch needs to be bumped

smoser (approved):
lgtm

PR #27196, opened 2024-08-27
py3-mdurl: previous change didn't bump epoch

justinvreeland:


smoser (changes requested):
Can you fix the update section  please?
you probably just need to add 'use-tag: true'

smoser (approved):

PR #27245, opened 2024-08-28
fix(logstash CVE): GHSA-2m96-52r3-2f3g

debasishbsws:
update the Gemfile template to force using the fix version of fugit gem


jamie-albert (approved):
Remaining CVE ticket already in queue

PR #27256, opened 2024-08-28
py3-cachecontrol: Multiversionize

justinvreeland:


py3-cachecontrol.yaml
```diff
@@ -31,10 +37,55 @@ pipeline:
       tag: v${{package.version}}
       expected-commit: e2be0c2f8e7c61ab5e2611dd58748f803f520d4d
 
-  - name: Python Build
-    uses: python/build-wheel
+subpackages:
+  - range: py-versions
+    name: py${{range.key}}-CacheControl
```
pnasrat, line 42: recommend normalize the package to lower case and we should fix anything that requires the CamelCase by name.

pnasrat (changes requested):

pnasrat (approved):
LGTM

PR #27278, opened 2024-08-28
gobject-introspection - depend on py3-setuptools-73

smoser:
gdk-pixbuf was found to be FTBFS, failing with a stacktrace like:

> ```
> Traceback (most recent call last):
>   File "/usr/bin/g-ir-scanner", line 103, in <module>
>     from giscanner.scannermain import scanner_main
>   File "/usr/lib/gobject-introspection/giscanner/scannermain.py", line 35, in <module>
>     from giscanner.ast import Include, Namespace
>   File "/usr/lib/gobject-introspection/giscanner/ast.py", line 27, in <module>
>     from .sourcescanner import CTYPE_TYPEDEF, CSYMBOL_TYPE_TYPEDEF
>   File "/usr/lib/gobject-introspection/giscanner/sourcescanner.py", line 25, in <module>
>     from .ccompiler import CCompiler
>   File "/usr/lib/gobject-introspection/giscanner/ccompiler.py", line 29, in <module>
>     from distutils.msvccompiler import MSVCCompiler
> ModuleNotFoundError: No module named 'distutils.msvccompiler'
> ```

Upstream issue to follow gobject-introspection removeing use of distutils is
 https://gitlab.gnome.org/GNOME/gobject-introspection/-/issues/515
or https://gitlab.gnome.org/GNOME/gobject-introspection/-/issues/395

### DIFF
--- a/gobject-introspection.yaml
+++ b/gobject-introspection.yaml
@@ -1,7 +1,7 @@
 package:
   name: gobject-introspection
   version: 1.80.1
-  epoch: 0
+  epoch: 1
   description: Introspection system for GObject-based libraries
   copyright:
     - license: LGPL-2.0-or-later AND GPL-2.0-or-later AND MIT
@@ -49,6 +49,10 @@ pipeline:
     with:
       expected-sha256: a1df7c424e15bda1ab639c00e9051b9adf5cea1a9e512f8a603b53cd199bc6d8
       uri: https://download.gnome.org/sources/gobject-introspection/${{vars.mangled-package-version}}/gobject-introspection-${{package.version}}.tar.xz
+
+  - uses: patch
+    with:
+      patches: fix-giscanner-with-setuptools-74.patch
 
   - uses: meson/configure
     with:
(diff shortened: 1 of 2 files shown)


py3-setuptools-73.yaml
```diff
@@ -0,0 +1,72 @@
+package:
+  name: py3-setuptools-73
```
luhring, line 2: Maybe a dumb question, if this is a version stream, should it have a `provides` of `py3-setuptools=73` or something like that?

smoser: Not a dumb question.
We could do that. I think the answer is no, because we do not really want to have this package used.  I would hope we can drop it as soon as gobject-introspection changes.

luhring: Ah okay, got it! What would you think of putting a comment somewhere toward the top here to explain that? Like "hey, you might want to version-stream-ize this, but don't! we want to get rid of this package as soon as X....."

smoser: @luhring ... good news. by waiting, now we have a better fix.

luhring: Yay!

luhring (approved):

PR #27281, opened 2024-08-28
meson: use py/pip-build-install pipeline for meson

justinvreeland:


meson.yaml
```diff
@@ -7,7 +7,7 @@ package:
     - license: Apache-2.0
   dependencies:
     runtime:
-      - python3
+      - python-3.12-base
```
smoser, line 10: i think the shebang will find this dep, so you should not need to add if you have python-3.12-base added below.

meson.yaml
```diff
@@ -25,9 +25,9 @@ pipeline:
       expected-sha256: 567e533adf255de73a2de35049b99923caf872a455af9ce03e01077e0d384bed
       uri: https://github.com/mesonbuild/meson/releases/download/${{package.version}}/meson-${{package.version}}.tar.gz
 
-  - runs: python3 setup.py build
+  - runs: python3.12 setup.py build
 
-  - runs: python3 setup.py install --prefix=/usr --root=${{targets.destdir}} --skip-build
+  - runs: python3.12 setup.py install --prefix=/usr --root=${{targets.destdir}} --skip-build
```
smoser, line 30: i think if you change this to py/pip-build-install you can just add a 'python3' to environment packages and it will DTRT.

smoser (dismissed):

smoser (changes requested):
I'd click merge, except 2 things.
1. the commit message doesn't make any sense.
2. you need to bump epoch

smoser (changes requested):
I'm sorry, i knoiw this is painful, but your commit message doesn't make any sense now.

smoser (approved):

PR #27282, opened 2024-08-28
Build curl with kerberbos support.

smoser:


justinvreeland (approved):

smoser:
> @smoser I can't see this in the repos -r4, how can I have the customer test it?

I just landed with @justinvreeland's approval.  I would have liked to test _before_ adding to archive, as I'm not sure it does what it is supposed to do.  I'd prefer not landing changes that don't accomplish their goal.

cjohannsen81:
@smoser Cloudera confirmed that it's working for them!

PR #27033, opened 2024-08-24
py3.12: cherrypick fix for CVE-2024-7592

imjasonh:

PR #27288, opened 2024-08-29
python-3.11/CVE-2024-7592 patch implementation

jamie-albert:
Cherry picking in the git-checkout block causes a out of memory error as the buffer size is exceeded when pulling in the two massive branches of 3.11 and 3.12, instead of adding in the logic of increasing buffer size and the subsequent increased build time, I created a patch of the backported changes instead. The patch is identical between all python versions and has been merged [in 3.13](https://github.com/python/cpython/pull/123103) and [3.12](https://github.com/wolfi-dev/os/pull/27033/files), the only reason this remains an open PR is the fact that it requires a special level of [privilege to merge on these branches](https://github.com/python/cpython/issues/123067#issuecomment-2308560855). All [tests pass on 3.11 commit](https://github.com/python/cpython/pull/123105) that will implement this eventually

### DIFF
--- a/python-3.11.yaml
+++ b/python-3.11.yaml
@@ -1,7 +1,7 @@
 package:
   name: python-3.11
   version: 3.11.9
-  epoch: 8
+  epoch: 9
   description: "the Python programming language"
   copyright:
     - license: PSF-2.0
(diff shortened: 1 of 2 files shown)


xnox (changes requested):
Why would we ship unreviewed backport ahead of cpython upstream shipping it for python3.11 for a low severity CVE? Or any other distro?

Risks being, unreproducible behaviour change not present in any other python 3.11 distribution.

Example status from Debian - https://security-tracker.debian.org/tracker/CVE-2024-7592
[bookworm] - python3.11 <postponed> (Minor issue, wait until merged into 3.11 branch)

jamie-albert:
I agree with your points, the only way it would make sense would maybe be for a **critical** CVE approaching SLA. Here is [the advisory ](https://github.com/wolfi-dev/advisories/pull/7330)

PR #27311, opened 2024-08-29
FTBS: lm-sensors

justinvreeland:


lm-sensors/fix-char-const.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/prog/sensord/rrd.c b/prog/sensord/rrd.c
+index a531362b..36d685fc 100644
```
xnox, line 2: is this artisian patch, or available upstream? cause we can also do cherry-picks if available. And would be nice to submit upstream if they don't already have this.

xnox (changes requested):
Also needs epoch bump

lm-sensors/fix-char-const.patch
justinvreeland: Upstream has two version of this patch actually. Neither has been accepted yet.

https://github.com/lm-sensors/lm-sensors/pull/497
https://github.com/lm-sensors/lm-sensors/pull/483

xnox: @justinvreeland trivial, cool thanks. Sometimes we add provenance which one we take; or even use cherry pick pipeline option to fetch those. For example

```yaml
  - uses: git-checkout
    with:
      expected-commit: 4f269258b4237a342da3e9891045cdd51debb27c
      repository: https://github.com/mamba-org/mamba
      tag: ${{package.version}}
      cherry-picks: |
        pull/3386/head/c1fabca58d05ac192466420659abd922b4481012: Backport fmt 11 support from main
```

xnox (dismissed):

lm-sensors/fix-char-const.patch
```diff
@@ -0,0 +1,22 @@
...
+--- a/prog/sensord/rrd.c
++++ b/prog/sensord/rrd.c
+@@ -299,7 +299,7 @@ int rrdInit(void)
```
xnox, line 5: Please git rm the patch then

xnox (changes requested):

xnox (approved):

PR #27445, opened 2024-09-03
kubeflow-pipelines: fix GHSA-952p-6rrq-rcjv

hectorj2f:
fix GHSA-952p-6rrq-rcjv


hectorj2f:
I think I have missed one more dependency of this library.

cpanato (approved):

PR #27470, opened 2024-09-03
added a libestr library package needed for rsyslog build

dustinkirkland:
This library is a build dependency of rsyslog, which I'm trying to package.


dustinkirkland:
@EyeCantCU You might want to look at this one too, as it's also related to the rsyslog build deps...

EyeCantCU (approved):

PR #27471, opened 2024-09-03
added a libfastjson package, a build dependency of rsyslog

dustinkirkland:
This package is a build dependency of rsyslog


libfastjson.yaml
```diff
@@ -0,0 +1,65 @@
...
+
+  - runs: |
+      libtoolize --copy
+      aclocal -I m4 --install
+      autoheader
+      autoconf
+      automake --foreign --add-missing --force-missing --copy
+      ./configure --prefix=/usr
+      make -j $(nproc)
+      make install
```
ajayk, line 34:
```suggestion
      make DESTDIR="${{targets.destdir}}" install
```

dustinkirkland: Yes!  Thanks!!!  I was trying to figure that out just now...

libfastjson.yaml
```diff
@@ -0,0 +1,65 @@
...
+      repository: https://github.com/rsyslog/libfastjson
+      tag: v${{package.version}}
+
+  - runs: |
+      libtoolize --copy
+      aclocal -I m4 --install
+      autoheader
+      autoconf
+      automake --foreign --add-missing --force-missing --copy
+      ./configure --prefix=/usr
```
jamonation, line 32: I think this (these packages in general) might default to statically linking based on the alpine configure flag here: https://git.alpinelinux.org/aports/tree/main/libfastjson/APKBUILD#n30

I'm not sure though if the make command builds both the .so and .a files. I can check in a bit to confirm

libfastjson.yaml
```diff
@@ -0,0 +1,65 @@
...
+      make -j $(nproc)
```
EyeCantCU, line 33: We can use our `autoconf/make` pipeline here

libfastjson.yaml
```diff
@@ -0,0 +1,65 @@
...
+      make DESTDIR="${{targets.destdir}}" install
```
EyeCantCU, line 34: `autoconf/make-install` can be used here

libfastjson.yaml
```diff
@@ -0,0 +1,65 @@
...
+      make check
```
EyeCantCU, line 35: Any validation we do should ideally happen in our `test` pipeline

EyeCantCU (changes requested):

libfastjson.yaml
```diff
@@ -0,0 +1,65 @@
...
+  epoch: 0
+  description: a fast json library for C
+  copyright:
+    - license: MIT
+
+environment:
+  contents:
+    packages:
+      - autoconf
+      - automake
```
EyeCantCU, line 13: Oh, and when using our `autoconf` pipelines, we don't need to include either of these. We pull them in as part of the pipelines

libfastjson.yaml
dustinkirkland: Hmm, I agree that run-time checks should happen in the "test" pipeline.  But this is a build-time check.

The source code and Makefile have to be available to run this set of tests supplied by upstream.  We have this in the build pipeline, but this source code isn't available to us in the run time check.  Respectfully, I disagree and recommend keeping the "make check" tests here.

libfastjson.yaml
dustinkirkland: We do get the .so and .la files from this build, but we don't get the .a files...  Do we need those too, @jamonation ?
```

/usr/lib/libfastjson.so.4.3.0
/usr/lib/libfastjson.so.4
/usr/lib/libfastjson.la

```

libfastjson.yaml
dustinkirkland: I sure wish our `autoconf/make` pipeline did `-j $(nproc)` by default when building locally for larger builds...  (Not this one, but others...)

libfastjson.yaml
EyeCantCU: Sounds good to me :)

libfastjson.yaml
EyeCantCU: It does :) - https://github.com/chainguard-dev/melange/blob/4ecc73b9b01f3a4f996836ccbc10dcac60ea3014/pkg/build/pipelines/autoconf/make.yaml#L20

libfastjson.yaml
EyeCantCU: We should remove the *.la files, those are only useful for libtool

EyeCantCU: We shouldn't need static libs, shared is fine

libfastjson.yaml
dustinkirkland: Okay, updated to use `autoconf/make`

libfastjson.yaml
dustinkirkland: Updated to use `autoconf/make-install`

libfastjson.yaml
dustinkirkland: I've updated to use the auto pipelines, but I've left the upstream `make check` as part of the build.  I'm happy for the build to fail, if upstreams build-time tests are not able to pass cleanly.

libfastjson.yaml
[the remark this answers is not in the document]
dustinkirkland: Updated to use `autoconf/configure`

libfastjson.yaml
dustinkirkland: Done.  However, I did have to leave `libtool` in there, as the build fails otherwise.  Should we have our `autoconf` pipeline pull in `libtool` too?

EyeCantCU (approved):

libfastjson.yaml
EyeCantCU: Our `autoconf/make-install` pipeline removes them for us, marking as resolved - https://github.com/chainguard-dev/melange/blob/4ecc73b9b01f3a4f996836ccbc10dcac60ea3014/pkg/build/pipelines/autoconf/make-install.yaml#L26

PR #27515, opened 2024-09-04
Add kserve package with kserve-agent kserve-router

smoser:


rawlingsj:
Just going to pop this into draft and collaborate on this.  Will likely change this to use a `data:` and `range:` as there are more subpackages to add from this repo.

smoser:
@rawlingsj go-components added as you suggested (thanks for that good idea). looking at storage-controller now.

rawlingsj (approved):

PR #27522, opened 2024-09-04
test build glibc

smoser:


smoser:
this was just to see if glibc was building on aarch64. it seems fine.

PR #27523, opened 2024-09-04
simiplify this build using our autoconf pipeline

dustinkirkland:
Use our autoconf pipeline as much as possible

### DIFF
--- a/liblogging.yaml
+++ b/liblogging.yaml
@@ -1,7 +1,7 @@
 package:
   name: liblogging
   version: 1.0.7
-  epoch: 0
+  epoch: 1
   description: an easy to use and lightweight signal-safe logging library
   copyright:
     - license: BSD-2-Clause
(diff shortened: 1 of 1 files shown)


dustinkirkland:
@EyeCantCU You might want to have a look at this PR, based on your feedback on the libfastjson one...

liblogging.yaml
```diff
@@ -26,16 +24,16 @@ pipeline:
...
+  - uses: autoconf/configure
+
   - runs: |
-      libtoolize --copy
-      aclocal -I m4 --install
-      autoupdate
-      autoheader
-      autoconf
-      automake --foreign --add-missing --force-missing --copy
       ./configure --prefix=/usr --disable-journal --enable-stdlog
```
EyeCantCU, line 30: We can remove this as it's handled by the configure pipeline above it

EyeCantCU (changes requested):

liblogging.yaml
dustinkirkland: Hmm, how to I add the specific --disable and --enable options?

dustinkirkland: I figured it out!

EyeCantCU (approved):

PR #27532, opened 2024-09-04
Add kserve-modelmesh-serving.

wlynch:
Related: https://github.com/chainguard-dev/image-requests/issues/3818

#### For new package PRs only

### DIFF
--- a/kserve-modelmesh-serving.yaml
+++ b/kserve-modelmesh-serving.yaml
@@ -16,7 +16,12 @@ pipeline:
   - uses: go/build
     with:
       packages: .
-      output: modelmesh-manager
+      output: manager
+
+  - name: Copy default config
+    runs: |
+      mkdir -p ${{targets.destdir}}/etc/model-serving/config/
+      cp ./config/default/config-defaults.yaml ${{targets.destdir}}/etc/model-serving/config/default
 
 test:
   environment:
@@ -78,11 +83,11 @@ test:
     - name: Verify installation
       runs: |
         # This works
-        go build -trimpath -o bin/manager main.go
-        ./bin/manager &
+        # go build -trimpath -o bin/manager main.go
+        # ./bin/manager &
 
         # This does not work
-        # modelmesh-manager &
+        manager &
 
         sleep 5
         curl -s localhost:8080/metrics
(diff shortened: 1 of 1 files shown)


kserve-modelmesh-serving.yaml
```diff
@@ -0,0 +1,34 @@
+package:
+  name: kserve-modelmesh-serving
+  version: 0.12.0
+  epoch: 0
+  description: ModelMesh Serving is the Controller for managing ModelMesh, a general-purpose model serving management/routing layer.
+  copyright:
+    - license: Apache-2.0
+
+environment:
```
developer-guy, line 9: you don't need any of these o7

kserve-modelmesh-serving.yaml
```diff
@@ -0,0 +1,34 @@
...
+      repository: https://github.com/kserve/modelmesh-serving.git
+      tag: v${{package.version}}
+      expected-commit: d0e56fa4ac7547e644c87d63dc97a14f612c0391
+
+  - uses: go/build
+    with:
+      packages: .
+      output: manager
+
+  - uses: strip
```
developer-guy, line 28: I think this one also not needed, go/build uses a "-w" flag to strip the binary, afaik.

developer-guy (dismissed):

kserve-modelmesh-serving.yaml
wlynch: Done.

jamonation:
Could this benefit from using the `kwok` test pipeline? The [helm-operator package](https://github.com/wolfi-dev/os/blob/2e718418f25c851dd05271ea014bd3342797b51a/helm-operator.yaml#L49-L67) uses it to good effect. 

At a minimum it would be helpful to add a package level test to validate the binary runs, and ideally a more complete functional test like the kwok example. That said, for this there might need to be some configMap loading, so if it gets too complicated stubbing things out it might not be worth it.

wlynch:
@jamonation Can you point me to more documentation on that? I don't see any mention of it in the melange docs or the wiki.

jamonation:
There's no doc on how to use it @wlynch - but there are a few packages here in Wolfi to crib from as examples. @joshrwolf added the pipeline I believe, so it might be worth asking him for more guidance if there aren't any good examples that would work with this package.

joshrwolf (approved):
I LOVE THESE TESTS

PR #27533, opened 2024-09-04
add tests for grafana

joshrwolf:
adds tests for grafana

also moves around our packaging a bit to match other managers (`homebrew`: https://github.com/Homebrew/homebrew-core/blob/250e7f2925a7e60a58f18488be15043494ead877/Formula/g/grafana.rb)

#13623


cpanato (approved):
cool

PR #27550, opened 2024-09-04
kserve - add kserve-storage-controller, run strip

smoser:
based on
  https://github.com/kserve/kserve/blob/master/python/storage-initializer.Dockerfile

Some notes:
1. storage-initializer says it is <3.12, so we're using 3.11 here.

2. we do not have python3.11 poetry, so a. let 'python3.11 -m pip' install poetry b. let that poetry to build a wheel c. install that wheel with 'python3.11 -m pip'

3. This _does_ use pip to install into the system path.  There are lots of deps we already have that we could use, but for now let pip do it.

### DIFF
--- a/kserve.yaml
+++ b/kserve.yaml
@@ -1,7 +1,7 @@
 package:
   name: kserve
   version: 0.13.1
-  epoch: 1
+  epoch: 2
   description: "Standardized Serverless ML Inference Platform on Kubernetes"
   copyright:
     - license: Apache-2.0
@@ -10,6 +10,8 @@ environment:
   contents:
     packages:
       - go
+      - py3.11-pip
+      - python-3.11-dev
 
 pipeline:
   - uses: git-checkout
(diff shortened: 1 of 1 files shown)


kserve.yaml
```diff
@@ -36,6 +37,7 @@ subpackages:
           modroot: .
           packages: ./cmd/${{range.key}}
           output: ${{range.key}}
+      - uses: strip
```
EyeCantCU, line 40: Definitely a nit, but I think we handle stripping with ldflags we pass in the `go/build` pipeline (`-w`)

EyeCantCU (dismissed):

kserve.yaml
smoser: yeah. on intell it got wheels everywhere.

smoser: fixed.

smoser:
With regard to the CVE scan failures:

```
 🔎 Scanning "/tmp/artifacts-1/packages/x86_64/kserve-storage-controller-0.13.1-r2.apk"
├── 📄 /usr/lib/python3.11/site-packages/ray/_private/runtime_env/agent/thirdparty_files/aiohttp-3.9.3.dist-info/METADATA, 
       /usr/lib/python3.11/site-packages/ray/_private/runtime_env/agent/thirdparty_files/aiohttp-3.9.3.dist-info/RECORD, 
       /usr/lib/python3.11/site-packages/ray/_private/runtime_env/agent/thirdparty_files/aiohttp-3.9.3.dist-info/top_level.txt
│       📦 aiohttp 3.9.3 (python)
│           High CVE-2024-30251 GHSA-5m98-qgg9-wh84 fixed in 3.9.4
│           Medium CVE-2024-27306 GHSA-7gpw-8wmc-pm8g fixed in 3.9.4
│           Medium CVE-2024-42367 GHSA-jwhx-xcg6-8xhj fixed in 3.10.2
│
├── 📄 /usr/lib/python3.11/site-packages/ray/_private/runtime_env/agent/thirdparty_files/idna-3.6.dist-info/METADATA, 
       /usr/lib/python3.11/site-packages/ray/_private/runtime_env/agent/thirdparty_files/idna-3.6.dist-info/RECORD
│       📦 idna 3.6 (python)
│           Medium CVE-2024-3651 GHSA-jjg7-2v4v-x38h fixed in 3.7
│
└── 📄 /usr/lib/python3.11/site-packages/virtualenv/seed/wheels/embed/setuptools-68.0.0-py3-none-any.whl
        📦 setuptools 68.0.0 (python)
            High CVE-2024-6345 GHSA-cx63-2mw6-8hw5 fixed in 70.0.0
```

1. Direct dependency 'ray' has a enbedded copy of aiohttp and idna.  Installed version of ray is 2.10.0. Latest is [2.35.0](https://github.com/ray-project/ray/tags).  kserve pins `ray[serve]<2.11.0,>=2.10.0"`
2. Indirect dependency virtualenv (kserve -> ray -> virtualenv) has a embedded copy of setuptools wheel

EyeCantCU (approved):

PR #27570, opened 2024-09-04
Adds Kserve/modelmesh package

dakaneye:
Package for: https://github.com/chainguard-dev/image-requests/issues/3819

### DIFF
--- a/kserve-modelmesh.yaml
+++ b/kserve-modelmesh.yaml
@@ -1,0 +1,34 @@
+package:
+  name: kserve-modelmesh
+  version: 0.12.0
+  epoch: 0
+  description: The ModelMesh framework is a mature, general-purpose model serving management/routing layer designed for high-scale, high-density and frequently-changing model use cases.
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - maven-3.9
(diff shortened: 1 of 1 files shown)


kserve-modelmesh.yaml
```diff
@@ -0,0 +1,34 @@
...
+    with:
+      repository: https://github.com/kserve/modelmesh.git
+      tag: v${{package.version}}
+      expected-commit: f8212c75fffba9af22c3f3831ea0a8caade518d2
+
+  - name: Compile
+    runs: |
+      export JAVA_HOME=/usr/lib/jvm/java-17-openjdk
+      mvn -B package -Dfile.encoding=UTF8 -DskipTests=true --file pom.xml
+
```
EyeCantCU, line 28: Could we add a package test here?

dakaneye: @EyeCantCU see comment, what I'd do here would likely duplicate: https://github.com/wolfi-dev/os/pull/27532/files. 

Talked w/ @jamonation and we likely don't want to throw unit tests via `mvn test` in here. One alternative is to play w/ bubblewrap/docker to create a container on the fly and then deploy this alongside `modelmesh-serving` and verify services come up but I thought that we could probably do that all together in the image test.

EyeCantCU (approved):

PR #27577, opened 2024-09-04
gcc: enable zstd compression support for LTO metadata

xnox:
Currently gcc build only supports zlib, with this change zstd is also
supported as typically produced by toolchains from other vendors.

This enables feature parity, and ability to reuse foreign libraries
with LTO information.

    $ gcc --version --verbose 2>&1 | grep 'Supported LTO'
    Supported LTO compression algorithms: zlib zstd

### DIFF
--- a/gcc.yaml
+++ b/gcc.yaml
@@ -1,7 +1,7 @@
 package:
   name: gcc
   version: 14.2.0
-  epoch: 2
+  epoch: 3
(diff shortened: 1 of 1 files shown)


joshrwolf:
missing a lint otherwise lgtm

pnasrat (approved):
LGTM

PR #27596, opened 2024-09-04
add a basic rsyslog package

dustinkirkland:
Adding an rsyslog package.  Very basic, default configure options.  There's a chance we're going to need some more later (e.g. for mysql or other database connectivity too).


dustinkirkland:
There is one (old) [CVE](https://nvd.nist.gov/vuln/detail/CVE-2015-3243) against rsyslog -- potentially world-readable permissions on log files.  Red Hat solved it by creating the log file in the distro with 600 root:root permissions.  Presumably we'll do the same?

EyeCantCU (approved):

PR #27655, opened 2024-09-05
gcc: improve openssf-compiler-options integration

xnox:
Turns out even an empty /specs file, causes significant changes to
built-in specs (i.e. default linked libraries) and thus it has to be a
complete spec file.

During compile time dump the full spec file, ship it as default.specs.

Always ship "specs" file, which is the same as default.specs, but with
extra include from /etc.

Later openssf-complier-options will switch to /etc location to
complete the transition.


xnox:
We use wrapper for now; and await better upstream integration for later.

PR #27678, opened 2024-09-05
add argo-events

cpanato:


argo-events.yaml
```diff
@@ -0,0 +1,54 @@
...
+  - uses: git-checkout
+    with:
+      expected-commit: 8dd63f1305746e6503269a7675288b1f65e53df5
+      repository: https://github.com/argoproj/argo-events
+      tag: v${{package.version}}
+
+  - uses: go/build
+    with:
+      ldflags: |
+        -s -w
```
EyeCantCU, line 26: We can remove these. `-s` will disable the symbol table which is useful for govulncheck, `-w` is passed by default in the `go/build` pipeline

argo-events.yaml
```diff
@@ -0,0 +1,54 @@
...
+        -X github.com/argoproj/argo-events.version=${{package.version}}
+        -X github.com/argoproj/argo-events.gitCommit=$(git rev-parse --short=8 HEAD)
+        -X github.com/argoproj/argo-events.buildDate=$(date -u -d "@${SOURCE_DATE_EPOCH:-$(date +%s)}" "+%Y-%m-%dT%H:%M:%SZ")
+        -X github.com/argoproj/argo-events.gitTreeState=clean
+      output: argo-events
+      packages: ./cmd
+
+  - uses: strip
```
EyeCantCU, line 34: Strip is handled via `-w` in ldflags within `go/build`

argo-events.yaml
```diff
@@ -0,0 +1,54 @@
...
+  description: Event-driven Automation Framework for Kubernetes.
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - go
```
EyeCantCU, line 14: Environment can be removed, handled with `go/build`

EyeCantCU (changes requested):

argo-events.yaml
cpanato: I was using the same approach as the other argo package. Should we start removing this from the other packages as well to make it all equal? this can confuse people like me

cpanato: thanks for the review !!

EyeCantCU (approved):

PR #27761, opened 2024-09-06
heimdal - fix FTBFS with gcc~13, add openssf-compiler-options.

smoser:
package does not build with gcc-14.
So build with gcc~13.
Then, while we're there
 * convert configure to 'uses: autoconf/configure' and drop the use of the unset env variable CBUILD and CHOST.
 * add openssf-compiler-options

### DIFF
--- a/heimdal.yaml
+++ b/heimdal.yaml
@@ -1,7 +1,7 @@
 package:
   name: heimdal
   version: 7.8.0
-  epoch: 4
+  epoch: 5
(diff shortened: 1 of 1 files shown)


mamccorm (approved):

smoser:
>  * add openssf-compiler-options

um.... you didn't commit that part @smoser.  
here is the rest: https://github.com/wolfi-dev/os/pull/27783

PR #27784, opened 2024-09-06
loki-3 - fix ftbfs with go-1.23

smoser:
I'm interested in what @xnox feels about this. Move forward to go-1.23  with go-bump  of a [dep](https://github.com/grafana/pyroscope-go/releases/tag/godeltaprof%2Fv0.1.8 ) versus just changing 'go' go 'go-1.22'.

### DIFF
--- a/loki-3.yaml
+++ b/loki-3.yaml
@@ -1,7 +1,7 @@
 package:
   name: loki-3
   version: 3.1.1
-  epoch: 1
+  epoch: 2
   description: Like Prometheus, but for logs.
   copyright:
     - license: AGPL-3.0-or-later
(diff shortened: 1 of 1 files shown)


xnox:
I err on the side of the future. As it is inevitable, and it was not fun unpinning obsolete golangs last week, because they would have become under CVE deadline starting yesterday as vulnerabilities there did affect all prior go toolchains.

xnox (approved):

loki-3.yaml
```diff
@@ -30,6 +30,12 @@ pipeline:
       replaces: go.etcd.io/bbolt=go.etcd.io/bbolt@v1.3.6
       show-diff: true
 
+  # fix compilation with go-1.23
+  # https://github.com/grafana/pyroscope-go/releases/tag/godeltaprof%2Fv0.1.8
+  - uses: go/bump
```
xnox, line 35: @smoser Note for commentary one can always also use "name: Fix compilation failure with go-1.23" instead of a comment. In case you want to be more poetic.

PR #27799, opened 2024-09-06
perl-tk - fix ftbfs for gcc-14.

smoser:
Changes here:
1. apply patches from alpine for gcc-14 and some others.
2. drop an attempt to use CFLAGS.
       this does not work.  replacing that line with
         export CFLAGS="THIS DOES NOT WORK"
       has the exact same behavior as the previous line.
       (the default perl builds get $Config{ccflags}')

### DIFF
--- a/perl-tk.yaml
+++ b/perl-tk.yaml
@@ -1,7 +1,7 @@
 package:
   name: perl-tk
   version: "804.036"
-  epoch: 2
+  epoch: 3
   description: Tk - a Graphical User Interface Toolkit
   copyright:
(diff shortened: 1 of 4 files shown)


smoser:
I put this in draft because I should not need the 'chmod o+w' changes.
I've seen them locally in [db](https://github.com/wolfi-dev/os/pull/27805) also, but they do not represent in c-i.

smoser:
Dropped the chmod o+w change and moving back to review.  I'll chase the bug elsewhere.

justinvreeland (approved):

PR #27832, opened 2024-09-06
Split postgresql-16-bitnami-compat subpackage

jonjohnsonjr:
This is causing a cycle preventing me from merging some [changes into wolfictl](https://github.com/wolfi-dev/wolfictl/pull/1150) because:

postgresql-16 contains subpackage
postgresql-16-bitnami-compat, which depends on
pgaudit-16, which depends on
postgresql-16-dev, which is a subpackage of
postgresql-16

### DIFF
--- a/postgresql-16-bitnami-compat.yaml
+++ b/postgresql-16-bitnami-compat.yaml
@@ -1,0 +1,53 @@
+package:
+  name: postgresql-16-bitnami-compat
+  version: "16.4"
+  epoch: 1
+  description: "compat package with postgresql image"
+  copyright:
+    - license: BSD-3-Clause
+  dependencies:
+    provides:
+      - postgresql-bitnami-compat=${{package.full-version}}
+    runtime:
+      - bash
+      # Required by startup scripts
+      - busybox
+      - net-tools
+      - pgaudit-16
+      - postgresql-16
+
+environment:
+  contents:
+    packages:
+      - busybox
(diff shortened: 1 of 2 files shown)


postgresql-16-bitnami-compat.yaml
```diff
@@ -0,0 +1,53 @@
+package:
+  name: postgresql-16-bitnami-compat
+  version: "16.4"
+  epoch: 0
```
xnox, line 4: Bump this to 1.

postgresql-16.yaml
```diff
@@ -190,47 +190,6 @@ subpackages:
           ln -s /usr/libexec/${{vars.mangled-package-name}}/docker-ensure-initdb.sh ${{targets.subpkgdir}}/usr/bin/docker-ensure-initdb.sh
           ln -s /usr/libexec/${{vars.mangled-package-name}}/docker-ensure-initdb.sh ${{targets.subpkgdir}}/var/lib/postgres/initdb/docker-ensure-initdb.sh
```
xnox, line 192: This package should also be bumped epoch to 1.

Such that there is no set of melange files that build the same full-version that build the same .apk.

xnox (changes requested):
Please bump both packages to epoch r1 such that, there is no set of melange files that both build conflicting .apk by full-version.

xnox (approved):

PR #27864, opened 2024-09-07
py3-gcsfs: fix update block

rawlingsj:
Update check is expected to fail as there is a new version available upstream.  Once this PR merged we will get an automated update PR.

### DIFF
--- a/py3-gcsfs.yaml
+++ b/py3-gcsfs.yaml
@@ -41,9 +41,9 @@ pipeline:
 
 update:
   enabled: true
-  manual: false
   github:
     identifier: fsspec/gcsfs
+    use-tag: true
 
 test:
   pipeline:


py3-gcsfs.yaml
```diff
@@ -41,9 +41,9 @@ pipeline:
 
 update:
   enabled: true
-  manual: false
   github:
     identifier: fsspec/gcsfs
+    use-tag: true
```
kranurag7, line 46: there are two similar tags [2024.9.0post1](https://github.com/fsspec/gcsfs/releases/tag/2024.9.0post1) and [2024.9.0](https://github.com/fsspec/gcsfs/releases/tag/2024.9.0)
thinking if we should avoid the one that have `post1` as suffix.

rawlingsj: yeah - the current config will detect that [2024.9.0](https://github.com/fsspec/gcsfs/releases/tag/2024.9.0) is latest...

as a test to confirm...
```
2024/09/17 11:51:10 INFO attempting to bump version to 2024.9.0
2024/09/17 11:51:10 INFO processing git-checkout node
2024/09/17 11:51:10 INFO   expected-commit: 5cb4479b0ac9737ed99b906befc88c9250cbd21c
```

kranurag7 (approved):

PR #27868, opened 2024-09-07
need libgallium for another build

dustinkirkland:
Add a libgallium package with libgallium.so, needed for another build


dustinkirkland:
Sorry, don't need this.  It's already included in mesa package itself.

PR #27959, opened 2024-09-09
py3-cffi: Enable -supported packages

justinvreeland:


py3-cffi.yaml
```diff
@@ -63,6 +64,20 @@ subpackages:
...
+    dependencies:
+      runtime:
+        - py3.10-${{vars.pypi-package}}
+        - py3.11-${{vars.pypi-package}}
+        - py3.12-${{vars.pypi-package}}
+    test:
+      pipeline:
+        - uses: python/import
+          with:
+            python: python3.10
```
EyeCantCU, line 78:
```suggestion
            python: python${{range.key}}
```

EyeCantCU (changes requested):

py3-cffi.yaml
justinvreeland: This is not correct. This is outside the range specifier so won't build.

```
2024/09/09 08:58:05 ERRO failed to build package: compiling build: compiling subpackage "py3-supported-cffi" tests: compiling Pipeline[0]: mutating with: variable range.key not defined
make[1]: *** [packages/aarch64/py3-cffi-1.17.1-r1.apk] Error 1
```

The goal here is to make sure that the metapackage is installing the libraries. Just testing 3.10 is sufficient for that.

EyeCantCU (approved):

PR #27968, opened 2024-09-09
loki-3: rename to loki-3.1

imjasonh:
Also clean up loki-3.0 directory no longer used.


loki-3.yaml
EyeCantCU, line 1: We should drop this

EyeCantCU (approved):

PR #28014, opened 2024-09-10
chore(gdal): Build for all supported Python versions in single package def

EyeCantCU:
Restructure gdal to build for all supported Python versions within a single package definition


debasishbsws:
Thanks for this @EyeCantCU. Looks like there are some conflicts that need some work.

gdal.yaml
```diff
@@ -1,10 +1,20 @@
...
+  dependencies:
+    provider-priority: 0
+
+data:
+  - name: py-versions
+    items:
+      3.10: "310"
+      3.11: "311"
+      3.12: "312"
+      3.13: "300"
```
philroche, line 17: This looks odd - why 300?

philroche (changes requested):

gdal.yaml
philroche: Do we need to add it then because will `300` need updating once 3.13 is ready?

EyeCantCU: 3.13 builds have been added for most of our existing Python packages. If someone were to use 3.13, then they'd be able to do so without having multiple versions installed. I'm indifferent, but this is how it's handled elsewhere. I think the idea is to mass update 3.13's priority to 313 from 300 when ready

philroche (approved):

debasishbsws (approved):

gdal.yaml
```diff
@@ -66,37 +76,77 @@ pipeline:
...
+    dependencies:
+      provides:
+        - py3-gdal
+      provider-priority: ${{range.value}}
     pipeline:
       - runs: |
-          mkdir -p ${{targets.subpkgdir}}/usr/lib/
-          mv ${{targets.destdir}}/usr/lib/python3* ${{targets.subpkgdir}}/usr/lib/
+          mkdir -p ${{targets.contextdir}}/usr/lib/
+          mv ${{targets.outdir}}/gdal-py${{range.key}}/usr/lib/python3* ${{targets.contextdir}}/usr/lib/
```
imjasonh, line 127: `targets.outdir` is not defined, and our linter didn't catch it. I'm trying to fix that in https://github.com/wolfi-dev/os/pull/32135 but it's failing because that path doesn't exist under `targets.destdir`. Do you happen to see what the error is?

PR #28086, opened 2024-09-10
docker/docker-cli - Resolve circular dep by dropping compat things.

smoser:
This removes the packages:
1. docker-cli-compat
2. docker-modprobe-compat

This should be safe because:
 * nothing rdepends on docker-cli-compat and no images reference docker-modprobe-compat
 * docker-modprobe-compat was being installed in /usr/bin. If you actually tried to call it from /usr/bin, it provides a forkbomb.  It re-execs itself infinitely after trying to use 'ip' and 'lsmod' which aren't installed.

### DIFF
--- a/docker-cli.yaml
+++ b/docker-cli.yaml
@@ -1,7 +1,7 @@
 package:
   name: docker-cli
   version: 27.2.1
-  epoch: 0
+  epoch: 1
   description: The Docker CLI
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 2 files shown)


smoser:
This should resolve the docker* circular deps that are listed https://github.com/chainguard-dev/internal-dev/issues/980

jonjohnsonjr (approved):

jonjohnsonjr:
Thank you!

PR #28104, opened 2024-09-11
add git-sync package

kranurag7:
adds kubernetes git-sync package.

### DIFF
--- a/git-sync.yaml
+++ b/git-sync.yaml
@@ -1,0 +1,32 @@
+package:
+  name: git-sync
+  version: 4.2.4
+  epoch: 0
+  description: A sidecar app which clones a git repo and keeps it in sync with the upstream.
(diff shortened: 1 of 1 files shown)


git-sync.yaml
```diff
@@ -0,0 +1,32 @@
...
+update:
+  enabled: true
+  github:
+    identifier: kubernetes/git-sync
+    strip-prefix: v
+
+test:
+  pipeline:
+    - runs: |
+        git-sync --version
```
Dentrax, line 32: Could we add a one more test here to sync a real repo? https://github.com/kubernetes/git-sync?tab=readme-ov-file#usage

Dentrax (dismissed):

EyeCantCU (approved):

PR #28127, opened 2024-09-11
add libcld2

Dentrax:
Adds `libcld2` and `libcld2-dev` from https://salsa.debian.org/debian/cld2/-/tree/master?ref_type=heads

### DIFF
--- a/libcld2.yaml
+++ b/libcld2.yaml
@@ -1,0 +1,111 @@
+#nolint:valid-pipeline-git-checkout-tag
+package:
+  name: libcld2
+  description: Compact Language Detector 2
+  url: https://github.com/CLD2Owners/cld2
+  version: 0_git20240911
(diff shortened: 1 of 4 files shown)


libcld2.yaml
```diff
@@ -0,0 +1,107 @@
...
+    description: "libcld2 development headers"
+    pipeline:
+      - uses: split/dev
+    dependencies:
+      runtime:
+        - libcld2
+
+update:
+  enabled: false
+  exclude-reason: No releases or tags
```
developer-guy, line 74: this is perfect case where you can use git: {} and schedule for example: https://github.com/wolfi-dev/os/pull/26391

Dentrax: TIL! That's great.

developer-guy (approved):

PR #28144, opened 2024-09-11
add packer-plugin-googlecompute package

kranurag7:
tests are missing because we don't have packer.

### DIFF
--- a/packer-plugin-google.yaml
+++ b/packer-plugin-google.yaml
@@ -1,0 +1,29 @@
+package:
+  name: packer-plugin-googlecompute
+  version: 1.1.6
+  epoch: 0
+  copyright:
+    - license: MPL-2.0
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/hashicorp/packer-plugin-googlecompute
+      tag: v${{package.version}}
+      expected-commit: 901527ae067872d869ea41d2e3aab1db10092626
+
+  - uses: go/bump
+    with:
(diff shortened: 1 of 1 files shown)


packer-plugin-google.yaml
```diff
@@ -0,0 +1,29 @@
...
+  - uses: go/bump
+    with:
+      deps: github.com/hashicorp/go-retryablehttp@v0.7.7
+
+  - uses: go/build
+    with:
+      packages: .
+      output: packer-plugin-google
+      ldflags: -X github.com/hashicorp/packer-plugin-googlecompute/version.Version=${{package.version}}
+
```
EyeCantCU, line 24: We should add some form of a package test here

EyeCantCU (changes requested):

packer-plugin-google.yaml
mamccorm: Re: testing. In the absence of a --help or a --version, can you share info on how we were able to manually test this? It seems like it requires packer to operate, and I don't believe we have a packer apk today

kranurag7: Yes that's right. Given packer is BUSL we don't have packer in here but plugins are MPL so plugins can be added to wolfi but open to thoughts. Plugins can be migrated to other repo too.

mamccorm: Have we been able to do any manual validation that the plugins are detected and usable by packer?

kranurag7:
closing this for now given this requires more work and we don't have confirmation yet on the tag.

PR #28146, opened 2024-09-11
add packer-plugin-azure package

kranurag7:
no tests given we don't have packer

### DIFF
--- a/packer-plugin-azure.yaml
+++ b/packer-plugin-azure.yaml
@@ -1,0 +1,29 @@
+package:
+  name: packer-plugin-azure
+  version: 2.1.8
+  epoch: 0
+  copyright:
+    - license: MPL-2.0
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/hashicorp/packer-plugin-azure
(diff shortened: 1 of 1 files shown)


packer-plugin-azure.yaml
```diff
@@ -0,0 +1,29 @@
...
+  - uses: go/bump
+    with:
+      deps: github.com/hashicorp/go-retryablehttp@v0.7.7
+
+  - uses: go/build
+    with:
+      packages: .
+      output: packer-plugin-azure
+      ldflags: -X github.com/hashicorp/packer-plugin-azure/version.Version=${{package.version}}
+
```
mamccorm, line 24: Re: testing. In the absence of a --help or a --version, can you share info on how we were able to manually test this? It seems like it requires packer to operate, and I don't believe we have a packer apk today

mamccorm: Have we been able to do any manual validation that the plugins are detected and usable by packer?

kranurag7:
closing this for now given this requires more work and we don't have confirmation yet on the tag.

PR #28211, opened 2024-09-12
postgresql: support libxml

imjasonh:


xnox:
Checks out https://salsa.debian.org/postgresql/postgresql-common/-/blob/master/server/postgresql.mk?ref_type=heads#L40

Separately maybe we shall check/benchmark the features we have on; versus debian/ubuntu/fedora/suse/PostgresPRO.

xnox (approved):

PR #28269, opened 2024-09-12
py3-aiohttp: Multiversionize

justinvreeland:


smoser (changes requested):
Can you drop the `charset-normalizer` and `multidict` deps?

they're not listed as required https://pypi.org/project/aiohttp/ . I'm guessing they got dropped somewhere. 

other than that, lgtm.

smoser (approved):

PR #28311, opened 2024-09-13
chore(kafka): Stream

EyeCantCU:
Create Kafka 3.8 stream and use correct version in Bitnami compat package


kafka-3.8.yaml
```diff
@@ -50,13 +58,16 @@ pipeline:
...
     description: "compat package with bitnami/kafka image"
+    dependencies:
+      provides:
+        - kafka-bitnami-compat=${{package.full-version}}
     pipeline:
       - uses: bitnami/compat
         with:
           image: kafka
-          version-path: 3.6/debian-12
+          version-path: 3.8/debian-12
```
mamccorm, line 70: Can we use ${{vars.major-minor-version}} here?

EyeCantCU: We can't use it here unfortunately. We have to be explicit

mamccorm (approved):

PR #28322, opened 2024-09-13
py3-seceretstorage: Mulitversionize

justinvreeland:
This is a runtime dep for py3-keyring which is a runtime dep for py3-poetry

### DIFF
--- a/py3-secretstorage.yaml
+++ b/py3-secretstorage.yaml
@@ -2,17 +2,22 @@
 package:
   name: py3-secretstorage
   version: 3.3.3
-  epoch: 4
+  epoch: 5
   description: Python bindings to FreeDesktop.org Secret Service API
   copyright:
     - license: BSD-3-Clause
   dependencies:
-    runtime:
-      - py3-cryptography
-      - py3-jeepney
-      - python3
-    provides:
-      - py3-SecretStorage=${{package.full-version}}
+    provider-priority: 0
+
+vars:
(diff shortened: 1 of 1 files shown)


py3-secretstorage.yaml
```diff
@@ -33,13 +35,38 @@ pipeline:
...
+          with:
+            python: python${{range.key}}
+            import: secretstorage
 
-  - uses: strip
+test:
+  pipeline:
+    - uses: python/import
+      with:
+        import: secretstorage
```
jamonation, line 69: I'm more curious since the ranging over versions thing is somewhat new to me - but is this test pipeline needed? It looks like all the versions should already be tested in the subpackage test that does the import?

justinvreeland: It's not strictly speaking needed but it is helpful.  This file produces a package called `py3-secretstorage` that's empty and mostly useless.   I don't know how to stop this file from producing this package and there's always the possibility someone is using `py3-secretstorage` as a dep.  This test ensures that if you install `py3-secretstorage` you are given a package that actually installs the library and a workable python version.  So it's testing the `py3-secretstorage` metapackage/priority provider functionality rather than the `py3.XX-secretstorage` packages.  I don't specify a python version because it should only pull in a single version. If multiple are being pulled in to satisfy the deps it should fail and be investigated.

jamonation (approved):

PR #28359, opened 2024-09-13
cni-plugins: add compat for aws-k8s-cni

sanghanan:
- Added missing [`tap`](https://www.cni.dev/plugins/current/main/tap/) and [`vrf`](https://www.cni.dev/plugins/current/meta/vrf/) to the plugin groups.
- Added a compat package for `aws-k8s-cni-init`. All the plugins need to be copied into the `/init` directory. The entrypoint script then copies them to the host, so we can't use `ln`. See: https://github.com/chainguard-dev/image-requests/issues/4138
- Added simple `--version` test
- Refactored the plugins list into `vars`

### DIFF
--- a/cni-plugins.yaml
+++ b/cni-plugins.yaml
@@ -55,44 +55,38 @@ data:
       tap: Creates a tap device inside the container namespace.
       vrf: Creates a VRF in the network namespace and assigns it the interface passed in the arguments.
 
+vars:
+  all-plugins: bridge bandwidth dhcp dummy firewall host-device host-local ipvlan loopback macvlan ptp portmap sbr static tap tuning vrf vlan
+  main-plugins: bridge dummy host-device ipvlan loopback macvlan ptp tap vlan
+  ipam-plugins: dhcp host-local static
+  meta-plugins: bandwidth firewall portmap sbr tuning vrf
+
 # CNI Plugins, separated into groups
 subpackages:
   - name: "cni-plugins-main"
     pipeline:
       - runs: |
           mkdir -p "${{targets.subpkgdir}}"/usr/bin
-          for f in bridge \
-                    ipvlan \
-                    loopback \
-                    macvlan \
-                    ptp \
-                    vlan \
-                    host-device \
-                    tap \
-                    dummy; do
-            cp bin/$f "${{targets.subpkgdir}}"/usr/bin/
+          for binary in ${{vars.main-plugins}}; do
+            cp bin/$binary "${{targets.subpkgdir}}"/usr/bin/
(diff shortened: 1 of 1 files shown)


cni-plugins.yaml
```diff
@@ -96,6 +100,10 @@ subpackages:
       - runs: |
           mkdir -p "${{targets.subpkgdir}}"/usr/bin
           cp bin/${{range.key}} "${{targets.subpkgdir}}"/usr/bin
+    test:
+      pipeline:
+        - runs: |
+            ${{range.key}} --version
   - range: plugins
```
EyeCantCU, line 107: nit: let's add spaces between these subpackages

cni-plugins.yaml
```diff
@@ -105,6 +113,31 @@ subpackages:
       - runs: |
           mkdir -p "${{targets.subpkgdir}}"/opt/cni/bin
           ln -s /usr/bin/${{range.key}} "${{targets.subpkgdir}}"/opt/cni/bin/${{range.key}}
+  - name: "cni-plugins-aws-k8s-compat"
+    description: "Compatibility package for aws-k8s-cni"
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/init
+          for f in bandwidth \
```
EyeCantCU, line 121: We should place all of these in a var above and loop over that

EyeCantCU: Example with PostgreSQL: https://github.com/wolfi-dev/os/blob/2f02bc0659d420eab283f8553856a450f1ab9b4d/postgresql-16.yaml#L88

sanghanan: that's neat, will change it!

cni-plugins.yaml
sanghanan: `yam` removes the space between `subpackages`, so CI fails 😞 
adding the space would definitely help readability though

cni-plugins.yaml
sanghanan: ended up doing this for the other subpackages too, definitely makes it easier to maintain

cni-plugins.yaml
sanghanan: The `.yam.yaml` accounts for a gap between the subpackages: https://github.com/wolfi-dev/os/blob/main/.yam.yaml#L3, which makes this confusing

EyeCantCU: That is incredibly strange, but given everything else is in order, let's get this through

EyeCantCU (approved):

PR #28410, opened 2024-09-15
add kapp-controller package

kranurag7:
Fixes:

#### For new package PRs only


developer-guy (dismissed):

kapp-controller.yaml
```diff
@@ -0,0 +1,49 @@
...
+  enabled: true
+  github:
+    identifier: carvel-dev/kapp-controller
+    strip-prefix: v
+
+test:
+  pipeline:
+    - name: help message
+      runs: |
+        kapp-controller --help
```
Dentrax, line 49: Can we add another test using the `test/daemon-check-output` pipeline if possible?

kranurag7:
the CI is failing because I've included some runtime dependencies and those patches needs to be merged first. 
ref: https://github.com/wolfi-dev/os/pull/28426 https://github.com/wolfi-dev/os/pull/28427 https://github.com/wolfi-dev/os/pull/28428

kapp-controller.yaml
kranurag7: Thanks for the review @Dentrax, I'm inclined to adding tests using `test/kwok/cluster` pipeline but given there's some problem w.r.t. that pipeline, I've avoided adding the tests for now. Once that's resolved, happy to add the tests here via subsequent commits/patches.

wlynch:
Did some digging - I don't think we can test this effectively at the package level.

The controller uses a [default server configuration](https://github.com/carvel-dev/kapp-controller/blob/68347a9701ea2bfa14e2ed4677c040c3fd38db77/pkg/apiserver/apiserver.go#L170). By [default](https://github.com/kubernetes/kubernetes/blob/e8d0e006bee3a2d8cdb87f0cfc638bea01f75d75/staging/src/k8s.io/apiserver/pkg/server/options/authentication.go#L242) this does not respect the KUBECONFIG envvar or --kubeconfig flag.
This results in [the server always trying to load an in-cluster config ](https://github.com/diamanticom/apiserver/blob/f2a92f83c30cfb3169b95f548cc98f02e1a1abc5/pkg/server/options/authentication.go#L362-L377), which causes the controller to fail with the error:

2024/09/18 14:17:03 INFO > {"level":"error","ts":"2024-09-18T18:17:03Z","logger":"kc.main","msg":"Exited run with error","error":"Building API server: failed to get delegated authentication kubeconfig: failed to get delegated authentication kubeconfig: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined","stacktrace":"main.main\n\tcarvel.dev/kapp-controller/cmd/controller/main.go:49\nruntime.main\n\truntime/proc.go:272"}
even when trying to set the KUBERNETES_SERVICE_HOST environment variables, it's expecting serviceaccount token projection auth instead of the local kwok cert based auth.

The easiest thing to do would be to run the controller in a proper cluster instead of out-of-band with just the package binary, which is more effective to do in image testing.

wlynch (approved):

PR #28426, opened 2024-09-15
add carvel-dev/vendir package

kranurag7:

Ref: 


wlynch:
@Dentrax can you PTAL here since this is related to the other carvel-dev packages? 🙏

jamonation (approved):

PR #28427, opened 2024-09-15
add carvel-dev/kapp package

kranurag7:

Ref:

### DIFF
--- a/kapp.yaml
+++ b/kapp.yaml
@@ -1,0 +1,38 @@
+package:
+  name: kapp
+  version: 0.63.3
+  epoch: 0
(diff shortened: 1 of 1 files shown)


kapp.yaml
```diff
@@ -0,0 +1,38 @@
...
+  enabled: true
+  github:
+    identifier: carvel-dev/kapp
+    strip-prefix: v
+
+test:
+  pipeline:
+    - name: version check
+      runs: |
+        kapp version | grep ${{package.version}}
```
Dentrax, line 38: Is it possible to run `kapp deploy` here using kwok?

wlynch (approved):

Dentrax (approved):
Nice tests! @kranurag7 🙌

PR #28428, opened 2024-09-15
add carvel-dev/kbld package

kranurag7:

Ref:

### DIFF
--- a/kbld.yaml
+++ b/kbld.yaml
@@ -1,0 +1,34 @@
+package:
+  name: kbld
+  version: 0.44.1
+  epoch: 0
+  description: kbld seamlessly incorporates image building and image pushing into your development and deployment workflows
+  copyright:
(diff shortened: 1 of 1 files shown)


kbld.yaml
```diff
@@ -0,0 +1,34 @@
...
+  enabled: true
+  github:
+    identifier: carvel-dev/kbld
+    strip-prefix: v
+
+test:
+  pipeline:
+    - name: version check
+      runs: |
+        kbld version | grep ${{package.version}}
```
Dentrax, line 34: Is it possible to run simple example on kwok using `kblf -f` as in the [example](https://github.com/carvel-dev/simple-app-on-kubernetes#step-3-building-container-images-locally) without need having `kapp` dep? (or we should merge it first)

jamonation (approved):

PR #28466, opened 2024-09-16
add a pipeline for testing executables

joshrwolf:
pipeline form of https://github.com/wolfi-dev/os/pull/28463.

I don't love that we need to specify `uses` for each call, so maybe this is something we can just bake into `melange`. I recall there being ideas for doing this a while back and we pushed back, but I can't remember why

### DIFF
--- a/pipelines/test/execs.yaml
+++ b/pipelines/test/execs.yaml
@@ -1,0 +1,41 @@
+name: Test package binaries execution
+
+inputs:
+  package:
+    description: The package to test
+    required: true
+  test-params:
+    description: Space-separated list of parameters to test
+    default: "--version --help -v -h -V -H version help"
+  timeout:
+    description: Timeout for each command execution (in seconds)
+    default: "10"
+
+pipeline:
+  - runs: |
+      set -eu
+
+      pkg="${{inputs.package}}"
+      try="${{inputs.test-params}}"
+
+      path_regex=$(echo "$PATH" | sed -e 's/:/|/g' -e 's/^/^/' -e 's/$/\//')
+
+      apk info -L "$pkg" | grep -E "$path_regex" | while read -r bin; do
+        [ -x "/$bin" ] || continue
+        echo "PACKAGE [$pkg] with BINARY [$bin]"
+        ls -halF "/$bin"
+        cmd=$(basename "/$bin")
+        success=false
+        for param in $try; do
+          if timeout ${{inputs.timeout}}s "$cmd" $param; then
+            success=true
+            break
+          fi
+        done
+        if ! $success; then
+          echo "ERROR: All attempts failed for $cmd"
+          exit 1
+        fi
+      done
+
+      echo "Testing complete for package [$pkg]"
--- a/7zip.yaml
+++ b/7zip.yaml
@@ -49,3 +49,9 @@ update:
   enabled: true
   release-monitor:
     identifier: 265148
+
+test:
+  pipeline:
+    - uses: test/execs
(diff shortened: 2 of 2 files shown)


pipelines/test/execs.yaml
```diff
@@ -0,0 +1,46 @@
...
+        [ -x "/$bin" ] || continue
+        echo "PACKAGE [$pkg] with BINARY [$bin]" >&2
+        ls -halF "/$bin" >&2
+        cmd=$(basename "/$bin")
+        for param in $try; do
+          if runcmd "$cmd $param"; then
+            echo "SUCCESS: $cmd $param" >&2
+            break
+          else
+            echo "FAILED: $cmd $param" >&2
```
philroche, line 41: This will result in a lot more failures than we expect. Many commands will not support all of `--version --help -v -h -V -H version help` options but for those that it does support we should test - the others we can ignore - not fail. 

We need to know whether the command supports the option/flag before we can test it. 

Or have I misunderstood how this pipeline is implemented, and only legit failures will be surfaced?

philroche (changes requested):

pipelines/test/execs.yaml
smoser: > only legit failures will be surfaced?

I think as it is, _no_ failures will be surfaced.

joshrwolf: I changed it to just call `runcmd` and let the exit code be parsed.

the hoisting of `test-params` into an input means we can ultimately configure it as we need, but keeping the same defaults as dustins original script

dustinkirkland: Yeah, there's a lot to LOVE about building this into the pipeline itself.  But there are a few other things to solve for...

I do think there's a lot of value in being "explicit" about the binary/executable that we're testing, and the exact parameter that we're calling it with.  My script was intended as a "one-shot" to detect a bunch of those that are currently "successful", and then bake them into the packaging, such that regressions would be detected in the future, if any of those fail after today.

As implemented here, I think you could automatically test: for any given package that installs binaries, find and run all binaries with some innocuous parameters that might succeed (like help and version), and report success when you find one or more of those.  But I don't think this will be able to detect failures or regressions, without storing some state about what's expected to succeed for any given package.  I think that's the albeit very basic "intelligence" that's baked into [this](https://github.com/wolfi-dev/os/pull/28463/).

To do that with your pipeline, I think we'd actually need to pass a dict into each exec test, with both cmd: 7z and param: --help, as well as cmd: 7zz and param: --help.  Then, take a look at the bluez [package](https://gist.github.com/dustinkirkland/7546196e03770e5f5e96bebbcdd9106b#file-pkg_tests-diff-L727), which has a dozen commands, most of which accept --version, but there's one in there that accepts --help.

At that point, I dunno, just having the explicit actual shell commands `rfcomm --version` and `sdptool --help` make a little more sense to me.  Certainly easier to read and understand what they do.

All that said, I do think there's a GREAT opportunity to introduce a new lint check, that:
1. upon building a package
2. looks for any/all binaries that get installed in the path
3. checks for tests for each binary installed in the path
4. for those that have no invocations in the `test:` block (or has no test block)
5. tries running each installed binary with this set of possible candidate parameters (per the logic/script here in this PR)
6. and prints to screen an updated, suggested test block that one should add to their package.yaml
7. and perhaps even (gently) fails the lint check until those binaries that CAN be easily tested, are easily tested in the yaml

Do it slightly more explicitly this way would establish a better baseline for regressions, I think...

PR #28474, opened 2024-09-16
mariadb: provide mysql-client and add tests

Dentrax:
Provide `mysql-client` on `mariadb-client` package as it's the same package.

### DIFF
--- a/mariadb-11.5.yaml
+++ b/mariadb-11.5.yaml
@@ -1,7 +1,7 @@
 package:
   name: mariadb-11.5
   version: 11.5.2
-  epoch: 1
+  epoch: 2
   description: "The MariaDB open source relational database"
   copyright:
     - license: GPL-3.0-or-later
@@ -187,6 +187,7 @@ subpackages:
     dependencies:
       provides:
         - mariadb-client=${{package.full-version}}
+        - mysql-client=${{package.full-version}}
     pipeline:
       - runs: |
           mkdir -p "${{targets.subpkgdir}}"/usr/bin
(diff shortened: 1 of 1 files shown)


mariadb-11.5.yaml
```diff
@@ -187,6 +187,7 @@ subpackages:
     dependencies:
       provides:
         - mariadb-client=${{package.full-version}}
+        - mysql-client=${{package.full-version}}
```
mamccorm, line 190: Hey @Dentrax, any more context on this one? Did we remove a duplicate mysql-client package and add this provides instead?

mamccorm: Could we add a code comment as well to explain why this provides two separate names?

mariadb-11.5.yaml
```diff
@@ -232,3 +233,42 @@ update:
...
+        mysql --version
+        mariadb --version
+        mariadbd --version
+        mysqladmin --version
+        mariadb-admin --version
+    - name: Test mariadbd start and stop
+      uses: test/daemon-check-output
+      with:
+        setup: |
+          useradd $DBUSER
```
developer-guy, line 261: if this is something required maybe we could use post-install scriplet for mariadb-client? does that make sense?

Dentrax: Actually its just for testing, we shouldn't need this in the image since there will be  `accts` set.

xnox (changes requested):
this will break all of our mysql packaging.

PR #28531, opened 2024-09-16
add flag to display the dependency tree

hectorj2f:
Testing displaying the dependency graph before running bump programs.

### DIFF
--- a/pipelines/go/bump.yaml
+++ b/pipelines/go/bump.yaml
@@ -30,6 +30,9 @@ inputs:
   tidy-compat:
     description: "Set the go version for which the tidied go.mod and go.sum files should be compatible"
     default: ""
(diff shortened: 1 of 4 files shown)


pipelines/maven/pombump.yaml
```diff
@@ -0,0 +1,68 @@
+name: Run pombump tool to update versions and properties in a Maven POM file
```
hectorj2f, line 1: I'll remove the pipeline from here. It doesn't belong here.

hectorj2f:
Tested in https://github.com/wolfi-dev/os/actions/runs/10892773416/job/30226452413?pr=28531.

rawlingsj (approved):

PR #28550, opened 2024-09-17
Enable auto-updates for haproxy, bump to v3.0.4

mamccorm:
Switch to using fetch and grab the source tarball, as git clone does not work with the upstream git repo reliably (https://git.haproxy.org/git/haproxy-3.0.git). Namely:

> failed to validate pipeline git-checkout: failed to list refs for https://git.haproxy.org/git/haproxy-3.0.git/: pkt-line 1: invalid hash text: encoding/hex: invalid byte

Re-enables automatic updates

### DIFF
--- a/haproxy-3.0.yaml
+++ b/haproxy-3.0.yaml
@@ -1,6 +1,6 @@
 package:
   name: haproxy-3.0
-  version: 3.0.2
+  version: 3.0.4
   epoch: 0
   description: "A TCP/HTTP reverse proxy for high availability environments"
   copyright:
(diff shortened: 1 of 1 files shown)


mattmoor (dismissed):

xnox:
https://github.com/orgs/haproxytech/repositories?type=all

Is where they pull data plane from. And which has lots of additional things - helm charts and what not

In their dockerfile they use tarball download rather than git checkout https://github.com/haproxytech/haproxy-docker-alpine-quic/blob/main/3.0/Dockerfile

It does seem like they are transitioning to GitHub and haproxy.com rather than haproxy.org.

Possibly worth asking them about it.

Possibly worth considering to monitor git for tags, but download via fetch a tarball.

ajayk (approved):

PR #28595, opened 2024-09-17
krb5: improve configuration options

xnox:
Remove nonexistent without-tcl configure option that does nothing.

Previously TLS configuration was set to automatic, and could be
skipped if OpenSSL somehow becomes incompatible. Explicitely set to
OpenSSL such that it is guaranteed that OpenSSL support is not lost.

Previously built-in cryptography was used in libk5crypto.so.3, but
given openssl is in use for TLS, it is no additional dependency to
switch libk5crypto to use openssl as well.

### DIFF
--- a/krb5.yaml
+++ b/krb5.yaml
@@ -1,7 +1,7 @@
 package:
   name: krb5
   version: 1.21.3
-  epoch: 0
+  epoch: 1
   description: The Kerberos network authentication system
   copyright:
     - license: MIT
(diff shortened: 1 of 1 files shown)


jamonation:
What about `--with-spake-openssl` ?

k4leung4 (approved):

jamonation:
@xnox for spake we should switch to openssl as well. By default it will use an unsupported x25519 curve: https://web.mit.edu/kerberos/krb5-devel/doc/admin/spake.html

xnox:
> @xnox for spake we should switch to openssl as well. By default it will use an unsupported (in some environments) 25519 curve: https://web.mit.edu/kerberos/krb5-devel/doc/admin/spake.html

Most things should have ed25519 these days. Cue openssl. I wonder if that is a dated default.

But yes, should toggle more options if more are needed. Or can be turned on.

PR #28615, opened 2024-09-17
llvm-19: add llvm 19.1 release

ajayk:
Fixes: https://github.com/chainguard-dev/internal-dev/issues/1470

### DIFF
--- a/openssf-compiler-options.yaml
+++ b/openssf-compiler-options.yaml
@@ -1,7 +1,7 @@
 package:
   name: openssf-compiler-options
   version: 20240627
-  epoch: 2
+  epoch: 3
   description: "Compiler Options Hardening Guide for C and C++"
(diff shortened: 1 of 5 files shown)


openssf-compiler-options.yaml
```diff
@@ -17,7 +17,7 @@ pipeline:
     runs: |
       mkdir -p "${{targets.destdir}}"
       cp -r etc usr "${{targets.destdir}}"
```
xnox, line 19: Needs epoch bump.

xnox (changes requested):

openssf-compiler-options.yaml
justinvreeland: 😡 every time I'm so bad at remembering that.

xnox (approved):

PR #28616, opened 2024-09-17
Add python-3.13 rc build.

smoser:
here is a diff against current python-3.12.yaml:

```diff
diff --git a/python-3.12.yaml b/python-3.13.yaml
index 10b30250e..be8ade5e4 100644
--- a/python-3.12.yaml
+++ b/python-3.13.yaml
@@ -1,6 +1,6 @@
 package:
-  name: python-3.12
-  version: 3.12.6
+  name: python-3.13
+  version: 3.13.0_rc2
   epoch: 0
   description: "the Python programming language"
   copyright:
@@ -10,8 +10,8 @@ package:
     memory: 8Gi
   dependencies:
     provides:
-      - python3=${{package.full-version}}
-      - python-3=${{package.full-version}}
+      - python3=0
+      - python-3=0
     runtime:
       - ${{package.name}}-base=${{package.full-version}}

@@ -28,6 +28,7 @@ environment:
       - linux-headers
       - mpdecimal-dev
       - ncurses-dev
+      - openssf-compiler-options
       - openssl-dev
       - readline-dev
       - sqlite-dev
@@ -41,16 +42,24 @@ var-transforms:
     replace: ''
     to: python
   - from: ${{package.version}}
-    match: (\d).(\d+).(\d+)
+    match: (\d).(\d+).(\d+).*
     replace: '$1.$2'
     to: pyversion
+  - from: ${{package.version}}
+    match: '_'
+    replace: ''
+    to: tagname
+  - from: ${{package.version}}
+    match: (\d).(\d+).(\d+).*
+    replace: '$1.$2.$3'
+    to: versionMajMinMic

 pipeline:
   - uses: git-checkout
     with:
-      expected-commit: a4a2d2b0d85273f746c3834ce2753e19c898949d
+      expected-commit: ec610069637d56101896803a70d418a89afe0b4b
       repository: https://github.com/python/cpython.git
-      tag: v${{package.version}}
+      tag: v${{vars.tagname}}

   - name: Force use of system libraries
     runs: |
@@ -150,7 +159,7 @@ subpackages:
     test:
       pipeline:
         - runs: |
-            ${{vars.python}} version-check.py ${{package.version}}
+            ${{vars.python}} version-check.py ${{vars.versionMajMinMic}}
         - name: Verify venv installs expected packages
           runs: |
             set +x
@@ -195,8 +204,8 @@ subpackages:
     description: "python3 development headers"
     dependencies:
       provides:
-        - python3-dev=${{package.full-version}}
-        - python-3-dev=${{package.full-version}}
+        - python3-dev=0
+        - python-3-dev=0
       runtime:
         - ${{package.name}}=${{package.full-version}}
         - ${{package.name}}-base-dev=${{package.full-version}}
@@ -231,8 +240,8 @@ test:
   pipeline:
     - runs: |
         # main package should provide 'python' and 'python3'.
-        python version-check.py ${{package.version}}
-        python3 version-check.py ${{package.version}}
+        python version-check.py ${{vars.versionMajMinMic}}
+        python3 version-check.py ${{vars.versionMajMinMic}}
     - name: Verify working python3 -m venv
       runs: |
         d=$(mktemp -d)

```

### DIFF
--- a/python-3.13.yaml
+++ b/python-3.13.yaml
@@ -1,0 +1,260 @@
+package:
+  name: python-3.13
+  version: 3.13.0_rc2
+  epoch: 0
+  description: "the Python programming language"
+  copyright:
+    - license: PSF-2.0
+  resources:
+    cpu: 8
+    memory: 8Gi
+  dependencies:
+    provides:
+      - python3=0
+      - python-3=0
+    runtime:
+      - ${{package.name}}-base=${{package.full-version}}
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - bzip2-dev
+      - ca-certificates-bundle
+      - expat-dev
+      - gdbm-dev
+      - libffi-dev
+      - linux-headers
+      - mpdecimal-dev
+      - ncurses-dev
+      - openssf-compiler-options
+      - openssl-dev
+      - readline-dev
+      - sqlite-dev
+      - xz-dev
+      - zlib-dev
+
+# creates helpfull python3.M and 3.M variables
+var-transforms:
+  - from: ${{package.name}}
+    match: '-'
+    replace: ''
+    to: python
+  - from: ${{package.version}}
+    match: (\d).(\d+).(\d+).*
+    replace: '$1.$2'
+    to: pyversion
+  - from: ${{package.version}}
+    match: '_'
+    replace: ''
+    to: tagname
+  - from: ${{package.version}}
+    match: (\d).(\d+).(\d+).*
+    replace: '$1.$2.$3'
+    to: versionMajMinMic
+
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: ec610069637d56101896803a70d418a89afe0b4b
+      repository: https://github.com/python/cpython.git
+      tag: v${{vars.tagname}}
+
+  - name: Force use of system libraries
+    runs: |
+      rm -rf Modules/expat \
+        Modules/_ctypes/darwin* \
+        Modules/_ctypes/libffi*
+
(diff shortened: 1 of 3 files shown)


python-3.13.yaml
```diff
@@ -0,0 +1,263 @@
...
+  copyright:
+    - license: PSF-2.0
+  resources:
+    cpu: 8
+    memory: 8Gi
+  dependencies:
+    provider-priority: 0
+    provides:
+      - python3=${{package.full-version}}
+      - python-3=${{package.full-version}}
```
imjasonh, line 15:
```suggestion
```

I don't think we want this yet, until it's meant to become the latest version.

python-3.13.yaml
```diff
@@ -0,0 +1,263 @@
...
+    pipeline:
+      - uses: split/manpages
+
+  - name: "${{package.name}}-dev"
+    description: "python3 development headers"
+    dependencies:
+      provider-priority: 0
+      provides:
+        - python3-dev=${{package.full-version}}
+        - python-3-dev=${{package.full-version}}
```
imjasonh, line 209:
```suggestion
```

ditto, and below.

python-3.13.yaml
smoser: thank you. you're right.

xnox: Or you also rebuild all previous pythons and add a provider priority on them.

Or you make these python3 provides to be =0 rather than `=${{package.full-version}}`

smoser: i went with python3=0 and that seems to work.

xnox (approved):

PR #28631, opened 2024-09-17
apache2 changes

roaksoax:
Add an entrypoint script and customize default config file to run as user.


apache2.yaml
```diff
@@ -129,6 +129,29 @@ subpackages:
...
+    dependencies:
+      runtime:
+        - ${{package.name}}
+        - dash-binsh
+    pipeline:
+      - runs: |
+          set -x
+          # Add the entrypoint
+          mkdir -p ${{targets.subpkgdir}}/usr/bin
+          mv entrypoint.sh ${{targets.subpkgdir}}/usr/bin/entrypoint.sh
```
smoser, line 143: weird to move this. you could copy or `install`. wierd is just strange because i think the file came from the source.

apache2.yaml
```diff
@@ -129,6 +129,29 @@ subpackages:
...
+          chmod 0755 ${{targets.subpkgdir}}/usr/bin/entrypoint.sh
+          # Modify the default config
+          sed -ri \
+            -e 's!^(\s*CustomLog)\s+\S+!\1 /proc/self/fd/1!g' \
+            -e 's!^(\s*ErrorLog)\s+\S+!\1 /proc/self/fd/2!g' \
+            -e 's!^(\s*TransferLog)\s+\S+!\1 /proc/self/fd/1!g' \
+            -e 's!^(\s*User)\s+daemon\s*$!\1 www-data!g' \
+            -e 's!^(\s*Group)\s+daemon\s*$!\1 www-data!g' \
+            "${{targets.destdir}}"/etc/httpd.conf \
+            "${{targets.destdir}}"/etc/extra/httpd-ssl.conf;
```
smoser, line 153: "sedit" changes are hard to read. and can fail silently.
it would be nice if we had some sane way to validate that the changes you intend to make actually were made.

```
update() {
  local cfg="$1" var="$2" val="$3" found=""
  sed -i "s!^(\s*$var)(\s+)\S+!\1\2$val!g' "$cfg" || { echo "failed update $cfg"; return 1; }
  found=$(sed -n "s!^\s*$var\s+(\S+).*!\1!p")
  [ "$val" = "$found" || { echo "tried to update $var to $val, but it didnt work. found $found"; return 1; }
}

for f in /etc/httpd.conf /etc/extra/httpd-ssl.conf; do
  cp "$f" /tmp/dist 
  update CustomLog /proc/self/fd/1
  update ErrorLog /proc/self/fd/2
  ...
  # just to show the  diff output
  diff -u /tmp/dist "$f"
done
```

its not great, but then you don't silently fail to make the edit you want.
alternatively, you could just use patch. I don't know if that would be brittle acrros updates or not.

  
}

smoser (commented):
some way to validate the changes would be really nice.

apache2.yaml
```diff
@@ -83,6 +84,25 @@ pipeline:
...
+      # Modify User/Group and test to ensure changes are applied.
+      sed -ri \
+        -e 's!^(\s*User)\s+daemon\s*$!\1 www-data!g' \
+        -e 's!^(\s*Group)\s+daemon\s*$!\1 www-data!g' \
+        "${{targets.destdir}}"/etc/httpd.conf
+
+      grep -E '^\s*User www-data$' "${{targets.destdir}}"/etc/httpd.conf;
+      grep -E '^\s*Group www-data$' "${{targets.destdir}}"/etc/httpd.conf;
+
+      # Modify CustomLog/ErrorLog/TransferLog (it's ok it fails, no need to test).
```
smoser, line 98: If it is OK if it fails, then why do we do it?
Given the option between 
1. software that isn't necessary and only _might_ do something
2. no software

I will pick 2.

apache2/entrypoint.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+set -e
```
smoser, line 2: set -e is pointless here in a one line script. but ok.

smoser (approved):
thanks @roaksoax

PR #28865, opened 2024-09-19
this commit adds 2207 simple tests across 625 packages

dustinkirkland:
This branch adds 2207 simple tests, across 625 packages that don't yet have any tests.  All of these tests were very simply generated by looking for executables installed in the $PATH, which can take a common parameter such as "--version" or "--help" or similar, and exits with return code 0.  Generally that should give us some regression protection and signal that binaries installed continue to be executable...  At @joshrwolf request, I've added an # AUTOGENERATED comment inline, and I did NOT bump the epochs (so as not to trigger a rebuild).


joshrwolf (dismissed):
I'm going to make the case to force merge this in (after a rebase). The primary goal here is to add some breadth to our archive testing in a basic yet better than nothing way.

Dustin walked me through the setup for how he's testing these packages locally and has managed to reliably make them pass. Given the way our current presubmit is set up its extremely likely things will pass in CI (the disk will fill up long before we make it through the diff). 

This PR intentionally doesn't bump the epoch of any packages, and will rely on the _next_ time the given package is bumped to run this test and surface any errors that may have sneaked through. This is intentional, as its sustaining work that we already commit to. The way our presubmit CI is set up means packages are being rebuilt before being tested in presubmit, and there are a few packages that seem to be failing a rebuild (`cyrus-sasl` the latest). Landing this PR won't trigger a rebuild in main, but will surface as package build (or test if something sneaks in) failures during the bump, which is work we've already committed to.

All that being said, I think as is this PR does far more good than bad. There *may* be a very small set of tests that end up breaking we don't catch here, but they're trivial to fix individually, and the package build errors surfaced here is already work we're committed to resolving.

We *could* batch this in, but I'd argue that's unnecessary toil fighting with our existing presubmit CI limitations, and is needless CI pageantry for things we can catch once this lands as individual package by package cases during the next bump.

philroche (dismissed):
I am +1 on adding these tests. This enables a lot more automated version bumps with >0 confidence that the version bump did not regress. 

Conditional approval assuming conflicts are resolved

joshrwolf (approved):

smoser (approved):
Thanks.

PR #28930, opened 2024-09-19
add akhq package

cpanato:
xref: https://github.com/chainguard-dev/image-requests/issues/4179

### DIFF
--- a/akhq.yaml
+++ b/akhq.yaml
@@ -1,0 +1,71 @@
+package:
+  name: akhq
+  version: 0.25.1
+  epoch: 0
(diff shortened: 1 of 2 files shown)


jamonation (dismissed):

jamonation:
I don't think we have a gradle bump yet @cpanato - but could we try adding bumps to the `gradle.properties` files for commons-compress and vertx?

https://github.com/tchiotludo/akhq/blob/dev/gradle.properties

jamonation (approved):
Nice, 0 CVEs :)

PR #28946, opened 2024-09-19
lmdb: cherry-pick Debian patch to provide a soname

xnox:
Without a soname, package splits; linking; SCA dependencies do not work.

Add "0" soname to lmdb by cherry-picking patch from debian.

Also fix the pc file, as lots of variables got interpreted prior to writing them out. Possibly it would have been a better idea to simply commit .pc file in the overlayd dir.

Once all that is in place, rebuild reverse dependencies to get correct linking with liblmdb.so.0 and runtime dependencies.

This makes postmap be executable without undefined symbols:

before:
```
postmap lmdb:/etc/postfix/aliases
postmap: fatal: load_library_symbols: dlopen failure loading /usr/lib/postfix/postfix-lmdb.so: /usr/lib/postfix/postfix-lmdb.so: undefined symbol: mdb_drop
```

after
```
 ldd /usr/lib/postfix/postfix-lmdb.so 
	linux-vdso.so.1 (0x00007ffd80c9a000)
	liblmdb.so.0 => /usr/lib/liblmdb.so.0 (0x00007c60f87b2000)
	libc.so.6 => /lib/libc.so.6 (0x00007c60f85bd000)
	/lib64/ld-linux-x86-64.so.2 (0x00007c60f87d4000)
```

### DIFF
--- a/lmdb.yaml
+++ b/lmdb.yaml
@@ -29,18 +29,18 @@ pipeline:
   - runs: |
       cd libraries/liblmdb
 
-      cat > lmdb.pc << EOF
+      cat > lmdb.pc << 'EOF'
       prefix=/usr
-      exec_prefix=\${prefix}
-      libdir=\${exec_prefix}/lib
-      includedir=\${prefix}/include
+      exec_prefix=${prefix}
+      libdir=${exec_prefix}/lib
+      includedir=${prefix}/include
 
       Name: liblmdb
       Description: Lightning Memory-mapped key-value database
       URL: http://symas.com/mdb/
       Version: @@VERSION@@
-      Libs: -L\${libdir} -llmdb
-      Cflags: -I\${includedir}
+      Libs: -L${libdir} -llmdb
+      Cflags: -I${includedir}
       EOF
 
       sed -i "s|@@VERSION@@|${{package.version}}|" -i *.pc
--- a/lmdb.yaml
+++ b/lmdb.yaml
@@ -1,8 +1,7 @@
-# Generated from https://git.alpinelinux.org/aports/plain/main/lmdb/APKBUILD
 package:
   name: lmdb
   version: 0.9.31
-  epoch: 0
+  epoch: 1
   description: Lightning Memory-Mapped Database
   copyright:
     - license: OLDAP-2.8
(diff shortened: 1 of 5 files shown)


lmdb.yaml
```diff
@@ -23,21 +22,25 @@ pipeline:
...
       expected-commit: ce201088de95d26fc0da36ba805bf2ddc2ba74ff
 
+  - uses: patch
+    with:
+      patches: 0001-Debian-quirks-to-make-shared-library-versioned.patch
+
   - runs: |
       cd libraries/liblmdb
 
       cat > lmdb.pc << EOF
```
smoser, line 32: fwiw, if you just use `cat >lmbd.pc << "EOF"` this would work.

i think i'd prefer that to more shell escaping.

xnox: TIL!

xnox: >  If any part of word is quoted, the delimiter is the result of quote removal on word, and the lines in the here-document are not expanded. If word is unquoted, all lines of the here-document are subjected to parameter expansion, command substitution, and arithmetic expansion, the character sequence \newline is ignored, and ‘\’ must be used to quote the characters ‘\’, ‘$’, and ‘`’.

from https://www.gnu.org/software/bash/manual/bash.html#Here-Documents

smoser (commented):
looks good. thank you.

To satisfy curiosity, I went looking for some background on this.

Upstream Issue  [7794](https://bugs.openldap.org/show_bug.cgi?id=7794) [2014-01-29] from a fedora maintainer seems to be the first request for adding a version to lmdb.  I'm guessing all distros are just carrying a patch now.  

It is not clear to me if [8034](https://bugs.openldap.org/show_bug.cgi?id=8034#c1) is out of date now or not, but there a maintainer says:

> Just a note:
>
> LMDB should not be pulled separately from OpenLDAP.  I.e., only the bundled 
version of LMDB should be used with a given version of OpenLDAP.

Oh well.
Thank you, @xnox

lmdb.yaml
smoser: shell!

smoser (approved):

smoser (approved):

PR #28944, opened 2024-09-19
Withdraw chromium packages which say they provide libvulkan.so.1

murraybd:
Two gtk-4 tests were recently added by @dustinkirkland and are failing due to libvulkan.so.1 being missing from the test environment. Looking at the package installation log we can see that an old version of chromium is being installed but what we really want is for vulkan-loader to be installed.

My understanding (thanks @smoser) is that withdrawing the versions of chromium which purport to provide libvulkan.so.1 will sort the issue out and the tests will pass.

Test log:
```
2024/09/19 12:51:53 INFO installing chromium (126.0.6478.126-r0)
2024/09/19 12:51:53 INFO installing gtk-4 (4.16.1-r0)
2024/09/19 12:51:54 INFO installing wolfi-keys (1-r8)
2024/09/19 12:51:54 INFO installing apk-tools (2.14.4-r0)
2024/09/19 12:51:54 INFO installing wolfi-base (1-r6)
2024/09/19 12:51:55 INFO built image layer tarball as /tmp/apko-temp-1458869052/apko-x86_64.tar.gz
2024/09/19 12:51:55 INFO using /tmp/apko-temp-1458869052/apko-x86_64.tar.gz for image layer
2024/09/19 12:51:59 INFO populating workspace /tmp/melange-workspace-2728922673 from ./gtk-4/
2024/09/19 12:51:59 INFO ImgRef = /tmp/melange-guest-1559818651
2024/09/19 12:51:59 INFO running the main test pipeline
2024/09/19 12:51:59 WARN + '[' -d /home/build ]
2024/09/19 12:51:59 WARN + cd /home/build
2024/09/19 12:51:59 WARN + gtk4-broadwayd --help
2024/09/19 12:51:59 INFO Usage:
2024/09/19 12:51:59 INFO   gtk4-broadwayd [OPTION?] [:DISPLAY] - broadway display daemon
2024/09/19 12:51:59 INFO
2024/09/19 12:51:59 INFO Help Options:
2024/09/19 12:51:59 INFO   -h, --help                   Show help options
2024/09/19 12:51:59 INFO
2024/09/19 12:51:59 INFO Application Options:
2024/09/19 12:51:59 INFO   -p, --port=PORT              Httpd port
2024/09/19 12:51:59 INFO   -a, --address=ADDRESS        Ip address to bind to 
2024/09/19 12:51:59 INFO   -u, --unixsocket=ADDRESS     Unix domain socket address
[...]

PR #28953, opened 2024-09-19
Add vulkan-loader as a test dep temporarily

murraybd:
An older version of chromium is currently being installed because it purports to provide libvulkan.so.1 (see https://github.com/wolfi-dev/os/pull/28944) while we really need vulkan-loader installed. Until those versions of chromium are withdrawn we can workaround the issue with the change in this PR.


murraybd:
The chromium packages have been withdrawn so this is obsolete.

PR #28955, opened 2024-09-19
neuvector - be explicit about phython version

smoser:
Explicitly set the python version being used here.
Otherwise, the build-time installs things into one version (3.12)
and then later runtime resolution may install 3.13 for dependencies.

Note, the npm install section actually needs python3 executable and
setuptools. Also here is dropping the 'supervisor' dep of the 
neuvector-manager-cli package.  I could not find why that was there,
and it mean tthat we would get 2 python stacks installed.

### DIFF
--- a/neuvector-manager.yaml
+++ b/neuvector-manager.yaml
@@ -1,7 +1,7 @@
 package:
   name: neuvector-manager
   version: 5.4.0
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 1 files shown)


smoser:
This is obnoxious. @EyeCantCU do you have any idea what is going on here? builds just get killed after 360m (quite reasonable). There is only output of the build for the first 3 minutes.

PR #28956, opened 2024-09-19
python deps cleanups

smoser:
*  py3-python-dateutil - provide py3-python-dateutil
    
    This package is poorly named. I'm not sure why it originally
    got the redundant py3-python-dateutil name, but some other packages
    (py3-jupyter-client) have a dependency on that name.
    
    Because the versioned packages did not provide py3-python-dateutil
    (the did provide py3-dateutil and py3.XX-dateutil) installing
    py3-jupyter-client would just leave you with the skeleton un-versioned
    python package.

 *  py3-pbr - make py3.XX-package provide py3-package
    
    These were missing boilerplate from the multi-version python
    packaging.
    
    The result is that if you 'apk add py3-pbr' you only
    get the empty parent package. Also update variable name for 
    consistency.

 *  py3-cryptography - add runtime depends on cffi
    
    This was discovered by py3-paramiko failing.
    Improve this package's test to also fail.

### DIFF
--- a/py3-cryptography.yaml
+++ b/py3-cryptography.yaml
@@ -2,7 +2,7 @@
 package:
   name: py3-cryptography
   version: 43.0.1
-  epoch: 1
+  epoch: 2
   description: cryptography is a package which provides cryptographic recipes and primitives to Python developers.
   copyright:
     - license: Apache-2.0 OR BSD-3-Clause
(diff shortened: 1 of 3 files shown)


smoser:
I split the py3-pygments to https://github.com/wolfi-dev/os/pull/28993 because it was creating a dependency loop.

imjasonh (approved):

PR #28984, opened 2024-09-20
add aws-gateway-controller package

kranurag7:
#### For new package PRs only

### DIFF
--- a/aws-gateway-controller.yaml
+++ b/aws-gateway-controller.yaml
@@ -1,0 +1,46 @@
+package:
+  name: aws-gateway-controller
+  version: 1.0.7
+  epoch: 0
+  description: A Kubernetes controller for Amazon VPC Lattice
+  copyright:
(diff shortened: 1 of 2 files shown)


aws-gateway-controller.yaml
```diff
@@ -0,0 +1,46 @@
+package:
+  name: aws-gateway-controller
```
kranurag7, line 2: the name of the repo is different which is https://github.com/aws/aws-application-networking-k8s/ but I chose this name because the container image was named `public.ecr.aws/aws-application-networking-k8s/aws-gateway-controller`

Dentrax: Let's rename the package name to `aws-application-networking-k8s` for consistency and add `aws-gateway-controller=${{package.full-version}}` to `provides:`.

mamccorm (approved):

PR #28992, opened 2024-09-20
ts-patch: new package

paulgibert:
Adds new package ts-patch to Wolfi.

#### For new package PRs only


ts-patch.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - name: Test tspc version
+      runs: |
+        tspc --version
+    - name: Test tspc compiles
+      runs: |
+        tsfile=$(mktemp /tmp/tempfile.XXXXXX.ts)
+        echo "console.log('Green eggs and ham.');" > $tsfile
+        jsfile=$(mktemp /tmp/tempfile.XXXXXX.js)
+        tspc $tsfile --outFile $jsfile
+        node $jsfile
```
jamonation, line 63:
```suggestion
        node $jsfile |grep "Green eggs and ham."
```

Super small suggestion here - since the application is logging this output, the test can check for it. This ensures that it is actually running the ts application, as opposed to say, silently outputting nothing. Looks great!

jamonation (approved):

PR #29005, opened 2024-09-20
Fix invalid period in schedule

xnox:
Fails wolfictl lint for all


imjasonh (dismissed):

xnox:
sadness need to either fix expected commits or force merge.

imjasonh:
Can you update the commits? These are going to block us anyway otherwise, and we might as well avoid the force merge.

ajayk (approved):

PR #29037, opened 2024-09-20
init apache hadoop package

kranurag7:
Fixes:

Related:

#### For new package PRs only


hadoop.yaml
```diff
@@ -0,0 +1,86 @@
...
+  epoch: 0
+  description: Open Source Identity and Access Management For Modern Applications and Services
+  copyright:
+    - license: Apache-2.0
+  target-architecture:
+    # note: skipping aarch64 because frontend-maven-plugin doesn't work on aarach arch.
+    - x86_64
+  dependencies:
+    runtime:
+      - openjdk-8
```
EyeCantCU, line 13: We don't need the full JDK. Just `openjdk-8-default-jvm`

hadoop.yaml
```diff
@@ -0,0 +1,86 @@
...
+  contents:
+    packages:
+      - bash
+      - busybox
+      - ca-certificates-bundle
+      - maven
+      - nodejs
+      - openjdk-8
+      - wolfi-base
+      - wolfi-baselayout
```
EyeCantCU, line 25: We don't need these

hadoop.yaml
```diff
@@ -0,0 +1,86 @@
...
+    # note: skipping aarch64 because frontend-maven-plugin doesn't work on aarach arch.
+    - x86_64
+  dependencies:
+    runtime:
+      - openjdk-8
+
+environment:
+  contents:
+    packages:
+      - bash
```
EyeCantCU, line 18: We can remove this

hadoop.yaml
```diff
@@ -0,0 +1,86 @@
...
+
+  - uses: maven/pombump
+    with:
+      patch-file: pombump-deps.yaml
+      properties-file: pombump-properties.yaml
+      pom: hadoop-project/pom.xml
+
+  - runs: |
+      tar --version
+      mkdir -p ${{targets.contextdir}}/
```
EyeCantCU, line 46: The destination directory (destdir) already exists. We only need to do this for subpackages. Not to mention, this is handled by the command below it

hadoop.yaml
```diff
@@ -0,0 +1,86 @@
...
+      mkdir -p ${{targets.contextdir}}/usr/share/java
+      mkdir -p ${{targets.contextdir}}/usr/bin
+      mvn -Pdist,src,yarn-ui -DskipTests -Dtar -Dmaven.javadoc.skip=true clean package -e -X
+      tar -xvf hadoop-dist/target/hadoop-${{package.version}}.tar.gz -C ${{targets.contextdir}}/usr/share/java
+      for f in ${{targets.contextdir}}/usr/share/java/hadoop-${{package.version}}/bin/*; do
+        ln -sf $f ${{targets.contextdir}}/usr/bin/$(basename $f)
```
EyeCantCU, line 52: This is no-op. It's going to create links to files that don't exist as contextdir will be accounted for

hadoop.yaml
```diff
@@ -0,0 +1,86 @@
...
+    contents:
+      packages:
+        - bash
+        - openjdk-8-jre
+    environment:
+      JAVA_HOME: /usr/lib/jvm/java-1.8-openjdk
+  pipeline:
+    - name: version check
+      runs: |
+        /usr/share/java/hadoop-${{package.version}}/bin/hadoop version
```
EyeCantCU, line 83: This should be `/usr/share/java/hadoop`, without the package version

hadoop/pombump-properties.yaml
```diff
@@ -0,0 +1,20 @@
+properties:
+  # the following deps can't be bumped because of signature errors.
+  # - property: avro.version
+  #   value: "1.11.3"
+  # - property: dnsjava.version
+  #   value: "3.6.0":
+  # - property: commons-compress.version
+  #   value: "1.26.0"
```
EyeCantCU, line 8: We can remove these, we'll tackle these when we file advisories

EyeCantCU (changes requested):

hadoop.yaml
```diff
@@ -0,0 +1,86 @@
...
+  version: 3.4.0
+  epoch: 0
+  description: Open Source Identity and Access Management For Modern Applications and Services
+  copyright:
+    - license: Apache-2.0
+  target-architecture:
+    # note: skipping aarch64 because frontend-maven-plugin doesn't work on aarach arch.
+    - x86_64
+  dependencies:
+    runtime:
```
EyeCantCU, line 12: There's bundled bash scripts so we should add `bash` at runtime

hadoop.yaml
kranurag7: the following line extracts the built hadoop tarball at `/usr/share/java` and the main hadoop binary lives at `/usr/share/java/hadoop-3.4.0/bin/hadoop` that's being used in tests. Please let me know if I should change something here in terms of placing the binary at some other location either directly or via symlink. 

```bash
tar -xvf hadoop-dist/target/hadoop-${{package.version}}.tar.gz -C ${{targets.destdir}}/usr/share/java
```

EyeCantCU: We should definitely exclude version in the path. I don't see a reason to have multiple versioned installations of Hadoop and this will complicate things when we set the working directory as well need to pass the Hadoop version

hadoop.yaml
```diff
@@ -0,0 +1,115 @@
...
+          cp scripts/transformation.py ${{targets.contextdir}}/opt/transformation.py
+          cp krb5.conf ${{targets.contextdir}}/opt/krb5.conf
+          ln -sf /usr/lib/async-profiler.jar ${{targets.contextdir}}/opt/profiler/build/async-profiler.jar
+          ln -sf /usr/lib/libasyncProfiler.so ${{targets.contextdir}}/opt/profiler/build/libasyncProfiler.so
+
+  - name: ${{package.name}}-compat
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/opt
+          ln -sf /usr/share/java/hadoop ${{targets.contextdir}}/opt
```
EyeCantCU, line 89: Shouldn't this be `/opt/hadoop`?

EyeCantCU (changes requested):

hadoop.yaml
```diff
@@ -0,0 +1,115 @@
...
+        - async-profiler
+    pipeline:
+      - uses: git-checkout
+        with:
+          repository: https://github.com/apache/hadoop
+          # this branch is where the hadoop-runner base image comes from.
+          # for update please look into the PR that are updating the base image.
+          # previous work: https://github.com/apache/hadoop/pull/6910
+          branch: docker-hadoop-runner-jdk11-u2204
+          expected-commit: 5bb6f450340fa3deb931571d08d42a3f358af8bf
```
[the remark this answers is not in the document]
kranurag7: given this is a branch where upstream supports docker related configuration. I didn't find a way to poll based on branches. For now, I think this is stable because they are making new changes on another branch but if there's a way then please let me know about it. I'll implement the same. The central point here is that we need to poll on branches that start with `docker-` and then checkout that and build it.
From my attempt of looking into it, I find that doing it manually is the way to go for now because they also remove something in new branches which we depend on. 
For example: This (ref: https://github.com/apache/hadoop/tree/docker-hadoop-3.4) is latest `docker-` branch but python configuration on which we depend is missing from this branch.

hadoop.yaml
EyeCantCU, line 1: Rename as `hadoop-3.4`

hadoop.yaml
```diff
@@ -0,0 +1,147 @@
+package:
+  name: hadoop
```
EyeCantCU, line 2: Update the name here too

hadoop.yaml
```diff
@@ -0,0 +1,147 @@
...
+      - libtool
+      - maven
+      - openjdk-8
+      - openssl-dev
+      - protobuf-dev~3.21
+      - protobuf~3.21
+      - protoc~3.21
+      - snappy-dev
+      - snappy-static
+      - zlib
```
EyeCantCU, line 49: `zlib-dev` pulls in `zlib`

hadoop.yaml
```diff
@@ -0,0 +1,147 @@
...
+
+  - uses: maven/pombump
+    with:
+      patch-file: pombump-deps.yaml
+      properties-file: pombump-properties.yaml
+      pom: hadoop-project/pom.xml
+
+  - runs: |
+      mkdir -p ${{targets.destdir}}/usr/share/java/hadoop
+      mkdir -p ${{targets.destdir}}/usr/bin
```
EyeCantCU, line 76: This folder is empty, we can remove this

hadoop.yaml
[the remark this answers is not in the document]
EyeCantCU: I'm not sure I follow. We'd create a separate package and then just poll like this:

```yaml
update:
  enabled: true
  schedule: daily
  git: {}
```

We shouldn't need to pass anything else as it will leverage whatever we pass to `git-checkout` for polling

hadoop/fix-build-openssl.patch
EyeCantCU, line 1: Drop this

EyeCantCU (changes requested):

hadoop.yaml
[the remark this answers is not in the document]
kranurag7: thank you for the review. New package is tracked [here](https://github.com/wolfi-dev/os/pull/30294) and I've included the same as runtieme dependency to this package (https://github.com/wolfi-dev/os/pull/30294)

kranurag7:
CI is failing because `hadoop-runner` is not there which was include as runtime dependency to this package. 

Logs: 

`failed to bundle: unable to build graph: hadoop-3.4-3.4.0-r0: unable to resolve dependency hadoop-runner-3.4: could not find package that provides hadoop-runner-3.4 in indexes`

kranurag7:
closing given this is not longer tracked here.

PR #29043, opened 2024-09-20
Add consul-k8s, go-discovery

wlynch:
Testing looks like its better suited for a full k8s environment, so holding off on more detailed functional tests here.

Also includes go-discovery, [which is a runtime dependency of consul-k8s](https://github.com/hashicorp/consul-k8s/blob/0c3745134cbb81198758c442b8d80d61bb351bed/control-plane/Dockerfile#L58). go-discovery is unversioned upstream.

Related: https://github.com/chainguard-dev/image-requests/issues/4271

#### For new package PRs only

Not eol.date, but looks like multiple version streams supported: 1.1 - 1.5

#### For new version streams

### DIFF
--- a/consul-k8s.yaml
+++ b/consul-k8s.yaml
@@ -1,0 +1,57 @@
+package:
+  name: consul-k8s
+  version: 1.5.3
+  epoch: 0
+  description: The consul-k8s includes first-class integrations between Consul and Kubernetes.
+  copyright:
+    - license: MPL-2.0
+  dependencies:
+    runtime:
+      - go-discover
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/hashicorp/consul-k8s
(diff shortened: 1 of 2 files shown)


shimmerjs (dismissed):

jamonation (dismissed):

consul-k8s.yaml
```diff
@@ -0,0 +1,61 @@
...
+      - uses: go/build
+        with:
+          modroot: ./cli
+          output: consul-k8s
+          packages: .
+    test:
+      environment:
+        contents:
+          packages:
+            - go
```
Dentrax, line 44: I think we should not to add `go` to runtime deps, is this required?

wlynch: Good catch. Removed.

jamonation (approved):

PR #29044, opened 2024-09-20
openssf-compiler-options: give up on warnings

xnox:
Turning on warnings causes build failures for no extra gain.

The packages that fail, are the ones that actually already turn on
lots of warnings, and cruitially also turn on Werror. Adding more
warnings (Wall, Wconversion, Wimplicit-fallthrough, Whardened) ends up
causing failures for things that upstream do not monitor yet, or chose
not to monitor.

This is preventing mass enabling the rest of the hardening flags, that
do not fail the builds and genuinely improve binary hardening.

Also build logs are huge as genuinly it seems like everything trips up
Wconversion and Wsign-compare.

Remove addition of any new warnings.


smoser (approved):
this is fine.
We need to follow on to this and drop the places where (mostly me) added CFLAGS or other ways added -Wno-error-<thing> to avoid failure.

PR #29150, opened 2024-09-22
add a series of pkg-config tests for dev libraries

dustinkirkland:
Here's another batch of automatically generated tests.

These all test `pkg-config --modversion` on dev libraries.  This patch affects ~80 library packages, adding roughly the same number of tests.  Note that I did `make test/*` on every one of these and they do in fact pass here locally.


dustinkirkland:
Note that I can also add pkg-config --libs as well as --modversion, if that's at all helpful?

dustinkirkland:
One thing super interesting about this, is the list of lib*.yaml files that are not represented here...  My script exposes a bunch of lib*-dev packages that don't seem to ship a .pc file, and thus the pkg-config commands fail.  Perhaps we should look more into those?

dustinkirkland:
Hmm, FYI, I may be changing this over to a pipeline instead.  Please hold off on reviewing/merging.

PR #29155, opened 2024-09-23
Add full bash-completion package

jamonation:
There are lots of per-package completion subpackages, but wolfi is missing `bash-completion` itself.

Try it out - run `bash`, source `/etc/profile.d/bash-completion.sh`, then apk add `xz` as an example and run `xz -<tab>`.

#### For new package PRs only

### DIFF
--- a/bash-completion.yaml
+++ b/bash-completion.yaml
@@ -48,7 +48,7 @@ pipeline:
 
 update:
   enabled: true
-  github: 
+  github:
     identifier: scop/bash-completion
     use-tag: true
     tag-filter: v
--- a/bash-completion.yaml
+++ b/bash-completion.yaml
@@ -51,7 +51,7 @@ update:
   github:
     identifier: scop/bash-completion
     use-tag: true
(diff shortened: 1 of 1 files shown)


bash-completion.yaml
```diff
@@ -0,0 +1,61 @@
...
+  - uses: autoconf/make-install
+
+  - uses: strip
+
+update:
+  enabled: true
+  github:
+    identifier: scop/bash-completion
+    use-tag: true
+    tag-filter: v
```
kranurag7, line 54: I think `tag-filter` can be removed looking at recent tags https://github.com/scop/bash-completion/tags

jamonation: Good call, `2.` is a better filter here (without it we get some of the old junk tags like `rel20050121`).

jamonation, line 54:
```suggestion
    tag-filter: 2.
```

kranurag7 (approved):
Thanks Jamon, this is an excellent addition.

PR #29171, opened 2024-09-23
bincapz: rename to malcontent

egibs:
Related: https://github.com/chainguard-dev/malcontent/pull/464

#### For version bump PRs

### DIFF
--- a/malcontent.yaml
+++ b/malcontent.yaml
@@ -41,7 +41,7 @@ test:
     - name: Verify malcontent on itself
       runs: |
         set -o pipefail
-        output=$(mal analyze mal)
+        output=$(mal analyze /usr/bin/mal)
         test -z "$output"
     - name: Verify malcontent on yara
       runs: |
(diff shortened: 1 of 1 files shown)


egibs:
@ajayk -- thanks!

malcontent.yaml
```diff
@@ -38,11 +38,11 @@ test:
         - openssl
         - crane
   pipeline:
-    - name: Verify malcontent on itself
+    - name: Verify malcontent version
       runs: |
         set -o pipefail
-        output=$(mal analyze /usr/bin/mal)
```
egibs: This output should be empty so I'll need to take a closer look at the regression. Either way, this takes too long and we removed this as a test in the `malcontent` repository.

egibs: This output should be empty so I'll need to take a closer look at the regression. Either way, this takes too long and we removed this as a test in the `malcontent` repository.

egibs: PR incoming to fix this behavior. I'll update this PR to the new hotfix release once it merges.

ajayk (approved):

k4leung4:
should the bincapz package be withdrawn?
https://github.com/wolfi-dev/os/blob/main/WITHDRAWING_PACKAGES.md

PR #29194, opened 2024-09-23
python-3.12, python-3.13 - build tkinter with python rather than separately.

smoser:
We were building py3-tkinter separate from the python build.
The big issue with that was that it was only providing
one version of it (only 3.11, since that was being manually updated).

py3-tkinter was actually failing to build from source since tcl and tk
moved to version 9.

In order to correctly use tkinter (and pass the provided test),
we have to add explicit tcl and tk dependencies because melange does
not identify the dependency that
_tkinter.cpython-312-x86_64-linux-gnu.so (312, 313) has.

```
$ cd /usr/lib/python3.12
$ ldd ./lib-dynload/_tkinter.cpython-310-x86_64-linux-gnu.so
        linux-vdso.so.1 (0x00007ffd1fdfc000)
        libtk8.6.so => not found
        libtcl8.6.so => not found
        libc.so.6 => /lib/libc.so.6 (0x000078bac419b000)
        /lib64/ld-linux-x86-64.so.2 (0x000078bac43a3000)
```

And since it is no longer needed, delete py3-tkinter.yaml


smoser:
Ultimately this iddn't work because of recursive dependencies.
I was trying to add to python a  libx11 and tk-dev dependency.  something in libx11's stack had a dependency on python, thus a circle.

python-3.13.yaml
```diff
@@ -188,6 +195,30 @@ subpackages:
...
+          fromd="${{targets.destdir}}-base"
+          tod=${{targets.contextdir}}
+          d=usr/lib/${{vars.python}}
+
+          mkdir -p "$tod/$d/lib-dynload"
+          mv "$fromd/$d/lib-dynload"/_tkinter.*.so "$tod/$d/lib-dynload/"
+          mv "$fromd/$d/tkinter" "$tod/$d/"
+    test:
+      pipeline:
+        - runs: |
```
dannf, line 219: why not `uses: python/import` (here and above)?

smoser: good suggestion.

dannf (dismissed):
Do we also need a `py3-supported-tk`? I suppose we could make it a standalone meta package. Updating the lists of supported python versions wouldn't be any more difficult than it is for all of the other `py3-*` packages.

python-3.13.yaml
smoser: done. ecafe19fd4dd591fa1a48d80a5a1be625a0c9019

smoser:
> Do we also need a `py3-supported-tk`? I suppose we could make it a standalone meta package. Updating the lists of supported python versions wouldn't be any more difficult than it is for all of the other `py3-*` packages.

I don't think so. Its not a "real" module. things that would depend on it would actually probably just think it was part of standard library.

dannf (commented):
+1

dannf (approved):

PR #29247, opened 2024-09-24
added pkgconf test pipeline and the first two tests

dustinkirkland:
Added a new test pipeline which will be useful for testing any dev libraries with pkgconf.  I added the first two tests here, but will add a bunch more shortly...

### DIFF
--- a/pipelines/test/pkgconf.yaml
+++ b/pipelines/test/pkgconf.yaml
@@ -17,5 +17,4 @@ pipeline:
         # other packages that have build dependencies on this dev package
         pkgconf --modversion $lib_name
         pkgconf --libs $lib_name
-        #pkgconf --cflags $lib_name
       done
--- a/pipelines/test/pkgconf.yaml
+++ b/pipelines/test/pkgconf.yaml
@@ -1,0 +1,21 @@
+name: pkgconf
+
+needs:
+  packages:
+    - pkgconf
+
+pipeline:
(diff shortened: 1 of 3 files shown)


pipelines/test/pkgconf.yaml
```diff
@@ -0,0 +1,21 @@
...
+      dev_pkg=$(basename ${{targets.contextdir}})
+      # Find all pc files installed by this dev_pkg
+      for pc_file in $(apk info -L "$dev_pkg" | grep "\.pc$"); do
+        # Try to get the library name from the pc filename
+        lib_name=$(basename "$pc_file" ".pc")
+        # Run some package config tests, which will likely be used by
+        # other packages that have build dependencies on this dev package
+        pkgconf --modversion $lib_name
+        pkgconf --libs $lib_name
+        #pkgconf --cflags $lib_name
```
cpanato, line 20: Is this a leftover? can be removed?

cpanato: or uncommented

dustinkirkland: Maybe, not sure.  @xnox previously thought that --cflags should succeed everywhere but my testing is that it doesn't necessarily.  I'll remove it for cleaner code and just take a mental note to maybe readd it later.

xnox (approved):

PR #29251, opened 2024-09-24
feat(package): add dhclient package

89luca89:
Add dhclient package


xnox (changes requested):
This often does more harm than good. Is systemd-networkd, and it's built-in dhcpclient that can handle renewals and network changes, handles single and dual stack ipv4/v6 and integrates with DNS and timesync usable at all? systemd-networkd can also be configured via kernel cmdline too.

I would prefer to just use that, with wide open network file that attempts to DHCP all interfaces.

Dhcpclient needs hand holding for ipv4, ipv6, and DNS and everything. Despite being flexible, it often requires a lot of additional scaffolding around it.

xnox (approved):
Never mind we don't have systemd in most containers, doh.

imjasonh (approved):

PR #29259, opened 2024-09-24
Adds mattermost v10.0, which is the latest version

mamccorm:
This PR adds mattermost v10.0, which is the latest release which landed recently. It also removes the (now non-latest) version of mattermost v9.

### DIFF
--- a/mattermost-10.0.yaml
+++ b/mattermost-10.0.yaml
@@ -1,13 +1,15 @@
 package:
-  name: mattermost-9
-  version: 9.11.1
-  epoch: 1
+  name: mattermost-10.0
+  version: 10.0.0
+  epoch: 0
   description: "Mattermost is an open source platform for secure collaboration across the entire software development lifecycle."
   copyright:
     - license: MIT
     - license: Apache-2.0
     - license: AGPL-3.0-only
   dependencies:
+    provides:
+      - mattermost=${{package.full-version}}
     runtime:
       - bash
       - tzdata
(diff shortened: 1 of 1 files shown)


mattermost-10.0.yaml
```diff
@@ -97,7 +94,7 @@ update:
   github:
     identifier: mattermost/mattermost
     strip-prefix: v
-    tag-filter: v9.
+    tag-filter: v10.0
```
kranurag7, line 97: should we enable minor version updates here? Using `v10.` we will get update for v10.1.0 as well. Upstream is also creating some `rc` releases which can skip here.

jamonation (approved):
10.1 will be a new release based on their supported versioning scheme, so +1 here

mattermost-10.0.yaml
mamccorm: Good questions - as we don't have 'use-tag:true' set here, it defaults to the GitHub releases. We filter out non-GA releases here automatically when polling releases instead of tags.

v10.0 is the version stream: https://endoflife.date/mattermost. So when v10.0.1 and v10.0.2 come along, this will auto-increment. if v10.1.0 comes along, thats a new version stream, and isn't built by this melange. Our automation will take care of creating a new melange for that version stream

kranurag7: Ahh missed the filename :( 

this looks good. Thanks Mark.

kranurag7 (approved):

PR #29260, opened 2024-09-24
Feat/python 3.13 the things

smoser:


py3-certifi.yaml
```diff
@@ -24,16 +24,17 @@ vars:
...
+      3.10: '310'
+      3.11: '311'
+      3.12: '312'
+      # certifi uses a special versioning scheme where the version is a date like "2023.07.22". During the
+      # build process, the version is "normalized" (and importantly, changed) to a form like "2023.7.22",
+      # which creates problems with vulnerability scanners, since vulnerability data for certifi is filed
+      # using its actual version numbers (the "2023.07.22" form). This transform creates the "normalized"
+      # version, so that we can use it when needed to update the version string back to its original form in
+      # the Python package's metadata.
+      3.13: '313'
```
justinvreeland, line 36: Nit: probably want to put this comment outside the list

py3-ply.yaml
```diff
@@ -14,15 +14,16 @@ vars:
 var-transforms:
   - from: ${{package.version}}
     match: _git(.*)
-    replace: ""
+    replace: ''
```
justinvreeland, line 17: Just remove this? I guess the script caught it?

justinvreeland (commented):
Looks good two nits and it'd be nice to add the test while we're here. Annoying that the automation seems to not pick the ordering up correctly.

py3-ply.yaml
smoser: I dont' know why py-ruamel did that.

py3-certifi.yaml
smoser: ruamel di that too.

py3-ply.yaml
justinvreeland: it removed all the quoted string so I assume it just has a rule to unquote strings and then quote this as a char? make sense to me if a bit weird.

smoser:
I'm going to close this mega-change and do it in smaller hunks. 
c-i runs 'make package/name ' then 'make test/name'
and that is causing unrecoverable loops (py3-pip gets put to build before setuptools but its test ends up requiring it, so it fails with 'can't find 3.13').

smoser:
closing this. almost all are in now.

PR #29269, opened 2024-09-24
spark-3.5/3.5.3 package update patch

jamie-albert:
In the most recent release of spark-3.5 (3.5.3) there are missing imports in DelegatingCatalogExtension.java that results in a build failure. There is an [upstream fix merged](https://github.com/apache/spark/commit/ba374c673634511373af551ba55691debd35904f) that is awaiting release. I have implemented that patch below and removed an old patch that has since been merged and released.


jamie-albert:
After the build issue was fixed, the CI failed due to tests erroring out. After investigation it was found that these tests were never functional and should not have ever been merged. Only passed CI previously due to parsing bug. Adequate image level tests exist which is the only reason it is deemed plausible to comment these out at this time. The test.environment.contents.packages range.key has been fixed and this will be ready to uncomment and verify function when the following PR is released in melange: https://github.com/chainguard-dev/melange/pull/1519

jamie-albert:
CVEs will be addressed after this version bump is implemented.

imjasonh (approved):

PR #29278, opened 2024-09-25
ci-build - faster determining packages to build

smoser:
'Build package list' determines the order in which to build packages by first running 'wolfictl text' to create a list to build all packages and then getting the intersection of that list with the list of packages that had changed.

The 'all-packages' list is currently ~2570 and creating that list takes somewhere between 5 and 30 seconds.

The previous solution looped over each of the changed packages for each of the 2570 total packages.  That took 8 minutes in one pr with 145 packages changed.

On my laptop, even a simple loop like this would take over 1 minute to run:
```
   time sh -c '
      p=$(seq 1 145);
      for f in $(seq 1 2570); do
         for n in ${p}; do
            n=$(echo n);
         done;
         echo $f;
      done'
    real   0m54.332s
    user   0m31.770s
    sys    0m32.135s
```

If you're curious, the real cause of pain is the fork 'n=$(echo n)'. Dropping just the fork makes the above finish in ~ .15s real.

The big change here is to just let 'grep' handle it all, and it does that well.

```
    time bash -c '
      changed=$(seq 100 245 | sed 's,$,.yaml,');
      seq 1 2570 > all-packages;
      for f in ${changed}; do
         pkg=${f##*/};
         pkg=${pkg%.yaml};
         pkg=${pkg%.melange.yaml};
         echo $pkg ; done > changed-packages;
      grep --fixed-strings --line-regexp \
        --file=changed-packages all-packages'

    real  0m0.007s
    user  0m0.004s
    sys   0m0.004s
```

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -45,24 +45,44 @@ jobs:
       - name: Build package list
         id: package-list
         run: |
-          printf "packages=" >> $GITHUB_OUTPUT
-
+          set +f # disable pathname expansion
           wolfictl text -t name --pipeline-dir=./pipelines/ \
               -r https://packages.wolfi.dev/bootstrap/stage3 \
-              -k https://packages.wolfi.dev/bootstrap/stage3/wolfi-signing.rsa.pub > packages-list
-          while read pkg; do
-            for file in ${{ steps.changes.outputs.melange_all_changed_files }}; do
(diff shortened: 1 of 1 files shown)


xnox (approved):

xnox:
In `id:changes` we search and allow for recursive */*/melange.yaml but we don't use such files.

Can we please remove melange.yaml search in changes step too?

imjasonh (approved):

PR #29298, opened 2024-09-25
add mysql-connector-c++

Dentrax:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/mysql-connector-cpp.yaml
+++ b/mysql-connector-cpp.yaml
@@ -1,0 +1,92 @@
+package:
+  name: mysql-connector-cpp
+  version: 9.0.0
+  epoch: 0
+  description: MySQL Connector/C++ is a MySQL database connector for C++
+  copyright:
+    - license: GPL-3.0-or-later
+
+environment:
+  contents:
+    packages:
+      - abseil-cpp-dev
+      - autoconf
+      - automake
+      - boost-dev
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - cmake
+      - lz4-dev
+      - openssl-dev
+      - protobuf-dev
+      - zlib-dev
+      - zstd-dev
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/mysql/mysql-connector-cpp
+      tag: ${{package.version}}
+      expected-commit: ac218f0d2d5b7957a8a2efee148bc8d7a1c623a5
+
+  - uses: cmake/configure
+    with:
+      opts: |
+        -DWITH_BOOST=system \
+        -DWITH_LZ4=system \
+        -DWITH_SSL=system \
+        -DWITH_ZLIB=system \
+        -DWITH_ZSTD=system
+
+  - uses: cmake/build
+
+  - uses: cmake/install
+
+  - working-directory: /home/build/output
+    runs: |
+      mkdir -p ${{targets.destdir}}/usr/lib
+      mv *.so* ${{targets.destdir}}/usr/lib/
+
+  - uses: strip
+
+subpackages:
+  - name: ${{package.name}}-dev
+    pipeline:
+      - uses: split/dev
+    dependencies:
+      replaces:
+        - ${{package.name}}
+    description: ${{package.name}} dev
+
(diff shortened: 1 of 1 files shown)


mysql-connector-cpp.yaml
```diff
@@ -0,0 +1,82 @@
...
+  contents:
+    packages:
+      - autoconf
+      - automake
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - cmake
+      - openssl-dev
+      - zlib-dev
```
xnox, line 19: Please use as many system packages as possible. I.e. protobuf zlib lz4 zstd. We do not want vendored copies of those; if at all possible to use system ones.

mysql-connector-cpp.yaml
```diff
@@ -0,0 +1,82 @@
...
+        - ${{package.name}}
+    description: ${{package.name}} dev
+
+  - name: libmysqlclient
+    description: MySQL Connector/C++ shared libraries
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/usr/lib
+          mv ${{targets.destdir}}/usr/lib/*.so* ${{targets.contextdir}}/usr/lib/
+          chmod 755 ${{targets.subpkgdir}}/usr/lib/*.so*
```
xnox, line 56: After these splits. The main package contains incorrectly installed cmake files and INFO_BIN and INFO_SRC files. Imho the main package should just have the shared library.

The cmake files should be instructed to be correctly installed into -dev package. Ditto INFO_BIN and INFO_SRC files, not sure how they are used even, if at all.

And imho libmysqlclient as a sub-pckage should not exist at all.

xnox (changes requested):

xnox:
> 2024/10/01 18:44:31 INFO Package: mysql-connector-cpp: [no-repeated-deps]: package zlib-dev is duplicated in environment (ERROR)

mysql-connector-cpp.yaml
Dentrax: Thanks done. Also passed the following flags:
```
-DWITH_BOOST=system \
-DWITH_LZ4=system \
-DWITH_SSL=system \
-DWITH_ZLIB=system \
-DWITH_ZSTD=system
```

But passing `-DWITH_PROTOBUF=system` fails with syntax error due to probuf version, i guess.

```
/home/build/cdk/protocol/mysqlx/protocol.cc:131:17: error: 'log_handler' was not declared in this scope
/home/build/cdk/protocol/mysqlx/protocol.cc:131:3: error: 'SetLogHandler' was not declared in this scope
```

mysql-connector-cpp.yaml
Dentrax: How should we proceed here? I was following the official build doc: https://dev.mysql.com/doc/connector-cpp/9.0/en/connector-cpp-installation-source-cpp.html#connector-cpp-installation-source-build

Do you think `libmysqlclient` package would redundant?

mysql-connector-cpp.yaml
xnox: sadness, likely everything needs to be regenerated or it is incompatible. Leave it as it is, we have tried =)

mysql-connector-cpp.yaml
xnox: the guide only lists libraries, it doesn't tell you the name of the subpackage to use, and how one splits them or not.

So yes, I do believe the libmysqlclient subpackage should be re-dundant, and the default files in destdir should be cleaned up. (correct cmake files locations, and likely purge the weird INFO files if they are unreferenced otherwiese).

Dentrax: Thanks @xnox, I have removed the sub-package. But I'm not sure if I understand the "INFO infos", could you please take a look when possible?

Dentrax: Kindly ping here. 🙃 @xnox

EyeCantCU (dismissed):

xnox (approved):

PR #29299, opened 2024-09-25
python 3.13 for flit-core, install, pip, wheel.

smoser:


py3-wheel.yaml
```diff
@@ -17,6 +17,7 @@ data:
       3.10: "310"
       3.11: "311"
       3.12: "312"
+      3.13: "313"
```
xnox, line 20: this priority will make `apk add py3-wheel` install the 3.13 version, is that what is desired here?

smoser: thanks. i literally had just hit 'convert to draft' to stop that.

smoser: we are broken in that regard right now with setuptools :-(

xnox (approved):

PR #29331, opened 2024-09-25
add py3-robotframework package

kranurag7:
required deps for [hadoop package](https://github.com/wolfi-dev/os/pull/29037). 

ref: https://github.com/apache/hadoop/blob/5bb6f450340fa3deb931571d08d42a3f358af8bf/Dockerfile#L33

### DIFF
--- a/py3-robotframework.yaml
+++ b/py3-robotframework.yaml
@@ -1,0 +1,43 @@
+# Generated from https://pypi.org/project/robotframework/
+package:
+  name: py3-robotframework
+  version: "7.1"
+  epoch: 0
+  description: Generic automation framework for acceptance testing and robotic process automation (RPA)
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - python-3
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - wolfi-base
+
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: 8ce9a227bf31bf0d723e540b31e1d4ef9128d1b8
+      repository: https://github.com/robotframework/robotframework
+      tag: v${{package.version}}
+
+  - name: Python Build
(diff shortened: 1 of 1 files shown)


py3-robotframework.yaml
```diff
@@ -0,0 +1,76 @@
...
+package:
+  name: py3-robotframework
+  version: "7.1"
+  epoch: 0
+  description: Generic automation framework for acceptance testing and robotic process automation (RPA)
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - python-3
```
EyeCantCU, line 11: Remove this and set provider priority to 0

py3-robotframework.yaml
```diff
@@ -0,0 +1,76 @@
...
+    test:
+      pipeline:
+        - uses: python/import
+          with:
+            python: python${{range.key}}
+            import: robot
+
+update:
+  enabled: true
+  manual: false
```
EyeCantCU, line 61: We don't need to explicitly set this to false

py3-robotframework.yaml
```diff
@@ -0,0 +1,76 @@
...
+      3.12: "312"
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - py3-supported-pip
+      - wolfi-base
```
EyeCantCU, line 30: We shouldn't need to include this

py3-robotframework.yaml
```diff
@@ -0,0 +1,76 @@
...
+  - name: py-versions
+    items:
+      3.10: "310"
+      3.11: "311"
+      3.12: "312"
+
+environment:
+  contents:
+    packages:
+      - build-base
```
EyeCantCU, line 26: Don't believe we need this either, worth a check if these were copied

EyeCantCU (changes requested):

EyeCantCU (approved):

PR #29341, opened 2024-09-26
kafka-3.8: fix bitnami compat

srishtih:
- FQIW the server.properties file states it is to be used in ZK mode and the server.properties in /opt/bitnami/kafka/config/kraft states it must be used with kraft mode.
- Its mostly a permission related issue and this change works (as confirmed by the package level tests, image-level tests(locally) and docker-compose test). When the image is ran as kafka, a new server.properties is created with proper ownership and permissions after the entrypoint scripts runs.

### DIFF
--- a/kafka-3.8.yaml
+++ b/kafka-3.8.yaml
@@ -2,7 +2,7 @@ package:
   name: kafka-3.8
   # When bumping check to see if the CVE mitigation can be removed.
   version: 3.8.0
-  epoch: 3
+  epoch: 4
   description: Apache Kafka is a distributed event streaming platform
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


xnox (approved):
Looks good. But still in draft, so not merging it.

kranurag7 (approved):
The scans are failing because there was a recent CVE in one of the dependencies (protobuf-java) that is used by kafka. ref: https://github.com/protocolbuffers/protobuf/security/advisories/GHSA-735f-pc8j-v9w8

I tried to patch the [dependencies.gradle](https://github.com/apache/kafka/blob/3.8.0/gradle/dependencies.gradle) by pinning the protobuf-java version to 3.25.5 but it's a shaded dependency (ref: https://github.com/apache/kafka/blob/3.8.0/build.gradle#L1575-L1585)
I think for now we'll have to proceed with filing an advisory for this.

PR #29345, opened 2024-09-26
Jaeger Component Build

utieyin:
#### For new package PRs only


jaeger-all-in-one.yaml
```diff
@@ -1,59 +0,0 @@
-package:
```
Dentrax, line 1: Should we add this to `withdrawn-packages.txt`?

jaeger.yaml
```diff
@@ -0,0 +1,71 @@
...
+            find cmd/query/app/ui/actual -type f | grep -v .gitignore | xargs gzip --no-name
+            touch -t $(date -r jaeger-ui/packages/jaeger-ui/build/index.html '+%Y%m%d%H%M.%S') cmd/query/app/ui/actual/index.html.gz
+          fi
+      - uses: go/build
+        with:
+          modroot: cmd/${{range.key}}
+          packages: .
+          ldflags: -X github.com/jaegertracing/jaeger/pkg/version.latestVersion=v${{package.version}}
+          output: ${{range.key}}
+
```
Dentrax, line 66: Can we add simple tests? i.e. `--help` or `version`.

Preferably `test/daemon-check-output`.

utieyin: Have added tests that encompass both version and help for those packages that have them.

Dentrax: Nice, thanks! You can create the tests under subpackages-level so you don't need to write every subpackage under the test block. See this, for example: https://github.com/wolfi-dev/os/blob/4bdb602db133a030a33f39f946d893c686530d52/nvidia-container-toolkit.yaml#L57-L61

In this case, you can move the `version` and `help` checks to subpackage block:

```
test:
      pipeline:
        - runs: |
            ${{range.key}} version 2>&1 | grep ${{package.version}}
            ${{range.key}} --help
```

utieyin: The version test is not available for all the components. Moving this into the subpackage block will break the tests as not all the components have the version command. 

A conditional statement on the version test will be unwieldly

jaeger.yaml
```diff
@@ -0,0 +1,133 @@
...
+      tracegen: "."
+
+subpackages:
+  - range: jaeger-components
+    name: "${{package.name}}-${{range.key}}"
+    pipeline:
+      - runs: |
+          ## Only run this step if building for all-in-one package
+          ## This builds the ui component of jaeger with yarn
+          if [[ "${{range.key}}" = "all-in-one" ]]; then
```
EyeCantCU, line 45: We can also add the condition to the pipeline itself. I.E.:

```
if: "${{range.key}}" == "all-in-one"
runs: blah
```

jaeger.yaml
EyeCantCU: This looks fine to me

EyeCantCU (approved):

Dentrax (approved):

PR #29349, opened 2024-09-26
add byteman package

kranurag7:
required for hadoop package. 

#### For new package PRs only

### DIFF
--- a/byteman.yaml
+++ b/byteman.yaml
@@ -15,10 +15,9 @@ environment:
       - zip
       - unzip
       - maven
-      - openjdk-9
-      - openjdk-9-default-jvm
+      - openjdk-11
   environment:
-    JAVA_HOME: /usr/lib/jvm/java-1.9-openjdk
+    JAVA_HOME: /usr/lib/jvm/java-11-openjdk
     LANG: "en_US.UTF-8"
 
 pipeline:
@@ -55,8 +54,8 @@ test:
   environment:
     contents:
       packages:
-        - openjdk-9
-        - openjdk-9-default-jvm
+        - openjdk-11
+        - openjdk-11-default-jvm
   pipeline:
     - name: jar listing
       runs: |
--- a/byteman.yaml
+++ b/byteman.yaml
@@ -58,6 +58,6 @@ test:
         - openjdk-9
         - openjdk-9-default-jvm
   pipeline:
-    - name: Version Check
+    - name: jar listing
       runs: |
         jar -tf /usr/share/java/byteman/byteman.jar
--- a/byteman.yaml
+++ b/byteman.yaml
@@ -4,8 +4,7 @@ package:
   epoch: 0
   description: Simplify Java tracing, monitoring and testing with Byteman
   copyright:
-    - license: GPL-2.0-or-later
-      license-path: copyright.txt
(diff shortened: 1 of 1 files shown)


byteman.yaml
```diff
@@ -0,0 +1,63 @@
...
+# NOTE: JRE runtime dependency required but only jar is needed at this point for using this package as part of hadoop.
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - zip
+      - unzip
+      - maven
+      - openjdk-9
```
EyeCantCU, line 18: Is `openjdk-9` a typo?

kranurag7: RJ, from upstream reference https://github.com/bytemanproject/byteman/blob/4e8bd4985ed3110ff0a2e226e154a72c0d7b8579/bin/bminstall.sh#L40-L52 I found out that only two versions are supported for building it. I first tried it with openjdk-8 and that fails. Later I switched to using openjdk-9 here. 

I also checked the docs but was not able to find supported openjdk version. (ref: https://downloads.jboss.org/byteman/4.0.23/byteman-programmers-guide.html#building-byteman-from-the-sources)

byteman.yaml
```diff
@@ -0,0 +1,63 @@
...
+test:
+  environment:
+    contents:
+      packages:
+        - openjdk-9
+        - openjdk-9-default-jvm
+  pipeline:
+    - name: jar listing
+      runs: |
+        jar -tf /usr/share/java/byteman/byteman.jar
```
jamonation, line 63: Is there a way to run the main class in the jar, or import it in a test and run a program that relies on it?

kranurag7: Thanks Jamon, I've addressed this via https://github.com/wolfi-dev/os/pull/29349/commits/0bb32bc9479af2774d5aa9f552b3629dba62c5f1 (it runs a simple hello world app using byteman agent)

jamonation: Terrific, that's exactly the kind of test I was hoping for. Now we know it works as intended right at package build time.

jamonation (dismissed):

byteman.yaml
EyeCantCU: JDK 9 is EOL, let's use 11 as it supports 9+

byteman.yaml
```diff
@@ -0,0 +1,74 @@
...
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - zip
+      - unzip
+      - maven
+      - openjdk-9
+      - openjdk-9-default-jvm
```
EyeCantCU, line 19: We can remove this

byteman.yaml
```diff
@@ -0,0 +1,74 @@
...
+  github:
+    identifier: bytemanproject/byteman
+    use-tag: true
+
+test:
+  environment:
+    contents:
+      packages:
+        - openjdk-9
+        - openjdk-9-default-jvm
```
EyeCantCU, line 59: And this

EyeCantCU (changes requested):

byteman.yaml
EyeCantCU: Confirmed the JVM is actually used in tests

EyeCantCU (approved):

PR #29382, opened 2024-09-26
fix(gitlab-toolbox-scripts): Correct directory for scripts

EyeCantCU:
The scripts for GitLab toolbox are nested in a bin directory. Move these toplevel to /scripts


jamonation (approved):
Looks good, thanks for tracking down where those scripts come from

PR #29398, opened 2024-09-26
py3-h5py: multiverisionize

justinvreeland:


py3-h5py.yaml
```diff
@@ -3,25 +3,40 @@ package:
...
+      - py3-supported-cython
+      - py3-supported-numpy
+      - py3-supported-pip
+      - py3-supported-pkgconfig
+      - py3-supported-python-dev
+      - py3-supported-tomli
       - wolfi-baselayout
+  environment:
+    H5PY_SYSTEM_LZF: 1
+    H5PY_SETUP_REQUIRES:
```
smoser, line 39: setting this explicitly to '0' seems more obvious than haivng an empty string .
if you wanted/needed to set it to an empty string could you _set_ it to an empty string? or maybe I just don't understand what this does here.

smoser (changes requested):
minor nit/question

smoser (approved):

PR #29400, opened 2024-09-26
Add datawire-envoy

wlynch:
This is a fork of envoy with patches applied by Ambassador for their emissary project.
See https://github.com/emissary-ingress/emissary/blob/ac2dc64c6621cd8ec5617f3328544364bdd3fb01/DevDocumentation/ARCHITECTURE.md#envoy

The only modification here from base envoy is the source location. These should try and stay in sync, but I don't know if there's a better way to represent this.

#### For new package PRs only

#### For new version streams

I believe this supports the same upstream version streams as envoy.


datawire-envoy-1.31.yaml
```diff
@@ -0,0 +1,108 @@
...
+  - uses: strip
+
+subpackages:
+  - name: ${{package.name}}-oci-entrypoint
+    description: Entrypoint for using Envoy in OCI containers
+    dependencies:
+      runtime:
+        - busybox
+      provides:
+        - envoy-oci-entrypoint=${{package.full-version}}
```
ajayk, line 79: This can't provide envoy-oci-entrypoints needs changes

datawire-envoy-1.31.yaml
```diff
@@ -0,0 +1,108 @@
...
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/var/lib/envoy/init
+          cp envoy/ci/docker-entrypoint.sh ${{targets.subpkgdir}}/var/lib/envoy/init/envoy-entrypoint.sh
+          chmod +x ${{targets.subpkgdir}}/var/lib/envoy/init/envoy-entrypoint.sh
+
+  - name: ${{package.name}}-config
+    description: Default Envoy configuration
+    dependencies:
+      provides:
+        - envoy-config=${{package.full-version}}
```
ajayk, line 90: This can't provide envoy-config needs changes

datawire-envoy-1.31.yaml
wlynch: I think we only need the actual envoy binary and can probably reuse the existing envoy subpackages, so removed.

datawire-envoy-1.31.yaml
wlynch: Removed.

ajayk (approved):

PR #29410, opened 2024-09-27
jitsucom-jitsu/GHSA-gp8f-8m3g-qvj9 remediation

jamie-albert:
Added the next npm package and the fix version to the overrides section of the patch file. Only patch version changes from next @ 14.2.5 > 14.2.10. Remaining CVE in subpackage jitsucom-jitsu-console-2.8.2-r0 (html-minifier 4.0.0 (npm) CVE-2022-37620) existed before this dependency version update and has no fix, will require advisory to be filed.

### DIFF
--- a/jitsucom-jitsu.yaml
+++ b/jitsucom-jitsu.yaml
@@ -1,7 +1,7 @@
 package:
   name: jitsucom-jitsu
   version: 2.8.2
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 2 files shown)


philroche (changes requested):
@jamie-albert we need an epoch bump too don't we?

jamie-albert:
aye good catch thank you!

philroche (approved):

PR #29167, opened 2024-09-23
add apache tez package

kranurag7:
Fixes:

Related:

#### For new package PRs only

- [ ] This PR is marked as fixing a pre-existing package request bug
  - [ ] Alternatively, the PR is marked as related to a pre-existing package request bug, such as a dependency
- [x] REQUIRED - The package is available under an OSI-approved or FSF-approved license
- [x] REQUIRED - The version of the package is still receiving security updates
- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`)

PR #29412, opened 2024-09-27
feat(hive): add apache hive package

maxgio92:
This PR introduces the package for [Apache Hive](https://hive.apache.org/).

Related: #29167

#### For new package PRs only

- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`):
  There's no support policy available.


hive/hadoop-3.4.0-compat.patch
```diff
@@ -0,0 +1,32 @@
+diff --git a/common/src/java/org/apache/hadoop/hive/common/JvmMetrics.java b/common/src/java/org/apache/hadoop/hive/common/JvmMetrics.java
```
maxgio92, line 1: Becasuse the `EventCounter` class has been removed since Hadoop v3.4.0 because was using `log4j` 1.x, the following code for JVM metrics no longer compiles.

The following patch removes the event count metric (error, warning, and info logs) from the JVM metrics, that is no longer supported in Hadoop 3.4.0.

The change has been tracked in Hadoop in the issue [HADOOP-17524](https://issues.apache.org/jira/browse/HADOOP-17524), of which the PR [here](https://github.com/apache/hadoop/pull/2909).

Moreover, a related change in Hadoop is tracked [here](https://issues.apache.org/jira/browse/HADOOP-17642).

maxgio92:
Build and tests pass with the patch. But since the upstream work is in progress to support Hadoop 3.4.0 in Hive [here](https://issues.apache.org/jira/browse/HIVE-28191), we cannot ensure fully compatibility.

hive.yaml
```diff
@@ -0,0 +1,58 @@
...
+      expected-commit: 183f8cb41d3dbed961ffd27999876468ff06690c # rel/release-4.0.0
+
+  - uses: patch
+    with:
+      # This patch covers license check for the goal org.apache.rat:apache-rat-plugin:0.13:check
+      # because the .gitconfig file for safe directories at build time should be ignored.
+      patches: rat.patch
+
+  - runs: |
+      mvn package -Dhadoop.version=3.3.6 -DskipTests -Dmaven.javadoc.skip=true -Pdist
```
maxgio92, line 38: Hadoop 3.4.0 is not yet supported. Work to support it is in progress [here](https://github.com/apache/hive/pull/5187).

hive.yaml
```diff
@@ -0,0 +1,102 @@
...
+        echo "JAVA_HOME=/usr/lib/jvm/java-1.8-openjdk" >> hadoop-${HADOOP_VERSION}/etc/hadoop-env.sh
+        echo "127.0.0.1 $(hostname)" >> /etc/hosts
+    # This test suite starts Beeline and HiveServer2 in the same process.
+    # Ref: https://cwiki.apache.org/confluence/display/Hive/HiveServer2+Overview
+    - name: Create a table in Hive
+      runs: |
+        export HADOOP_HOME=$(pwd)/hadoop-${HADOOP_VERSION}
+
+        # Prepare HDFS directories
+        echo "build:x:$(id -u):$(id -g):build:/home/build:/bin/sh:" >> /etc/passwd
```
maxgio92, line 81: This tweak is there because Hadoop uses passwd DB to get the user name during authentication and access control checks.

hive.yaml
```diff
@@ -0,0 +1,101 @@
...
+    - name: Prepare Hadoop standalone
+      runs: |
+        echo "JAVA_HOME=/usr/lib/jvm/java-1.8-openjdk" >> hadoop-${HADOOP_VERSION}/etc/hadoop-env.sh
+        echo "127.0.0.1 $(hostname)" >> /etc/hosts
+    # This test suite starts Beeline and HiveServer2 in the same process.
+    # Ref: https://cwiki.apache.org/confluence/display/Hive/HiveServer2+Overview
+    - name: Create a table in Hive
+      runs: |
+        # Prepare HDFS directories
+        echo "test:x:$(id -u):$(id -g):test user:/:/bin/sh:" >> /etc/passwd
```
maxgio92, line 80: This tweak is there because Hadoop uses passwd DB to get the user name during authentication and access control checks for setting up the HDFS.

hive.yaml
```diff
@@ -0,0 +1,105 @@
...
+        rm *.tar.gz*
+
+    - name: Prepare Hadoop standalone
+      runs: |
+        echo "JAVA_HOME=/usr/lib/jvm/java-1.8-openjdk" >> hadoop-${HADOOP_VERSION}/etc/hadoop-env.sh
+        echo "127.0.0.1 $(hostname)" >> /etc/hosts
+
+    - name: Prepare HDFS directories
+      runs: |
+        echo "test:x:$(id -u):$(id -g):test user:/:/bin/sh:" >> /etc/passwd
```
maxgio92, line 80: This tweak is there because Hive uses the user name from the passwd DB to set up the HDFS ownership, during the authentication.

hive.yaml
```diff
@@ -0,0 +1,106 @@
...
+    packages:
+      - bash
+      - busybox
+      - maven
+      - openjdk-8
+  environment:
+    LANG: en_US.UTF-8
+    JAVA_HOME: /usr/lib/jvm/java-1.8-openjdk
+    HIVE_VERSION: 4.0.0
+    HADOOP_VERSION: 3.3.6
```
maxgio92, line 24: Hadoop 3.4.0 is not yet supported. Work to support it is in progress (https://github.com/apache/hive/pull/5187).

hive.yaml
```diff
@@ -0,0 +1,100 @@
...
```
EyeCantCU, line 24: Since we need to explicitly set the version here, when we package 3.3 could we make sure we're pinned to this version so that there isn't drift?

I.E.,

```
dependencies:
  runtime:
    - hadoop=3.3.6
```

Same for Tez

maxgio92: Totally makes sense, thanks @EyeCantCU.

smoser, line 24: you can replace this environment variable `HADOOP_VERSION` with a melange variable.

```yaml
vars:
   hadoop-version: 3.3.6
```

then you use `${{vars.hadoop-version}}` and you don't have to repeat yourself in the runtime environment below (if they are intended to be the same thing).

hive.yaml
```diff
@@ -0,0 +1,100 @@
...
+        - curl
+    environment:
+      JAVA_HOME: /usr/lib/jvm/java-1.8-openjdk
+      HADOOP_VERSION: 3.3.6
+      HADOOP_HOME: /home/build/hadoop
+      HIVE_HOME: /usr/share/java/hive
+  pipeline:
+    - name: Download Hadoop
+      runs: |
+        gpg --keyserver hkps://keyserver.ubuntu.com --recv-key CD32D773FF41C3F9E74BDB7FB362E1C021854B9D
```
smoser, line 65: I think that using `fetch` pipeline would be preferable here.

it will need updating any time HADOOP_VERSION needs updating (versus the gpg path that would accept it if it was signed), but fetch is used in more places, so we can improve that pipeline for any reason and get benefit everywhere.

then you can drop the gpg, gpg-agent, gnupg-dirmngr and curl from above.
```yaml
test:
  pipeline:
    - uses: fetch
      with:
        uri: https://dlcdn.apache.org/hadoop/common/hadoop-${{vars.hadoop-version}}/hadoop-${{vars.hadoop-version}}.tar.gz
        expected-sha256: f5195059c0d4102adaa7fff17f7b2a85df906bcb6e19948716319f9978641a04

```

smoser (changes requested):
some minor things.

hive.yaml
maxgio92: Love it! Thank you @smoser.

PR #29415, opened 2024-09-27
datadog-jmxfetch: fix update config block

rawlingsj:
Presubmit check should have caught this, that's now been fixed so future checks will fail..

### DIFF
--- a/datadog-jmxfetch.yaml
+++ b/datadog-jmxfetch.yaml
@@ -34,7 +34,7 @@ pipeline:
 update:
   enabled: true
   github:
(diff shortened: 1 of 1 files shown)


hectorj2f (approved):

datadog-jmxfetch.yaml
```diff
@@ -34,7 +34,7 @@ pipeline:
 update:
   enabled: true
   github:
-    identifier: https://github.com/DataDog/jmxfetch
+    identifier: DataDog/jmxfetch
```
hectorj2f, line 37: Do we need to bump the epoch?

PR #29572, opened 2024-09-27
datawire-envoy: add setcap subpackage

wlynch:
This adds a subpackage that runs setcap to enable certain capabilities on the binary. This is a requirement for downstream Ambassador dependencies (see
https://github.com/emissary-ingress/emissary/blob/ac2dc64c6621cd8ec5617f3328544364bdd3fb01/build-aux/Dockerfile#L129-L132).

### DIFF
--- a/datawire-envoy-1.31.yaml
+++ b/datawire-envoy-1.31.yaml
@@ -1,7 +1,7 @@
 package:
   name: datawire-envoy-1.31
   version: 1.31.2
-  epoch: 0
+  epoch: 1
   description: Ambassador fork of Envoy Proxy.
   copyright:
     - license: Apache-2.0
@@ -25,6 +25,7 @@ environment:
       - cmake
       - coreutils
       - git
+      - libcap-utils
       - libtool
       - llvm-libcxx-15
       - llvm-libcxx-15-dev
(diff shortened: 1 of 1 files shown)


datawire-envoy-1.31.yaml
```diff
@@ -69,6 +69,15 @@ pipeline:
...
   - uses: strip
 
+subpackages:
+  - name: ${{package.name}}-setcap-ei
+    description: Envoy binary with CAP_NET_BIND_SERVICE capabilities enabled, required for downstream Ambassador usage.
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/usr/bin
+          cp -p "${{targets.destdir}}"/usr/bin/envoy "${{targets.subpkgdir}}"/usr/bin/envoy
+          setcap cap_net_bind_service=ei /usr/local/bin/envoy
```
jamonation, line 79: Should both of these be /usr/bin/envoy or /usr/local/bin/envoy?

wlynch: Good catch. The upstream dockerfile uses /usr/local/bin, but IIUC we prefer /usr/bin because we consider these distribution provided?

Changed to /usr/bin.

jamonation (dismissed):

datawire-envoy-1.31.yaml
```diff
@@ -69,6 +70,16 @@ pipeline:
 
   - uses: strip
 
+subpackages:
+  - name: ${{package.name}}-setcap-ei
```
EyeCantCU, line 74: Sorry to nitpick, but can we call this `-privileged` instead?

wlynch: Done.

EyeCantCU (approved):

PR #29584, opened 2024-09-27
updated pkgconf test pipeline to be more portable

dustinkirkland:
This pkgconf test pipeline does a better job of checking what should be working (and skipping things that should not).

There are two sample tests added... One is to openssl.yaml which succeeds.  While the other is to libarchive.yaml fails.  However, the libarchive failure actually looks like a real failure that needs to be addressed.

See:
```

make test/openssl
make test/libarchive
```


dustinkirkland:
Note that in a followup commit, I actually added the missing run time dependency in libarchive.yaml.  @xnox says he has some melange magic that might be able to fix / solve these more automatically, too.

xnox (approved):

PR #29585, opened 2024-09-27
Add emissary

wlynch:
https://github.com/emissary-ingress/emissary

Dependent on: https://github.com/wolfi-dev/os/pull/29572

#### For new package PRs only

### DIFF
--- a/emissary.yaml
+++ b/emissary.yaml
@@ -108,7 +108,7 @@ subpackages:
           #
           # We can't use --ignore-installed for the list above, because it won't remove the old versions of the packages and the scanners will
           # complain about the old versions being present.
-          # We also need to run rm -rf existing setuptools first because of this, because of the weird "pip only installs 1 version" behavior described above.
+          # Running rm -rf existing setuptools first is a workaround for this.
           rm -rf ${{targets.contextdir}}/usr/lib/python${{range.key}}/site-packages/setuptools*
           python${{range.key}} -m pip install --no-cache-dir --prefix=/usr --root=${{targets.contextdir}} --ignore-installed --upgrade setuptools==70.0.0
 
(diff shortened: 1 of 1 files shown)


emissary.yaml
```diff
@@ -0,0 +1,100 @@
...
+
+test:
+  environment:
+    contents:
+      packages:
+        - curl
+  pipeline:
+    - uses: test/kwok/cluster
+    - runs: |
+        exit 1
```
EyeCantCU, line 94: Let's remove that :laughing:

emissary.yaml
```diff
@@ -0,0 +1,100 @@
...
+  dependencies:
+    runtime:
+      - datawire-envoy-privileged
+      - git
+      - python3
+
+environment:
+  contents:
+    packages:
+      - go
```
EyeCantCU, line 17: Can be removed. Fetched via `go/build`

emissary.yaml
```diff
@@ -0,0 +1,100 @@
...
+    with:
+      packages: ./cmd/kubestatus
+      output: kubestatus
+
+  - name: Python Build
+    runs: |
+      make python/ambassador.version
+
+      # Install Python dependencies
+      python3 -m venv venv
```
EyeCantCU, line 61: We should install deps without a venv now

EyeCantCU (changes requested):

emissary.yaml
wlynch: We should update https://eng.inky.wtf/docs/images/package-and-image-guides/building-package-walkthrough-python/ - that's what made me think we needed venv.

What's the recommended way to do this?

emissary.yaml
wlynch: Removed.

emissary.yaml
wlynch: d'oh. removed.

emissary.yaml
EyeCantCU: Let's definitely update the docs. It's the same process. You just wouldn't create the virtual - environment; you'd use the hosts pip directly

wlynch: Got things working. PTAL!

emissary.yaml
```diff
@@ -0,0 +1,115 @@
...
+        imports: from ambassador import Version
+    - uses: test/kwok/cluster
+    - runs: |
+        /usr/bin/entrypoint.sh &
+
+        # Wait for server to come up
+        sleep 10
+
+        curl -i http://127.0.0.1:8004/ambassador/v0/check_alive
+        curl -i http://127.0.0.1:8004/ambassador/v0/check_ready
```
Dentrax, line 115: Let's use `test/daemon-check-output` here:

```suggestion
    - name: "Test entrypoint"
      uses: test/daemon-check-output
      with:
        start: /usr/bin/entrypoint.sh
        timeout: 30
        expected_output: |
          expected_log1
          expected_log2
          expected_log3
        post: |
          curl -i http://127.0.0.1:8004/ambassador/v0/check_alive
        curl -i http://127.0.0.1:8004/ambassador/v0/check_ready
```

wlynch: Done, though had to modify the error strings a bit since there are some "error" logs being output due to the binary not running in a pod.

srishtih (dismissed):

srishtih (approved):

wlynch:
@EyeCantCU The PR is not letting me submit, even though there isn't any outstanding comments. I think it's carrying over your last "Needs Changes" status, so you may need to review this in order for this to submit.

EyeCantCU (approved):
Looks great!

PR #29644, opened 2024-09-28
bootstrap dart

kranurag7:
bootstrap dart 

#### For new package PRs only

### TODO
- [x] add update block (need some var-transforms) 
- [x] fix arm build

### DIFF
--- a/dart.yaml
+++ b/dart.yaml
@@ -81,8 +81,8 @@ update:
     - "merge*"
   git: {}
   schedule:
-    period: monthly
-    reason: upstream releases stable version every 3 months.
+    period: weekly
+    reason: upstream releases stable version every 3 months but polling can be done weekly so that we don't miss things.
 
 test:
   environment:
--- a/dart.yaml
+++ b/dart.yaml
@@ -64,6 +64,12 @@ pipeline:
       cp -R /usr/share/ca-certificates /runtime/usr/share
       ln -sf /usr/lib/dart/bin/dart "${{targets.destdir}}"/usr/bin/dart
 
+  - uses: git-checkout
+    with:
+      repository: https://github.com/dart-lang/sdk
+      tag: ${{package.version}}
+      expected-commit: 179da3ba67ead156a3ba718bd798cb7a1728e741
+
 update:
   enabled: true
   ignore-regex-patterns:
@@ -73,9 +79,10 @@ update:
     - "analyzer-*"
     - "meta-*"
     - "merge*"
-  github:
-    identifier: dart-lang/sdk
-    use-tag: true
+  git: {}
+  schedule:
+    period: monthly
+    reason: upstream releases stable version every 3 months.
 
 test:
   environment:
--- a/dart.yaml
+++ b/dart.yaml
@@ -42,26 +42,27 @@ pipeline:
     runs: |
       ./tools/build.py --mode release --arch x64 create_sdk
       mv out/ReleaseX64/dart-sdk ${{targets.destdir}}/usr/lib/dart
-      cp /lib64/ld-linux-x86-64.so.2 ${{targets.destdir}}/runtime/
+      ln -sf /lib64/ld-linux-x86-64.so.2 ${{targets.destdir}}/runtime/ld-linux-x86-64.so.2
 
   - if: ${{build.arch}} == "aarch64"
     working-directory: /home/dart-sdk/sdk
     runs: |
       ./tools/build.py --mode release --arch arm64 create_sdk
       mv out/ReleaseARM64/dart-sdk ${{targets.destdir}}/usr/lib/dart
-      cp /lib/ld-linux-aarch64.so.1 ${{targets.destdir}}/runtime/
+      ln -sf /lib/ld-linux-aarch64.so.1 ${{targets.destdir}}/runtime/ld-linux-aarch64.so.1
 
-  - name: "symlink dart binary and copy required libraries to runtime dir"
+  - name: "symlink dart binary and required libraries to runtime dir"
     runs: |
       mkdir -p ${{targets.destdir}}/usr/bin
+      mkdir -p ${{targets.destdir}}/runtime/etc/ssl
+      mkdir -p ${{targets.destdir}}/runtime/usr/share
+      mkdir -p ${{targets.destdir}}/runtime/lib
+      ln -sf /etc/ssl/certs ${{targets.destdir}}/runtime/etc/ssl
+      ln -sf /usr/share/ca-certificates ${{targets.destdir}}/runtime/usr/share
       FILES="/etc/nsswitch.conf /lib/libc.so.6 /lib/libdl.so.2 /lib/libm.so.6 /lib/libnss_dns.so.2 /lib/libpthread.so.0 /lib/libresolv.so.2 /lib/librt.so.1"
       for file in $FILES; do
-        cp -L $file ${{targets.destdir}}/runtime/
+        ln -sf $file ${{targets.destdir}}/runtime$file
       done
-      mkdir -p /runtime/etc/ssl
-      mkdir -p /runtime/usr/share
-      cp -R /etc/ssl/certs /runtime/etc/ssl
-      cp -R /usr/share/ca-certificates /runtime/usr/share
       ln -sf /usr/lib/dart/bin/dart "${{targets.destdir}}"/usr/bin/dart
 
   - uses: git-checkout
@@ -89,7 +90,6 @@ test:
     contents:
       packages:
         - ca-certificates-bundle
-        - git
   pipeline:
     - name: "dart version"
       runs: /usr/bin/dart --version
@@ -126,12 +126,9 @@ test:
         webdev build
     - name: unix and windows compile check
       runs: |
-        git clone https://github.com/dart-lang/samples --depth=1
-        cd samples/native_app
-        dart compile exe bin/main.dart -o hello_world
-        ./hello_world
-        dart compile exe bin/main.dart -o hello_world.exe
-    - name: dart testing
-      runs: |
-        cd samples/server/simple
+        dart create -t console demo-cli
+        cd demo-cli
         dart test
+        dart compile exe bin/demo_cli.dart -o demo_cli
+        ./demo_cli
+        dart compile exe bin/demo_cli.dart -o demo_cli.exe
--- a/dart.yaml
+++ b/dart.yaml
@@ -67,9 +67,9 @@ pipeline:
 update:
   enabled: true
   ignore-regex-patterns:
-    - "*.dev"
-    - "*.alpha"
-    - "*.beta"
+    - '.*\.dev$'
+    - '.*\.alpha$'
+    - '.*\.beta$'
     - "analyzer-*"
     - "meta-*"
     - "merge*"
--- a/dart.yaml
+++ b/dart.yaml
@@ -57,11 +57,11 @@ pipeline:
       mkdir -p ${{targets.destdir}}/runtime/etc/ssl
       mkdir -p ${{targets.destdir}}/runtime/usr/share
       mkdir -p ${{targets.destdir}}/runtime/lib
-      ln -sf /etc/ssl/certs ${{targets.destdir}}/runtime/etc/ssl
-      ln -sf /usr/share/ca-certificates ${{targets.destdir}}/runtime/usr/share
+      cp -R /etc/ssl/certs ${{targets.destdir}}/runtime/etc/ssl
+      cp -R /usr/share/ca-certificates ${{targets.destdir}}/runtime/usr/share
       FILES="/etc/nsswitch.conf /lib/libc.so.6 /lib/libdl.so.2 /lib/libm.so.6 /lib/libnss_dns.so.2 /lib/libpthread.so.0 /lib/libresolv.so.2 /lib/librt.so.1"
       for file in $FILES; do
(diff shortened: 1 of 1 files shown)


dart.yaml
```diff
@@ -0,0 +1,81 @@
...
+      - git
+      - python3
+      - xz
+
+pipeline:
+  - name: "clone the source code"
+    runs: |
+      mkdir -p ${{targets.destdir}}/usr/lib/
+      mkdir -p ${{targets.destdir}}/runtime
+      git clone --depth=1 https://chromium.googlesource.com/chromium/tools/depot_tools.git /home/depot_tools
```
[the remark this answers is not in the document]
kranurag7: I think upstream docs mentions to use gclient specifically and during sync. I think it does something extra apart from `git-checkout` 

ref: https://github.com/dart-lang/sdk/blob/main/docs/Building.md#getting-the-source 

After fetching the code, if I don't do `gclient sync` then I run into errors.

dart.yaml
```diff
@@ -0,0 +1,137 @@
...
+      mkdir -p /runtime/usr/share
+      cp -R /etc/ssl/certs /runtime/etc/ssl
+      cp -R /usr/share/ca-certificates /runtime/usr/share
+      ln -sf /usr/lib/dart/bin/dart "${{targets.destdir}}"/usr/bin/dart
+
+  - uses: git-checkout
+    with:
+      repository: https://github.com/dart-lang/sdk
+      tag: ${{package.version}}
+      expected-commit: 179da3ba67ead156a3ba718bd798cb7a1728e741
```
kranurag7, line 71: this is a hack to get update on dart. with git backend, we require `git-checkout` pipeline and this is just for the purpose of update. 

ref: https://dart.googlesource.com/sdk/+refs

dart.yaml
```diff
@@ -73,9 +79,10 @@ update:
     - "analyzer-*"
     - "meta-*"
     - "merge*"
-  github:
-    identifier: dart-lang/sdk
-    use-tag: true
+  git: {}
+  schedule:
+    period: monthly
```
wlynch, line 84: We may want to make this weekly? I worry that a monthly cadence would make us very slow to pick up new versions, since you could have up to a 30 day lag time before a re-poll is done.

dart.yaml
```diff
@@ -0,0 +1,137 @@
...
+      git fetch --tags --all
+      git checkout tags/${{package.version}} -b ${{package.version}}
+      gclient sync
+
+  - if: ${{build.arch}} == "x86_64"
+    working-directory: /home/dart-sdk/sdk
+    runs: |
+      ./tools/build.py --mode release --arch x64 create_sdk
+      mv out/ReleaseX64/dart-sdk ${{targets.destdir}}/usr/lib/dart
+      cp /lib64/ld-linux-x86-64.so.2 ${{targets.destdir}}/runtime/
```
xnox, line 45: why is this not a symlink too?

dart.yaml
```diff
@@ -0,0 +1,137 @@
...
+
+  - if: ${{build.arch}} == "aarch64"
+    working-directory: /home/dart-sdk/sdk
+    runs: |
+      ./tools/build.py --mode release --arch arm64 create_sdk
+      mv out/ReleaseARM64/dart-sdk ${{targets.destdir}}/usr/lib/dart
+      cp /lib/ld-linux-aarch64.so.1 ${{targets.destdir}}/runtime/
```
xnox, line 52: why is this not a symlink too?

dart.yaml
```diff
@@ -0,0 +1,137 @@
...
+  schedule:
+    period: weekly
+    reason: upstream releases stable version every 3 months but polling can be done weekly so that we don't miss things.
+
+test:
+  environment:
+    contents:
+      packages:
+        - ca-certificates-bundle
+        - git
```
xnox, line 92: I don't see direct calls to these in the test cases. it means dart itself calls it. Please add them as runtime dependencies of dart; otherwise one will have to repeat this again in the image building; any custom image that needs dart; and any other higher deployment that tries to depend on and use dart.

dart.yaml
kranurag7: > I'd rather have our ca-certificate package ship symlink (not copy)

following our conversation, I am now using symlinks everywhere.

dart.yaml
kranurag7: Thanks, I was using `git` to clone samples repo from dart and run some tests but now I've converted those to use dart itself. `git` is not longer required. 
`ca-certificates-bundle` is still required because by default test-environment is still coming with `r0` as of now. I think we need digests update here after https://github.com/wolfi-dev/os/pull/29948 was merged to get `r1` propogated in test environment.

dart.yaml
```diff
@@ -0,0 +1,134 @@
+package:
+  name: dart
+  version: 3.5.3
+  epoch: 0
+  description: The Dart SDK, including the VM, JS and Wasm compilers, analysis, core libraries, and more.
+  copyright:
+    - license: BSD-3-Clause
+  dependencies:
+    runtime:
+      - ca-certificates-bundle
```
kranurag7, line 10: _NOTE: info not a blocker for this package._ 

In dart ecosystem, it's very [common](https://github.com/search?q=%22COPY+--from%3Dbuild+%2Fruntime%2F+%2F%22+AND+%22dart%22&type=code) to use the following. Many template generators generate this including the official one. 

following is the second stage of the build pipeline:
```docker
FROM scratch AS server
COPY --from=build /runtime/ /
```

Now, as long as `scratch` is used, it's fine here. If any image is used that already ships with existing ca-certificates like wolfi, alpine etc. then the following error will occur.

```bash
 => ERROR [stage-1 2/3] COPY --from=build /runtime/ /                                                                                                                                    0.0s
------
 > [stage-1 2/3] COPY --from=build /runtime/ /:
------
Dockerfile:16
--------------------
  14 |     # and the pre-built AOT-runtime in the `/runtime/` directory of the base image.
  15 |     FROM alpine
  16 | >>> COPY --from=build /runtime/ /
  17 |     COPY --from=build /app/bin/server /app/bin/
  18 |
--------------------
ERROR: failed to solve: cannot replace to directory /var/lib/docker/fuse-overlayfs/yy0qk73q5lwnk1o8hmm1856pc/merged/etc/ssl/certs with file
```

I don't think, as per my understanding currently, there's way we can avoid this, ca-certificates-bundle is a runtime dependency for dart.
so far the image is working good and can used directly as long as `scratch` is used in the second stage.

dart.yaml
xnox: @kranurag7 don't care what upstream do. In the end we ship an image. Doing a copy like this, destroys provenance & sbom.

Do they do chroot syscall into `/runtime` ? are they distributing `/runtime` as a container? well, for us container start at `/`.

dart.yaml
xnox: ouch.

I almost want to make "/runtime" a symlink to "/" or "/usr"

Can you point at sample, demo apps that do that?

Cause if we need to revendor runtime like this to be self-contained, we would also need re-vendor relevant json sbom. And apk  database will look odd.

xnox: hang on, in a distroless world, our dart image would be minimal. It can contain /runtime symlink to either / or /usr, would that be enough?

Cause I want to avoid building nested containers.

kranurag7: I agree and like the idea, I'm trying to think and cover the case where people are already using a multistage pipeline and given upstream vendors libraries under `/runtime` and users are copying that runtime inside scratch image using `COPY --from=build /runtime/ /` 

ref: https://github.com/dart-lang/dart-docker/blob/a8ada07907fbfa139635fe90bfb31a236b7c2f34/Dockerfile-debian.template#L36-L51

If we don't ship the libraries and just copy the binary then existing Dockerfile won't work. From distroless world, if we use any of `scratch` or static images from cgr or gcr it won't work because underlying libraries are missing in these images. 

the binary is dynamically linking to some system libraries.
A simple binary compiled using binary links to the following:
```bash
 $ dart compile exe bin/server.dart  -o bin/server
Generated: /home/user/examples/webserver/bin/server
$ ldd bin/server
        linux-vdso.so.1 (0x00007ffe7a520000)
        libdl.so.2 => /lib/libdl.so.2 (0x00007cbcfa090000)
        libpthread.so.0 => /lib/libpthread.so.0 (0x00007cbcfa08b000)
        libm.so.6 => /lib/libm.so.6 (0x00007cbcf9fa9000)
        libc.so.6 => /lib/libc.so.6 (0x00007cbcf980b000)
        /lib64/ld-linux-x86-64.so.2 (0x00007cbcfa098000)
```

Usage on GitHub shows that this image is used in multistage pipeline so we'll either need a `-compat` vendoring libraries (hack) or a direct copy of that at runtime compatibility with existing distroless usage. 

I've reverted symlinking for now locally but open to more discussion here on the best way to proceed here. We can ship with our best practices and write a migration note if required.

dart.yaml
kranurag7: It's not chroot but a `cp` https://github.com/dart-lang/dart-docker/blob/a8ada07907fbfa139635fe90bfb31a236b7c2f34/Dockerfile-debian.template#L36-L51 (and this copying is what is being used at runtime inside scratch/static images via multistage builds) 

> are they distributing /runtime as a container?

Not as a separate container but they have docs for using image on dockerhub which does this and I think people are using that as a best practice.
https://hub.docker.com/_/dart

but to the original review here, I am thinking whether we should revert to copy again for these shared libraries (`/lib/ld-linux-aarch64.so.1` and `/lib64/ld-linux-x86-64.so.2`) because these two are also missing from scratch and static image.
_(mostly thinking about cases where the final executables are getting copied along with shared libraries including the above two in static/scratch image)_

xnox: @kranurag7 

What is the difference between using `FROM cgr.dev/chainguard/dart` rather than from static and copy /runtime?

Note that our images are so minimal, that most of the time, when we build a runtime image, it really is "static + runtime" and nothing else.

In this current packaging do you have anything out-side of runtime? Cause potentially, I want to build two images here - "dart" that can be used for development and "dart-runtime" which one can use for deployment like our "jre" images and like a replacement for "FROM static + cp /runtime". Because then this dart-runtime image can have correct sbom; not vendor in shared library dependencies; and be rebuilt/upgraded automatically as glibc is upgraded.

Can you please create subpackage `name: ${{package.name}}-runtime` ship the dart files that are runtime only inside there, so split /runtime which contains dart original files only, without any `cp`. For any required things add symlinks + runtime dependencies. Ensure that main dart package runtime depends on dart-runtime. When building images, we will need to build two images dart and dart-runtime. In the dockerfile example our distroless way of building things then would be use `dart` image as build, and then copy to `dart-runtime`.

Your dart-runtime image, in sbom and installed database must have libc listed.

Also looking at the structure of runtime package, it looks more like regular package split with just a compat symlink from "/runtime -> /".

Does any of the above make sense at all? Or should I attempt to show you a prototype as a PR to your fork?

dart.yaml
xnox: I think i replied in the wrong thread https://github.com/wolfi-dev/os/pull/29644#discussion_r1787707357

xnox:
@kranurag7 just to be clear runtime package is really just a meta-package (empty / dependencies only)? There is nothing darty inside of it? as in there is no common runtime binaries or libraries? 

I am not familiar with dart, is end user programs generate and run correctly in such environment? (as in they are like kind of like go binaries more or less, built for linux, with cgo, and only often really need glibc at runtime)

dart.yaml
```diff
@@ -71,6 +58,17 @@ pipeline:
...
       expected-commit: 179da3ba67ead156a3ba718bd798cb7a1728e741
 
+subpackages:
+  - name: ${{package.name}}-runtime
+    dependencies:
+      runtime:
+        - ld-linux
+        - glibc
+        - wolfi-baselayout
+    pipeline:
```
xnox, line 68: no ca-certificates-bundle? no cryptography / openssl?

i wonder if packaging one of the tests apps you mentioned as a subpackage here would be useful, to have a test pipeline here in runtime to run one of them.

kranurag7: missed ca-certificates-bundle, will look if openssl is required here in the morning and will add a test too.

kranurag7: I think above dependencies including above should suffice for x64 and aarch64. 

There's an issue about including libatomic as well (ref: https://github.com/dart-lang/dart-docker/issues/141) but it's only required for risc-v arch. 

from upstream reference I find following info about dart-runtime: 
- https://dart.dev/tools/dart-compile#exe (self contained executables that depend on dart-runtime)
- https://dart.dev/overview#runtime (provides gc, type checks at runtime) _[I was not able to fully understood this part please let me know if looking at the description it requires more shared libraries at runtime, I think above should suffice for gc and type-checks. [Isolates](https://dart.dev/language/isolates) looks more like goroutine from a quick skim so I think nothing extra required there.]_ 

> no cryptography / openssl?

I think it's similar to Go. 
ref: https://github.com/dart-lang/crypto (A set of cryptographic functions implemented in pure Dart.)

xnox: If needed, can be added later. note our static image (no deps of anything) still ships ca-certificates-bundle as it's difficult to add them if one copies in a static go binary, precisely to support static apps that will arrive later.

xnox (approved):
This can suppport lifecycle updates and CVE triange. Thus this is good to go.

PR #29645, opened 2024-09-28
Backfill some more packages

mattmoor:
Downloaded from https://cloudlogging.app.goo.gl/74RW47vS7PJXh44B8 and processed with:

```
cat foo.csv | cut -d',' -f 6 | gsed -E 's#^https://apk.cgr.dev/[0-9a-f]+/(aarch64|x86_64)/##g' | gsed -E 's|\?mount=.+$||g' | sort -u | grep -E '.apk$'
```

Confirmed these are all from Wolfi with:
```
for x in $(cat missing.txt); do echo === $x ===; curl --fail --head https://packages.wolfi.dev/os/x86_64/$x; done | less
```


mattmoor:
Need to run [this](https://github.com/wolfi-dev/os/actions/workflows/backfill.yaml) once this lands

k4leung4 (approved):

PR #28327, opened 2024-09-13
grafana-11.2: make compat version specific

ajayk:

PR #29677, opened 2024-09-29
Add a missing `provides` for `grafana-oci-compat`

mattmoor:
Our public images builds weren't updated to reflect this change, so they were still picking up a stale version of things.

ref: https://github.com/wolfi-dev/os/pull/28327

### DIFF
--- a/grafana-11.2.yaml
+++ b/grafana-11.2.yaml
@@ -1,7 +1,7 @@
 package:
   name: grafana-11.2
   version: 11.2.1
-  epoch: 0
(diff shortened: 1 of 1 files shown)


mattmoor:
I suspect that we need to update https://github.com/chainguard-images/images-private/blob/b4b078225a4b93fa79a13e673dcf8c7d54a753b4/images/grafana/main.tf#L15 due to the original change as well @ajayk

ajayk (approved):

PR #29705, opened 2024-09-30
python 3.13 - batch 1 packages.

smoser:


py3-distro.yaml
```diff
@@ -57,6 +62,34 @@ subpackages:
...
+      runtime:
+        - py${{range.key}}-${{vars.pypi-package}}
+      provides:
+        - py3-${{vars.pypi-package}}-bin
+        - py3-${{vars.pypi-package}}
+      provider-priority: ${{range.value}}
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/usr/
+          mv ./cleanup/${{range.key}}/bin ${{targets.contextdir}}/usr/
```
roaksoax, line 78: instead of doing the `.cleanup` thing here and in the section above for this package, could you have done this here instead in the subpkg

```
          mkdir -p "${{targets.subpkgdir}}"/usr/
          mv "${{targets.destdir}}"/usr/bin "${{targets.subpkgdir}}"/usr/
```

roaksoax (approved):
lgtm!

py3-distro.yaml
smoser: @justinvreeland did this first, and I remember wondering the same thing.
Its not super clear, but I don't think there is a better way right now.

The problem is that you want to get the files here in this package (py3.11-distro-bin) , and they were in py3.11-distro.
But we don't actually have any supported mechanism from melange to reference py3.11-distro.  So we move it to a "global" space.

Ie, we actually _could_ do this: 

    mv ${{targets.contextdir}}/../py${{range.key}}-${{vars.pypi-package}/usr/bin ${{targets.contextdir}}/usr/

Where, rendered that would look more like this:

    mv ${{targets.contextdir}}/../py3.11-distro/usr/bin ${{targets.contextdir}}/usr/
    
But that is making assumptions on the relationship of one "contextdir" to another.  We know that py3.11-distro is a sibling directory to py3.11-distro-bin, but nothing actually promises us that.

PR #29723, opened 2024-09-30
python 3.13 - batch 3 packages.

smoser:
This just adds python 3.13 to builds of a set of multi-versioned py3-* packages.

### DIFF
--- a/py3-exceptiongroup.yaml
+++ b/py3-exceptiongroup.yaml
@@ -2,7 +2,7 @@
 package:
   name: py3-exceptiongroup
   version: 1.2.2
-  epoch: 1
+  epoch: 2
   description: Backport of PEP 654 (exception groups)
   copyright:
     - license: MIT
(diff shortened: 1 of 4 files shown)


py3-exceptiongroup.yaml
```diff
@@ -25,9 +25,10 @@ vars:
 data:
   - name: py-versions
     items:
-      3.10: "310"
-      3.11: "311"
-      3.12: "312"
+      3.10: '310'
+      3.11: '311'
+      3.12: '312'
+      3.13: '300'
```
justinvreeland, line 31: why not 313?  IDK why I didn't notice that before.

smoser: @xnox saved us there.

because if it is 313 then 'apk add py3-exceptiongroup' will get you python 3.13, and we don't want that now.  The first reason we don't want it is that 3.13 isn't released, so we don't want it as "default".

Then we have other "does'nt build yet" reasons.

justinvreeland: Ahh that is a good check.  I'd rather we just not produce the unversioned packages and make people specify the version tbh.

justinvreeland (approved):

PR #29737, opened 2024-09-30
py3-rich: multiversionize

justinvreeland:


py3-rich.yaml
```diff
@@ -1,25 +1,30 @@
 package:
   name: py3-rich
   version: 13.8.1
-  epoch: 0
+  epoch: 2
```
smoser, line 4: double bump?

py3-rich.yaml
```diff
@@ -44,19 +78,7 @@ update:
...
-        python=$(set +x; m=/usr/bin/python3.*; set -- $m
-            [ $# -eq 1 -a -x "$1" ] && echo "$1" && exit 0
-            echo "found $# matches to $m"; exit 1;)
-        $python -c "from rich import print; print('${RICH_TEST_STRING}')"
-
-        # Test Rich Formatting Capabilities
-        $python -c "from rich.console import Console; console = Console(); console.print('${RICH_TEST_STRING}', style='bold red')"
-
-        # Test Rich JSON Rendering
-        $python -c "from rich import print; print({'json': ['Renders', 'well', 'in', 'Rich']})"
```
smoser, line 62: Did you mean to drop this? it does probably provide some benifit more than just 'import'.
 you could move it to the subpackage.

smoser (changes requested):

py3-rich.yaml
justinvreeland: Just being overly cautious since I normally forget I suppose.

py3-rich.yaml
justinvreeland: Seems to fall more on the side of unit testing than verifying the package is installed which was my understanding of the kind of testing we wanted.

smoser: I guess i wuoldn't even qualify it as unit testing because it doesn't really even check proper behavior.

There are absolutely cases where something like this would fail but simply 'import rich' would fail.  For some python packages, 'import <thing>' does almost literally nothing.

I guess at this poitn I'm just not quite to "throw away existing tests".
If they are flakey or wrong, then I understand.

justinvreeland: They're not even good test. Maybe wrong is a strong word but I don't find these tests much better than import.  As far as I can tell they're not testing a regression we've hit before and we don't compare the output for accuracy just success.  Not much that we package seems to do more than import and there doesn't seem to be a reason for this project to get special treatment.  I'd rather have consistent tests in our python packages than partial test suites we're not interested in developing and maintaining and which don't reflect the priorities of the upstream community we're pulling it from.  I'll add them back but from my perspective tests like these are just waiting to become bitrot or give us false reassurance that we're "tested" it.  If we want to test the project we should run the test suite if we want to make sure it's installed and available import seems sufficient even for the case where import does ostensibly nothing because it shows the files are in the right place and available which is what a package does. Though it would be nice if import could verify all the runtime deps were present and correct as well.

smoser (dismissed):
can you add 3.13 for me? I _think_ it should just work.

justinvreeland:
No prob!

py3-rich.yaml
```diff
@@ -29,13 +35,57 @@ pipeline:
...
+            import: ${{vars.pypi-package}}
 
-  - uses: strip
+  - name: py3-supported-${{vars.pypi-package}}
+    description: meta package providing ${{vars.pypi-package}} for supported python versions.
+    dependencies:
+      runtime:
+        - py3.10-${{vars.pypi-package}}
+        - py3.11-${{vars.pypi-package}}
+        - py3.12-${{vars.pypi-package}}
```
smoser, line 82: need py3.13 here.

smoser (approved):

Issue #26401, opened 2024-08-15
FTBFS 2024-08-15


xnox:
Generated with:

```
./cg build log --build-id=6a9cd064-7533-4046-ad92-44caa14b07b0 --search 'ERROR: failed to build package' --attrs pkg,arch | sort | cut -d\  -f2,14- | sed 's|^|- [ ] |'
```

```[tasklist]
### Tasks
- [x] 🔗 pkg=apache-arrow https://github.com/wolfi-dev/os/pull/26492
- [x] 🔗 pkg=cargo-auditable https://github.com/wolfi-dev/os/pull/26412
- [x] 🔗 pkg=clang-17
- [x] 🔗 pkg=clang-18
- [x] 🔗 pkg=collectd https://github.com/wolfi-dev/os/pull/26484
- [x] 🔗 pkg=composer
- [x] 🔗 pkg=delta
- [x] 🔗 pkg=freerdp
- [x] 🔗 pkg=gnupg https://github.com/wolfi-dev/os/pull/26406
- [x] 🔗 pkg=gtk-2.0
- [x] 🔗 pkg=helix
- [x] 🔗 pkg=hickory-dns
- [x] 🔗 pkg=ingress-nginx-controller-1.11
- [x] 🔗 pkg=ivykis
- [x] 🔗 pkg=jitsucom-bulker
- [x] 🔗 pkg=kube-fluentd-operator
- [x] 🔗 pkg=libcmis https://github.com/wolfi-dev/os/pull/30226
- [x] 🔗 pkg=libevdev https://github.com/wolfi-dev/os/pull/26423
- [x] 🔗 pkg=libffi https://github.com/wolfi-dev/os/pull/26422
- [x] 🔗 pkg=llvm-lld-18
- [x] 🔗 pkg=meilisearch
- [x] 🔗 pkg=neuvector-sigstore-interface
- [x] 🔗 pkg=nfs-utils
- [x] 🔗 pkg=njs
- [x] 🔗 pkg=nodejs-16
- [ ] 🔗 pkg=nodejs-19
- [x] 🔗 pkg=openjdk-10
- [x] 🔗 pkg=openjdk-12
- [x] 🔗 pkg=openjdk-7 https://github.com/wolfi-dev/os/pull/30228
- [x] 🔗 pkg=openjdk-8
- [x] 🔗 pkg=percona-server-8.3
- [x] 🔗 pkg=php-8.1
- [x] 🔗 pkg=pulseaudio
[...]

PR #29744, opened 2024-09-30
suricata-update: Fix build

justinvreeland:
Related: #26401


imjasonh:
You'll need to update the `epoch` or else this won't rebuild the package

imjasonh (approved):

PR #29753, opened 2024-09-30
py3-pandas: multiversionize

justinvreeland:


smoser (approved):

smoser (changes requested):
I think you need changes in 'dependencies' (top level... line 8).
1. add provider-priority to the base package
2. drop the runtime dependencies from that thing.

smoser (approved):

PR #29756, opened 2024-09-30
feat: Add OpenJDK with CRaC patches

EyeCantCU:
Adds OpenJDK 17 and 21 with CRaC patches

### DIFF
--- a/openjdk-crac-17.yaml
+++ b/openjdk-crac-17.yaml
@@ -1,0 +1,301 @@
+package:
+  name: openjdk-crac-17
+  version: 17.0.25
+  epoch: 0
+  description:
+  copyright:
+    - license: GPL-2.0-only
+  dependencies:
+    runtime:
+      - openjdk-crac-17-jre
+
+var-transforms:
+  - from: ${{package.version}}
+    match: (\d+)\.\d+\.(\d+)
+    replace: $1
+    to: mangled-package-version
+
+environment:
+  contents:
+    packages:
+      - alsa-lib-dev
+      - autoconf
+      - automake
+      - bash
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - cups-dev
+      - file
+      - fontconfig-dev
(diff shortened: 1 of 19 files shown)


xnox (dismissed):
ship it

openjdk-crac-17.yaml
```diff
@@ -0,0 +1,295 @@
...
+        # Test the jar
+        java -jar app.jar
+
+        # Test jlink
+        jlink --verbose --module-path "app.jar:$JAVA_HOME/jmods" \
+          --add-modules test.modules \
+          --output test-project-jre
+
+        # Test custom JRE
+        test-project-jre/bin/java -jar app.jar
```
xnox, line 295: Are there any symlistic java crac-criu processes that we can checkpoint restart?

Something like "sleep 10; print hello world", start it, checkpoint it, kill it, restart it, observe that it comes back and prints hello world? As that's the point of this build that one can do this.

openjdk-crac-17.yaml
```diff
@@ -0,0 +1,295 @@
...
+      - libxext-dev
+      - libxi-dev
+      - libxrandr-dev
+      - libxrender-dev
+      - libxt-dev
+      - libxtst-dev
+      - openjdk-16
+      - openjdk-16-default-jvm
+      - zip
+      - crac-criu
```
xnox, line 47: needs wolfictl lint

openjdk-crac-17.yaml
EyeCantCU: Wrote a small Java program that does this, but I can't get a package test working here as invoking CRIU for checkpoint/restore requires it to be running in a container with elevated privileges (or the NET_ADMIN cap set). We'll need to cover this in our image tests. Testing with a privileged wolfi-base container with OpenJDK CRaC and CRaC CRIU layered currently

xnox: fair enough!

PR #29757, opened 2024-09-30
php-8.1: Fix build

justinvreeland:
update deps add compiler option

Related: #26401

### DIFF
--- a/php-8.1.yaml
+++ b/php-8.1.yaml
@@ -1,7 +1,7 @@
 package:
   name: php-8.1
   version: 8.1.29
-  epoch: 4
+  epoch: 5
   description: "the PHP programming language"
   copyright:
     - license: PHP-3.01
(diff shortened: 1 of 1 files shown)


php-8.1.yaml
```diff
@@ -49,6 +52,8 @@ environment:
       - sqlite-dev
       - unixodbc-dev
       - wget
+  environment:
+    CFLAGS: -Wno-incompatible-pointer-types
```
xnox, line 56: This implies serious bugs though =/ is there no cherry-pick from upstream to resolve this? is it just a few things or many?

shyim: We should open maybe an upstream issue ticket, I tried to cherry pick some commits, but could not solve it still

xnox (changes requested):
Needs epoch bump, but otherwise good to go.

xnox (approved):

PR #29760, opened 2024-10-01
python 3.13 - batch 6 packages.

smoser:
This just adds python 3.13 to builds of a set of multi-versioned py3-* packages.

### DIFF
--- a/py3-platformdirs.yaml
+++ b/py3-platformdirs.yaml
@@ -2,7 +2,7 @@
 package:
   name: py3-platformdirs
   version: 4.3.6
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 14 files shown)


py3-platformdirs.yaml
```diff
@@ -10,14 +10,15 @@ package:
     provider-priority: 0
 
 vars:
-  pypi-packagename: platformdirs
```
imjasonh, line 13: Need to bump the epoch on this package

smoser: :heavy_check_mark:

xnox (approved):

PR #29800, opened 2024-10-01
python-3.13: move to rc3

ajayk:


python-3.13.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: python-3.13
-  version: 3.13.0_rc2
-  epoch: 1
+  version: 3.13.0_rc3
+  epoch: 2
```
smoser, line 4: You do not need the epoch bump here, unless you  know something that I do not.

@xnox told me about apk's knowledge of 'rc' and such. https://github.com/alpinelinux/apk-tools/blob/74de0e9bd73d1af8720df40aa68d472943909804/src/version.c#L39

It does appear to do the right thing.


Also neat, you can ask apk what it thinks like this:
```
# apk version -t 3.13.0_rc2-r1 3.13.0_rc3-r0
<
```

smoser (dismissed):

xnox (approved):
Yeap, it is a new version; with epoch reset to zero =)

Issue #27342, opened 2024-08-30
crac support in wolfi java sdk/jre


tuananh:
https://openjdk.org/projects/crac/ 

> The CRaC (Coordinated Restore at Checkpoint) Project researches coordination of Java programs with mechanisms to checkpoint (make an image of, snapshot) a Java instance while it is executing. Restoring from the image could be a solution to some of the problems with the start-up and warm-up times. The primary aim of the [Project](https://openjdk.org/projects/index.html) is to develop a new standard mechanism-agnostic API to notify Java programs about the checkpoint and restore events. Other research activities will include, but will not be limited to, integration with existing checkpoint/restore mechanisms and development of new ones, changes to JVM and JDK to make images smaller and ensure they are correct.

not sure what it takes to get this in wolfi java sdk/jre. looks like these are what we need to do according to this https://github.com/CRaC/openjdk-builds

- build jdk as usual
- build a modified criu https://github.com/CRaC/openjdk-builds
- copy it over to jdk/lib

some other information:

- azul jdk supports it https://www.azul.com/products/components/crac/ . They added support very recently https://github.com/zulu-openjdk/zulu-openjdk/pull/280/files
- this feature will make startup quite a lot faster, especially in serverless workload
- spring already support crac https://docs.spring.io/spring-framework/reference/integration/checkpoint-restore.html

tuananh:
@rawlingsj what do you think? is this something Wolfi team interested it?

xnox:
Unfortunately the mentioned repositories do not provided tagged release matching upstream releases. An alternative source of this feature got packaged however and delivered as an image.

If interested, please see https://images.chainguard.dev/directory/image/jdk-crac/versions and https://images.chainguard.dev/directory/image/jre-crac/versions .

PR #29829, opened 2024-10-01
chore(crac-criu): Use Canonical's fork, remove Python bindings

EyeCantCU:
This updates CRaC CRIU to track Canonical's fork instead of upstream as it is more actively maintained and LTS IoenJDK CRaC releases will track Canonical's OoenJDK CRaC forks, ensuring CRaC CRIU is compatible with any changes Canonical may make

Additionally disable everything that isn't directly needed for OpenJDK CRaC, including Python bindings, as this isn't useful elsewhere. Upstream CRIU should be packaged and used if needed elsewhere. CRaC CRIU is only needed for CRaC OpenJDK so there is no reason to provide support for anything that isn't directly used by it

### DIFF
--- a/crac-criu.yaml
+++ b/crac-criu.yaml
@@ -30,6 +30,7 @@ environment:
       - nftables
       - nftables-dev
       - nftables-static
+      - perl
       - pkgconf
       - protobuf-c-dev
       - protobuf-dev
--- a/crac-criu.yaml
+++ b/crac-criu.yaml
@@ -1,6 +1,6 @@
 package:
   name: crac-criu
-  version: 1.4
+  version: 1.4.3
   epoch: 0
   description: A project to implement checkpoint/restore functionality for Linux
   copyright:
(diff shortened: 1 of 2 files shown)


tuananh:
related: https://github.com/wolfi-dev/os/issues/27342

@EyeCantCU may i ask how do we build a jdk image with crac support?

EyeCantCU:
@tuananh the build process for OpenJDK with CRaC support is inline with OpenJDK. The upstream repository can be found here: https://github.com/openjdk/crac

After that has been built, CRaC patched CRIU can be dropped into the `/path/to/jvm/installation/lib` folder

philroche (approved):
This seems like a big change. Can you link to a bug or provide context on why this change is being made?

philroche (changes requested):
This seems like a big change. Can you link to a bug or provide context on why this change is being made?

tuananh:
> @tuananh the build process for OpenJDK with CRaC support is inline with OpenJDK. The upstream repository can be found here: https://github.com/openjdk/crac
> 
> After that has been built, CRaC patched CRIU can be dropped into the `/path/to/jvm/installation/lib` folder

do we have any plan to update openjdk here to add support for that?

EyeCantCU:
@philroche there is no issue tracking this, I've updated my PR description

@tuananh as upstream doesn't cut tags for the latest version of OpenJDK, this isn't something we plan to offer in Wolfi

philroche (approved):

PR #29846, opened 2024-10-01
py3-pandas: restore missing runtime dep

justinvreeland:


xnox:
that's odd that we even have py3-tzdata packaged; as normally one only has regular tzdata installed. py3-tzdata is a shim for when linux/container doesn't have normal tzdata present.

Also, wouldn't it make more sense to add such runtime dep on py3-tz itself, rather than panda? Does panda need tzdata, or does py3-tz need py3-tzdata?

In other distributions py3-tz depends on tzdata (NB! not py3-tzdata)

py3-pandas.yaml
```diff
@@ -54,6 +54,7 @@ subpackages:
         - py${{range.key}}-numpy
         - py${{range.key}}-pytz
         - py${{range.key}}-dateutil
+        - py${{range.key}}-tzdata
     pipeline:
```
xnox, line 58: Should it be here, or runtime dep on py3-pytz packages?

justinvreeland: It should be here. It's specifying the deps for the correct python version in each package runtime deps. The top level package `py3-pytz` is really just an empty package and I wish we didn't have to generate it all the magic is done by the subpackages provides.

justinvreeland:
> Also, wouldn't it make more sense to add such runtime dep on py3-tz itself, rather than panda? Does panda need tzdata, or does py3-tz need py3-tzdata?

According to upstream it needs the python modules tzdata which is py3-tzdata for us as far as I'm aware.

It seemed like a straightforward fix when I saw the failure and noticed I missed this package when I converted it.  I saw that upstream specifies a dep on tzdata in their pyproject.yaml as well so I didn't really look into it much.

py3-pandas.yaml
xnox: indeed they do conditional direct import at https://github.com/pandas-dev/pandas/blob/f598670353311a6fff4e6e1e96074ccf0737e6b7/pandas/_libs/tslibs/timezones.pyx#L64

xnox (approved):

PR #29926, opened 2024-10-02
python 3.13 - batch 13 packages.

smoser:
This just adds python 3.13 to builds of a set of
multi-versioned py3-* packages.

The change here for cassandra-driver was to correctly name the pypi package 'cassandra-driver' , which causes the 'import' change needed.

### DIFF
--- a/py3-cassandra-driver.yaml
+++ b/py3-cassandra-driver.yaml
@@ -2,7 +2,7 @@
 package:
   name: py3-cassandra-driver
   version: 3.29.2
-  epoch: 1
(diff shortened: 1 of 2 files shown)


smoser:
this is going to fail until the build of https://github.com/wolfi-dev/os/pull/29925 publishes (https://github.com/wolfi-dev/os/runs/30984440732).

But after that, i think hooray done.

roaksoax (approved):
lgtm!

imjasonh (approved):

PR #29933, opened 2024-10-02
Do not leave tmp.XXXXX directories around on each 'make' invocation.

smoser:
Every time `make` was invoked we would create:
 * a temp dir for local-wolfi repos
 * a temp dir for dev-container-wolfi repos
 * a temp dir for dev-container-out

Now, instead of always creating them, create them only if you're in the
given target.  The use of `--tmpdir` allows us to provide a template
(`$@.XXXXXX`) while respecting TMPDIR.  The template means that we get tmpdirs
named `local-wolfi.5Sr9nC` rather than just `tmp.5Sr9nC`.

I don't love the complexity in the setting of OUT_DIR, but it appears
that the ?= operator in make will still evaluate the right side of that
conditional assignment even if its set.
    
    FOO ?= $(shell echo bar; echo this-gets-called-even-if-FOO-is-set 1>&2)

So, without the shell magic, `make OUT_DIR=./my.out-dir` would still create the
temp dir while respecting and keeping OUT_DIR.

We *do* leak OUT_DIR tmp directories. I can fix that if someone wants
it fixed. Not bothering as the intent of OUT_DIR is to have data from
inside the container persisted after the container exist.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,23 +144,23 @@ dev-container:
 	    ghcr.io/wolfi-dev/sdk:latest@sha256:6c001d29fe246547dd24b52682770e4d7770ced4d1f25e534f31237d7308a1ce
 
 PACKAGES_CONTAINER_FOLDER ?= /work/packages
-TMP_REPOSITORIES_DIR := $(shell mktemp -d)
-TMP_REPOSITORIES_FILE := $(TMP_REPOSITORIES_DIR)/repositories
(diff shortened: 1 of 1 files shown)


justinvreeland:
🙇🏻‍♂️

justinvreeland (dismissed):

xnox (approved):

PR #29948, opened 2024-10-02
ca-certificates-bundle: add compat with Fedora-like expectations

xnox:
Currently Chainguard Images universally set SSL_CERT_FILE pointing at
a Debian-like path of the bundle.

A cert directory with by-c_rehash symlinks of certificates is not
shipped by default, as it is incompatible with apko image building
which prohibits executing any package scriplets. Also by-c_rehash is
incompatible with incert solution to append/extend certificates.

Add one more symlink to support any Fedora-like hardcoded expectations
for certificate bundles.

This fixes out of the box compat with dart-lang, which doesn't by
default support SSL_CERT_FILE variable like golang does.

References:
- https://github.com/dart-lang/sdk/blob/4fe2944776a0f8b24a7b8a63f67c06b82a808070/runtime/bin/security_context_linux.cc#L65
- https://github.com/golang/go/blob/95936844387c0158b773afa4ee6f99bd430791cf/src/crypto/x509/root_unix.go#L19
- https://askubuntu.com/questions/342484/etc-pki-tls-certs-ca-bundle-crt-not-found

### DIFF
--- a/ca-certificates.yaml
+++ b/ca-certificates.yaml
@@ -2,7 +2,7 @@ package:
   name: ca-certificates
   # manual: update java-cacerts
   version: "20240705"
-  epoch: 0
+  epoch: 1
   description: "CA certificates from the Mozilla trusted root program"
   copyright:
     - license: MPL-2.0 AND MIT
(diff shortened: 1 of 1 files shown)


k4leung4 (approved):

kranurag7 (approved):
confirming it works for dart. Thanks a ton. :)

PR #29970, opened 2024-10-03
logstash/GHSA-9hf4-67fc-4vf4

jamie-albert:
Simple bump of gem from puma dependency from 6.4.2 -> 6.4.3 to remediate GHSA-9hf4-67fc-4vf4. Done via replacing the value of the dependency version in the gemfile template before build.

### DIFF
--- a/logstash.yaml
+++ b/logstash.yaml
@@ -69,6 +69,7 @@ pipeline:
     runs: |
       echo "gem 'fugit', '1.11.1'" >> Gemfile.template
       echo "gem 'rexml', '3.3.6'" >> Gemfile.template
+      echo "gem 'puma', '6.4.3'" >> Gemfile.template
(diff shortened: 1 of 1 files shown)


logstash.yaml
```diff
@@ -69,6 +69,7 @@ pipeline:
     runs: |
       echo "gem 'fugit', '1.11.1'" >> Gemfile.template
       echo "gem 'rexml', '3.3.6'" >> Gemfile.template
+      echo "gem 'puma', '6.4.3'" >> Gemfile.template
```
kranurag7, line 72: given we are bumping a dependency and want to build the package again, I think we should bump the epoch too.

jamie-albert: Thx for catching that

debasishbsws (approved):

PR #29995, opened 2024-10-03
spark-3.5: add openjdk-21 support

mark-thm:
Fixes:

Related: https://github.com/wolfi-dev/os/pull/26803

#26803 failed to include jdk21 support because it's not strictly documented. In reality, Spark runs fine on jdk21 (and the new tests added in the PR should demonstrate that). Unfortunately, #26803 now forces inclusion of _some_ jdk so realistically it should allow for additional, valid jdks.


spark-3.5.yaml
```diff
@@ -53,14 +54,15 @@ data:
...
       17: "java-17-openjdk"
+      21: "java-21-openjdk"
 
 subpackages:
   - range: openjdk-versions
     name: ${{package.name}}-openjdk-${{range.key}}
     dependencies:
-      # Both `python` and `openjdk-{11,17}-default-jvm` are required for running Spark,
-      # but will be added in the image, since upstream supports different Python versions
-      # and Java versions. So we don't need to specify them here to avoid conflicts.
```
mark-thm, line 63: this comment doesn't make sense any more because prior PR forces inclusion of openjdk

spark-3.5.yaml
```diff
@@ -72,7 +74,6 @@ subpackages:
         - net-tools
         - logrotate
         - linux-pam
-        - procps
```
mark-thm, line 75: this is a duplicate entry

mamccorm:
Doesn't look like spark v3.5 supports java21, there is a beta being worked on for spark v4, which looks to be the first instance upstream will support. We'll not jump ahead of the gun here as we'd be on the hook to support a version upstream isn't yet, but appreciate the interest

PR #30030, opened 2024-10-03
Here are a bunch more smoke tests for subpackages (a-j)

dustinkirkland:
Recently, we merged a bunch of simple, cheap smoke tests for binaries installed by main packages in the default $PATH.

I've expanded my script to identify another ~2500 or so binaries that can be tested similarly, in subpackages.

I'm going through these a little more slowly, carefully this time around, reviewing every single change, and testing every single one of them in a local make test environment.  This first batch includes all packages that start with 'a' through 'j'.

Based on my local testing, I expect all of these to pass at this point.


joshrwolf:
this lgtm but will need another force merge.

we're not bumping the epoch here so nothing will be rebuilt postsubmit, but our presubmit CI is too crude to _only_ run the tests. I'm recommending we do the same thing we did for the last batch and catch the testing errors in postsubmit.

joshrwolf (dismissed):

philroche:
@dustinkirkland If the new binary found is a sub package is there a way to add a comment 

```
# below tests are for binaries built by subpackages
```

because we can have per sub package tests but this is adding those tests to the origin package test.

PR #30033, opened 2024-10-04
python-3.13: enable opt-in JIT

xnox:
Python 3.13 features [an experimental just-in-time (JIT) compiler](https://docs.python.org/3.13/whatsnew/3.13.html#whatsnew313-jit-compiler).

Compile support for it, but keep it disabled by default at runtime,
allowing users to opt into it.

> The --enable-experimental-jit option takes these (optional) values, defaulting to yes if --enable-experimental-jit is present without the optional value.
> yes-off: Build the JIT but disable it by default. To enable the JIT at runtime, pass the environment variable PYTHON_JIT=1.

### DIFF
--- a/python-3.13.yaml
+++ b/python-3.13.yaml
@@ -1,7 +1,7 @@
 package:
   name: python-3.13
   version: 3.13.0_rc3
-  epoch: 0
+  epoch: 1
   description: "the Python programming language"
   copyright:
     - license: PSF-2.0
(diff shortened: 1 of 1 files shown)


smoser (commented):
Is there a way that we can test that it gets enabled? Even just running with the env set and unset and then querying that Python says on or off as expected.

xnox:
> Is there a way that we can test that it gets enabled? Even just running with the env set and unset and then querying that Python says on or off as expected.

ack, also fails to build.

PR #30068, opened 2024-10-04
Add py3-supported-build-base package

smoser:
This will allow us to simplify the build environment dependencies
similar to the way 'build-base' does for C.

### DIFF
--- a/py3-supported-python.yaml
+++ b/py3-supported-python.yaml
@@ -1,7 +1,7 @@
 package:
   name: py3-supported-python
   version: 1
-  epoch: 2
+  epoch: 3
   description: metapackage to install all supported versions of python
   copyright:
     - license: "MIT"
@@ -15,8 +15,15 @@ package:
 environment:
   contents:
     packages:
-      - build-base
       - busybox
+
+data:
+  - name: py-versions
+    items:
+      3.10: '310'
+      3.11: '311'
+      3.12: '312'
+      3.13: '300'
 
 subpackages:
   - name: ${{package.name}}-dev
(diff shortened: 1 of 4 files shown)


smoser:
I think this covers a reasonable set without going too far:
I ran the python program below on 'py3-*.yaml' (537 packages). it ignores anything that isn't multi-versioned, so 159 py3-* packages.

These are all values > 6:
```
py3-supported-poetry-core: 7
cython: 7
openssf-compiler-options: 7
py3-supported-cython: 7
py3-supported-hatch-fancy-pypi-readme: 7
maturin: 7
py3-supported-maturin: 7
py3-supported-setuptools-scm: 9
py3-supported-hatch-vcs: 12
py3-supported-installer: 17
py3-supported-flit-core: 20
py3-supported-hatchling: 22
py3-supported-build: 26
py3-supported-python-dev: 26
py3-supported-setuptools: 48
py3-supported-wheel: 69
py3-supported-python: 81
wolfi-base: 106
ca-certificates-bundle: 108
build-base: 134
py3-supported-pip: 149
busybox: 156
```

```python
#!/usr/bin/python3

import yaml
import sys

totals = {}
for f in sys.argv[1:]:
    if not f.endswith(".yaml"):
        continue
    with open(f, "rb") as fp:
        yamdata = yaml.safe_load(fp)

    if "py-versions" not in [v.get("name") for v in yamdata.get("data", {})]:
        # print("skip %s" % f)
        continue

    envp = yamdata["environment"]["contents"]["packages"]

    for p in envp:
        if p not in totals:
            totals[p] = 0
        totals[p] = totals[p] + 1

for (p, num) in sorted(totals.items(), key=lambda i: i[1]):
    print("%s: %d" % (p, num))
```

py3-supported-python.yaml
```diff
@@ -28,6 +35,38 @@ subpackages:
...
 
+  - range: py-versions
+    name: py${{range.key}}-build-base
+    description: ${{vars.pypi-package}} core/base build packages.
+    dependencies:
+      runtime:
+        - py${{range.key}}-build
+        - py${{range.key}}-installer
+        - py${{range.key}}-pip-base
+        - py${{range.key}}-setuptools
```
pnasrat, line 46: You probably want setuptools-scm by default here too to ensure version correctness from git-checkout.

pnasrat (dismissed):

py3-supported-python.yaml
smoser: I did consider setuptools-scm as you're right it is useful.
i left it out because of its dependency tree. just checked, and it just adds typping-extensions, so i'll add it.

justinvreeland (approved):

PR #30081, opened 2024-10-04
Add py3-fromager and py3-version its dependencies

pnasrat:
- **py3-wrapt: py-versioned**
- **py3-webencodings: py-versioned**
- **py3-html5lib: py-versioned**
- **py3-pynacl: python versions** includes fixes for missing cffi dep and build with system libsodium
- **py3-deprecated: py-versioned**
- **py3-pyjwt: py-versioned**
- **py3-github: py-versioned**
- **new py3 versioned package: python-pypi-mirror**
- **new py3 versioned package: py3-elfdeps**
- **new py3 versioned package: py3-resolvelib**
- **new py3 versioned package: py3-stevedore**
- **new py3 versioned package: py3-fromager**


py3-elfdeps.yaml
```diff
@@ -0,0 +1,72 @@
...
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - py3-supported-pip
+      - wolfi-base
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/python-wheel-build/elfdeps.git
```
xnox, line 35: we have no linting for this; but most github.com urls are without ".git" suffix, note this url is used to generate PURLs and download urls, and sometimes visible as human clickable too. Thus prefer the without ".git". It then is also easier to parse.

Note this is pedantic/cosmetic request, as there are redirect via any protocol both ways between suffix-less and .git suffix urls.

xnox (dismissed):

py3-elfdeps.yaml
pnasrat: Done.

xnox (dismissed):
Reviewing changes since my last review => three .git gone, thanks.

xnox (approved):

PR #30156, opened 2024-10-07
Add tests for yarn, and fix missing runtime dep

dustinkirkland:
Added tests for yarn, which found a missing runtime dependency on nodejs

### DIFF
--- a/yarn.yaml
+++ b/yarn.yaml
@@ -44,8 +44,8 @@ update:
 
 test:
   pipeline:
-    runs: |
-      yarn --version
-      yarn --help
-      yarnpkg --version
-      yarnpkg --help
+    - runs: |
+        yarn --version
+        yarn --help
+        yarnpkg --version
+        yarnpkg --help
--- a/yarn.yaml
+++ b/yarn.yaml
@@ -1,10 +1,13 @@
 package:
   name: yarn
   version: 1.22.22
-  epoch: 1
+  epoch: 2
   description: Fast, reliable, and secure dependency management for Node.js
   copyright:
     - license: BSD-2-Clause
+  dependencies:
+    runtime:
+      - nodejs
 
 environment:
   contents:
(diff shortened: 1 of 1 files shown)


mamccorm (dismissed):
Nice, thanks!

yarn.yaml
```diff
@@ -1,10 +1,13 @@
...
   name: yarn
   version: 1.22.22
-  epoch: 1
+  epoch: 2
   description: Fast, reliable, and secure dependency management for Node.js
   copyright:
     - license: BSD-2-Clause
+  dependencies:
+    runtime:
+      - nodejs
```
mamccorm, line 10: It might be safer to drop this and just define in your test - I don't disagree this is the more correct thing to do, but we could have other packages using yarn, with a different version of nodejs.

If we add this as a runtime dep, we could break those packages / images and end up installing multiple versions of node

dustinkirkland: Hmm, maybe.  It would be easy to just fix the test.  But are we treating the symptoms or the condition?  It's weird to me that I could "apk add yarn", and it not work at all out of the box...

mamccorm: Does look like we have different melange packages that require different versions of nodejs. theres a couple here pinned to nodejs-20:

 - https://github.com/wolfi-dev/os/blob/main/argo-workflows.yaml
 - https://github.com/wolfi-dev/os/blob/main/argo-cd-2.12.yaml
 
Whereas some other packages are always happy with the latest. We have multiple node versions in wolfi. If you were to include nodejs as a runtime dep here, it'd mean those melanges would attempt to re-install a later nodejs version, so this change would break those builds

dustinkirkland: Okay -- so to be clear, you'd be happy with this merge proposal if I change the runtime dep to a specific (known, working) version of nodejs?

mamccorm:
Closing this - as we've now got functional tests for yarn which were added in a later commit:
 - https://github.com/wolfi-dev/os/pull/32159

PR #30167, opened 2024-10-07
add openbao package

wojciechka:
#### For new package PRs only


wojciechka:
Working on adding an advisory

kranurag7 (commented):
Great work!! looks like we need will also a compat package for this linking binary at /bin/bao and a further linking of the same at /bin/vault.

wojciechka:
> Great work!! looks like we need will also a compat package for this linking binary at /bin/bao and a further linking of the same at /bin/vault.

@kranurag7 added - please review the compat package definition

openbao.yaml
```diff
@@ -0,0 +1,51 @@
...
+  - runs: |
+      mkdir -p ${{targets.destdir}}/usr/bin
+      install -m755 -D src/bin/bao "${{targets.destdir}}/usr/bin/"
+
+update:
+  enabled: true
+  github:
+    identifier: openbao/openbao
+    strip-prefix: v
+    use-tag: true
```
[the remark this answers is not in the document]
wojciechka: Can you please make a suggested change for this? I don't think there's an easy way to test it and I based it on other packages that use GitHub releases, so safer you add what you meant here and I just incorporate it into PR.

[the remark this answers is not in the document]
wojciechka: Actually, nevermind, I think it was enough to remove `use-tag`. PTAL and let me know if it's enough.

kranurag7:
@wojciechka thanks for the work here. 

We'll need the following change:
`/bin/bao` should link to `/usr/bin/bao` and `/bin/vault` should link to `/bin/bao`

powersj:
@cipherboy FYI :)

cipherboy:
@wojciechka @powersj Thank you both for getting this started!

> * This PR links to the upstream project's support policy (e.g. `endoflife.date`)

As a FYI, this will be discussed on this Thursday's [OpenBao TSC call](https://lf-edge.atlassian.net/wiki/spaces/OP/pages/62586881/2024-10-10+OpenBao+TSC+Meeting). I suspect since the community is smaller without many commercial backers, we'll likely stick to an only-latest support policy.

But once that decision is made we'll publish something officially. I think https://github.com/openbao/openbao/issues/426 tracks this on our side.

openbao.yaml
```diff
@@ -19,25 +19,23 @@ environment:
...
   - runs: |
-      # echo ".cache" >>.git/info/exclude
-      # echo ".config" >>.git/info/exclude
-      # echo ".gitconfig" >>.git/info/exclude
-      # echo ".npm" >>.git/info/exclude
-      # echo ".yarn" >>.git/info/exclude
-      # echo "melange-out" >>.git/info/exclude
       export LD_FLAGS="-X github.com/openbao/openbao/version.fullVersion=${{package.version}}"
-      make ember-dist bin GOARCH=$(go env GOARCH)
+      make -C src ember-dist
```
cipherboy, line 31: Just a note but we don't yet support the Web UI because it still contains HashiCorp branded assets. I'd suggest not building it until a release with the UI re-enabled is released.

wojciechka: The problem is building the binary fails if the UI is not present:
```
http/assets.go:16:12: pattern web_ui/*: no matching files found
```

It's possible to workaround this by creating an empty file called `http/web_ui/index.html`, but that seems like a hacky solution and we'd need to re-add `make ember-dist` in the next release.

When is this planned to be updated so we can ship the UI? Perhaps we can consider holding off on merging this and building images until the UI is ready?

If not, then we can go ahead with an empty index.html file, but is it possible to let us know once we should start including the UI?

wojciechka:
> We'll need the following change:
> `/bin/bao` should link to `/usr/bin/bao` and `/bin/vault` should link to `/bin/bao`

@kranurag7 done, PTAL

openbao.yaml
```diff
@@ -0,0 +1,64 @@
...
+      install -m755 -D src/bin/bao "${{targets.destdir}}/usr/bin/"
+
+  - uses: strip
+
+subpackages:
+  - name: openbao-compat
+    description: "Compatibility package to allow replacing vault image"
+    pipeline:
+      - runs: |
+          mkdir -p mkdir -p ${{targets.contextdir}}/bin ${{targets.contextdir}}/usr/bin
```
kranurag7, line 51:
```suggestion
          mkdir -p ${{targets.contextdir}}/usr/bin
          mkdir -p ${{targets.contextdir}}/bin 
```

kranurag7 (commented):
thanks for the work again, the symlink looks good now.

openbao.yaml
wojciechka: Thanks. Updated.

openbao.yaml
cipherboy: Ahh, I see, this is because [`make bin`](https://github.com/openbao/openbao/blob/main/Makefile#L26-L28) does a `ui`-enabled build. We should fix that upstream, but my suggestion to [others doing distributions](https://github.com/openbao/openbao/issues/431#issuecomment-2252912056) was just to call `go build` directly as that's what the go-releaser tooling we use for releases does.

You'll have to plumb through the hashes and commit date though, but you'll want to use these over build time for reproducibility. I've done something similar for GitLab here: https://gitlab.com/gitlab-org/govern/secrets-management/openbao-internal/-/blob/main/Makefile?ref_type=heads#L21 

(Though, note that we're doing a patching based flow, so you won't need the two separate commits which complicates things.)

wojciechka: @cipherboy Good point that we can just use `go build` directly and pass the version details instead of calling Makefile that in turn calls the script and passes the `ui` tag. I switched to this approach just now.

openbao.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - license: MPL-2.0
+
+environment:
+  environment:
+    CGO_ENABLED: "0"
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - go
```
ajayk, line 16: minor nit `go` is provided by go/build 
```suggestion
```

kranurag7 (changes requested):
we can consider using go/bump pipeline to fix the CVEs in here? 

```yaml
  - uses: go/bump
    with:
      deps: github.com/anchore/archiver/v3@v3.5.2
      replaces: github.com/mholt/archiver/v3=github.com/anchore/archiver/v3@v3.5.2
```

wojciechka:
> we can consider using go/bump pipeline to fix the CVEs in here?
> 
> ```yaml
>   - uses: go/bump
>     with:
>       deps: github.com/anchore/archiver/v3@v3.5.2
>       replaces: github.com/mholt/archiver/v3=github.com/anchore/archiver/v3@v3.5.2
> ```

No, I would not do that. That is a fork, the project is aware of the issue (I emailed openbao-security list yesterday). It is also confirmed that it is a false positive in our case - it is used by debug subcommand and the package is only writing archives, the CVE talks about user-injected archives.

I believe It's better to add an advisory for this and mark it as false positive.

ajayk (dismissed):

kranurag7 (dismissed):

openbao.yaml
```diff
@@ -0,0 +1,64 @@
...
+environment:
+  environment:
+    CGO_ENABLED: "0"
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+      - nodejs
+      - npm
+      - yarn
```
EyeCantCU, line 18: We should be able to drop the environment here. Since we aren't building the UI, `go/build` includes any dependencies we need at buildtime. We also shouldn't need to explicitly disable CGO. We could leave `nodejs`, `npm`, and `yarn` here commented out with a note that they're to be used when the UI is available

openbao.yaml
```diff
@@ -0,0 +1,64 @@
...
+        -X github.com/openbao/openbao/version.BuildDate='$(git show --no-show-signature -s --format=%cd --date=format:"%Y-%m-%dT%H:%M:%SZ" HEAD)'
+
+  - uses: strip
+
+subpackages:
+  - name: openbao-compat
+    description: "Compatibility package to allow replacing vault image"
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/usr/bin
```
EyeCantCU, line 50: Since we aren't linking anything here, we can remove this line

EyeCantCU (changes requested):

openbao.yaml
wojciechka: Good point. Refactored commenting out of UI - adding a note at the top and then same comment in all the places that need uncommenting to build the UI in the future.

Re CGO - this is how it's also built upstream:
https://github.com/openbao/openbao/blob/main/.goreleaser-template.yaml#L31
https://github.com/openbao/openbao/blob/main/Makefile#L18

openbao.yaml
wojciechka: Good point. Done.

openbao.yaml
wojciechka: After giving it some thought I noticed that not disabling CGO may be a better choice. OpenBao seems to work without it and is built in both modes based on Makefile.

EyeCantCU (dismissed):

EyeCantCU (approved):

PR #30212, opened 2024-10-07
bump some packages for rebuild that failed 'apk fetch pkg'

smoser:
Of py3-*.yaml packages, these failed download with the loop above.

    $ pkgs=$(for f in py3-*.yaml; do echo ${f%.yaml}; done)
    $ for pkg in $pkgs; do apk fetch $pkg; done
    ...
    > Downloading py3-s3transfer-0.10.2-r0
    > ERROR: py3-s3transfer-0.10.2-r0: BAD archive
    ...

### DIFF
--- a/py3-boolean.py.yaml
+++ b/py3-boolean.py.yaml
@@ -2,7 +2,7 @@
 package:
   name: py3-boolean.py
   version: "4.0"
-  epoch: 1
+  epoch: 2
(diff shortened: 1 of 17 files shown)


smoser:
py3-ml-metadata as also failing with fetch, but it needed an update, so i just let https://github.com/wolfi-dev/os/pull/29842 bump it.

EyeCantCU (approved):

PR #30217, opened 2024-10-07
Make py3-rich installation get python 3.12, not 3.13

smoser:
The '313' here in the py-versions key means that installation of py3-rich will get py3.13-rich.

I noticed this because py3-hatch's test will fail. The reason for failure is that 2 different pythons are installed and the python/import test correctly expects only one to be installed.

### DIFF
--- a/py3-rich.yaml
+++ b/py3-rich.yaml
@@ -1,7 +1,7 @@
 package:
   name: py3-rich
   version: 13.9.2
-  epoch: 0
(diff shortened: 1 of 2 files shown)


smoser:
I hope that this PR and a withdraw packages run will result in https://github.com/wolfi-dev/os/pull/30212 passing c-i.

justinvreeland (approved):

PR #30218, opened 2024-10-07
test/pkgconf pipelines for all dev packages that currently succeed

dustinkirkland:
This set of tests uses the new test/pkgconf pipeline.

All ~263 of these are confirmed passing right now in make test/*


justinvreeland (dismissed):
Love adding more tests

justinvreeland (dismissed):

joshrwolf (dismissed):
this lgtm, but similar to the rest of these batches it'll need a force merge. cc @mamccorm

dustinkirkland:
@joshrwolf @murraybd @xnox Is there anything else I need to do here?  I'd love to get this merged, so that I can start working on the next batch...

murraybd:
@dustinkirkland My investigation indicates that the CI failure is just with cyrus-sasl failing to build and I'll have an independent fix for that shortly. So if you just drop the test changes for cyrus-sasl I'll include that in a PR fixing the build failure.

dustinkirkland:
Sweet!  Done!

xnox:
Automatic dependencies have landed. I would rather do epoch bump-full rebuilds for all the things. Because then the next batch should be enabled strict pipeline checks for all packages.

dustinkirkland:
Okay -- so I'll do the work to do this (epoch bumps) -- if both @murraybd and @joshrwolf agree?

joshrwolf (approved):
like the other batches of these test only changes, we're not bumping the epoch here so this doesn't trigger a package build in postsubmit, so I'm +1 for force merging this in and catching any errors (which should be only due to discrepancies in our local/prod build environments) in post

xnox (approved):
Test only change.

imjasonh:
Force merging because some packages fail to rebuild now. We'll test everything after this and fix up anything that fails.

PR #30225, opened 2024-10-07
python-3.13: move from rc to release version

ajayk:


python-3.13.yaml
```diff
@@ -10,8 +10,8 @@ package:
     memory: 8Gi
   dependencies:
     provides:
-      - python3=0
-      - python-3=0
+      - python3=${{package.full-version}}
+      - python-3=${{package.full-version}}
     runtime:
```
xnox, line 15: this and the other -dev one will make it the default. Are we ready? has everything completed rebuilding for 3.13?

smoser: I doing think we should do this yet.

smoser (changes requested):
Let's land the update change (tag filter, I'm bummed I messed that up) and the version change.

I'm indifferent on the vars/ majorminormicro change. It works either way.

smoser (commented):
Can you drop the definitions of the transforms that you have stopped using?

`versionMajMinMic` and `tagname`.

smoser (approved):
thank you.

PR #30254, opened 2024-10-08
Add new python package - py3-propcache

mamccorm:
Adds a new python package: py3-propcache. The primary reason for adding this, was due to a new buildtime dependency introduced by the following (existing) package: https://github.com/wolfi-dev/os/pull/30252.

Theres more info about the buildtime [dependency introduction here](https://github.com/aio-libs/yarl/pull/1169).

### DIFF
--- a/py3-propcache.yaml
+++ b/py3-propcache.yaml
@@ -1,0 +1,84 @@
+package:
+  name: py3-propcache
+  version: 0.2.0
+  epoch: 0
+  description: Fast property caching
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    provider-priority: 0
+
+vars:
+  pypi-package: propcache
+
+data:
+  - name: py-versions
+    items:
+      3.10: '310'
+      3.11: '311'
+      3.12: '312'
+      3.13: '300'
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - py3-supported-cython
+      - py3-supported-expandvars
+      - py3-supported-pip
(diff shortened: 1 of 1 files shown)


py3-propcache.yaml
```diff
@@ -0,0 +1,84 @@
...
+vars:
+  pypi-package: propcache
+
+data:
+  - name: py-versions
+    items:
+      3.10: '310'
+      3.11: '311'
+      3.12: '312'
+      3.13: '300'
```
mamccorm, line 20: the `3.13: '300' entry here is correct and consistent with all our other multi-versioned python packages in wolfi.

py3-propcache.yaml
```diff
@@ -0,0 +1,84 @@
...
+    - uses: python/import
+      with:
+        import: ${{vars.pypi-package}}
+
+update:
+  enabled: true
+  github:
+    identifier: aio-libs/propcache
+    strip-prefix: v
+    tag-filter: v
```
kranurag7, line 84: can be omitted considering upstream just name the releases as 0.2.0 (ref: https://github.com/aio-libs/propcache/releases) or we can append `use-tag: true`

Dentrax (approved):

py3-propcache.yaml
mamccorm: The release names follow 0.2.0 but the tag associated are still pre-pended with a v. If we omit then the build will fail to find the release tag unfortunatelt. We also prefer to poll releases over tags (if they exist). We often find pre-release tags and failing package PRs in sustaining that wouldn't have been an issue if we'd polled releases instead :)

PR #30285, opened 2024-10-08
Move building of python arrow modules from pyarrow to apache-arrow.

smoser:
A couple things here:
1. fix FTBFS state that apache-arrow was currently in.

  * need zstd-static because:

       CMake Error at /usr/lib64/cmake/zstd/zstdTargets.cmake:107 (message):
         The imported target "zstd::libzstd_static" references the file
            "/usr/lib/libzstd.a" uses=cmake/configure
         but this file does not exist.  Possible reasons include:

2. cherry-pick a patch (from 2016) to rapidjson that allows a user of it
   (apache-arrow) to build with gcc-14.

3. Add environment variables to build so the python build modules
   have access.  I had hoped to localize these or use substitution
   in the environment block, but that is not supported.

   https://github.com/chainguard-dev/melange/issues/1548
   https://github.com/chainguard-dev/melange/issues/1402

4. drop unnecessary runtime dependency from libarrow to libarrow.

5. remove pyarrow.

6. bump the epoch on apache-arrow to 3 to be newer than pyarrow was.

7. provide multiple versions of the python modules, and make them provide 'pyarrow' for backwards compat.


smoser:
it looks to me right now that apache-arrow is just FTBFS. :cry:

justinvreeland (approved):

PR #30287, opened 2024-10-08
samba-server-libs: fix misplaced libraries

gdonval:
Server-related libs were copied from `origin:/usr/lib/samba` to `dest:/usr/lib`. This breaks building SSSD (which I'm trying to build for Wolfi).


gdonval:
Re package update config check: let's do the update to the newer version in another PR?

PR #30294, opened 2024-10-08
add hadoop-runner package

kranurag7:
follow up from https://github.com/wolfi-dev/os/pull/29037 

this initialized hadoop-runner for both x64 and arm given we don't have problems here. In future, we want to proceed with arm then we can use this as runtime.


kranurag7:
closing given this is no longer tracked here.

PR #30305, opened 2024-10-08
Add otel-plugin-nginx, otel-cpp

wlynch:
- Adds opentelemetry-cpp: c++ client for otel (dependency for otel-plugin-nginx).
- Adds otel-plugin-nginx
- Modifies protobuf to disable generation of libupb. This is a static internal dependency of protobuf. If this ends up being load bearing for php and python protobuf packages, we should repackage this into a subpackage.

> [!NOTE]  
> This is the [community otel nginx plugin](https://github.com/open-telemetry/opentelemetry-cpp-contrib/tree/main/instrumentation/nginx) that was [used in ingress-nginx](https://github.com/kubernetes/ingress-nginx/blob/f369ffb0734cc63fdf85959c4e41be0606ca8f1e/images/opentelemetry/rootfs/build.sh#L129-L143). This is **not** the [official nginx otel plugin](https://github.com/nginxinc/nginx-otel).

Related: https://github.com/chainguard-dev/image-requests/issues/4407

#### For new package PRs only

### DIFF
--- a/opentelemetry-cpp.yaml
+++ b/opentelemetry-cpp.yaml
@@ -1,0 +1,49 @@
+package:
+  name: opentelemetry-cpp
+  version: 1.17.0
+  epoch: 0
+  description: The OpenTelemetry C++ Client
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - abseil-cpp-dev
+      - benchmark-dev
+      - busybox
+      - c-ares-dev
+      - cmake
+      - gcc
+      - glibc-dev
+      - grpc-dev
+      - gtest-dev
+      - icu-dev
+      - make
+      - openssl-dev
+      - pkgconf
+      - protobuf-dev
+      - re2-dev
+      - systemd-dev
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/open-telemetry/opentelemetry-cpp
+      tag: v${{package.version}}
+      expected-commit: fa0a5200ddbe9ae6f487a228e658aedc102dab56
+
+  - runs: |
+      mkdir build
+      cd build
+      cmake -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DBUILD_SHARED_LIBS=ON -DWITH_OTLP_GRPC=ON -DWITH_ABSEIL=ON ..
+      cmake --build . --target all
+      cmake --install . --prefix ${{targets.contextdir}}/usr
+
+  - uses: strip
+
+update:
+  enabled: true
+  github:
+    identifier: open-telemetry/opentelemetry-cpp
+    strip-prefix: v
--- a/opentelemetry-cpp.yaml
+++ b/opentelemetry-cpp.yaml
@@ -32,13 +32,25 @@ pipeline:
       repository: https://github.com/open-telemetry/opentelemetry-cpp
       tag: v${{package.version}}
       expected-commit: fa0a5200ddbe9ae6f487a228e658aedc102dab56
+
   - uses: cmake/configure
     with:
       opts: -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DBUILD_SHARED_LIBS=ON -DWITH_OTLP_GRPC=ON -DWITH_ABSEIL=ON
+
   - uses: cmake/build
+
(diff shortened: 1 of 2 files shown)


opentelemetry-cpp.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+  - runs: |
+      mkdir build
+      cd build
+      cmake -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DBUILD_SHARED_LIBS=ON -DWITH_OTLP_GRPC=ON -DWITH_ABSEIL=ON ..
+      cmake --build . --target all
+      cmake --install . --prefix ${{targets.contextdir}}/usr
+
+  - uses: strip
+
```
EyeCantCU, line 44: Any docs/infodirs we can split into subpackages, development files. etc?

opentelemetry-plugin-nginx.yaml
```diff
@@ -0,0 +1,125 @@
...
+      mkdir build
+      cd build
+      cmake ..
+      make
+
+      mkdir -p ${{targets.contextdir}}/usr/share/nginx/modules
+      cp -p otel_ngx_module.so ${{targets.contextdir}}/usr/share/nginx/modules/
+
+  - uses: strip
+
```
EyeCantCU, line 53: Same here

opentelemetry-cpp.yaml
```diff
@@ -0,0 +1,49 @@
...
+  - uses: git-checkout
+    with:
+      repository: https://github.com/open-telemetry/opentelemetry-cpp
+      tag: v${{package.version}}
+      expected-commit: fa0a5200ddbe9ae6f487a228e658aedc102dab56
+
+  - runs: |
+      mkdir build
+      cd build
+      cmake -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DBUILD_SHARED_LIBS=ON -DWITH_OTLP_GRPC=ON -DWITH_ABSEIL=ON ..
```
EyeCantCU, line 39: We may be able to leverage our cmake pipelines here

opentelemetry-plugin-nginx.yaml
```diff
@@ -0,0 +1,125 @@
...
+      branch: main
+      expected-commit: f6d29426ee9b4d6b476c09ca3cb9bed3cf23906f
+
+  - runs: |
+      cd instrumentation/nginx
+
+      mkdir build
+      cd build
+      cmake ..
+      make
```
EyeCantCU, line 47: We have an `autoconf/make` pipeline we may be able to use

opentelemetry-plugin-nginx.yaml
```diff
@@ -0,0 +1,125 @@
...
+      repository: https://github.com/open-telemetry/opentelemetry-cpp-contrib
+      branch: main
+      expected-commit: f6d29426ee9b4d6b476c09ca3cb9bed3cf23906f
+
+  - runs: |
+      cd instrumentation/nginx
+
+      mkdir build
+      cd build
+      cmake ..
```
EyeCantCU, line 46: `cmake/build` should cover this

opentelemetry-cpp.yaml
wlynch: Done.

opentelemetry-plugin-nginx.yaml
wlynch: This can't use `cmake/install` though, since this cmake pipeline outputs directly at the prefix dir and we need to move this to the nginx modules directory.

opentelemetry-cpp.yaml
EyeCantCU: We have pipelines in melange that can be used to split things away: https://github.com/chainguard-dev/melange/tree/main/pkg/build/pipelines/split

Typically, we split whatever we can. I.E., those headers (the `.h` files) would be split away with `split/dev` and docs would be split with `split/manpages` and `split/info` if an info directory exists. It looks like in your case, you just need to create a `-dev` subpackage that uses the `split/dev` pipeline

wlynch: 💯 exactly what I wanted. Thanks! 🎉

EyeCantCU (approved):

PR #30331, opened 2024-10-08
mesa: update for llvm-19

justinvreeland:
The patch variable is leftover from being imported from alpine and we never set it. We don't need it for our builds.


xnox (changes requested):
epoch bump

xnox (approved):

PR #30337, opened 2024-10-08
imath: Remove uneeded llvm dep

justinvreeland:
It's been here from first time the package was imported but it doesn't seem to be used.


justinvreeland:
some of these changes ended up in the buck2 commit

mamccorm (approved):

PR #30338, opened 2024-10-08
buck2: Remove uneeded llvm dep

justinvreeland:


xnox (changes requested):
changes to melange file; that affect buildability require epoch bump.

we want to verify that this change does indeed still builds.

xnox (approved):

PR #30396, opened 2024-10-09
python/multi-version batch 07

smoser:
Convert packages to python multi-version.

This is part of a series of programatic changes to
add the ability to support python packages for multiple
python versions.

Other changes to py3-face.yaml, py3-requests-oauthlib.yaml, py3-trio.yaml
were to drop an additional 'provides'.  That was only supposed to be
added if the package was not 'py3-<pypi-pkgname>' and for these it is.

### DIFF
--- a/py3-face.yaml
+++ b/py3-face.yaml
@@ -1,7 +1,7 @@
 package:
   name: py3-face
   version: 22.0.0
-  epoch: 3
+  epoch: 4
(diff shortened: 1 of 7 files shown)


justinvreeland (dismissed):

smoser:
these all landed elsewhere.

PR #30419, opened 2024-10-09
Drop py3-typing. It is obsolete since python 3.5.

smoser:
Per its pypi index page https://pypi.org/project/typing/ :

> Python 3.5 and later, the typing module lives in the stdlib
> and installing this package has NO EFFECT, because stdlib
> takes higher precedence than the installation directory.


pnasrat (approved):
LGTM

PR #30428, opened 2024-10-09
Libreoffice: Unlocked python 3.11 dependency, switched to python3(.12)

hbh7:
Fixes: Libreoffice held an older Python version (3.11), which lead to some difficulties and appears to not actually be needed. This bumps to python3, which currently resolves to 3.12. 

Related: https://github.com/chainguard-dev/enterprise-packages/pull/8284


hbh7:
Apparently there's a version bump PR already, should probably integrate these changes here: https://github.com/wolfi-dev/os/pull/27736

justinvreeland:
I'd rather keep the changes separate so we can see the history more clearly unless the python causes a build issue.

justinvreeland (approved):

PR #30438, opened 2024-10-09
protobuf: disable upb

wlynch:
Removes upb which is causing issues with downstream dependencies (otel-nginx - see https://github.com/wolfi-dev/os/pull/30305). This is documented as being an [internal dependency of grpc](https://github.com/protocolbuffers/protobuf/blob/6690ab42d855ea19d9a24cd99b0375910ea772ca/cmake/libupb.cmake#L14-L15).
I did a spot check locally and py3-protobuf and php-protobuf can still build.

Related: https://github.com/chainguard-dev/image-requests/issues/4407, https://github.com/wolfi-dev/os/pull/30305

There is a newer version of protobuf available, but this breaks other packages and requires additional changes.

### DIFF
--- a/protobuf.yaml
+++ b/protobuf.yaml
@@ -1,7 +1,7 @@
 package:
   name: protobuf
   version: 3.27.4
-  epoch: 0
+  epoch: 1
   description: Library for extensible, efficient structure packing
   copyright:
     - license: BSD-3-Clause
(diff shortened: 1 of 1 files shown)


cpanato (approved):

cpanato:
can this impact any existing packages that use this?

wlynch:
> can this impact any existing packages that use this?

I thought based on the docs that this might impact python or php protobuf libraries, but these packages are still green even with this change.

PR #30441, opened 2024-10-09
python/multi-version batch 10

smoser:
Convert packages to python multi-version.

This is part of a series of programatic changes to
add the ability to support python packages for multiple
python versions.


py3-jmespath.yaml
```diff
@@ -1,45 +1,110 @@
...
+          mv ./cleanup/${{range.key}}/bin ${{targets.contextdir}}/usr/
+    test:
+      environment:
+        contents:
+          packages:
+            - apk-tools
+      pipeline:
+        - runs: |
+            apk info -L py${{range.key}}-${{vars.pypi-package}}-bin > "pkg.list"
+            echo "Please write a test for these:"
```
justinvreeland, line 80: Is this a test that warns people to write a test if the binaries don't already have them?

justinvreeland (approved):

py3-jmespath.yaml
smoser: thats the point of it yea, to just get some output and at least then see if '--help' llooks like it works.

PR #30459, opened 2024-10-09
grpc: Version-streamify grpc-1.66

wlynch:
There are breaking changes in grpc-1.67 that break falco. [grpc support policy says that they support n-1 versions](https://grpc.io/docs/what-is-grpc/faq/#how-long-are-grpc-releases-supported-for), so version-streamify the existing grpc-1.66 package in preparation for a separate grpc-1.67 package.

#### For new package PRs only

#### For new version streams

Note: Had to deviate from the naming conventions somewhat for py3 subpackage. lmk if there's a better way to do this.

### DIFF
--- a/grpc-1.66.yaml
+++ b/grpc-1.66.yaml
@@ -12,6 +12,12 @@ package:
     provides:
       - grpc=${{package.full-version}}
 
+var-transforms:
+  - from: ${{package.version}}
+    match: ^(\d+\.\d+)\.\d+$
+    replace: "$1"
+    to: major-minor-version
+
 environment:
   contents:
     packages:
@@ -99,7 +105,7 @@ pipeline:
       sh -c 'rm $0 && ln -s /etc/ssl/certs/ca-certificates.crt $0' "{}" \;
 
 subpackages:
-  - name: ${{package.name}}-py3-grpcio
+  - name: py3-grpcio-${{vars.major-minor-version}}
     description: "gRPC Python HTTP/2-based RPC framework"
     dependencies:
       runtime:
--- a/grpc-1.66.yaml
+++ b/grpc-1.66.yaml
@@ -1,5 +1,5 @@
 package:
-  name: grpc
+  name: grpc-1.66
   version: 1.66.2
   epoch: 0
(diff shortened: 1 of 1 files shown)


ajayk (changes requested):
Needs tag filter changes

wlynch:
Good catch! Done.

grpc-1.66.yaml
```diff
@@ -96,22 +99,26 @@ pipeline:
       sh -c 'rm $0 && ln -s /etc/ssl/certs/ca-certificates.crt $0' "{}" \;
 
 subpackages:
-  - name: py3-grpcio
+  - name: ${{package.name}}-py3-grpcio
```
ajayk, line 102: I am not completely sure about this package name probably we can use py-3-grpcio-${{vars.major-minor}}
instead of grpc-1.66-py3-grpcio

wlynch: Can we update the checklist with more detailed naming styling? 🙏
I originally did something like that, but ended up not doing it because of the checklist recommendation.

For `major-minor` - is there a built-in variable I can use, or do I need to do the var transformation myself?

ajayk: You need to use var transformation

https://github.com/wolfi-dev/os/blob/main/gitlab-kas-17.4.yaml#L12-L15

ajayk: For regular sub-package ${{package.name}}-xyz would work , but this seems like a py3-package  , would defer to @EyeCantCU or @pnasrat for the naming styling here

pnasrat: I think py3-grcpio-${{vars.major-minor}} for the versioned one makes sense. We wouldn't expect the two versions of the python module to be co-installable

The checklist is somewhat simplified and predates version streaming I believe - I'm not sure who owns that but there are multiple cases we do other things

EyeCantCU: In agreement with Pris

ajayk (approved):

EyeCantCU (approved):

PR #30470, opened 2024-10-09
percona-server-8.4: use system dependecies

Dentrax:
Percona Server 8.4 was using bundled dependencies. This PR changes it to use system dependencies instead.


EyeCantCU (changes requested):
Needs an epoch bump, but otherwise looks great!

EyeCantCU (approved):

PR #30474, opened 2024-10-09
diffoscope et all: Update for multiversion and misc cleanup

justinvreeland:
It seems a bit ridiculous but it's how we've done other packages. These were mostly re-done with my draft melange convert code with some minor updates. I mostly just wanted liabrchive to be installed when I install diffoscope


justinvreeland:
@smoser These were generated with the updates from melange. Small improvements.

PR #30481, opened 2024-10-10
New package libsdl3

xnox:
Whilst it is pre-release, it has declared to have stable abi https://github.com/libsdl-org/SDL/releases/tag/preview-3.1.3

And we are starting to see cutting edge software releases that require
libsdl3. It is likely that libsdl2 will be required for a long time,
even when libsdl3 goes stable.

Package both during this transition period.

Needed for:
- https://github.com/wolfi-dev/os/pull/27343

### DIFF
--- a/libsdl3.yaml
+++ b/libsdl3.yaml
@@ -1,0 +1,59 @@
+package:
+  name: libsdl3
+  version: 3.1.3
+  epoch: 0
+  description: Simple DirectMedia Layer is a cross-platform development library designed to provide low level access to audio, keyboard, mouse, joystick, and graphics hardware via OpenGL and Direct3D.
+  copyright:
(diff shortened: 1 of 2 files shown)


libsdl3.yaml
```diff
@@ -0,0 +1,59 @@
...
+  - name: ${{package.name}}-dev
+    dependencies:
+      runtime:
+        - libsdl3
+    pipeline:
+      - uses: split/dev
+
+  - name: ${{package.name}}-doc
+    pipeline:
+      - uses: split/dev
```
ajayk, line 40:
```suggestion
      - uses: split/manpages
```

ajayk (dismissed):

ajayk (approved):

PR #30483, opened 2024-10-10
Spark 3.5 CVE-GHSA-r7pg-v2c8-mfg3 and enabled testing

jamie-albert:
Enabled the tests again after an accidental close of a PR and fixed the way the tests pull in the openjdk-default-jvm. Tests are in line with basic testing from the setup documentation [outlined here.](https://github.com/apache/spark#:~:text=Interactive%20Scala%20Shell) Due to the guava upgrade, I had to also bring along [the missing method](https://github.com/apache/spark/blob/1f24b2d72ed6821a6cc6d1d22683d2f3ba2326a2/pom.xml#L3423) from an upstream commit. The PR I took for [guidance can be found here](https://github.com/apache/spark/commit/1f24b2d72ed6821a6cc6d1d22683d2f3ba2326a2#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R3423). The internal-access patch should be removed when v4.0.0 releases. This missing method integration should have been done when guava was bumped from 14.x to 32.0.1 and would have been caught if the testing was functional. In the future we consider not introducing major test changes and major version dependency changes in the same PR unless we are able to prove the efficacy of the test. Bumped the avro version from 1.11.3 to 1.11.4

### DIFF
--- a/spark-3.5.yaml
+++ b/spark-3.5.yaml
@@ -1,7 +1,7 @@
 package:
   name: spark-3.5
   version: 3.5.3
-  epoch: 2
+  epoch: 3
   description: Unified engine for large-scale data analytics
   copyright:
     - license: Apache-2.0
@@ -42,7 +42,7 @@ pipeline:
 
   - uses: patch
     with:
-      patches: make-distribution.patch
+      patches: make-distribution.patch internal-access.patch
 
   - uses: maven/pombump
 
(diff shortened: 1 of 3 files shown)


powersj (commented):
Looks great, thanks for the background in the commit message. One YAML question and this should be good to go...

edit: once tests work on elastic build as it seems they failed with name lookup?

philroche (approved):

PR #30535, opened 2024-10-10
ensure fluent-operators config can replace

joshrwolf:
this adds a new `fluent-watcher-config` subpackage to `fluent-operator` to replicate the upstream behavior of using a different config file for `fluent-watcher`.

Since `fluent-watcher` depends on `fluent-bit`, both of which provide valid `/fluent-bit/etc/{fluent.conf,parsers.conf,fluent-bit.yaml}`, we use a `replaces` block here to ensure `fluent-watcher` prefers its own config.

### DIFF
--- a/fluent-operator.yaml
+++ b/fluent-operator.yaml
@@ -1,7 +1,7 @@
 package:
   name: fluent-operator
   version: 3.2.0
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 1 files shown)


joshrwolf:
the failing elastic build step is due to the failing test in a known bubblewrap issue (processes not being shared between pipeline steps). its passing in the docker based GHA tests.

developer-guy (approved):

PR #30562, opened 2024-10-10
feat: add openbao-k8s package

wojciechka:


openbao-k8s.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+  - uses: go/bump
+    with:
+      deps: golang.org/x/net@v0.23.0
+      replaces: golang.org/x/net=golang.org/x/net@v0.23.0
+
+  - uses: go/bump
+    with:
+      deps: google.golang.org/protobuf@v1.33.0
+      replaces: google.golang.org/protobuf=google.golang.org/protobuf@v1.33.0
```
kranurag7, line 24: deps and replaces takes more than one args. I think we can combine both of this block.

wojciechka: Good point. Thanks, looked at other examples and made it 1 action and simpler.

kranurag7 (approved):

PR #30569, opened 2024-10-10
portieris: Add package

IdlePhysicist:
This commit adds the portieris package https://github.com/IBM/portieris

#### For new package PRs only

### DIFF
--- a/portieris.yaml
+++ b/portieris.yaml
@@ -1,0 +1,45 @@
+package:
+  name: portieris
+  version: 0.13.19
+  epoch: 0
+  description: A Kubernetes Admission Controller for verifying image trust.
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  environment:
+    CGO_ENABLED: "0"
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/IBM/portieris.git
+      tag: v${{package.version}}
+      expected-commit: d0705039d060ab11b8fa598b79647ac0008af57e
+
+  - uses: go/build
+    with:
+      packages: "./cmd/portieris"
+      ldflags: "-X github.com/IBM/portieris/internal/info.Version=v${{package.version}}"
+      tags: containers_image_openpgp
+      output: portieris
+
+subpackages:
+  - name: portieris-compat
+    description: Compat package for portieris
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/
+          ln -s /usr/bin/portieris "${{targets.subpkgdir}}"/portieris
+
+update:
+  enabled: true
+  github:
+    identifier: IBM/portieris
+    strip-prefix: v
+
+test:
+  pipeline:
+    - runs: |
+        which portieris
+        portieris --help


portieris.yaml
```diff
@@ -0,0 +1,46 @@
...
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/IBM/portieris.git
+      tag: v${{package.version}}
+      expected-commit: d0705039d060ab11b8fa598b79647ac0008af57e
+
+  - runs: |
+      VERSION=v${{package.version}} make portieris
+
```
smoser, line 25: did you try using go/build ?

portieris.yaml
```diff
@@ -0,0 +1,46 @@
...
+  - runs: |
+      mkdir -p "${{targets.destdir}}"/usr/bin/
+      install -m755 ./portieris ${{targets.destdir}}/usr/bin/
+
+data:
+  - name: components
+    items:
+      # Only the keys matter
+      portieris:
```
smoser, line 34: we should probalby standardize on something here.

I'd prefer a 'NOT_USED' value or something rather than a comment. thats due to having been editing several yamls with an program and the comments don't always get handled right.

portieris.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+update:
+  enabled: true
+  github:
+    identifier: IBM/portieris
+    strip-prefix: v
+    tag-filter: v0.13.
```
kranurag7, line 41: tag-filter can be removed here given upstream is using semver naming and no version streams.

portieris.yaml
IdlePhysicist: I did not.  Is there a documented standard for when we should be using go/build vs make?  My only other experience with Wolfi packages is another Go package that uses make for its build.

portieris.yaml
IdlePhysicist: The melange [docs](https://github.com/chainguard-dev/melange/blob/main/docs/BUILD-FILE.md#data) are not clear on what the `data` block does.

Do we even need this block?

IdlePhysicist: With a bit of grepping it looks like a lot of packages have the block populated.

portieris.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - license: Apache-2.0
+
+environment:
+  environment:
+    CGO_ENABLED: "0"
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - go
```
kranurag7, line 16: this can be removed go/build pipeline already includes this. (https://github.com/chainguard-dev/melange/blob/c63354ae90657f4b28e71aaaa9f4543273116a6b/pkg/build/pipelines/go/build.yaml#L3-L7)

Note: If we need `build-base` at build time and then we can only include that in here.

portieris.yaml
```diff
@@ -0,0 +1,46 @@
...
+    with:
+      packages: "./cmd/portieris"
+      ldflags: "-X github.com/IBM/portieris/internal/info.Version=v${{package.version}}"
+      tags: containers_image_openpgp
+      output: portieris
+
+data:
+  - name: components
+    items:
+      portieris: portieris
```
kranurag7, line 35: can be removed, we are no longer using this.

kranurag7 (changes requested):

kranurag7:
great work!! the changes looks good, I think looking at upstream image we will need at compat package for this too because upstream puts the binary at `/` 

ref: https://github.com/IBM/portieris/blob/d0705039d060ab11b8fa598b79647ac0008af57e/Dockerfile#L69

The new tags are not pushed on DockerHub (https://hub.docker.com/r/ibmcom/portieris/tags) and I'm not able to pull because of auth issues from the icr.io registry but older images also have binaries at `/`

portieris.yaml
IdlePhysicist: Works with go/build

IdlePhysicist:
> great work!! the changes looks good, I think looking at upstream image we will need at compat package for this too because upstream puts the binary at `/`
> 
> ref: https://github.com/IBM/portieris/blob/d0705039d060ab11b8fa598b79647ac0008af57e/Dockerfile#L69
> 
> The new tags are not pushed on DockerHub (https://hub.docker.com/r/ibmcom/portieris/tags) and I'm not able to pull because of auth issues from the icr.io registry but older images also have binaries at `/`

Maybe I'm missing something, but if our image sets its CMD & ENTRYPOINT to the correct path e.g. `/bin` then why would we want the compat package?

kranurag7:
> Maybe I'm missing something, but if our image sets its CMD & ENTRYPOINT to the correct path e.g. `/bin` then why would we want the compat package?

We maintain the consistency while placing the binary so that it goes well with upstream image which may be defined at other places. 

for example this about the following being defined in some pod manifest:
```yaml
spec:
  containers:
  - command:
    - /foo
    - bar
    - --config
    - /etc/config/moo.json
```
And if we ship our application with placing the binary at `/usr/bin/foo` then our image is not compatible with upstream one.

portieris.yaml
```diff
@@ -0,0 +1,43 @@
...
+      ldflags: "-X github.com/IBM/portieris/internal/info.Version=v${{package.version}}"
+      tags: containers_image_openpgp
+      output: portieris
+
+subpackages:
+  - name: portieris-compat
+    description: Compat package for portieris
+    pipeline:
+      - runs: |
+          ln -s /usr/bin/portieris "${{targets.subpkgdir}}"/
```
kranurag7, line 32:
```suggestion
          mkdir -p ${{targets.subpkgdir}}/
          ln -s /usr/bin/portieris ${{targets.subpkgdir}}/portieris
```

IdlePhysicist: oh man thanks.  I was thinking "surely / already exists".

IdlePhysicist: running a test locally, then I will push.  Thanks for coming along for the ride

portieris.yaml
smoser: There is not a documented standard.

portieris.yaml
kranurag7: subpackage in this case bundle things from `/home/build/melange-out/portieris-compat` and that doesn't exist out of the box. 
So what above command is doing is following: 

```bash
mkdir -p /home/build/melange-out/portieris-compat/
ln -s /usr/bin/portieris /home/build/melange-out/portieris-compat/portieris
```
You can put `exit 1` at places and run `make debug/foo` for debugging these cases.

kranurag7 (approved):
LGTM, great work on this. Thank you!!

smoser (approved):

PR #30593, opened 2024-10-10
wasmedge: Update to use llvm-19 for build

justinvreeland:


wasmedge.yaml
```diff
@@ -46,10 +47,14 @@ pipeline:
       patches: llvm-include.patch
 
   - runs: |
+      ln -s /usr/lib/llvm-19/lib/LLVMgold.so /usr/lib/ # I cannot convince ld to find this here
```
justinvreeland, line 50: @xnox Can you think of a reason I shouldn't symlink this in out clang package instead?

justinvreeland: Silly question it'll conflict which is why we moved it to begin with.

ajayk (approved):

PR #30609, opened 2024-10-11
checkov: update dependency to python-3.12

davzucky:
The current version of checkov depends on Python 3.11. however, the latest version is Python 3.12 https://pypi.org/project/checkov, and it should be updated

Fixes:
Update the dependency of checkov to be python-3.12

- [x] REQUIRED cgr.dev/chainguard/wolfi-base and ghcr.io/wolfi-dev/sdk images successfully build
- [x] REQUIRED cgr.dev/chainguard/wolfi-base and ghcr.io/wolfi-dev/sdk contain no obsolete (no longer built) packages

### DIFF
--- a/checkov.yaml
+++ b/checkov.yaml
@@ -7,7 +7,7 @@ package:
     - license: MIT
   dependencies:
     runtime:
-      - python-3.11
+      - python-3.12
 
 environment:
   contents:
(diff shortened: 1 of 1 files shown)


kranurag7 (changes requested):
changes in build time dependencies requires epoch bump so in this case you'll require to bump the epoch from `epoch: 0` to `epoch: 1` but I think you already have a PR that bumps that version https://github.com/wolfi-dev/os/pull/30610 

I think we can tackle this there only, we can move to python-3.12 if upstream is building with python-3.12

davzucky:
Closed as merge in another PR in https://github.com/wolfi-dev/os/pull/30869

PR #30615, opened 2024-10-11
druid avro fix for GHSA-r7pg-v2c8-mfg3

jamie-albert:
Resolve what can be fixed for druid/GHSA-r7pg-v2c8-mfg3 there still is a avro dependency as a part of hadoop-client-runtime however we are not able to remedy that. The avro dependency is referred throughout the various druid build components so it is set as a property variable in the parent pom.xml. To bump that env variable we use the maven/pombump properties flag.

### DIFF
--- a/druid.yaml
+++ b/druid.yaml
@@ -1,7 +1,7 @@
 package:
   name: druid
   version: 30.0.1
-  epoch: 1
+  epoch: 2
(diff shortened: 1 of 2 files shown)


kranurag7 (approved):
lgtm given druid main is already using 1.11.4.

PR #30656, opened 2024-10-11
Update clang to provide clang=<version>

justinvreeland:
Also paramterise and remove the clang-dev package which isn't needed anymore.

Fixes: https://github.com/chainguard-dev/internal-dev/issues/1642

### DIFF
--- a/clang-16.yaml
+++ b/clang-16.yaml
@@ -1,7 +1,7 @@
 package:
   name: clang-16
   version: 16.0.6
-  epoch: 3
+  epoch: 4
   description: "C language family frontend for LLVM"
   copyright:
     - license: Apache-2.0
@@ -11,6 +11,8 @@ package:
   dependencies:
     runtime:
       - libLLVM-16
+    provides:
+      - clang=${{package.full-version}}
 
 environment:
   contents:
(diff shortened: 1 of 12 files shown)


justinvreeland:
@xnox This is probably of interest to you

xnox:
Can use elastic build up listed repo, to figure out what the two test failures are.

xnox:
Can use elastic build published repo, to figure out what the two test failures are? As packages are all built there.

justinvreeland:
> Can use elastic build published repo, to figure out what the two test failures are? As packages are all built there.

Ahh this is kind of annoying. I don't know if something changed but it looks like on linux melange/bwrap doesn't mount /sys so stress-ng tests don't work. 

`2024/10/16 11:56:50 INFO stress-ng: error: [3] no CPUs found in /sys/devices/system/cpu name="Test CPU stressor"`

I'm not sure I understand the point of that test but it's there in both snmalloc and scudo-malloc.

xnox:
Thanks! I can poke it. I am on good terms with stressng and its author.

xnox (approved):

PR #30666, opened 2024-10-11
rust-1.81: update to llvm 19

justinvreeland:

### DIFF
--- a/rust-1.81.yaml
+++ b/rust-1.81.yaml
@@ -10,9 +10,12 @@ package:
     memory: 16Gi
   dependencies:
     runtime:
-      - libLLVM-18
+      - libLLVM-${{vars.llvm-vers}}
     provides:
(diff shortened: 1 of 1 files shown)


rust-1.81.yaml
```diff
@@ -10,27 +10,30 @@ package:
     memory: 16Gi
   dependencies:
     runtime:
-      - libLLVM-18
+      - libLLVM-${{vars.llvm-vers}}
     provides:
       - rust=${{package.full-version}}
 
+vars:
+  llvm-vers: 18
```
ajayk, line 18:
```suggestion
  llvm-vers: 19
```

justinvreeland: Ahh yeah that would be important and rust doesn't build with llvm-19 it looks like : (.

justinvreeland:
Doesn't build with llvm-19

PR #30674, opened 2024-10-11
gnucobol: new package

willswire:
Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/gnucobol.yaml
+++ b/gnucobol.yaml
@@ -8,10 +8,6 @@ package:
 
 environment:
   contents:
-    repositories:
-      - https://packages.wolfi.dev/os
-    keyring:
-      - https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
     packages:
       - autoconf
       - automake
--- a/gnucobol.yaml
+++ b/gnucobol.yaml
@@ -15,7 +15,6 @@ environment:
     packages:
       - autoconf
       - automake
-      # - bash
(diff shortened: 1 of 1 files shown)


gnucobol.yaml
```diff
@@ -0,0 +1,46 @@
...
+        --without-db
+
+  - uses: autoconf/make
+
+  - uses: autoconf/make-install
+
+  - uses: strip
+
+update:
+  enabled: false
```
mamccorm, line 41: We can use release-monitor to keep a check for any future updates:
 - https://release-monitoring.org/project/8854/


```suggestion
update:
  enabled: true
  release-monitor:
    identifier: 8854
```

gnucobol.yaml
```diff
@@ -0,0 +1,46 @@
...
+  description: "compiler from COBOL to C"
+  copyright:
+    - license: GPL-3.0-or-later
+
+environment:
+  contents:
+    repositories:
+      - https://packages.wolfi.dev/os
+    keyring:
+      - https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
```
mamccorm, line 14: We can delete these as they are defaults:

```suggestion
```

mamccorm (commented):
Thanks for the contribution, inline suggestions

mamccorm (approved):

PR #30755, opened 2024-10-14
refactor kubernetes-pause and update to 3.10

kranurag7:
this commit bumps the kubernetes pause version to 3.10 which is what
    is being used by kubernetes-1.31 latest release and also adds a
    compat package and tests for kubernetes-pause package.

ref: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-node/updating-pause-images.md#first-level-core-projects (on k/k, it's already updated to 3.10)

the usage of which version of pause is being used can also be verified by kubeadm.
```bash
/ # apk add kubeadm
/ # kubeadm version
kubeadm version: &version.Info{Major:"1", Minor:"31", GitVersion:"v1.31.1", GitCommit:"948afe5ca072329a73c8e79ed5938717a5cb3d21", GitTreeState:"clean", BuildDate:"2024-10-13T19:50:13Z", GoVersion:"go1.23.2", Compiler:"gc", Platform:"linux/amd64"}
/ # kubeadm config images list
registry.k8s.io/kube-apiserver:v1.31.1
registry.k8s.io/kube-controller-manager:v1.31.1
registry.k8s.io/kube-scheduler:v1.31.1
registry.k8s.io/kube-proxy:v1.31.1
registry.k8s.io/coredns/coredns:v1.11.3
registry.k8s.io/pause:3.10
registry.k8s.io/etcd:3.5.15-0
```

```

### CODE REVIEW

File: kubernetes-1.31.yaml [hunk 1]
```diff
@@ -161,17 +161,28 @@ subpackages:
 
           install -d ${{targets.subpkgdir}}/var/log/kube-apiserver
 
-  - name: kubernetes-pause-3.9
+  - name: kubernetes-pause-${{vars.pause-version}}
```
kranurag7, line 164: version can be removed from the name but keeping it because we had maintained the same previously. I'm more inclined to removing it but keeping it open for discussion. This version (3.10) is different from kubernetes version(1.31.x)

kranurag7: same goes with binary name as well, we are naming it `pause-3.10` as of now but I think we can keep it simple and name it pause.

kranurag7: I removed tracking kubernetes pause version on it's own. Now we are tracking and trusting the main k/k repo and the pause is always build from source so we can append kubernetes-version next to `kubernetes-pause`

kubernetes-1.31.yaml
```diff
@@ -161,17 +163,26 @@ subpackages:
 
           install -d ${{targets.subpkgdir}}/var/log/kube-apiserver
 
-  - name: kubernetes-pause-3.9
+  - name: kubernetes-pause-${{vars.kubernetes-version}}
```
EyeCantCU, line 166: Why do we put the pause version in the package name instead of following the kubernetes version? It's built out of the same repository, and we aren't really doing anything too special with it

kubernetes-1.31.yaml
```diff
@@ -161,17 +163,26 @@ subpackages:
...
-      provides:
-        - kubernetes-pause=3.9
     pipeline:
       - working-directory: /home/build/build/pause
         runs: |
           mkdir -p ${{targets.subpkgdir}}/usr/bin/
-          CFLAGS="$CFLAGS -static -DVERSION=${{vars.pause-version}}-${{package.version}}"
-          gcc -o "${{targets.subpkgdir}}"/usr/bin/pause-3.9 linux/pause.c
+          # the executable is always build from source but version is hardcoded and should be updated manually. please check var-transforms.
+          CFLAGS="$CFLAGS -static -DVERSION=v${{vars.pause-version}}-${{package.version}}"
```
EyeCantCU, line 173: We can fetch the pause version dynamically by reading it from within the tree. No reason to statically set it

EyeCantCU (changes requested):

kubernetes-1.31.yaml
EyeCantCU: On top of that, if the actual version drifts, we aren't accounting for that

kubernetes-1.31.yaml
kranurag7: thanks for the review.
we were using the naming because pause comes from k/k repository but follows a different versioning pattern. e.g. all components of k/k gets releases under 1.31.x as of now but pause version is 3.10. It's basically a single pause.c file. 

In this patch, I moved from having `3.9` in the name to appending `-${{vars.kubernetes-version}}` because it was coming from the same repository and given we are always building the pause binary from source we can trust this. As per the docs [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-node/updating-pause-images.md#first-level-core-projects), pause gets updated first in k/k repo and then ecosystem follows so we can always trust this one.

kubernetes-1.31.yaml
kranurag7: In k/k repo, the version is hardcoded within the tree as well (https://github.com/kubernetes/kubernetes/blob/023cd33d23067908d08bd21affa43140fb5d7791/build/pause/Makefile#L20). I can use grep on Makefile but looks like hacky if the underlying name changes. 

> On top of that, if the actual version drifts, we aren't accounting for that

I didn't understand this part completely, we are always building from source so the behavior of the executable will always be same as the one in k/k repo. The hardcoding part is only 3.10 at -DVERSION. `make` target from upstream is not useful because it pulls a 1.5GB container for cross-compilation.

EyeCantCU: To be honest, I think it's ridiculous that something simple like this even has an internal version. The problem with doing this like this is we can not guarantee that 3.10 won't become 3.11. I would consider reading the tag from the Makefile much more reliable here

kubernetes-1.31.yaml
EyeCantCU: It is wrong we put the version in the package name here regardless though. Especially when it's primarily only used to tag the image. We should be getting that with the image module and not relying on an improper package name to decipher it. This will lead to drift when pause updates

Look at how we handle this with EKS Distro. None of those components diverge from the EKS and kubernetes version at all in spite of having different internal versions. We can do terraform magic instead if we need this for the tag but we shouldn't be doing it at the package level

kranurag7: we are now using the name which matches the kubernetes version `v1.31.x` so no more dealing with 3.9 and 3.10 because we are always building it from source and given we are grepping the TAG from Makefile, the version is also set correctly. 

It also matches the convention that we are following with other packages _(kube-proxy-1.31)_. Now, the packages are named something like `kubernetes-pause-1.31`, `kubernetes-pause-1.30` etc.

kubernetes-1.31.yaml
```diff
@@ -161,17 +156,27 @@ subpackages:
...
-          CFLAGS="$CFLAGS -static -DVERSION=${{vars.pause-version}}-${{package.version}}"
-          gcc -o "${{targets.subpkgdir}}"/usr/bin/pause-3.9 linux/pause.c
+          CFLAGS="$CFLAGS -static -DVERSION=v$(grep '^TAG ?=' Makefile | awk '{print $3}')-${{package.version}}"
+          gcc ${CFLAGS} -o "${{targets.subpkgdir}}"/usr/bin/pause linux/pause.c
+    test:
+      pipeline:
+        - name: kubernetes pause version check
+          runs: /usr/bin/pause -V
+
+  - name: kubernetes-pause-compat-${{vars.kubernetes-version}}
```
EyeCantCU, line 174: Where is this used?

kranurag7: Apologies upfront because you're not going to like this answer. :( 

```bash
$ crane config registry.k8s.io/pause:3.10 | jq .config.Entrypoint
[
  "/pause"
]
```

Ahh, just having some fun before I go to bed, following is the detailed answer: 
```bash
$ kind create cluster
$ docker exec -it kind-control-plane bash
root@kind-control-plane:/# ctr --namespace=k8s.io container ls | grep pause
20807e1436af1b80ac01d970f82a34b86f53a7e3ef06d41cb26b5bd5e20d1250    registry.k8s.io/pause:3.10                                     io.containerd.runc.v2
3209afc8f95c83d3c21f385483eaf6f17769f2b51b27c6eb58b0ecb845212479    registry.k8s.io/pause:3.10                                     io.containerd.runc.v2
5dfb4e335ac6fe4747299255791bfa971135715fa3995f37e7c9845f8b1eeb7a    registry.k8s.io/pause:3.10                                     io.containerd.runc.v2
68bcace43965b022f1e4b8f207582ddd35b5f5c14298c8a11c42fa5bbfae7b4e    registry.k8s.io/pause:3.10                                     io.containerd.runc.v2
6bfa094639f1acb6887e4bfe26209b0274986f5fadae72b7ac716748e8d58406    registry.k8s.io/pause:3.10                                     io.containerd.runc.v2
6cf9d697c3c7635d1e28d3e787bd69c57dff2bd07e2f6a337667f519f9553316    registry.k8s.io/pause:3.10                                     io.containerd.runc.v2
e054de8075e5e368e1c3b0e2dc040fd2579ed711955ab69f7a03dca3090d0b99    registry.k8s.io/pause:3.10                                     io.containerd.runc.v2
e7ef02fefec6a2c698ac7adac1c26b4bc0baf46c06d0ec1dc0bc1a5d657a6bc4    registry.k8s.io/pause:3.10                                     io.containerd.runc.v2
f49d08d1ed70f6e5472f581bb016795771a15932dc687b0edb0f22079ea377ff    registry.k8s.io/pause:3.10                                     io.containerd.runc.v2

root@kind-control-plane:/# ctr --namespace=k8s.io container info 20807e1436af1b80ac01d970f82a34b86f53a7e3ef06d41cb26b5bd5e20d1250 | jq .Spec.process.args
[
  "/pause"
]
```
the process (containerd in this case) at runtime calls the binary by absolute path.

EyeCantCU (approved):

PR #30758, opened 2024-10-14
add linkerd CLI package

kranurag7:
this adds linkerd CLI from edge releases and thus includes `edge` as
suffix in the package name.

the naming of the package and file is done this way because this is coming from edge branch which is not stable release from LinkerD. It's the latest release from main and can contain bugs. Maintainers of LinkerD try their best to avoid this but still it's not stable. I am appending -edge is suffix for the same reason.


kranurag7:
elastic build failure is unrelated because it's failing because the test is not using docker runner.

kranurag7:

PR #30767, opened 2024-10-14
py3-argon2-cffi-bindings - fix import error, add test.

smoser:
This module has both a runtime and buildtime dependency on py3-cffi.

The import test is re-added here.  It would fail without the fix and passes with it.

Also drop the duplicate 'provides' (py3-)
 py3-argon2-cffi-bindings == ${{package.name}} == py3-${{vars.pypi-package}}


roaksoax (approved):
lgtm!

PR #30776, opened 2024-10-14
apache-nifi: fix `Could not acquire read lock` on postsubmit builds

rawlingsj:
We are seeing lots of errors in postsubmit builds:

```
java.lang.IllegalStateException: Attempt 1: Could not acquire read lock for 'artifact:org.apache.nifi:nifi-api:1.27.0' in 30 SECONDS
```

By removing parallel threads for arm as well this seems to make the build more reliable.

### DIFF
--- a/apache-nifi.yaml
+++ b/apache-nifi.yaml
@@ -1,7 +1,7 @@
 package:
   name: apache-nifi
   version: 1.27.0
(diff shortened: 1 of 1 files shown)


xnox:
v2 upgrade has now landed

PR #30777, opened 2024-10-14
py3-google-apis-common-protos - multiversion; py3-protobuf minor updates

smoser:
Add multiversion python for py3-googleapis-common-protos. Update py3-protobuf to use use py3-supported-build-base-dev package and the 'import' name to import.


roaksoax (approved):
lgtm!

PR #30782, opened 2024-10-14
Gecko driver 3765

utieyin:
* "geckodriver-0.35.0: new package"
This is a component of https://github.com/orgs/chainguard-dev/projects/45/views/31?pane=issue&itemId=70997726&issue=chainguard-dev%7Cimage-requests%7C3765
Related: docker-selenium #3765

#### For new package PRs only


kranurag7 (commented):
I think we should pull in the source code which is available in the releases https://github.com/mozilla/geckodriver/releases/tag/v0.35.0 and try to build from source. 

cargo install will always pull in the latest which is available on crates.io 

We can fetch the source code, untar and build it from source to verify that we are really packaging 0.35.0

utieyin:
> I think we should pull in the source code which is available in the releases https://github.com/mozilla/geckodriver/releases/tag/v0.35.0 and try to build from source.
> 
> cargo install will always pull in the latest which is available on crates.io
> 
> We can fetch the source code, untar and build it from source to verify that we are really packaging 0.35.0

@kranurag7 

> I think we should pull in the source code which is available in the releases https://github.com/mozilla/geckodriver/releases/tag/v0.35.0 and try to build from source.
> 
> cargo install will always pull in the latest which is available on crates.io
> 
> We can fetch the source code, untar and build it from source to verify that we are really packaging 0.35.0

@kranurag7 , the `build` command actually do this. I have updated the PR accordingly. This allows us to build only the version that is in version control

geckodriver.yaml
```diff
@@ -0,0 +1,42 @@
...
+      repository: https://github.com/mozilla/geckodriver
+      tag: v${{package.version}}
+      expected-commit: 53f2b320033668644eae58a8f3af5c94f0efc0f5
+
+  - name: Configure and Build
+
+  - runs: |
+      mkdir -p ${{targets.destdir}}/usr/bin/
+      cargo build --release
+      mv target/release/geckodriver ${{targets.destdir}}/usr/bin/
```
kranurag7, line 29: this both block can be merged as something like this otherwise the `Configure and Build` is an empty step: 

```suggestion
  - name: Configure and Build 
    runs: |
      mkdir -p ${{targets.destdir}}/usr/bin/
      cargo auditable build --release
      mv target/release/geckodriver ${{targets.destdir}}/usr/bin/
```

geckodriver.yaml
```diff
@@ -0,0 +1,42 @@
...
+  description: A tool for exploring each layer in a docker image
+  copyright:
+    - license: MPL-2.0
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - openssl-dev
```
kranurag7, line 14: let's build it using cargo-auditable as per our best practices. 
```suggestion
      - busybox
      - cargo-auditable
      - openssl-dev
```

geckodriver.yaml
```diff
@@ -0,0 +1,42 @@
...
+    with:
+      repository: https://github.com/mozilla/geckodriver
+      tag: v${{package.version}}
+      expected-commit: 53f2b320033668644eae58a8f3af5c94f0efc0f5
+
+  - name: Configure and Build
+
+  - runs: |
+      mkdir -p ${{targets.destdir}}/usr/bin/
+      cargo build --release
```
EyeCantCU, line 28: We've a `cargo/build` pipeline in melange that will produce an auditable rust binary and includes necessary deps (part of the environment above can be removed): https://github.com/chainguard-dev/melange/tree/main/pkg/build/pipelines/cargo

geckodriver.yaml
```diff
@@ -0,0 +1,42 @@
...
+  github:
+    identifier: mozilla/geckodriver
+    strip-prefix: v
+
+test:
+  pipeline:
+    # AUTOGENERATED
+    - runs: |
+        geckodriver --version
+        geckodriver --help
```
EyeCantCU, line 42: Can we do a selenium test with geckodriver here?

geckodriver.yaml
utieyin: updated

geckodriver.yaml
utieyin: done

geckodriver.yaml
utieyin: done

kranurag7:
we can also append `uses: strip` in here because by default this is not enabled in `cargo/build` pipeline.

EyeCantCU (approved):

PR #30790, opened 2024-10-14
add linkerd2 policy controller package

kranurag7:
adds linkerd2 policy controller package

### DIFF
--- a/linkerd2-policy-controller-edge.yaml
+++ b/linkerd2-policy-controller-edge.yaml
@@ -1,0 +1,59 @@
+package:
+  name: linkerd2-policy-controller-edge
+  version: 24.10.2
+  epoch: 0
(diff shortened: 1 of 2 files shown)


linkerd2-policy-controller-edge.yaml
```diff
@@ -0,0 +1,59 @@
+package:
+  name: linkerd2-policy-controller-edge
```
kranurag7, line 2: the naming of the package and file is done this way because this is coming from edge branch which is not stable release from LinkerD. It's the latest release from main and can contain bugs. Maintainers of LinkerD try their best to avoid this but still it's not stable. I am appending `-edge` is suffix for the same reason.

kranurag7:

PR #30797, opened 2024-10-15
mult-version python py3-matplotlib-inline and py3-ipython

smoser:
py3-matplotlib-inline is cyclic dependency on ipython.
Upstream pypi just resolves that by not recording the dependency in
py3-matplotlib-inline.  So we will follow suite.
    
The test here will verify that pypi's dependencies are met.
It is commented out now to allow this package in so that
ipython can be put in.


roaksoax (approved):
lgtm!

PR #30800, opened 2024-10-15
py3-sqlalchemy - allow updates, multi-version

smoser:
The change here to 'update' section allows picking up updates. The other changes move to multi-version.  2.0.35 supports python 3.13 when the older version did not.


roaksoax (approved):
lgtm!

PR #30810, opened 2024-10-15
add linkerd2 controller package

kranurag7:
adds linkerd2 core controller package. 

`-edge` is suffixed because we are building from edge releases.


kranurag7:

PR #30834, opened 2024-10-15
hugo: new upstream release, and build with extended tag.

roaksoax:
Build hugo with the extended tag, and updates to the new upstream release.


smoser (approved):

kranurag7 (approved):
thanks for this, it's really good that we are building now using extended tag. this will also cover all the cases where one is using wolfi-base CI and `apk add hugo`

PR #30869, opened 2024-10-15
checkov: Update checkov config and add testing

davzucky:
- Remove hard dependencies on python 3.11
- Add dependencies on python 3.11
- Update version of Checkov used to be latest
- Enable auto update
- Add test

#### For new version streams

- [x ] The upstream project actually supports multiple concurrent versions.

#### For version bump PRs

- [0] The `epoch` field is reset to 0


checkov.yaml
```diff
@@ -1,13 +1,13 @@
...
   name: checkov
-  version: 3.0.34
+  version: 3.2.262
   epoch: 0
   description: "static code and composition analysis tool for IaC"
   copyright:
     - license: MIT
   dependencies:
     runtime:
-      - python-3.11
```
jamie-albert, line 10: I disagree with the removal of the version pinning here, it will default to python-3.13 [as seen here](https://github.com/wolfi-dev/os/blob/8de808be6257315d5132458a862423fb53ab2f15/python-3.13.yaml#L12) and the upstream states that [python-3.12 is less stable](https://github.com/bridgecrewio/checkov/pull/6702#issuecomment-2347324389) than python-3.11 and here is a PR mentioning that as of September 15, the [maintainers are "preparing" for python-3.13](https://github.com/bridgecrewio/checkov/pull/6721#:~:text=preparation%20for%20Python%203.13). Due to the noted differences in behavior between python-3.11 and 3.12, bugs that appear far more in depth than what the below testing covers, I would advise against removing the version pinning at this time.

jamie-albert (changes requested):

checkov.yaml
davzucky: @jamie-albert, thank you for sharing all this information. It is useful. However, I need help solving my problem. I want to build a Docker-based image that requires different Python versions for dependencies. The current dependencies are Python applications with different base Python dependencies, not libraries, usually flagged as ```py3-library-name```. 
Should we create a new pipeline for Python in ```melange``` that allows the installation of a Python-based application with the correct Python version segregated from the host version?
[melange python pipeline](https://github.com/chainguard-dev/melange/blob/0900229dc8a4b90e8ad950ffb44e0acee502b996/pkg/build/pipelines/python/README.md).

This would allow us to simulate what a ```pipx``` or ```uvx``` can do.

PR #30881, opened 2024-10-16
py3-keras multi-version

smoser:
Note that there is no test added here.
Import test fails.  Per pypi:

> To use keras, you should also install the backend of choice:
> tensorflow, jax, or torch. Note that tensorflow is required for using
> certain Keras 3 features: certain preprocessing layers as well as
> tf.data pipelines.

The idea is that the user would supply the correct backend and then can configure that via KERAS_BACKEND or config file.

tensorflow is not built multi-versioned yet.
wolfi does not have torch or jax builds.

So we're not really able to even import-test this at the moment.

Also note, while that seems terrible, this is strictly better than what _was_ there, which had no test and lack of _identified_ runtime dependencies.

### DIFF
--- a/py3-keras.yaml
+++ b/py3-keras.yaml
@@ -1,37 +1,77 @@
-# Generated from https://pypi.org/project/keras/
 package:
   name: py3-keras
   version: 3.6.0
-  epoch: 0
+  epoch: 1
   description: Deep learning for humans.
   copyright:
     - license: Apache-2.0
   dependencies:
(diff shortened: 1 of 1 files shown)


smoser:
this depends on https://github.com/wolfi-dev/os/pull/30878 . but after that lands we can just hit 'update branch' and this should build.

roaksoax (approved):
lgtm!

PR #30882, opened 2024-10-16
py3-pymongo - multi-version , py3-hatch-requirements-txt - add supported

smoser:
I failed to add the 'supported' package to py3-hatch-requirements-txt so it wasn't really able  to be used from py3-mongo build.


roaksoax (approved):
lgtm!

PR #30904, opened 2024-10-16
feat: Create initial kubernetes-replicator package

philroche:
This includes remediating all CVEs found after initial build.

Smoke test added to ensure build was successful.

### DIFF
--- a/kubernetes-replicator.yaml
+++ b/kubernetes-replicator.yaml
@@ -17,7 +17,6 @@ pipeline:
     with:
       packages: .
       ldflags: |
-        -s -w
         -X main.version=${{package.version}}
       output: kubernetes-replicator
 
--- a/kubernetes-replicator.yaml
+++ b/kubernetes-replicator.yaml
@@ -16,8 +16,6 @@ pipeline:
   - uses: go/build
     with:
       packages: .
-      ldflags: |
-        -X main.version=${{package.version}}
       output: kubernetes-replicator
 
 update:
--- a/kubernetes-replicator.yaml
+++ b/kubernetes-replicator.yaml
@@ -1,0 +1,38 @@
+package:
+  name: kubernetes-replicator
+  version: 2.10.2
+  epoch: 0
+  description: Kubernetes controller for synchronizing secrets & config maps across namespaces
+  copyright:
+    - license: Apache-2.0
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/mittwald/kubernetes-replicator
+      tag: v${{package.version}}
+      expected-commit: 8f4de965a57b974b446fae30048269894fd4ff3d
+
+  - uses: go/build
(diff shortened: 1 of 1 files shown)


kubernetes-replicator.yaml
```diff
@@ -0,0 +1,38 @@
...
+    with:
+      repository: https://github.com/mittwald/kubernetes-replicator
+      tag: v${{package.version}}
+      expected-commit: 8f4de965a57b974b446fae30048269894fd4ff3d
+
+  - uses: go/build
+    with:
+      packages: .
+      ldflags: |
+        -s -w
```
kranurag7, line 20:
```suggestion

```
`-s` is something we avoid and `-w` comes out of the box from `go/build` pipeline. ref: https://github.com/chainguard-dev/melange/blob/0900229dc8a4b90e8ad950ffb44e0acee502b996/pkg/build/pipelines/go/build.yaml#L58-L62

kubernetes-replicator.yaml
```diff
@@ -0,0 +1,38 @@
...
+        -X main.version=${{package.version}}
+      output: kubernetes-replicator
+
+  - uses: strip
```
kranurag7, line 24:
```suggestion

```
this can be removed as well given we already have `-w` via go/build pipeline.

kubernetes-replicator.yaml
kranurag7: looks like upstream don't set any version related ldflags in their goreleaser config and I was not able to find the other variable as well so we can omit the other ldflag as well.  

https://github.com/mittwald/kubernetes-replicator/blob/master/.goreleaser.yml

philroche: Thank you. Now resolved.

kubernetes-replicator.yaml
philroche: Thank you. Now resolved.

kubernetes-replicator.yaml
```diff
@@ -0,0 +1,35 @@
...
+    with:
+      repository: https://github.com/mittwald/kubernetes-replicator
+      tag: v${{package.version}}
+      expected-commit: 8f4de965a57b974b446fae30048269894fd4ff3d
+
+  - uses: go/build
+    with:
+      packages: .
+      ldflags: |
+        -X main.version=${{package.version}}
```
kranurag7, line 20: I don't see upstream setting this or exposing this variable to set the variable in their main.go at root of the repo so I think we can omit this. 

But this is good to have not a blocker given `go build` with wrong/any ldflags fails silently.

jamonation (dismissed):

kubernetes-replicator.yaml
philroche: All build flags now removed. Thank you

kranurag7 (approved):

PR #30921, opened 2024-10-16
opentelemetry-plugin-nginx-compat: Remove dependency on main package

wlynch:
This allows the compat package to be used with either the full library or nodeps variants (alternatively we can create a `-nodeps-compat` variant as well).

Related: https://github.com/chainguard-dev/image-requests/issues/4407


Aditvil-Dev:
Did you try to test it locally for image tests?

EyeCantCU (approved):
This is correct, and typically how we handle compat packages

PR #30941, opened 2024-10-16
vitess-20: start version stream and reduce package size by splitting binaries to subpkg

Dentrax:
See: https://vitess.io/docs/releases/

* Upstream maintain multiple major versions, so it should be version-streamed.
* Upstream doesn't use all binaries, so get rid of some them to reduce package size
* Use `go/build` pipeline

### DIFF
--- a/vitess-20.yaml
+++ b/vitess-20.yaml
@@ -6,6 +6,8 @@ package:
   copyright:
     - license: Apache-2.0
   dependencies:
+    provides:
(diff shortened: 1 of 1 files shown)


jamonation:
I think this needs some provides for the main and subpackages since it is version streamed now?

jamonation (approved):

developer-guy (approved):

PR #30942, opened 2024-10-16
fix(provides): do not provide default jvm/jdks

dakaneye:


mattmoor (dismissed):

withdrawn-packages.txt
```diff
@@ -10,3 +10,4 @@ modelmesh-runtime-adapter-torchserve-adapter-0.12.0-r1.apk
 modelmesh-runtime-adapter-puller-0.12.0-r1.apk
 modelmesh-runtime-adapter-ovms-adapter-0.12.0-r1.apk
 modelmesh-runtime-adapter-triton-adapter-0.12.0-r1.apk
+openjdk-24-24-r0.apk
```
xnox, line 13: this will not do much, did you mean to withdraw all -r0 packages? all packages that had undesired provides? As in you need to list them all i.e. openjdk-24-default-jvm-24-r0.apk and so on.

Also we should look into adding provider-priorities and add back provides, but once all version streams have priority set on them.

dakaneye: ah ok didn't realize it needed all subpackages as well

xnox: it is individual apk withdrawals from index, there is no magic to it, nor does it have ability to reference old melange files to figure out what was or wasn't build by old melange yaml.

xnox (approved):

PR #30995, opened 2024-10-17
Updated commons-io dependency in required opensearch-2 subpackages: GHSA-78wr-2p64-hpwj

jamie-albert:
This is a more complicated package build than what is regularly seen, opensearch-2 has many plugins that exist in both its own repository and outside of it. For example one of the commons-io vulnerabilities is brought in via [performance-analyzer-rca](https://github.com/opensearch-project/performance-analyzer-rca) however performance-analyzer-rca is built as part of the [performance-analyzer
](https://github.com/opensearch-project/performance-analyzer) (Different repo) build process and requires injecting a patch into the build.gradle that references another patch.

### DIFF
--- a/opensearch-2.yaml
+++ b/opensearch-2.yaml
@@ -5,7 +5,7 @@
 package:
   name: opensearch-2
   version: 2.17.1
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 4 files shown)


philroche (approved):
neat

PR #31030, opened 2024-10-17
pdfcpu: new package

hbh7:
Gotenberg now wants pdfcpu as a dependency with their latest version update. This adds the package so the following PR can be resolved. 

Related: https://github.com/chainguard-dev/enterprise-packages/pull/8477

#### For new package PRs only

- [ ] This PR links to the upstream project's support policy (pdfcpu is NOT present on `endoflife.date`)

### DIFF
--- a/pdfcpu.yaml
+++ b/pdfcpu.yaml
@@ -7,9 +7,19 @@ package:
     - license: Apache-2.0
 
 pipeline:
-  - uses: go/install
+  - uses: git-checkout
     with:
-      package: github.com/pdfcpu/pdfcpu/cmd/pdfcpu@v${{package.version}}
+      repository: https://github.com/pdfcpu/pdfcpu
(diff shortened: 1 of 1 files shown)


pdfcpu.yaml
```diff
@@ -0,0 +1,24 @@
+package:
+  name: pdfcpu
+  version: 0.8.1
+  epoch: 0
+  description: A Go PDF processor and CLI
+  copyright:
+    - license: Apache-2.0
+
+pipeline:
+  - uses: go/install
```
ajayk, line 10: can we switch to go/build instead and ad the necessary build instructions here 
```suggestion
  - uses: go/build
```

hbh7: Sure, I'll swap it shortly.

ajayk (approved):

PR #31162, opened 2024-10-18
vitess: withdrawn old versions

Dentrax:
Now its renamed to `vitess-20.0`, old packages creates non-exist symlinks under `/vt/bin/*` when installing with `apk add vitess vitess-compat`, since it fetches `20.0.2-r5` instead of `(20.0.2-r1)`.


kranurag7 (approved):
looks good to me, we already have [provides in vitess-20.0](https://github.com/wolfi-dev/os/blob/main/vitess-20.0.yaml) so removing this would work when adding `apk add vitess vitess-compat` 
 

thanks!!

PR #31194, opened 2024-10-18
openssf-compiler-options: add -Wno-hardened to avoid causing problems…

justinvreeland:
… with redefined FORTIFY_SOURCE

Remove lines adding this manually from ruby.

### DIFF
--- a/openssf-compiler-options.yaml
+++ b/openssf-compiler-options.yaml
@@ -1,7 +1,7 @@
 package:
   name: openssf-compiler-options
   version: 20240627
-  epoch: 3
+  epoch: 4
   description: "Compiler Options Hardening Guide for C and C++"
   url: https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.html
   copyright:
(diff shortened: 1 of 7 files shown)


justinvreeland:
@xnox Thoughts? This is a persistent issue I find and some things like python packages (in particular py3-ml-metadata) seem to go out of their way to ignore options passed other ways.

xnox:
> @xnox Thoughts? This is a persistent issue I find and some things like python packages (in particular py3-ml-metadata) seem to go out of their way to ignore options passed other ways.

Yes please, as that's what Wno-error=hardened was supposed to do.

Unless I got the syntax wrong.

Plus many of them are false alarms too. Not adding blah because it was already specified to the same safe value because us or upstream were already doing best practice.

justinvreeland:
> Yes please, as that's what Wno-error=hardened was supposed to do.
Yeah judging from where I see this problem that turns off the error but it still keeps it as a warning then if they have `-Werr` the warning will cause an error : (. There's also a fair amount of failure from additional configure output which need to be entirely suppressed to work correctly.

gcc-13.yaml
```diff
@@ -28,6 +28,7 @@ environment:
       - make
       - mpc-dev
       - mpfr-dev
+      - openssf-compiler-options
       - patch
       - texinfo
```
xnox, line 33: Is this hunk intentional? Why no epoch bump?

xnox (dismissed):

gcc-13.yaml
justinvreeland: This is intentional.  I initially had more extensive gcc-13 changes but merged it into here to build successfully. I'll bump it.

justinvreeland: Actually I'll remove it and fix it in the gcc pr I already have.

xnox (approved):

PR #31206, opened 2024-10-18
mountpoint-s3 package build

developer-guy:
Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/mountpoint-s3.yaml
+++ b/mountpoint-s3.yaml
@@ -1,0 +1,47 @@
+package:
+  name: mountpoint-s3
+  version: 1.10.0
+  epoch: 0
+  description: A simple, high-throughput file client for mounting an Amazon S3 bucket as a local file system.
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - clang-19
+      - cmake
+      - fuse3
+      - fuse3-dev
+      - libclang-cpp-19
+      - pkgconf
+      - rust
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/awslabs/mountpoint-s3
+      tag: v${{package.version}}
+      expected-commit: 2b0161600400c7ab7ccb1d6811abb9774facef5d
+      recurse-submodules: true
+
+  - runs: |
+      mkdir -p "${{targets.contextdir}}"/usr/bin
+      cargo build --release
+      install -Dm755 target/release/mount-s3 "${{targets.contextdir}}"/usr/bin/mount-s3
+
+  - uses: strip
+
+test:
+  pipeline:
+    - runs: |
+        mount-s3 --version
+
+update:
+  enabled: true
+  github:
+    identifier: awslabs/mountpoint-s3
+    strip-prefix: mountpoint-s3-


Dentrax (dismissed):

mountpoint-s3.yaml
```diff
@@ -0,0 +1,47 @@
...
+      mkdir -p "${{targets.contextdir}}"/usr/bin
+      cargo build --release
+      install -Dm755 target/release/mount-s3 "${{targets.contextdir}}"/usr/bin/mount-s3
+
+  - uses: strip
+
+test:
+  pipeline:
+    - runs: |
+        mount-s3 --version
```
jamonation, line 41: Would there be a way to use minio here to create a bucket and mount it? That seems like a solid way to test things

developer-guy: I tried but their client is not suitable with any other s3 compatible sdks, I was tried to use it with minio but it failed with an error like:

```bash
[sdk] ❯ mount-s3 --region us-east-2 --endpoint-url http://localhost:9000 my-test-bucket /tmp
Error: Failed to create S3 client

Caused by:
    0: initial ListObjectsV2 failed for bucket my-test-bucket in region us-east-2
    1: Client error
    2: Unknown CRT error
    3: CRT error 1059: aws-c-io: AWS_IO_DNS_INVALID_NAME, Host name was invalid for dns resolution.
Error: Failed to create mount process
```

jamonation: One last try with the IP address of minio instead of the DNS name? e.g. I see this in an unprivileged container:
```
eee5ce3e6889:~# mount-s3 --region=us-east-1 --endpoint-url http://172.17.0.3:9000 myminio /tmp/minio
fuse: device not found, try 'modprobe fuse' first
Error: Failed to create FUSE session
```

developer-guy: oh cmon! it worked, wow!

```
[sdk] ❯ mount-s3 --region=us-east-1 --endpoint-url=http://172.17.0.2:9000 myminio /tmp/minio
bucket myminio is mounted at /tmp/minio
[sdk] ❯
```

developer-guy: but it didn't work in the tests due to that: https://github.com/s3fs-fuse/s3fs-fuse/issues/647#issuecomment-330398877.

I think it needs privileged container so I'm afraid that we cannot add more tests :/

Screenshot 2024-10-19 at 16 06 02

mountpoint-s3.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+update:
+  enabled: true
+  github:
+    identifier: awslabs/mountpoint-s3
+    strip-prefix: mountpoint-s3-
```
kranurag7, line 47: the releases contains a space [mountpoint-s3 v1.10.0](https://github.com/awslabs/mountpoint-s3/releases/tag/mountpoint-s3-1.10.0) in there wondering if we will be able to get update on this here if we use `-` 

looking at API output from GitHub, If we are filtering based on tag_name then this looks good but if based on name then maybe we can use tags. 

`gh api /repos/awslabs/mountpoint-s3/releases` 
```json
    "node_id": "RE_kwDOH3Saf84KuwPA",
    "tag_name": "mountpoint-s3-1.10.0",
    "target_commitish": "main",
    "name": "mountpoint-s3 v1.10.0",
```

kranurag7 (dismissed):

kranurag7 (approved):

PR #31207, opened 2024-10-18
set file capabilities on some iptables executables

kranurag7:
these file capabilities are required by linkerd-proxy-init at runtime to run as non-root user with UID of 65532.

### DIFF
--- a/iptables.yaml
+++ b/iptables.yaml
@@ -72,7 +72,7 @@ pipeline:
 
 subpackages:
   - name: iptables-cap_set_admin
-    description: iptables with extra file capabilities on executables in /sbin only to be used with linkerd.
+    description: iptables with cap_net_raw and cap_net_admin file capabilities on executables in /sbin
     options:
       no-provides: true
     pipeline:
--- a/iptables.yaml
+++ b/iptables.yaml
@@ -77,9 +77,7 @@ subpackages:
       no-provides: true
     pipeline:
       - runs: |
-          mkdir -p ${{targets.contextdir}}/sbin
-          cp ${{targets.destdir}}/sbin/xtables-legacy-multi ${{targets.contextdir}}/sbin/xtables-legacy-multi
-          cp ${{targets.destdir}}/sbin/xtables-nft-multi ${{targets.contextdir}}/sbin/xtables-nft-multi
+          cp -r ${{targets.destdir}} ${{targets.contextdir}}
           setcap cap_net_raw,cap_net_admin+eip ${{targets.contextdir}}/sbin/xtables-legacy-multi
           setcap cap_net_raw,cap_net_admin+eip ${{targets.contextdir}}/sbin/xtables-nft-multi
 
--- a/iptables.yaml
+++ b/iptables.yaml
@@ -1,7 +1,7 @@
 package:
   name: iptables
   version: 1.8.10
-  epoch: 4
+  epoch: 5
   description: Linux kernel firewall, NAT and packet mangling tools
   copyright:
     - license: GPL-2.0-or-later
@@ -17,6 +17,7 @@ environment:
       - busybox
       - ca-certificates-bundle
       - flex
+      - libcap-utils
       - libmnl-dev
       - libnftnl-dev
       - libtool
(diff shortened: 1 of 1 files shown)


iptables.yaml
```diff
@@ -70,6 +71,18 @@ pipeline:
       install -D -m644 ebtables.confd "${{targets.destdir}}"/etc/conf.d/ebtables
 
 subpackages:
+  - name: iptables-privileged
```
kranurag7, line 74: the patch is in draft mode because we can discuss more on the package name here. I was thinking if we can name it as `iptables-linkerd-compat` similar to our bitnami variant of package where we name a package as `*-bitnami-compat`

kranurag7:
package update config check issue looks like one off and elastic build logs looks good from a quick skim.

xnox (commented):
I am not sure if it has to be a separate package or not.

If container is in a username space those capabilities will not grant anything.

If docker / containerd / k8s / bubble wrap / flat pack / nspawn gave the namespace network capability things will just work.

I will check what other things use, but it is safer to grant capability on select binaries, rather than all binaries in a container / whole container. Cause I would have hoped we can get away with creating duplicate package because of permission settings like this.

iptables.yaml
```diff
@@ -70,6 +71,18 @@ pipeline:
...
+    description: iptables with extra file capabilities on executables in /sbin only to be used with linkerd.
+    options:
+      no-provides: true
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/sbin
+          cp ${{targets.destdir}}/sbin/xtables-legacy-multi ${{targets.contextdir}}/sbin/xtables-legacy-multi
+          cp ${{targets.destdir}}/sbin/xtables-nft-multi ${{targets.contextdir}}/sbin/xtables-nft-multi
+          setcap cap_net_raw,cap_net_admin+eip ${{targets.contextdir}}/sbin/xtables-legacy-multi
+          setcap cap_net_raw,cap_net_admin+eip ${{targets.contextdir}}/sbin/xtables-nft-multi
```
EyeCantCU, line 84: We likely need more than just xtables here so we'd probably need to do something like this:

```
cp -r ${{targets.destdir}} ${{targets.contextdir}}
setcap cap_net_raw,cap_net_admin+eip ${{targets.contextdir}}/sbin/xtables-legacy-multi
setcap cap_net_raw,cap_net_admin+eip ${{targets.contextdir}}/sbin/xtables-nft-multi
```

iptables.yaml
jamonation: I'd call it what it is - something like `iptables-cap_set_admin`. Then it can be used for any package or image that needs it, not just linkerd

iptables.yaml
```diff
@@ -70,6 +71,16 @@ pipeline:
       install -D -m644 ebtables.confd "${{targets.destdir}}"/etc/conf.d/ebtables
 
 subpackages:
+  - name: iptables-cap_set_admin
+    description: iptables with extra file capabilities on executables in /sbin only to be used with linkerd.
```
jamonation, line 75:
```suggestion
    description: iptables with cap_net_raw and cap_net_admin file capabilities on executables in /sbin
```

iptables.yaml
```diff
@@ -70,6 +71,16 @@ pipeline:
       install -D -m644 ebtables.confd "${{targets.destdir}}"/etc/conf.d/ebtables
 
 subpackages:
+  - name: iptables-cap_set_admin
```
EyeCantCU, line 74: What about `iptables-xtables-privileged`? Since we're only setting caps for xtables

iptables.yaml
```diff
@@ -70,6 +71,16 @@ pipeline:
...
+    description: iptables with cap_net_raw and cap_net_admin file capabilities on executables in /sbin
```
EyeCantCU, line 75:
```suggestion
    description: iptables with cap_net_raw and cap_net_admin capabilities set for xtables
```

EyeCantCU (changes requested):

EyeCantCU (approved):

PR #31210, opened 2024-10-18
libreoffice-24.2 - use boost lib bundled by libreoffice.

smoser:
We saw bug using convert with the system boost (currently 1.86.0). Switching to using libreoffice's bundled version fixed.

The problem was as easy to recreate as:

    $ echo "1,scott,taco" > my.csv
    $ /usr/lib/libreoffice/program/soffice.bin --convert-to pdf my.csv
    Unspecified Application Error
    $ echo $?
    1

After compiling with `--enable-dbgutil`, then we got some more information out that pointed me in this direction.

    warn:legacy.tools:249:249:sfx2/source/appl/appuno.cxx:452:
      invalid type for Stream
    warn:sal.osl:249:249:sal/osl/unx/module.cxx:100:
      dlopen(/binds/libreoffice/instdir/program/libscfiltlo.so, 1):
       /binds/libreoffice/instdir/program/liborcus-0.18.so.0:
         undefined symbol: _ZN5boost9iostreams4zlib10stream_endE
    warn:sal.osl:249:249:sal/osl/unx/module.cxx:100:
      dlopen(libscfiltlo.so, 1):
       /binds/libreoffice/instdir/program/liborcus-0.18.so.0:
        undefined symbol: _ZN5boost9iostreams4zlib10stream_endE
    soffice.bin: /binds/libreoffice/sc/source/ui/docshell/impex.cxx:2668:
       ScFormatFilter::Get()::<lambda()>: Assertion `false' failed.

### DIFF
--- a/libreoffice-24.2.yaml
+++ b/libreoffice-24.2.yaml
@@ -1,7 +1,7 @@
 package:
   name: libreoffice-24.2
   version: 24.2.7.1
-  epoch: 1
+  epoch: 2
   description:
   # https://www.libreoffice.org/about-us/licenses
   copyright:
@@ -14,7 +14,6 @@ package:
   dependencies:
     runtime:
       - argon2
-      - boost
       - cairo
       - cups-libs
       - dbus-libs
@@ -47,7 +46,6 @@ environment:
       - automake
       - bash
       - bison
-      - boost-dev
       - build-base
       - busybox
       - cairo-dev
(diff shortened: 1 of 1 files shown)


xnox (approved):

xnox:
I wonder if it is worth reporting upstream, and if they care or not. At least they could block building against a bad version of system boost.

PR #31259, opened 2024-10-20
kafka_exporter: add compat and expand tests

Dentrax:
Adds `-compat` for `kafka_exporter`: https://github.com/danielqsj/kafka_exporter/blob/v1.8.0/Dockerfile


kranurag7 (approved):
lgtm, checked it once again, build is passing is elastic and errors on tests side are because of different runners.

PR #31306, opened 2024-10-21
feat: add systemd units for docker, containerd, sshd, logind and systemd stubs configs for network and logind

89luca89:
Useful in case of systemd containers, we want actual services to work
We also add some of the manual config into split packages in order to create these containers more granular and reproducible


89luca89:
Contextually bumping containerd to 1.7.23 to please the CI

dustinkirkland (dismissed):

xnox (changes requested):
why hand crafted services used instead of upstream ones?
why none of the hardening options applied in the service units?

89luca89:
> why hand crafted services used instead of upstream ones? why none of the hardening options applied in the service units?

mostly for user/groups, It's not clear to me how I should "adduser" from apks, given that apko will not run pre/post scriptlets

dustinkirkland:
I'm not clear on this either (adduser in packages).  Would love some help understanding this too!

89luca89:
I suppose we should keep track of it "manually" and add them in the apko manifests when/if needed
But this makes the image creation a bit more involved than what it should

Maybe a split package (dockerd-user, for example) can be useful, but still I don't know how that should work then

89luca89:
> > > why hand crafted services used instead of upstream ones? why none of the hardening options applied in the service units?
> > 
> > 
> > mostly for user/groups, It's not clear to me how I should "adduser" from apks, given that apko will not run pre/post scriptlets
> 
> apko has configuration to define users at image build time; you will define stock os users there. Please do not pollute those in packaging. That should help for many of these things. More-over there is a set of stock users we always provide already anyway, and we can expand that static list too - note expanding default static list means we should never reuse the same user or group ids from the reserved range again.

That's ok will make them at apko level
just wanted to be sure I'm not missing something :+1:

89luca89:
Corrected to use upstream services, with just a patch to fix PATHs where needed

89luca89:
@dustinkirkland @xnox updated with suggestions (and bumped epochs) can you please take a look? Thanks!

89luca89:
@xnox @dustinkirkland friendly ping :heart:

systemd.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: systemd
   version: "256.7"
-  epoch: 4
+  epoch: 5
```
jamonation, line 4:
```suggestion
  epoch: 0
```

systemd.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: systemd
   version: "256.7"
```
jamonation, line 3:
```suggestion
  version: "256.8"
```

They just released an hour ago, expected-commit is a2240d1cf9f3515728186c2c98a6f4b64a40e4da

89luca89:
Thanks @jamonation :heart: 

The CI is blocking containerd as there is an update 1.73.0->2.0.0 seems a bump big enough to deserve it's own PR

jamonation (approved):
@89luca89 ah yes I see that now too. Well +1 for these changes in any case

xnox (approved):
Some of these can be improved and get cleaner. The custom setting of path looks like bugs elsewhere in systemd config.

xnox:
Also setting approved to run label for elastic-build result. Or please make your wolfi-dev membership public.

PR #31310, opened 2024-10-21
py3-fastbencode - correctly build and test the extension.

smoser:
Previously, fastbencode use would warn:

    $ apk info -v | grep fastb
    py3.12-fastbencode-0.3.1-r1
    $ python3.12 -c 'import fastbencode'
    /usr/lib/python3.12/site-packages/fastbencode/__init__.py:51:
       UserWarning: failed to load compiled extension:
         No module named 'fastbencode._bencode_pyx'

We are intentionally building the extension so we want that warning
to be fatal.

Directly importing the pyx 'import fastbencode._bencode_pyx' will
instead fail.


smoser:
https://github.com/chainguard-dev/melange/pull/1579 is a fix for the failures here.

xnox (approved):

PR #31325, opened 2024-10-21
delete disabled workflows

imjasonh:
These workflows have been disabled, in some cases for many months: https://github.com/wolfi-dev/os/actions

To avoid confusion, we should just delete these.

I've kept `build.yaml` for now, which is also disabled, in case we need to migrate back from elastic postsubmit builds


ajayk (approved):

xnox:
Thank you!

PR #31612, opened 2024-10-22
Transition from eudev to systemd

xnox:
eudev is musl port of systemd project udev component. As Wolfi is
glibc based and can support systemd, there is no need to build or use
eudev. Start transition from eudev to systemd, with goal of removing eudev.

Note, currently dependencies on so:udev.so can install eudev, even when built against systemd-dev.

- **FTBFS: hwloc: remove sphinx dependencies, which appear to not be needed to build manpages**
Note current set of build dependencies is not co-installable as sphinx-7 & sphinx-8 are pulled in at the same time.

- **FTBFS: systemd: build-depend on libbpf-dev for C headers, rather than runtime libs only**

- **Skip transition for multipath-tools & thingsboard**
    thingsboard fails to download node tarball in elastic builds.
    multipath-tools fails to test in GHA builds.

Remaining packages will be fixed, after this initial batch of majority of fixes lands.

### DIFF
--- a/bluez.yaml
+++ b/bluez.yaml
@@ -2,7 +2,7 @@
 package:
   name: bluez
   version: "5.79"
-  epoch: 0
+  epoch: 1
   description: Tools for the Bluetooth protocol stack
   copyright:
     - license: GPL-2.0-or-later AND BSD-2-Clause AND MIT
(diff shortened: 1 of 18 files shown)


mattmoor (approved):
There's a CVE finding in the py-udev package

PR #31658, opened 2024-10-23
py3-soupsieve - multi-version

smoser:
Note:
 soupseive has circular dependency with beautifulsoup
 so there is no import test here as it would not run
 without running a runtime dep on py3-beautifulsoup.

See https://github.com/wolfi-dev/os/pull/30668
and https://github.com/wolfi-dev/os/pull/30667 for example of attempts to recognize the dependency.

### DIFF
--- a/py3-soupsieve.yaml
+++ b/py3-soupsieve.yaml
@@ -1,27 +1,33 @@
-# Generated from https://pypi.org/project/soupsieve/
+# Note: soupseive has circular dependency with beautifulsoup
+#       so there is no import test here as it would not run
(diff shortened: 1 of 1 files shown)


justinvreeland (approved):

py3-soupsieve.yaml
```diff
@@ -1,27 +1,33 @@
-# Generated from https://pypi.org/project/soupsieve/
+# Note: soupseive has circular dependency with beautifulsoup
```
justinvreeland, line 1: 👍🏻

PR #31680, opened 2024-10-23
fix logstash bitnami

cpanato:
follow up of https://github.com/wolfi-dev/os/pull/31650

xref: https://github.com/chainguard-dev/image-requests/issues/4094


jamonation (approved):
LGTM, not sure why the elastic build version wouldn't be running with this change

PR #31729, opened 2024-10-23
tclx & tcl-tls: Remove tcl-tls and tclx

justinvreeland:
Neither are used in any packages or images nor are they included in the image
they were originally added for.


justinvreeland:
Tcl 9 issue not an issue with out options.

xnox (approved):
Confirmed as unused in all package repositories and images.

PR #31771, opened 2024-10-24
opentofu: add version stream support

ajayk:


cpanato (approved):

cpanato:
need to update the image to use the new version stream
@ajayk

PR #31831, opened 2024-10-24
Change permissions within /opt/kserve/mmesh

srishtih:
- Creates /etc/environment  expected at runtime
- Creates build-version at /opt/kserve/mmesh expected at runtime

### DIFF
--- a/kserve-modelmesh.yaml
+++ b/kserve-modelmesh.yaml
@@ -1,7 +1,7 @@
 package:
   name: kserve-modelmesh
   version: 0.12.0
-  epoch: 2
+  epoch: 3
   description: The ModelMesh framework is a mature, general-purpose model serving management/routing layer designed for high-scale, high-density and frequently-changing model use cases.
   dependencies:
     runtime:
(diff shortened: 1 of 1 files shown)


kserve-modelmesh.yaml
```diff
@@ -30,23 +30,28 @@ pipeline:
...
       mkdir -p ${{targets.destdir}}/opt/kserve/mmesh
-      mv /home/build/target/dockerhome ${{targets.destdir}}/opt/kserve/mmesh/
-      chmod -R 775 ${{targets.destdir}}/opt/kserve/mmesh/dockerhome
+      mv /home/build/target/dockerhome/* ${{targets.destdir}}/opt/kserve/mmesh/
+
+      echo "$(date +%Y%m%d)-$(git rev-parse --short HEAD)" > ${{targets.destdir}}/opt/kserve/mmesh/build-version
+      mkdir -p ${{targets.destdir}}/etc
+      mkdir -p ${{targets.destdir}}/opt/kserve/mmesh/log
+      chmod -R 775 ${{targets.destdir}}/opt/kserve/mmesh/
+      touch ${{targets.destdir}}/etc/environment
```
kranurag7, line 39: can we pull in `linux-pam` as runtime deps instead of creating this file. 

$ apk info -W /etc/environment
/etc/environment is owned by linux-pam-1.6.1-r1

kserve-modelmesh.yaml
```diff
@@ -30,23 +31,27 @@ pipeline:
       export JAVA_HOME=/usr/lib/jvm/java-17-openjdk
       mvn -B package -Dfile.encoding=UTF8 -DskipTests=true --file pom.xml
       mkdir -p ${{targets.destdir}}/opt/kserve/mmesh
-      mv /home/build/target/dockerhome ${{targets.destdir}}/opt/kserve/mmesh/
-      chmod -R 775 ${{targets.destdir}}/opt/kserve/mmesh/dockerhome
+      mv /home/build/target/dockerhome/* ${{targets.destdir}}/opt/kserve/mmesh/
```
EyeCantCU, line 34: If this works independently, it might be worth splitting into a separate package - something like `kserve-modelmesh-docker` with `bash` removed from runtime deps above and added to the subpackage?

kserve-modelmesh.yaml
```diff
@@ -1,11 +1,12 @@
 package:
   name: kserve-modelmesh
   version: 0.12.0
-  epoch: 2
+  epoch: 3
   description: The ModelMesh framework is a mature, general-purpose model serving management/routing layer designed for high-scale, high-density and frequently-changing model use cases.
   dependencies:
     runtime:
       - bash # entrypoint script uses bash
+      - linux-pam
```
kranurag7, line 9: from a quick skim, it looked like the package needed linux-pam at runtime but looking more deeply into it that's not the case.

upstream is using ubi-minimal as base which comes with `/etc/environment` out of the box. This comes from `setup` package which configures the system settings at start.

```bash
$ docker run --rm -it registry.access.redhat.com/ubi8/ubi-minimal:latest
[root@0ec0def7e2db /]# rpm -qf /etc/environment
setup-2.12.2-9.el8.noarch
```

So, I think upstream `/etc/environment` is different considering it comes from `setup` than our so I'll be more comfortable creating `/etc/environment` using touch than shipping this package with linux-pam because linux-pam comes with more shared-libraries and auth related binaries.

I think the original changes were good enough for creating `/etc/environment`

TLDR is: we cannot pull in `/etc/environment` for this package because linux-pam is not a runtime dependency for this package and there's variation b/w who owns `/etc/environment` in our base image vs upstream base image. 

Sorry but for now it looks like a revert of https://github.com/wolfi-dev/os/pull/31831/commits/1f2788a87da76f9b0f1dc9876015bb67680a4b1c will be a better approach here.

kserve-modelmesh.yaml
srishtih: I'm not sure I fully follow why we need the subpackage

kranurag7: > If this works independently, it might be worth splitting into a separate package

I think this looks like something that's core to this package so we can keep it here probably. It's a group of jars and few scripts and I'm not sure if we can place it at some other location. 

Screenshot 2024-10-25 at 1 17 24 AM

please correct if that's not the case.

EyeCantCU: We don't need one. The idea was to keep Docker bits separate from the framework so that it could be installed independently without adding things that would only be useful in Docker

EyeCantCU (approved):

PR #31856, opened 2024-10-24
abseil-cpp.yaml: Add openssf-options

justinvreeland:


tuananh:
@justinvreeland when we do this, are the env var defined at https://github.com/wolfi-dev/os/blob/main/build-x86_64.env be used or config file in openssf package?

justinvreeland:
@tuananh The options provided by the openssf-compiler-options package will added to the command.  The exported environment variables will also still be in the environment for tooling to consider if it uses them. It is not unlikely that both options will have an impact on the compilation.

xnox:
> @justinvreeland when we do this, are the env var defined at https://github.com/wolfi-dev/os/blob/main/build-x86_64.env be used or config file in openssf package?

the environment variables will be removed once openssf-compiler-options becomes the default and pulled in by all compilers.
unfortunately environment variables are used by minority of build systems and do not end up taking any effect on many binaries.

justinvreeland:
Superseded by: https://github.com/wolfi-dev/os/pull/34201

PR #31896, opened 2024-10-25
add necessary missing files/folders to mountpoint-s3

developer-guy:
To provide compatibility between these lines in Dockerfile:

- https://github.com/awslabs/mountpoint-s3-csi-driver/blob/main/Dockerfile#L59-L62

### DIFF
--- a/mountpoint-s3.yaml
+++ b/mountpoint-s3.yaml
@@ -1,7 +1,7 @@
 package:
   name: mountpoint-s3
   version: 1.10.0
-  epoch: 1
+  epoch: 2
   description: A simple, high-throughput file client for mounting an Amazon S3 bucket as a local file system.
   copyright:
     - license: Apache-2.0
@@ -14,9 +14,9 @@ environment:
       - ca-certificates-bundle
       - clang-19
       - cmake
-      - fuse3
-      - fuse3-dev
+      - fuse2-dev
       - libclang-cpp-19
+      - libgcc
       - pkgconf
       - rust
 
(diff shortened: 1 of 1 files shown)


mountpoint-s3.yaml
```diff
@@ -14,9 +14,10 @@ environment:
       - ca-certificates-bundle
       - clang-19
       - cmake
-      - fuse3
-      - fuse3-dev
+      - fuse2
+      - fuse2-dev
```
kranurag7, line 18: fuse2-dev will pull in fuse2, we can omit the first one. 

```bash
/ # apk add fuse2-dev
(1/3) Installing fuse-common (3.16.2-r3)
(2/3) Installing fuse2 (2.9.9-r0)
(3/3) Installing fuse2-dev (2.9.9-r0)
OK: 15 MiB in 18 packages
```
also there's `fuse2-static` in index. I think that's not required in this case.

mountpoint-s3.yaml
```diff
@@ -36,6 +36,16 @@ pipeline:
...
 
+subpackages:
+  - name: ${{package.name}}-compat
+    description: "Compatibility package to place binaries in the location expected by upstream helm charts"
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/mountpoint-s3/bin
+          ln -sf /usr/bin/mount-s3 ${{targets.contextdir}}/mountpoint-s3/bin/mount-s3
+          cp /usr/lib/libgcc_s.so.1 ${{targets.contextdir}}/mountpoint-s3/bin/libgcc_s.so.1
+          cp /usr/lib/libfuse.so.2 ${{targets.contextdir}}/mountpoint-s3/bin/libfuse.so.2
```
kranurag7, line 47: from my previous reviews with @EyeCantCU we always symlink shared libraries and don't use cp but I'll wait for RJ to confirm here once again if this is something we can proceed with.

developer-guy: yeah its possible though, let me do it real quick.

developer-guy: good point, thanks!

kranurag7 (approved):
thank you very much!! looks good.

PR #31939, opened 2024-10-25
ruby3.2-kubeclient - Loosen restriction on recursive-open-struct version

### LINKED ISSUE #31938: ruby3.2-kubeclient fails to test due to ruby3.2-recursive-open-struct that is too new
When recursive-open-struct moved to version 2.0, the gem requirements from kubeclient are no longer matched.
This causes FTBFS (test failure).

smoser:
ruby3.2-kubeclient is currently not passing a test. That means it can't get any new version should one be created. I hit this when trying to create ruby 3.3 versions of it.

If this does not work, then we will need to create a ruby3.2-recursive-open-struct-1.3 package and depend on it.

I proposed upstream https://github.com/ManageIQ/kubeclient/pull/640.

Fixes: https://github.com/wolfi-dev/os/issues/31938

Note, that  currently https://github.com/aetherknight/recursive-open-struct does not have any maintenance branches.  Each of the versions are just tags on main.

### DIFF
--- a/ruby3.2-kubeclient.yaml
+++ b/ruby3.2-kubeclient.yaml
@@ -1,7 +1,7 @@
 package:
   name: ruby3.2-kubeclient
   version: 4.12.0
-  epoch: 1
+  epoch: 2
   description: An efficient, thread safe LRU cache.
   copyright:
     - license: MIT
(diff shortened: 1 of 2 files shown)


smoser:
This did land as part of the big pr I did.

PR #31941, opened 2024-10-25
apache-nifi.yaml: Add openssf-options

justinvreeland:


xnox (dismissed):

xnox:
v2 upgrade has now landed, so this needs to be re-created.

xnox (approved):

PR #31995, opened 2024-10-25
make clickhouse link to openssl dynamically

kranurag7:
so that we can do fips magic


clickhouse-24.9.yaml
```diff
@@ -76,6 +79,11 @@ subpackages:
     pipeline:
       - uses: split/dev
 
+  - name: "${{package.name}}-debug"
```
kranurag7, line 82: I think we can omit this, looking at build output of ninja it contains `/usr/lib/debug` but that's not coming up in the package. It looks like a split/debug is required but the package is empty which is not desirable. 💭

EyeCantCU: Let's continue removing the directory above or manually move it here

Dentrax (approved):

EyeCantCU (approved):

PR #32004, opened 2024-10-26
python multi-version packages bs4 and others.

smoser:
multi-version:
 - py3-beautiful-soup4
 - py3-hatch-nodejs-version.yaml
 - py3-influxdb-client.yaml
 - py3-jaeger-client.yaml
 - py3-pip-tools.yaml
 - py3-reactivex.yaml - added as dep for py3-influxdb-client
 - py3-threadloop.yaml - added as dep for jaeger-client
 - thrift.yaml

pip-tools is now a "regular" python package, before it was a fat pip install
which had bins for pyproject-build and wheel that were its dependencies.


roaksoax (approved):
lgtm!

PR #32068, opened 2024-10-28
fix(CI): do not fail CVE scan and soname check if only build for one arch

debasishbsws:
If a package is only build for one arch(x86_64), the CVE scan CI fail because of the artifact is missing from any of one arch. With these change the Ci will not fail if ony one artifact is present it will only fail if no artifact is present

We have already changes the CI pipeline in enterprise. https://github.com/chainguard-dev/enterprise-packages/pull/7273

/CC @hectorj2f


tuananh:
can we check `target-architecture` in package yaml?

PR #32097, opened 2024-10-28
add crossplane-provider-keycloak package

cpanato:
xref: https://github.com/chainguard-dev/image-requests/issues/4530

### DIFF
--- a/crossplane-provider-keycloak.yaml
+++ b/crossplane-provider-keycloak.yaml
@@ -1,0 +1,70 @@
+package:
+  name: crossplane-provider-keycloak
+  version: 1.7.0
+  epoch: 0
+  description: Official Keycloak Provider for Crossplane by Upbound
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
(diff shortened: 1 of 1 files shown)


crossplane-provider-keycloak.yaml
```diff
@@ -0,0 +1,70 @@
...
+      export PATH=$PATH:$GOPATH/bin
+      go install golang.org/x/tools/cmd/goimports@latest
+      go run cmd/generator/main.go $(pwd)
+      make generate
+
+  - uses: go/build
+    with:
+      output: crossplane-keycloak-provider
+      packages: ./cmd/provider
+
```
Dentrax, line 50: Should we also pass `LDFLAGS` for _version_?

Also it seems they build the `generator` package, do we also need it?

cpanato: i was looking the image and there is no generator, i think that is only used to generate the code gen

Dentrax (dismissed):

jamonation (approved):

PR #32105, opened 2024-10-28
feat: create cloud-provider-aws package

sanghanan:
This package has two subpackages: aws-cloud-controller-manager and ecr-credential-provider

https://github.com/chainguard-dev/image-requests/issues/4932

Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches


cloud-provider-aws.yaml
```diff
@@ -0,0 +1,67 @@
+package:
+  name: cloud-provider-aws
```
kranurag7, line 2: this package needs to be version streamed across supported kubernetes versions. We can keep the package and file name as `cloud-provider-aws-1.31` here.

cloud-provider-aws.yaml
```diff
@@ -0,0 +1,67 @@
...
+  version: 1.31.1
+  epoch: 0
+  description: The AWS cloud provider provides the interface between a Kubernetes cluster and AWS service APIs.
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  environment:
+    CGO_ENABLED: "0"
+    GO111MODULE: "on"
```
kranurag7, line 12: we can omit this because go/build pipeline will handle this. We don't set this explicitly. Please let me know if there is any explicit reason for this?

cloud-provider-aws.yaml
```diff
@@ -0,0 +1,67 @@
...
+subpackages:
+  - name: aws-cloud-controller-manager
+    description: The AWS Cloud Controller Manager is the controller that is primarily responsible for creating and updating AWS loadbalancers (classic and NLB) and node lifecycle management.
+    pipeline:
+      - uses: go/build
+        with:
+          packages: ./cmd/aws-cloud-controller-manager
+          output: aws-cloud-controller-manager
+          ldflags: |
+            -w -s
```
kranurag7, line 30: -s is something we don't use and -w is handled by go/build pipeline. We can omit `-s -w` from ldflags.

cloud-provider-aws.yaml
```diff
@@ -0,0 +1,67 @@
...
+            -X k8s.io/component-base/version.gitVersion=v${{package.version}}
+            -X main.gitVersion=v${{package.version}}
+    test:
+      pipeline:
+        - runs: |
+            aws-cloud-controller-manager --help
+            aws-cloud-controller-manager --version
+
+  - name: ecr-credential-provider
```
kranurag7, line 39: ecr-credential-provider is not part the image so I think we can omit that part from the package. (this is not a blocker, we can package both inside here as well and use `${{package.name}}-foo`

cloud-provider-aws.yaml
```diff
@@ -0,0 +1,67 @@
...
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.contextdir}}/bin"
+          ln -sf /usr/bin/aws-cloud-controller-manager ${{targets.contextdir}}/bin/aws-cloud-controller-manager
+
+update:
+  enabled: true
+  github:
+    identifier: kubernetes/cloud-provider-aws
+    strip-prefix: v
```
kranurag7, line 67: if we are switching to version streams then we also need to filter based on tags here so we'll append `tag-filter: v1.31` here for 1.31 version stream.

kranurag7 (changes requested):
great work!! requesting some changes inline.

cloud-provider-aws.yaml
sanghanan: I wasn't sure that `GO111MODULE` was set by default. Will remove this!

cloud-provider-aws.yaml
sanghanan: Do we add `uses: strip`?

kranurag7: -w does the stripping for us so explicitly using `-w` is not required anytime we are using go/build pipeline.

cloud-provider-aws.yaml
sanghanan: I added it `ecr-credential-provider`, in case it was needed in the future.

kranurag7 (approved):
looks good to me, verified the build time ldflags upstream once again and now we are ready to 🚢 📦

PR #32110, opened 2024-10-28
karpenter: version stream and add test

Dentrax:
Start version stream for `kapenter`, add test and also provide the package as `karpenter-provider-aws` since upstream renamed it after version 1.

### DIFF
--- a/karpenter-1.0.yaml
+++ b/karpenter-1.0.yaml
@@ -1,10 +1,14 @@
 package:
-  name: karpenter
+  name: karpenter-1.0
   version: 1.0.6
   epoch: 0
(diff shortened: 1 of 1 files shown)


karpenter-1.0.yaml
```diff
@@ -1,10 +1,14 @@
 package:
-  name: karpenter
+  name: karpenter-1.0
   version: 1.0.6
   epoch: 0
```
jamonation, line 4: Still needs an epoch bump to make sure this one with provides wins out right?

Dentrax: I am not so sure about this, since we renamed the package, do we still need epoch bump?

ajayk: No epoch is needed here

ajayk (approved):

PR #32144, opened 2024-10-28
crac-criu.yaml: Add openssf-options

justinvreeland:


xnox:
See https://github.com/wolfi-dev/os/pull/33051

Issue #29568, opened 2024-09-27
exim - external DB lookups support not working


willdeane:
The Exim Makefile includes support for a number of external lookups but they aren't making into the final binary or being added as dynamic libs.  I'm specifically looking for postgresql support.
https://github.com/wolfi-dev/os/blob/main/exim/exim.Makefile

The Makefile includes config to build the lookup modules/support:
```
LOOKUP_CDB=2
LOOKUP_DBM=2
LOOKUP_DNSDB=2
LOOKUP_DSEARCH=yes
LOOKUP_LSEARCH=yes
LOOKUP_MODULE_DIR=/usr/lib/exim/
LOOKUP_MYSQL=2
LOOKUP_MYSQL_INCLUDE=-I/usr/include/mysql
LOOKUP_MYSQL_LIBS=-Wl,--no-as-needed -lmysqlclient
LOOKUP_PASSWD=yes
LOOKUP_PGSQL=2
LOOKUP_PGSQL_INCLUDE=-I/usr/include/postgresql
LOOKUP_PGSQL_LIBS=-Wl,--no-as-needed -lpq
LOOKUP_SQLITE=2
LOOKUP_SQLITE_LIBS=-Wl,--no-as-needed -lsqlite3
```
The `yes` indicates the support should be statically compiled into exim and the `2` indicates that they should be complied as dynamic libs. The `LOOKUP_MODULE_DIR=/usr/lib/exim/` value specifies where the dynamic libs should be deployed

Building exim from within the wolfi-os dev container (`make dev-container`) with `make package/exim` produces a package which which doesn't include any of the dynamic modules:
```
~ # apk info -L exim
exim-4.98-r1 contains:
etc/exim/exim.conf
etc/mail/aliases
usr/sbin/exim
usr/sbin/mailq
usr/sbin/newaliases
usr/sbin/rmail
usr/sbin/rsmtp
usr/sbin/runq
usr/sbin/sendmail
usr/share/man/man8/exim.8
var/lib/db/sbom/exim-4.98-r1.spdx.json
~ # 
```

Logs from the build suggest the shared libs are being built:
```
2024/10/04 17:06:21 INFO cc -shared cdb.c
2024/10/04 17:06:21 INFO cc -shared dbmdb.c
2024/10/04 17:06:22 INFO cc -shared dnsdb.c
2024/10/04 17:06:22 INFO cc -shared mysql.c
2024/10/04 17:06:22 INFO cc -shared pgsql.c
[...]

PR #30647, opened 2024-10-11
exim: add subpackages, for lookups, scripts and utils.

patdowney:
This PR adds sub-packages that are present in the alpine package (https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/community/exim/APKBUILD) that weren't in the Wolfi package. Specifically it adds sub-packages for lookups (postgres, mysql and others) as well as packaging the `exim-utils` and `exim-scripts` too. 
Additionally, as also present in the alpine package, it adds a pre-install script to configure the `exim` user upon package installation.

Fixes:
* https://github.com/wolfi-dev/os/issues/29568

#### For new package PRs only

- [ ] This PR is marked as fixing a pre-existing package request bug
  - [x] Alternatively, the PR is marked as related to a pre-existing package request bug, such as a dependency
- [x] REQUIRED - The package is available under an OSI-approved or FSF-approved license
- [x] REQUIRED - The version of the package is still receiving security updates
- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`)

PR #32155, opened 2024-10-28
exim: remove empty subpackage

justinvreeland:
Surprised the linter allows it tbh.

### DIFF
--- a/exim.yaml
+++ b/exim.yaml
@@ -1,7 +1,7 @@
 package:
   name: exim
   version: "4.98"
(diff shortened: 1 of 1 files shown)


willdeane:
@justinvreeland, there is an open PR waiting for review to build the subpackages which are missing from the current exim build - https://github.com/wolfi-dev/os/pull/30647, this fixes issue https://github.com/wolfi-dev/os/issues/29568 where the external lookup packages are built but not packaged.

PR #32180, opened 2024-10-29
feat: add promxy package

wojciechka:
#### For new package PRs only


promxy.yaml
```diff
@@ -0,0 +1,49 @@
...
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/jacksontj/promxy
+      tag: v${{package.version}}
+      expected-commit: 114942b664615bc68a67f56d2e458d7ca88f18d2
+
+  - runs: |
+      # remove vendored dependencies
+      rm -fR vendor/
```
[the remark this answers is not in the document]
wojciechka: +1 to this, I'll remove it ; thanks

promxy.yaml
```diff
@@ -0,0 +1,45 @@
...
+      output: promxy
+
+  - uses: go/build
+    with:
+      packages: ./cmd/remote_write_exporter
+      output: remote_write_exporter
+
+subpackages:
+  - name: promxy-compat
+    description: "Compatibility package to allow replacing promxy image"
```
kranurag7, line 29: we can write something on this line. 
```suggestion
    description: "Compatibility package to symlink binaries to match upstream deployment"
```

promxy.yaml
```diff
@@ -0,0 +1,45 @@
...
+update:
+  enabled: true
+  github:
+    identifier: jacksontj/promxy
+    strip-prefix: v
+
+test:
+  pipeline:
+    - runs: |
+        promxy --version
```
kranurag7, line 45: can we considering adding a help message test for the other binary here? by default it's giving `exit 1` on invoking `--help` but we can do something like this. 

```bash
/ # remote_write_exporter --help | grep drop-stale
      --drop-stale       drop metrics that are written with StaleNaN
/ # echo $?
0
```

promxy.yaml
wojciechka: Thank you. Updated it via another commit.

promxy.yaml
wojciechka: Good point! I was checking `--help` but did not think to `| grep` it. Done.

promxy.yaml
```diff
@@ -0,0 +1,47 @@
+package:
+  name: promxy
+  version: 0.0.90
+  epoch: 0
+  description: An aggregating proxy to enable HA prometheus.
+  copyright:
+    - license: MIT
```
kranurag7, line 7: I was doing a final check and noticed that upstream image is also pulling in ca-certificates-bundle in the final image. 
Should we also include ca-certificates-bundle as runtime dependency here looking at upstream image config? https://github.com/jacksontj/promxy/blob/a622d5348e93c81d4e929ddb12a8bf6d25199337/Dockerfile#L16

wojciechka: Good point. I'll add it as:

```
package:
  # ...
  dependencies:
    runtime:
      - ca-certificates-bundle
```

Dentrax (approved):

kranurag7 (approved):

PR #32191, opened 2024-10-29
freerdp: fix FTBFS with gcc-14

xnox:
Fix FTBFS with gcc-14, enable optimised build and hardening flags

### DIFF
--- a/freerdp.yaml
+++ b/freerdp.yaml
@@ -1,7 +1,7 @@
 package:
   name: freerdp
   version: 2.11.7
-  epoch: 2
+  epoch: 3
   description: FreeRDP client
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


freerdp.yaml
```diff
@@ -27,12 +27,13 @@ environment:
       - libxext-dev
       - libxi-dev
       - libxinerama-dev
-      # - libxkbfile-dev
       - libxkb-dev
       - libxkbcommon-dev
       - libxrender-dev
       - libxv-dev
       - linux-headers
+      # - libxkbfile-dev
```
imjasonh, line 35: Do we still need this if it's commented out?

xnox: i have no idea, but yam moved it. Imho commented out stuff is never needed as git history exists.

imjasonh (approved):

PR #32207, opened 2024-10-29
feat(makefile): set default melange runner to docker

maxgio92:
Continuous integration and automation runs pipelines in Docker with Melange, in order to provide consistency and homogeneous environment, it sets the default runner to Docker. The Make variable can be inherited and overridden.

  Related:
  - https://github.com/chainguard-dev/melange/pull/1611


xnox (approved):
This is nice! as this makes it easier to use preffered runner. And makes local default be the current production default.

89luca89 (approved):
I like this too :+1: Seems a sensible thing to do

PR #32227, opened 2024-10-29
fix(makefile): make QEMU easier to use, fetch kernel and set env variables

89luca89:
Use:

`eval $(make fetch-kernel` or just `make fetch-kernel` and do the exports printed manually

then you can simply use

`make package/foo` and it will use QEMU with local downloaded kernel

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,6 +97,15 @@ list-yaml:
 	$(info $(addsuffix .yaml,$(shell $(PKGLISTCMD))))
 	@printf ''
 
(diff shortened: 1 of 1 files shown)


philroche (approved):

xnox:
Silly question -> are we not able to use our wolfi kernel from enterprise packages that we are already building and fetch that?

Also can we make it have built-in modules for all the things without need to have modules?

xnox (approved):

PR #32228, opened 2024-10-29
add libnvme package to wolfi

kranurag7:
packaging this given this is required for https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver incoming package.

PR #32234, opened 2024-10-29
add nvme-cli package

kranurag7:
adds nvme-cli package

depends on https://github.com/wolfi-dev/os/pull/32228 

CI will fail for now given libnvme is not available as of now until the above package is merged.

### DIFF
--- a/nvme-cli.yaml
+++ b/nvme-cli.yaml
@@ -11,7 +11,6 @@ environment:
     packages:
       - bash
       - build-base
-      - json-c
       - libnvme-dev
       - meson
 
@@ -57,3 +56,8 @@ test:
   pipeline:
     - name: version test
       runs: nvme --version | grep ${{package.version}}
+    - name: more tests
+      runs: |
+        nvme list
+        nvme help | grep discover
+        nvme gen-tls-key
--- a/nvme-cli.yaml
+++ b/nvme-cli.yaml
@@ -1,0 +1,61 @@
+package:
+  name: nvme-cli
+  version: 2.10.2
+  epoch: 0
+  description: NVM-Express user space tooling for Linux
+  copyright:
+    - license: GPL-2.0-only
+
+environment:
+  contents:
+    packages:
+      - bash
+      - build-base
+      - json-c
+      - libnvme
+      - libnvme
+      - libnvme-dev
+      - meson
+
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: eeaa08c9a0e9184f3889df0bff3d2a23db6d6294
+      repository: https://github.com/linux-nvme/nvme-cli
+      tag: v${{package.version}}
+
+  - uses: meson/configure
+    with:
+      opts: |
+        -Ddocs=man \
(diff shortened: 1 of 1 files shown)


nvme-cli.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+  - uses: meson/compile
+
+  - uses: meson/install
+
+  - runs: |
+      rm -rf ${{targets.destdir}}/usr/lib/systemd
+      rm -rf ${{targets.destdir}}/usr/lib/udev
+      rm -rf ${{targets.destdir}}/usr/share/bash-completion
+      rm -rf ${{targets.destdir}}/usr/share/zsh
```
kranurag7, line 41: the package was picking up things under `/usr/lib/systemd` and `/usr/lib/dev` so removed that. 

bash and zsh completion were removed given this is a niche CLI and I was not sure if adding a completion for this would be useful here but in case if it is required happy to add that in subsequent commits.

nvme-cli.yaml
```diff
@@ -0,0 +1,59 @@
...
+  description: NVM-Express user space tooling for Linux
+  copyright:
+    - license: GPL-2.0-only
+
+environment:
+  contents:
+    packages:
+      - bash
+      - build-base
+      - json-c
```
developer-guy, line 14: we don't need json-c unless you specify a build option `-Djson-c=enabled`

nvme-cli.yaml
```diff
@@ -0,0 +1,59 @@
...
+  enabled: true
+  manual: false
+  github:
+    identifier: linux-nvme/nvme-cli
+    strip-prefix: v
+
+test:
+  pipeline:
+    - name: version test
+      runs: nvme --version | grep ${{package.version}}
```
Dentrax, line 59: Are we able to run `nvme list`?

nvme-cli.yaml
kranurag7: thanks for the review. Removed `json-c` given `libnvme-dev` already pulls in json-c. The binary is dynamically linking to json-c out of the box.

```bash
55307245e915:/work/packages# ldd $(command -v nvme)
        linux-vdso.so.1 (0x00007ffc9c7d0000)
        libnvme.so.1 => /usr/lib/libnvme.so.1 (0x00007c90d1d4b000)
        libnvme-mi.so.1 => /usr/lib/libnvme-mi.so.1 (0x00007c90d1d41000)
        libc.so.6 => /lib/libc.so.6 (0x00007c90d1b4c000)
        libjson-c.so.5 => /usr/lib/libjson-c.so.5 (0x00007c90d1b3c000)
        libcrypto.so.3 => /usr/lib/libcrypto.so.3 (0x00007c90d1600000)
        libkeyutils.so.1 => /usr/lib/libkeyutils.so.1 (0x00007c90d1b33000)
        /lib64/ld-linux-x86-64.so.2 (0x00007c90d1dfa000)
```

Dentrax (approved):

developer-guy (approved):

PR #32344, opened 2024-10-30
use a variable for php version (phpMM) and utilize pipelines in php files.

smoser:
Update the php-8.*-*.yaml files to have use a var transform based on package name for the php version, and then use replace any use of php version with that variable.

Also here:
 * drop a comment that causes issue with wolfictl bump 
   https://github.com/wolfi-dev/wolfictl/issues/1282
 * update to use pecl/install and pecl/phpize pipeline where possible
 * update php8.*-redis.yaml to use phpMM name, as it had already used php-version.

### DIFF
--- a/php-8.1-amqp.yaml
+++ b/php-8.1-amqp.yaml
@@ -1,15 +1,21 @@
 package:
   name: php-8.1-amqp
   version: 2.1.2
-  epoch: 1
+  epoch: 2
   description: "PHP extension to communicate with any AMQP compliant server"
   copyright:
     - license: PHP-3.01
   dependencies:
     runtime:
       - ${{package.name}}-config
-      - php-8.1
+      - php-${{vars.phpMM}}
       - rabbitmq-c
+
+var-transforms:
+  - from: ${{package.name}}
+    match: ^php-(\d+\.\d+)-.*$
+    replace: "$1"
+    to: phpMM
 
 environment:
   contents:
@@ -17,8 +23,8 @@ environment:
       - autoconf
       - build-base
       - busybox
-      - php-8.1
-      - php-8.1-dev
+      - php-${{vars.phpMM}}
+      - php-${{vars.phpMM}}-dev
       - rabbitmq-c-dev
 
 pipeline:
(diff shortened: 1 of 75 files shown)


smoser:
python program that made the bulk of these changes (inserting phpMM), executed as `python3 ./tmpl-phps php-8.*-*.yaml`:


```python
#!/usr/bin/python3
import re
import sys
import subprocess
from ruamel.yaml import YAML

re.compile(r"abc")
namepre = "php-8."
varname = "phpMM"
vartok = "${{vars." + varname + "}}"
repl = re.compile(r"([^a-zA-Z]php[-]{0,1})(\d[.]\d+)")

blob = r"""
var-transforms:
  - from: ${{package.name}}
    match: ^php-(\d+\.\d+)-.*$
    replace: "$1"
    to: phpMM
"""

changed = []
yaml = YAML(typ='rt')
for fname in sys.argv[1:]:
    if not fname.startswith(namepre):
        print(f"SKIP: {fname} not {namepre}")
        continue
    if not fname.endswith(".yaml"):
        print(f"SKIP: {fname} not yaml")
        continue
    with open(fname, "r") as fp:
        ocontent = fp.read()
    if varname in ocontent:
        print(f"SKIP: {fname} already had {varname}")
        continue
    pname = fname[0:-len(".yaml")]

    ncontent = repl.sub(r"\1" + vartok, ocontent)

    # now substitute *back* the 'name: php-' as we don't
    # want it updated.
    pver = fname[len("php-"):len("php-")+len("8.3")]
    ncontent = ncontent.replace(f"name: php-{vartok}-",
                                f"name: php-{pver}-", 1)

    # and insert the var-transform blob
    pos = ncontent.index("\nenvironment:")
    ncontent = ncontent[0:pos] + blob + ncontent[pos+1:]
    print(f"UPDATE: {fname}")
    with open(fname, "w") as fp:
        fp.write(ncontent)
    changed.append(fname)
    continue


subprocess.check_output(["yam"] + changed)
subprocess.check_output(["wolfictl", "bump"] + changed)

sys.exit(0)
```

ajayk (approved):

PR #32345, opened 2024-10-30
add gcp-compute-persistent-disk-csi-driver package

kranurag7:
this patch adds gcp-compute-persistent-disk-csi-driver package

### DIFF
--- a/gcp-compute-persistent-disk-csi-driver-1.15.yaml
+++ b/gcp-compute-persistent-disk-csi-driver-1.15.yaml
@@ -67,7 +67,7 @@ pipeline:
 
 subpackages:
   - name: "${{package.name}}-compat"
-    description: "compatibility package to place binaries at / as expected by upstream deployment"
+    description: "compatibility package to place binaries and files at places expected by upstream deployment"
     pipeline:
       - runs: |
           mkdir -p ${{targets.subpkgdir}}/lib/udev_containerized
--- a/gcp-compute-persistent-disk-csi-driver-1.15.yaml
+++ b/gcp-compute-persistent-disk-csi-driver-1.15.yaml
@@ -9,6 +9,7 @@ package:
     runtime:
       - bash
       - blkid
+      - blockdev
       - btrfs-progs
       - busybox
       - ca-certificates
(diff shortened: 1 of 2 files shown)


gcp-compute-persistent-disk-csi-driver-1.15.yaml
```diff
@@ -0,0 +1,89 @@
...
+    with:
+      repository: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver
+      tag: v${{package.version}}
+      expected-commit: 8c21c23edf516250992f2b9a0a2cd667489d30b8
+
+  - uses: go/build
+    with:
+      packages: ./cmd/gce-pd-csi-driver/main.go
+      ldflags: |
+        -X main.version=$(git describe --long --tags --match='v*' --dirty 2>/dev/null || git rev-list -n1 HEAD)
```
developer-guy, line 65: can't we just use ${{package.version}} here?

kranurag7: Yes, we can. This is what upstream was using. It contains a mix of version and commit if done in dev but we are not doing that so we can proceed with using `v${{package.version}}`

but agree we can update it looking at logs. 

```bash
$ docker run --rm registry.k8s.io/cloud-provider-gcp/gcp-compute-persistent-disk-csi-driver:v1.15.1 -v 10
I1031 12:55:34.509737       1 main.go:114] Operating compute environment set to: production and computeEndpoint is set to: <nil>
I1031 12:55:34.510037       1 main.go:123] Sys info: NumCPU: 11 MAXPROC: 1
I1031 12:55:34.510043       1 main.go:128] Driver vendor version v1.15.1
```

gcp-compute-persistent-disk-csi-driver-1.15.yaml
```diff
@@ -0,0 +1,89 @@
...
+  - uses: go/build
+    with:
+      packages: ./cmd/gce-pd-csi-driver/main.go
+      ldflags: |
+        -X main.version=v${{package.version}}
+      output: gce-pd-csi-driver
+
+subpackages:
+  - name: "${{package.name}}-compat"
+    pipeline:
```
EyeCantCU, line 70: Missing package description

EyeCantCU (changes requested):

EyeCantCU (approved):

PR #32378, opened 2024-10-30
dotnet-7.yaml: Add openssf-options

justinvreeland:


xnox:
dotnet 8 update last week landed in https://github.com/wolfi-dev/os/pull/33965 and is fine.

Note sure we care about dotnet-7; as i think that was building fine too.

PR #16420, opened 2024-04-05
php-8.3: add php provide

shyim:
As the provide is missing, I cannot install php-8.3 + composer togeher right now

#### For new package PRs only

- [ ] This PR is marked as fixing a pre-existing package request bug
  - [ ] Alternatively, the PR is marked as related to a pre-existing package request bug, such as a dependency
- [X] REQUIRED - The package is available under an OSI-approved or FSF-approved license
- [X] REQUIRED - The version of the package is still receiving security updates
- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`)

#### For new version streams

- [ ] The upstream project actually supports multiple concurrent versions.
- [ ] Any subpackages include the version string in their package name (e.g. `name: ${{package.name}}-compat`)
- [ ] The package (and subpackages) `provides:` logical unversioned forms of the package (e.g. `nodejs`, `nodejs-lts`)

#### For security-related PRs

- [ ] The security fix is recorded in the [advisories](https://github.com/wolfi-dev/advisories) repo

#### For version bump PRs

- [ ] The `epoch` field is reset to 0

#### For PRs that add patches

- [ ] Patch source is documented

PR #32398, opened 2024-10-30
php modules - Drop virtual provides , replace remaining 8.X occurences.

smoser:
We currently have 3 packages of each php module, one for each version of PHP.  For amqp, that is:

   php-8.1-amqp
   php-8.2-amqp
   php-8.3-amqp

Each of these has a dependency on the corresponding php-8.X.

Before this commit, the module for the newest php version was also providing 'php-amqp', such that if something just wanted 'php-amqp', it would get php-8.3-amqp.

That is somewhat fragile though and would require us to remove the provides from an 8.3 package when 8.4 is released. (see https://github.com/wolfi-dev/os/pull/16420)

We don't want to have to re-build php-8.3-* modules in order to add php-8.4 versions.

The other option here is to use provider-priority as we do in python. But in order to get there, we would probably have to remove all the previously published apks as they would 'provide' an explicit version.

### DIFF
--- a/php-8.1-amqp.yaml
+++ b/php-8.1-amqp.yaml
@@ -1,7 +1,7 @@
 package:
   name: php-8.1-amqp
   version: 2.1.2
-  epoch: 2
+  epoch: 3
   description: "PHP extension to communicate with any AMQP compliant server"
   copyright:
     - license: PHP-3.01
@@ -51,9 +51,6 @@ pipeline:
 
 subpackages:
   - name: ${{package.name}}-config
-    dependencies:
-      provides:
-        - php-amqp-config=${{package.full-version}}
     pipeline:
       - runs: |
           mkdir -p "${{targets.subpkgdir}}/etc/php/conf.d"
--- a/php-8.1-igbinary.yaml
+++ b/php-8.1-igbinary.yaml
@@ -1,7 +1,7 @@
 package:
   name: php-8.1-igbinary
   version: 3.2.16
-  epoch: 1
+  epoch: 2
   description: "Igbinary is a drop in replacement for the standard php serializer."
   copyright:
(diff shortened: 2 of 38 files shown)


smoser:
I'm wondering what we should do here.

The general problem is that we want to have something like the [worldpress image](https://github.com/chainguard-images/images/blob/main/images/wordpress/config/main.tf) "move forward" automatically to a newer php version.

That moving forward is non-trivial:
* wordpress needs 23(?) php modules.  If one of those doesn't get updated right away for php-8.N+1, then  it would break.
* those modules are not likely to be availailable immediately upon release of php-8.N+1

Thoughts? @xnox or @ajayk ?

ajayk:
> I'm wondering what we should do here.
> 
> The general problem is that we want to have something like the [worldpress image](https://github.com/chainguard-images/images/blob/main/images/wordpress/config/main.tf) "move forward" automatically to a newer php version.
> 
> That moving forward is non-trivial:
> 
> * wordpress needs 23(?) php modules.  If one of those doesn't get updated right away for php-8.N+1, then  it would break.
> * those modules are not likely to be availailable immediately upon release of php-8.N+1
> 
> Thoughts? @xnox or @ajayk ?

Couple of days of non-availability might be fine , @joshrwolf is working on making the version streams decoupled  from failing the the whole batch for eol grace period , with that the previous php versions will be still built

smoser:
I'm going to close this PR.
The replacment of 8.x with phpMM is done in https://github.com/wolfi-dev/os/pull/33015 .

I'll address the other changes to virtual provides under  https://github.com/wolfi-dev/os/issues/32874 but not on a friday.

smoser:
closing. replaced with https://github.com/wolfi-dev/os/pull/34680

PR #32441, opened 2024-10-30
git-bootstrap.yaml: Add openssf-options

justinvreeland:


dannf (approved):
Thanks for cleaning up the CHOST thing. I'm still unsure when we should be adding `openssf-compiler-options` explicitly, but it's not a problem.

justinvreeland:
> Thanks for cleaning up the CHOST thing. I'm still unsure when we should be adding `openssf-compiler-options` explicitly, but it's not a problem.

I'm just resolving PRs that have been open since before it was the default option. It's not needed but since this was the purpose of the PR and is reflected in all the similarly tagged PRs and I like maintaining the consistency.

PR #32446, opened 2024-10-30
glibc.yaml: Add openssf-options (and disable it)

justinvreeland:


dannf (commented):
Thanks @justinvreeland. I haven't had time to dig into the root cause yet, so let's go with this. 

Would you mind adding a link to https://github.com/chainguard-dev/internal-dev/issues/7756 above the GCC_SPEC_FILE setting?

Should we guard this to only apply to the ARM builds?

justinvreeland:
> Would you mind adding a link to https://github.com/chainguard-dev/internal-dev/issues/7756 above the GCC_SPEC_FILE setting?

Done.


> Should we guard this to only apply to the ARM builds?

I'm not sure if we can do that in the yaml tbh. But in general I prefer keeping things as close as possible in the architectures. Don't have particularly strong feelings about that though.

dannf:
> > Would you mind adding a link to [chainguard-dev/internal-dev#7756](https://github.com/chainguard-dev/internal-dev/issues/7756) above the GCC_SPEC_FILE setting?
> 
> Done.
> 
> > Should we guard this to only apply to the ARM builds?
> 
> I'm not sure if we can do that in the yaml tbh. 

Yeah, probably not in the `environment` section, so we'd have to do something like this in a `runs:`

```
[ "${{build.arch}}" = "aarch64" ] || export GCC_SPEC_FILE=/dev/null
```

> But in general I prefer keeping things as close as possible in the architectures. Don't have particularly strong feelings about that though.

Fair enough, let's just merge this as-is. Thanks!

dannf (approved):

justinvreeland:
Rebasing to see if that fixes whatever is stopping the merge

xnox:
rebased PR here https://github.com/wolfi-dev/os/pull/38280 hopefully it will pass CI.

justinvreeland:
I rebased this one but looks like we have SBOM issues. Checked the maintainer edit button.

jonjohnsonjr (approved):

PR #32475, opened 2024-10-30
hdf5.yaml: Add openssf-options

justinvreeland:


justinvreeland:
I think the new options are causing problems with the configure script but I can't quite figure out how or why. I had this marked to not use the openssf_spec_file but don't seem to have added a comment for why I choose that : (

justinvreeland:
I can confirm that our options are causing problems with this configure script. I'm not really sure I want to fix it given that it cakes and autoconf and seems fragile overall.

dannf (approved):
Yeah, fixes a FTBFS for me, and I couldn't quickly find a flag we could invert to make it work w/o disabling everything. I might continue to poke, but this is definitely better than a broken build. Thanks @justinvreeland !

PR #32497, opened 2024-10-30
ingress-nginx-controller-1.11.yaml: Add openssf-options

justinvreeland:


justinvreeland:
The test is failing because the nginx module directory it's pointing to doesn't exist

justinvreeland:
We're currently distributing similar files. They fail in malcontent because they're tests for vulnerabilities and thus seem like sketch exploits which is reasonable since tests for exploits are pretty much indistiguishable from exploits.

mamccorm:
Thanks @justinvreeland - have tagged in @egibs re: the malcontent scans. Once his team add their stamp of approval (and note down any rule tweaks that can be made), then this one should be cleared for merge

egibs:
The malcontent findings were keying off the OWASP CRS: https://github.com/coreruleset/coreruleset/tree/main

mamccorm (approved):

PR #32505, opened 2024-10-30
istio-envoy-1.23.yaml: Add openssf-options

justinvreeland:


justinvreeland:
Something weird with downloading internal to the build system here it looks like

justinvreeland:
Looks like we don't have istio-envoy-1.23 anymore it was replaced by 1.24 which builds with the default options enabled.

PR #32701, opened 2024-10-31
Check for .pc files in usr/lib and usr/share

murraybd:
This fixes the following test failure with pkgconf:

```
   2024/10/31 12:56:35 WARN + echo usr/share/pkgconfig/catch2-with-main.pc uses=test/pkgconf name="pkgconf build dependency check"
   2024/10/31 12:56:35 WARN + grep -q '^usr/lib/pkgconfig/catch2-with-main.pc$' uses=test/pkgconf name="pkgconf build dependency check"
   2024/10/31 12:56:35 INFO ERROR: failed to test package. the test environment has been preserved:

```

### DIFF
--- a/glew.yaml
+++ b/glew.yaml
@@ -58,9 +58,6 @@ subpackages:
       - uses: split/dev
       - runs: |
           rm -rf "${{targets.subpkgdir}}"/usr/lib/cmake
-    test:
-      pipeline:
-        - uses: test/pkgconf
 
 update:
   enabled: true
--- a/glew.yaml
+++ b/glew.yaml
@@ -58,6 +58,9 @@ subpackages:
       - uses: split/dev
       - runs: |
           rm -rf "${{targets.subpkgdir}}"/usr/lib/cmake
+    test:
+      pipeline:
+        - uses: test/pkgconf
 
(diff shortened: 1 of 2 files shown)


dustinkirkland:
Okay, with that patch, these package tests now succeed:

Catch2
bash-completion
cli11
eigen
eudev
fish
gi-docgen
gtk-doc
hicolor-icon-theme
hwdata
iso-codes
kmod
libbpf
mimalloc
mimalloc2
nfs-ganesha
nlohmann-json
numactl
samba
scdoc
shared-mime-info
spirv-headers
systemd
util-macros
wayland-protocols
xbitmaps
xcb-proto
xkeyboard-config
xorgproto
xtrans
yajl

dustinkirkland (changes requested):
Hmm....I think this grep regex would be better like this:
`
          echo "$pc_file" | grep -q "^usr/lib/pkgconfig/${lib_name}.pc$\|^usr/share/pkgconfig/${lib_name}.pc$"
`

dustinkirkland (dismissed):

dustinkirkland (approved):

PR #32784, opened 2024-11-01
add libyang package to wolfi

kranurag7:
adds libyang package to wolfi

### DIFF
--- a/libyang.yaml
+++ b/libyang.yaml
@@ -14,6 +14,7 @@ environment:
       - cmocka-dev
       - doxygen
       - graphviz
+      - openssf-compiler-options
       - pcre2-dev
       - tcl
       - valgrind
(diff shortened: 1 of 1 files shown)


kranurag7:
going through the build logs I find the following:

```bash
2024/11/01 10:56:52 INFO -- Looking for strnstr uses=cmake/configure
2024/11/01 10:56:52 INFO -- Looking for strnstr - not found uses=cmake/configure
```
but I think we can skip this because we are only building for linux.

This is project APKBUILD that upstream uses to build libyang library. (ref: https://github.com/FRRouting/frr/blob/master/docker/alpine/libyang/APKBUILD)

developer-guy (approved):

Dentrax (approved):

PR #32818, opened 2024-11-01
nodejs-19.yaml: Add openssf-options

justinvreeland:


justinvreeland:
The failure to build we're seeing seem unrelated to our Openssf compiler options. However I believe we'll have the same abseil issue that other projects have it as well.

justinvreeland:
Since I'm reasonably sure this isn't because of the Openssf compiler options I'm closing this in favor of https://github.com/wolfi-dev/os/issues/36399 for tracking

PR #32827, opened 2024-11-01
nvidia-device-plugin.yaml: Add openssf-options

justinvreeland:


xnox:
Fixed by https://github.com/wolfi-dev/os/pull/34569 due to gcc upstream bug.

PR #32836, opened 2024-11-01
openipmi.yaml: Add openssf-options

justinvreeland:


justinvreeland:
```
datahash = 39e25f55cea35f1d3df129d2cafbe6437fd62512ca61d1099c49fd90df5d9502
  pc:ncurses (virtual):
    note: please select one of the 'provided by'
          packages explicitly
    provided by: ncurses-dev-6.5_p20240629-r2
                 ncurses-dev-6.5_p20241006-r3
    required by: openipmi-dev-2.0.36-r1[pc:ncurses]
```

justinvreeland:
@xnox  is this not resolving to r3 because the version ignores everything after the `_` are all of our git hash versioned packages going to fail to resolve epoch updates because of the githash?

xnox:
> @xnox is this not resolving to r3 because the version ignores everything after the `_` are all of our git hash versioned packages going to fail to resolve epoch updates because of the githash?

does adding "dependencies: runtime: - ncurses-dev" help at all on the openmpi-dev subpackage?

justinvreeland:
> does adding "dependencies: runtime: - ncurses-dev" help at all on the openmpi-dev subpackage?

Yes I updated the PR with the fix, I was curious if I understood the cause correctly and if we'll have to keep a watch out for this in other packages.

xnox:
> > does adding "dependencies: runtime: - ncurses-dev" help at all on the openmpi-dev subpackage?
> 
> Yes I updated the PR with the fix, I was curious if I understood the cause correctly and if we'll have to keep a watch out for this in other packages.

I believe this is mostly expected in this case - but commented this in detail at https://github.com/chainguard-dev/melange/pull/1522#issuecomment-2458262019 

I'm not too sure what else might generate similar multiple provides which are unresolvable.

xnox (approved):

PR #32883, opened 2024-11-01
py3-ml-metadata.yaml: Add openssf-options

justinvreeland:


justinvreeland:
This is a bit annoying. We swapped to overrides with `GCC_SPEC_FILE` because some build systems ignore environment variables setting via common compiler flags.  It seems here we have a case of py3-ml-metadata under certain circumstances ignoring all unknown environment variables.

I can't override the FORTIFY_SROUCE through flags, when I update our OPENSSF_COMPILER_OPTIONS to point to one without fortify source I get a new error: 

```
2024/12/09 16:28:23 WARN   gcc: error: unrecognized command-line option '-Wthread-safety' subpackage=py3.10-ml-metadata uses=py/pip-build-install name="pip build"
2024/12/09 16:28:23 WARN   gcc: error: unrecognized command-line option '-Wself-assign' subpackage=py3.10-ml-metadata uses=py/pip-build-install name="pip build"
2024/12/09 16:28:23 WARN   gcc: error: unrecognized command-line option '-fcolor-diagnostics' subpackage=py3.10-ml-metadata uses=py/pip-build-install name="pip build"
```

While compiling one of the vendors deps. Which is unpleasant because it seems our compiler options are causing Bazel problems by being different depending on if the environment variable is seen by a particular build. And also because protobuf is on the list of projects that we can't build with our options right now: https://github.com/wolfi-dev/os/issues/34075

The only way I've found to get this package to build so far is to create an empty `openssf-compiler-options-null` package that provides `openssf-compiler-options` which is really what we're trying to avoid.

Using the various Bazel options to try and pass additional environment variables does not seem to have an impact and while there's and example of updated dependencies in our patches I haven't found a combination that works for a full build.

xnox:
Please add explicit build-depends on openssf-compiler-options; as i see in the logs that -r5 is getting installed, instead of -r6.

xnox:
Also, in the future, to support https://github.com/chainguard-dev/internal/discussions/3879 i am hoping to do opportunistic inclusion of melange.spec file.

I wonder if we should really create a null.spec file; and make the config file specify the spec file to use (or symlink the active spec file?!) such that one can self-set that; or we can package those as things we can depend on.

Cause in the future, i hope that non-hardened builds, pick up ELF notes automatically still.

justinvreeland:
That's a bit annoying it looks like it's still getting the current version of openssf-compiler-options.

justinvreeland:
Yay!

openssf-compiler-options.yaml
```diff
@@ -57,6 +57,14 @@ test:
         touch foo.c
         gcc -v foo.c &>stderr.log || true
         grep gcc_s stderr.log
+    - name: Ensure etc specfile is used
+      runs: |
+        touch /etc/gcc_spec_file
+        gcc -v 2>&1 | grep "/etc/gcc_spec_file"
+    - name: Ensure GCC_SPEC_FILE is used over "/etc/gcc_spec_file"
+      runs: |
```
smoser, line 65: can you create /etc/gcc_spec_file here ? to be clear that it exists ?

smoser: and maybe just rm it above to make the tests more standalone.

py3-ml-metadata.yaml
```diff
@@ -49,6 +50,9 @@ pipeline:
     with:
       series: series
 
+  - runs: |
+      touch /etc/gcc_spec_file
```
smoser, line 54: i think i would prefer Setting the environment var to /dev/null.
I'm interested in your thoughts on that versus this.

openssf-compiler-options.yaml
justinvreeland: Can do.

py3-ml-metadata.yaml
justinvreeland: Doesn't work which is why we have I'm making the file. The build eventually clears all the environment variables.

justinvreeland: Which is to say I also would've preferred the environment variable.

xnox (approved):

PR #33025, opened 2024-11-01
openssf-compiler-options: add -Wno-error

justinvreeland:


justinvreeland:
Imaginary option that should exist

PR #33043, opened 2024-11-01
openssf-compiler-options: add bypass

xnox:
If for any reason a bypass is desired add environment variable support
in the gcc wrapper to use any other spec file for example /dev/null.

We don't have wrappers for clang, and it doesn't have env variable to
change config directly, but it does have command-line option to
override configs location at runtime, which will be documented.

Implement test cases in test/compiler-hardening-check that checks that
both bypass methods work.

### DIFF
--- a/openssf-compiler-options/usr/local/bin/gcc-wrapper
+++ b/openssf-compiler-options/usr/local/bin/gcc-wrapper
@@ -1,2 +1,2 @@
 #!/bin/sh
-exec /usr/bin/${0##*/} -specs openssf.spec "$@"
+exec /usr/bin/${0##*/} -specs "${GCC_SPEC_FILE:-openssf.spec}" "$@"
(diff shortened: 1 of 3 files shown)


openssf-compiler-options/usr/local/bin/gcc-wrapper
```diff
@@ -1,2 +1,2 @@
 #!/bin/sh
-exec /usr/bin/${0##*/} -specs openssf.spec "$@"
+exec /usr/bin/${0##*/} -specs ${GCC_SPEC_FILE:-openssf.spec} "$@"
```
smoser, line 2: might as well quote that.

"${GCC_SPEC_FILE:-openssf.spec}"

smoser (approved):
so the user has to know that gcc is being used ? versus clang?  does it work for

PR #33149, opened 2024-11-04
add libyang2.1 package

kranurag7:
this is a follow-up of https://github.com/wolfi-dev/os/pull/32784 

the reason to package this old non-supported version is because https://github.com/CESNET/libyang/releases/tag/v2.2.8 which is a breaking change was introduced here. 

this library is required for frr and then seem to be pinning this version and without building this we cannot build frr.

### DIFF
--- a/libyang-2.1.yaml
+++ b/libyang-2.1.yaml
@@ -66,6 +66,7 @@ update:
   github:
     identifier: CESNET/libyang
     strip-prefix: v
+    tag-filter: v2.1.
 
 test:
   environment:
--- a/libyang-2.1.yaml
+++ b/libyang-2.1.yaml
@@ -1,0 +1,85 @@
+package:
+  name: libyang-2.1
+  version: 2.1.148
+  epoch: 0
+  description: YANG data modeling language library
+  copyright:
+    - license: BSD-3-Clause
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - cmake
+      - cmocka-dev
+      - doxygen
+      - graphviz
+      - openssf-compiler-options
+      - pcre2-dev
+      - tcl
+      - valgrind
+
+pipeline:
+  - uses: git-checkout
+    with:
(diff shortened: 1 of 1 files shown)


developer-guy (approved):

jamonation:
I'm not sure this needs to be pinned - I've built frr using libyang 3.4.5 with no compilation issues?

jamonation:
Very much just built frr with the default 3.4.2:

```
[14:22:48|24-11-04]user@jamon /home/user/frr/libyang
 0 % yanglint -v
yanglint 3.4.2
```

And then frr libraries are linked against it:

```
ldd ./eigrpd/.libs/eigrpd |grep yang
	libyang.so.3 => /usr/lib/libyang.so.3 (0x000079774ec28000)
```

kranurag7:
@jamonation You're right, I was also able to build frr-10.1.1 with libyang but for any previous version which is 10.0.2, 9.0.4, 9.1.2, 8.4.6 and 8.5.6 I'm not able to build without using this version of library. 

With this version all the remaining 5 are building okay. 

TLDR is: for 10.1.1 we are still using libyang but for rest we need libyang of version 2.x.x.

upstream is using this version of library https://github.com/FRRouting/frr/blob/master/docker/alpine/libyang/APKBUILD#L4

the issue I was running into in the morning was following: 
- lua libs were missing out even though they were present (I raised with maintainer of the project here 
- not able to built with gcc-14, going through this issue I found that the version that needs to be there is libyang 2.x.x (from here I started building with libyang 2 and everything worked smoothly https://github.com/FRRouting/frr/issues/15855#issuecomment-2080149731 this is the exact error I was getting) 

I'm opening all the PRs after this if you want to run a `make debug/frr` with `libyang` library. I opened a PR here https://github.com/wolfi-dev/os/pull/33165 for 10.1 version stream.

kranurag7:
thank you for the reviews.
closing this one given in wolfi we only track latest versions and this is now moved to epkgs.

PR #33210, opened 2024-11-04
openssl: pin versions of libcrypto3 and libssl3 for CLI package

lyoung-confluent:
If a Wolfi image already has an older version of the `libcrypto3` and/or `libssl3` package installed, and the `openssl` CLI is installed at runtime (or as say a `docker build` step), the `openssl` package will install correctly but will crash when used:
```
/ # apk list --installed | grep libcrypto
libcrypto3-3.3.2-r2 aarch64 {openssl} (Apache-2.0) [installed]
/ # apk add openssl
(1/2) Installing openssl-provider-legacy (3.4.0-r1)
(2/2) Installing openssl (3.4.0-r1)
OK: 15 MiB in 20 packages
/ # openssl
openssl: /usr/lib/libssl.so.3: version `OPENSSL_3.4.0' not found (required by openssl)
openssl: /usr/lib/libcrypto.so.3: version `OPENSSL_3.4.0' not found (required by openssl)
```
This is because the `openssl` CLI must use the exact same version (or at least the same `<major>.<minor>.<patch>`) of `libcrypto3` and `libssl3` that the `openssl` CLI was built with.

Currently, the relationship between `openssl` and `libcrypto3` and `libssl3` is picked up automatically by melange's SCA:
```
openssl-3.4.0-r1 depends on:
openssl-provider-legacy
so:ld-linux-aarch64.so.1
so:libc.so.6
so:libcrypto.so.3
so:libssl.so.3
```
However, this dependency will be satisfied by _any_ version of `libcrypto3` and `libssl3`.

This PR adds an explicit dependency on the matching versions which should prevent the installation (or force an upgrade) of the `openssl` CLI if the already installed `libcrypto3` and/or `libssl3` versions mismatch.

### DIFF
--- a/openssl.yaml
+++ b/openssl.yaml
@@ -19,8 +19,8 @@ package:
       # However, the version of OpenSSL must match exactly or the CLI will crash due to missing symbols
       # Example: '/usr/lib/libcrypto.so.3: version `OPENSSL_3.4.0' not found (required by openssl)'
       # To avoid this, we explicitly pin the versions of libcrypto3 and libssl3 to match the CLI version
-      - libcrypto3=${{package.full-version}}
-      - libssl3=${{package.full-version}}
+      - libcrypto3>=${{package.version}}
+      - libssl3>=${{package.version}}
 
 environment:
   contents:
--- a/openssl.yaml
+++ b/openssl.yaml
@@ -2,7 +2,7 @@
 package:
   name: openssl
   version: 3.4.0
-  epoch: 1
+  epoch: 2
   description: "the OpenSSL cryptography suite"
   copyright:
     - license: Apache-2.0
@@ -15,6 +15,12 @@ package:
   dependencies:
     runtime:
       - openssl-provider-legacy
+      # Melange will automatically add a dependency on 'so:libcrypto.so.3' and 'so:libssl.so.3' via SCA
+      # However, the version of OpenSSL must match exactly or the CLI will crash due to missing symbols
+      # Example: '/usr/lib/libcrypto.so.3: version `OPENSSL_3.4.0' not found (required by openssl)'
+      # To avoid this, we explicitly pin the versions of libcrypto3 and libssl3 to match the CLI version
+      - libcrypto3=${{package.full-version}}
+      - libssl3=${{package.full-version}}
 
 environment:
   contents:
--- a/openssl.yaml
+++ b/openssl.yaml
@@ -16,9 +16,11 @@ package:
     runtime:
       - openssl-provider-legacy
       # Melange will automatically add a dependency on 'so:libcrypto.so.3' and 'so:libssl.so.3' via SCA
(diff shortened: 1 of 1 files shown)


openssl.yaml
```diff
@@ -15,6 +15,12 @@ package:
   dependencies:
     runtime:
       - openssl-provider-legacy
+      # Melange will automatically add a dependency on 'so:libcrypto.so.3' and 'so:libssl.so.3' via SCA
+      # However, the version of OpenSSL must match exactly or the CLI will crash due to missing symbols
+      # Example: '/usr/lib/libcrypto.so.3: version `OPENSSL_3.4.0' not found (required by openssl)'
+      # To avoid this, we explicitly pin the versions of libcrypto3 and libssl3 to match the CLI version
+      - libcrypto3=${{package.full-version}}
+      - libssl3=${{package.full-version}}
```
xnox, line 24: strict pin is not needed, as ">=" should be sufficient. I have pondered of doing this, but it will do very little.

Note all containers built prior to 3.4.0-r0 landing, have libcrypto3 & libssl3 pinned to ~3.3 version. And whilst 3.4.0-r2 will be excluded, instead 3.4.0-r1 will be installed. Unless we somehow rebuild all past openssl versions; or at least withdraw 3.4.0-r0 and 3.4.0-r1 (which breaks image rebuild reproducibility).

I'm happy to take this change, but it will likely only have meaningful effect with next upgrade to 3.5 in 6 months time. To avoid this same "experience" then.

lyoung-confluent: > strict pin is not needed, as ">=" should be sufficient. I have pondered of doing this, but it will do very little.

Would you prefer pinning to the same `<major>.<minor>.<patch>` version which is _technically_ all that's required for it to work:
```
      - libcrypto3>=${{package.version}}
```
Or instead the full-version including epoch:
```
      - libcrypto3>=${{package.full-version}}
```

lyoung-confluent: > I'm happy to take this change, but it will likely only have meaningful effect with next upgrade to 3.5 in 6 months time. To avoid this same "experience" then.

That's fine, we've worked around the issue by strictly pinning the installed version of `openssl` in the `apk add` command to match, this PR was mostly to avoid anyone having the same experience in the future if possible

xnox: This by-hand change will fix installing openssl tool; which is a common operation but it will not fix this issue for all other packages. See for example "apk add python does not work" https://github.com/wolfi-dev/os/issues/33218 which is exactly this same issue.

I wish for melange to automatically generate versioned `so:LIB` provides, such that one can do depends `so:LIB>=$libversion` automatically. i opened https://github.com/chainguard-dev/melange/issues/1621 to track suck feature.

Please note this experience is not present with images built with apko; which always build images with a consistent set of packages; including self-service way of doing that for Chainguard Images customers.

PR #33407, opened 2024-11-05
py3-bokeh multiversion

dannf:
Add multiversion support to `py3-bokeh`

### DIFF
--- a/py3-bokeh.yaml
+++ b/py3-bokeh.yaml
@@ -2,22 +2,23 @@
 package:
   name: py3-bokeh
   version: 3.6.1
-  epoch: 0
+  epoch: 1
   description: Interactive plots and applications in the browser from Python
   copyright:
     - license: BSD-3-Clause
   dependencies:
-    runtime:
-      - py3-jinja2
-      - py3-contourpy
-      - numpy
-      - py3-packaging
-      - py3-pandas
-      - py3-pillow
-      - py3-pyyaml
-      - py3-tornado
-      - py3-xyzservices
-      - python-3
+    provider-priority: 0
+
+vars:
+  pypi-package: bokeh
+
+data:
+  - name: py-versions
+    items:
+      3.10: "310"
+      3.11: "311"
+      3.12: "312"
+      3.13: "300"
 
 environment:
   contents:
@@ -25,8 +26,10 @@ environment:
       - build-base
       - busybox
       - ca-certificates-bundle
-      - py3-setuptools
-      - python-3
+      - py3-supported-cython
+      - py3-supported-pip
+      - py3-supported-python-dev
+      - py3-supported-setuptools-git-versioning
       - wolfi-base
 
 pipeline:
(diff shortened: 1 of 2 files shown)


smoser:
Just repeating here so you see it.  I have a branch of files converted to mult-version committed that you can start from.
it may or may not help you learn.
https://github.com/chainguard-dev/internal-dev/issues/5334#issuecomment-2460116967

py3-bokeh.yaml
```diff
@@ -38,8 +107,6 @@ pipeline:
   - name: Python Build
     uses: python/build-wheel
 
-  - uses: strip
-
```
smoser, line 42: i think we're leaving these in. they don't do anything if there aren'[t c modules, but its not harmful (and if it was, we need to fix it).

smoser (changes requested):
one change request.

py3-setuptools-git-versioning.yaml
```diff
@@ -0,0 +1,113 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: 1a2b9d0674b881e04b2a1238372b74020558274e
+      repository: https://github.com/dolfinus/setuptools-git-versioning
+      tag: v${{package.version}}
+
+  - name: Python Build
+    uses: python/build-wheel
```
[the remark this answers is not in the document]
dannf: nice catch!

py3-setuptools-git-versioning.yaml
```diff
@@ -0,0 +1,114 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: 1a2b9d0674b881e04b2a1238372b74020558274e
+      repository: https://github.com/dolfinus/setuptools-git-versioning
+      tag: v${{package.version}}
+
+  - name: Python Build
+    uses: py/pip-build-install
```
smoser, line 105:
```yaml
        with:
          python: python${{range.key}}
```

dannf: Actually, do we just need to kill this section now that the builds are happening in the `subpackages:` ?

smoser: actually, yes. good catch.

smoser: and can you move the pipeline above the subpackages ? i really wich 'yam' would do all the things, but it does not.

dannf: Could we add a jsonschema for the YAML, and and a `pre-commit` hook to validate it? That could enforce ordering, as well as using `yam`

smoser (approved):
Thanks

smoser (changes requested):
actually. i'm going to ask one more change.
Can you add a variable 'import' and use it?

All of the recently modified py3-* files are using that.  (see [py3-absl-py](https://github.com/wolfi-dev/os/blob/main/py3-absl-py.yaml) for example).  That just gives us better data to work with if we're refactoring things later.

py3-setuptools-git-versioning.yaml
smoser: huh. i did not realize jsonschema could enforce ordering. I opened https://github.com/chainguard-dev/yam/issues/71 a while ago, I think people would generally be open to computer-enforced formatting (ie, 'go fmt').

dannf: You're right - `jsonschema` doesn't enforce ordering (at least, not AFAIK). What I should've said is that I've written `pre-commit` hooks to enforce that YAML keys are sorted in the order that they are defined in a `jsonschema` file. I suppose the schema isn't strictly needed for the sorting, but schema validation is useful as well.

smoser (approved):

PR #33408, opened 2024-11-05
add runtime dependency on python3

joshrwolf:
:old-man-yells-at-hardcoded-shbangs:


joshrwolf:
going to address this in the image

PR #33418, opened 2024-11-06
Draft: First pass at automated ruby test

justinvreeland:


justinvreeland:
The pipeline in https://github.com/chainguard-dev/melange/pull/1483 is better than this one.  When that lands we can script pre like this one and resolve issues that come up.

PR #33493, opened 2024-11-06
gow: new package

blaine-arcjet:
This adds the [gow](https://github.com/mitranim/gow) command line tool as a package to the repository. We use this tool to run our Go servers in our development containers and it currently needs to be installed separately. Adding it to the wolfi repository would reduce our custom container configuration.

#### For new package PRs only

### DIFF
--- a/gow.yaml
+++ b/gow.yaml
@@ -1,0 +1,31 @@
+package:
+  name: gow
+  version: '20231006'
+  epoch: 0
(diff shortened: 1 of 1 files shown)


mamccorm:
Unfortunately, the project does not have a releasing strategy, and relies on pinning to main builds, which we avoid and mean we can't do auto-updates. Theres also not been any active commits since October 2023.

PR #33509, opened 2024-11-06
Add pkgconf test pipeline to some more packages

murraybd:
Now that https://github.com/wolfi-dev/os/pull/33417 has been merged these packages will all pass.

### DIFF
--- a/audit.yaml
+++ b/audit.yaml
@@ -96,6 +96,9 @@ subpackages:
         - audit
         - linux-headers
     description: audit dev
+    test:
+      pipeline:
+        - uses: test/pkgconf
 
   - name: audit-doc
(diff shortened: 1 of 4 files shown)


ajayk (approved):

xnox:
Hmmmm

xnox:
I would say audit package is buggy; doesn't pass our standard flags to ./configure; should be converted to use autoconf/configure pipeline, which should be setting prefix=/usr and then pkgconfig files should end up in /usr/, not /usr/local.

As otherwise end-users do not have anywhere to compile a custom patched audit and drop in on top. Which is a major usability issue; and also makes our packages & images not be path compatible with all other distros.

PR #33540, opened 2024-11-07
refactor: add libglvnd as dependency for mesa

wojciechka:


mesa.yaml
EyeCantCU, line 1: Needs an epoch bump

EyeCantCU (changes requested):
Mesa needs an epoch bump, otherwise changes look great!

mesa.yaml
wojciechka: Updated

EyeCantCU (approved):

PR #33564, opened 2024-11-07
dotnet-8: use a bootstrap package to decouple the requirement artifacts

ader1990:
To build dotnet in a more controlled way, I propose to have a bootstrap package and a main package that uses the bootstrap one.

The workflow:

  * build dotnet-bootstrap-8
  * build dotnet-source-8

The `dotnet-bootstrap-8` package bundles the artifacts from the `prep.sh` run. Those files consist of:

  * .dotnet folder - the `prep.sh` downloads and extracts the following https://dotnetcli.azureedge.net/dotnet/Sdk/8.0.110/dotnet-sdk-8.0.110-linux-x64.tar.gz. The size is ~564MB, License is MIT.
  * prereqs/packages/archive/Private.SourceBuilt.Artifacts.Bootstrap.tar.gz. This files is a bootstrap resulted from the download and extraction of https://dotnetcli.azureedge.net/source-built-artifacts/assets/Private.SourceBuilt.Artifacts.8.0.110-servicing.24474.1.centos.9-x64.tar.gz (size ~705MB). The bootstrap result is prereqs/packages/archive/Private.SourceBuilt.Artifacts.Bootstrap.tar.gz (~1.4GB). License - this is a very big list of .nupkg, need to check in-depth.

The `dotnet-source-8` package uses the bootstrap package and does not need the `prep.sh` anymore, as the `.dotnet` folder is symlinked to the bootstrapped `/usr/share`. Also the `prereqs/packages/archive/Private.SourceBuilt.Artifacts.Bootstrap.tar.gz` is symlinked.

The outcome is that the build of `dotnet-source-8` is done offline (excluding the actual git clone of the dotnet/dotnet repository).

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/dotnet-bootstrap-8.yaml
+++ b/dotnet-bootstrap-8.yaml
@@ -1,0 +1,61 @@
+package:
+  name: dotnet-bootstrap-8
+  version: 8.0.11
+  epoch: 0
+  description: ".NET 8 SDK Bootstrap"
+  copyright:
+    - license: MIT
+  resources:
+    cpu: 2
+    memory: 32Gi
+  dependencies:
+    runtime:
+      - icu
+
+environment:
+  contents:
+    packages:
+      - bash
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - curl
+      - glibc-locale-en
+      - icu-dev
+      - krb5-dev
+      - libunwind-dev
+      - lttng-ust-dev
+      - ncurses-dev
+      - openssl-dev
+      - wolfi-base
+      - zlib-dev
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/dotnet/dotnet
+      tag: v${{package.version}}
+      expected-commit: d5f3d652f9266d600777f626a9650a273419859b
+      destination: /home/build/src
+
+  - working-directory: /home/build/src
+    pipeline:
+      - runs: |
+          ./prep.sh
+      - runs: |
+          mkdir -p "${{targets.destdir}}"/usr/share/dotnet-bootstrap-8
+      - runs: |
+          # see https://github.com/jellyfin/jellyfin/pull/11063/
+          rm -f .dotnet/shared/Microsoft.NETCore.App/8.0.*/libcoreclrtraceptprovider.so
+          mv .dotnet "${{targets.destdir}}"/usr/share/dotnet-bootstrap-8/dotnet
+          mv prereqs/packages/archive/Private.SourceBuilt.Artifacts.Bootstrap.tar.gz "${{targets.destdir}}"/usr/share/dotnet-bootstrap-8/
+
+  - uses: strip
+
+update:
+  enabled: false
+  github:
+    identifier: dotnet/dotnet
+    strip-prefix: v
+    use-tag: true
+    tag-filter: "v8"
--- a/dotnet-8.yaml
+++ b/dotnet-8.yaml
@@ -1,7 +1,7 @@
 package:
   name: dotnet-8
   version: 8.0.11
-  epoch: 1
+  epoch: 2
   description: ".NET SDK"
   copyright:
     - license: MIT
@@ -23,6 +23,7 @@ environment:
       - clang-15-default
       - cmake
       - curl
+      - dotnet-bootstrap-8
       - glibc-locale-en
       - icu-dev
       - krb5-dev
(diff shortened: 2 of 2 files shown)


dotnet-source-8.yaml
```diff
@@ -0,0 +1,210 @@
+package:
+  name: dotnet-source-8
```
ader1990, line 2: An option would be to leave the current file `dotnet-8.yaml` to use the bootstrap package.

dotnet-bootstrap-8.yaml
```diff
@@ -0,0 +1,71 @@
...
+  - uses: strip
+
+update:
+  enabled: true
+  github:
+    identifier: dotnet/dotnet
+    strip-prefix: v
+    use-tag: true
+    tag-filter: "v8"
+
```
xnox, line 62: we could also keep update: enabled: false & have update-excuse: saying this is a bootstrap package.
as we typically do not want to upgrade bootstrap packages by default.

dotnet-bootstrap-8.yaml
```diff
@@ -0,0 +1,71 @@
...
+test:
+  environment:
+    contents:
+      packages:
+        - dotnet-8-sdk-bootstrap
+  pipeline:
+    - name: Basic .NET command test
+      runs: |
+        echo 1 # TBD
```
xnox, line 71: it's also sort of ok to not have tests for bootstrap; as it's sole purpose is ability to build the real package.

dotnet-source-8.yaml
xnox: Ack.

dotnet-source-8.yaml
```diff
@@ -0,0 +1,210 @@
...
+        - aspnet-8-targeting-pack
+
+  - name: dotnet-8-sdk-default
+    dependencies:
+      runtime:
+        - dotnet-8-sdk
+        - dotnet-8-runtime-default
+        - aspnet-8-runtime-default
+      provides:
+        - dotnet-sdk=8
```
xnox, line 160: How come this builds both runtime & sdk? Isn't SDK somewhere separate at a much higher version?

My expectation for this package to build runtime only, i.e. 8.0.10 like what everyone else has.

And then some other .yaml build would be doing sdk build to get us up to higher version numbers. as in 8.0.403

Where does .NET SDK 8.0.403 come from, the one that PowerShell releases claim to use?

ader1990: The https://github.com/dotnet/dotnet is a source based monolithic repository which aggregates all: sdk, runtime, aspnet. It is pinning the versions of the all 3 elements (and more) to have a unified working matrix.

The .NET SDK 8.0.403 comes from the https://github.com/dotnet/installer/tree/v8.0.403 / https://github.com/dotnet/sdk/tree/v8.0.403, and the PowerShell version uses that version independently of the dotnet/dotnet latest matrix.

The .NET SDK 8.0.403 has not been yet released (picked up) into the monolithic repository, and that is why the latest PowerShell version cannot be built with the SDK from the dotnet/dotnet repository.

From what I understand, it is just a matter of time until that version gets picked up in the monolith.

An option would be to not use the monolithic repository and try to mix and match versions of the subprojects, but the downside would be that there will be less testing (which is done in the dotnet/dotnet), thus less reliable composition.
We would have to do the work of the maintainers / CI of dotnet/dotnet repo to make sure that our matrix works well.

For example, the last commit of the SDK present in the dotnet/dotnet repository (branch release/8.0.1xx), was
https://github.com/dotnet/sdk/commit/d90806d013db39755cda355f42015d6a248456f5, equivalent to SDK version v8.0.110.

It is a matter of choice, depending on how much upstream testing is deemed good-enough.

xnox: THANKS! Let me ponder all that. I do think we need/want monolithic build for the bootstrap - aka known good point / LTS like thing. That makes all the things available.

And then it's likely on chainguard to build version-streamed non-monolithic sdk/aspnet/powershell etc.

dotnet-source-8.yaml
```diff
@@ -0,0 +1,211 @@
...
+          ln -sf /usr/share/dotnet-bootstrap/Private.SourceBuilt.Artifacts.Bootstrap.tar.gz prereqs/packages/archive/Private.SourceBuilt.Artifacts.Bootstrap.tar.gz
+      - runs: |
+          ./build.sh --clean-while-building \
+            -- \
+            /v:n \
+            /p:ContinueOnPrebuiltBaselineError=true \
+            /p:MinimalConsoleLogOutput=false \
+            /p:SkipPortableRuntimeBuild=true \
+            /p:BuildWithOnlineSources=false \
+            /p:Configuration=Debug
```
ader1990, line 67: to be reverted, added it here to have a way of build if anyone wants.

Size differences:

# release version of dotnet
du -sh /usr/share/dotnet-source/
481.2M  /usr/share/dotnet-source/

# debug version of dotnet
du -sh /usr/share/dotnet-source/
559.3M  /usr/share/dotnet-source/

xnox: do not want, too big.

dotnet-bootstrap-8.yaml
```diff
@@ -0,0 +1,71 @@
+package:
+  name: dotnet-bootstrap-8
+  version: 8.0.10
```
xnox, line 3: Please bump to .11 =) now that it is out.

xnox (approved):

PR #33634, opened 2024-11-08
add py3-ansible-core package

kranurag7:
add py3-ansible package

notes around why we are building with a patch:

        this support is required because we have latest resolvlib and that's
        supported on the upstream project devel branch too.
        the tag we are checking out is using older version of resolvlib and
        using that is leading to build failure unless we use the old version of
        resolvlib which we don't have packaged in wolfi.

### DIFF
--- a/py3-ansible-core.yaml
+++ b/py3-ansible-core.yaml
@@ -1,5 +1,5 @@
 package:
-  name: py3-ansible
+  name: py3-ansible-core
   version: 2.18.0
   epoch: 0
   description: Ansible is a radically simple IT automation platform
@@ -9,8 +9,7 @@ package:
     provider-priority: 0
 
 vars:
-  pypi-package: ansible
-  import: ansible
+  pypi-package: ansible-core
 
 data:
   - name: py-versions
@@ -38,8 +37,8 @@ subpackages:
     description: python${{range.key}} version of ${{vars.pypi-package}}
     dependencies:
       runtime:
-        - py3-jinja2
-        - py3-yaml
+        - py${{range.key}}-jinja2
+        - py${{range.key}}-pyyaml
       provider-priority: ${{range.value}}
       provides:
         - py3-${{vars.pypi-package}}
@@ -58,8 +57,8 @@ test:
   environment:
     contents:
       packages:
-        - py3-yaml
         - py3-jinja2
+        - py3-pyyaml
   pipeline:
     - name: version tests
       runs: |
--- a/py3-ansible-core.yaml
+++ b/py3-ansible-core.yaml
@@ -30,6 +30,10 @@ pipeline:
       repository: https://github.com/ansible/ansible
       tag: v${{package.version}}
 
+  - uses: patch
+    with:
+      patches: resolvlib.patch
+
 subpackages:
   - range: py-versions
     name: py${{range.key}}-${{vars.pypi-package}}
--- a/py3-ansible-core.yaml
+++ b/py3-ansible-core.yaml
@@ -36,8 +36,11 @@ subpackages:
     description: python${{range.key}} version of ${{vars.pypi-package}}
     dependencies:
       runtime:
+        - py${{range.key}}-cryptography
         - py${{range.key}}-jinja2
+        - py${{range.key}}-packaging
         - py${{range.key}}-pyyaml
+        - py${{range.key}}-resolvelib
       provider-priority: ${{range.value}}
       provides:
         - py3-${{vars.pypi-package}}
--- a/py3-ansible-core.yaml
+++ b/py3-ansible-core.yaml
@@ -36,6 +36,7 @@ subpackages:
     description: python${{range.key}} version of ${{vars.pypi-package}}
     dependencies:
       runtime:
+        - openssl # requires because of py3-cryptography
         - py${{range.key}}-cryptography
         - py${{range.key}}-jinja2
         - py${{range.key}}-packaging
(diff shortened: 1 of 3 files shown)


py3-ansible.yaml
```diff
@@ -0,0 +1,75 @@
...
+  github:
+    identifier: ansible/ansible
+    strip-prefix: v
+
+test:
+  environment:
+    contents:
+      packages:
+        - py3-yaml
+        - py3-jinja2
```
EyeCantCU, line 62: No need to redefine package runtime dependencies

py3-ansible.yaml
```diff
@@ -0,0 +1,75 @@
...
+        uses: python/build-wheel
+      - uses: strip
+
+update:
+  enabled: true
+  github:
+    identifier: ansible/ansible
+    strip-prefix: v
+
+test:
```
EyeCantCU, line 57: We should add an import test as well. See other versioned Python packages as examples

py3-ansible.yaml
```diff
@@ -0,0 +1,75 @@
...
+  version: 2.18.0
+  epoch: 0
+  description: Ansible is a radically simple IT automation platform
+  copyright:
+    - license: GPL-3.0
+  dependencies:
+    provider-priority: 0
+    runtime:
+      - py3-jinja2
+      - py3-yaml
```
EyeCantCU: No need to define here, the empty main package will never be used

py3-ansible.yaml
```diff
@@ -0,0 +1,75 @@
...
+    dependencies:
+      runtime:
+        - py3-jinja2
+        - py3-yaml
+      provider-priority: ${{range.value}}
+      provides:
+        - py3-${{vars.pypi-package}}
+    pipeline:
+      - name: Python Build
+        uses: python/build-wheel
```
EyeCantCU, line 48: Should probably be `py/pip-build-install`

py3-ansible.yaml
```diff
@@ -0,0 +1,75 @@
...
+      repository: https://github.com/ansible/ansible
+      tag: v${{package.version}}
+
+subpackages:
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}
+    description: python${{range.key}} version of ${{vars.pypi-package}}
+    dependencies:
+      runtime:
+        - py3-jinja2
```
EyeCantCU, line 41: Should always be `py${{range.key}}-*` for versioned deps

EyeCantCU (dismissed):

py3-ansible.yaml
kranurag7: RJ, when I try to use this pipeline then I'm not able to build for python-3.10.

```bash
2024/11/08 12:47:49 WARN   creating '/tmp/pip-modern-metadata-cbsh28yi/ansible_core-2.18.0.dist-info' subpackage=py3.10-ansible-core uses=py/pip-build-install name="pip build"
2024/11/08 12:47:49 INFO   Preparing metadata (pyproject.toml): finished with status 'done' subpackage=py3.10-ansible-core uses=py/pip-build-install name="pip build"
2024/11/08 12:47:49 INFO INFO: pip is looking at multiple versions of ansible-core to determine which version is compatible with other requirements. This could take a while. subpackage=py3.10-ansible-core uses=py/pip-build-install name="pip build"
2024/11/08 12:47:49 WARN ERROR: Package 'ansible-core' requires a different Python: 3.10.15 not in '>=3.11' subpackage=py3.10-ansible-core uses=py/pip-build-install name="pip build"
```

EyeCantCU: We can't build with 3.10 as it needs at least 3.11 so let's remove 3.10

py3-ansible-core.yaml
```diff
@@ -0,0 +1,74 @@
...
+  github:
+    identifier: ansible/ansible
+    strip-prefix: v
+
+test:
+  environment:
+    contents:
+      packages:
+        - py3-jinja2
+        - py3-pyyaml
```
EyeCantCU, line 61: We can drop this

py3-ansible-core/resolvlib.patch
```diff
@@ -0,0 +1,166 @@
...
+ create mode 100644 changelogs/fragments/update-resolvelib-lt-1_2_0.yml
+
+diff --git a/changelogs/fragments/update-resolvelib-lt-1_2_0.yml b/changelogs/fragments/update-resolvelib-lt-1_2_0.yml
+new file mode 100644
+index 00000000000000..461f436339b4c5
+--- /dev/null
++++ b/changelogs/fragments/update-resolvelib-lt-1_2_0.yml
+@@ -0,0 +1,2 @@
++minor_changes:
++  - ansible-galaxy - support ``resolvelib >= 0.5.3, < 1.2.0`` (https://github.com/ansible/ansible/issues/84217).
```
maxgio92, line 27: Because I see that it's been patched to `2.0.0` in the upstream and [merged](https://github.com/ansible/ansible/pull/84218/files#diff-065abed58d290df207524c494d40352b8dc2fb11ecb86b3e58553c9a3888656fR15) to the `devel` branch,
even if not setting the update to manual, maybe we should highlight this in case the patch will no longer apply in the next release?

kranurag7: I think this will be removed as part of automation in the next update.

maxgio92 (approved):
LGTM

maxgio92:
I think that the filed low CVEs should be marked as false positive. Only `/usr/lib/python3.12/site-packages/ansible_core-2.18.0.dist-info/METADATA` file is there once packages are installed and it's a text file.

PR #33641, opened 2024-11-08
feat: add tileserver-gl package

wojciechka:
#### For new package PRs only


tileserver-gl.yaml
```diff
@@ -0,0 +1,168 @@
...
+      - python-3.11
+      - xorg-server-dev
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/maptiler/tileserver-gl
+      tag: v${{package.version}}
+      expected-commit: 6cda7a0b384940b781838be7415b2ae448edcd29
+      destination: app
```
developer-guy, line 55: instead of doing that we have `working-directory:` option where you can use during the pipeline steps, it helps you to avoid using cd app/ in every step, also it makes your build config lot easier to read and follow in my humble opinion, I don't want to be a pedantic here but you can look for other examples which uses working-directory option, it looks a bit lean to me, does that make sense?

wojciechka: Thanks. Updated.

developer-guy (approved):

PR #33756, opened 2024-11-10
feat: Add runtimes for LLVM 19

EyeCantCU:
Adds runtimes for LLVM 19

### DIFF
--- a/llvm-runtimes-19.yaml
+++ b/llvm-runtimes-19.yaml
@@ -1,0 +1,186 @@
+package:
+  name: llvm-runtimes-19
+  version: 19.1.3
+  epoch: 0
+  description: LLVM runtimes
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    provides:
+      - llvm-runtimes=${{package.full-version}}
+
+var-transforms:
+  - from: ${{package.version}}
+    match: ^(\d+).*
+    replace: $1
+    to: major-version
+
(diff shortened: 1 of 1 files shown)


EyeCantCU:
Retriggering elastic

xnox (changes requested):
cc: @justinvreeland 

Previously these three components were packaged into these two source packages, which we seem to be missing for 18 & 19 llvm projects

```
$ ls | grep -e llvm-libunwind -e llvm-libcxx
llvm-libcxx-15.yaml
llvm-libcxx-16.yaml
llvm-libcxx-17.yaml
llvm-libunwind-15.yaml
llvm-libunwind-16.yaml
llvm-libunwind-17.yaml
```

Could we introduce above two yamls for 18 & 19 please? Then all of them will be the same.

EyeCantCU:
@xnox would we rather do that than move existing packages over to this format? I think it makes more sense to build the runtimes together

xnox:
> @xnox would we rather do that than move existing packages over to this format? I think it makes more sense to build the runtimes together

If we collapse things, just collapse everything.

A single llvm-19.yaml should be able to build llvm,clang,lld,CMake,tools, all libraries, etc. No other distribution builds them in stages. Just one build of everything in one go.

EyeCantCU:
Sounds good to me. I'll consolidate all of it

PR #33842, opened 2024-11-11
openssl: cherry-pick Jitter entropy source compat for old providers

xnox:
Cherry-pick patch from a draft PR to ensure old providers with new
openssl use jitter entropy. No effect on any v3.2+ providers as all of
them use runtime configured callbacks instead.

Add test case that verifies with gdb and internal function name
knowledge, that indeed syscall/os-entropy is used as before with
default openssl provider.


smoser (approved):
this is great. thank you for the gdb test.

PR #33860, opened 2024-11-11
Adding a test use case for creating

utieyin:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches


test-spark.yaml
```diff
@@ -0,0 +1,121 @@
+package:
+  name: sparkr
+  version: 3.5.3
+  epoch: 1
```
Dentrax, line 4:
```suggestion
  epoch: 0
```

test-spark.yaml
```diff
@@ -0,0 +1,121 @@
+package:
+  name: sparkr
```
Dentrax, line 2: Should we add version suffix here?

test-spark.yaml
```diff
@@ -0,0 +1,121 @@
...
+  version: 3.5.3
+  epoch: 1
+  description: Unified engine for large-scale data analytics
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    provides:
+      - spark=${{package.full-version}}
+      - spark-3.5=${{package.full-version}}
+      - spark-3.5-scala-2.13=${{package.full-version}}
```
Dentrax, line 12: Those are doesn't match with actual package name.

test-spark.yaml
```diff
@@ -0,0 +1,121 @@
...
+    runtime:
+      - openjdk-17
```
Dentrax, line 14: Is this project supports multiple openjdk versions? If so, we should build for all supported JDKs, example here: https://github.com/wolfi-dev/os/blob/cb8b130b6739d596fab0688fb200a24123743e7f/spark-3.5.yaml#L49-L54

spark-3.5.yaml
```diff
@@ -120,7 +131,7 @@ subpackages:
                 val sum = rdd.reduce(_ + _)
                 assert(sum == 15)
                 EOF
-                cat SimpleJob.scala | /usr/lib/spark/bin/spark-shell
+                cat SimpleJob.scala | /usr/lib/spark/bin/spark-shell --conf spark.jars.ivy=/tmp/.ivy
```
EyeCantCU, line 134: I'm not sure I understand why we need pass --conf

spark-3.5.yaml
```diff
@@ -224,6 +245,94 @@ subpackages:
           ln -s /usr/lib/spark/sbin ${{targets.subpkgdir}}/opt/bitnami/spark/sbin
           ln -s /usr/lib/spark/yarn ${{targets.subpkgdir}}/opt/bitnami/spark/yarn
 
+  - name: ${{package.name}}-scala-2.13
+    dependencies:
+      runtime:
+        - openjdk-17-default-jvm
```
EyeCantCU, line 251: We build with 8 and 17. We should drop this

spark-3.5.yaml
```diff
@@ -224,6 +245,94 @@ subpackages:
...
+          cp pom.xml.orig pom.xml
+          mvn clean
+          ./dev/change-scala-version.sh 2.13
+          ./build/mvn -DskipTests -Pscala-2.13 clean package
+          mkdir -p ${{targets.contextdir}}/usr/lib/spark
+          mkdir -p ${{targets.contextdir}}/usr/lib/spark/work-dir
+          mv bin/ ${{targets.contextdir}}/usr/lib/spark
+          mv sbin/ ${{targets.contextdir}}/usr/lib/spark
+          mv target ${{targets.contextdir}}/usr/lib/spark
+          mv assembly ${{targets.contextdir}}/usr/lib/spark
```
EyeCantCU, line 267: Why does this deviate from the build process for 2.12 above?

spark-3.5.yaml
```diff
@@ -224,6 +245,94 @@ subpackages:
...
+          cp resource-managers/kubernetes/docker/src/main/dockerfiles/spark/entrypoint.sh ${{targets.contextdir}}/usr/lib/spark/
+
+  - name: ${{package.name}}-scala-2.13-compat
```
EyeCantCU, line 270: Why do we need another compat package?

spark-3.5.yaml
```diff
@@ -224,6 +245,94 @@ subpackages:
...
+        - bash
+        - ${{package.name}}-scala-2.13
+    environment:
+      LANG: en_US.UTF-8
+      SCALA_VERSION: 2.13
+      JAVA_HOME: /usr/lib/jvm/java-17-openjdk
+      SPARK_LOCAL_IP: 127.0.0.1
+      SPARK_LOCAL_HOSTNAME: localhost
+      SPARK_HOME: /usr/lib/spark
+      PATH: $PATH:${SPARK_HOME}/bin
```
EyeCantCU, line 297: This doesn't work, use an export

EyeCantCU (changes requested):
Great effort. It looks like we're recreating packages that contain the same logic. Ideally, we'd want to use a range with opt-in logic when the version is 2.13. Appears as though this breaks the package for OpenJDK 8 as well

spark-3.5.yaml
utieyin: @EyeCantCU The `entrypoint.sh` is sourced from a different location for `2.12` which I suspect is as a result of the build process.

EyeCantCU: If we loop over a range of supported configurations (scala 2.12-2.13, jdk8-17), we can accommodate that in the same subpackage definition without creating a new one with much of the same logic

spark-3.5.yaml
```diff
@@ -21,7 +21,7 @@ environment:
       - maven
       - openjdk-11
       - openjdk-17
-      # Only 8 is used during the build process
+      - openjdk-17-default-jvm
```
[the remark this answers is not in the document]
utieyin: Removed

spark-3.5.yaml
utieyin: @EyeCantCU Ranging over two different data items here might be difficult or impossible from my understanding. Is nested ranging supported?

spark-3.5.yaml
utieyin: @EyeCantCU because of this error `2024/11/19 00:35:51 WARN Exception in thread "main" java.lang.IllegalArgumentException: basedir must be absolute: ?/.ivy2/local name="Test spark-3.5 with OpenJDK 11" name="Run a simple Scala test script"`

spark-3.5.yaml
utieyin: Replaced this with a range and conditional build

spark-3.5.yaml
utieyin: Removed

spark-3.5.yaml
utieyin: @EyeCantCU scala-2.13 expects to build with java-17

spark-3.5.yaml
```diff
@@ -96,9 +96,20 @@ subpackages:
             - python3
             - openjdk-${{range.key}}-default-jvm
             - R
+            - bash
+            - wolfi-base
+            - libstdc++-11
+            - libavcodec61
```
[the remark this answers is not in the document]
utieyin: Removed

spark-3.5-scala-2.13.yaml
EyeCantCU, line 1: We should probably rename the existing `spark-3.5` to `spark-3.5-scala-2.12` as well for alignment

spark-3.5-scala-2.13.yaml
```diff
@@ -0,0 +1,145 @@
...
+      repository: https://github.com/apache/spark
+      tag: v${{package.version}}
+      expected-commit: 32232e9ed33bb16b93ad58cfde8b82e0f07c0970
+
+  - runs: |
+      export JAVA_HOME=/usr/lib/jvm/java-17-openjdk
+      export MAVEN_OPTS="-Xss64m -Xmx2g -XX:ReservedCodeCacheSize=1g"
+      ## pom bump seem to mess with the ability to switch scala version
+      ## the solution is to restore the original pom file at the package build stage
+      mvn clean
```
EyeCantCU, line 52: We shouldn't need to do this anymore

spark-3.5-scala-2.13.yaml
```diff
@@ -0,0 +1,145 @@
...
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/apache/spark
+      tag: v${{package.version}}
+      expected-commit: 32232e9ed33bb16b93ad58cfde8b82e0f07c0970
+
+  - runs: |
+      export JAVA_HOME=/usr/lib/jvm/java-17-openjdk
+      export MAVEN_OPTS="-Xss64m -Xmx2g -XX:ReservedCodeCacheSize=1g"
```
EyeCantCU, line 49: These can go in the environment, since you install `-default-jdk` though, setting JAVA_HOME shouldn't be necessary

spark-3.5-scala-2.13.yaml
utieyin: Yep, removed

spark-3.5-scala-2.13.yaml
utieyin: Moved to environment

EyeCantCU (approved):

spark-3.5-scala-2.13.yaml
mamccorm: Nice work. Whats the process going forward when a new scala version is released? Will this require a manual effort each time to create separate spark scala packages? If so we should create a runbook with some pointers for sustaining here:
- https://eng.inky.wtf/docs/squads/sustaining/playbooks/

utieyin: It was agreed that we rename the spark-3.5 package after closing this issue.
@mamccorm I think spark going forward (v4) builds exclusively on scala-2.13+ as shown in the version change script here https://github.com/apache/spark/blob/0cb7a42731076b9960eb9ad27067cab7ae570356/dev/change-scala-version.sh#L40. So building newer versions of spark will likely move this into version streamed territory with only a single version of scala at least until apache decides to change its mind. If they do, we will have to rebuild v4 with the specific version of scala

PR #33888, opened 2024-11-12
refactor(xorg-server): use mesa built against glvnd

wojciechka:


jamonation (dismissed):
Discussed this and since `mesa-glx` is now built with glvnd, looks good to merge

jamonation (approved):

PR #33890, opened 2024-11-12
refactor(libreoffice-24.8): update to mesa with glvnd

wojciechka:


libreoffice-24.8.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: libreoffice-24.8
-  version: 24.8.3.1
+  version: 24.8.3.2
```
rawlingsj, line 3: this needs to remain the same, when making changes to the melange yaml we need to bump the epoch below.

PR #33946, opened 2024-11-12
add py3-requests-unixsocket package

kranurag7:
dependency for ansible-runner-http package. 

using `py/pip-build-install` pipeline was not possible given the project is still using setup.py and haven't made the transition to using pyproject.toml as of now.

### DIFF
--- a/py3-requests-unixsocket.yaml
+++ b/py3-requests-unixsocket.yaml
@@ -6,16 +6,30 @@ package:
   copyright:
     - license: Apache-2.0
   dependencies:
-    runtime:
-      - py3-requests
+    provider-priority: 0
 
 environment:
   contents:
     packages:
+      - build-base
       - busybox
-      - py3-pip
-      - py3-setuptools
-      - python3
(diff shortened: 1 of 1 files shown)


kranurag7:
Notes: 
we are not using multi-version python here. I really want to use that but `py/pip-build-install` pipeline breaks. Upstream project doesn't use `pyproject.toml` as of now. 

The disadvantage of not doing multi-version will mean that all the package that'll pull this as runtime dependency will not have multi-version python either. :( 

This is what I found out during my building/testing cycle; reviews/suggestion appreciated if there's a workaround for this. 

I can try using `range` and doing exactly similar package naming to what our `py/pip-build-install` pipeline does but not sure if that's a good way to go about it.

pnasrat:
@kranurag7 please see the updated version that is using py3-versioned fine with this project. I'm not sure what your issue was but it was not the pipeline or lack of pyproject.toml. If you have specific errors you encountered we can discuss

developer-guy (approved):

PR #33969, opened 2024-11-12
add lsof package

kranurag7:
adds lsof package

⚠️ licensing concerns:

This is not OSI Approved license. (ref: https://spdx.org/licenses/)
Wondering if are safe to land this in wolfi or not. 
The project also says that there's no license of the project (ref: https://github.com/lsof-org/lsof/blob/master/docs/faq.md#is-there-an-lsof-license)

### DIFF
--- a/lsof.yaml
+++ b/lsof.yaml
@@ -27,6 +27,22 @@ pipeline:
       tag: ${{package.version}}
       expected-commit: 2e4c7a1a9bc7258dc5b6a3ab28ebca44174279a8
 
+  - name: "update release version in configure.ac"
+    runs: |
+      # configure.ac in release tarballs has updated major.minor.micro (4.99.3)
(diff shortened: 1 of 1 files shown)


lsof.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+update:
+  enabled: true
+  github:
+    identifier: lsof-org/lsof
+
+test:
+  pipeline:
+    - name: version test
+      runs: /usr/bin/lsof -v
```
kranurag7, line 70: One thing that's not working as expected for now is that, this version is wrong. It returns 4.99.0 whereas we are building for 4.99.3

smoser:
@kranurag7 thanks.
I've made a PR to your branch https://github.com/kranurag7/os/pull/1/files .
You can merge that in and it addresses the 4.99.3

smoser (approved):

PR #33992, opened 2024-11-13
libsdl2-ttf: add libsdl2-ttf package

ajayk:
freedrp-3 needs libsdl2-ttf packages 

required to fix https://github.com/wolfi-dev/os/pull/31386


xnox (approved):

xnox:
I was puzzled before where this comes from. Didn't bother to check or find that it is, in fact a separate library.

PR #34028, opened 2024-11-13
add istio ztunnel package

kranurag7:
adds istio ztunnel package

### DIFF
--- a/ztunnel-1.24.yaml
+++ b/ztunnel-1.24.yaml
@@ -10,14 +10,11 @@ package:
       - ztunnel=${{package.full-version}}
     runtime:
       - ca-certificates-bundle
-      - glibc
       - ip6tables
       - iptables
-      - libgcc
       - libmnl
       - libnetfilter_conntrack
       - libnfnetlink
-      - wolfi-baselayout
 
 environment:
   contents:
@@ -44,9 +41,9 @@ pipeline:
       export ZTUNNEL_BUILD_buildGitRevision=$(git rev-parse HEAD)
       export ZTUNNEL_BUILD_buildTag=$(git describe --tags --abbrev=0)
       export ZTUNNEL_BUILD_buildVersion=$(git rev-parse HEAD)
-       cargo auditable build --release
-       mkdir -p ${{targets.destdir}}/usr/bin
-       cp out/rust/release/ztunnel ${{targets.destdir}}/usr/bin/ztunnel
+      cargo auditable build --release
+      mkdir -p ${{targets.destdir}}/usr/bin
+      cp out/rust/release/ztunnel ${{targets.destdir}}/usr/bin/ztunnel
 
   - uses: strip
 
--- a/ztunnel-1.24.yaml
+++ b/ztunnel-1.24.yaml
@@ -57,6 +57,9 @@ subpackages:
       - runs: |
           mkdir -p "${{targets.contextdir}}"/usr/local/bin
           ln -sf /usr/bin/ztunnel "${{targets.contextdir}}"/usr/local/bin/ztunnel
+    dependencies:
+      provides:
+        - ztunnel-compat=${{package.full-version}}
 
 update:
   enabled: true
--- a/ztunnel-1.24.yaml
+++ b/ztunnel-1.24.yaml
@@ -1,0 +1,67 @@
+package:
+  name: ztunnel-1.24
+  version: 1.24.0
+  epoch: 0
+  description: The `ztunnel` component of istio ambient mesh.
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    provides:
+      - ztunnel=${{package.full-version}}
+    runtime:
+      - ca-certificates-bundle
+      - glibc
+      - ip6tables
+      - iptables
+      - libgcc
+      - libmnl
+      - libnetfilter_conntrack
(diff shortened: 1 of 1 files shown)


ztunnel-1.24.yaml
```diff
@@ -0,0 +1,75 @@
...
+      tag: ${{package.version}}
+      expected-commit: 1226c1b35f50938f428c71f7dcad3602ea991675
+
+  - runs: |
+      export ZTUNNEL_BUILD_buildStatus=clean
+      export ZTUNNEL_BUILD_buildGitRevision=$(git rev-parse HEAD)
+      export ZTUNNEL_BUILD_buildTag=$(git describe --tags --abbrev=0)
+      export ZTUNNEL_BUILD_buildVersion=$(git rev-parse HEAD)
+       cargo auditable build --release
+       mkdir -p ${{targets.destdir}}/usr/local/bin
```
kranurag7, line 48:
```suggestion
       mkdir -p ${{targets.destdir}}/usr/bin
```

developer-guy (dismissed):

ztunnel-1.24.yaml
```diff
@@ -0,0 +1,78 @@
...
+  epoch: 0
+  description: The `ztunnel` component of istio ambient mesh.
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    provides:
+      - ztunnel=${{package.full-version}}
+    runtime:
+      - ca-certificates-bundle
+      - glibc
```
EyeCantCU, line 13: We shouldn't need to explicitly include this

ztunnel-1.24.yaml
```diff
@@ -0,0 +1,78 @@
...
+      - ip6tables
+      - iptables
+      - libgcc
+      - libmnl
```
EyeCantCU, line 17: I'd also think these would be transient and added via SCA

ztunnel-1.24.yaml
```diff
@@ -0,0 +1,78 @@
...
+      - libnetfilter_conntrack
+      - libnfnetlink
+      - wolfi-baselayout
```
EyeCantCU, line 20: Images should already provide the baselayout. Why do we need this?

ztunnel-1.24.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - cargo-auditable
```
EyeCantCU, line 28: OUr pipeline doesn't work here?

ztunnel-1.24.yaml
```diff
@@ -0,0 +1,78 @@
...
+      expected-commit: 1226c1b35f50938f428c71f7dcad3602ea991675
+
+  - runs: |
+      export ZTUNNEL_BUILD_buildStatus=clean
+      export ZTUNNEL_BUILD_buildGitRevision=$(git rev-parse HEAD)
+      export ZTUNNEL_BUILD_buildTag=$(git describe --tags --abbrev=0)
+      export ZTUNNEL_BUILD_buildVersion=$(git rev-parse HEAD)
+       cargo auditable build --release
+       mkdir -p ${{targets.destdir}}/usr/bin
+       cp out/rust/release/ztunnel ${{targets.destdir}}/usr/bin/ztunnel
```
EyeCantCU, line 49: Looks like there's additional indentation here

ztunnel-1.24.yaml
```diff
@@ -0,0 +1,78 @@
...
+  enabled: true
+  ignore-regex-patterns:
+    - '-rc'
+    - '-beta'
+  github:
+    identifier: istio/ztunnel
+    tag-filter-prefix: 1.24.
+    use-tag: true
+
+test:
```
EyeCantCU, line 74: Anything more we can do in tests?

kranurag7: the application tries to create `/var/run/ztunnel/ztunnel.sock` and fails. I was not able to add anything more in package level tests here because of same.

ztunnel-1.24.yaml
kranurag7: Unfortunately no, the default build output goes in `out` directory instead of `target/releases` that our pipeline uses.

ztunnel-1.24.yaml
```diff
@@ -0,0 +1,75 @@
...
+      export ZTUNNEL_BUILD_buildTag=$(git describe --tags --abbrev=0)
+      export ZTUNNEL_BUILD_buildVersion=$(git rev-parse HEAD)
+      cargo auditable build --release
+      mkdir -p ${{targets.destdir}}/usr/bin
+      cp out/rust/release/ztunnel ${{targets.destdir}}/usr/bin/ztunnel
+
+  - uses: strip
+
+subpackages:
+  - name: ${{package.name}}-compat
```
Dentrax, line 51: Should this be instead the following and the omit `dependencies` section entirely?
```suggestion
  - name: ztunnel-compat
```

kranurag7: I think this is going to be the default so this will be very much a version stream package as starting 1.25 version of istio so the current version stream format will be more future proof IMHO. 💭

EyeCantCU (approved):

PR #34092, opened 2024-11-13
Change default value for go telemetry from 'local' to 'off'

smoser:
This will set the default value for go telemetry to 'off' rather than 'local'.

Go 1.23 introduced a telemetry feature that collects local audit data about the Go toolchain, storing it by default in
$HOME/.config/go/telemetry. While this data is not sent externally by default, the local storage path can trigger security alerts in tools like Falco, as it writes to a sensitive location under /root.

The behavior can be disabled with 'go telemetry off', which writes to the config file above, but that means the user needs to do so before calling 'go' in any other manner.  Doing so for a container is non-obvious.  We could build /root/.config/go/telemetry into a 'go' image, but that would still provide a problem for any user other than uid 0.

There is no mechanism to change the behavior "system wide" or an environment variable that can set the value.

See https://github.com/golang/go/issues/68960 and
https://github.com/golang/go/issues/69113. The second one requests that env GOTELEMETRY=off would disable telemetry. That would be easy for us to utilize but it was rejected upstream.

Instead, we just change the default value returned if there is no .config/go/telemetry file present.

### DIFF
--- a/go-1.23.yaml
+++ b/go-1.23.yaml
@@ -1,7 +1,7 @@
 package:
   name: go-1.23
   version: 1.23.3
-  epoch: 0
+  epoch: 1
   description: "the Go programming language"
   copyright:
     - license: BSD-3-Clause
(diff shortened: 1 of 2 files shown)


go-1.23.yaml
```diff
@@ -146,3 +148,41 @@ test:
 
         # Run the Go program with cgo and check the output
         go run hello_cgo.go | grep "Hello from cgo!"
+    - name: Test telemetry settings
```
vaikas, line 151: ❤️

xnox (approved):

Issue #34075, opened 2024-11-13
abseil-cpp and rdeps FTBFS


dannf:
Since the introduction of `openssf-compiler-options`, `abseil-cpp` FTBFS. This [appears to impact](https://console.cloud.google.com/logs/query;query=logName%3Dprojects%2Fprod-wolfi-os%2Flogs%2Fstderr%0Aresource.type%3D%22k8s_container%22%0Aresource.labels.namespace_name%3D%22world-wolfi%22%0A%22is%20not%20a%20constant%20expression%22;summaryFields=labels%252F%2522k8s-pod%252Fapp_kubernetes_io%252Fcomponent%2522,labels%252F%2522k8s-pod%252Fmelange_chainguard_dev%252Farch%2522,labels%252F%2522k8s-pod%252Fmelange_chainguard_dev%252Fbuild-id%2522:false:32:beginning;cursorTimestamp=2024-11-09T08:45:49.249694420Z;startTime=2024-10-01T22:39:49.296Z;endTime=2024-11-12T23:39:49.296Z?project=prod-wolfi-os&rapt=AEjHL4ORwE0hBTJtgqW8rLuTsoooP9B2539dayjiy33GHQyDxBeCVwORKexsKjqdxXFunKK88tA2NAHZJmuODVRyMLIEYIittJny6Y1092Iay4MYrqwSGoU) the following packages:

- abseil-cpp
- grpc-1.66
- grpc-1.67
- libreoffice-24.8
- nodejs-22
- nodejs-23
- protobuf
- protobuf-c
- py3-grpcio-tools
- re2

As @justinvreeland [surmised](https://github.com/wolfi-dev/os/pull/31754#issuecomment-2438710205), this is believed to be a [GCC bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71962). Upstream's last reply on that bug was over a year ago, so it seems we need to do something else.

So what do we do in the meantime?

While the current LTS version of `abseil-cpp` - `20240722` is impacted, the previous LTS version - `20240116` appears not to be. `abseil-cpp` upstream [supports LTS releases for at least 2 years](https://abseil.io/about/releases#long-term-support-lts-branches), so multi-versioning that package seems like a way to unblock rdeps that are compatible with it.

[...]

PR #34108, opened 2024-11-13
Workaround GCC issue to fix FTBFS with abseil-cpp and reverse deps

dannf:
Related: #34075 https://github.com/chainguard-dev/internal-dev/issues/5334

This is part of the `py3.13-google-cloud-spanner` dependency chain.

### DIFF
--- a/abseil-cpp.yaml
+++ b/abseil-cpp.yaml
@@ -1,7 +1,7 @@
 package:
   name: abseil-cpp
   version: "20240722.0"
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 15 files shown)


dannf:
Moving to draft status. I'm unable to build `py3-grpcio-gcp` against it, likely because using the vendored `protobuf` is too old. I'm working on fixing our `abseil-cpp` FTBFS now instead of working around it.

dannf:
This has changed from its original intent and is now superseded by https://github.com/wolfi-dev/os/pull/34201

PR #34130, opened 2024-11-14
refactor(kubernetes): remove setcap

wojciechka:
Removing setcap as this prevents the build from working in bubblewrap as non-root.

The `kube-apiserver` is running as uid 0 and setcap should not be needed.

### DIFF
--- a/kubernetes-1.31.yaml
+++ b/kubernetes-1.31.yaml
@@ -1,7 +1,7 @@
 package:
   name: kubernetes-1.31
   version: 1.31.2
-  epoch: 1
+  epoch: 2
   description: Production-Grade Container Scheduling and Management
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


kubernetes-1.31.yaml
```diff
@@ -62,11 +62,6 @@ pipeline:
 
       make WHAT="$WHAT"
 
-  - runs: |
```
maxgio92, line 65: This may introduce a regression for unprivileged deployments that relied on the capability on the binary file to be able to run on the host network and bind on the usual privileged port 6443.

Nevertheless, I think it's a good refactoring because this way we are giving least privilege by default.
For cases where the API server will be exposed on unprivileged ports (>1024), this capability isn't needed at all.
In cases where it is, the capability can be added at runtime at thread level.

maxgio92: But I think we should test these scenarios before introducing this change.

wojciechka: I checked and the extended attribute (`security.capabilities`) is not persisted, so this step is not changing anything in the output apk file.

I am confident it is safe to remove it from the `kubernetes` package.

maxgio92: Good catch! That's unexpected, I'd say

wojciechka:
No longer needed

PR #34145, opened 2024-11-14
fix(py3-conda-libmamba-solver): Cherrypick upstream commits to add support for libmamba 2.0

philroche:
We have already released packages updates to libmamba and subpackages which is a major version jump from 1.5 to 2.0.

This caused failures in the conda packages and images due to py3-conda-libmamba-solver not yet supporting libmamba 2.0.

There is not yet an upstream release of conda-libmamba-solver with mamba 2.0 support but there is a release candidate
branch @ https://github.com/conda/conda-libmamba-solver/tree/24.11.0rc awaiting release.

We have broken dependent packages so we cannot wait for upstream to release.

As such I have cherrypicked the relevant 2.0 support commits from the 24.11.0rc branch.

This resolves the issues seen with conda and libmamba.

Full diff for conda-libmamba-solver @  https://github.com/conda/conda-libmamba-solver/compare/24.9.0...24.11.0rc

https://github.com/mamba-org/mamba/issues/3506 was the most helpful in finding the work to support libmamba 2.0

### DIFF
--- a/py3-conda-libmamba-solver.yaml
+++ b/py3-conda-libmamba-solver.yaml
@@ -1,7 +1,7 @@
 package:
   name: py3-conda-libmamba-solver
   version: 24.9.0
-  epoch: 0
+  epoch: 1
   description: The libmamba based solver for conda.
   copyright:
     - license: BSD-3-Clause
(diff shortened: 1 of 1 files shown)


jamonation (approved):
Discussed this with @philroche and we decided that while these are rc cherry picks, they unbreak existing use of the package, so +1 to merge

PR #34201, opened 2024-11-15
Add abseil-cpp spec to workaround GCC issue w/ openssf-compiler-options

### LINKED ISSUE #34075: abseil-cpp and rdeps FTBFS

dannf:
Fixes: #34075

Also merges a few automated package update PRs that were blocked by this.

Blocked by https://github.com/chainguard-dev/melange/issues/1651


justinvreeland (dismissed):

dannf:
The elastic build of libreoffice failed because it didn't actually have a build-dep on `abseil-cpp-dev`, and therefore couldn't find the spec file. I had pre-installed a locally built `abseil-cpp-dev` when I tested locally, so it worked for me. `libreoffice` vendors a copy of `abseil-cpp`, but does support a `--use-system-abseil` configure flag. I've turned that on and force pushed.

dannf:
I've pushed a new version that puts the spec files in a new `abseil-cpp-compiler-options` package instead of bundling them with `abseil-cpp-dev`. This reduces the dependency chain on packages that vendor in their own `abseil-cpp` source.

grpc-1.66.yaml
```diff
@@ -82,7 +86,7 @@ pipeline:
       	-DgRPC_BUILD_TESTS=OFF
       cmake --build _build
 
-      GRPC_PYTHON_CFLAGS="-std=c++17" \
+      GRPC_PYTHON_CFLAGS="-std=c++17 -specs abseil-cpp.spec" \
```
justinvreeland, line 89: Does GCC_SPEC_FILE as mentioned in the announcement not work?

https://github.com/orgs/wolfi-dev/discussions/33052

justinvreeland: Oh huh, I'm not sure it's working for me anymore either.

dannf: I believe `GCC_SPEC_FILE` worked for me when I tested it by pointing it at a different real spec file. I didn't try pointing it at `/dev/null` + and adding other `CFLAGS` though as that document describes. My goal was to avoid diverting from `openssf.spec` as much as possible. This should only filter out the problematic flag, leaving all of the rest of the `openssf` flags in-tact.

dannf:
Superseded by https://github.com/wolfi-dev/os/pull/38781

PR #34224, opened 2024-11-15
add ansible-operator package

kranurag7:
CI will fail for now given it's dependent on https://github.com/wolfi-dev/os/pull/33634

PR #34225, opened 2024-11-15
add awx-operator package

kranurag7:
Ci will fail given it's dependent on https://github.com/wolfi-dev/os/pull/34224 and PR linked.

### DIFF
--- a/awx-operator.yaml
+++ b/awx-operator.yaml
@@ -1,0 +1,42 @@
+package:
+  name: awx-operator
+  version: 2.19.1
+  epoch: 0
+  description: An Ansible AWX operator for Kubernetes built with Operator SDK and Ansible.
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - ansible-operator
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - py3-ansible-core
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/ansible/awx-operator
+      tag: ${{package.version}}
+      expected-commit: dd37ebd440edf953d822f2c134833a44a8e77532
+
+  - runs: |
+      mkdir -p ${{targets.contextdir}}/opt/ansible
+      cp requirements.yml ${{targets.contextdir}}/opt/ansible/requirements.yml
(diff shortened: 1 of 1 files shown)


awx-operator.yaml
```diff
@@ -0,0 +1,54 @@
...
+      runs: |
+        stat /opt/ansible/requirements.yml
+        stat /opt/ansible/watches.yaml
+        stat /opt/ansible/roles
+        stat /opt/ansible/playbooks/awx.yml
+    - name: Smoke test run
+      runs: |
+        mkdir -p /var/run/secrets/kubernetes.io/serviceaccount
+        echo dummy >/var/run/secrets/kubernetes.io/serviceaccount/token
+        export KUBERNETES_SERVICE_HOST=localhost KUBERNETES_SERVICE_PORT=6443 KUBERNETES_MASTER=https://localhost:6443
```
mritunjaysharma394, line 50: nb but are we simply invoking k8s env variable and not using our kwok test pipeline, is it important for this test to run?

maxgio92: It's not vital because the controller will start anyway retrying to connect to the API server (we don't need to test the reconciliation), but that's a good point. I'd try to leverage it

awx-operator.yaml
```diff
@@ -0,0 +1,52 @@
...
+    identifier: ansible/awx-operator
+
+test:
+  pipeline:
+    - name: Check Ansible files
+      runs: |
+        stat /opt/ansible/requirements.yml
+        stat /opt/ansible/watches.yaml
+        stat /opt/ansible/roles
+        stat /opt/ansible/playbooks/awx.yml
```
wojciechka, line 45: `test -f` (test if a file) and `test -d` (directory) would be a better and more explicit test.

But you're explicitly copying most things so I'd only leave the test for `playbooks/awx.yaml` ass you're copying an entire directory. You could do the same for one of the files in `roles` directory.

awx-operator.yaml
maxgio92: Addressed @mritunjaysharma394. Thank you! TIL

awx-operator.yaml
maxgio92: Good point. As discussed in general maintaining specific file checks over time can be cumbersome. In this case I'd just test `watches.yaml` playbook file as it's the one expected by default by the `run` command.

wojciechka (approved):

PR #34268, opened 2024-11-15
datadog-agent cleanups and do not install .yaml.default files

smoser:


maxgio92:
LGTM, let's the CI run.

This as discussed should remove the errors trying to always load those checks because of the related default configuration present by default.

datadog-agent.yaml
```diff
@@ -537,25 +548,28 @@ test:
...
     - name: Ensure the agent's requirements.txt for integrations is included
       runs: |
-        ls /opt/datadog-agent/requirements-agent-release.txt
+        ls ${{vars.destd}}/requirements-agent-release.txt
     - name: Ensure the agent integration subcommand works
       runs: |
-        /opt/datadog-agent/bin/agent/agent integration freeze >/dev/null
+        agent integration freeze >/dev/null
     - name: Ensure checks can be loaded
+        ${{vars.destd}}/embedded/bin/python3 -c "import datadog_checks.base"
```
smoser, line 560: I left this to be clear that python3 being executed is not /usr/bin/python3.

maxgio92 (approved):

PR #34281, opened 2024-11-15
build-base: provide build-essential

xnox:
Provide meta package build-essential, which helps people transition from deb-like operating systems.

There have been over 30 requests before asking where is build-essential.


gdal.yaml
```diff
@@ -125,7 +125,7 @@ subpackages:
     pipeline:
       - runs: |
           mkdir -p ${{targets.contextdir}}/usr/lib/
-          mv ${{targets.outdir}}/gdal-py${{range.key}}/usr/lib/python3* ${{targets.contextdir}}/usr/lib/
+          mv ${{targets.destdir}}/gdal-py${{range.key}}/usr/lib/python3* ${{targets.contextdir}}/usr/lib/
```
xnox, line 129: this is broken; i was doing this to see if it helps elastic builds or not.

build-essential.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name: build-essential
+  url: https://packages.debian.org/sid/build-essential
+  version: 12.12
+  epoch: 0
+  description: "virtual compat package for build-base"
+  copyright:
+    - license: MIT
+  dependencies:
+    runtime:
+      - build-base
```
imjasonh, line 11: Is there value in having `build-base` `provides=build-essential`?

imjasonh: Or, if we expect this to diverge from `build-base`, I think we should just explicitly list the constituent packages, instead of having two layers of indirection when someone wants to see what packages are in `build-essential`.

xnox: @imjasonh currently elastic builds are unable to build `build-base: provides: build-essential` at all =) github actions can.

xnox: Also in Debian/Ubuntu build-essential implies coreutils bash dash-binsh so not sure if we should honor that or not.

justinvreeland: I'd like to. If we're going to have the package it might as well match and I don't think I'm the only ones whose added coretuils to get more feature full commands during builds.

xnox: decisions; should `coreutils bash-binsh gnu-awk gnu-find` simply be what build-base depends on? cause that would make things just work.

smoser:
> There have been over 30 requests before asking where is build-essential.

what do these requests look like? mostly just curious.  Its one of those things that there isn't a 1:1 mapping (build-base on debian gets you one version of gcc and will get you another here). its not going to solve a _lot_ of problems.

Generally i'm in the range of "this is fine", but I don't want to deal with someone asking for a fix because debian's build-essential does one thing and we don't act just like that.

xnox:
> > There have been over 30 requests before asking where is build-essential.
> 
> what do these requests look like? mostly just curious. Its one of those things that there isn't a 1:1 mapping (build-base on debian gets you one version of gcc and will get you another here). its not going to solve a _lot_ of problems.
> 
> Generally i'm in the range of "this is fine", but I don't want to deal with someone asking for a fix because debian's build-essential does one thing and we don't act just like that.

bits and pieces failing to build; and all queries have been resolved successfully with `use build-base` as in all cases people wanted to get a C compiler. Follow on requests sometimes ended up being weeks later `how do i find libfoo` which have been resolved with `apk add pc:libfoo`.

Details are not clear; but i believe people have been transitioning dockerfiles from debian/ubuntu and previously used build-essential, and build-base is direct equivalent. As on both OSes, it gives one "the default c,c++ compiler and make" whichever versions of those are for a given debian or ubuntu release. As build-essential in debian/ubuntu is a rolling meta package, without any specific versions.

cmwilson21:
Thanks for the contribution!

PR #34294, opened 2024-11-16
bun

xnox:
- **New package bun**

### DIFF
--- a/bun.yaml
+++ b/bun.yaml
@@ -1,0 +1,82 @@
+package:
+  name: bun
+  version: 1.1.34
+  epoch: 0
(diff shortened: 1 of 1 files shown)


tuananh:
@xnox it seems we're using 2 conventions here: both bootstrap and stage0 :)

xnox:
> @xnox it seems we're using 2 conventions here: both bootstrap and stage0 :)

I know we don't enforce it at all. But I tend to use stage0 meaning there maybe N other stages; and that package is compiled somehow very differently (less features, less runtime dependencies, statically linked and so on). Whereas bootstrap packages, typically are equivalent, but came compiled from somewhere else. But yeah, it is confusing overall to use more than one name for these things.

developer-guy (approved):

PR #34425, opened 2024-11-18
switch http git checkouts for github to https

rawlingsj:
This should probably be a lint check too.

Not bumping the epoch as no changes to the packages.


cpanato (approved):

xnox:
Also, i want to drop ".git" from the end of urls, to ensure all of them are consistent.

PR #34427, opened 2024-11-18
linux-headers: upgrade to v6.12

xnox:
linux-headers are UAPI headers with strong backwards guarantees. Thus it is generally safe to always upgrade them.

Filter out rc releases, but otherwise unpin to track latest stable releases.


89luca89 (dismissed):

89luca89:
Do you think we can add a test using https://docs.kernel.org/dev-tools/checkuapi.html ?

imjasonh (approved):

xnox:
> Do you think we can add a test using https://docs.kernel.org/dev-tools/checkuapi.html ?
> 

That's might not achieve anything. As by the time UAPI header is modified and made it to a stable tag it is the new frozen ABI that is backwards compatible.

As in any changes to those already pass that, or intentionally deemed compatible (obsolete fields renamed or ignored, but keep the same bitfield slots). Thus even when/if checkuapi check fails it likely to be a false negative.

Running that would be nice if we ever need to fork headers; apply out of tree changes; or backport features. So far we only ever used stable tags, meaning the exposed headers is the de facto status quo.

PR #34454, opened 2024-11-18
datadog-agent: add missing directory for system probe sockets

maxgio92:
This PR adds the required directory for the system probe agents' IPC socket files.
In particular:
* system probe (default: /opt/datadog-agent/run/sysprobe.sock)
* security runtime agent (default: /opt/datadog-agent/run/runtime-security.sock)

The /opt/datadog-agent/run directory was missing leading to the impossibility to create the IPC socket files for the system probes, for which the following error log is printed:
```
error while starting api server, exiting: error creating IPC socket:
can't listen unix /opt/datadog-agent/run/sysprobe.sock: bind: no such file or directory.
```

### DIFF
--- a/datadog-agent.yaml
+++ b/datadog-agent.yaml
@@ -3,7 +3,7 @@ package:
   # This package has two git checkouts. For each new release, the commit SHA for
   # DataDog/integrations-core must also be updated.
   version: 7.59.0
-  epoch: 1
+  epoch: 2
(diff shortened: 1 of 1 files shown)


EyeCantCU (changes requested):
Missing epoch bump

EyeCantCU (approved):

smoser (approved):
Looks good. Thank you.

PR #34541, opened 2024-11-18
gatekeeper-3.17: Add crds package

lyoung-confluent:
In addition to the [primary Dockerfile](https://github.com/open-policy-agent/gatekeeper/blob/master/Dockerfile) the upstream [open-policy-agent/gatekeeper](https://github.com/open-policy-agent/gatekeeper) repository build, an additional [CRD image](https://github.com/open-policy-agent/gatekeeper/blob/master/crd.Dockerfile) is built. 

This image is pushed to `openpolicyagent/gatekeeper-crds` and is [referenced/used by the gatekeeper Helm chart](https://github.com/open-policy-agent/gatekeeper/blob/56eac9f4bf4e53dd6bf99fe79af5fc1b84e525bb/charts/gatekeeper/templates/upgrade-crds-hook.yaml#L112-L115) to install/replace the Kubernetes CRDs before installing/upgrading the gatekeeper service.

This PR adds a matching `gatekeeper-crds` package containing these CRDs, allowing a similar wolfi-based image to be created containing just `gatekeeper-crds` and `kubectl`.


gatekeeper-3.17.yaml
```diff
@@ -59,6 +59,15 @@ subpackages:
       provides:
         - gatekeeper-gator=${{package.full-version}}
 
+  - name: ${{package.name}}-crds
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.destdir}}/crds
+          cp charts/gatekeeper/crds/* ${{targets.destdir}}/crds
```
kranurag7, line 66:
```suggestion
          mkdir -p ${{targets.contextdirdir}}/crds
          cp charts/gatekeeper/crds/* ${{targets.contextdir}}/crds
```
given this is a subpackage we can consider copying things over here else this package is coming out empty.

kranurag7 (changes requested):
thanks for the patch, requesting some changes inline.

kranurag7 (approved):

PR #34554, opened 2024-11-18
apache2: Move conf to /etc/apache2/ instead of /etc

roaksoax:
apache2: move conf to /etc/apache2 instead of /etc


roaksoax:
Seems like adding lua5.4 explicitly as a dependency causes the build to fail, which is more than just the presumit check.

`failed to test package: unable to build guest: unable to generate image: installing apk packages: installing packages: installing lua-rrd (ver:1.8.0-r0 arch:x86_64): unable to install files for pkg lua-rrd: unable to install file over existing one, different contents: usr/lib/lua5.4/liblua-5.4.so.0.0.0`

smoser:
I put up https://github.com/wolfi-dev/os/pull/34786 with a solution to the immediate problem.

smoser (approved):

PR #34625, opened 2024-11-19
py3-fabric - multi-version

smoser:
use of openssl-legacy-provider is to avoid the:

>  RuntimeError: OpenSSL 3.0's legacy provider failed to load. This is
>  a fatal error by default, but cryptography supports running without
>  legacy algorithms by setting the environment variable
>  CRYPTOGRAPHY_OPENSSL_NO_LEGACY. If you did not expect this error, you
>  have likely made a mistake with your OpenSSL configuration.

The use of test/populated-etc-passwd is to cover python 3.13. os.getuid() requires. either USER set or user in /etc/passwd.

> ```
>  + fab --version
>  Traceback (most recent call last):
>    File "/usr/lib/python3.13/getpass.py", line 173, in getuser
>      return pwd.getpwuid(os.getuid())[0]
>             ~~~~~~~~~~~~^^^^^^^^^^^^^
>  KeyError: 'getpwuid(): uid not found: 1000'
> ```

py3-deprecated needed changing to not provide py3-Deprecated
at a specific version.

### DIFF
--- a/py3-deprecated.yaml
+++ b/py3-deprecated.yaml
@@ -1,8 +1,7 @@
-# Generated from https://pypi.org/project/Deprecated/
 package:
   name: py3-deprecated
   version: 1.2.15
-  epoch: 0
+  epoch: 1
   description: Python @deprecated decorator to deprecate old python classes, functions or methods.
   copyright:
     - license: MIT
@@ -10,8 +9,8 @@ package:
     provider-priority: "0"
 
 vars:
-  module_name: deprecated
   pypi-package: deprecated
+  import: deprecated
 
 data:
   - name: py-versions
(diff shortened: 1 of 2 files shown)


py3-fabric.yaml
```diff
@@ -28,21 +32,85 @@ pipeline:
...
+      provider-priority: ${{range.value}}
+      provides:
+        - py3-${{vars.pypi-package}}
+        - py3-${{vars.pypi-package}}-bin
+      runtime:
+        - py${{range.key}}-${{vars.pypi-package}}
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/usr/
+          mv ./cleanup/${{range.key}}/bin ${{targets.contextdir}}/usr/
```
dannf, line 82: Are subpackages guaranteed to be built in order of definition, and serialized? If not, this would be racy.

dannf (dismissed):
Thanks! Just a couple nits (take or leave) and a question

py3-deprecated.yaml
```diff
@@ -2,7 +2,7 @@
 package:
```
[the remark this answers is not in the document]
smoser: i dont know. they got dropped when i did the conversion . but since this one was done by had its still there. i have no strong feeling on it.

py3-fabric.yaml
smoser: they're built in order. i admit that is kind of weird.

dannf (approved):

PR #34649, opened 2024-11-19
libisoburn - fix duplicate copies of xorriso, make 'tk' provide cmd:wish

smoser:
Both xorriso package and libisoburn package were providing xorriso
binary.  That surely wasn't as intended.   Fix that so that xorriso
has only it's binaries and the dependencies are found by melange.

Other change here is to split xorriso-tcltk out into its own package. Then let it depend on tk, and tell tk that it provides 'cmd:wish'.

'cmd:wish' is not automatically added by melange due to it being a symlink to wish9.0


justinvreeland (dismissed):

justinvreeland (approved):

xnox:
Nice!

PR #34680, opened 2024-11-20
Remove php provides with a version, replace with providerPriority

### LINKED ISSUE #32874: php virtual provides should not include version
php packages currently have virtual provides with a full package version.
This causes issue when multiple php versions exist as discussed some https://github.com/wolfi-dev/os/pull/32398.
```yaml
    dependencies:
      provides:
        - php-igbinary-config=${{package.full-version}}
```

The biggest problem here is:
 1. you have 2 packages php8.3-ampq and php8.2-ampq both at version 2.1.2 at epoch 0.  They both provide `php-amqp=${{package.full-version}}`
 2. A change is made to php8.2-ampq, and the epoch bumped.  

Now, if something tries to install the virtual `php-ampq`, they'll get the php8.2 version rather than the 8.3 version.

We were lucky and avoided that problem in python because of the multi-versioning.  You cannot currently bump py3.12-foo without bumping py3.13-foo (they are both built by py3-foo).

So I think that:
 * **virtual provides are only for humans**.  They should never be allowed to be used by image builds. 
 *  We should not provide versions on virtual provides, but rather rely on provider-priority. (php-igbinary-config is fine, but php-igbinary-config-${{package.version}} is not).  This decision is primarily made for consistency with py3-*).  I do not have a reason why one option should be strongly preferred over the other, so I'm choosing consistency.

In order to accomplish this, we have to
* drop the version on php-* virtual provides
* bump and build and withdraw all older versions php-* . This is required so that users would not continue to get the older version as version provider-priority is only consulted when version is equal.

smoser:
In all the php-8.*-*.yaml files:
1. replace all of the 'phpMM' definitions with simple var rather than transform based name.
2. replace any virtual provides with a version with an empty version and set providerPriority to a value based on the PHP version (instead of package version).

   This takes care to handle PHP versions 8.10 by using major*100 + minor (803 for 8.3 , which would mean 810 for 8.10).

The first item isn't a huge win, but the second is needed if we're going to have virtual provides.

Fixes: https://github.com/wolfi-dev/os/issues/32874

### DIFF
--- a/php-8.1-igbinary.yaml
+++ b/php-8.1-igbinary.yaml
@@ -1,7 +1,7 @@
 package:
   name: php-8.1-igbinary
   version: 3.2.16
-  epoch: 2
+  epoch: 3
   description: "Igbinary is a drop in replacement for the standard php serializer."
   copyright:
     - license: BSD-3-Clause
@@ -10,11 +10,9 @@ package:
       - ${{package.name}}-config
       - php-${{vars.phpMM}}
 
-var-transforms:
-  - from: ${{package.name}}
-    match: ^php-(\d+\.\d+)-.*$
-    replace: "$1"
-    to: phpMM
+vars:
+  phpMM: "8.1"
+  phpPriority: 801
 
 environment:
   contents:
@@ -51,17 +49,19 @@ subpackages:
   - name: ${{package.name}}-config
     dependencies:
       provides:
-        - php-igbinary-config=${{package.full-version}}
+        - php-igbinary-config
+      provider-priority: ${{vars.phpPriority}}
     pipeline:
       - runs: |
           mkdir -p "${{targets.subpkgdir}}/etc/php/conf.d"
           echo "extension=igbinary.so" > "${{targets.subpkgdir}}/etc/php/conf.d/igbinary.ini"
 
   - name: ${{package.name}}-dev
-    description: PHP ${{vars.phpMM}} igbinary development headers
+    description: "PHP ${{vars.phpMM}} igbinary development headers"
     dependencies:
       provides:
-        - php-igbinary-dev=${{package.full-version}}
+        - php-igbinary-dev
+      provider-priority: ${{vars.phpPriority}}
     pipeline:
       - uses: split/dev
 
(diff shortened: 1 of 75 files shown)


smoser:
We will need a withdrawal after these packages land of all the packages that had a provides.
This is the list of origin packages that had a phpPriority added:

php-8.1-amqp.yaml
php-8.1-igbinary.yaml
php-8.1-redis.yaml
php-8.2-amqp.yaml
php-8.3-amqp.yaml
php-8.3-apcu.yaml
php-8.3-excimer.yaml
php-8.3-grpc.yaml
php-8.3-igbinary.yaml
php-8.3-imagick.yaml
php-8.3-memcached.yaml
php-8.3-msgpack.yaml
php-8.3-opentelemetry.yaml
php-8.3-redis.yaml
php-8.3-swoole.yaml

php-8.1-igbinary.yaml
```diff
@@ -10,11 +10,9 @@ package:
       - ${{package.name}}-config
       - php-${{vars.phpMM}}
```
xnox, line 11: If you remove variable, you need to stop using it too

xnox (changes requested):

xnox (approved):

xnox:
Provides for humans work, if they are consistent:

```
$ ls php-8.[0-9]-*.yaml | wc -l
75

$ git grep -l provides php-8.[0-9]-*.yaml | wc -l
17
```

Out of 75 addons we have, only 17 have some provides. I wonder if we should drop provides all together; and make everyone be explicit as to what they want. Or add provides consistently everywhere else.

Removing provides is sort of a breaking change; thus likely if we choose to go that route, we should drop them with php8.4.

smoser:
> Out of 75 addons we have, only 17 have some provides. I wonder if we should drop provides all together; and make everyone be explicit as to what they want. Or add provides consistently everywhere else.

I had that thought too.  my only reason for keeping them was that they'd be an indication of "the right way" to do it.  (And if it wasn't actually the right way, then we'd fix it so it was).

smoser:
I'm going to push another PR that drops the provides entirely.

ajayk (approved):

PR #34683, opened 2024-11-20
Add php-8.4 modules

smoser:
First, just add the modules, copying from 8.3 versions.

    $ for p in php-8.3-*.yaml; do
      sed -e 's,epoch: .*,epoch: 0,' \
          -e 's,name: php-8.3-,name: php-8.4-,' \
         "$p" > php-8.4${p#php-8.3};
    done

Then fix some to build with 8.4, cherry picking from upstream.

 * php-8.4-pecl-mcrypt - cherry-pick a PR to build for 8.4
   Grab https://github.com/php/pecl-encryption-mcrypt/pull/19
   which just uses php_mt_rand_range rather than now-removed php_rand.

 * php-8.4-imagick - cherry pick pr to build for 8.4
   Grab https://github.com/Imagick/imagick/pull/690 
   to replace php_strlower with zend_str_tolower

 * php-8.4-xdebug - cherry pick several upstream changes
   NOTE: I have very low confidence in this, other than it builds.
   xdebug is active and likely to release a 3.4.0 release soon
   that has php 8.4 support properly.

   By adding php-8.4-xdebug with updates enabled, we will get
   a PR automatically made and hopefully landed just by dropping
   the cherry-picks here.

 * Drop php-8.4-swoole - not supported upstream yet.
    
   Several changes would be needed to make the 5.1.5
   version of swoole build with php 8.4.
   Upstream explicitly removed support from
   the 5.1 branch https://github.com/swoole/swoole-src/pull/5525
   
   There is likely 6.0.0 release coming soon (beta released 2024-10-17).
   We will have to manually re-add at that point in time.

 * php-8.4-memcached and php-8.4-redis are separated out to another PR
   as they are dependent on php-8.4-igbinary output.


dannf (approved):
I diff'd the modules that exist for both 8.3 and 8.4 and those changes look straightfoward. I didn't double-check the cherry-picks, but I trust the given justifications.

PR #34720, opened 2024-11-20
Drop virtual provides from php module packages

smoser:
Some of the php modules had virtual provides. The intent of such a thing would be so that a process could 'apk add php-amqp' and get a sane result.

The implementation had a problem where by doing so would choose whichever php-8.X-amqp version had the newest package version. That is almost certainly _not_ what the process wanted (one day to get php-8.2-ampq and the next to get php-8.1-ampq).

https://github.com/wolfi-dev/os/issues/32874 describes the problem and a solution. https://github.com/wolfi-dev/os/pull/34680 implemented that solution.  In the PR we realized that the provides were so inconsistent that it would be had to use the. So then, given the option of making them consistent or dropping them, we chose to drop them.

### DIFF
--- a/php-8.1-amqp.yaml
+++ b/php-8.1-amqp.yaml
@@ -1,7 +1,7 @@
 package:
   name: php-8.1-amqp
   version: 2.1.2
-  epoch: 2
+  epoch: 3
   description: "PHP extension to communicate with any AMQP compliant server"
   copyright:
     - license: PHP-3.01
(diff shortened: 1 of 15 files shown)


xnox (dismissed):

smoser (commented):
I'm going to fix the little nits that the machine-edit added. there are a couple.

xnox (approved):

PR #34721, opened 2024-11-20
py3-grpcio-gcp: Add python multiversion support

dannf:

From the individual commits:

-  pip-build-install pipeline: Introduce "source" input variable
    -  Python modules may be provided in subdirectories of the upstream source.
-  py3-grpcio-gcp: Build from github repository using py/pip-build-install
    -  Building from the tarball results in an unusable package because  the included grpc_gcp_pb2.py is generated from an old protobuf. The tarball includes the .proto file we need to generate a new one.
-  py3-grpcio-gcp: Add python multiversion support

### DIFF
--- a/py3-grpcio-gcp.yaml
+++ b/py3-grpcio-gcp.yaml
@@ -17,22 +17,34 @@ environment:
       - build-base
       - busybox
       - ca-certificates-bundle
+      - py3-build-base
+      - py3-grpcio-tools
       - py3-setuptools
       - python-3
       - wolfi-base
 
 pipeline:
-  - uses: fetch
+  - uses: git-checkout
     with:
-      expected-sha256: e292605effc7da39b7a8734c719afb12ec4b5362add3528d8afad3aa3aa9057c
-      uri: https://files.pythonhosted.org/packages/source/g/grpcio-gcp/grpcio-gcp-${{package.version}}.tar.gz
+      repository: https://github.com/GoogleCloudPlatform/grpc-gcp-python
+      tag: v${{package.version}}
+      expected-commit: 8ef890243664ebd95ae0225cdf87e7442b570f9e
+
+  - runs: |
+      cd src
+      # From src/setup.sh
+      cp -f ../template/version.py version.py
+      python3 -m grpc_tools.protoc -I. --python_out=grpc_gcp/proto grpc_gcp.proto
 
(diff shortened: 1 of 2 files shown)


xnox (approved):
NICE!!!! love the copy template bit.

py3-grpcio-gcp.yaml
```diff
@@ -1,38 +1,90 @@
...
+      tag: v${{package.version}}
+      expected-commit: 8ef890243664ebd95ae0225cdf87e7442b570f9e
 
-  - name: Python Build
-    uses: python/build-wheel
+  - runs: |
+      cd src
+      # From src/setup.sh
+      cp -f ../template/version.py version.py
+      python3 -m grpc_tools.protoc -I. --python_out=grpc_gcp/proto grpc_gcp.proto
```
smoser, line 41: this is really weird, because the only reason you have `python3` is because `py3-grpcio-tools` has a manually added runtime dependency on it.

so if you multi-version py3-grpcio-tools, it will break this.

py3-grpcio-gcp.yaml
```diff
@@ -1,38 +1,90 @@
...
+      provides:
+        - py3-${{vars.pypi-package}}
+      runtime:
+        - py${{range.key}}-grpcio
+        - py${{range.key}}-protobuf
+    pipeline:
+      - uses: py/pip-build-install
+        with:
+          python: python${{range.key}}
+          source: ./src
```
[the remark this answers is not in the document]
dannf: ah - do you think I should revert and switch to `working-directory` to minimize pattern divergence?

py3-grpcio-gcp.yaml
dannf: Well noticed, I'll fix that :)

dannf: https://github.com/wolfi-dev/os/pull/34780/commits/b80c0466c790ce7a792608dc13f914ea10363885

py3-grpcio-gcp.yaml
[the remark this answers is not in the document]
dannf: https://github.com/wolfi-dev/os/pull/34780/commits/f902408c13b09311ebe62c9d8db8f8482a1aa34e

PR #34730, opened 2024-11-20
Update docs to remove reference to a wolfictl update check subcommand

mamccorm:
This command no longer exists in wolfictl, so updating the docs accordingly. Also made some small tweaks re: suggestions


kranurag7 (approved):
thank you!!

PR #34750, opened 2024-11-20
fix/apache nifi 2.0 support

philroche:
**fix(apache-nifi.yaml): Use single thread in both architectures**

Use single thread to avoid `java.lang.IllegalStateException: Could not acquire lock(s)` build error

**fix(apache-nifi.yaml): Catch error when testing apache-nifi server start script**

More accurately represent a correct apache-nifi server start - just as our container images do.

The start script will not exit non zero even if it fails. Previous test would always pass as long
as the script existed and was executable.

As such we need to ensure we start the same we to with the apache-nifi container image entrypoint
by setting the expected environment variables and by ensuring the expected log message is present.

**fix(apache-nifi.yaml): Use the same start script as the container image uses as its entrypoint**

This ensures we are being consistent in testing and ensuring that we block on the correct failure which would block image build/test.

**fix(apache-nifi.yaml): Add python3 as a runtime dependency**

With Apachae nifi 2.0+ - python3 is now a runtime dependency.

Without this error below will occur when attempting start

```
Caused by: java.lang.IllegalStateException: Failed to communicate with Python Controller
Caused by: java.io.IOException: Cannot run program "python3": error=2, No such file or directory
```

### DIFF
--- a/apache-nifi.yaml
+++ b/apache-nifi.yaml
@@ -156,6 +156,16 @@ update:
 test:
   pipeline:
     - runs: |
+        # ensure the same environment variables are set as the container image
         export JAVA_HOME=/usr/lib/jvm/java-21-openjdk
+        export NIFI_BASE_DIR="/usr/share/nifi"
+        export NIFI_HOME="/usr/share/nifi/nifi-current"
+        export NIFI_TOOLKIT_HOME="/usr/share/nifi/nifi-toolkit-current"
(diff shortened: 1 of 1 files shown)


olabadmus (approved):

mamccorm (approved):

philroche:
CVE reported can be addressed with follow on PR - see https://github.com/wolfi-dev/os/pull/34688

PR #34786, opened 2024-11-20
Withdraw lua-rrd-1.8.0-r0 due to it providing liblua-5.4.so

smoser:
The 1.8.0-r0 version of lua-rrd provided so:liblua-5.4.so That causes install problems for anything with a dependency so:liblua-5.4.so as it might get the lua-rrd version.

The 1.9 versions are still providing it, but it gets marked as 'vendored' now so it doesn't get selected.

It still should not provide the paths below, as it will then conflict with lua5.4 or that has a liblua-5.4 dependency.

    # apk info -L lua-rrd
    lua-rrd-1.9.0-r1 contains:
    usr/lib/lua5.4/liblua-5.4.so.0
    usr/lib/lua5.4/liblua-5.4.so.0.0.0
    usr/lib/lua5.4/liblua.a
    usr/lib/lua5.4/liblua.la
    usr/lib/lua5.4/liblua.so

### DIFF
--- a/withdrawn-packages.txt
+++ b/withdrawn-packages.txt
@@ -1,4 +1,1 @@
-istio-install-cni-1.22-1.23.0-r0.apk
-istio-install-cni-1.22-compat-1.23.0-r0.apk
-istio-install-cni-1.22-1.23.0-r1.apk
-istio-install-cni-1.22-compat-1.23.0-r1.apk
+lua-rrd-1.8.0-r0.apk


smoser:
rrdtool still needs fixing i think.  Why would anyone want to `--enable-lua-site-install` (or `--enable-perl-site-install`)?

smoser:
https://github.com/chainguard-dev/melange/pull/1662 is the fix to melange so that lua5.4 will start to provide so:liblua5.4.so.1

smoser:
closing this. I'm going to do it as part of https://github.com/wolfi-dev/os/pull/34944

PR #34848, opened 2024-11-21
openssl: do not install obsolete cryptography by default

xnox:
Make it opt-in. Known users are py3-cryptography and webpack v4 (uses md4 as a oneway compression function).

This means algorithms provided by legacy provider will not be available, unless one explicitely installs or depends on openssl-provider-legacy.

The algorithms in question are all broken, obsolete and no longer in active use. For a full list please see:

https://docs.openssl.org/3.4/man7/OSSL_PROVIDER-legacy/

Note that many other security focused distributions have already stopped shipping legacy provider altogether (i.e. OpenBSD).

### DIFF
--- a/kubeflow-centraldashboard.yaml
+++ b/kubeflow-centraldashboard.yaml
@@ -1,7 +1,7 @@
 package:
   name: kubeflow-centraldashboard
   version: 1.9.2
-  epoch: 1
+  epoch: 2
   description: Landing page and central dashboard for Kubeflow deployments
   copyright:
     - license: MIT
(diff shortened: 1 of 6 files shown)


justinvreeland (dismissed):

justinvreeland (dismissed):

xnox:
i was waiting on malcontent, that clearly hang; and now there is conflict again =(

xnox:
@justinvreeland we should just merge this before it gets stale, yet again.

justinvreeland (approved):

PR #34893, opened 2024-11-21
fix(trino.yaml): Ensure build time and runtime use the same version of JVM and JDK

philroche:
fix(trino.yaml): Ensure build time and runtime use the same version of JVM and JDK

Without this there will be runtime errors during image test like:

```
Exception in thread "main" java.lang.UnsupportedClassVersionError: io/trino/server/Server has been compiled by a more recent version of the Java Runtime (class file version 67.0), this version of the Java Runtime only recognizes class file versions up to 66.0
```

This fix addresses trino container image test failures blocking publication.

### DIFF
--- a/trino.yaml
+++ b/trino.yaml
@@ -1,7 +1,7 @@
 package:
   name: trino
   version: "464"
-  epoch: 2
+  epoch: 3
   description: The distributed SQL query engine for big data, formerly known as PrestoSQL
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


philroche:
Package update check CI failure will be resolved when https://github.com/wolfi-dev/os/pull/34808 lands. Can be ignored for this fix.

joshrwolf (approved):

imjasonh:
- ci-diff-report times out on these APKs
- ci-mal-report will almost certainly time out too
- ci-apk-add seems to have flaked with a `bus error`

PR #34897, opened 2024-11-21
Add php-8.4-memcached.yaml and php-8.4-redis.yaml.

smoser:
These were split from https://github.com/wolfi-dev/os/pull/34683 due to failing to pass c-i. that seemed to be due to the fact that these two packages depend on php-8.4-igbinary.

They're just copy and 'sed'it from the php-8.3-<module> name.

### DIFF
--- a/php-8.4-memcached.yaml
+++ b/php-8.4-memcached.yaml
@@ -1,0 +1,67 @@
+package:
+  name: php-8.4-memcached
+  version: 3.3.0
+  epoch: 0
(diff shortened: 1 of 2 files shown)


smoser:
once https://github.com/wolfi-dev/os/pull/34683 lands this should build with just a merge.

dannf (approved):
Yep, confirmed it is a simple sed + epoch reset.

PR #34944, opened 2024-11-22
rrdtool - do include lua5.4 libs, withdraw rrdtool-1.8.0-r0, other cleanups

smoser:
The thing that brought me here was that rrdtool was not co-installable with lua5.4, which didn't make any sense.  It was actually copying the installed lua5.4 libraries to the target (not building them).

The fixes here are then:
a. do not include lua5.4 itself
b. build the rrd lua module and include it
c. move things to the -doc package
d. separate out python bindings into py3.12-rrdtool e. add more tests


roaksoax (dismissed):
lgtm!

jonjohnsonjr (approved):

PR #34985, opened 2024-11-22
lua5.3, lua5.4 - package cleanups and rebuild

smoser:
- lua5.4 - updates to package and a rebuild.

  A build of lua5.4 right now will not 'provide: so:liblua5.4.so' due
  to /usr/lib/liblua* being symlinks.
  We need a build https://github.com/chainguard-dev/melange/pull/1662
  to fix that.

  Other things here are:
   * build 5.4 with V=5.4 (not V=5.3)
   * build 5.4 with compath for 5.3 not the non-existant 5.2
     https://gitlab.alpinelinux.org/alpine/aports/-/issues/13257
   * split libs out from the main package
     (main package will get a dep on the libs so that will be fine)
   * rename patches in lua5.4/ to not be redundantly lua-5.4-thing
   * define and use a var 'luaMM' for the major.minor everywhere.
   * use one patch section with 3 patches rather than 3 patch sections
   * do not install liblua.la
   * improve tests a bit to check for version in '-v' output
   * add a test to make sure lua5.4-libs provides the desired so:
   * fix typo in configure.ac (top_builddir, not 'top_buildir').

- lua5.3 - apply the same changes, but leave it building with linenoise
  instead of readline.


smoser:
OK, I'm going to write a summary of the lua5.4/rrdtool madness here.

1. recent builds of lua5.4 do no not provide `so:liblua-5.4.so.0=0` as they should.  That is the install layout where they install their libraries in /usr/lib/lua5.4/ and provide symlinks to those in /usr/lib.  melange SCA ignored the symlinks and ignored the actual libraries for `provides` consideration because those libraries are not in /usr/lib.  That is fixed https://github.com/chainguard-dev/melange/pull/1662 .  The melange change is making its way into builds (currently used in GHA but not elastic builds).

  A rebuild, and a test to ensure the so is provided is included in this PR.

2. rrdtool build is currently badly broken. It contains a package 'lua-rrd' which contains nothing other than the libraries from the liblua5.4 that it used in it's build environment.  The _old_ versions of lua-rrd (1.8.0-r0) were identified as providing `so:liblua-5.4.so.0=0` because they were built with a melange that did not limit libraries for provides to /usr/lib. 

  The fix for rrdtool is https://github.com/wolfi-dev/os/pull/34944 .  They depend on the lua5.4 rebuild so they can install.

3. As a result of 1 and 2 above, anything that has a runtime dependency on liblua5.4 gets some random behavior on installation, etiher picking an old version of lua5.4 or an old version of rrdtool.  One such example is https://github.com/wolfi-dev/os/pull/34554 .

To get this resolved we will
1. wait until elastic builds has melange code at [v0.15.13](https://github.com/chainguard-dev/melange/releases/tag/v0.15.13) or later
2. update / sync this PR to rebuild and land this
3. update / sync https://github.com/wolfi-dev/os/pull/34944
4. update / sync  https://github.com/wolfi-dev/os/pull/34554

xnox (approved):
A bit too verbose and too generic, but it is fine.

smoser:
> A bit too verbose and too generic, but it is fine.

I'm interested in some more info on that.
General changes, line numbers for lua5.4
1. [56-72] Pass `-DLUA_COMPAT_<LAST_VERSION>` and validate that maybe it will do something.   
2. [153-177] Validate that the lua `provides` the correct so:.  I added this to make sure it wouldn't land until melange was fixed.  I can see it viewed as knee-jerk response to specific bug.
3. [182-192] verify 'lua -v' outputs contains expected version.  This is as short as you can do this in shell without false positives or false negatives.  
4. use ${{vars.luaMM}} everywhere. 

I agree that testing in shell is long, error-prone, hard to read and hard to write.  I want to fix that with sane re-usable testing tools written in a sane language.

smoser:
> ✅ 1. wait until elastic builds has melange code at [v0.15.13](https://github.com/chainguard-dev/melange/releases/tag/v0.15.13) or later
> ✅ 2. update / sync this PR to rebuild
> ✅ 3. land this PR ([lua5.3, lua5.4 - package cleanups and rebuild #34985](https://github.com/wolfi-dev/os/pull/34985))
> ✅ 4. update / sync [rrdtool - do include lua5.4 libs, withdraw rrdtool-1.8.0-r0, other cleanups #34944](https://github.com/wolfi-dev/os/pull/34944)
> ✅ 5. update / sync  [apache2: Move conf to /etc/apache2/ instead of /etc #34554](https://github.com/wolfi-dev/os/pull/34554)

After #34944 landed, i ran withdraw workflow [here](https://github.com/wolfi-dev/os/actions/runs/12015445546/job/33493515801).

PR #34992, opened 2024-11-22
fix(sustaining): mitigate GHSA-8495-4g3g-x7pr in dask-gateway

ritster:


jamie-albert (approved):
Please leave a description of the changes, even if it seems trivial. I need to close this out asap but next time it will require changing before approval.

PR #34996, opened 2024-11-22
fix(sustaining): bump airflow epoch to mitigate GHSA-27mf-ghqm-j3j8 a…

ritster:
…nd GHSA-8495-4g3g-x7pr


egibs:
> Looking into this one ...

Removed the label so that the bincapz-check bot runs again.

tstromberg:
Reviewing this now, will make judgement within the next hour.

This appears to be unchanged from previous releases, so we probably shouldn't care, but checking it out anyways.

tstromberg:
This is probably a false positive, but I need to compare against a similarly compiled executable. We are not compiling OpenSSL ourselves here:

```
OpenSSL 3.0.15 3 Sep 2024
GCC: (GNU) 8.5.0 20210514 (Red Hat 8.5.0-22.0.1)
```

PR #35016, opened 2024-11-22
add lazy test case

xnox:


imjasonh:
Updating a test pipeline won't actually rebuild or test any packages, so it may be hard to tell if the change is correct. Do you want to bump an epoch to see the diff?

PR #35114, opened 2024-11-24
opensearch-dashboards-2/cve-2024-21534: remediation

rawlingsj:


opensearch-dashboards-2.yaml
```diff
@@ -63,6 +63,10 @@ pipeline:
       cherry-picks: |
         main/5e19749ec40230316ba2688c38e5c62f74ddb71d: CVE-2024-37890
 
+  - uses: patch
+    with:
+      patches: cve-2024-21538.patch
```
hectorj2f, line 68: This is good, but I am still not convinced with adding patches which aren't cleaned it up :/.

hectorj2f (approved):

opensearch-dashboards-2.yaml
rawlingsj: yeah though it's no different from what's happening today when someone adds any node CVE remediation

cleaning these up definitely needs to be automated too.

PR #35136, opened 2024-11-25
Modification in bats.yaml

anushkamittal20:
Adds a more detailed bats package to match upstream


bats.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: bats
   version: 1.11.0
-  epoch: 1
+  epoch: 0
```
kranurag7, line 4: I think this should be 2, otherwise existing one will be pulled only unless you've removed the r0 one. 💭

anushkamittal20: You are right, I think I reset it midway. Let me change that.

kranurag7 (dismissed):

bats.yaml
```diff
@@ -28,9 +29,28 @@ pipeline:
       expected-commit: 5da66876b8b619235aee1eb3e54954eaca88059b
 
   - runs: |
-      ./install.sh  "${{targets.contextdir}}/usr"
+      ./install.sh "${{targets.contextdir}}/usr"
+      mkdir -p "${{targets.contextdir}}/tmp/"
+      mv ./docker "${{targets.contextdir}}/tmp/docker"
+      ls -la "${{targets.contextdir}}/tmp/docker/"
```
EyeCantCU, line 35: Can remove

bats.yaml
```diff
@@ -28,9 +29,28 @@ pipeline:
...
+      "${{targets.contextdir}}/tmp/docker/install_libs.sh" support 0.3.0
+      "${{targets.contextdir}}/tmp/docker/install_libs.sh" file 0.4.0
+      "${{targets.contextdir}}/tmp/docker/install_libs.sh" assert 2.1.0
+      "${{targets.contextdir}}/tmp/docker/install_libs.sh" detik 1.3.1
+      mkdir -p "${{targets.contextdir}}/usr/local/bin"
+      ln -s /opt/bats/bin/bats "${{targets.contextdir}}/usr/local/bin/bats"
```
EyeCantCU, line 41: Why is this needed?

bats.yaml
```diff
@@ -28,9 +29,28 @@ pipeline:
...
+      find "${{targets.contextdir}}" -type f -perm /2000 -exec chmod g-s {} \;
+      rm -rf "${{targets.contextdir}}/tmp/"
 
 subpackages:
+  - name: bats-compat
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.contextdir}}/opt/bats/"
+          find . -maxdepth 1 -not -name '.' -not -name 'melange-out' -not -name '*.tmp' -not -name '*.log' \
+            -exec mv {} "${{targets.contextdir}}/opt/bats/" \;
```
EyeCantCU, line 51: Is it intended that we move over everything?

anushkamittal20: Hey RJ, thanks for the review. The reason we needed to add all of this was to match the upstream image. https://github.com/bats-core/bats-core/blob/master/Dockerfile where the entire repo is shifted and the link that I created above is also created.

bats.yaml
anushkamittal20: Agreed, i will correct this

bats.yaml
EyeCantCU: In that case, could we remove the .git folder? And anything else it may not actually be using at runtime?

anushkamittal20: I compared the dir in my image:
```
bash-5.2$ ls /opt/bats
AUTHORS                       docker                        man
Dockerfile                    docker-compose.override.dist  package.json
LICENSE.md                    docker-compose.yml            shellcheck.sh
README.md                     docs                          test
SECURITY.md                   install.sh                    uninstall.sh
bin                           lib
contrib                       libexec
```
vs upstream
```
/code # ls /opt/bats
AUTHORS                       bin                           docs                          package.json
Dockerfile                    contrib                       install.sh                    shellcheck.sh
LICENSE.md                    docker                        lib                           test
README.md                     docker-compose.override.dist  libexec                       uninstall.sh
SECURITY.md                   docker-compose.yml            man
```
both are the same

anushkamittal20: Should I explicitly look at removing items?

bats.yaml
```diff
@@ -28,9 +26,27 @@ pipeline:
       expected-commit: 5da66876b8b619235aee1eb3e54954eaca88059b
 
   - runs: |
-      ./install.sh  "${{targets.contextdir}}/usr"
+      ./install.sh "${{targets.contextdir}}/usr"
+      mkdir -p "${{targets.contextdir}}/tmp/"
+      cp -r ./docker "${{targets.contextdir}}/tmp/docker"
```
developer-guy, line 31: https://github.com/bats-core/bats-core/blob/89a7fae3b651d5c2f42a3874ec32e821014814eb/docker/install_libs.sh#L32 I saw that it downloads the files from their own repositories like https://github.com/bats-core/bats-support, so creating a separate packages for them like support, file, assert and detik would be great cuz in that case it'd be easier to keep these things in sync when there is a new version come up, but I'm not certain though.

bats.yaml
```diff
@@ -1,24 +1,22 @@
 package:
```
developer-guy, line 1: writing some tests would be great here:

for example:
cat <<EOF > example.sh
#!/bin/bash

# Function to add two numbers
add_numbers() {
    echo $(( $1 + $2 ))
}

# Function to check if a file exists
file_exists() {
    if [[ -f "$1" ]]; then
        echo "true"
    else
        echo "false"
    fi
}

# Function to check if a string contains another string
string_contains() {
    if [[ "$1" == *"$2"* ]]; then
        echo "true"
    else
        echo "false"
    fi
}
EOF

cat <<EOF >
#!/usr/bin/env bats

# Load the script to test
load_file() {
    source "./example.sh"
}

setup() {
    # This runs before every test
    load_file
}

teardown() {
    # This runs after every test
    rm -f testfile.txt
}

@test "add_numbers should correctly add two numbers" {
    run add_numbers 3 5
    [ "$status" -eq 0 ]
    [ "$output" -eq 8 ]
}

@test "file_exists should return true for existing file" {
    touch testfile.txt
    run file_exists testfile.txt
    [ "$status" -eq 0 ]
    [ "$output" = "true" ]
}

@test "file_exists should return false for non-existing file" {
    run file_exists non_existing_file.txt
    [ "$status" -eq 0 ]
    [ "$output" = "false" ]
}

@test "string_contains should return true if substring is found" {
    run string_contains "hello world" "hello"
    [ "$status" -eq 0 ]
    [ "$output" = "true" ]
}

@test "string_contains should return false if substring is not found" {
    run string_contains "hello world" "bye"
    [ "$status" -eq 0 ]
    [ "$output" = "false" ]
}
EOF

anushkamittal20: Sure thing!

bats.yaml
anushkamittal20: you mean like subpackages or completely new packages?

developer-guy: I mean completely new packages, yeah but we can do that later on for sure.

anushkamittal20: I can make an internal dev issue for it to track and get it in!

bats.yaml
```diff
@@ -28,9 +26,27 @@ pipeline:
...
-      ./install.sh  "${{targets.contextdir}}/usr"
+      ./install.sh "${{targets.contextdir}}/usr"
+      mkdir -p "${{targets.contextdir}}/tmp/"
+      cp -r ./docker "${{targets.contextdir}}/tmp/docker"
+      "${{targets.contextdir}}/tmp/docker/install_libs.sh" support 0.3.0
+      "${{targets.contextdir}}/tmp/docker/install_libs.sh" file 0.4.0
+      "${{targets.contextdir}}/tmp/docker/install_libs.sh" assert 2.1.0
+      "${{targets.contextdir}}/tmp/docker/install_libs.sh" detik 1.3.1
+      mkdir -p "${{targets.contextdir}}/usr/local/bin"
+      ln -s /opt/bats/bin/bats "${{targets.contextdir}}/usr/local/bin/bats"
```
kranurag7, line 37: we can consider moving this to compat package given most of `/opt` fs is within the compat package. 💭

anushkamittal20: Great suggestion!

kranurag7 (changes requested):

bats.yaml
```diff
@@ -44,6 +60,82 @@ update:
 
 test:
   pipeline:
-    # AUTOGENERATED
     - runs: |
         bats --version
+
+        echo 'add_numbers() {' > /tmp/example.sh
```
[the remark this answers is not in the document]
anushkamittal20: I tried doing that I kept getting errors, let me try again

bats.yaml
```diff
@@ -1,24 +1,22 @@
 package:
   name: bats
   version: 1.11.0
-  epoch: 1
+  epoch: 4
```
mritunjaysharma394, line 4: Why epoch 4? I think it should be 2

bats.yaml
mritunjaysharma394: Anushka did you address this?

bats.yaml
anushkamittal20: you are right!

bats.yaml
anushkamittal20: Yes Mritunjay, I have now! Thanks @kranurag7

mritunjaysharma394 (approved):
Great work, looks good to me then!

kranurag7 (approved):
🚢 📦 (thanks!!)

PR #35140, opened 2024-11-25
New package rust-audit-info

xnox:
Small binary from cargo-auditable collection of tools that helps to
parse & read cargo audit json information from a binary. Helps with
inspecting SBOM of a given binary.

### DIFF
--- a/rust-audit-info.yaml
+++ b/rust-audit-info.yaml
@@ -1,0 +1,45 @@
+package:
+  name: rust-audit-info
+  version: 0.5.4
+  epoch: 0
+  description: Read audit information from rust binaries
+  copyright:
+    - license: MIT OR Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - cargo-auditable
+      - openssf-compiler-options
+      - rust
+
+pipeline:
+  - uses: git-checkout
+    with:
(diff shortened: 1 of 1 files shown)


rust-audit-info.yaml
```diff
@@ -0,0 +1,45 @@
+package:
+  name: rust-audit-info
+  version: 0.5.4
```
kranurag7, line 3: I'm thinking about the reason to pin to this version given we have minor releases for the upstream project as well after this version. https://github.com/rust-secure-code/cargo-auditable/releases 

Also thinking if we can add it as a subpackage using cargo/build pipeline to existing cargo-auditable package given both comes from same repo.

xnox: @kranurag7 

it is not pinned. the subdirectory has its own versioning; and it's own tags.

See the update section; see the prefixed tags; and the version set of this binary inside its subdirectory.

v0.5.4 is latest rust-audit-info

https://crates.io/crates/rust-audit-info
https://github.com/rust-secure-code/cargo-auditable/releases/tag/rust-audit-info%2Fv0.5.4
https://github.com/rust-secure-code/cargo-auditable/blob/83713a1c609b28d93660c64b8974b0226fe2c7a4/rust-audit-info/Cargo.toml#L3

kranurag7 (approved):

PR #35145, opened 2024-11-25
update fluent-operator replaces to 3.2

jamonation:
The `fluent-watcher-config` subpackage notes that the `fluent-bit` version in the replaces block needs to be updated with version-stream updates. So this PR bumps the version from 3.1 -> 3.2.

### DIFF
--- a/fluent-operator.yaml
+++ b/fluent-operator.yaml
@@ -1,7 +1,7 @@
 package:
   name: fluent-operator
   version: 3.2.0
-  epoch: 1
+  epoch: 2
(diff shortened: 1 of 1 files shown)


mamccorm (approved):
Wonder if we could use a var-transforms to get '3.2' from package.version? But not blocking to prevent the image being fixed up, thanks for taking a look

jamonation:
> Wonder if we could use a var-transforms to get '3.2' from package.version? But not blocking to prevent the image being fixed up, thanks for taking a look

Oh I like it!

PR #35146, opened 2024-11-25
Add tests for ruby packages

mamccorm:
Add tests for ruby packages.
Also adds a missing ruby-3.3 package which we only had a ruby v3.2 version for.


luhring (approved):
I have thoroughly reviewed each line of code and applied my Ruby expertise to evaluating the quality of these tests.

PR #35123, opened 2024-11-25
remove conflicts files from fluent-watcher-config subpackage

kranurag7:
removes conflicting files

PR #35147, opened 2024-11-25
Use variable transform for fluent-bit-watcher subpackage version

jamonation:
Related: #35145 

Uses variable transform for fluent-bit-watcher subpackage to avoid manual updates on fluent-bit version bumps.

### DIFF
--- a/fluent-operator.yaml
+++ b/fluent-operator.yaml
@@ -1,7 +1,7 @@
 package:
   name: fluent-operator
   version: 3.2.0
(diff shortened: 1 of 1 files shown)


kranurag7 (approved):
looks good and I trust you with regex. :)

I'll also validate image building locally but I think you've already done that and conflicting files error are gone now with replaces in place. I'll close https://github.com/wolfi-dev/os/pull/35123 

Thank you!!

PR #34624, opened 2024-11-19
dotnet-9: use a bootstrap package to decouple the requirement artifacts

ader1990:
The first package - dotnet-bootstrap-9 - contains the downloadable artifacts
from the dotnet online repository, which are needed to build dotnet-9
fully offline from source. It packages the artifacts obtained after running
`./prep-source-build.sh`.

The second stage package, dotnet-9, will use the dotnet-bootstrap-9 packages
via a symlink, thus the `./prep-source-build.sh` execution is no longer needed.

PR #35150, opened 2024-11-25
dotnet-9: boostrap sdk with 2 stages

ader1990:
To be merged after: https://github.com/wolfi-dev/os/pull/34624


xnox:
Cute, but too much! We are ok with trusting MS bootstrap.

PR #35158, opened 2024-11-25
rustls-ffi: enable cargo auditable capi

xnox:
cargo-audtiable in latest upstream release fixed ability to have
auditable capi builds. Start using these.


kranurag7 (approved):
thank you!! `ci-apk-add` check looks like/is a flake.

ajayk (approved):

PR #35164, opened 2024-11-25
Add libquadmath as runtime dep for gcc

jonjohnsonjr:
This is silly, but the alternative to fix this will take a while, so I am proposing this in the meantime. We currently (incorrectly) assume that we will be able to solve for identical package sets across architectures, but that's not correct for per-arch things like libquadmath. I want to unwedge some things with this change, which I can revert once we fix these assumptions downstream.

### DIFF
--- a/gcc.yaml
+++ b/gcc.yaml
@@ -1,7 +1,7 @@
 package:
   name: gcc
   version: 14.2.0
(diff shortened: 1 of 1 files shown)


smoser (approved):
lgtm.

PR #35267, opened 2024-11-26
Migrate teleport package to use go/build

mamccorm:
The teleport build was leveraging the upstream projects make build, to produce the binaries. This meant we were missing binary optimisations by not leveraging our own go/build pipeline.

The changes in this PR flip the package over to using go/build for the various components.


developer-guy (dismissed):

teleport.yaml
```diff
@@ -35,32 +37,62 @@ pipeline:
...
+      tags: kustomize_disable_go_plugin_support
+
+  - uses: go/build
+    with:
+      packages: ./tool/teleport-update
+      output: teleport-update
+      tags: kustomize_disable_go_plugin_support
+
+  - runs: |
+      cd tool/fdpass-teleport && cargo build --release --locked
```
xnox, line 94:
```suggestion
      cd tool/fdpass-teleport && cargo auditable build --release --locked
```

teleport.yaml
```diff
@@ -35,32 +37,62 @@ pipeline:
...
+
+  - uses: go/build
+    with:
+      packages: ./tool/tctl
+      output: tctl
+      tags: kustomize_disable_go_plugin_support
+
+  - uses: go/build
+    with:
+      packages: ./tool/tsh
```
xnox, line 77: note mutliple packages can be specified together, and output can be specified as "./" as default target directory, especially when package name and output matches. This would remove duplication, especially since tags and destinations are the same.

teleport.yaml
```diff
@@ -35,32 +37,62 @@ pipeline:
...
-      # This is a bit of a hack, but it's the easiest way to get the right version of rustc and cargo in the path.
-      export PATH="$HOME/.rustup/toolchains/stable-${{host.triplet.rust}}/bin:$PATH"
-      make full
+      pnpm config set package-import-method copy
 
-      # create me a for loop that iterates over the binaries in build/
-      # and installs them to "${{targets.contextdir}}"/usr/local/bin
-      for bin in build/*; do
-        echo "Installing $bin to /usr/local/bin"
-        install -Dm755 $bin -t "${{targets.contextdir}}"/usr/local/bin
```
xnox, line 62: Note previously the binaries were installed directly into /usr/local/bin. For them to stay directly there, without need to create compat packages. In the go/build pipelines, one should set

```
with:
  prefix: usr/local
```

As otherwise go/build default to usr prefix. (note that /bin/ and output name will be appended)

teleport.yaml
```diff
@@ -35,32 +37,62 @@ pipeline:
...
-      done
+      # Install dependencies and build web assets
+      make ensure-js-deps
+      make ensure-webassets
+
+  - uses: go/build
+    with:
+      packages: ./tool/teleport
+      output: teleport
+      tags: webassets_embed,kustomize_disable_go_plugin_support
```
xnox, line 67:
```suggestion
      tags: webassets_embed,kustomize_disable_go_plugin_support
      prefix: usr/local
```

teleport.yaml
xnox: Note that one can check it was built correctly with:

```
$ rust-audit-info /usr/local/bin/fdpass-teleport 
No audit data found in the binary! Was it built with 'cargo auditable'?
```

Also I wonder if cargo auditable should ship wrapper in /usr/local/bin like with do with openssf compiler options; such that auditing is picked up automatically. Will follow up on this separately.

teleport.yaml
mamccorm: Interesting, was not aware of this - will take a look

teleport.yaml
mamccorm: ack - updated

mamccorm:
Ready for another review @xnox, @philroche. All PR feedback should be addressed

philroche (approved):

xnox (approved):

PR #35281, opened 2024-11-26
Adding java contrib as package

utieyin:
java-contrib: new package (opentelemetry jmx metrics)

Fixes:

Related:

#### For new package PRs only


java-contrib.yaml
```diff
@@ -0,0 +1,75 @@
+package:
+  name: java-contrib
```
jamonation, line 2: The package name should probably be something like `opentelemetry-java-contrib` to make its purpose clear. Otherwise I'd read it as being something more generic for the java ecosystem.

utieyin: @jamonation Very long name but renamed all packages :-)

jamonation (approved):

Issue #34568, opened 2024-11-19
openssf-compiler-options: `-Wl,-z,now` causes nvidia-device-plugin to fail to load


dannf:
I'm reporting this per https://github.com/orgs/wolfi-dev/discussions/33052.

I found that while a rebuild of `nvidia-device-plugin` w/ `openssf-compiler-flags` succeeds, the tests will fail:

```
2024/11/19 00:38:19 WARN + nvidia-device-plugin --version
2024/11/19 00:38:19 WARN nvidia-device-plugin: symbol lookup error: nvidia-device-plugin: undefined symbol: nvmlGpuInstanceGetComputeInstanceProfileInfoV
```

[nvidia-device-plugin-build-and-test-fail.txt](https://github.com/user-attachments/files/17807775/nvidia-device-plugin-build-and-test-fail.txt)
[nvidia-device-plugin-no-rebuild-test-ok.txt](https://github.com/user-attachments/files/17807776/nvidia-device-plugin-no-rebuild-test-ok.txt)

tuananh:
could have just change it to `-Wl,-z,lazy` and it should work.

i guess that this pkg need to link with `libnvidia-ml` on the host, that's why lazy is being used.

dannf:
> could have just change it to `-Wl,-z,lazy` and it should work.

Yeah, that did work when I hacked it onto the end of the options in `openssf.spec` - but I didn't identify a way to do it with build commands/environment variables. Neither `LDFLAGS` nor `CGO_LDFLAGS` did the trick.

>  i guess that this pkg need to link with `libnvidia-ml` on the host, that's why lazy is being used.

That's right.

tuananh:
ah i think it's because we use `make` to build. maybe convert this to use `go/build` and then we can use `ldflags` override in the pipeline

https://github.com/chainguard-dev/melange/blob/main/pkg/build/pipelines/go/build.yaml

dannf:
> ~ah i think it's because we use `make` to build (https://github.com/NVIDIA/k8s-device-plugin/blob/main/Makefile) . maybe convert this to use `go/build` and then we can use `ldflags` override in the pipeline~
> 
> https://github.com/chainguard-dev/melange/blob/main/pkg/build/pipelines/go/build.yaml

Thanks @tuananh. That would provide a hook for passing a clean `-ldflags`, but the problem persists:

```yaml
[...]
  - uses: go/build
    with:
      packages: ./cmd/nvidia-device-plugin
[...]

PR #35302, opened 2024-11-27
ossf lazy

xnox:
- **gcc: cherrypick fix for pr117739**
  Cherry-pick fix for:
  - https://github.com/wolfi-dev/os/issues/34568
  - https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117739
  - https://gcc.gnu.org/pipermail/gcc-patches/2024-November/670120.html

This change will enable using "-fhardened -Wl,-z,lazy" however, it will not yet enable using "-specs openssf.spec -Wl,-z,lazy" will need to open a new bug report about that. Still this should unblock openssf-ication of the affected packages.

- **add lazy test case**
  Note, checked that all existing packages that call this test pipeline do not regress with proposed change.

Example on how to use this in practice is shown at https://github.com/wolfi-dev/os/pull/35325 which does work locally and passes build + test (which i think is the original reproducer that triggered this). Note that package uses cgo, and thus CGO_CFLAGS and CGO_LDFLAGS should be used to affect the build.


pipelines/test/compiler-hardening-check.yaml
```diff
@@ -54,5 +59,8 @@ pipeline:
       # Test disabling hardening flags
       hardening-check --nostackprotector $arch_skip ${{inputs.args}} --color hello-disabled && exit 1
 
+      # Test disabling bindnow
+      hardening-check --nostackprotector $arch_skip ${{inputs.args}} --color hello-lazy && exit 1
```
dannf, line 63: Thanks for adding this test!

dannf (approved):
Approved w/ an optional suggestion inline. Thanks @xnox !

PR #35317, opened 2024-11-27
refactor parallal package to reduce size and runtime deps

kranurag7:
refactor parallal package to reduce size and runtime deps

### DIFF
--- a/parallel.yaml
+++ b/parallel.yaml
@@ -41,8 +41,7 @@ pipeline:
       ${{targets.contextdir}}/usr/bin/env_parallel.ksh \
       ${{targets.contextdir}}/usr/bin/env_parallel.mksh \
       ${{targets.contextdir}}/usr/bin/env_parallel.pdksh \
-      ${{targets.contextdir}}/usr/bin/env_parallel.tcsh \
-      ${{targets.contextdir}}/usr/bin/env_parallel.dash
+      ${{targets.contextdir}}/usr/bin/env_parallel.tcsh
 
 subpackages:
   - name: parallel-doc
--- a/parallel.yaml
+++ b/parallel.yaml
@@ -38,6 +38,7 @@ pipeline:
       ${{targets.contextdir}}/usr/bin/env_parallel.tcsh \
       ${{targets.contextdir}}/usr/bin/env_parallel.dash \
       ${{targets.contextdir}}/usr/bin/env_parallel.fish \
+      ${{targets.contextdir}}/usr/bin/env_parallel.sh \
       ${{targets.contextdir}}/usr/bin/env_parallel.zsh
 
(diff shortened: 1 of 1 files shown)


dannf:
Thanks @kranurag7!

I agree that we need to avoid pulling in all of those deps. But I do also see potential value in keeping those snippets around. I'm not sure if this is the *best* option - but we could just "go manual" and do something like:

```
options:
  no-depends: true
dependencies:
  runtime:
    - busybox
    - <anything else really needed>
```

Also, could you change the description of the -doc package to say `documentation` instead of just `manpages`?

parallel.yaml
```diff
@@ -35,7 +41,20 @@ pipeline:
       ${{targets.contextdir}}/usr/bin/env_parallel.ksh \
       ${{targets.contextdir}}/usr/bin/env_parallel.mksh \
       ${{targets.contextdir}}/usr/bin/env_parallel.pdksh \
-      ${{targets.contextdir}}/usr/bin/env_parallel.tcsh
+      ${{targets.contextdir}}/usr/bin/env_parallel.tcsh \
+      ${{targets.contextdir}}/usr/bin/env_parallel.dash \
+      ${{targets.contextdir}}/usr/bin/env_parallel.fish \
+      ${{targets.contextdir}}/usr/bin/env_parallel.sh \
+      ${{targets.contextdir}}/usr/bin/env_parallel.zsh
```
dannf, line 48: Since these will no longer result in a large dependency tree, can we leave them?

dannf: I think we should continue to omit the `env_parallel.*` for shells that we do not ship. But I think it would be good to continue shipping the scripts for the shells that we *do* ship. In which case, the number of provides should remain unchanged from the current version.

dannf: For the sake of completeness, I'll note that one option here is to move all of these snippets into individual subpackages - `parallel-shell-zsh`, `parallel-shell-mksh`, etc, each depending on the corresponding shell binary package. Each of these could provide `parallel-shell`, which `parallel` itself could depend on. That feels like overkill for this case, but worth explicitly mentioning even if we agree it's too much.

kranurag7: I have added it again for `fish`, `zsh` and `sh` shell. On adding separate compat subpackages, we can do that if we require it sometimes in future, for now I think we're not using them anywhere.

dannf: I see `dash` continues to be omitted, why is that?

kranurag7: missed it, now we are shipping env_parallel.dash too.

dannf (approved):

parallel.yaml
dannf: thanks!

PR #35344, opened 2024-11-27
dotnet-8: powershell: build separate package dotnet-sdk-8.0.404 to build powershell 7.4.6

ader1990:
Add dotnet-sdk only package 8.0.404 that is installed alongside 8.0.110.
The SDK  8.0.404 is required to build PowerShell 7.4.6

### DIFF
--- a/dotnet-sdk-8.yaml
+++ b/dotnet-sdk-8.yaml
@@ -1,0 +1,148 @@
+package:
+  name: dotnet-sdk-8-only
+  version: 8.0.404
+  epoch: 1
+  description: ".NET SDK"
+  copyright:
+    - license: MIT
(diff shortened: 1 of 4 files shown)


dotnet-sdk-8.yaml
```diff
@@ -0,0 +1,148 @@
+package:
+  name: dotnet-sdk-8-only
```
xnox, line 2: the .yaml filename and the first package name must match. I'm not sure if you wanted this with or without "-only".

ader1990: can we require in the build deps a specific version of an apk package? if so, that would be rather easy.

PR #35384, opened 2024-11-27
ruby3.X: faraday-net_http: fix runtimedeps

justinvreeland:
A runtime dep were mistakenly added as a test deps.

Faraday-net_http is the default faraday net provider and will be
automatically pulled in by faraday.  This packages is mostly uesless without
faraday so does not need to depend on it. This also avoids a dep cycle


roaksoax (approved):
lgtm!

PR #35385, opened 2024-11-27
Add tests to four packages that were missing them.

dlorenc:
Also fix the aerospike-7 build.


mepi262:
@dlorenc 
It seems that this pull request has conflicts.
Would you resolve it?

PR #35494, opened 2024-11-28
add lazydocker package

kranurag7:
continuation of an old trend to be little lazy in terminal https://github.com/wolfi-dev/os/pull/11922 :)

### DIFF
--- a/lazydocker.yaml
+++ b/lazydocker.yaml
@@ -17,7 +17,6 @@ pipeline:
     with:
       packages: .
       ldflags: |
-        -s -w
         -X main.version=${{package.version}}
         -X main.commit=$(git rev-parse HEAD)
         -X main.date=$(date -u +"%Y-%m-%dT%H:%M:%SZ")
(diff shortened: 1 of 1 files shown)


lazydocker.yaml
```diff
@@ -0,0 +1,37 @@
...
+    with:
+      repository: https://github.com/jesseduffield/lazydocker
+      tag: v${{package.version}}
+      expected-commit: be051153525b018a46f71a2b2ed42cde39a1110c
+
+  - uses: go/build
+    with:
+      packages: .
+      ldflags: |
+        -s -w
```
anushkamittal20, line 20: Do we need -w? I thought it is set by default 🤔

kranurag7: good catch, copy paste mistake from lazygit, correcting it.

kranurag7, line 20:
```suggestion

```

anushkamittal20 (approved):
Thanks for the work on this @kranurag7! LGTM

PR #35508, opened 2024-11-29
Add postgresql-client

anushkamittal20:


kranurag7 (changes requested):
given we want to build this again, this would require an epoch bump.

EDIT: looks like similar to  https://github.com/wolfi-dev/os/pull/35509

wojciechka:
I added that and symlinks in `/opt/bitnami` in #35508 35509

PR #35604, opened 2024-12-02
add pngcrush package

kranurag7:
required for discourse


pngcrush.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+pipeline:
+  - uses: fetch
+    with:
+      expected-sha256: 3b4eac8c5c69fe0894ad63534acedf6375b420f7038f7fc003346dd352618350
+      uri: https://downloads.sourceforge.net/pmt/pngcrush-${{package.version}}-nolib.tar.xz
+
+  - uses: patch
+    with:
+      patches: alder32.patch
```
EyeCantCU, line 26: Could you provide context for why we vendor this patch?

kranurag7: without this patch, we're running into build failure issues. I tried to pickup a patch upstream for this but wasn't able to find one. So commented out the part.

EyeCantCU: Nit, but could we vendor Alpine's patch instead?: https://gitlab.alpinelinux.org/alpine/aports/-/tree/master/community/pngcrush

EyeCantCU (approved):

PR #35623, opened 2024-12-02
add psmisc package

kranurag7:
required for discourse


kranurag7:
the `malcontent` looks like a false positive `killall` is the name of the binary that comes from this package.

EyeCantCU (approved):

PR #35640, opened 2024-12-02
Adding compat package to mirror upstream image

utieyin:
Compat package to allow calling the splunk otel binary from a different location

Fixes:

Related:

### DIFF
--- a/splunk-otel-collector.yaml
+++ b/splunk-otel-collector.yaml
@@ -1,7 +1,7 @@
 package:
   name: splunk-otel-collector
   version: 0.113.0
-  epoch: 1
+  epoch: 2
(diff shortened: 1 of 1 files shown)


splunk-otel-collector.yaml
```diff
@@ -40,6 +40,12 @@ pipeline:
       cp cmd/otelcol/config/collector/ecs_ec2_config.yaml ${{targets.contextdir}}/etc/otel/collector/ecs_ec2_config.yaml
 
 subpackages:
+  - name: ${{package.name}}-compat
+    description: Symlink for the otel binary in the home folder
```
EyeCantCU, line 44: This is at root, not in home

EyeCantCU (approved):

splunk-otel-collector.yaml
utieyin: @EyeCantCU This was a mixup. Upstream expects the binary in the root directory. Fixed the description

PR #35642, opened 2024-12-02
repmgr: remove non exist package

Dentrax:
Fixes `could not find package that provides pgaudit-16 in indexes` on HEAD


Dentrax:
Not its version streamed, lets close.

PR #35651, opened 2024-12-02
postgresql-bitnami: provide missing dirs

Dentrax:
On the upstream postgres-ha image, there is conf.d dirs maintained. And also those packages were missing lots of other binaries, so let's symlink the entire bin and lib dirs for .so files.

This is not a breaking change.


kranurag7 (approved):
thank you!!

PR #35679, opened 2024-12-03
Refactor docker package to reuse in fips image

anushkamittal20:
Refacored docker package- removed docker dependency to reuse in fips image. 

Note: Some tests were removed as well, checking for version/help. I have verified no image used these subpackages so we will just have to be careful with our docker and docker fips image module


kranurag7 (changes requested):
needs epoch bump as well.

developer-guy (approved):

kranurag7 (approved):

PR #35742, opened 2024-12-03
feat(ruby): Make coinstallable

EyeCantCU:
This makes Ruby 3.1-3.3 coinstallable, allowing for multiple versions of Ruby to be installed at the same time. This will help us simplifiy a lot of package manifests we have for Ruby as we'll be able to build gems for multiple Ruby versions within the same definition, without creating an entirely separate manifest

To full leverage this, we'll need to make SCA changes: chainguard-dev/melange#1676

And then rebuild the world for Ruby (purging the cmd:ruby dependency that all packages have right now, causing the resolver to fail unless you explicitly request a version of Ruby that aligns with the gem)

### DIFF
--- a/ruby-3.3.yaml
+++ b/ruby-3.3.yaml
@@ -1,14 +1,16 @@
 package:
   name: ruby-3.3
   version: 3.3.6
-  epoch: 0
-  description: "the Ruby programming language"
+  epoch: 1
+  description: "The Ruby ${{vars.major-minor-version}} programming language"
   copyright:
     - license: Ruby
     - license: BSD-2-Clause
   dependencies:
     provides:
       - ruby=${{package.full-version}}
+    runtime:
+      - ${{package.name}}-base=${{package.full-version}}
 
 environment:
   contents:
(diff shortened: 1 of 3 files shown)


EyeCantCU:
Note, making these changes now, before SCA changes are merged, will not impact existing packages. But after SCA changes land, it would be ideal that we rebuild the Ruby world

ruby-3.3.yaml
```diff
@@ -88,42 +99,74 @@ pipeline:
...
     pipeline:
       - uses: split/dev
+        with:
+          package: "${{package.name}}-base"
     test:
       pipeline:
         - uses: test/pkgconf
 
+  # Empty package, ensures ruby-dev always installs development files for latest ruby version
+  - name: "${{package.name}}-dev"
```
smoser, line 162: This is fine, but I think we are generally trying to move away from "virtual provides".
https://github.com/wolfi-dev/os/issues/32874 describes some reasons.  Those reasons do not completely line up with ruby-dev (using full-version here is _fine_, but using full-version in a ruby-pkgname is _not_ fine), but the point "virtual provides are only for humans" is relevant here.

So, i'm fine to keep this here, but I think we probably want to move away from virtual-provides like this.

smoser (approved):

ruby-3.3.yaml
EyeCantCU: I wasn't actually aware that is a problem with virtual provides. I had always assumed that the highest version always took precedence. Completely agree that these should never be used when building images or as explicit dependencies in packages

PR #35761, opened 2024-12-03
Add QEMU package

89luca89:
This is useful for our melange QEMU builds

### DIFF
--- a/qemu.yaml
+++ b/qemu.yaml
@@ -1,0 +1,87 @@
+package:
+  name: qemu
+  version: 9.1.2
+  epoch: 1
+  description: "fast processor emulator"
+  copyright:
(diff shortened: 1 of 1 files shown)


git-lfs.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: git-lfs
   version: 3.6.0
-  epoch: 1
+  epoch: 2
```
EyeCantCU, line 4: This should be split out

89luca89: Sorry needed to rebase, solved now

EyeCantCU (approved):

mattmoor:
Awesome, this should make it easier to leverage the `melange` QEMU runner 🤞

PR #35770, opened 2024-12-03
repmgr: rename package for version stream with postgresql

Dentrax:
* prepare package for version stream
* omit empty dev package


repmgr-17.yaml
```diff
@@ -216,11 +211,10 @@ test:
   environment:
     contents:
       packages:
-        - postgresql-16
+        - repmgr
```
kranurag7, line 214:
```suggestion
        - repmgr-17
```
should we also pull in specific `repmgr-17` here for testing? I was mostly thinking on line that if we copy paste then this will always pull in the latest?

Dentrax: `repmgr-17` package already provides `repmgr` but let's omit it, since `test:` pipeline fetches the _current_ package already. Thanks.

kranurag7 (approved):

PR #35796, opened 2024-12-04
Add run time dep for docker

anushkamittal20:
Updates runtime deps for docker

### DIFF
--- a/docker.yaml
+++ b/docker.yaml
@@ -11,9 +11,9 @@ package:
       # Include busybox because dockerd-oci-entrypoint is a shell script that execs the input params given.
       - busybox
       - ca-certificates
+      - containerd
       # let's pin this to containerd version < 2 because containerd 2.0.0 had some breaking and we can move to 2.0.0 once upstream image (library/docker:dind) moves.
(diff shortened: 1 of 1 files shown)


docker.yaml
```diff
@@ -1,15 +1,39 @@
...
   description: A meta package for Docker Engine and Docker CLI
   copyright:
     - license: Apache-2.0
   dependencies:
     runtime:
+      - btrfs-progs
+      # Include busybox because dockerd-oci-entrypoint is a shell script that execs the input params given.
+      - busybox
+      - ca-certificates
+      # let's pin this to containerd version < 2 because containerd 2.0.0 had some breaking and we can move to 2.0.0 once upstream image (library/docker:dind) moves.
```
kranurag7, line 14: Let's verify if we're able to make this work with updated containerd version, if yes we can bump the containerd version without pinning it explicitly.

anushkamittal20: I tested in the image, looks like it worked! Thanks for the review @kranurag7 !

EyeCantCU (approved):

PR #35825, opened 2024-12-04
font-ubuntu: glow up

xnox:
Use git-ubuntu repositories for build. Update to latest spdx license
version. Enable git monitoring.


xnox:
Update check failed =(

PR #35827, opened 2024-12-04
Fail the pkgconf test if the package does not contain any .pc files

murraybd:
There are some packages which have included the pkgconf test pipeline but the test doesn't actually do anything because the package does not provide any .pc files. So running the test doesn't accomplish anything and could actually skew stats leading us to believe there are more tests than there really are.


murraybd:
Given that the package tests are not run pre-submit the test failing will not block merge proposals and slow down development so I think exiting 1 is fine.

joshrwolf (approved):

PR #35853, opened 2024-12-04
add a really good test

joshrwolf:
@murraybd 

#13623

### DIFF
--- a/libsass.yaml
+++ b/libsass.yaml
@@ -48,3 +48,124 @@ update:
   enabled: true
   github:
     identifier: sass/libsass
+
+test:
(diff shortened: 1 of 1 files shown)


murraybd:
I think it'd be helpful to label these tests as "AI generated" so that future travelers have that context when reviewing or debugging them.

joshrwolf:
> I think it'd be helpful to label these tests as "AI generated" so that future travelers have that context when reviewing or debugging them.

good catch, I usually always link these to #13623 but forgot to do so this time, updated!

murraybd (approved):

PR #35854, opened 2024-12-04
Drop pkgconf tests that fail because the package provides no .pc files

dannf:
Determined by grep'ing the YAML for 'test/pkgconf', running the tests for matching packages, and then grep'ing those logs for the string "please remove the pkgconf test".


murraybd:
The package build never actually happened in production. Perhaps if you merge with main and push your branch again the tests will really run?

murraybd (approved):

PR #35869, opened 2024-12-04
ruby 3.4: interpreter & standard gems

justinvreeland:
Add ruby 3.4 bootstrap that has all of the default and bundled gems so that we can build ruby 3.4 and the same bundled and default gems ourselves.   I have not split out all of them here yet. I focused on the gems that we'd already separately packaged and dependencies to build those gems.

Also add a meta package to make it easy to install the set of standard gems people would expect.

The standard gems do change between releases so unfortunately we can't use the same meta package for all of our ruby versions.


justinvreeland:
Huh I can deal with the package update config stuff easily but I don't understand why the build failed.  It says:
```
failed to build package: unable to build guest: unable to lock image configuration: resolving apk packages: for arch "amd64": solving "ruby3.4-bundler" constraint: could not find package that provides ruby3.4-bundler in indexes
```
But bundler did build and locally I seem to be building fine.. though I also thought I'd built the previous ones locally without issues.

PR #35880, opened 2024-12-04
remove GHA workflows

imjasonh:
ci-build.yaml has been disabled since Saturday, and build.yaml since September. This PR "burns the boats" and lets us drop the GH runner infra needed to run these builds.


jonjohnsonjr (approved):

tuananh:
@imjasonh may i ask why we do this ?

PR #26482, opened 2024-08-16
pybind11: py versioned

pnasrat:
merge dev and versioned packages

validated prior package pybind11-config was using the site-package installed includes and cmakedirs not the duplicate ones in /usr

bump packages building with pybind11

Signed-off-by: Pris Nasrat <pris.nasrat@chainguard.dev>

PR #35958, opened 2024-12-05
py3-pybind11: Fix priorities of provides

dannf:
When py3-bind11 was python multi-versioned, all provides for py3-pybind11 and py3-pybind11-dev were assigned the same priority. This looks like just a cut & paste error - fix it.

Fixes: commit d76de48b0 ("pybind11: py versioned (#26482)")


pnasrat (approved):
LGTM

PR #36013, opened 2024-12-06
openjdk-24: enable linkable runtime

xnox:
Enable support for [JEP-493](https://openjdk.org/jeps/493)

Once this lands, can figure out image tests & package tests & dependencies to assert and validate this works.

Note, needs pandoc to build manpages now, which we don't have as we don't have haskell.


xnox:
Merged upon 3rd attempt:

PR #36047, opened 2024-12-07
py3-tensorflow-metadata: python multiversioning

dannf:
related: https://github.com/chainguard-dev/internal-dev/issues/5334

### DIFF
--- a/py3-tensorflow-metadata.yaml
+++ b/py3-tensorflow-metadata.yaml
@@ -1,16 +1,24 @@
 package:
   name: py3-tensorflow-metadata
   version: 1.16.1
-  epoch: 0
+  epoch: 1
   description: Utilities for passing TensorFlow-related metadata between tools
   copyright:
     - license: Apache-2.0
   dependencies:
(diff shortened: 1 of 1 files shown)


py3-tensorflow-metadata.yaml
```diff
@@ -34,37 +41,69 @@ pipeline:
...
+            python: python3.13
+            import: ${{vars.import}}
 
 update:
   enabled: true
   manual: false
   github:
     identifier: tensorflow/metadata
     strip-prefix: v
-
```
justinvreeland, line 51: I'd prefer we keep a top level tests so that when the package with no version is installed we know *a* compatible version was installed.

justinvreeland (dismissed):

py3-tensorflow-metadata.yaml
dannf: good call. adding that back actually highlighted that I'd missed a provides.

justinvreeland (approved):

PR #36053, opened 2024-12-07
add openssl-config to docker runtime deps

kranurag7:
else we run into following: 
```bash
Can't open "/etc/ssl/openssl.cnf" for reading, No such file or directory
803B09B2977B0000:error:80000002:system library:BIO_new_file:No such file or directory:crypto/bio/bss_file.c:67:calling fopen(/etc/ssl/openssl.cnf, r)
803B09B2977B0000:error:10000080:BIO routines:BIO_new_file:no such file:crypto/bio/bss_file.c:75:
```

### DIFF
--- a/docker.yaml
+++ b/docker.yaml
@@ -1,7 +1,7 @@
 package:
   name: docker
   version: 27.3.1
-  epoch: 6
+  epoch: 7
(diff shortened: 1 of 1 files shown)


kranurag7:
closing this is addressed already via other patch.

PR #36071, opened 2024-12-07
tensorflow-core: Add python multiversion support

dannf:
wolfi currently contains`py3.10-tensorflow-core.yaml` and `tensorflow-core.yaml`, which builds the libs and a `py3-tensorflow-core` that was built for python 3.12. This PR adds python-multiversion support to `tensorflow-core.yaml` so that it emits `py3.10`, `py3.11` and `py3.12`  packages, which lets us get rid of `py3.10-tensorflow-core.yaml`.

Both were missing basic runtime dependencies that needed to be installed by hand. `py3.10-tensorflow-core` would smoke test w/ `import tensorflow`, but `py3-tensorflow-core`, which was nominally built for Python 3.12, couldn't do a basic `import tensorflow`:

```
4c947c77c612:/work/packages# python3.12 -c 'import tensorflow'
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.12/site-packages/tensorflow/__init__.py", line 40, in <module>
    from tensorflow.python import pywrap_tensorflow as _pywrap_tensorflow  # pylint: disable=unused-import
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/tensorflow/python/pywrap_tensorflow.py", line 34, in <module>
    self_check.preload_check()
  File "/usr/lib/python3.12/site-packages/tensorflow/python/platform/self_check.py", line 63, in preload_check
    from tensorflow.python.platform import _pywrap_cpu_feature_guard
ImportError: Python version mismatch: module was compiled for Python 3.10, but the interpreter version is incompatible: 3.12.8 (tags/v3.12.8-1-g9208400-dirty:9208400, Dec 10 2024, 20:13:08) [GCC 14.2.0].
```

So the risk of regression here is really just with `py3.10-tensorflow-core`. The version of that package that we'll have after this PR will smoke test, but its contents are not identical. Different packages are vendored vs. system, and various `/usr/bin` commands that were there before have been removed - those were commands provided by vendored libraries (like `wheel`), and really didn't belong in this package.


tensorflow-core.yaml
```diff
@@ -41,19 +54,44 @@ environment:
...
+      - py3-supported-keras-preprocessing
+      - py3-supported-packaging
+      - py3-supported-pybind11
+      - py3-supported-requests
+      - py3-supported-setuptools
       - py3-wheel
       - python-3-dev
       - wolfi-base
       - zlib-dev
+  environment:
```
justinvreeland, line 70: ahh nice change here!

justinvreeland (approved):

tensorflow-core.yaml
dannf: thanks!

flatbuffers.yaml
```diff
@@ -2,11 +2,23 @@
...
+  pypi-package: flatbuffers
+  import: flatbuffers
+
+data:
+  - name: py-versions
+    items:
+      3.10: "310"
+      3.11: "311"
+      3.12: "312"
+      3.13: "300"
```
smoser, line 20: 3.13 is what all the cool kids are running now.

PR #36152, opened 2024-12-09
Refactor docker.yaml

anushkamittal20:
Add openssl-config, remove fuse-overlayfs

EDIT: we need fuse-overlayfs as a backup if overlay2 fails.

Added iproute2 as we get an unnecessary error in container logs otherwise

### DIFF
--- a/docker.yaml
+++ b/docker.yaml
@@ -1,7 +1,7 @@
 package:
   name: docker
   version: 27.4.0
-  epoch: 1
+  epoch: 4
   description: A meta package for Docker Engine and Docker CLI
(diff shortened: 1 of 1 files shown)


docker.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: docker
   version: 27.4.0
-  epoch: 0
+  epoch: 4
```
kranurag7, line 4:
```suggestion
  epoch: 1
```
this should be 1 💭

kranurag7 (approved):
great work!! thanks.

PR #36183, opened 2024-12-09
Mamba split

dannf:
We've been using upstream date-based tags for our versions, but that's apparently not what those tags are for. Currently that has CI wanting to move us *backwards*. We should use the semver tags instead.

Since `libmamba`, `libmambapy` and `micromamba` are all tagged separately, it's not clear which tag prefix we should be using. My thought was that we should treat them as separate sources even though they are from the same repo, and even though their releases always seem to point to the same hash.

If we did that, we'd build `libmamba` and `libmamba-dev` from `libmamba.yaml`, and then have `micromamba.yaml`, `py3-libmambapy.yaml`, and `mamba-package.yaml` files that build those binaries against the existing `libmamba-dev`.

But there are some problems here.

1. I don't think we have a defined way to migrate from `2024.10.02` to `2.0.4` version strings, since they sort lesser. We discussed options here, and withdrawing the date-versioned apks seems like the way to go.
2. `mamba-package` doesn't have a tag stream. I'm not sure which tag stream we should be using to build it.
3. `micromamba` is supposed to be statically linked, according to https://github.com/mamba-org/mamba?tab=readme-ov-file#micromamba. Ours is not, and looks like it will require some CMake work to do so.


philroche:
Latest package update PR @ https://github.com/wolfi-dev/os/pull/36642

philroche:
Also see slack thread (internal to chainguard) RE discussion static vs dynamic https://chainguard-dev.slack.com/archives/C0627BW5GLF/p1730889041303439

debasishbsws:
There is a Latest update PR https://github.com/wolfi-dev/os/pull/38633

cmwilson21:
Thanks for the contribution!

PR #36199, opened 2024-12-10
[draft] ruby3.4 preview2

justinvreeland:


justinvreeland:
It seems like the build is timing out. I'll split these into smaller PRs to make them pass

justinvreeland:
It's weird that I can build and install all of these packages in ~10mins locally but it looks like we're being timed-out or stuck in our CI for hours. Maybe it killed docker somehow? I'll look into the dependencies and split this up so it can pass CI.

```
ERRO 🔗 errors pretty printing info pkg=ruby3.4-stringio arch=arm64
ERRO 🔗 ERROR: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running? pkg=ruby3.4-net-http arch=arm64
ERRO 🔗 errors pretty printing info pkg=ruby3.4-net-http arch=arm64
```

PR #36210, opened 2024-12-10
kyverno-notation-aws package build

developer-guy:
Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/kyverno-notation-aws.yaml
+++ b/kyverno-notation-aws.yaml
@@ -1,0 +1,48 @@
+#nolint:git-checkout-must-use-github-updates,valid-pipeline-git-checkout-tag
+package:
+  name: kyverno-notation-aws
+  version: 1.0.1_git20241210
+  epoch: 0
+  description: Kyverno extension service for Notation and the AWS signer
+  copyright:
+    - license: Apache-2.0
+
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: efda86b8e6a97938c3caafcb5c98a5aede2b97e1
+      repository: https://github.com/nirmata/kyverno-notation-aws
+      branch: main
+
+  - uses: go/bump
+    with:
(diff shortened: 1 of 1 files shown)


kyverno-notation-aws.yaml
```diff
@@ -0,0 +1,49 @@
...
+      modroot: .
+      output: kyverno-notation-aws
+      packages: .
+
+subpackages:
+  - name: ${{package.name}}-compat
+    description: "Kyverno extension service for Notation and the AWS signer (compatibility)"
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/
```
kranurag7, line 33:
```suggestion

```
nit: we can omit this line given we're already creating a nested directory on the next line.

kyverno-notation-aws.yaml
```diff
@@ -0,0 +1,49 @@
...
+          mkdir -p ${{targets.contextdir}}/plugins/com.amazonaws.signer.notation.plugin
+          ln -sf /usr/bin/kyverno-notation-aws ${{targets.contextdir}}/kyverno-notation-aws
+          ln -sf /usr/bin/notation-com.amazonaws.signer.notation.plugin ${{targets.contextdir}}/plugins/com.amazonaws.signer.notation.plugin/notation-com.amazonaws.signer.notation.plugin
```
kranurag7, line 36: should we include https://github.com/wolfi-dev/os/blob/main/aws-signer-notation-plugin.yaml as a runtime dependency to this subpackage if they're related else we can pull in image without being specific to a version which is always latest.

developer-guy: I can do that but I was planning to use the same -compat sub-package for FIPS image, this is why I didn't specify it as a runtime dependency, does that make sense?

kranurag7: perfect that looks good to me.

kranurag7 (approved):
thanks for being patient with reviews. Looks good.

kyverno-notation-aws.yaml
developer-guy: removed brother

PR #36285, opened 2024-12-10
python-3.10, python-3.11 - build tkinter with python

smoser:
This is similar to the python-3.12 and python-3.12 change to build tkinter inside python 
at https://github.com/wolfi-dev/os/pull/29194.

The big difference is that we have to use tcl and tk at major version 8 because
python-3.10 and python-3.11 do not support tcl/tk at version 9.

The proper way to do this would be to version tcl and tk (tcl-8.yaml, tk-8.yaml).

In order to correctly use tkinter (and pass the provided test), we have to add explicit runtime tcl and tk dependencies because melange does not identify the dependency that

### DIFF
--- a/python-3.10.yaml
+++ b/python-3.10.yaml
@@ -1,7 +1,7 @@
 package:
   name: python-3.10
   version: 3.10.16
-  epoch: 0
+  epoch: 1
   description: "the Python programming language"
   copyright:
     - license: PSF-2.0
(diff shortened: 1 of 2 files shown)


smoser:
**Note**: it is perfectly acceptable to not take this.  We would lose tkinter support in python-3.10 and python-3.11, but the support we had didn't work very well and had FTBFS ever since tcl and tk moved to version 9.

dannf:
I'm a +0 on doing this at all, bu I feel like if we're going to do it, we should go ahead and version stream tk.

smoser:
I'm going to close this for now. 
If we had a need for python-3.10 and python-3.11 with tkinter, then @dannf 's suggestion is the right way to do it.

Thanks for the review.

PR #36325, opened 2024-12-11
don't require build maintainers approval

imjasonh:
Historically, we imposed a limit on who could approve changes in `Makefile` and `.github/workflows/` to prevent PRs from breaking CI.

Now that builds happen in elastic-build, it's a lot harder for randos to break the build (not impossible; _this does not constitute a challenge_), and we should let others contribute to the Makefile and GHA workflows. The blast radius of changes to those are a lot smaller now.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,7 +1,3 @@
-# Require review by build system owners for workflow and Makefile changes.
-Makefile @wolfi-dev/wolfi-build-maintainers
-/.github/workflows/ @wolfi-dev/wolfi-build-maintainers
-
 # Require review by repo owners of changes to CODEOWNERS
 CODEOWNERS @wolfi-dev/wolfi-owners
 


cpanato (approved):

EyeCantCU (approved):

joshrwolf (approved):
I agree, this shouldn't need **build** maintainers review anymore which has historically been specifically for CI related things, but there's definitely still places in the `Makefile` that need reviewers

imjasonh:
> I agree, this shouldn't need **build** maintainers review anymore which has historically been specifically for CI related things, but there's definitely still places in the `Makefile` that need reviewers

Yeah I think it makes sense to require some special folks to approve `Makefile` changes, I'm just not sure it needs to be only me, @pdeslaur or @mattmoor to approve those.

It probably makes sense to have that be a responsibility of @murraybd @roaksoax @dannf @sil2100 's team?

pdeslaur (approved):

PR #36408, opened 2024-12-11
ruby 3.4: interpreter & gem batch 1

justinvreeland:
This was created by renaming the 3.3 file to 3.4.

Turning the gemspec patches into files. See https://github.com/wolfi-dev/os/blob/main/ruby3.2-bouncy-castle-java/0001-add-gemspec.patch

Providing the base64, csv, and drb package from the main ruby package because they now conflict and some packages have implicit dependencies on them. Which means we either need to spot update failing packages or provide it from the interpreter that is already installing these gems.


smoser (approved):
thanks.

justinvreeland:
If you want to see the actual diff from the 3.3 copy you can look at https://github.com/justinvreeland/wolfi/commit/21c0c0d8850c62d6d534441db0c28d15e114bbc4

imjasonh (approved):
This will need a force merge, since this many files makes `ci-sbom-validity` time out and die.

PR #36409, opened 2024-12-11
Fix: CVE GHSA-rcjc-c4pj-xxrp in druid

debasishbsws:
This fix needs java version upgrade to 21 which is not what upstream mentioned https://github.com/apache/druid/blob/master/docs/operations/java.md\#selecting-a-java-runtime

### DIFF
--- a/druid.yaml
+++ b/druid.yaml
@@ -1,7 +1,7 @@
 package:
   name: druid
   version: 31.0.0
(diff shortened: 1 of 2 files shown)


debasishbsws:
Druid requires Java 8, 11, or 17. Your current version is: 21.0.5. can't remediate  this CVE

PR #36330, opened 2024-12-11
Make python-3.13 version of PYPI_PKG provide virtual py3-PYPI_PKG

smoser:
Make python-3.13 version of PYPI_PKG provide virtual py3-PYPI_PKG

This just makes the python v3.13 version of a package the highest
version.  So now, if you do `apk add py3-appnope`, you'll get
`py3.13-appnope` where before you would get py3.12-appnope.

The 'set-313.sh' script used to do this looks like:

    #!/bin/sh -e
    r=' \(["'\'']*3[.]13["'\'']*\): \(["'\'']*\)[0-9]\+\(["'\'']*\)'
    n=313

    files=$(git grep -l "$r")
    set -- $files
    echo "$# files"
    sed -i -e "s,$r, \1: \2${n}\3," "$@"
    wolfictl bump "$@"

I then filtered out those packages that were failing to build/test:
  py3-ml-metadata apache-arrow py3-meson-python py3-build
  py3-google-apitools py3-cairo libmamba

And those that caused issues with the build solver:
  py3-pip py3-supported-python py3-wheel

PR #36414, opened 2024-12-11
py3-ml-metadata - make py3.13-ml-metadata the default.

smoser:
See similar changes https://github.com/wolfi-dev/os/pull/36330


OddBloke:
https://github.com/wolfi-dev/os/pull/38016 applied this same change.

PR #36419, opened 2024-12-11
Update python 3.13 priority in 4 packages.

smoser:
These packages did not "just work" with some sed changes to move 3.13 as the highest priority python packages.

py3-build - update python 3.13 version, fix test, improve test 
py3-cairo -  need multiple versions of the py3-cairo-dev 
libmamba - drop python test from base package
py3-google-apitools - do not build 3.13, does not work 

See similar changes #36330

### DIFF
--- a/py3-cairo.yaml
+++ b/py3-cairo.yaml
@@ -87,19 +87,6 @@ subpackages:
           packages:
             - py3-supported-${{vars.pypi-package}}
 
-  - range: py-versions
-    name: py${{range.key}}-${{vars.pypi-package}}-dev
-    description: py${{range.key}} dev files for ${{vars.pypi-package}}
-    dependencies:
-      runtime:
-        - cairo-dev
-        - py${{range.key}}-${{vars.pypi-package}}
-      provides:
-        - py3-${{vars.pypi-package}}-dev
-      provider-priority: ${{range.value}}
-    pipeline:
-      - uses: split/dev
-
   - name: py3-supported-${{vars.pypi-package}}
     description: meta package providing ${{vars.pypi-package}} for supported python versions.
     dependencies:
--- a/py3-cairo.yaml
+++ b/py3-cairo.yaml
@@ -1,7 +1,7 @@
 package:
   name: py3-cairo
   version: 1.27.0
-  epoch: 2
+  epoch: 3
   description: Python3 bindings for the cairo graphics library
   copyright:
     - license: LGPL-2.0-or-later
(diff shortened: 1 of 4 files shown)


smoser:
Note, the package update config check is bogus.  mamba https://github.com/mamba-org/mamba/tags uses semantic versioning, but somehow we got stuck on their YYYY.MM.DD based releases.

xnox:
Note @dannf was proposing to downgrade mamba

py3-cairo.yaml
```diff
@@ -114,23 +127,6 @@ subpackages:
...
-    pipeline:
-      - uses: split/dev
-    test:
-      pipeline:
-        - uses: test/pkgconf
-        - uses: py/one-python
-          with:
-            content: |
-              include="$(python3 -c 'import ${{vars.import}}; print(${{vars.import}}.get_include())')"
-              test -f "$include/py3cairo.h"
```
dannf, line 132: The reason for this package was to keep the `/usr/include/pycairo/py3cairo.h` file around. I'm not sure if anything uses it, but it could be a regression. The test doesn't seem appropriate though.

py3-cairo.yaml
```diff
@@ -80,13 +80,26 @@ subpackages:
...
+    description: py${{range.key}} dev files for ${{vars.pypi-package}}
+    dependencies:
+      runtime:
+        - cairo-dev
+        - py${{range.key}}-${{vars.pypi-package}}
+      provides:
+        - py3-${{vars.pypi-package}}-dev
+      provider-priority: ${{range.value}}
+    pipeline:
+      - uses: split/dev
```
dannf, line 101: When I build this locally, these are empty.

dannf (dismissed):
Thanks @smoser ! I highlighted a couple of things w/ pycairo inline.

py3-cairo.yaml
smoser: then we should just straight out test for that path. because cairo.get_include() returns `/usr/lib/python3.13/site-packages/cairo/include`

```
$ apk add py3.13-cairo
...
(26/27) Installing cairo (1.18.2-r1)
(27/27) Installing py3.13-cairo (1.27.0-r2)
$  python3.13 -c 'import cairo; print(cairo.get_include())'
/usr/lib/python3.13/site-packages/cairo/include
```

py3-cairo.yaml
smoser: yeah, i guess. and i guess i'm fine to drop this then. 
i'll drop it.
thank you for paying attention.

py3-cairo.yaml
smoser: i think it makes most sense to just drop this 'py3-cairo-dev' package. if something needs it then we can add it. nothing in the archive right now seems to.

xnox (approved):

PR #36423, opened 2024-12-11
py3-meson-python - make virtual package default to python 3.13

smoser:


smoser:
shoved this change in to https://github.com/wolfi-dev/os/pull/36431 . it was py3-pyproject-metadata that was the problem.

PR #36430, opened 2024-12-11
Make python 3.13 version of virtual packages the default. - batch 03

smoser:
Make python-3.13 version of PYPI_PKG provide virtual py3-PYPI_PKG

This just makes the python v3.13 version of a package the highest
version.  So now, if you do `apk add py3-appnope`, you'll get
`py3.13-appnope` where before you would get py3.12-appnope.

The 'set-313.sh' script used to do this looks like:

    #!/bin/sh -e
    r=' \(["'\'']*3[.]13["'\'']*\): \(["'\'']*\)[0-9]\+\(["'\'']*\)'
    n=313

    files=$(git grep -l "$r")
    set -- $files
    echo "$# files"
    sed -i -e "s,$r, \1: \2${n}\3," "$@"
    wolfictl bump "$@"


justinvreeland (dismissed):

justinvreeland (approved):

imjasonh (approved):
Force merging since ci-apk-add and ci-sbom-validity don't currently like large PRs.

PR #36450, opened 2024-12-12
 ruby 3.4: gems batch 2

justinvreeland:
Copied ruby3.3 files to 3.4 sed'd the version, removed sed patch for the key that've been integrated into the pipeline. Also re add console test.


dannf (commented):
Most are all simple sed's w/ some patch drops as it says on the tin.

Of the others, the following have no `3.3` counterpart:

- ruby3.4-rspec-core.yaml
- ruby3.4-rspec-expectations.yaml
- ruby3.4-rspec-mocks.yaml

why are those being introduced?

and the following have additional changes:
- ruby3.4-faraday.yaml

This is using version 2.12.1, while ruby3.3 was at 2.12.2. It also dropped a lot of tests. Perhaps it was updated while your branch was in devel?

justinvreeland:
> Most are all simple sed's w/ some patch drops as it says on the tin.
> 
> Of the others, the following have no `3.3` counterpart:
> 
> * ruby3.4-rspec-core.yaml
> * ruby3.4-rspec-expectations.yaml
> * ruby3.4-rspec-mocks.yaml
> 
> why are those being introduced?
> 
> and the following have additional changes:
> 
> * ruby3.4-faraday.yaml
> 
> This is using version 2.12.1, while ruby3.3 was at 2.12.2. It also dropped a lot of tests. Perhaps it was updated while your branch was in devel?

Ahh thanks for finding that. I'm not sure why that was missing. Console was missing one too. I'll put up a copy and diff branch to double check.

justinvreeland:
> This is using version 2.12.1, while ruby3.3 was at 2.12.2. It also dropped a lot of tests. Perhaps it was updated while your branch was in devel?


Huh, I'm pushing my local branch but I'm pretty sure I haven't changed anything;

```
~/work/os jvreeland/ruby-batch-2 ≡
.venv ❯ git diff chainguard/main -- ruby3.3-faraday.yaml

~/work/os jvreeland/ruby-batch-2 ≡
.venv ❯ diff ruby3.3-faraday.yaml ruby3.4-faraday.yaml
2c2
<   name: ruby3.3-faraday
---
>   name: ruby3.4-faraday
```

justinvreeland:
@dannf I think we noticed the faraday thing at the same time. Looks like you caught it betwixt my updates. Thanks for the attentive review!

dannf (approved):
LGTM, thanks @justinvreeland !

PR #36573, opened 2024-12-12
Converted mercurial package to build with each supported python version

hbh7:
Fixes: Firefox needs an older Python version to resolve a build issue, but it also relies on Mercurial being built with that same older Python version. This provides those older python version builds without any impact on other packages using Mercurial. 

Related: https://github.com/wolfi-dev/os/pull/36188

I definitely learned a lot during this conversion, so please fact check everything I did to make sure I haven't missed anything. Thanks!

### DIFF
--- a/mercurial.yaml
+++ b/mercurial.yaml
@@ -50,7 +50,7 @@ subpackages:
         - py3-${{vars.pypi-package}}
         - mercurial
       runtime:
-        - py${{range.key}}
+        - python-${{range.key}}
     pipeline:
       - runs: |
           # Build the wheel for this python version
--- a/mercurial.yaml
+++ b/mercurial.yaml
@@ -59,13 +59,13 @@ subpackages:
             --output-fd 3 3>&1 >&2
 
           # Install the built wheel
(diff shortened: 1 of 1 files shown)


mercurial.yaml
```diff
@@ -31,27 +40,48 @@ pipeline:
...
-      for man in doc/*.?; do
-        install -Dm644 "$man" \
-          "${{targets.destdir}}"/usr/share/man/man${man##*.}/${man##*/}
-      done
+          # Copy the hg bins to the right location
+          install -Dm755 contrib/hg-ssh hgeditor -t "${{targets.subpkgdir}}/usr/bin"
 
-  - uses: strip
+          # Cleanup the wheel so the other subpackages don't try to install it too
+          rm -f dist/mercurial-6.9-cp*.whl
```
debasishbsws, line 68: This will result in failure in future when there will be a new version. `mercurial-${{package.version}}-cp*.whl` would be better

debasishbsws (changes requested):

mercurial.yaml
hbh7: Oops, of course it would, good catch, thanks!

debasishbsws (approved):
Looks good

PR #36593, opened 2024-12-12
openssh: support config snippets includes

xnox:
This allows to ship config snippets for hardened configurations.

This also creates config parity with Debian & Ubuntu.

### DIFF
--- a/openssh.yaml
+++ b/openssh.yaml
@@ -1,7 +1,7 @@
 package:
   name: openssh
   version: 9.9_p1
-  epoch: 2
+  epoch: 3
   description: "the OpenBSD SSH implementation"
   copyright:
     - license: ISC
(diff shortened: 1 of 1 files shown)


justinvreeland (approved):

openssh.yaml
```diff
@@ -122,6 +122,8 @@ subpackages:
           mkdir -p "${{targets.subpkgdir}}"/etc/ssh
           mv "${{targets.destdir}}"/etc/ssh/moduli "${{targets.subpkgdir}}"/etc/ssh/
           mv "${{targets.destdir}}"/etc/ssh/ssh_config "${{targets.subpkgdir}}"/etc/ssh/
+          mkdir -p "${{targets.subpkgdir}}"/etc/ssh/ssh_config.d
+          sed -i 's|\(^# Host\)|Include /etc/ssh/ssh_config.d/*.conf\n\n\1|' "${{targets.subpkgdir}}"/etc/ssh/ssh_config
```
smoser, line 126: this is not your fault. but so hard to read. we some sort of solution here.

maybe we should look at augeas/augtool or something? I've never really looked at it.

This is just so brittle. 
ssh_config looks like:
```
# Site-wide defaults for some commonly used options.  For a comprehensive
# list of available options, their meanings and defaults, please see the
# ssh_config(5) man page.

# Host *
#   ForwardAgent no
...
```

so you're adding 'Include /etc/ssh/ssh_config.d/*.conf' before '# Host *' line.
But if the preceeding commentn block changes to have a line like:

    # Hosts are ipv4 or ipv6 hostnames or addresses.

then we'll false-positive and double insert.

smoser (approved):

openssh.yaml
xnox: would you preffer it as a regular patch? or even a hard-coded ssh_config file that we package. Note there is no consensus what ssh_config should be and various distributions preset various things to their liking.

PR #36606, opened 2024-12-12
ruby 3.4: interpreter & gem batch .25

justinvreeland:


justinvreeland:
The update will be resolved when 3.4 is released and we're not pulling from the `_preview2` tag

justinvreeland:
The bigger batches work now

PR #36623, opened 2024-12-12
chore(gitsign,guac): fix ghsa-v778-237x-gjrc

maxgio92:
This commit bumps golang.org/x/crypto to 0.31.0

#### For security-related PRs

### DIFF
--- a/gitsign.yaml
+++ b/gitsign.yaml
@@ -1,7 +1,7 @@
 package:
   name: gitsign
   version: 0.11.0
-  epoch: 1
+  epoch: 2
   description: Keyless Git signing with Sigstore!
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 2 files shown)


kranurag7 (changes requested):
requires epoch bump given we want to build it again with bumped deps.

gitsign.yaml
```diff
@@ -37,6 +37,9 @@ subpackages:
   - name: gitsign-credential-cache
     description: "helper binary that allows users to cache signing credentials"
     pipeline:
+      - uses: go/bump
+        with:
+          deps: golang.org/x/crypto@v0.31.0
```
kranurag7, line 42: one go/bump should be enough, it's the same build environment. 💭

kranurag7 (approved):
thank you!!

PR #36624, opened 2024-12-12
Fix a ftbfs with meson due to an incorrect dep

murraybd:
Before this change the package build would fail in the following way:

```
2024-12-12T07:22:17Z INFO running step "py/pip-build-install"
2024-12-12T07:22:17Z INFO running step "pip build"
2024-12-12T07:22:17Z INFO /usr/bin/python3.12 is 3.12 with site_packages dir 'usr/lib/python3.12/site-packages'
2024-12-12T07:22:17Z INFO execute: /usr/bin/python3.12 -m pip wheel --verbose --wheel-dir=./.wheels/3.12 --find-links=/tmp/tmp.GprFUG/dist-wheels --no-index --no-build-isolation --no-deps .
2024-12-12T07:22:17Z WARN /usr/bin/python3.12: No module named pip
2024-12-12T07:22:17Z INFO deleting guest dir /tmp/melange-guest-130102806
2024-12-12T07:22:17Z INFO deleting workspace dir /tmp/melange-workspace-226502872
2024-12-12T07:22:17Z INFO removing image path /tmp/melange-guest-1547025742
2024-12-12T07:22:18Z ERRO failed to build package: unable to run package meson pipeline: unable to run pipeline: unable to run pipeline: exit status 1
```

### DIFF
--- a/meson.yaml
+++ b/meson.yaml
@@ -15,8 +15,8 @@ environment:
       - busybox
       - ca-certificates-bundle
       - openssf-compiler-options
-      - py3-supported-pip
       - py3-setuptools
+      - py3-supported-pip
       - python3
       - samurai
 
--- a/meson.yaml
+++ b/meson.yaml
@@ -1,7 +1,7 @@
 package:
   name: meson
   version: 1.6.0
-  epoch: 1
+  epoch: 2
   description: Fast and user friendly build system
   copyright:
     - license: Apache-2.0
--- a/meson.yaml
+++ b/meson.yaml
@@ -15,7 +15,7 @@ environment:
       - busybox
       - ca-certificates-bundle
       - openssf-compiler-options
-      - py3-pip
+      - py3-supported-pip
       - py3-setuptools
       - python3
       - samurai


dannf (commented):
You've hit an "archive out of sync" condition I think.
The problem is that `py3-pip` prefers `py3.13-pip`, while `python3` prefers `python3.12`. So yes, this will get the build working again, but it is fragile. The real answer is to multiversion this package. If you want a more robust bandaid, then you can also use `py3-supported-setuptools`. Or, even better, replace both `pip` and `setuptools` with `py3-supported-build-base`.

tuananh:
@dannf can you explain abit about the py3-supported-* convention? this one seems to be added recently?

dannf:
> @dannf can you explain abit about the py3-supported-* convention? this one seems to be added recently?

`py3-supported-X` is a metapackage that depends on all of the supported `py3.x-X` packages. For a python package `foo` that support all of the versions of python currently in wolfi, installing `py3-supported-foo` will bring in `py3.10-foo`, `py3.11-foo`, `py3.12-foo` and `py3.13-foo`. For this PR, it solves the problem by making sure there is a `py3.x-pip` installed for whatever version of python `python3` points to. But that's still just a partial solution. The full solution is to multiversion this package so that it builds a `py3.x-meson` for every supported version of python, and provides a `py3-supported-meson` package that depends on all of those. That way `meson` is available for any python stack, not just whatever happened to be the default version in wolfi the day it was built.

dannf:
@murraybd how about this instead? https://github.com/wolfi-dev/os/pull/36981

tuananh:
> > @dannf can you explain abit about the py3-supported-* convention? this one seems to be added recently?
> 
> `py3-supported-X` is a metapackage that depends on all of the supported `py3.x-X` packages. For a python package `foo` that support all of the versions of python currently in wolfi, installing `py3-supported-foo` will bring in `py3.10-foo`, `py3.11-foo`, `py3.12-foo` and `py3.13-foo`. For this PR, it solves the problem by making sure there is a `py3.x-pip` installed for whatever version of python `python3` points to. But that's still just a partial solution. The full solution is to multiversion this package so that it builds a `py3.x-meson` for every supported version of python, and provides a `py3-supported-meson` package that depends on all of those. That way `meson` is available for any python stack, not just whatever happened to be the default version in wolfi the day it was built.

so `py3-supported-x` will install all possible versions? I would imaging it causing images size to increase.

dannf:
> > > @dannf can you explain abit about the py3-supported-* convention? this one seems to be added recently?
> > 
> > 
> > `py3-supported-X` is a metapackage that depends on all of the supported `py3.x-X` packages. For a python package `foo` that support all of the versions of python currently in wolfi, installing `py3-supported-foo` will bring in `py3.10-foo`, `py3.11-foo`, `py3.12-foo` and `py3.13-foo`. For this PR, it solves the problem by making sure there is a `py3.x-pip` installed for whatever version of python `python3` points to. But that's still just a partial solution. The full solution is to multiversion this package so that it builds a `py3.x-meson` for every supported version of python, and provides a `py3-supported-meson` package that depends on all of those. That way `meson` is available for any python stack, not just whatever happened to be the default version in wolfi the day it was built.
> 
> so `py3-supported-x` will install all possible versions? I would imaging it causing images size to increase.

@tuananh they are only intended to be used as build-time dependencies for other apks, they should not be installed in images. What we're doing here is making our library of python packages *available* for whatever python version a given image needs whereas, in the past, python packages were typically only available for the current default python version in wolfi. 

We're implementing this by having every `py3-y` package build for every supported version of python. This produces  `py3.10-y`, `py3.11-y, `py3.12-y` and `py3.13-y` apks. However, to do those 4 builds, we need to have the python dependency packages for `py3-y` installed for each interpreter version. This is the use case for `py3-supported-x`. `py3-y` need only declare a build dependency on `py3-supported-x`, instead of explicitly listing that it needs `py3.10-x, `py3.11-x`, etc.

If anything, this should make images *smaller*. For example, take this PR. If you have an image for a `python3.10`-based app, and you also wanted to include the `meson` tool, today you'll also need the `python3.13` interpreter. This is because we only have 1 `meson` package, and it was built for `python3.13`:

```console
68c5d811206c:/work/packages# apk add python-3.10 meson
WARNING: opening /work/packages: No such file or directory
(1/18) Installing py3-pip-wheel (24.3.1-r1)
(2/18) Installing libbz2-1 (1.0.8-r9)
(3/18) Installing libexpat1 (2.6.4-r0)
(4/18) Installing libffi (3.4.6-r5)
(5/18) Installing gdbm (1.24-r1)
(6/18) Installing xz (5.6.3-r2)
(7/18) Installing libstdc++ (14.2.0-r6)
(8/18) Installing mpdecimal (4.0.0-r3)
(9/18) Installing ncurses-terminfo-base (6.5_p20241006-r4)
(10/18) Installing ncurses (6.5_p20241006-r4)
(11/18) Installing readline (8.2.13-r1)
(12/18) Installing sqlite-libs (3.47.2-r0)
(13/18) Installing python-3.12-base (3.12.8-r2)
(14/18) Installing samurai (1.2-r3)
(15/18) Installing meson (1.6.0-r1)
(16/18) Installing py3-setuptools-wheel (75.6.0-r0)
(17/18) Installing python-3.10-base (3.10.16-r0)
(18/18) Installing python-3.10 (3.10.16-r0)
Executing glibc-2.40-r3.trigger
Executing busybox-1.37.0-r0.trigger
OK: 100 MiB in 33 packages
68c5d811206c:/work/packages# 
```

Once this is merged, we'll have a `meson` binary for `python3.10`, `python3.11`, `python3.12` and `python3.13`, so we could install the one that uses the same interpreter as our app:

```console
43fd9788ad64:/work/packages# apk add python-3.10 meson py3.10-meson-bin
(1/18) Installing py3-pip-wheel (24.3.1-r1)
(2/18) Installing py3-setuptools-wheel (75.6.0-r0)
(3/18) Installing libbz2-1 (1.0.8-r9)
(4/18) Installing libexpat1 (2.6.4-r0)
(5/18) Installing libffi (3.4.6-r5)
(6/18) Installing gdbm (1.24-r1)
(7/18) Installing xz (5.6.3-r2)
(8/18) Installing libstdc++ (14.2.0-r6)
(9/18) Installing mpdecimal (4.0.0-r3)
(10/18) Installing ncurses-terminfo-base (6.5_p20241006-r4)
(11/18) Installing ncurses (6.5_p20241006-r4)
(12/18) Installing readline (8.2.13-r1)
(13/18) Installing sqlite-libs (3.47.2-r0)
(14/18) Installing python-3.10-base (3.10.16-r0)
(15/18) Installing py3.10-meson (1.6.0-r2)
(16/18) Installing samurai (1.2-r3)
(17/18) Installing py3.10-meson-bin (1.6.0-r2)
(18/18) Installing python-3.10 (3.10.16-r0)
Executing glibc-2.40-r3.trigger
Executing busybox-1.37.0-r0.trigger
OK: 64 MiB in 33 packages
43fd9788ad64:/work/packages# 
```

tuananh:
Thanks @dannf

PR #36708, opened 2024-12-12
Add test/ldd-check pipeline, replace ldd 'tests' with it.

smoser:
I happened to see the test that aws-c-s3 had in place
and realized that it does not actually notice failure.
Running `ldd` on a program or library with missing dependencies
will exit zero unless ldd itself fails.
    
The added 'test/ldd-check' pipeline will check and list missing
dependencies.

### DIFF
--- a/pipelines/test/ldd-check.yaml
+++ b/pipelines/test/ldd-check.yaml
@@ -1,0 +1,59 @@
+name: ldd-check
+
+needs:
+  packages:
+    - busybox
+    - posix-libc-utils
+
+inputs:
+  files:
+    description: |
+      The files to run `ldd` on and check for missing deps.
+    required: true
+  verbose:
+    description: |
+      Should the full ldd output be shown
+    required: false
+    default: false
+
+pipeline:
+  - name: "run ldd on provided files"
+    runs: |
+      set +x
+      set -f
+      error() { echo "ERROR[ldd-check]:" "$@"; exit 1; }
+      fail() { echo "FAIL[ldd-check]:" "$@"; fails=$((fails+1)); }
+      pass() { echo "PASS[ldd-check]:" "$@"; passes=$((passes+1)); }
+      cleanup() { [ -n "$tmpd" -o -z "$tmpd" ] && return 0; rm -Rf "$tmpd"; }
+
+      tmpd=$(mktemp -d) || fail "ERROR: failed to create tmpdir"
+      trap cleanup EXIT
+
+      fails=0
+      passes=0
+      files="${{inputs.files}}"
+      verbose="${{inputs.verbose}}"
+      case "$verbose" in
+        true|false) :;;
+        *) error "verbose must be 'true' or 'false'. found '$verbose'";;
+      esac
+
+      export LANG=C
+      set -- $files
+      outf="$tmpd/out"
+      for f in "$@"; do
+        [ -e "$f" ] || { fail "$f: does not exist"; continue; }
+        [ -f "$f" ] || { fail "$f: not a file"; continue; }
+        ldd "$f" > "$outf" || { fail "$f: ldd exited $?"; continue; }
+        missing=$(awk \
+          '$0 ~ /=> not found/ { miss = miss " " $1; }; END { printf("%s\n", miss); }' \
+          "$outf") || error "$f: parsing with awk failed $?";
+        if [ "$verbose" = "true" ]; then
+          echo "> $ ldd $f"
+          sed 's,^,> ,' "$outf"
+        fi
+        [ -z "$missing" ] && { pass "$f"; continue; }
+        fail "$f: missing ${missing# }"
+      done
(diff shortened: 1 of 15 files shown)


pipelines/test/ldd-check.yaml
```diff
@@ -0,0 +1,59 @@
...
+  verbose:
+    description: |
+      Should the full ldd output be shown
+    required: false
+    default: false
+
+pipeline:
+  - name: "run ldd on provided files"
+    runs: |
+      set +x
```
justinvreeland, line 22: .... did you mean to leave this here?

justinvreeland (approved):

pipelines/test/ldd-check.yaml
```diff
@@ -0,0 +1,59 @@
...
+needs:
+  packages:
+    - busybox
+    - posix-libc-utils
+
+inputs:
+  files:
+    description: |
+      The files to run `ldd` on and check for missing deps.
+    required: true
```
dannf, line 12: Thanks for doing this! Is there a way to use this interface w/o knowing the list of files in advance? I'd coincidentally added the following test to a package in my local tree:

```yaml
test:
  environment:
    contents:
      packages:
        - apk-tools
        - posix-libc-utils # for ldd
    environment:
      LD_LIBRARY_PATH: /usr/local/cuda-12.5/lib64:/usr/local/cudnn-9/lib64:/opt/tensorrt_llm/libs:/opt/tritonserver/lib
  pipeline:
    - name: Scan .so files for undetected runtime dependencies
      runs: |
        ret=0
        apk info -L "${{package.name}}" | grep \.so$ | sed 's,^,/,' | while read obj; do
          echo -n "Scanning $obj... "
          if ldd "$obj" | grep "not found"; then
            echo "failed"
            ret=1
            continue
          fi
          echo "ok"
        done
        exit $ret
```

Here, I don't know the `.so*` files in advance, so I'm just searching the entire package. Ultimately, I think we should teach `melange` to `shlibdeps` these on its own, allowing users to add exceptions.

pipelines/test/ldd-check.yaml
```diff
@@ -0,0 +1,59 @@
...
+  verbose:
+    description: |
+      Should the full ldd output be shown
+    required: false
+    default: false
+
+pipeline:
+  - name: "run ldd on provided files"
```
dannf, line 20: Can you document if the files should be provided in a YAML list, or a space-separated list?

vaikas: Just curious if we anticipate many files being in the list, might be easier to visualize if it's a yaml list. <- drive by commentary no need for any action just thinking out loud. But +1 to what the format is.

pipelines/test/ldd-check.yaml
smoser: Yes. It just makes noise. Makes output harder to read

pipelines/test/ldd-check.yaml
smoser: I didn't do that, but I wouldn't be opposed to adding the functionality. 

I agree with the fact that something else other than crappy shell scripts should do this.

smoser: Also, I explicitly did not allow wildcards, but I'm not opposed to that either.

pipelines/test/ldd-check.yaml
smoser: https://github.com/chainguard-dev/melange/issues/1388

smoser: updated doc string.

pipelines/test/ldd-check.yaml
dannf: > Thanks for doing this! Is there a way to use this interface w/o knowing the list of files in advance? I'd coincidentally added the following test to a package in my local tree:
> 
> ```yaml
> test:
>   environment:
>     contents:
>       packages:
>         - apk-tools
>         - posix-libc-utils # for ldd
>     environment:
>       LD_LIBRARY_PATH: /usr/local/cuda-12.5/lib64:/usr/local/cudnn-9/lib64:/opt/tensorrt_llm/libs:/opt/tritonserver/lib
>   pipeline:
>     - name: Scan .so files for undetected runtime dependencies
>       runs: |
>         ret=0
>         apk info -L "${{package.name}}" | grep \.so$ | sed 's,^,/,' | while read obj; do
>           echo -n "Scanning $obj... "
>           if ldd "$obj" | grep "not found"; then
>             echo "failed"
>             ret=1
>             continue
>           fi
>           echo "ok"
>         done
>         exit $ret
> ```
> 
> Here, I don't know the `.so*` files in advance, so I'm just searching the entire package. Ultimately, I think we should teach `melange` to `shlibdeps` these on its own, allowing users to add exceptions.

fyi: https://github.com/chainguard-dev/internal-dev/issues/7609

PR #36864, opened 2024-12-13
revert: icu 76.1 package update as contains ABI breakage, bumps dependants that have already been built with new ABI

rawlingsj:
see https://github.com/wolfi-dev/os/pull/35274

package update check is expected to fail as we are rolling back an update.

### DIFF
--- a/icu.yaml
+++ b/icu.yaml
@@ -1,7 +1,7 @@
 package:
   name: icu
-  version: "76.1"
-  epoch: 0
+  version: "75.1"
+  epoch: 4
   description: "International Components for Unicode library"
   copyright:
     - license: MIT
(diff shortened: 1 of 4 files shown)


xnox (dismissed):

xnox:
Also likely want to withdraw 76.1 APKs

rawlingsj:
> Also likely want to withdraw 76.1 APKs

yeah lemme push another commit to this PR

maxgio92 (dismissed):

sil2100 (commented):
LGTM

xnox (dismissed):

maxgio92:
Tensorflow build was complaining during Bazel build because the Python version in the hermetic environment was defaulted to the latest 3.13 while Tensworflow's Bazel workspace doesn't support a requirements_lock.txt for 3.13.

xnox:
The builds are good; i cannot tell if they are out of order or not; we may need to bump/rebuild a few more things; but I would rather land this sooner than later. As a lot more things are currently broken; or can get more broken as things are landing.

xnox (approved):

PR #36891, opened 2024-12-13
Add python-as-wrapper package.

smoser:
Many programs or build systems really want to use 'python3',
but we really want them to use 'python3.X'.

This wrapper package provides a compromise that gets rid of the
"sticky" results of calling 'python3'.  The big change is that
now a program that uses 'python3' to write its scripts will
write a shbang as '#!/usr/bin/python3.XX'

See python-as-wrapper/README.md for more information.


smoser:
I feel like this can actually replace a lot of the shell that I've written in pipelines/py/, but even without that, it allows more build systems to "just work" by providing a `python3` or `python` that works we want it to.

python-as-wrapper.yaml
```diff
@@ -0,0 +1,176 @@
...
+  version: 0.1.0
+  epoch: 0
+  description: Shell script wrapper to exec python3.XX
+  copyright:
+    - license: Apache-2.0
+
+# In a "normal" installation of python, 'python3' is a symlink
+# to 'pythonX.Y' where X.Y is Major.Minor (3.13).  In that scenario,
+# if caller invokes 'python3', then when python reads its sys.executable,
+# he executable is 'python3'.
```
xnox, line 12: the

smoser: fixed. thank you for reading.

python-as-wrapper.yaml
```diff
@@ -0,0 +1,176 @@
...
+    description: provides /usr/bin/python3 that execs python${{range.key}}
+    dependencies:
+      runtime:
+        - python-${{range.key}}-base
+    pipeline:
+      - runs: |
+          ./write-strict "python${{range.key}}" ${{targets.contextdir}}/usr/bin/python3
+    test:
+      pipeline:
+        - name: validate wrapper 'python3' for ${{range.key}}
```
dannf, line 90: to avoid code duplication, maybe we could put this in a `python-as-wrapper-tests` package and ship it?

python-as-wrapper.yaml
```diff
@@ -0,0 +1,176 @@
...
+      python3: PYTHON3_AS
+
+environment:
+  contents:
+    packages:
+      - busybox
+
+pipeline:
+  - runs: |
+      cat > write-strict <<"EOF"
```
dannf, line 44: Is there a reason to prefer HERE documents vs. adding flat files to the tree?

xnox:
silly question, why shell and not python https://docs.python.org/3/library/os.html#os.execv ?

python-as-wrapper.yaml
```diff
@@ -0,0 +1,176 @@
...
+#
+# That value of sys.executable is "sticky" and finds its way into shbangs that
+# are installed with 'pip install' or other places.
+#
+# In the vast majority of places, we do not want shebang with
+# /usr/bin/python3 because at runtime:
+#  * python3 might be a symlink to another python version
+#  * python3 may not exist at all - with only python3.XX installed.
+#
+# Using the wrapper that this package provides if you do
```
dannf, line 22: I feel this might be a simpler description:

> This wrapper will cause the scripts generated by `python3 -m pip install pkg` to be generated with a shbang of `/usr/bin/python3.Y` instead of `/usr/bin/python3`.

It might also be worth noting (if this is indeed your intent) that:

>  `python3.X -m pip install pkg` should normally be used. Use of these wrappers is intended only as a way to work around packages that still end up generating `/usr/bin/python3?` shbangs.

dannf (dismissed):
The concept makes sense to me, nice! Should we add some bumpers to try and avoid this getting shipped in packages/images? Maybe emit a warning when used?

smoser:
It is probably difficult to tell exactly what  python-as-wrapper does by reading python-as-wrapper.yaml.  I encourange you to pull this branch, `make package/python-as-wrapper` and then `make local-wolfi` and `apk add python-as-3.13` or `python3-as-3.12`.

What you'll see is:
* `python3-as-python3.13` package
  * depends on `python-3.13-base` (for /usr/bin/python3.13)
  * installs `/usr/bin/python3` with:
    ```bash
    #!/bin/sh
    exec "python3.13" "$@"
    ```
* `python-as-python3.13` package
  * depends on `python-3.13-base`
  * installs `/usr/bin/python` with the same 2 line shell script.

The `python3-as-env` and `python-as-env` packages are a bit more complicated.
They install `/usr/bin/python3` and `/usr/bin/python` with the content below.
 * `python-as-env` uses the variable `PYTHON_AS`
 * `python3-as-env` uses the variable `PYTHON3_AS`.
 * this is fully posix shell.
 * all of the code is really just to give good error messages so people will be
   able to diagnose what went wrong.

```bash
#!/bin/sh
fail() { echo "FATAL:" "$0" "$@"; exit 9; }
[ -n "${PYTHON3_AS}" ] || fail "invoked without PYTHON3_AS set"
[ "$PYTHON3_AS" -ef "$0" ] && fail "invoked with PYTHON3_AS = $0"
[ "${PYTHON3_AS}" != "${0##*/}" ] || fail "invoked with PYTHON3_AS = basename($0)"
command -v "$PYTHON3_AS" >/dev/null ||
  fail "invoked with PYTHON3_AS=$PYTHON3_AS, but $PYTHON3_AS not found in PATH"
exec "${PYTHON3_AS}" "$@"
```

The overhead of both the `-env` and `-3.XX` versions is minimal.
The following is collected on a workstation by running `time` like below
with PPATH set to the wrapper /usr/bin/python3 and /usr/bin/python3.13 directly.
```bash
nums=$(seq 1 1000);
PPATH=/usr/bin/python3
time sh -c 'p=$1; shift; for n in "$@"; do $p -c pass; done' PPATH $nums
```


| invocation          | PYTHON3_AS          | real     | user     | sys      |
| -------------       | ------              | ------   | ---      | ---      |
| python3-as-env      | /usr/bin/python3.13 | `14.55s` | `10.14s` | ` 3.53s` |
| python3-as-env      | python3.13          | `14.56s` | `10.13s` | ` 3.55s` |
| python3-as-3.13     | n/a                 | `15.16s` | `10.16s` | ` 3.86s` |
| /usr/bin/python3.13 | n/a                 | `12.92s` | ` 9.69s` | ` 2.79s` |

The numbers aren't scientific.  Prices and participation my vary, but they do
show that the overhead is small.  Maybe < 3 seconds on 1000 invocations.
I don't know why, but in the two times I ran the `python3-as-3.13` was slower
than python3-as-env.  That probably indicates that there is more noise in the
tests on a virtual machine in GCE.

smoser:
> silly question, why shell and not python https://docs.python.org/3/library/os.html#os.execv ?

It would be longer text-wise and slower.  It would drop a shell depends, but ... we're not really going to get rid of /bin/sh in our build environments.

Here is a 'time' run of 1000 invocations of the noops `python3.13 -c pass' and 'sh -c :'.  Basically, it shows the overhead of bringup of those interpreters.

```
$ nums=$(seq 1 1000); time sh -c 'for n in "$@"; do python3.13 -c pass ; don
e' -- $nums
real    0m 13.26s
user    0m 9.80s
sys     0m 2.91s

$ nums=$(seq 1 1000); time sh -c 'for n in "$@"; do sh -c :; done' -- $nums
real    0m 1.84s
user    0m 0.45s
sys     0m 0.90s
```

python-as-wrapper.yaml
smoser: I did consider that. I can do that if you'd like.

python-as-wrapper.yaml
smoser: i dont know.  I'm fine changing that to files in `python-as-wrapper/` if you'd like.  it is quite arguably more readable that way.

xnox:
> > silly question, why shell and not python https://docs.python.org/3/library/os.html#os.execv ?
> 
> It would be longer text-wise and slower. It would drop a shell depends, but ... we're not really going to get rid of /bin/sh in our build environments.
> 
> Here is a 'time' run of 1000 invocations of the noops `python3.13 -c pass' and 'sh -c :'. Basically, it shows the overhead of bringup of those interpreters.
> 
> ```
> $ nums=$(seq 1 1000); time sh -c 'for n in "$@"; do python3.13 -c pass ; don
> e' -- $nums
> real    0m 13.26s
> user    0m 9.80s
> sys     0m 2.91s
> 
> $ nums=$(seq 1 1000); time sh -c 'for n in "$@"; do sh -c :; done' -- $nums
> real    0m 1.84s
> user    0m 0.45s
> sys     0m 0.90s
> ```

ok, rust binary? c binary? go binary? i do understand those are harder to write over shell script.

python-as-wrapper.yaml
smoser: ok. split everything out.

python-as-wrapper.yaml
smoser: :heavy_check_mark:

python-as-wrapper.yaml
smoser: done.

smoser:
@dannf 
> Should we add some bumpers to try and avoid this getting shipped in packages/images? Maybe emit a warning when used?

ideas on how?

Also, thanks for the push to split out the HERE docs, this is more readable now.

smoser:
I added a README.md . Please read it to get a summary of what the packages do.
https://github.com/smoser/wolfi-os/blob/feat/python-as-wrapper/python-as-wrapper/README.md

smoser:
Any reviewers willing to ACK this ?

xnox (commented):
```
$ docker run -it --entrypoint sh cgr.dev/chainguard/python
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: exec: "sh": executable file not found in $PATH: unknown

Run 'docker run --help' for more information
```

Our python runtime images do not have shell today. Thus this will not be possible to use in the python runtime image, is that ok?

Alternatively if we will need this script in our python runtime container image, this script would need to be rewritten in a compiled language or in python itself.

smoser:
> ```
> $ docker run -it --entrypoint sh cgr.dev/chainguard/python
> docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: exec: "sh": executable file not found in $PATH: unknown
> 
> Run 'docker run --help' for more information
> ```
> 
> Our python runtime images do not have shell today. Thus this will not be possible to use in the python runtime image, is that ok?

The goal is for this to only ever be used in build systems. So at this point, that is unrecognized feature not a bug.  :partying_face: 

> 
> Alternatively if we will need this script in our python runtime container image, this script would need to be rewritten in a compiled language or in python itself.

Yeah.  I think if that is needed, we can absolutely write it in C or rust.

xnox (approved):

PR #36898, opened 2024-12-13
Bump/confluent docker utils 0.0.129

smoser:
This is a different take on https://github.com/wolfi-dev/os/pull/36827

Python transition is painful for us.
We're 80% through the default 'py3-<package>' installing the py3.13-<package> version.
Much of that pain really is just that using un-versioned package names doesn't work well.

The changes here:
1. use 'py-version' variable, to make a change easier
2. use a new `python3-as-3.12` package that I had just put up at https://github.com/wolfi-dev/os/pull/36891

The `python3-wrapper` package allows us to get the behavior we want from invokers of 'python3' without the pain of changing them.

### DIFF
--- a/confluent-docker-utils.yaml
+++ b/confluent-docker-utils.yaml
@@ -1,7 +1,7 @@
 #nolint:git-checkout-must-use-github-updates
 package:
   name: confluent-docker-utils
-  version: 0.0.127
+  version: 0.0.129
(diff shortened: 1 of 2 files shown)


smoser:
I'm going to close this, please put any discussion at https://github.com/wolfi-dev/os/pull/36827

Issue #36914, opened 2024-12-13
icu 76 transition


xnox:
rough list of things to rebuild

```
$ apkrane ls --json --latest https://packages.wolfi.dev/os/x86_64/APKINDEX.tar.gz | jq -r 'select(.Dependencies) | select(.Dependencies | index("so:libicuuc.so.75")) | .Origin' | sort -u
apache-arrow
couchdb
couchdb-3.3
falco
freerdp-3
gitlab-cng-17.1
gitlab-cng-17.2
gitlab-cng-17.3
gitlab-cng-17.4
gitlab-cng-17.5
gitlab-cng-17.6
grpc-1.66
grpc-1.67
grpc-1.68
harfbuzz
libical
mozjs91
neon
nodejs-16
nodejs-18
nodejs-20
nodejs-21
nodejs-22
nodejs-23
percona-server-8.4
percona-server-9.0
percona-xtrabackup-8.4
php-8.1
php-8.1-pecl-http
php-8.2
php-8.2-pecl-http
php-8.3
php-8.3-pecl-http
php-8.4
php-8.4-pecl-http
postfix
postfix-3.9
postgresql-16
postgresql-17
qt5-qtbase
R
re2
ruby3.2-charlock_holmes
ruby3.3-charlock_holmes
tesseract
tileserver-gl
```

hbh7:
I can work on the epoch bumps for these today if we want to proceed.

PR #36915, opened 2024-12-13
withdraw icu 76.1 packages (again)

rawlingsj:
`manual:true` will create an issue to crack the update manually and prevent automated PRs

package update check failure is expected as this is rolling back a version

The withdraw.txt doesn't need to be updated as they are already listed in there, but we will need to run the withdraw action again once this merges.

### DIFF
--- a/icu.yaml
+++ b/icu.yaml
@@ -1,7 +1,7 @@
 package:
   name: icu
-  version: "76.1"
-  epoch: 0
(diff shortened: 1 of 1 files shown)


maxgio92:
Related: https://github.com/wolfi-dev/os/issues/36914

maxgio92 (approved):

PR #36930, opened 2024-12-13
Drop pkg-config tests obsoleted by pkgconf tests

murraybd:
The pkg-config and pkgconf tests are the same so let's drop the pkg-config ones in favor of the pkgconf ones which use a pipeline.


dannf (approved):
thanks for cleaning this up @murraybd !

PR #36966, opened 2024-12-13
Apache Pulsar

MickeyPvX:
Related: https://github.com/chainguard-dev/image-requests/issues/1654

#### For new package PRs only


apache-pulsar.yaml
```diff
@@ -0,0 +1,123 @@
...
+  - name: ${{package.name}}-compat
+    pipeline:
+      - runs: ln -s /usr/share/java/pulsar ${{targets.contextdir}}/pulsar
+
+update:
+  enabled: true
+  github:
+    identifier: apache/pulsar
+    use-tag: true
+    strip-prefix: v
```
jamonation, line 50: There'll need to be a tag filter here to ignore things like candidate versions, e.g. https://github.com/apache/pulsar/releases/tag/v4.0.1-candidate-2

apache-pulsar.yaml
```diff
@@ -0,0 +1,123 @@
...
+
+test:
+  environment:
+    contents:
+      packages:
+        - ${{package.name}}-compat
+        - bash
+        - libstdc++
```
EyeCantCU, line 58: If these are needed by the package, let's add them as runtime deps

apache-pulsar.yaml
```diff
@@ -0,0 +1,123 @@
...
+        - openjdk-17-default-jvm
+        - openjdk-17-jre
```
EyeCantCU, line 60: Fetched by the package

apache-pulsar.yaml
```diff
@@ -0,0 +1,123 @@
...
+  name: apache-pulsar
+  version: 4.0.1
+  epoch: 0
+  description: Pulsar is a distributed pub-sub messaging platform with a very flexible messaging model and an intuitive client API.
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - openjdk-17-default-jvm
+      - openjdk-17-jre
```
EyeCantCU, line 11: `-default-jvm` will pull in `-jre`, we should probably use `-jre` and install `-default-jvm` at the image level

apache-pulsar.yaml
MickeyPvX: I figured out what I was doing wrong (or maybe not-quite-right?) - I'd been testing the package build with `make dev-container` but I think that has some extra packages on board that would make me miss some of the runtime dependencies.

Re-read the Building Packages docs, switched to testing with `make local-wolfi` and that actually replicates all the runtime failures I was seeing when doing `make test/apache-pulsar`

apache-pulsar.yaml
MickeyPvX: So I took out the `-default-jvm` leaving just the `-jre` and I get the common `Error: JAVA_HOME not set, and no java executable found in /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin` - I assume that's expected now until we install the `-default-jvm` in the image or specify it as part of what we need for testing?

jamonation (dismissed):

apache-pulsar.yaml
EyeCantCU: Yep, you'd either export `JAVA_HOME` or install `-default-jvm`

EyeCantCU (approved):

jamonation (approved):

PR #36981, opened 2024-12-13
meson: python multiversioning

dannf:

I also tweaked the demo test a bit to use a temporary directory instead of hardcoded paths in /tmp, and to truncate files instead of appending to them in HERE documents, which I imagine would break tests if ran multiple times.

### DIFF
--- a/meson.yaml
+++ b/meson.yaml
@@ -1,13 +1,48 @@
 package:
   name: meson
   version: 1.6.0
-  epoch: 1
+  epoch: 2
   description: Fast and user friendly build system
   copyright:
     - license: Apache-2.0
   dependencies:
-    runtime:
-      - samurai
+    provider-priority: 0
+
+vars:
+  pypi-package: meson
+  import: mesonbuild
+  test-version-help: |
+    meson --help
+  test-demo: |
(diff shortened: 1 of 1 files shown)


meson.yaml
```diff
@@ -26,9 +59,81 @@ pipeline:
...
+    description: meta package providing ${{vars.pypi-package}} for supported python versions.
+    dependencies:
+      runtime:
+        - py3.10-${{vars.pypi-package}}
+        - py3.11-${{vars.pypi-package}}
+        - py3.12-${{vars.pypi-package}}
+        - py3.13-${{vars.pypi-package}}
+    test:
+      pipeline:
+        - uses: python/import
```
smoser, line 120: I'd be fine with the pipeline here supporting "pythons" and doing the iterating in it. 

I'm not sure if you can reference "py-versions.keys()" or something to that effect.

smoser (approved):
One thought for brevity.  But this is fine as it is. 

Thank you

meson.yaml
dannf: I agree, but I couldn't find a way to do that. It appears that `range` is currently only permitted as a child of `subpackage`.

PR #37113, opened 2024-12-16
airflow: pin to python3.12, due to FTBFS with python3.13

xnox:
Currently airflow is failing to build from source in post-submit, due to python3 changing from 3.12 to 3.13.

This change happened after successful pre-submit and prior to successful build in post-submit. This race is possible as we do not require "up to date" PR prior to merging, which would be impractical to enforce.

In the future, prior to flipping default, we should detect FTBFS and pin packages to 3.12 prior to upgrading the default.

### DIFF
--- a/airflow.yaml
+++ b/airflow.yaml
@@ -1,7 +1,7 @@
 package:
   name: airflow
   version: 2.10.3
-  epoch: 1
+  epoch: 2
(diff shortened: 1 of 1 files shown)


kranurag7 (approved):
thank you!! looks good, Isolated package we're not adding airflow as runtime dependencies to any existing other python package.

PR #37132, opened 2024-12-16
fix(aws-cli-2): Provide aws-cli

EyeCantCU:
aws-cli should resolve to the latest version of the CLI


kranurag7 (approved):
thanks RJ, you're amazing!!

PR #37195, opened 2024-12-16
Re-name Jenkins package to match version stream

mamccorm:
Jenkins is referred to as 'jenkins-2' in version stream data. Re-naming the package to that affect, and adding a 'provides', to maintain backwards compatibility.

This aligns the package with others that have different version streams, in terms of the naming in wolfi.


jenkins-2.yaml
```diff
@@ -14,6 +16,8 @@ package:
       - glibc-locale-en
       # The entrypoint script is pulled from a separate repo, which we'e packaged.
       - jenkins-entrypoint
+      - openjdk-17
```
mamccorm, line 19: This is intentional. Today we define these as runtime deps at the image level - we talked about moving these to package level definitions recently.

EyeCantCU (approved):

PR #37222, opened 2024-12-17
cython-0 - fix ftbfs after python3.13 transition

smoser:
cython-0 would not build with python3 as python3.13. Pin it to python-3.12.


dannf (approved):
+1. I wonder if we are able to get rid of this version stream yet. It was created for `lxml`, which no longer needs it. `grpc-1.66` and `tensorflow-core` still use it, but perhaps that is just cargoculted. :shrug:

PR #37223, opened 2024-12-17
nodejs - be explicit about which python is used in build environment.

smoser:
When 'python3' and 'py3-jinja2' are installed in the build environment, there is not currently a way to guarantee that the 'python3' (a virtual package provided by python3.12, python3.13...) will be the same python version that fulfills the virtual package 'py3-jinja2'.

The easiest way to guarantee this is to pin the python version as done here.

python-3.12 (eol 2028) will outlive any of the currently available nodejs versions. So it would be sufficient for the 'py-version' value in all of these changes.  Moving the last node lts (22) to 3.13 will give us a neweer python moving forward.

### DIFF
--- a/nodejs-16.yaml
+++ b/nodejs-16.yaml
@@ -1,7 +1,7 @@
 package:
   name: nodejs-16
   version: 16.20.2
-  epoch: 7
+  epoch: 8
   description: "JavaScript runtime built on V8 engine"
   dependencies:
     provides:
@@ -12,6 +12,9 @@ package:
   resources:
     cpu: 33
     memory: 64Gi
+
+vars:
+  py-version: 3.12
 
 environment:
   contents:
@@ -27,9 +30,9 @@ environment:
       - nghttp2-dev
       - openssf-compiler-options
       - openssl-dev
-      - py3-jinja2
-      - py3-setuptools
-      - python3
+      - py${{vars.py-version}}-jinja2
+      - py${{vars.py-version}}-setuptools
+      - python-${{vars.py-version}}-base
       - samurai
       - wolfi-base
       - zlib-dev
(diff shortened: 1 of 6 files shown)


smoser:
nodejs-19 was failing with:

```
../../deps/v8/src/heap/cppgc/stats-collector.h:394:48: error: cannot convert '
al::StatsCollector::AllocationObserver*>::iterator' to 'const char*'          
  394 |         std::remove(allocation_observers_.begin(), allocation_observer
                                                                              
      |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~                     
      |                                                |                      
      |                                                std::vector<cppgc::inte
ocationObserver*>::iterator                                                   
In file included from /usr/include/c++/14/cstdio:42,                          
                 from /usr/include/c++/14/ext/string_conversions.h:45,        
                 from /usr/include/c++/14/bits/basic_string.h:4154,           
                 from /usr/include/c++/14/string:54,                          
                 from /usr/include/c++/14/bits/locale_classes.h:40,           
                 from /usr/include/c++/14/bits/ios_base.h:41,                 
                 from /usr/include/c++/14/ios:44,                             
                 from /usr/include/c++/14/istream:40,                         
                 from /usr/include/c++/14/sstream:40,                         
                 from ../../deps/v8/src/base/logging.h:10,                    
                 from ../../deps/v8/src/base/macros.h:12,                     
                 from ../../deps/v8/src/heap/cppgc/sweeper.h:10:   
...

  158 | extern int remove (const char *__filename) __THROW;                   
      |                    ~~~~~~~~~~~~^~~~~~~~~~                             
In file included from ../../deps/v8/src/heap/cppgc/process-heap-statistics.h:9
                 from ../../deps/v8/src/heap/cppgc/process-heap-statistics.cc:
../../deps/v8/src/heap/cppgc/stats-collector.h: In member function 
     'void cppgcr::ForAllAllocationObservers(Callback)':                                      
../../deps/v8/src/heap/cppgc/stats-collector.h:394:48: error: 
    cannot convert 'al::StatsCollector::AllocationObserver*>::iterator' to 'const char*'          
  394 |         std::remove(allocation_observers_.begin(), allocation_observer
      |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~                     
      |                                                |                      
      |                                                std::vector<cppgc::inte
rnal::StatsCollector::AllocationObserver*>::iterator          
```

justinvreeland (approved):

PR #37244, opened 2024-12-17
add py3-nbclassic package

kranurag7:
```bash
$ uv pip show nbclassic
Name: nbclassic
Version: 1.1.0
Location: /home/user/tmp/hacky_KRbiX7/.venv/lib/python3.12/site-packages
Requires: ipykernel, ipython-genutils, nest-asyncio, notebook-shim
Required-by:
```

### DIFF
--- a/py3-nbclassic/shlex.patch
+++ b/py3-nbclassic/shlex.patch
@@ -1,0 +1,31 @@
+diff --git a/setupbase.py b/setupbase.py
+index 80f28b6e..4e7a4fee 100644
+--- a/setupbase.py
++++ b/setupbase.py
+@@ -12,7 +12,7 @@ This includes:
+ 
+ import os
+ import re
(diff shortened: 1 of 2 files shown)


py3-nbclassic/shlex.patch
```diff
@@ -0,0 +1,31 @@
+diff --git a/setupbase.py b/setupbase.py
```
EyeCantCU, line 1: Could we add a comment above the diff explaining why we vendor this patch?

kranurag7: Thanks RJ, addressed it via https://github.com/wolfi-dev/os/pull/37244/commits/f3f570617a5c9ff4916a34aac7a1e619c703eb4d

kranurag7:
Closing given this is no longer required.

PR #37276, opened 2024-12-17
py3-pgcli, py3-pgspecial, py3-psycopg: Add python multiversion support.

sil2100:
Add multiversion support to `py3-pgcli` and all its dependencies that missed the same (`py3-pgspecial, py3-psycopg`)

#### For new version streams


py3-pgcli.yaml
```diff
@@ -36,13 +37,85 @@ pipeline:
...
+        - py${{range.key}}-sqlparse
+        - py${{range.key}}-cli-helpers
+        - py${{range.key}}-prompt-toolkit
+        - py${{range.key}}-pygments
+        - py${{range.key}}-pgspecial
+        - py${{range.key}}-pendulum
+        - py${{range.key}}-psycopg
+        - py${{range.key}}-setproctitle
+    pipeline:
+      - name: Python Build
```
smoser, line 59: i think `uses: py/pip-build-install` will work here. then its just one section rather than 2 and more things use it than the other.

py3-pgcli.yaml
```diff
@@ -36,13 +37,85 @@ pipeline:
...
+        - py3-${{vars.pypi-package}}-bin
+        - py3-${{vars.pypi-package}}
+        - ${{package.name}}
+      provider-priority: ${{range.value}}
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/usr/
+          mv ./cleanup/${{range.key}}/bin ${{targets.contextdir}}/usr/
+      # Change the interpreter to the appropriate python version
+      - runs: |
```
smoser, line 85: you should not have to do this... almost certain you will not have to if you use the py/pip-build-install pipeline.

smoser (commented):
this looks right, but try py/pip-build-install please.

py3-pgcli.yaml
```diff
@@ -36,13 +38,81 @@ pipeline:
...
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}-bin
+    description: Executable binaries for ${{vars.pypi-package}} installed for python${{range.key}}
+    dependencies:
+      runtime:
+        - py${{range.key}}-${{vars.pypi-package}}
+      provides:
+        - py3-${{vars.pypi-package}}-bin
+        - py3-${{vars.pypi-package}}
+        - ${{package.name}}
```
smoser, line 78: this is redundant i think.  package name is 'py3-pgcli' and pypi-package is 'pgcli', so they're the same, right?

smoser (changes requested):
one more nit.

smoser (dismissed):
I'm  happy with this. thank you.

just an fyi, we do have the `py3-supported-build-base` and `py3-supported-build-base-dev`. I'm sure you understand the tradeoffs of using the metapackage versus not, but I wanted to make sure you were aware of it


Thanks!

smoser (approved):

PR #37277, opened 2024-12-17
yamllint - specify python version

smoser:
When building a python apckage, it will install into a specific usr/lib/python3.XX/site-packages thus rendering the dependency on build-time of that specific python interpreter.

The runtime dependencies are resolved at install time, and at that point they may have moved.

This dependency is thus correct and required to keep the package running.


dannf (approved):
thx

PR #37299, opened 2024-12-17
py3-jupyter-console: add new package

maxgio92:
#### For new package PRs only

Notes:
BSD 3-Clause License
Notes:
Last release 6.6.3 of Mar 6, 2023.

### DIFF
--- a/py3-jupyter-console.yaml
+++ b/py3-jupyter-console.yaml
@@ -1,7 +1,7 @@
 package:
   name: py3-jupyter-console
   version: 6.6.3
-  epoch: 1
+  epoch: 0
   description: A terminal-based console frontend for Jupyter kernels.
   copyright:
     - license: BSD-3-Clause
--- a/py3-jupyter-console.yaml
+++ b/py3-jupyter-console.yaml
@@ -1,0 +1,120 @@
+package:
+  name: py3-jupyter-console
+  version: 6.6.3
+  epoch: 1
+  description: A terminal-based console frontend for Jupyter kernels.
+  copyright:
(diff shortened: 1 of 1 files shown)


py3-jupyter-console.yaml
```diff
@@ -0,0 +1,120 @@
+package:
+  name: py3-jupyter-console
+  version: 6.6.3
+  epoch: 1
```
jamonation, line 4:
```suggestion
  epoch: 0
```

This is a new package right?

py3-jupyter-console.yaml
```diff
@@ -0,0 +1,120 @@
...
+  description: A terminal-based console frontend for Jupyter kernels.
+  copyright:
+    - license: BSD-3-Clause
+  dependencies:
+    provider-priority: 0
```
jamonation, line 9: What does this setting do in this package?

EyeCantCU: Other packages that provide `py3-blah-blah-blah` will take priority based on the version of Python (latest will always be resolved to). This package is empty so we set a priority of 0 so that it never gets installed

py3-jupyter-console.yaml
maxgio92: Oh yes, good catch. Thanks

maxgio92: Oh yes, good catch. Thanks

py3-jupyter-console.yaml
jamonation: ahhh TIL!

jamonation (approved):

PR #37343, opened 2024-12-17
Fix Druid pombump configuration

dakaneye:
After analyzing the druid pom structure, and the stacktrace linked in the stale druid
issue, realized that we were getting a class compatibility issue with netty-common,
specifically the Timer class. This is because every other netty dependency was a
different version that netty-common. By setting the netty4.version property, we guarantee
that all netty libs are on the same version and are therefore compatible. And the reason
we added the pombump dependency in the first place (probably to solve a CVE) is preserved.

The class compatiblity issue was causing the imagetest for druid to fail, because services could not correctly come up with that issue, which is part of the Druid startup process.

Fixes https://github.com/chainguard-dev/image-release-stats/issues/3075

### DIFF
--- a/druid.yaml
+++ b/druid.yaml
@@ -1,7 +1,7 @@
 package:
   name: druid
   version: 31.0.0
-  epoch: 2
+  epoch: 3
   description: Apache Druid is a high performance real-time analytics database.
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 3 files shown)


pnasrat (approved):
LGTM

Issue #37197, opened 2024-12-16
python 3.13 transition 


smoser:
Last week we bumped the 'provider priority' value for the python 3.13 subpackage of packages named 'py3-<pkg>'.
That is to say that before last week, if you did `apk add py3-lxml` you'd end up with the python-3.12 version, a package named `py3.12-lxml`.  This week, you'd end up with `py3.13-lxml` and its dependencies installed.

There are several packages that have an un-versioned python package listed as a build-depends, runtime-dependency or test dependency.   Many of these packages will now be broken either in runtime or in the next build for reasons such as:
1. they installed python modules (.py files) into `/usr/lib/python3.12/site-packages` or a virtual env's `lib/python3.12/site-packages` using 'python' or 'python3' to do the install.
2. their build uses a `python3` or `python3-dev` package , but does not support python-3.13

I searched through existing melange files for use of un-versioned python 3 packages.
There are very few cases where an un-versioned python dependency (`python3-dev` rather than `python-3.XX-dev`) should be used.  Those are mostly limited to:

1. build systems that use only `python3` and no other python packages.
2. runtime dependencies that use _only_ the python executable and standard library. byobu is one such example.

To update packages, follow these bullet points:

 * Remove `python-3` and `python-3-dev` references. Either replace them with `python3` or `python3-dev`. In almost all cases, `python3` should not be listed in the runtime or test dependencies.
 * Packages that simply use `python` or `python3` in their build environment but do not produce packages with python files can continue to use `python3`.
 * If `python3` is listed only in the build environment (not elsewhere), you may well be able to just delete it.  If another package in the build environment needs python, it should be pulled in.
 * If a package has more than one python package (py3-<pkg>) in its build environment, it should pick a specific version (3.13 if supported) and list the dependencies with `py${{vars.py-version}}-pkg`.  This will guarantee that the dependencies are installed at a consistent version.
 * If there are more than one python build package (setuptools, wheel...) listed in the build environment, then replace those with `py3.XX-build-base`, then use the build-base package.
 * If a package has anything other than `python3` listed in its runtime depends, _or_ it uses a virtual env, then it needs to pick a specific python version.

 * If the build uses `python3` or `python` by name, the easiest thing to do is
   to use a `python-as-wrapper` package to provide that executable.
   https://github.com/wolfi-dev/os/pull/36891  . It ensures that shbang will get
   the correct executable value, and melange will then generate the correct
   dependendency.

[...]

PR #37370, opened 2024-12-17
Add python3 as a dep so g-ir-scanner can run

murraybd:
g-ir-scanner would previously fail to run because python3 was not available.

### DIFF
--- a/gobject-introspection.yaml
+++ b/gobject-introspection.yaml
@@ -1,7 +1,7 @@
 package:
   name: gobject-introspection
   version: 1.82.0
-  epoch: 1
+  epoch: 2
   description: Introspection system for GObject-based libraries
   copyright:
     - license: LGPL-2.0-or-later AND GPL-2.0-or-later AND MIT
@@ -9,6 +9,7 @@ package:
   # build it. This is a drop in replacement to make things work.
   dependencies:
     runtime:
+      - python3
       - py3-setuptools
 
 # creates a new var that contains only the major and minor version to be used in the fetch URL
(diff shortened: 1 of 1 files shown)


smoser (changes requested):
This isn't right. We need to fix the shbang so it uses the Python that it was installed for. 

A python3 drop will Get you a /usr/bbin/python3, but not necessarily the right python3. 

See my issue for unversioned Python and 313. Also related here is the python-as wrapper, it works likely fix this.

dannf:
For reference, here's the issue @smoser is referring to - https://github.com/wolfi-dev/os/issues/37197
According to that, I think we want to try to use the suggestion in the "pick a specific version" example, and then fallback to the `python-as-wrapper` thing if it fails to inherit a python3.13 shbang.

smoser:
> For reference, here's the issue @smoser is referring to - https://github.com/wolfi-dev/os/issues/37197
> 
> According to that, I think we want to try to use the suggestion in the "pick a specific version" example, and then fallback to the `python-as-wrapper` thing if it fails to inherit a python3.13 shbang.

If we land Python-as-wrapper (#36891) we can use that. I do think it's useful to generally "do the right thing".

smoser:
> For reference, here's the issue @smoser is referring to - https://github.com/wolfi-dev/os/issues/37197
> 
> According to that, I think we want to try to use the suggestion in the "pick a specific version" example, and then fallback to the `python-as-wrapper` thing if it fails to inherit a python3.13 shbang.

If we land Python-as-wrapper (#36891) we can use that. I do think it's useful to generally "do the right thing".

murraybd:
I'm going to close this in favor of sil2100's PR at https://github.com/wolfi-dev/os/pull/37393.

PR #37439, opened 2024-12-18
New package dotnet-8.0.404

xnox:
Packaging of optional SDK version of .net 8 tagged 8.0.404. Note this
is sdk only, for package build-time only for now. Runtime versions
unaffected.

### DIFF
--- a/dotnet-8.0.404.yaml
+++ b/dotnet-8.0.404.yaml
@@ -1,0 +1,99 @@
+package:
+  name: dotnet-8.0.404
+  version: 8.0.404
+  epoch: 1
+  description: ".NET SDK"
+  copyright:
+    - license: MIT
+  resources:
+    cpu: 2
+    memory: 32Gi
+  dependencies:
+    runtime:
+      - dotnet-8-sdk-default
+      - icu
+
+environment:
+  environment:
+    DOTNET_CLI_TELEMETRY_OPTOUT: 1
+    DISABLE_CROSSGEN: true
+    DisableArcade: 1
+  contents:
+    packages:
+      - bash-binsh
+      - brotli-dev
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - clang-15
+      - clang-15-default
+      - cmake
+      - curl
+      - dotnet-bootstrap-8
+      - glibc-locale-en
+      - icu-dev
+      - krb5-dev
+      - libunwind
+      - libunwind-dev
+      - llvm15
+      - llvm15-cmake-default
+      - llvm15-dev
+      - llvm15-tools
+      - lttng-ust-dev
+      - ncurses-dev
+      - nodejs-22
+      - openssl-dev
+      - posix-libc-utils
+      - python3
+      - rapidjson-dev
+      - samurai
+      - wolfi-base
+      - zlib-dev
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/dotnet/sdk
+      tag: v${{package.version}}
+      expected-commit: 5dd80ae4b107027f86bda68ce91542cad72b2868
+      destination: /home/build/src
+
+  - working-directory: /home/build/src
+    pipeline:
+      - runs: |
+          ln -sf /usr/share/dotnet-bootstrap-8/dotnet .dotnet
+          mkdir -p prereqs/packages/archive/
+          ln -sf /usr/share/dotnet-bootstrap-8/Private.SourceBuilt.Artifacts.Bootstrap.tar.gz prereqs/packages/archive/Private.SourceBuilt.Artifacts.Bootstrap.tar.gz
+          mkdir -p prereqs/packages/previously-source-built
+          tar -xzvf prereqs/packages/archive/Private.SourceBuilt.Artifacts.Bootstrap.tar.gz -C prereqs/packages/previously-source-built/
+      - runs: |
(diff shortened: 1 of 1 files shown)


dotnet-8.0.404.yaml
```diff
@@ -0,0 +1,99 @@
+package:
+  name: dotnet-8.0.404
```
imjasonh, line 2: This version stream form seems weird to me. Should this be -8.0, or -8? When 8.0.405 comes out, should we delete this?

pnasrat: Yeah its semver https://learn.microsoft.com/en-us/dotnet/core/versions/

> - MINOR is incremented when:
>    -  Public API surface area is added.
>    -  A new behavior is added.
>     - A newer MINOR version of an existing dependency is adopted.
>     - A new dependency is introduced.

So 8 or 8.00

dotnet-8.0.404.yaml
```diff
@@ -0,0 +1,99 @@
...
+  version: 8.0.404
+  epoch: 1
+  description: ".NET SDK"
+  copyright:
+    - license: MIT
+  resources:
+    cpu: 2
```
dakaneye, line 9: do we know that dotnet needs these resources to work at all?

dotnet-8.0.404.yaml
```diff
@@ -0,0 +1,99 @@
...
+    packages:
+      - bash-binsh
+      - brotli-dev
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - clang-15
+      - clang-15-default
+      - cmake
+      - curl
```
dakaneye, line 31: does dotnet need curl to work?

dotnet-8.0.404.yaml
```diff
@@ -0,0 +1,99 @@
...
+      - llvm15-tools
+      - lttng-ust-dev
+      - ncurses-dev
+      - nodejs-22
+      - openssl-dev
+      - posix-libc-utils
+      - python3
+      - rapidjson-dev
+      - samurai
+      - wolfi-base
```
dakaneye, line 50: this is just for me, but under what circumstances do we need build-base and wolfi-base?

dotnet-8.0.404.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+  - uses: strip
+
+update:
+  enabled: true
+  github:
+    identifier: dotnet/sdk
+    strip-prefix: v
+    use-tag: true
+    tag-filter: "v8.0.404"
```
dakaneye, line 92: similar to @imjasonh 's comment, if we match an exact tag then update is effectively disabled yes?

dotnet-8.0.404.yaml
xnox: these are build-dependencies; potentially we can trim them; it is likely that nuget uses curl.

dotnet-8.0.404.yaml
xnox: this is copy & paste from our existing dotnet-8.yaml and dotnet-9.yaml packaging. It is likely that many of these are redundant and we can trim them down.

dotnet-8.0.404.yaml
xnox: this is additional SDK that is needed to build curtain projects, specifically powershell 7.4.6.

dotnet-8 is the one that gets upgraded.

dotnet-8.0.404.yaml
```diff
@@ -0,0 +1,99 @@
...
+  epoch: 1
+  description: ".NET SDK"
+  copyright:
+    - license: MIT
+  resources:
+    cpu: 2
+    memory: 32Gi
+  dependencies:
+    runtime:
+      - dotnet-8-sdk-default
```
xnox, line 13: @pnasrat @imjasonh @dakaneye note this SDK depends on the default (longterm) SDK and the default runtime. This is not standalone version stream, but a supplement / build-time-only-additional-SDK.

dotnet-8.0.404.yaml
xnox: it downloads lots of stuff; and links lots of stuff; these resources were set by infra people before based on past builds.

xnox:
@pnasrat @dakaneye @imjasonh how this package will be used is shown here:
- https://github.com/wolfi-dev/os/pull/37445
in the powershell ugprade pr, which for 7.4.6 no longer can build with default dotnet-8 sdk; and it needs this extra version of SDK to build.

dotnet-8.0.404.yaml
pnasrat: @xnox given the confusion here can you add more explicit comments in the file itself for the usage and why it is numbered the way it is. So it is apparent to those reading the melange file

dotnet-8.0.404.yaml
pnasrat: As above document that in the melange file at the updatre block

sil2100 (commented):
Packaging looks fine. Only concerns are the small nitpick inline + as per other reviewers, a request to add some comments regarding the reasoning for the package's existence. Maybe even using the `description` field for that?

I'm also possibly missing some context here. I assume that we want to introduce the 8.0.404 SDK via a separate package instead of updating dotnet-8 because we don't want to break anyone in case this version bump introduces regressions? So basically at one moment we'll have a newer dotnet-8 version and this package here will be dropped, right?

xnox:
This got merged already; but need to address this review feedback and still get it merged.

Todo for 2025.

PR #37440, opened 2024-12-18
openssf-compiler-options: remove explicit dependency

justinvreeland:


xnox:
I don't know if we want to pointlessly rebuild these; or simply force-merge without any epoch bumps.

cc @imjasonh @ajayk

justinvreeland:
> I don't know if we want to pointlessly rebuild these; or simply force-merge without any epoch bumps.

I did batch the anticipating that someone might suggest the epoch bump. But if we're fine without the epoch bump I'm happy to submit a larger change.

justinvreeland:
@imjasonh @ajayk @xnox  I'm assuming there's no objections to pushing this without the EPOCH bump in which case I'll open a new PR removing them from all packages in wolfi today.

xnox (approved):

ajayk (approved):

PR #37500, opened 2024-12-19
Update glib and gobject-introspection for specific python version 3.13

smoser:
* gobject-introspection - use specific python version, test g-ir-scanner
   
   gobject-introspection build writes programs in /usr/bin/
   g-ir-scanner and g-ir-annotation-tool that used '#!/usr/bin/env python3'
   
   It installs python files and modules into /usr/lib/gobject-introspection
   and successfully finds that at runtime.  However, it built a c extension
   _giscanner.cpython-312-x86_64-linux-gnu.so which will only load with
   a specific python (the one that it was built against).
   
   So, update the shbang on g-ir-scanner and g-ir-annotation-tool to
   reference the specific python it was built against.

 * glib - use a specific python in build, drop runtime dep on python3
   
   the build system installs /usr/bin/gtester-report with
   /usr/bin/python3.XX shbang (based on the build version). This is good,
   but in this case it does appear not strictly necessary.
   
   The package installs a pure python, standard-lib-only module into
   usr/share/glib-2.0 and finds it correctly.  It _could_ work with "any
   old python".

### DIFF
--- a/glib.yaml
+++ b/glib.yaml
@@ -1,10 +1,15 @@
 package:
   name: glib
   version: 2.83.2
-  epoch: 0
+  epoch: 1
   description: Common C routines used by Gtk+ and other libs
   copyright:
     - license: LGPL-2.1-or-later
+
+vars:
+  # build uses rst2man from py3-docutils and also invokes 'python3'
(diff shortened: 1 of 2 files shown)


jonjohnsonjr (approved):

dannf (approved):
thanks!

PR #37597, opened 2024-12-19
tensorflow-cpu: add new package

maxgio92:
#### For new package PRs only

### DIFF
--- a/tensorflow-cpu.yaml
+++ b/tensorflow-cpu.yaml
@@ -1,7 +1,7 @@
 #nolint:valid-pipeline-git-checkout-tag
 package:
   name: tensorflow-cpu
-  version: 0_git20250107
+  version: 0_git20250108
   epoch: 0
   description: tensorflow-cpu
   copyright:
--- a/tensorflow-cpu.yaml
+++ b/tensorflow-cpu.yaml
@@ -1,7 +1,7 @@
 #nolint:valid-pipeline-git-checkout-tag
 package:
   name: tensorflow-cpu
-  version: 0_git20250106
+  version: 0_git20250107
   epoch: 0
   description: tensorflow-cpu
   copyright:
@@ -19,7 +19,7 @@ package:
       - unzip
 
 vars:
-  pypi-package: tensorflow-cpu
+  pypi-package: tensorflow
   prefix: /usr
 
 environment:
@@ -36,7 +36,7 @@ environment:
 pipeline:
   - uses: git-checkout
     with:
-      expected-commit: d311fe017c036143186f9185f491fa6d6461a84e
+      expected-commit: 35a7dbc754193523e5582ce5b5cbbb4c07cff7b2
       repository: https://github.com/tensorflow/build
       branch: master
 
--- a/tensorflow-cpu.yaml
+++ b/tensorflow-cpu.yaml
@@ -14,12 +14,13 @@ package:
       - build-base
       - curl
       - pkgconf
-      - py3.11-pip
-      - python-3.11
+      - py${{vars.py-version}}-pip
+      - python-${{vars.py-version}}-dev
       - unzip
 
 vars:
-  pypi-package: tensorflow
+  py-version: 3.11
+  pypi-package: tensorflow-cpu
   prefix: /usr
 
 environment:
@@ -27,8 +28,8 @@ environment:
     packages:
       - build-base
       - busybox
-      - py3.11-pip
-      - python-3.11
+      - py${{vars.py-version}}-pip
+      - python-${{vars.py-version}}
   environment:
     # This is needed to work around the error "ValueError: ZIP does not support timestamps before 1980"
     SOURCE_DATE_EPOCH: 315532800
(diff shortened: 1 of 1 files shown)


tensorflow-cpu.yaml
```diff
@@ -0,0 +1,121 @@
+package:
+  name: tensorflow-cpu
+  version: 0_git20241219
+  epoch: 1
```
developer-guy, line 4: epoch should be zero here 🫠

tensorflow-cpu.yaml
```diff
@@ -0,0 +1,118 @@
...
+          cp jupyter.readme.md ${{targets.contextdir}}/tf/tensorflow-tutorials/README.md
+      - name: Cleanup pycache
+        runs: |
+          find . -name '__pycache__' -exec rm -rf {} +
+
+test:
+  environment:
+    contents:
+      packages:
+        - curl
```
anushkamittal20, line 79: shouldnt we get curl in runtime dep : https://github.com/tensorflow/build/blob/master/tensorflow_runtime_dockerfiles/cpu.packages.txt

maxgio92: Right, good catch

tensorflow-cpu.yaml
```diff
@@ -0,0 +1,122 @@
...
```
anushkamittal20, line 83: Do we need this here if we have it as a runtime dep 🤔 

P.S. Please correct me if I got that wrong!

maxgio92: It's needed for the tests for Jupyter

anushkamittal20:
This PR looks good to me! Probably just needs a version bump.

tensorflow-cpu.yaml
```diff
@@ -0,0 +1,126 @@
...
+    with:
+      expected-commit: 35a7dbc754193523e5582ce5b5cbbb4c07cff7b2
+      repository: https://github.com/tensorflow/build
+      branch: master
+
+  - working-directory: tensorflow_runtime_dockerfiles
+    pipeline:
+      - name: Install cpu requirements
+        runs: |
+          pip${{vars.py-version}} install \
```
EyeCantCU, line 48: Instead of installing directly to site packages, we'll need to install to a virtual environment

tensorflow-cpu.yaml
EyeCantCU, line 1: I'd set `package.options.no-provides` and `no-depends` to `true` as we're bundling several packages together

tensorflow-cpu.yaml
```diff
@@ -0,0 +1,126 @@
...
+  target-architecture:
+    - x86_64
+  dependencies:
+    runtime:
+      - bash
+      - build-base
+      - curl
+      - pkgconf
+      - py${{vars.py-version}}-pip
+      - python-${{vars.py-version}}-dev
```
EyeCantCU, line 18: Is it intentional to ship the dev variant of Python?

tensorflow-cpu.yaml
```diff
@@ -0,0 +1,126 @@
...
+      - unzip
+
+vars:
+  py-version: 3.11
```
EyeCantCU, line 22: I'm sure there's a reason, but why do we pin to 3.11?

tensorflow-cpu.yaml
```diff
@@ -0,0 +1,126 @@
...
+        # Perform a simple computation
+        a = tf.constant([1.0, 2.0, 3.0])
+        b = tf.constant([4.0, 5.0, 6.0])
+        c = tf.add(a, b)
+        print("Result of addition:", c.numpy())
+        EOF
+        python${{vars.py-version}} test.py
+    - name: Test Jupyter notebook run
+      runs: |
+        jupyter notebook --notebook-dir=/tf --ip 0.0.0.0 --no-browser --allow-root >/dev/null 2>&1 &
```
EyeCantCU, line 114: We have a pipeline for daemons we could reuse here, something like `test/daemon-check-output`

tensorflow-cpu.yaml
```diff
@@ -0,0 +1,126 @@
...
+        pid=$!
+        sleep 5
+        curl --fail --location --request GET --head --silent localhost:8888
+        kill $pid
+    - runs: python${{vars.py-version}} -m pip --version
```
EyeCantCU, line 119: This looks safe to remove

tensorflow-cpu.yaml
```diff
@@ -0,0 +1,126 @@
...
+    contents:
+      packages:
+        - curl
+  pipeline:
+    - name: Test import of IPython kernel for Jupyter
+      uses: python/import
+      with:
+        imports: |
+          import ipykernel.kernelspec
+    - name: Test Tensorflow library
```
EyeCantCU, line 91: This is great!

EyeCantCU (changes requested):

tensorflow-cpu.yaml
maxgio92: Oh yeah!

tensorflow-cpu.yaml
maxgio92: Yep, I just stuck to what upstream does ([here](https://github.com/tensorflow/build/blob/35a7dbc754193523e5582ce5b5cbbb4c07cff7b2/tensorflow_runtime_dockerfiles/setup.python.sh#L31)). Do you think it's only needed at build time for the next pip-installs?

tensorflow-cpu.yaml
maxgio92: Also, it seems that 3.12 is not officially supported in their build setups (for which the [issue](https://github.com/tensorflow/build/issues/210)), even though I'm not sure it's an hard requirement.

maxgio92: I just added a check that will fail the build when we're using a different version from the default one supported by upstream, referencing their Dockerfile's ARG.

tensorflow-cpu.yaml
EyeCantCU: @maxgio92 I'm honestly unsure. Could we test with the non-dev variant?

tensorflow-cpu.yaml
EyeCantCU: Awesome!

tensorflow-cpu.yaml
maxgio92: Sure @EyeCantCU. Tests and build seem to pass even without that 🤔

EyeCantCU: Let's send it and come back if problematic

EyeCantCU (approved):

PR #37793, opened 2024-12-19
Drop unused python or python deps from build environment.

smoser:
*  fish - simplify build-deps on python in order to run sphinx.
    the build just used sphinx-build. no reason to specify its deps.
 *  fattrace - drop unused python3 from build
 *  clamav-1.4 - drop unused python3 from build env
 *  clamav - drop unused python3 from build env
 *  configurable-http-proxy - drop unused python build-time dep.

### DIFF
--- a/clamav-1.4.yaml
+++ b/clamav-1.4.yaml
@@ -1,7 +1,7 @@
 package:
   name: clamav-1.4
   version: 1.4.1
-  epoch: 1
+  epoch: 2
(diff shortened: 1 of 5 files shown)


roaksoax (approved):
lgtm! The only question is clamav update config check

dannf:
> lgtm! The only question is clamav update config check

It was failing a download so I re-ran it, but now it claims to have spotted a potential CVE.

PR #37807, opened 2024-12-19
xcb-proto - split python package out

smoser:
this was pulling in python in a lot of places where it probably was not necessary.


roaksoax (approved):
lgtm!

PR #38007, opened 2024-12-20
procps: fix tests by using absolute path for kill

sil2100:
No epoch bump needed as this is just a test fix. The tests are failing as executions of `kill` always prioritize the builtin implementation. From what I know this is not only the case for busybox, but also bash - where for bash a `enable -n kill` is needed to disable this behavior. Since we want to test the procps implementation, one way of doing it is just using the absolute path for the kill binary.

### DIFF
--- a/procps.yaml
+++ b/procps.yaml
@@ -71,9 +71,10 @@ test:
         ps --version || exit 1
         free --version
         free --help
-        kill --help
+        # Use full path as kill is always preferred via the shell builtin
+        /usr/bin/kill --help
         sleep 9999 &
-        kill -9 $$
+        /usr/bin/kill -9 $!
         pgrep --version
         pgrep --help
         pidof --version


sil2100:
There is a new procps here: https://github.com/wolfi-dev/os/pull/37498 . I'd recommend landing this here and then pulling in the new version. But I don't know the recommended procedures just yet.

dannf (approved):
This all LGTM. Are you able to add the `approved-to-run` tag to kick off CI? I think you may need to be granted permission to do that. Let me know, I can add it until we get you perms.

Also, would you mind noting the `$$` -> `$!` fix in your commit message?

dannf:
> There is a new procps here: #37498 . I'd recommend landing this here and then pulling in the new version. But I don't know the recommended procedures just yet.

I don't know what is "correct" here - but what I would do here is to merge this one, then rebase #37498 and force push to that PR so it can proceed.

PR #38076, opened 2024-12-20
nodejs-*: Use more generic -march for rosetta2 compatibility

### LINKED ISSUE #37196: nodejs edge case broken - x86 on apple silicon
This likely doesn't affect many people but I wanted to document it in case other people start googling for it. We're trialing chainguard and I noticed builds on my machine that use nodejs were hanging. Switching to standard node or ubuntu base images fixed this as did disabling Rosetta in docker which lead me to think chainguard/wolfi was doing something different. 

To reproduce, get an ARM mac device and set docker to use rosetta then:
```shell
# this will hang forever
docker run --rm -it --platform linux/amd64 cgr.dev/chainguard/node:latest-dev -e "console.log(10 % 4)"
```

As you can imagine, the `%` operator is used loads of places so this manifested for me in npm builds that would hang:
```shell
# this also hangs forever
$ npm i -ddd -g mimic-function
npm verbose cli /usr/bin/node /usr/bin/npm
...
npm silly idealTree buildDeps
npm silly fetch manifest mimic-function@*
npm silly packumentCache full:https://registry.npmjs.org/mimic-function cache-miss
```

Under gdb (against node@21.7.3) we can see the loop the process is stuck in:
image

It seems Rosetta doesn't translate `sahf` so the loop never exists. When compiled the same as official builds of nodejs are, they use stock gcc on ubuntu (so `march=x86-64 + mtune=generic`), we see the generated code avoids this instruction:
image

I believe this all is caused by the tunes selected in [gcc.yaml](../blob/main/gcc.yaml), specifically [mtune=sapphirerapids](https://github.com/wolfi-dev/os/blob/main/gcc.yaml#L71) enables [a bunch of newer instructions](https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html#:~:text=%E2%80%98-,sapphirerapids,-%E2%80%99) which [Rosetta won't support](https://developer.apple.com/documentation/apple-silicon/about-the-rosetta-translation-environment#:~:text=Rosetta%20translates%20all%20x86_64%20instructions%2C%20but%20it%20doesn%E2%80%99t%20support%20the%20execution%20of%20some%20newer%20instruction%20sets%20and%20processor%20features%2C%20such%20as%20AVX%2C%20AVX2%2C%20and%20AVX512%20vector%20instructions.). Curiously `sahf` isn't listed in the Rosetta docs as not working but [others](https://github.com/Vita3K/Vita3K/pull/2038#issuecomment-1231936944) seem to be seeing issues with it.

**Potential workarounds:**
1. Disable Rosetta (you can disable it for one container with `EXPERIMENTAL_DOCKER_DESKTOP_FORCE_QEMU=1`)
2. Fetch [official](https://nodejs.org/en/download/prebuilt-binaries) nodejs build instead

justinvreeland:
Building node with `-march=x86_64-v2` is causing our nodejs to hang on
rosetta2.   Setting `-march=x86_64` seems to fix this issue.

Also remove the extra `autoconfg/make` directive.  Nodes build system doesn't carry over the flags from configure from what I can tell so the build that matters is the one at the end of Configure and Build.

19 is currently failing to build so not included.

Fixes: https://github.com/wolfi-dev/os/issues/37196
Related: https://github.com/chainguard-dev/customer-issues/issues/1866

### DIFF
--- a/nodejs-16.yaml
+++ b/nodejs-16.yaml
@@ -1,7 +1,7 @@
 package:
   name: nodejs-16
   version: 16.20.2
-  epoch: 8
+  epoch: 9
   description: "JavaScript runtime built on V8 engine"
   dependencies:
     provides:
@@ -47,7 +47,14 @@ pipeline:
   - name: Configure and build
     runs: |
       # Add defines recommended in libuv readme.
-      common_flags="-D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64"
+      case "${{build.arch}}" in
+      "aarch64")
+         common_flags="-D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64"
+         ;;
+      "x86_64")
+         common_flags="-D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -march=x86-64"
+         ;;
+      esac
 
       # Fix the standard to work with newer code
       sed s/gnu++14/gnu++17/g -i ./common.gypi
@@ -74,8 +81,6 @@ pipeline:
         --without-corepack
 
         make BUILDTYPE=Release -j$(nproc)
-
-  - uses: autoconf/make
 
   - uses: autoconf/make-install
 
--- a/nodejs-18.yaml
+++ b/nodejs-18.yaml
@@ -1,7 +1,7 @@
 package:
   name: nodejs-18
   version: 18.20.5
-  epoch: 1
+  epoch: 2
   description: "JavaScript runtime built on V8 engine - LTS version"
   copyright:
     - license: MIT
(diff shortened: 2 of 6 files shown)


xnox (approved):

xnox:
This is sad though.

Wish we could report this to NodeJS upstream and Apple.

liath:
I think it would be fine keep these compiler tunes and document that wolfi/chainguard does not support running under Rosetta intentionally. Likely this affects many packages built with these tunes and playing instruction set whack-a-mole is going to be frustrating.

justinvreeland:
I'm not planning on doing this for every instance we find but for now this is a known issue we can fix and we know people hitting this issue. Until we figure out the precise cause and report it to the relevant party this is fine.

PR #38108, opened 2024-12-21
add tshark package

jamonation:
#### For new package PRs only

#### For new version streams

- [X] The upstream project actually supports multiple concurrent versions.
- [X] The package (and subpackages) `provides:` logical unversioned forms of the package (e.g. `nodejs`, `nodejs-lts`)

### DIFF
--- a/linkerd2.yaml
+++ b/linkerd2.yaml
@@ -1,7 +1,7 @@
 package:
   name: linkerd2
   version: 24.11.8
-  epoch: 2
+  epoch: 3
   description: "meta linkerd package"
   copyright:
     - license: Apache-2.0
--- a/linkerd2.yaml
+++ b/linkerd2.yaml
@@ -176,6 +176,28 @@ subpackages:
           mkdir -p ${{targets.contextdir}}
           ln -sf /usr/bin/tap ${{targets.contextdir}}/tap
 
+  # the linkerd debug image doesn't have linkerd2 binaries, but has a number of tools
+  - name: ${{package.name}}-debug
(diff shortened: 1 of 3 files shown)


tshark.yaml
```diff
@@ -0,0 +1,59 @@
+package:
+  name: tshark
+  version: 4.4.2
+  epoch: 0
+  description: "TShark is a network protocol analyzer to ump and analyze network traffic"
```
jamonation, line 5:
```suggestion
  description: "TShark is a network protocol analyzer to dump and analyze network traffic"
```

kranurag7 (commented):
looking at package contents, I think we can do a `split/doc` and `split/dev` as well. 💭

jamonation:
> looking at package contents, I think we can do a `split/doc` and `split/dev` as well. 💭

Ah I tried split/manpages and got null 🤦 

split/dev isn't helpful here - it ends up grabbing the .so symlinks. Since tshark doesn't install any development headers, that subpackage shouldn't be needed.

kranurag7 (dismissed):

linkerd2.yaml
```diff
@@ -176,6 +176,28 @@ subpackages:
           mkdir -p ${{targets.contextdir}}
           ln -sf /usr/bin/tap ${{targets.contextdir}}/tap
 
+  # the linkerd debug image doesn't have linkerd2 binaries, but has a number of tools
```
jamonation, line 179: Need an epoch bump to get this

kranurag7 (approved):

PR #38175, opened 2024-12-21
py3-cassandra-medusa: pin python version

dannf:


dannf:
Reopening - let's try to just fix the python issues with this one, and we can discuss further how to deal with the security aspects in https://github.com/wolfi-dev/os/pull/38209

murraybd (approved):

PR #38392, opened 2024-12-26
Build yara-x with increased memory, --target-dir for cargo-c, and optimizations

egibs:
I had some local Melange changes that made #38328 appear valid when in fact the build artifacts for `cargo-c` were still landing in `/tmp`.

This PR specifies `--target-dir` to redirect `cargo-c`'s build artifacts to a non-`/tmp` directory to avoid running in to QEMU's `noexec` tmpfs. 

Once this was fixed, the `cargo-c` install was OOM'ing with the default `-m 4000000k` configuration so I added a memory requirement of `6Gi` which allowed the build to succeed.  Finally, I added `opt-level=3` to build the `yara-x-capi` with all available optimizations.

### DIFF
--- a/yara-x.yaml
+++ b/yara-x.yaml
@@ -17,6 +17,8 @@ environment:
       - perl
       - rust
       - wolfi-base
+  environment:
+    RUSTFLAGS: "-C opt-level=3"
 
 pipeline:
   - uses: git-checkout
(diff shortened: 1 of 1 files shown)


yara-x.yaml
```diff
@@ -23,9 +25,9 @@ pipeline:
       expected-commit: d2b8358879c009f41f0e14847681f2f8dbe624cf
       tag: v${{package.version}}
 
-  - runs: cargo install cargo-c --features=vendored-openssl --root $HOME/.cargo
+  - runs: cargo install cargo-c --features=vendored-openssl --root $HOME/.cargo --target-dir $HOME/.cargo
 
-  - runs: cargo cinstall -p yara-x-capi --release --prefix=/usr --pkgconfigdir=/usr/lib/pkgconfig --includedir=/usr/include --libdir=/usr/lib
+  - runs: RUSTFLAGS="-C opt-level=3" cargo cinstall -p yara-x-capi --release --prefix=/usr --pkgconfigdir=/usr/lib/pkgconfig --includedir=/usr/include --libdir=/usr/lib
```
dannf, line 30: Nit: does it make sense to set this in the environment block?

dannf (dismissed):
just a nit inline

yara-x.yaml
egibs: Good catch -- moved in [`7b5fe52` (#38392)](https://github.com/wolfi-dev/os/pull/38392/commits/7b5fe5244ff1b64a8c485cb838b1bc93b608f393).

dannf (approved):

PR #38512, opened 2024-12-30
add necessary runtime dependencies to apache-tika

developer-guy:
Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/apache-tika-3.0.yaml
+++ b/apache-tika-3.0.yaml
@@ -1,13 +1,14 @@
 package:
   name: apache-tika-3.0
   version: 3.0.0
-  epoch: 2
+  epoch: 3
   description: The Apache Tika toolkit detects and extracts metadata and text from over a thousand different file types (such as PPT, XLS, and PDF).
   copyright:
     - license: Apache-2.0
   dependencies:
     runtime:
-      - openjdk-17-jre
+      - dash-binsh
+      - openjdk-17-default-jvm
     provides:
       - apache-tika=${{package.full-version}}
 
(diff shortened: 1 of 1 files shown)


apache-tika-3.0.yaml
```diff
@@ -1,13 +1,14 @@
...
-  epoch: 2
+  epoch: 3
   description: The Apache Tika toolkit detects and extracts metadata and text from over a thousand different file types (such as PPT, XLS, and PDF).
   copyright:
     - license: Apache-2.0
   dependencies:
     runtime:
-      - openjdk-17-jre
+      - dash-binsh
+      - openjdk-17-default-jvm
```
kranurag7, line 11: Is whole jvm required here or we can work with jre only? 💭

developer-guy: as they use "exec java" in their entrypoint in the image config https://github.com/apache/tika-docker/blob/main/full/Dockerfile#L77 its necessary to add jvm here 🫠

kranurag7: that looks good 🫠

the CVE check is failing as of now, do we need to file advisories for it if it's not fixable by bumping the deps?

developer-guy: so the PR is in draft mode, I'm going to mark it as ready for review once the local tests are succeed @kranurag7

developer-guy: its ready for review brother, thank you!

apache-tika-3.0.yaml
```diff
@@ -63,9 +64,9 @@ test:
   pipeline:
     - name: "start server and verify endpoint"
       runs: |
-        TIKA_JAR=$(ls /tika-server-standard-*.jar 2>/dev/null | head -n 1)
+        TIKA_JAR=$(ls /tika-server-standard.jar 2>/dev/null | head -n 1)
```
kranurag7, line 67: I think we can hardcode this now to `tika-server-standard.jar` after the symlink change was introduced. 💭 

`ls /tika-server-standard.jar 2>/dev/null | head -n 1` this will always return the same name.

developer-guy: definitely!

kranurag7 (approved):
thank you!! looks good. o7

PR #37393, opened 2024-12-18
gobject-introspection: rebuild against 3.13

sil2100:
We see a FTBFS of `at-spi2-core`. The main reason is that `gobject-introspection` was built against 3.12, and the build can't work with 3.13. Rebuild it and also switch to using the py-versioned versions of packages.

Fixes: https://github.com/wolfi-dev/os/issues/37369

Related: https://github.com/wolfi-dev/os/issues/37197

PR #38514, opened 2024-12-30
Add a disabled test of g-ir-scanner

murraybd:
The test currently fails as detailed in https://github.com/wolfi-dev/os/pull/37393#issuecomment-2552811148 but we don't want to lose track of the fact that we should be testing g-ir-scanner so let's add it as a comment.


dannf (commented):
I think @smoser fixed this test in commit 4780497f930689d30e4723027e0ccfd655e58a2a. It exists earlier in the pipeline uncommented, and WFM :)

murraybd:
Wow, my bad.

PR #38520, opened 2024-12-30
rebuild ruby3.4-fluentd-kubernetes-daemonset-1.17

Dentrax:
Rebuild the `ruby3.4-fluentd-kubernetes-daemonset-1.17` since it's built with corrupted ruby-3.4.


kranurag7:
do we want to fix the cves in here or file advisories for the same?

kranurag7 (approved):
advisories or fix will follow, unblocking this for now.

PR #38533, opened 2024-12-30
ruby 3.4: Bump gems for released ruby batch 6

justinvreeland:


justinvreeland:
Not sure why this PR is taking so long to build. I'll be splitting it into smaller pieces instead.

justinvreeland:
It really does seem like the gems in https://github.com/wolfi-dev/os/pull/38631 were causing the huge time issues.

imjasonh (approved):

PR #38558, opened 2024-12-31
dpkg improvements

dannf:
Make gnutar a runtime dependency + some minor test cleanups.

### DIFF
--- a/dpkg.yaml
+++ b/dpkg.yaml
@@ -8,6 +8,7 @@ package:
   dependencies:
     runtime:
       - gnutar # dpkg requires tar option '--warning=no-timestamp'
+      - xz
 
 environment:
   contents:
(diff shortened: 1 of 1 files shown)


dpkg.yaml
```diff
@@ -119,7 +122,6 @@ test:
         - bzip2
         - libmd
```
justinvreeland, line 123: While we're here a lot of these look wrong for tests. I don't see any of these used directly so I'd expect the package to bring them in as part of the install.

justinvreeland (dismissed):

dpkg.yaml
dannf: good call, fixed!

justinvreeland (dismissed):

justinvreeland (approved):

PR #38625, opened 2025-01-02
apache pulsar subpackages

MickeyPvX:
Adding a runtime dependency, two subpackages that are needed by the image to be built from this package.

- Pulsar startup scripts are just bash scripts that need tools like `ls` `dirname` `cd` etc. - added `coreutils` and `grep`
- `offloaders` and `connectors` are additional resources that aren't included in the main package tarball, but are needed in the `pulsar-all` container image (to-be-built).

Related: https://github.com/chainguard-dev/image-requests/issues/1654


apache-pulsar.yaml
```diff
@@ -51,6 +53,22 @@ subpackages:
     pipeline:
       - runs: ln -s /usr/share/java/pulsar ${{targets.contextdir}}/pulsar
 
+  - name: ${{package.name}}-connectors
```
mritunjaysharma394, line 56: Can we modify our existing tests in melange to capture these additional resources ( that is how we reached the conclusion that they are needed for the pulsar-all image? I think it will be great if we can verify both so that in case of future breakages we can identify the error here)

mritunjaysharma394: At the same time, I think both the additional resources are bring CVEs, can we patch them? If not, we can advisory for them later

MickeyPvX: Just saw that - moving to working on those CVEs

As far as the subpackages, I found them while starting to test the [Pulsar helm chart](https://github.com/apache/pulsar-helm-chart/tree/master/charts/pulsar) with an image that just had the `apache-pulsar` package installed, saw that the image being used differed from the one on the issue (`pulsar-all` vs. just `pulsar`) and then found the additional resources being loaded onto the `pulsar-all` [image](https://github.com/apache/pulsar/blob/master/docker/pulsar-all/Dockerfile#L27).

I'll try to see if there's a way to test the subpackages though

mritunjaysharma394: Thanks @MickeyPvX , yes testing would be great if possible, if not we can focus on mitigation of CVEs, probably a pom/bump or patch of gradle can come to rescue unless it is the case of nested jars and then probably we will have to move to advisory in worst case

PR #38631, opened 2025-01-02
ruby3.4-activemodel: Update to match activesupport version

justinvreeland:
Also bump them to rebuild with the correct path not that ruby has been updated.

### DIFF
--- a/ruby3.4-activemodel.yaml
+++ b/ruby3.4-activemodel.yaml
@@ -1,6 +1,6 @@
 package:
   name: ruby3.4-activemodel
-  version: 8.0.0
+  version: 8.0.1
   epoch: 0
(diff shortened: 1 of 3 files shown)


ruby3.4-activemodel.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: ruby3.4-activemodel
-  version: 8.0.0
-  epoch: 0
+  version: 8.0.1
+  epoch: 1
```
kranurag7, line 4: not a blocker but should this be 0 because we're also bumping the version for this package.

kranurag7 (dismissed):

kranurag7 (approved):
thank you!!

PR #38662, opened 2025-01-02
Provide udev package from systemd

justinvreeland:
multipath-tools: use systemd-udev
thingsboard:  use systemd udev
kmod: remove uneeded dependecies

Eudev will be removed once all packages are done.

### DIFF
--- a/systemd.yaml
+++ b/systemd.yaml
@@ -1,7 +1,7 @@
 package:
   name: systemd
   version: "257.2"
-  epoch: 3
+  epoch: 4
   description: The systemd System and Service Manager
   copyright:
     - license: LGPL-2.1-or-later AND GPL-2.0-or-later
@@ -301,6 +301,38 @@ subpackages:
             # # test if it works
             ls -la /var/lib/machines/test
 
+  - name: "udev"
+    description: "/dev hotplug management daemon"
+    dependencies:
+      runtime:
+        - libsystemd-shared
+    pipeline:
+      - runs: |
+          cd ${{targets.destdir}}/
+          find . -name '*udev*' -exec cp -v --parents -r '{}' ${{targets.subpkgdir}} \;
+          find . -name '*systemd-hwdb*' -exec cp -v --parents -r '{}' ${{targets.subpkgdir}} \;
+
+          find . -name '*udev*' -exec rm -f '{}' \;
+          find . -name '*systemd-hwdb*' -exec rm -f '{}' \;
+    test:
+      pipeline:
+        - runs: |
+            systemd-hwdb --version
+            systemd-hwdb --help
+            udevadm --help
+            udevadm --version
+        - name: Systemd-hwdb checks
+          runs: |
+            systemd-hwdb update -s
+
+            EXPECTED=$(cat <<EOF
+            ID_PCI_CLASS_FROM_DATABASE=Bridge
+            ID_PCI_SUBCLASS_FROM_DATABASE=ISA bridge
+            EOF
+            )
+
+            [ "$EXPECTED" = "$(systemd-hwdb query pci:v00001234d00007000sv00001AF4sd00001100bc06sc01i00)" ]
+
   - name: "systemd-init"
     description: "Configure systemd for use as an init system"
     dependencies:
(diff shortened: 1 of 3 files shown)


systemd.yaml
```diff
@@ -114,6 +114,27 @@ subpackages:
...
+      - runs: |
+          cd ${{targets.destdir}}/
+          find . -name '*udev*' -exec cp -v --parents -r '{}' ${{targets.subpkgdir}} \;
+          find . -name '*systemd-hwdb*' -exec cp -v --parents -r '{}' ${{targets.subpkgdir}} \;
+
+          find . -name '*udev*' -exec rm -f '{}' \;
+          find . -name '*systemd-hwdb*' -exec rm -f '{}' \;
+    test:
+      pipeline:
+        - runs: |
```
89luca89, line 132: we can add `systemd-hwdb update -s` to the tests, it seems it does not rely on running udev/systemd 
Also we can do a generic query and check if output is valid:

```console
systemd-hwdb query pci:v00001234d00007000sv00001AF4sd00001100bc06sc01i00
ID_PCI_CLASS_FROM_DATABASE=Bridge
ID_PCI_SUBCLASS_FROM_DATABASE=ISA bridge
```

for example

89luca89 (dismissed):

systemd.yaml
smoser: > we can add `systemd-hwdb update -s` to the tests, it seems it does not rely on running udev/systemd Also we can do a generic query and check if output is valid:
> 

looks done below, thanks @justinvreeland

smoser (dismissed):

89luca89 (dismissed):

89luca89 (dismissed):

xnox (changes requested):
libudev.so.1 is usable without udevd daemon/tools; as it provides netlink interfaces to the kernel provided events without need of udevd daemon running.

We have a lot more packages that use libudev.so.1 than those listed here.

Splitting udev deamon and tools into a subpackage is good, but can also remain in the main systemd package.

However, it would be very nice to split runtime libudev.so.1 into a standalone libudev package, to kill off eudev, and ensure that lots of dependants can continue using it, without needing to pull in udev binaries/daemon.

Please see:
- https://apk.dag.dev/https/packages.wolfi.dev/os/x86_64/APKINDEX.tar.gz@etag:1940b07138c37e1dc44c2aefddc97507/APKINDEX?full=false&search=&depend=so%3Alibudev.so.1&provide=

xnox:
Wait I am consufed, library split was already done in https://github.com/wolfi-dev/os/pull/31612/files

So this PR is really about the udev binaries and daemon.

xnox (dismissed):
Ok, this all looks good.

Once merged we should decide if we want to always only ship pre-computed hwdb.

systemd.yaml
```diff
@@ -273,6 +273,38 @@ subpackages:
             # # test if it works
             ls -la /var/lib/machines/test
 
+  - name: "udev"
```
xnox, line 276: Rebase dropped epoch bump unfortunately.

xnox (changes requested):

systemd.yaml
justinvreeland: Wonder if this fixes the build/test issue I couldn't reproduce.

89luca89 (dismissed):

xnox:
> textPayload: "> Invalid -Xlog option '-Xlog:gc*,heap*,age*,safepoint=debug:file=/var/log/tb-mqtt-transport/gc.log:time,uptime,level,tags:filecount=10,filesize=10M', see error log for details."

can we figure out what is wrong with thingsboard? or possibly drop it from this PR, to see if the rest of the things work?

justinvreeland:
> > textPayload: "> Invalid -Xlog option '-Xlog:gc*,heap*,age*,safepoint=debug:file=/var/log/tb-mqtt-transport/gc.log:time,uptime,level,tags:filecount=10,filesize=10M', see error log for details."
> 
> can we figure out what is wrong with thingsboard? or possibly drop it from this PR, to see if the rest of the things work?

Dropped it. I'll take a look at it separately. I don't seem to have problems with my workstation or locally but we've had falykness with this package before.

xnox (approved):
Malcontent issue is expected behaviour, and that package will not be shipped in production images.

PR #38673, opened 2025-01-02
systemd: Add container package for container related tools and daemons

justinvreeland:
This new package contains the follow pieces used for container and image management which are not needed for most systems.

Notably sysext and context are being left in the main systemd package but portablectl is being included here. Debian puts it in their system-container package but fedora doesn't. I think it fits well here so that's where I put it.

machine
portable
systemd-disect
vm/nspawn
import/export

### DIFF
--- a/systemd.yaml
+++ b/systemd.yaml
@@ -1,7 +1,7 @@
 package:
   name: systemd
   version: "257.2"
-  epoch: 1
+  epoch: 2
   description: The systemd System and Service Manager
   copyright:
     - license: LGPL-2.1-or-later AND GPL-2.0-or-later
@@ -148,6 +148,7 @@ subpackages:
     dependencies:
       runtime:
         - systemd
+        - systemd-container
     pipeline:
       - runs: |
           mkdir -p ${{targets.subpkgdir}}/usr/lib/systemd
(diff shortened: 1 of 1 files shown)


justinvreeland:
This is on top of https://github.com/wolfi-dev/os/pull/37794

smoser (dismissed):

89luca89 (dismissed):

justinvreeland:
This PR contains the changes in https://github.com/wolfi-dev/os/pull/37794 that still needs to be approved. Github doesn't really do dependent PRs.  I'm used to and prefer to get a few different commits rather than one big squashed one, though it looks like we do have merge commits enabled.  If anyone feels strongly I'm happy to just merge this one and close the other one.

justinvreeland:
Rebased now that the previous PR landed.

89luca89 (approved):

89luca89:
Added just an additional test, which uncovered a missing dep, LGTM :+1:

systemd.yaml
```diff
@@ -257,6 +263,14 @@ subpackages:
             /usr/lib/systemd/systemd-portabled --version
             /usr/lib/systemd/systemd-pull --help
             /usr/lib/systemd/systemd-pull --version
+        - runs: |
+            # Run a pull command
+            SYSTEMD_LOG_LEVEL=debug /usr/lib/systemd/systemd-pull tar https://dl-cdn.alpinelinux.org/alpine/v3.21/releases/x86_64/alpine-minirootfs-3.21.2-x86_64.tar.gz test --verify=checksum --verify=checksum --direct --force
+
```
xnox, line 270: does double --verify=checksum needed?

xnox (approved):

systemd.yaml
xnox: can resolve this later.

89luca89: oh that's a typo sorry

PR #38675, opened 2025-01-02
Adds luarocks

srishtih:


luarocks.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - uses: git-checkout
+    with:
+      repository: https://github.com/luarocks/luarocks
+      tag: v${{package.version}}
+      expected-commit: fa2736e7c68094710643577b0153dac72ce4bcdc
+
+  - runs: |
+      ./configure --prefix=/usr --with-lua=/usr
+      make build
+      DESTDIR="${{targets.destdir}}" make install
```
kranurag7, line 29:
```suggestion
  - runs: ./configure --prefix=/usr --with-lua=/usr
  - uses: autoconf/make
  - uses: autoconf/make-install
```
the host flag is something I was not able to hack on. Even passing `""` doesn't make the `autoconf/configure` pipeline happy. 

One other thing is that `apk add luarocks` in the current state pulls in fluentbit which is something I think is not required by this package. 💭

srishtih:
fluent-bit provides [luajit ](https://apk.chaindag.dev/https/packages.wolfi.dev/os/aarch64/fluent-bit-3.0.2-r0.apk@sha1:6c4a8a7dc94a9a0bd6a4f5c9d08f3be4ee03687e/.PKGINFO)

EyeCantCU:
Not missing anything. Looking at the PKGINFO here: https://apk.chaindag.dev/https/apk.cgr.dev/wolfi-presubmit/7adcf3524e4a76d18c7d74e463f9fec978ca5ead/x86_64/luarocks-3.11.1-r0.apk@etag:8a3abe44de40e7f716710cc39ceb0dfa/.PKGINFO

We generate a dependency on `cmd:luajit` because it's used in the shebang of a script. fluent-bit doesn't provide the same luajit implementation (it's tailored for use with fluent-bit).

We can ignore luajit when it's found in the shebang of scripts to stop generating a dependency on `cmd:luajit` to avoid the conflict. Short term, let's set `no-depends` here and explicitly add `luajit` at runtime

EyeCantCU:
Opened a PR to melange preventing SCA from generating a dependency on `cmd:luajit` when found in the shebang

luarocks.yaml
```diff
@@ -0,0 +1,55 @@
...
+  description: "Package manager for Lua modules"
+  copyright:
+    - license: MIT
+  options:
+    no-depends: true
+  dependencies:
+    runtime:
+      - curl
+      - gcc
+      - luajit
```
jamonation, line 14:
```suggestion
      - git
      - luajit
```

luarocks uses git for some downloads

maxgio92:
LGTM, just a small point on testing

luarocks.yaml
```diff
@@ -0,0 +1,56 @@
...
+test:
+  environment:
+    contents:
+      packages:
+        - luajit-dev
+        - glibc-dev
+  pipeline:
+    - runs: |
+        luarocks --version
+        luarocks install luasocket
```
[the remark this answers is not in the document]
jamonation: Yes I'd like to see a test that checks for the case that's led to adding this package, e.g. 
1. add openresty, 
2. add an openresty config that uses zlib/lzlib
3. create a curl request sending compressed data
4. check openresty log indicating it read and decompressed the data

maxgio92 (approved):
LGTM!
Thanks @srishti!

PR #38712, opened 2025-01-03
py3-cassandra-medusa: re-remove the grpc-health-probe dep

sil2100:
Possibly added by mistake. The symlink seems to be 'fine' when it's dangling. This also fixes some incompatibilities when fips versions are involved in images.


dannf (approved):
Thanks for noticing that this was intentionally removed in https://github.com/wolfi-dev/os/pull/20727.

PR #38717, opened 2025-01-03
systemd: add standalone binaries

justinvreeland:


justinvreeland:
This is really just about the top one. But it's built on the others.

xnox (dismissed):
looks good, but needs rebase with epochs.

systemd.yaml
```diff
@@ -85,6 +93,27 @@ pipeline:
...
+          mv ${{targets.destdir}}/usr/bin/${{range.value}} ${{targets.subpkgdir}}/usr/bin/${{range.key}}
+    test:
+      pipeline:
+        - runs: |
+            ${{range.key}} --help
+            ${{range.key}} --version
+
+  - name: systemd-shutdown-standalone
+    description: Standalone version of ${{range.key}}
+    # Can't test this with --version becuase the binary checks the pid before anything else
```
imjasonh, line 111:
```suggestion
    # Can't test this with --version because the binary checks the pid before anything else
```

xnox (approved):

PR #38777, opened 2025-01-04
datadog-agent: Re-enable GPU support and some python cleanup

dannf:
See the individual commit logs for details.


datadog-agent.yaml
```diff
@@ -81,6 +80,8 @@ environment:
     CGO_CFLAGS: "-Os -I/usr/include/"
     # Pass options to the linker.
     CGO_LDFLAGS: "-L/usr/lib/"
+    # See https://github.com/wolfi-dev/os/issues/34568
+    GCC_SPEC_FILE: /home/build/openssf.spec
```
justinvreeland, line 84: Should we be using absolute paths here? It doesn't seem likely that we won't change the path we build in. but I believe our patches and the like are all using relative paths. Similarly with writing git to the absolute path maybe we should just be reading/writing the files from the current directory?

justinvreeland (approved):

PR #38780, opened 2025-01-04
fix missing static/assets issue

developer-guy:
There was an issue with the superset package which is the static/assets folder does not exist, so, this PR aims to fix this.

### DIFF
--- a/superset.yaml
+++ b/superset.yaml
@@ -1,7 +1,7 @@
 package:
   name: superset
   version: 4.1.1
-  epoch: 0
+  epoch: 1
   description: Data Visualization and Data Exploration Platform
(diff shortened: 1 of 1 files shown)


kranurag7 (approved):
thank you, looked at [static assets](https://apk.chaindag.dev/https/apk.cgr.dev/wolfi-presubmit/0a103c763d972b2c28215adb4041c5e3a4c5327e/x86_64/superset-4.1.1-r1.apk@sha1:437ee5f1e9b19493f777792160c120f462bc5534/usr/share/superset/venv/lib/python3.11/site-packages/superset/static/assets/) of the package and looks good to me.

PR #38842, opened 2025-01-06
Feat: python multiversion build for gi-docgen

debasishbsws:
recently while working on gtk-4 https://github.com/wolfi-dev/os/pull/38405 package we encounter that there is a python version mismatch for gi-docgen and we are getting an older version of py3-gi-docgen due to matching version for some other package in the build env.

The best and optimal solution we discussed in slack is to multiversionised our gi-docgen and build it for every supported version of python available


gi-docgen.yaml
```diff
@@ -1,10 +1,22 @@
...
+
+vars:
+  pypi-package: gi-docgen
+
+data:
+  - name: py-versions
+    items:
+      3.11: '311'
+      3.12: '312'
+      3.13: '313'
```
debasishbsws, line 19: We are not building for python3.10 because of [toml version constrain ](https://gitlab.gnome.org/GNOME/gi-docgen/-/blob/main/meson.build#L42)

We are getting the error
```
meson.build:45:6: ERROR: Problem encountered: tomli 1.0 or newer required on Python 3.10 and older subpackage=py3.10-gi-docgen
```

gi-docgen.yaml
```diff
@@ -15,49 +27,84 @@ environment:
...
+            exit 1
+          }
+
+          echo "Using Python interpreter: $pyfullpath"
+
+          # Create a temporary directory and symlink for python3
+          tmp_dir=$(mktemp -d)
+          trap "rm -rf $tmp_dir" EXIT
+          ln -s "$pyfullpath" "$tmp_dir/python3"
+          export PATH="$tmp_dir:$PATH"
```
debasishbsws, line 76: meson build required python3 in the PATH, with `py3-supported-python` package it is not set with these we create a temp dir have a symlink of python3 for each version, add it in the PATH, complete the meson build
and cleanup the temp dir using `trap "rm -rf $tmp_dir" EXIT` at the end.

gi-docgen.yaml
```diff
@@ -15,49 +27,84 @@ environment:
...
+
+          # Cleanup previous build and compile
+          rm -rf output
```
debasishbsws, line 79: cleanup the previously compiled build if present

developer-guy (approved):

PR #38870, opened 2025-01-06
feat[package]: adds airflow-core

mritunjaysharma394:
Airflow-core package, based on upstream stating that their slim image is what airflow-core looks like https://airflow.apache.org/docs/docker-stack/index.html#docker-image-for-apache-airflow

Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches


airflow-core.yaml
```diff
@@ -0,0 +1,136 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/apache/airflow
+      tag: ${{package.version}}
+      expected-commit: c083e456fa02c6cb32cdbe0c9ed3c3b2380beccd
+
+  - runs: |
+      # by default airflow celery provider is not built,but running the upstream helm chart requires it
```
mamccorm, line 53: Theres more than just celery being inserted in patch.py. Can you share some more details about this requirement? Is there any upstream docs or Dockerfile we're referencing here? Also where should we check in future if there are issues maintaining this package and some of the settings in the python script may need to be changed?

airflow-core.yaml
```diff
@@ -0,0 +1,136 @@
...
+       rm -rf patch.py
+
+  - runs: |
+      # Set required build environment variables
+      export SOURCE_DATE_EPOCH=315532800
+      export MYSQLCLIENT_CFLAGS=$(mysql_config --cflags)
+
+      # Install airflow core, the upstream sets this as directory
+      export PYTHONUSERBASE=${{targets.destdir}}/home/airflow/.local
+      pip3.12 install --no-cache-dir --user apache-airflow[core]==${{package.version}}
```
mamccorm, line 65: Do we need to specify 'pip3.12' here or can we just use 'pip' given we've only one version installed? The latter would make maintaining this package easier in future, and less to update when we bump the python version.

Similar for the references below

airflow-core.yaml
```diff
@@ -0,0 +1,136 @@
...
+      cp scripts/docker/airflow-scheduler-autorestart.sh ${{targets.destdir}}/airflow-scheduler-autorestart
+      chmod 755 ${{targets.destdir}}/airflow-scheduler-autorestart
+
+      # unlike airflow, airflow-core needs only these scripts
+      cp scripts/docker/common.sh ${{targets.destdir}}/scripts/docker
+      cp scripts/docker/install_mssql.sh ${{targets.destdir}}/scripts/docker
+      cp scripts/docker/install_mysql.sh ${{targets.destdir}}/scripts/docker
+      cp scripts/docker/install_os_dependencies.sh ${{targets.destdir}}/scripts/docker
+      cp scripts/docker/install_postgres.sh ${{targets.destdir}}/scripts/docker
+      chmod 755 ${{targets.destdir}}/scripts/docker/*
```
mamccorm, line 105: I don't believe permissions are persisted in melange, this would need to be done at the image level, so can remove this line

airflow-core.yaml
```diff
@@ -0,0 +1,136 @@
...
+  environment:
+    contents:
+      packages:
+        - python3
+  pipeline:
+    - runs: |
+        export PATH=/root/bin:/home/airflow/.local/bin:/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+        export PYTHONPATH=/home/airflow/.local/lib/python3.12/site-packages
+        airflow version
+        python3.12 -c "import airflow"
```
mamccorm, line 136: Can we switch 'python3.12' references to 'python' - it'll make it easier to maintain this package when we swap out python versions in future

airflow-core.yaml
```diff
@@ -0,0 +1,136 @@
...
+      # `$AIRFLOW_HOME` directory
+      # However, for production case it is advised to generate the configuration
+      export PYTHONPATH=${{targets.destdir}}/home/airflow/.local/lib/python3.12/site-packages
+
+      ${{targets.destdir}}/home/airflow/.local/bin/airflow config list --defaults > ${{targets.destdir}}/opt/airflow/"airflow.cfg"
+
+      cp airflow/config_templates/default_webserver_config.py ${{targets.destdir}}/
+
+      cp scripts/docker/entrypoint_prod.sh ${{targets.destdir}}/entrypoint
+      chmod 755 ${{targets.destdir}}/entrypoint
```
mamccorm, line 91: I don't believe permissions are persisted in melange, this would need to be done at the image level, so can remove this line

(same for similar references below)

airflow-core.yaml
```diff
@@ -0,0 +1,136 @@
...
+      pip3.12 install --no-cache-dir --user apache-airflow[core]==${{package.version}}
+
+      pip3.12 install --no-cache-dir --user aiohttp>=3.10.11 tornado>=6.4.2
+
+      pip3.12 install --no-cache-dir --user --force-reinstall packaging statsd
+
+  - runs: |
+      # Cleanup unnecessary dependencies
+      pip3.12 uninstall --yes setuptools
+      find . -name '__pycache__' -exec rm -rf {} +
```
mamccorm, line 74: Do we need this? it'd be removing setuptools from the build host for this melange, not sure how that'd make any difference to the size of the final apk?

airflow-core.yaml
mritunjaysharma394: That's a good call, Mark, althought airflow-core will remain a minimal package so it might not even need this provider (that is something that we will know when we reach the stage of helm chart tests). My intention to have it here comes from the existing airflow package that reports this issue and verify if it works or not 
https://github.com/wolfi-dev/os/blob/66c102ad5a2b5defd16ce394e330b498efbaff4f/airflow.yaml#L50

airflow-core.yaml
mritunjaysharma394: Thanks Mark, yes I don't think that should be a problem, i though to have it explicitly 3.12 can help know that we are using that as that is a requirement by upstream but having it in environment should be enough too

airflow-core.yaml
mritunjaysharma394: Sure, will do that, thanks Mark

airflow-core.yaml
mritunjaysharma394: Actually the reference comes from here as airflow did that so I was not sure if we had some historical context to do it, we might have to refactor this too https://github.com/wolfi-dev/os/blob/66c102ad5a2b5defd16ce394e330b498efbaff4f/airflow.yaml#L113

airflow-core.yaml
```diff
@@ -0,0 +1,136 @@
...
+      export SOURCE_DATE_EPOCH=315532800
+      export MYSQLCLIENT_CFLAGS=$(mysql_config --cflags)
+
+      # Install airflow core, the upstream sets this as directory
+      export PYTHONUSERBASE=${{targets.destdir}}/home/airflow/.local
+      pip3.12 install --no-cache-dir --user apache-airflow[core]==${{package.version}}
+
+      pip3.12 install --no-cache-dir --user aiohttp>=3.10.11 tornado>=6.4.2
+
+      pip3.12 install --no-cache-dir --user --force-reinstall packaging statsd
```
mamccorm, line 69: We have py2-aiohttp, py3-tornado, py3-packaging and py3statsd already packaged up in wolfi. Any reason we are not defining our existing APKs as builtime dependencies instead of this?

airflow-core.yaml
mritunjaysharma394: I saw this context and kept it but probably we can dig deep and know if we should have it or not  
https://github.com/wolfi-dev/os/blob/66c102ad5a2b5defd16ce394e330b498efbaff4f/airflow.yaml#L73

airflow-core.yaml
```diff
@@ -0,0 +1,136 @@
...
+
+update:
+  enabled: true
+  ignore-regex-patterns:
+    - 'rc\d+$'
+    - 'helm-chart*'
+  github:
+    identifier: apache/airflow
+
+test:
```
mamccorm, line 126: Can we expand the tests to include any simple functional tests?

airflow-core.yaml
```diff
@@ -0,0 +1,136 @@
+package:
+  name: airflow-core
+  version: 2.10.4
+  epoch: 0
+  description: Minimal Airflow core - Platform to programmatically author, schedule, and monitor workflows
+  dependencies:
```
mamccorm, line 6: Do we need all of these as buildtime deps? bash and busybox and coreutils stood out, but curl etc?

airflow-core.yaml
mritunjaysharma394: Probably the pip install method (used here) is the reason to have these CVE fixes https://github.com/wolfi-dev/os/blob/66c102ad5a2b5defd16ce394e330b498efbaff4f/airflow.yaml#L79

airflow-core.yaml
mritunjaysharma394: Yes curl is in the upstream image, so is mariadb, etc 
```bash
(airflow)find curl
curl
(airflow)
```

airflow-core/patch.py
```diff
@@ -0,0 +1,52 @@
+import os
```
mamccorm, line 1: Previous comment in the melange invoking the patch - but would need some info here on the source of this patch / info where it was derived, and all the various settings it's adding in addition to head - so sustaining know what to do with this package

mamccorm (changes requested):
Thanks for this, some comments inline before we merge this one

airflow-core.yaml
mritunjaysharma394: From the discussion with Srishti and comments from this issue by RJ, I think we are good to go with this to align with what our full blown airflow package does

https://github.com/chainguard-dev/internal-dev/issues/8158#issuecomment-2580349351

airflow-core.yaml
mritunjaysharma394: Thanks, Mark, done

airflow-core/patch.py
mritunjaysharma394: Removed the patch as after multiple checks with upstream, airflow-core doesn't need it, if at all, as suggested by upstream project too, the user will have to add the providers on their own

airflow-core.yaml
mritunjaysharma394: Added context from our airflow package for this

airflow-core.yaml
```diff
@@ -0,0 +1,149 @@
...
+      - glibc-locale-en
+      - glibc-locales
+      - gnupg-utils
+      - gzip
+      - mariadb-dev
+      - mysql
+      - openssl
+      - postgresql-client
+      - postgresql-dev
+      - python-3.12
```
EyeCantCU, line 24: Let's create a var called `py-version`, set to 3.12, and reuse that. Will make it easier to transition to new versions of Python Also, is there a reason we've gone ahead with 3.12 here?

airflow-core.yaml
```diff
@@ -0,0 +1,149 @@
...
+      - glibc-dev
+      - mariadb-connector-c-dev
+      - mariadb-dev
+      - openssf-compiler-options
+      - pkgconf-dev
+      - postgresql-dev
+      - py3.12-build-bin
+      - py3.12-pip
+      - python-3.12
+      - python-3.12-dev
```
EyeCantCU, line 41: Same here, reference a global `py-version` variable

airflow-core.yaml
```diff
@@ -0,0 +1,149 @@
...
+      - wolfi-base
```
EyeCantCU, line 42: Can be dropped

airflow-core.yaml
```diff
@@ -0,0 +1,149 @@
...
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/apache/airflow
+      tag: ${{package.version}}
+      expected-commit: c083e456fa02c6cb32cdbe0c9ed3c3b2380beccd
+
+  - runs: |
+      # Set required build environment variables
+      export SOURCE_DATE_EPOCH=315532800
```
EyeCantCU, line 54: Set this in the package environment with an explanation on why it's needed

airflow-core.yaml
```diff
@@ -0,0 +1,149 @@
...
+
+      # Install airflow core, the upstream sets this as directory
+      export PYTHONUSERBASE=${{targets.destdir}}/home/airflow/.local
+      pip install --no-cache-dir --user apache-airflow[core]==${{package.version}}
+
+      #GHSA-8w49-h785-mj3c/GHSA-8495-4g3g-x7pr/GHSA-27mf-ghqm-j3j8 fixes
+      pip install --no-cache-dir --user aiohttp>=3.10.11 tornado>=6.4.2
+      pip install --no-cache-dir --user --force-reinstall packaging statsd
+
+  - runs: |
```
EyeCantCU, line 65: If we aren't breaking things into labelled steps, I'd prefer we just used whitespace to separate things logically rather than invoking run again

airflow-core.yaml
```diff
@@ -0,0 +1,149 @@
...
+      cp scripts/docker/airflow-scheduler-autorestart.sh ${{targets.destdir}}/airflow-scheduler-autorestart
+      chmod 755 ${{targets.destdir}}/airflow-scheduler-autorestart
+
+      # unlike airflow, airflow-core needs only these scripts
+      cp scripts/docker/common.sh ${{targets.destdir}}/scripts/docker
+      cp scripts/docker/install_mssql.sh ${{targets.destdir}}/scripts/docker
+      cp scripts/docker/install_mysql.sh ${{targets.destdir}}/scripts/docker
+      cp scripts/docker/install_os_dependencies.sh ${{targets.destdir}}/scripts/docker
+      cp scripts/docker/install_postgres.sh ${{targets.destdir}}/scripts/docker
+      chmod 755 ${{targets.destdir}}/scripts/docker/*
```
EyeCantCU, line 103: In the past, we've typically split this into a separate entrypoint package, under the assumption that the package itself can operate independently

airflow-core.yaml
EyeCantCU: Let's always handle in the package when we can, this works fine

airflow-core.yaml
```diff
@@ -0,0 +1,149 @@
...
+    - 'rc\d+$'
+    - 'helm-chart*'
+  github:
+    identifier: apache/airflow
+
+test:
+  environment:
+    contents:
+      packages:
+        - python3
```
EyeCantCU, line 118: We need to pin Python here

airflow-core.yaml
```diff
@@ -0,0 +1,149 @@
...
+        - bash
+  pipeline:
+    - runs: |
+        # Set environment variables
+        export PATH=/home/airflow/.local/bin:$PATH
+        export PYTHONPATH=/home/airflow/.local/lib/python3.12/site-packages
```
EyeCantCU, line 124: Another place to use a re-usable var (`py-version`)

EyeCantCU (changes requested):

airflow-core.yaml
mritunjaysharma394: Thanks @EyeCantCU , yes there's a reason as airflow most latest supported python version as we speak is 3.12 https://airflow.apache.org/blog/airflow-2.9.0
I have made the corresponding changes wrt using var `py-version`

airflow-core.yaml
mritunjaysharma394: Done, thanks RJ

airflow-core.yaml
mritunjaysharma394: Done

airflow-core.yaml
mritunjaysharma394: Done

airflow-core.yaml
mritunjaysharma394: done, thanks RJ

airflow-core.yaml
mritunjaysharma394: Thanks for the heads up, done

airflow-core.yaml
mritunjaysharma394: Done, splitted this in another package

airflow-core.yaml
EyeCantCU: Even PostgreSQl dev?

airflow-core.yaml
```diff
@@ -0,0 +1,193 @@
...
+      - cyrus-sasl-dev
+      - findutils
+      - glibc-locale-en
+      - glibc-locales
+      - gnupg-utils
+      - gzip
+      - mariadb-dev
+      - openssl
+      - postgresql-client
+      - postgresql-dev
```
EyeCantCU, line 22: Looking great, but is all of PostgreSQL dev needed?

airflow-core.yaml
```diff
@@ -0,0 +1,193 @@
...
+  epoch: 0
+  description: Minimal Airflow core - Platform to programmatically author, schedule, and monitor workflows
+  dependencies:
+    runtime:
+      - bash
+      - busybox
+      - coreutils
+      - curl
+      - cyrus-sasl
+      - cyrus-sasl-dev
```
EyeCantCU, line 13: Same for these other dev dependencies

airflow-core.yaml
mritunjaysharma394: They seem to have few pg related binaries but seems like I just checked again, I think using `postgres-client` can be enough for them after confirming with apk search:

To name a few:
```bash
(airflow)find psql
psql
(airflow)find pg_dump
pg_dump
(airflow)
```

airflow-core.yaml
mritunjaysharma394: Thanks for taking a look at them, veried and confirmed that we can proceed without dev, apologies for not taking a deeper look earlier with apk search

airflow-core.yaml
mritunjaysharma394: I see this in upstream, that confused me
```bash
(airflow)pwd
/lib/postgresql/17/lib
(airflow)ls
pgxs
(airflow)
```

mritunjaysharma394: And I found this reference https://stackoverflow.com/a/7379005

airflow-core.yaml
mritunjaysharma394: Also to note that upstream has this:
```bash
(airflow)pwd
/lib/postgresql/17/lib
(airflow)ls
pgxs
(airflow)
```
This made me think that we might still PostgreSQL-dev or simple PostgreSQL at least?

EyeCantCU: @mritunjaysharma394 I'd imagine we still need PostgreSQL, but we shouldn't need the dev dependency at runtime

airflow-core.yaml
mamccorm: Are we sure we need postgresql-dev at buildtime? does the package build without this dependency if we drop it? Had a look a the stackoverflow article and reads more like a runtime issue.

mritunjaysharma394: Actually, that's is a very good catch and I am sorry for missing it myself, we don't need most build time dep because of pip install method here that we follow, I just rebuilt without them and it worked fine, other than few python deps, it needed only busybox. Thanks again!

mritunjaysharma394:
Closing, following up in different repo

PR #38875, opened 2025-01-06
Builds kong

srishtih:


kong/http-archive.patch
```diff
@@ -0,0 +1,19 @@
...
+
+ def kong_github_repositories():
+-    maybe(
+-        github_release,
++    http_archive(
+         name = "kong_admin_gui",
+-        repo = "kong/kong-manager",
+-        tag = KONG_VAR["KONG_MANAGER"],
+-        pattern = "release.tar.gz",
++        url = "https://github.com/Kong/kong-manager/releases/download/nightly/release.tar.gz",
```
EyeCantCU, line 17: We should match the version upstream uses here. I'd define a dummy string in the patch itself:

```suggestion
+        url = "https://github.com/Kong/kong-manager/releases/download/KONG_MANAGER_VERSION/release.tar.gz",
```

kong.yaml
```diff
@@ -0,0 +1,84 @@
...
+      - yaml-dev
+      - zlib-dev
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/Kong/kong
+      tag: ${{package.version}}
+      expected-commit: 0118b3c220681d4f0327540ae1a5e8f271081bf6
+
```
EyeCantCU, line 47: And before applying the patch, replace the dummy (`KONG_MANAGER_VERSION`) (you may need GNU grep):

```suggestion

    - runs: |
        KONG_MANAGER_VERSION=$(grep -oP '^KONG_MANAGER=\K.*' .requirements)
        sed -i "s/KONG_MANAGER_VERSION/${KONG_MANAGER_VERSION}/g" http-archive.patch

```

EyeCantCU (changes requested):

kong/http-archive.patch
srishtih: I checked, it uses the nightly version. Also the version mapping between kong and kong-gui is not very clean, so using nightly seems right

srishtih: https://github.com/Kong/kong/blob/df6cc598b65d4d205ae148812d0a673d2fc098b9/.requirements#L26

EyeCantCU (approved):

PR #38880, opened 2025-01-06
py3-cassandra-medusa: pin poetry version during build

dannf:
Fixes a ftbfs introduced by poetry 2.0's recent release.

Note: `ci-cve-scan` issues are being discussed in https://github.com/chainguard-dev/internal-dev/issues/7891 and shouldn't block this fix IMHO.


murraybd:
I kind of think it would be better to link to an issue with the details of the FTBFS rather than using a comment in the yaml file.

dannf:
> I kind of think it would be better to link to an issue with the details of the FTBFS rather than using a comment in the yaml file.

Yeah, good point.

dannf:
> > I kind of think it would be better to link to an issue with the details of the FTBFS rather than using a comment in the yaml file.
> 
> Yeah, good point.

Turns out upstream has already fixed this, so I've updated the comment. I've also cleaned up some of the other pins we're doing - see the individual commits for details.

murraybd (approved):

PR #38913, opened 2025-01-07
Fix(CVE): GHSA-rhh4-rh7c-7r5v for mattermost-10.3

debasishbsws:
The fork we are using is a drop in replacement which includes one extra commit and have released a tag
[`github.com/anchore/archiver`](https://github.com/anchore/archiver/)

Until the version 4.x.x gets released for [`github.com/mholt/archiver`](https://github.com/mholt/archiver) we can use this fork temporary for these CVE


mamccorm (approved):
We've done similar to address this CVE for other packages - approving on that basis

PR #38929, opened 2025-01-07
py3-jupyterhub-idle-culler: only the -bin package should provide the original

sil2100:
When doing the multiversioning changes, I had to split idle-culler into a -bin package. But while doing the split, even though the -bin package correctly provides the original py3-jupyterhub-idle-culler package, I forgot to remove the same provides: from the versioned py<version>-jupyterhub-idle-culler one.


kranurag7 (approved):
thank you!! looks good.

PR #38957, opened 2025-01-07
rename subpackage to jupyter-docker-base

kranurag7:
rename the subpackage

### DIFF
--- a/jupyter-docker-stacks.yaml
+++ b/jupyter-docker-stacks.yaml
@@ -2,7 +2,7 @@
 package:
   name: jupyter-docker-stacks
   version: 7.3.2
-  epoch: 0
+  epoch: 1
   description: Ready-to-run images containing Jupyter applications
   copyright:
     - license: BSD-3-Clause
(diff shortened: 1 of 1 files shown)


EyeCantCU (approved):

EyeCantCU (changes requested):
`jupyter-docker-stacks-base-notebook` also seems wrong since it isn't installing anything from the repo. I'd create a separate package for it and name `jupyter-notebook-base` and have it track the version of notebook via GitHub

kranurag7:
> `jupyter-docker-stacks-base-notebook` also seems wrong since it isn't installing anything from the repo. I'd create a separate package for it and name `jupyter-notebook-base` and have it track the version of notebook via GitHub

RJ, we can do that. I was mostly keeping it as subpackages because entrypoint scripts come from `jupyter/docker-stacks` repo. We do a checkout once and based on `working-directory` we divide it into subpackages. 

```bash
# copy entrypoint scripts
cp start-notebook.py start-singleuser.py start-notebook.sh start-singleuser.sh ${{targets.contextdir}}/usr/local/bin
```
We're copying this files into `jupyter-docker-stacks-base-notebook` subpackage. 

please let me know if this should still be splitted into two, I'll do that and withdraw the existing packages because those are not being used by any image as of now.

PR #38999, opened 2025-01-07
openfga: healthcheck compat package

MickeyPvX:
Puts the `openfga` compat package for its healthcheck in the right spot.

Related: https://github.com/chainguard-dev/image-requests/issues/5204


kranurag7 (changes requested):

kranurag7 (approved):
thanks again Mike, looks good.

PR #39033, opened 2025-01-07
Test rebuild of packages for the poetry update

murraybd:
We want to see if anything fails a rebuild with the 2.0 version of poetry.


murraybd:
If everything passes, which it should this time, we should merge this so the packages use the new version of poetry.

sil2100 (dismissed):
LGTM!

I also see that `py3-cassandra-medusa` build-deps on poetry. Is that one omitted because the latest change I pushed already built against poetry 2.0?

murraybd:
> LGTM!
> 
> I also see that `py3-cassandra-medusa` build-deps on poetry. Is that one omitted because the latest change I pushed already built against poetry 2.0?

Ah, I grepped for "py3.*poetry" which didn't catch `py${{vars.py-version}}-poetry` and some others.

murraybd:
> > LGTM!
> > I also see that `py3-cassandra-medusa` build-deps on poetry. Is that one omitted because the latest change I pushed already built against poetry 2.0?
> 
> Ah, I grepped for "py3.*poetry" which didn't catch `py${{vars.py-version}}-poetry` and some others.

However, `py3-cassandra-medusa` is the only one I missed so this is good to be merged.

dannf (dismissed):

dannf (approved):

imjasonh (approved):
SBOM check is timing out, I'll force merge this.

PR #39036, opened 2025-01-07
Adds docker entrypoint for kong

srishtih:


kong-entrypoint.yaml
```diff
@@ -0,0 +1,30 @@
...
+  version: 3.9.0
+  epoch: 0
+  description: "Provides entrypoint for kong"
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - bash
```
EyeCantCU, line 12: If the script uses bash, I'd move to runtime deps

kong-entrypoint.yaml
```diff
@@ -0,0 +1,30 @@
...
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/Kong/docker-kong.git
+      tag: ${{package.version}}
+      expected-commit: d46098c09b6198efb71682fe7209bd6467bf3f67
+
+  - runs: |
+      install -Dm755 ./docker-entrypoint.sh  -t ${{targets.destdir}}/docker-entrypoint.sh
+
```
EyeCantCU, line 25: Would add a test, even if just a stat, as any packages without tests need manual intervention for updates

EyeCantCU (changes requested):

EyeCantCU (approved):

PR #38563, opened 2024-12-31
pipelines: test: ldd-check: Add "packages" input parameter

dannf:
`gcc` currently fails to test on ARM because some of the filenames we provide to `test/ldd-check` are architecture-specific.

This should unblock https://github.com/wolfi-dev/os/pull/35302

PR #39042, opened 2025-01-07
pipelines: test: ldd-check: cleanups

dannf:
Ref: https://github.com/wolfi-dev/os/pull/38563/files#r1906010589

### DIFF
--- a/pipelines/test/ldd-check.yaml
+++ b/pipelines/test/ldd-check.yaml
@@ -38,64 +38,94 @@ pipeline:
       passes=0
       files="${{inputs.files}}"
       packages="${{inputs.packages}}"
-      verbose="${{inputs.verbose}}"
-      case "$verbose" in
+      VERBOSE="${{inputs.verbose}}"
+      case "$VERBOSE" in
         true|false) :;;
-        *) error "verbose must be 'true' or 'false'. found '$verbose'";;
+        *) error "VERBOSE must be 'true' or 'false'. found '$VERBOSE'";;
       esac
       [ -n "${files}${packages}" ] || error "No files or packages specified"
 
       export LANG=C
-      test_file() {
-        outf="$tmpd/out"
-        [ -e "$f" ] || { fail "$f: does not exist"; continue; }
-        [ -f "$f" ] || { fail "$f: not a file"; continue; }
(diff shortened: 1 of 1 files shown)


pipelines/test/ldd-check.yaml
```diff
@@ -62,17 +62,22 @@ pipeline:
         fail "$f: missing ${missing# }"
       }
 
+      is_dyn_obj() {
+        # ldd has this code already, let's not NIH it
+        # note: the actual ldd check happens in test_file()
+        ldd "$1" > /dev/null 2>&1
```
smoser, line 68: Does it by chance edit with a different code for "not a dynamic object" than for "catastrophic failures"

dannf: It does not. But we can check it's error message. See the updated commit.

pipelines/test/ldd-check.yaml
```diff
@@ -62,6 +62,19 @@ pipeline:
         fail "$f: missing ${missing# }"
```
smoser, line 62: huh. i think the above line here should be 'return 0', not 'continue' as we're not in a loop here.

smoser (changes requested):

smoser:
I pushed some changes. please @dannf and others review.

smoser (dismissed):

pipelines/test/ldd-check.yaml
```diff
@@ -38,51 +38,94 @@ pipeline:
...
       for f in "$@"; do
-        test_file "$f"
+        check_file "$f" true
       done
       set -- $packages
       for pkg in "$@"; do
         test_files_in "$pkg"
       done
-      echo "tested $((passes+fails)) files with ldd. $passes passes. $fails fails."
+      echo "tested $((passes+fails)) files with ldd." \
```
murraybd, line 129: I really wish this  line was prefaced with "[ldd-check]" like some of the other lines so one could easily grep log files for "[ldd-check]".

dannf: https://github.com/wolfi-dev/os/pull/39042/commits/69f73a83c353e185990e438850681eeaffa657f3

murraybd (approved):

PR #39058, opened 2025-01-08
remove unnecessary .pem files under examples folder

developer-guy:
Fixes: This PR helps erlang to avoid having unnecessary *.pem files that cause false-positives during image scanning process.

Related:

#### For new package PRs only

### DIFF
--- a/erlang-26.yaml
+++ b/erlang-26.yaml
@@ -1,7 +1,7 @@
 package:
   name: erlang-26
   version: 26.2.5.6
-  epoch: 0
+  epoch: 1
   description: General-purpose programming language and runtime environment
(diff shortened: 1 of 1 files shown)


mamccorm (dismissed):
If these are in the upstream image, we'll need to discuss whether we should be deviating and removing these.

Please don't merge and lets collab via the ticket

xnox:
> If these are in the upstream image, we'll need to discuss whether we should be deviating and removing these.
> 
> Please don't merge and lets collab via the ticket

which ticket?

dannf (approved):
This rationale makes sense to me. Thanks all!

PR #39086, opened 2025-01-08
DNM: test if helix builds without ca-certificates

kranurag7:
testing only


kranurag7:
closing given we've the validation that we need to be explicit about `ca-certificates-bundle`

PR #39113, opened 2025-01-08
openssf-compiler-options: remove explict dependecy

justinvreeland:
for all packages except gcc and clang


justinvreeland:
Fish shell macOS version of sed.
```
for i in *.yaml;
      sed -i '' '/\- openssf-compiler-options/d' $i
  end
```

dannf (commented):
Nice! Should we also keep it in yaml files that explicitly reference `openssf.spec`? That may be just `datadog-agent.yaml` atm. Or should we make those usages conditional on the spec file being there?

justinvreeland:
> Nice! Should we also keep it in yaml files that explicitly reference `openssf.spec`? That may be just `datadog-agent.yaml` atm. Or should we make those usages conditional on the spec file being there?

Good catch, yeah there should be an explicit dep there.

dannf:
Sweet, thanks! This LGTM - I can't see why this would cause any issues - but let me double check for other concerns in our team SU tomorrow morning.

dannf (dismissed):
we're +1 :) Thanks for the tech debt reduction. @murraybd suggested noting what this means in eng-psa, since people may have grown accustomed to seeing it

lean4.yaml
```diff
@@ -10,7 +10,6 @@ package:
     memory: 16Gi
   dependencies:
     runtime:
-      - openssf-compiler-options # this package picks up our wrapper
```
xnox, line 13: This line should remain.

xnox (changes requested):

xnox (dismissed):
+1 to force merge

xnox:
but conflicts need resolving:

```
Conflicting files
dbus.yaml
ggshield.yaml
systemd.yaml
```

xnox (approved):

justinvreeland:
> we're +1 :) Thanks for the tech debt reduction. @murraybd suggested noting what this means in eng-psa, since people may have grown accustomed to seeing it

Good suggestion, and done!

PR #39136, opened 2025-01-08
Add the ldd-check test pipeline to more packages

murraybd:
Some packages had the `ldd-check` test pipeline added more than once so this cleans that up, it also adds the `ldd-check` test pipeline to some additional packages.


murraybd:
And of course this all passed when I ran it on a workstation.

dannf (approved):

PR #39152, opened 2025-01-09
Implement Package Metadata ELF notes by including melange generated gcc specs for C, C++, Rust, Go.

xnox:
Cannot land this PR in one go, thus split openssf-compiler-options (first commit) into separate PR to land first:
- https://github.com/wolfi-dev/os/pull/39242

- **openssf-compiler-options: include melange spec files**
  If exist, include melange generated spec file from workspace.
  No errors produced if the melange generated spec file is missing.
  
  Potentially worth exploring to also always load $HOME/.gcc.spec, which
  may make it easier to customize gcc spec files on per-package basis.
  

- **Use new spec file with optional include**
  As an e2e test for three packages (written in C/Go/Rust). This enables
  using the future spec file with melange generated spec file include.
  
  When these land in images, and SBOM/Vulnerability/Console all look
  good we can roll this out by default.
  

NB! This PR needs melange that generates the dynamic gcc spec file which is being optionally included.

Once this lands for three packages in question, images are tested, and any bugs/issues resolved, we can remove the opt-in, and make the ELF notes the default.

PR #39242, opened 2025-01-10
openssf-compiler-options: include melange spec files

xnox:
This is part 1 / pre-requisite to land:
- https://github.com/wolfi-dev/os/pull/39152

If exist, include melange generated spec file from workspace.
No errors produced if the melange generated spec file is missing.

Potentially worth exploring to also always load $HOME/.gcc.spec, which
may make it easier to customize gcc spec files on per-package basis.

### DIFF
--- a/openssf-compiler-options.yaml
+++ b/openssf-compiler-options.yaml
@@ -1,7 +1,7 @@
 package:
   name: openssf-compiler-options
   version: 20240627
-  epoch: 10
+  epoch: 11
   description: "Compiler Options Hardening Guide for C and C++"
   url: https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.html
   copyright:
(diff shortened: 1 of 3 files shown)


dannf (approved):
Thanks @xnox ! +1 to get this off the ground, but a couple of things I'd like to discuss, more related to the melange side of things:

I was wondering why we chose a workspace instead of a system path, but I see you raised the same question in your melange commit message. The workspace feels too volatile - something that should be owned by the package YAML. I think it could make developers more likely to consider in-place sed hacks to it instead of using one of the existing override mechanisms. (I'm a guilty party, but I'm thinking we could implement a pipeline for cases like mine). While I prefer `/etc/melange`, would `/melange` would be more idiomatic in a container world? That would reserve `/etc/melange` for a melange-running-in-wolfi use case.

I feel like we should add some explicit messages in the logs when this is in use, especially during this startup period. While we can likely search for melange version numbers, a directly "grep-able" string might be easier to use for some quick bulk searches.

xnox:
> I was wondering why we chose a workspace instead of a system path, but I see you raised the same question in your melange commit message.

see https://github.com/chainguard-dev/melange/issues/1750 because all paths inside the container are writable; currently we cannot build apko image with arbitrary extra text file which didn't come from an apk; and everything is mutable - so we need to do other build isolation work with most container files own by root (not the builder accounts uid) to actually make files in / to be not writable by the builder process. Hence all of this is moot for now until we fix lots of the pre-requisites to make any container/host files immutable.

xnox:
> I feel like we should add some explicit messages in the logs when this is in use, especially during this startup period. While we can likely search for melange version numbers, a directly "grep-able" string might be easier to use for some quick bulk searches.

ok. We can land all of this things for the three packages; and then demo it and see how much more verbose we can make this. Because yes, I agree more verbosage would be nice.

Cranking up gcc debug output is possible => but that's so much spam, it will be impossible to read the build logs.

I guess melange emitting that it is writting this file would help.

xnox:
forked tracing to https://github.com/chainguard-dev/melange/issues/1752

PR #39245, opened 2025-01-10
websockify: multiversion support

sil2100:
Currently pulls in 3.12 because the `/usr/bin/websockify` script tries to use the python3.12 interpreter (build time defined). Move to multiversion + split off to -bin (with provides).

### DIFF
--- a/websockify.yaml
+++ b/websockify.yaml
@@ -1,28 +1,34 @@
 package:
   name: websockify
   version: 0.12.0
-  epoch: 2
+  epoch: 3
   description: WebSockets support for any application/server
   copyright:
     - license: LGPL-3.0-or-later
   dependencies:
-    runtime:
-      - numpy
-      - py3-jwcrypto
-      - py3-requests
-      - py3-simplejson
+    provider-priority: 0
+
+vars:
+  pypi-package: websockify
+  import: websockify
+
+data:
+  - name: py-versions
+    items:
+      3.10: '310'
+      3.11: '311'
+      3.12: '312'
+      3.13: '313'
 
 environment:
   contents:
     packages:
-      - autoconf
-      - automake
       - build-base
       - busybox
       - ca-certificates-bundle
-      - py3-gpep517
-      - py3-setuptools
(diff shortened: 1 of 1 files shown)


websockify.yaml
```diff
@@ -31,18 +37,55 @@ pipeline:
...
-  - uses: strip
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}-bin
+    description: Executable binaries for ${{vars.pypi-package}} installed for python${{range.key}}
+    dependencies:
+      runtime:
+        - py${{range.key}}-${{vars.pypi-package}}
+      provides:
+        - py3-${{vars.pypi-package}}-bin
+        - websockify
```
kranurag7, line 75:
```suggestion
        - py3-${{vars.pypi-package}}
```
I think we should be using this package now as runtime dependency after migrating to multiversion support. websockify is empty [package](https://apk.chaindag.dev/https/apk.cgr.dev/wolfi-presubmit/9125a36bf69507ddb282df7987f7d1e869aeb356/aarch64/websockify-0.12.0-r3.apk@etag:7621302313b3155aa50af2af463a2ea7/) via this commit.

sil2100: I added a provides for `websockify` because of all the potential images and packages that could depend on the old `websockify` package name. Is it not practice to provide the old names for backwards compatibility?

kranurag7: thank you for the explanation, this looks good to me. Can we also consider adding `py3-${{vars.pypi-package}}` together with this so that we have it synced with other examples and if someone wants to pull in the latest then they can do `apk add py3-websockify` or specify the same as runtime dep.

I think we also don't need to check on dependents `docker-selenium` and `novnc` to check if they're building correctly after this migration because they're already building with latest python which is inline with this. most likely they'll build fine. 💭

sil2100: Aaah, good idea with adding that one, doing it! Thank you!

kranurag7 (approved):
thank you for adding another provides, looks good.

PR #39303, opened 2025-01-10
Enable ARM branch protection in gcc + restore OpenSSF hardening in glibc

dannf:
Fixes: https://github.com/chainguard-dev/internal-dev/issues/7940


pipelines/test/compiler-hardening-check.yaml
```diff
@@ -34,7 +34,6 @@ pipeline:
       EOF
       ${{inputs.cc}} -v -o hello-default hello.c
       out=$(./hello-default 1 || true)
-      [ "$out" = "hello-c" ]
```
xnox, line 37: this checks that compiled thing is executable.

xnox:
AMAZING

xnox (approved):

PR #39406, opened 2025-01-13
chore(llvm): Build everything together

EyeCantCU:
* Refactors our existing LLVM packages to build the world together (LLVM, Clang, runtimes, etc)
* Adds runtimes we had with prior LLVM versions
* Introduces the Clang compiler runtime


xnox (dismissed):
Looks good. Note we might not need all the subpackages so separate. In general llvm dev stuff; cmake all can be together. We only possibly want separate runtime libraries for which there are things that shared linked against them. And possibly lld / clang top level binaries; but even that usually doesn't hurt to have the rest of development llvm stuff there.

llvm-19.yaml
```diff
@@ -47,75 +65,473 @@ pipeline:
...
+            nvptx-arch --version
+            nvptx-arch --help
+
+  - name: llvm-cmake-${{vars.major-version}}
+    description: CMake macros for LLVM ${{vars.major-version}}
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/${{vars.llvm-prefix}}/share/cmake
+          mv ./cmake/* ${{targets.contextdir}}/${{vars.llvm-prefix}}/share/cmake/
+
```
xnox, line 394: i thought these need to be generated and are templated and need to be installed, not just copied. or are they generated in the current working directory? looks interesting.

xnox (dismissed):
This is good. And I think we can merge this and work out any issues with follow ups.

EyeCantCU:
The update for lld slid in there :slightly_frowning_face:. I'll rebase again real quick

llvm-19.yaml
EyeCantCU: I'll double check that

xnox: with         -DCMAKE_INSTALL_PREFIX=${{vars.llvm-prefix}} \ -DCMAKE_PREFIX_PATH=${{vars.llvm-prefix}} now added i expect this package to not exists. cmake files to be shipped in the top level dev package. And it can have provides llvm-cmake-${{vars.major-version}} for compat. But we really don't need this subpackage.

I guess worth rebuilding whatever was using this - rust?!

EyeCantCU: There isn't a package that explicitly depends on llvm-cmake-19 so I've omitted in the latest push. The only package I found that has an explicit dependency on LLVMs CMake headers is qdrant, but it was pinned to an older version so I've bumped it to use the latest Clang

EyeCantCU:
The libraries for `libunwind` and `libcxxabi` are the same version in LLVM 17 and 19. The SCA generated dependency takes priority here:

```
  llvm-libunwind-16-16.0.6-r5:
    conflicts: llvm-libunwind-19-19.1.7-r1[llvm-libunwind=16]
    satisfies: llvm-libcxxabi-19-19.1.7-r1[so:libunwind.so.1]
               llvm-libcxx-19-19.1.7-r1[so:libunwind.so.1]
  llvm-libcxxabi-19-19.1.7-r1:
    conflicts: llvm-libcxxabi-17-17.0.6-r1[llvm-libcxxabi=19.1.7-r1]
    satisfies: llvm-libcxx-19-19.1.7-r1[llvm-libcxxabi-19]
  llvm-libunwind-19-19.1.7-r1:
    conflicts: llvm-libunwind-16-16.0.6-r5[llvm-libunwind=19.1.7-r1]
    satisfies: llvm-libcxx-19-19.1.7-r1[llvm-libunwind-19]
  llvm-libcxxabi-17-17.0.6-r1:
    conflicts: llvm-libcxxabi-19-19.1.7-r1[llvm-libcxxabi=17]
    satisfies: llvm-libcxx-19-19.1.7-r1[so:libc++abi.so.1]
```

Looking into how we can fix this, cc @xnox

xnox:
> The libraries for `libunwind` and `libcxxabi` are the same version in LLVM 17 and 19. The SCA generated dependency takes priority here:
> 
> ```
> ERROR: unable to select packages:
>   llvm-libunwind-16-16.0.6-r5:
>     conflicts: llvm-libunwind-19-19.1.7-r1[llvm-libunwind=16]
>     satisfies: llvm-libcxxabi-19-19.1.7-r1[so:libunwind.so.1]
>                llvm-libcxx-19-19.1.7-r1[so:libunwind.so.1]
>   llvm-libcxxabi-19-19.1.7-r1:
>     conflicts: llvm-libcxxabi-17-17.0.6-r1[llvm-libcxxabi=19.1.7-r1]
>     satisfies: llvm-libcxx-19-19.1.7-r1[llvm-libcxxabi-19]
>   llvm-libunwind-19-19.1.7-r1:
>     conflicts: llvm-libunwind-16-16.0.6-r5[llvm-libunwind=19.1.7-r1]
>     satisfies: llvm-libcxx-19-19.1.7-r1[llvm-libunwind-19]
>   llvm-libcxxabi-17-17.0.6-r1:
>     conflicts: llvm-libcxxabi-19-19.1.7-r1[llvm-libcxxabi=17]
>     satisfies: llvm-libcxx-19-19.1.7-r1[so:libc++abi.so.1]
> ```
> 
> Looking into how we can fix this, cc @xnox. Quick fix would be no-depends and explicit runtime depends, but that feels wrong

Provider priorities?
Also ensure the ABIs are separate?
Also are you sure these should be part of the llvm mega stack?!
I thought these libraries are similar to like gcc_s where it is public library and only one of them should be present.
Or ensure they are in private toolchain location in all, which would be preferred. And only single one in public location.

I guess get inspired by the differences in packaging between gcc.yaml and gcc-13.yaml.

EyeCantCU:
> Also ensure the ABIs are separate?

From what I can tell, they are not. That's a great point. I'll clean this up

> Also are you sure these should be part of the llvm mega stack?!

At least present in the latest packaging for LLVM

> I guess get inspired by the differences in packaging between gcc.yaml and gcc-13.yaml.

I'll look at both, thank you!

EyeCantCU:
@xnox would really appreciate another review here when available. We should be good to move forward with what's here. I'll be filing a ticket and picking up remaining work (namely eliminating virtual provides in llvm after migrating packages that pin to older llvm versions to new llvm, and adding omp to the stack)

xnox:
@zmarano @a-crate @justinvreeland @dannf I trust in you all!

dannf (commented):
Thanks for taking this on @EyeCantCU ! This is clearly a lot of work, and would therefore take some time to review in detail, but here's some feedback:

Is your goal for the new packages to be drop in replacements for the previous, warts & all, with the exception of the things noted in the description (clang-rt + reintroduction of other runtimes), or are other changes expected? Upon spot checking, I noticed that the refactored `llvm15` installs 85 files in `/usr/bin`, whereas the older one installed 0. That will of course introduce a potential for conflicts. What differences do you expect to see?

Can you talk about what testing you've done to verify these changes? For example, did you try rebuilding everything that uses clang/llvm in the various repos and check for ftbfs regressions?

I'm not sure I understand the current-and-previous-major-version provides, I'm not sure I've seen that pattern before. I saw your thread about the SCA-added dependencies - e.g. `llvm-libunwind-16`. Could we rebuild the packages with those SCA-added dependencies w/ the new packaging in the same PR to get rid of that? I'm thinking about a request we have to add llvm-14 - to keep this pattern, we'd need to remember to bump every other llvm-* as well.

I wonder if we shouldn't name the library packages with the soname - e.g. llvm-unwind*1* for `llvm-unwind.so.1`. All llvm versions that use libunwind1 should be able to share the newest common version of that (>= ${{package.full-version}}) but, should an llvm 25 bump to .so.2, we could co-install llvm-unwind1~=24 with llvm-unwind2~=25.

I'm a bit nervous about withdrawing toolchain components - is that required here?. Those would seem to have forensic value, and give us a place to rollback to if we something goes wrong and we need to roll *this* back.

Finally, is it possible to do one clang/llvm version at a time instead of all at once?

EyeCantCU:
Thanks for taking a look Dann!

> Is your goal for the new packages to be drop in replacements for the previous, warts & all, with the exception of the things noted in the description (clang-rt + reintroduction of other runtimes), or are other changes expected? Upon spot checking, I noticed that the refactored `llvm15` installs 85 files in `/usr/bin`, whereas the older one installed 0. That will of course introduce a potential for conflicts. What differences do you expect to see?

Exactly, I don't want to break any packages so I've provided virtual provides for the packages I've refactored that now provide those components.

The refactored LLVM 15 contains symlinks to /usr/bin that used to be provided by a separate package that only contained symlinks. We provide that package as a virtual provides now with LLVM. So the composition of packages did change, but packages have virtual provides for those to keep up. We changed that composition in newer versions of LLVM/Clang we've packaged so I modified things to align there

> Can you talk about what testing you've done to verify these changes? For example, did you try rebuilding everything that uses clang/llvm in the various repos and check for ftbfs regressions?

Most packages across each version of LLVM in Wolfi/ent/extras. The number of packages that use LLVM right now is actually far fewer than I had thought. The only packages that fail to build are older versions of rust that we've bootstrapped. But those also fail in the same way with our current packages. I'm pretty sure we didn't intend to keep the manifests for older versions around anyway

> I'm not sure I understand the current-and-previous-major-version provides, I'm not sure I've seen that pattern before. I saw your thread about the SCA-added dependencies - e.g. `llvm-libunwind-16`. Could we rebuild the packages with those SCA-added dependencies w/ the new packaging in the same PR to get rid of that? I'm thinking about a request we have to add llvm-14 - to keep this pattern, we'd need to remember to bump every other llvm-* as well.

I can refactor existing packages in this PR, but I'd have to follow up with ent/extras after anyway. I think it makes sense to keep them separate. The idea was to land this, fixup packages that live in wolfi/ent/extras to use the current packages and attempt to move older packages to more recent LLVM in the process, and then kill virtual provides to prevent breaking the world. I wish I could do this in one sweep, but this isn't a monorepo

> I wonder if we shouldn't name the library packages with the soname - e.g. llvm-unwind_1_ for `llvm-unwind.so.1`. All llvm versions that use libunwind1 should be able to share the newest common version of that (>= ${{package.full-version}}) but, should an llvm 25 bump to .so.2, we could co-install llvm-unwind1~=24 with llvm-unwind2~=25.

Very good point! I like this! I'll make the change

> I'm a bit nervous about withdrawing toolchain components - is that required here?. Those would seem to have forensic value, and give us a place to rollback to if we something goes wrong and we need to roll _this_ back.

I'll set the provider priority for the new packages to avoid a withdraw, though that's a tad ugly to work around something broken historically

I also promise to own breakages here

> Finally, is it possible to do one clang/llvm version at a time instead of all at once?

We'd have to land the latest changes first, but it should be. I'll narrow the scope of this PR

javacruft:
Completed under a different PR this cycle - closing.

PR #39413, opened 2025-01-13
chrony: fix invalid configs, provide cloud configs

xnox:
Current packaging shipped unused, unusable, broken or depricated
config files. Clean all of them up.

Instead provide systemd units, and split out configs.

Provide config packages with preffered configuration for Precision
Time Protocol in public clouds. Also provide sources for in public
cloud Network Time Protocol fallback, as well as externally accessible
Network Time Protocol.

All of the above use leap second smearing.


89luca89 (approved):

89luca89:
:+1:

PR #39415, opened 2025-01-13
mysql-9.1: create default runtime directories during build

OddBloke:
`mysqld` won't start in the default configuration which we ship without these directories being present; this commit avoids users having to discover the issue and create them themselves.

To validate this change, this commit also updates the test pipeline to run with the default value for --datadir (which is the same as what we were passing anyway), and not to create the directories which the package now ships.

### DIFF
--- a/mysql-9.1.yaml
+++ b/mysql-9.1.yaml
@@ -1,7 +1,7 @@
 package:
   name: mysql-9.1
   version: 9.1.0
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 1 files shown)


jamonation:
I'd also like to see how this behaves when used to build an image - e.g. is this expected to be a no-op in terms of changes to existing images that rely on this package? I'll try getting a local build going here and report back.

jamonation (approved):
Tested this out as an image and it works 💯, thanks for the refactor work!

PR #39422, opened 2025-01-13
openjdk: Use JRE build for JRE subpackage

pnasrat:
Adds a new java-common-jre subpackage to reduce dangling symlinks if only JRE present

This removes some unnecessary java modules installed when just using a JRE


pnasrat (commented):
LGTM

Need another reviewer

jvanzyl (approved):

PR #39462, opened 2025-01-13
fix(world): Correctly use default-jvm and default-jdk

EyeCantCU:
default-jvm is meant for exclusive use for the JRE and default-jdk is for exclusive use with the JDK

Only rebuild packages where runtime dependencies have changed

### DIFF
--- a/R.yaml
+++ b/R.yaml
@@ -65,8 +65,7 @@ environment:
       - libxmu-dev
       - make
       - openblas-dev
-      - openjdk-17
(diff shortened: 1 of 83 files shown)


pnasrat (approved):
LGTM

Thanks for taking on this clean up and I agree with only bumping runtime changes here

As `default-jvm` and `default-jdk` currently setup a symlink to the same location this is currently a no-op for build deps where the prior requirement was both the jdk package and `default-jvm`

PR #39524, opened 2025-01-14
fix(sbt): Drop default-jdk

EyeCantCU:
Depend directly on OpenJDK 11 and allow other variants to be installed


xnox:
Verified that sbt images plan and build with this change in place.

xnox (approved):

PR #39558, opened 2025-01-14
fix(openjdk-17-jre): Use correct Java home

EyeCantCU:
This was pointing to OpenJDK 22, correctly point to 17


xnox (approved):

EyeCantCU:
Going with: https://github.com/wolfi-dev/os/pull/39559

PR #39589, opened 2025-01-15
Set manual update flag, unfortunately we can't enable auto-updates fo…

mamccorm:
release-monitor project returns v18:
 - https://release-monitoring.org/project/13891/
 
 But there is no stable v18, so the download link will fail:
 
```bash
ERRO chrpath: package chrpath: failed to validate update config, melange bump: got 404 Not Found when fetching https://alioth-archive.debian.org/releases/chrpath/chrpath/0.18/chrpath-0.18.tar.gz
```

Latest stable is v16. v18 is shown to be unstable. We can't use git poller here as the source is not hosted on a SCM. I couldn't see a better way to handle this in a new release-monitor project either.

Setting manual update flag - it'll appear on our list of manual packages to review.

### DIFF
--- a/chrpath.yaml
+++ b/chrpath.yaml
@@ -35,7 +35,8 @@ subpackages:
       - uses: split/manpages
 
 update:
-  enabled: true
+  enabled: false
+  exclude-reason: v16 is the latest stable at time of writing. v18 is unstable and returned by release-monitor project. Code is not in a SCM for us to use git poller.
   release-monitor:
     identifier: 13891
 


debasishbsws (commented):
Instead of update enabled false should we set the update manual = true ?

mamccorm:
> Instead of update enabled false should we set the update manual = true ?

Unfortunately not @debasishbsws , it's failing the update checks, re: https://github.com/chainguard-dev/internal-dev/issues/4914.

If you see another path forward open to suggestions, we're not regressing by this at the minute as the update check fails, but due to the conflicting data returned and lack of ability to use git and git poller, this looks like the only solution

debasishbsws (approved):
Ahh i get it now.

PR #39592, opened 2025-01-15
Update Git checkout URL

mamccorm:
This package currently leverages release-monitor and fetch to pull down the source code. The URL has the year hard-coded, which needs to be updated.

Longer term we might consider moving this to a git clone, but the reasons for not doing so originally would need to be understood. It's sometimes intentional for larger projects.

Additionally, upgrades to latest package version.

### DIFF
--- a/sqlite.yaml
+++ b/sqlite.yaml
@@ -28,7 +28,7 @@ var-transforms:
 pipeline:
   - uses: fetch
     with:
-      uri: https://www.sqlite.org/2024/sqlite-autoconf-${{vars.sqlite-download-version}}.tar.gz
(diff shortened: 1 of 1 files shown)


kranurag7:
thank you for the changes, this looks good. 

> Longer term we might consider moving this to a git clone, but the reasons for not doing so originally would need to be understood.

the reason for not proceeding ahead with git-checkout in the past was that the project uses fossil as choice of VCS (https://sqlite.org/src/doc/trunk/README.md) 
Having said that, if we want we can use https://github.com/sqlite/sqlite which is called Official mirror of the project and seems largely upto date with sqlite upstream. I'll leave this decision upto you if we want to rely on a GitHub mirror here.

philroche (approved):

PR #39639, opened 2025-01-15
kubeflow-pipelines-visualization-server: bump to use Python 3.11

sil2100:
Another package that was causing some images to pull in multiple Python bases.

This only bumps the used Python version to 3.11 as this is the max that google-cloud-sdk seems to support for now (see google-cloud-sdk.yaml). This also required bumping some pip dependencies.

### DIFF
--- a/kubeflow-pipelines-visualization-server.yaml
+++ b/kubeflow-pipelines-visualization-server.yaml
@@ -9,18 +9,21 @@ package:
     - x86_64 # Since tensorflow-data-validation is not available for aaarch64 yet
   dependencies:
     runtime:
-      - py3.11-google-cloud-sdk
-      - py3.11-typing-extensions
-      - python-3.11
+      - py${{vars.py-version}}-google-cloud-sdk
+      - py${{vars.py-version}}-typing-extensions
+      - python-${{vars.py-version}}
+
+vars:
+  py-version: 3.11
 
 environment:
   contents:
     packages:
       - build-base
       - ca-certificates-bundle
-      - py3.11-build-base-dev
-      - py3.11-crcmod
-      - py3.11-pip
+      - py${{vars.py-version}}-build-base-dev
+      - py${{vars.py-version}}-crcmod
+      - py${{vars.py-version}}-pip
 
 pipeline:
   - uses: git-checkout
@@ -42,9 +45,6 @@ pipeline:
 
       pip install --root=${{targets.destdir}} --prefix=/usr -r requirements.txt
       find ${{targets.destdir}} -name "*.pyc" -exec rm -rf '{}' +
-
-      # "collections" is deprecated in Python 3.10+, use "collections.abc" instead
-      sed -i 's/collections/collections.abc/g' ${{targets.destdir}}/usr/lib/python3.11/site-packages/gcsfs/mapping.py
 
       # CVE-2023-47248, GHSA-5wvp-7f3h-6wmm
       pip install pyarrow-hotfix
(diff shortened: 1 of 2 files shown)


sil2100:
Also, wanted to bump the tensorflow requirement to what we use, aka. 2.18.0, but actually one of the pip dependencies, I guess that was even tensorflow-data-validation 1.16.1 (latest), claimed it only supported tensorflow <2.17.0. So my bump in the requirements.txt had to be to 2.16.2 instead.

kubeflow-pipelines-visualization-server.yaml
```diff
@@ -1,27 +1,26 @@
...
   description: Machine Learning Pipelines for Kubeflow
   copyright:
     - license: Apache-2.0
   target-architecture:
     - x86_64 # Since tensorflow-data-validation is not available for aaarch64 yet
   dependencies:
     runtime:
-      - py3.10-google-cloud-sdk
-      - python-3.10
+      - py3.11-google-cloud-sdk
```
dannf, line 12: consider defining and using `vars.py-version`

kubeflow-pipelines-visualization-server.yaml
```diff
@@ -45,8 +43,8 @@ pipeline:
       pip install --root=${{targets.destdir}} --prefix=/usr -r requirements.txt
       find ${{targets.destdir}} -name "*.pyc" -exec rm -rf '{}' +
 
-      # "collections" is deprecated in Python 3.10, use "collections.abc" instead
-      sed -i 's/collections/collections.abc/g' ${{targets.destdir}}/usr/lib/python3.10/site-packages/gcsfs/mapping.py
+      # "collections" is deprecated in Python 3.10+, use "collections.abc" instead
+      sed -i 's/collections/collections.abc/g' ${{targets.destdir}}/usr/lib/python3.11/site-packages/gcsfs/mapping.py
```
dannf, line 47: do we need this anymore?

```
2ffbaddaf3fd:/work/packages# grep collections /usr/lib/python3.11/site-packages/gcsfs/mapping.py 
2ffbaddaf3fd:/work/packages# 
```

dannf (commented):
Wow, this one seems like a difficult case - thanks for tackling it!

I'm concerned with diverging from upstream to move to 3.11 support. I wondered why upstream hadn't moved forward, and took a quick look at the project. It seems very active. But where there used to be both `2.3.0` and `sdk-2.3.0` tags, there's now been 9 releases of just `sdk-*` tags (2.4->2.11). I wonder if maybe they have restructured things somewhat, and we're stuck on a dead release.

As for how to validate these changes, the best way I can think of is to build them into the target image and run the image test case.

sil2100:
Let me try building the image against the PR presubmit repo and see if that works as expected.

sil2100:
> I'm concerned with diverging from upstream to move to 3.11 support. I wondered why upstream hadn't moved forward, and took a quick look at the project. It seems very active. But where there used to be both `2.3.0` and `sdk-2.3.0` tags, there's now been 9 releases of just `sdk-*` tags (2.4->2.11). I wonder if maybe they have restructured things somewhat, and we're stuck on a dead release.

Regarding this: I don't think we're necessarily 'diverging' from upstream by moving from 3.10 to 3.11. I didn't dig down 
too deeply of course, but from all the release notes that I saw, there doesn't seem to be a recommended Python version to use. I think we used 3.10 because it was probably what google-cloud-sdk was using before. The only requirement that pipelines has is for Python to be 3.7 or above.
Did you see any mention of 3.10 being recommended that I missed?

sil2100:
hm, for unknown reasons I can't get the relevant image to build - both with the old and new package. It just gets stuck on "Creating...". Will investigate.

OddBloke (approved):

PR #39643, opened 2025-01-15
Add Spark 3.5 enhancements including dependency updates and script fixes

utieyin:
Refactored the spark-3.5-scala-2.12 package to allow the building of spark-3.5 with both scala-2.12 the default and scala-2.13. Files needed for patches and pombumps have also been copied across in the new spark-3.5


mamccorm (dismissed):
Thanks for this - making sure we don't merge this today (Friday). Defering to next week

wojciechka (approved):
LGTM, but no spark/scala expert

jamonation:
Looks good here - only question I have is when 4.x is released, how will this package fare? I'd hope it fails to build since it will need a refactor?

jamie-albert (approved):

PR #39684, opened 2025-01-15
Adds it-tools

srishtih:
Fixes:

Related:

#### For new package PRs only

#### For version bump PRs

### DIFF
--- a/it-tools.yaml
+++ b/it-tools.yaml
@@ -1,7 +1,7 @@
 package:
   name: it-tools
   version: 2024.10.22
-  epoch: 1
+  epoch: 0
   description: Collection of handy online tools for developers, with great UX
   copyright:
     - license: GPL-3.0
@@ -9,6 +9,7 @@ package:
     no-depends: true
   dependencies:
     runtime:
+      - ${{package.name}}-nginx-config
       - nginx
 
 environment:
@@ -19,6 +20,7 @@ environment:
       - ca-certificates-bundle
       - git
       - go
+      - nginx-package-config
       - node-gyp
       - nodejs
       - npm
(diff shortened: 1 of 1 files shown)


it-tools.yaml
```diff
@@ -0,0 +1,85 @@
...
+      - npm
+      - pnpm
+  environment:
+    NPM_CONFIG_LOGLEVEL: "warn"
+
+pipeline:
+  - runs: |
+      git clone --depth=1 https://github.com/CorentinTh/it-tools.git
+      cd it-tools
+      # rmdir it-tools
```
imjasonh, line 34: Is this needed?

Can you document why this isn't a `git-checkout` ?

srishtih: The version includes a commit from one of the pull requests that contribute to the new version. I can add that as a note here

it-tools.yaml
```diff
@@ -0,0 +1,85 @@
...
+pipeline:
+  - runs: |
+      # The version includes a commit from one of the pull requests that contribute to the new version
+      git clone --depth=1 https://github.com/CorentinTh/it-tools.git
+      cd it-tools
+      package_version="${{package.version}}"
+      git fetch --tags
+      latest_tag=$(git tag -l "v${package_version}-*" | sort -V | tail -n 1)
+
+      git checkout ${latest_tag}
```
Dentrax, line 39: Should we create a new ReleaseMonitor and use git-checkout` pipeline here instead? That way we able to access latest tag.

it-tools.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+  - runs: |
+      cd it-tools
+      npm install -g pnpm
```
Dentrax, line 43: Do we still need this since we already have `pnpm` in the build-time deps.

EyeCantCU (approved):

PR #39689, opened 2025-01-15
fix: Stream OTel plugin with ingress nginx

EyeCantCU:
The plugin needs to be built with the version set to the exact version of nginx shipped with the controller. Get rid of the standalone package and build inline with ingress nginx


ingress-nginx-controller-1.12.yaml
```diff
@@ -535,6 +546,33 @@ subpackages:
           modroot: ./images/kube-webhook-certgen/rootfs
           packages: .
 
+  - name: ingress-nginx-opentelemetry-plugin-${{vars.nginx-ingress-major-minor}}
+    description: OTel plugin for ingress nginx controller ${{vars.nginx-ingress-major-minor}}
+    dependencies:
+      provides:
```
Dentrax, line 552: Should this also provide `opentelemetry-plugin-nginx=${{package.full-version}}`?

Dentrax (approved):

ingress-nginx-controller-1.12.yaml
EyeCantCU: I'm going to address on the image side, for now it'll pull the current package

PR #39696, opened 2025-01-16
Switch some packages from using the ldd-check with specific files to the package

murraybd:
Only checking specific files with the ldd-check test pipeline is not sustainable as its possible for packages to build new files worth checking subsequently it would be more efficient to test the whole package.

Additionally, this PR splits out expat's docs to a separate package.


dannf (approved):
nice!

kranurag7 (approved):
thank you, this looks good. Merging given package update config issue looks independent of this change.

PR #39732, opened 2025-01-16
dbus: add systemd build depends

xnox:
This activates upstream build system to install systemd related units
and configuration files, such that dbus can work in a system
container.


89luca89 (approved):

sil2100 (approved):
LGTM.

PR #39733, opened 2025-01-16
add runtime deps to aspnet-runtime

developer-guy:
There were a failure in the aspnet-runtime due to missing runtime packages:

```bash
fail: Microsoft.AspNetCore.Server.Kestrel[13]                                                                                                                         
      Connection id "0HN9KCHVRM8HI", Request id "0HN9KCHVRM8HI:00000001": An unhandled exception was thrown by the application.       
      System.TypeInitializationException: The type initializer for 'SkiaSharp.SKData' threw an exception.
       ---> System.DllNotFoundException: Unable to load shared library 'libSkiaSharp' or one of its dependencies. In order to help diagnose loading problems, consider
 using a tool like strace. If you're using glibc, consider setting the LD_DEBUG environment variable: 
      libexpat.so.1: cannot open shared object file: No such file or directory  
```

so this PR aims to fix it.

### DIFF
--- a/dotnet-9.yaml
+++ b/dotnet-9.yaml
@@ -1,7 +1,7 @@
 package:
   name: dotnet-9
   version: 9.0.101
-  epoch: 1
+  epoch: 2
   description: ".NET SDK"
   copyright:
     - license: MIT
@@ -106,6 +106,8 @@ subpackages:
     dependencies:
       runtime:
         - dotnet-9-runtime
+        - libexpat1
+        - libbrotlidec1
 
   - name: aspnet-9-runtime-default
     dependencies:
--- a/dotnet-8.yaml
+++ b/dotnet-8.yaml
@@ -1,7 +1,7 @@
 package:
   name: dotnet-8
   version: 8.0.11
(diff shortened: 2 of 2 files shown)


dotnet-9.yaml
```diff
@@ -106,6 +106,8 @@ subpackages:
     dependencies:
       runtime:
         - dotnet-9-runtime
+        - libexpat1
+        - libbrotlidec1
```
sil2100, line 110: This seems correct. I remember `libbrotlidec1` as a dependency even in Ubuntu. Expat is something new, but the error seems to clearly indicate it is needed.

One question however: why is it a runtime dep here in `aspnet-9-runtime` and not `dotnet-9-runtime`?

sil2100 (commented):
This at least fits similar to what I've seen during my Ubuntu times. However, an inline question:

sil2100 (approved):
Let's re-visit the idea of moving those deps to dotnet-*-runtime some other time. This here is correct either way. Let's proceed.

PR #39742, opened 2025-01-16
package: netcdf

mesaglio:
This package is needed to build `vtk`, and `vtk` is needed to build `opencv`.

### DIFF
--- a/netcdf.yaml
+++ b/netcdf.yaml
@@ -1,0 +1,80 @@
+package:
+  name: netcdf
+  version: 4.9.2
+  epoch: 0
+  description: Libraries for the Unidata network Common Data Format
+  copyright:
+    - license: BSD-3-Clause
+
+environment:
+  contents:
+    packages:
+      - bash
+      - build-base
+      - busybox
+      - bzip2-dev
+      - cmake
+      - curl-dev
+      - hdf5-dev
+      - libaec-dev
+      - libxml2-dev
+      - m4
+      - samurai
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/Unidata/netcdf-c
+      tag: v${{package.version}}
+      expected-commit: 9328ba17cb53f13a63707547c94f4715243dafdf
+
+  - uses: patch
+    with:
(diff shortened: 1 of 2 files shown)


netcdf.yaml
```diff
@@ -0,0 +1,80 @@
...
+    with:
+      repository: https://github.com/Unidata/netcdf-c
+      tag: v${{package.version}}
+      expected-commit: 9328ba17cb53f13a63707547c94f4715243dafdf
+
+  - uses: patch
+    with:
+      patches: gcc14.patch
+
+  - runs: |
```
kranurag7, line 35: can we consider using the pipeline here? something like following:
```yaml
  - uses: cmake/configure
    with:
      opts: |
        -DCMAKE_BUILD_TYPE=None \
        -DENABLE_PERL=OFF \
        -DENABLE_TESTS=OFF \
        -DWITH_PCRE2=True
```

afterwards, we can use `cmake/build` and `cmake/install` command. 
You can refer to openscap melange config (ref:https://github.com/wolfi-dev/os/blob/main/openscap.yaml) for reference.

kranurag7 (changes requested):

netcdf.yaml
mesaglio: @kranurag7 Done 👍

netcdf.yaml
```diff
@@ -0,0 +1,80 @@
...
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/Unidata/netcdf-c
+      tag: v${{package.version}}
+      expected-commit: 9328ba17cb53f13a63707547c94f4715243dafdf
+
+  - uses: patch
+    with:
+      patches: gcc14.patch
```
kranurag7, line 33: I don't see this patch associated with this PR. 💭 Is this required for building this package?

kranurag7 (changes requested):
thank you, this looks good. Can we consider adding some tests for the package? We can use `test/pkgconf` pipeline for `-dev` package and `test/ldd-check` as well for the main package.

PR #39764, opened 2025-01-16
fix(yara-x): build+add primary yr cli tool to yara-x, drop compat package, use distro cargo-c

stevebeattie:


yara-x.yaml
```diff
@@ -29,37 +29,12 @@ pipeline:
 
   - runs: cargo install cargo-c --features=vendored-openssl --root $HOME/.cargo --target-dir $HOME/.cargo
```
kranurag7, line 30: non-blocking comment: I did a local build after removing `--features=vendored-openssl` and it seems to be working, I understand this is build-time only and this is not ending up in package. It would be great if we avoid any reliance of `vendored-openssl` 💭

stevebeattie: Yeah, I was curious if we could switch over to using the cargo-c apk instead of building cargo-c in the package itself.

stevebeattie: I converted over to depending on the cargo-c wolfi package rather than building cargo-c  from scratch, passes tests locally.

yara-x.yaml
```diff
@@ -86,3 +60,6 @@ test:
         EOF
 
         gcc `pkg-config --cflags yara_x_capi` `pkg-config --libs yara_x_capi` test.c
+    - runs: |
+        yr --help
+        yr --version
```
kranurag7, line 65:
```suggestion
        yr --version | grep ${{package.version}}
```

kranurag7 (approved):
Thank you Steve!! this looks good to me.

egibs (approved):
Amazing! 🚀

PR #39830, opened 2025-01-17
openjdk 24: split JRE and mark as beta

pnasrat:
Withdraw non beta versioned packages


openjdk-24.yaml
```diff
@@ -110,6 +118,10 @@ pipeline:
       cp -r build/*-server-release/images/jdk/* "${{targets.destdir}}"/$_java_home
       rm "${{targets.destdir}}"/$_java_home/lib/src.zip
 
+      # symlink to shared java cacerts store
+      rm -f "${{targets.contextdir}}"/$_java_home/lib/security/cacerts
+      ln -sf /etc/ssl/certs/java/cacerts \
```
EyeCantCU, line 123: Missing the destination

EyeCantCU (changes requested):

openjdk-24.yaml
pnasrat: doh ENOCOFFEE fixing

pnasrat:
@EyeCantCU fixed

EyeCantCU (dismissed):

imjasonh (approved):

openjdk-24.yaml
```diff
@@ -227,7 +202,7 @@ subpackages:
     description: "Use the openjdk-24 JVM as the default JVM with the JDK installed"
     dependencies:
       runtime:
-        - ${{package.name}}-default-jvm
+        - java-common
         - ${{package.name}}
```
stevebeattie, line 206: We create the `default-jvm` in the openjdk-24-default-jvm package, do we want to create the same here, like we did in the openjdk-23 and earlier packages?

```yaml
    pipeline:
      - runs: |
          mkdir -p "${{targets.subpkgdir}}"/usr/lib/jvm
          ln -sf java-24-openjdk "${{targets.subpkgdir}}"/usr/lib/jvm/default-jvm
```

stevebeattie: Addressed in #39878

PR #39831, opened 2025-01-17
Update Makefile

mfortin:
Makefile generate-index is currently not set to false as intented. This PR resolves this and disable generate-index.

Fixes:
Makefile `--generate-index false`

Related:
None
### Pre-review Checklist

- [X] None of the above


mfortin:
If you want to push a package to your own package repo, you may not want to have the index generated/updated by melange.

xnox:
> If you want to push a package to your own package repo, you may not want to have the index generated/updated by melange.

yes, but the way currently all developers work is by doing `wolfictl bump foo; make package/foo; make test/foo` and the latter relies on the default behaviour of generating and signing the index. As otherwise the local build will not be tested at all or used for local image builds.

I am afraid that this behaviour has been the expected default for too long.

if you want to disable local index generation, please export MELANGE_EXTRA_OPTS with a joint up = sign to override the setting (it can also be passed as make variable at the end of the make command).

dannf:
https://github.com/wolfi-dev/os/pull/50973

PR #39879, opened 2025-01-17
fix(logstash-8): build issue and missing deps - bump to 8.17.1

Dentrax:
Fixes build issue and adds some missing deps with `jdk` to fix: `WARNING: Unknown module: jdk.compiler specified to --add-exports`.


logstash-8.yaml
```diff
@@ -368,9 +384,9 @@ subpackages:
       environment:
         contents:
           packages:
-            - wolfi-base
+            - openjdk-17
         environment:
-          LS_JAVA_HOME: /usr/lib/jvm/default-jvm
```
EyeCantCU, line 373: `default-jvm` is still valid and should be preferred unless multiple JDKs are used

logstash-8.yaml
```diff
@@ -382,6 +398,12 @@ subpackages:
               grep 'message.*hello'
 
 test:
+  environment:
+    contents:
+      packages:
+        - openjdk-17
```
EyeCantCU, line 404: We could add `-default-jdk` here and still reference `-default-jvm` as the home

logstash-8.yaml
```diff
@@ -28,7 +28,8 @@ package:
     runtime:
       - bash # some helper scripts use bash and busybox utilities
       - busybox
-      - openjdk-17-default-jvm
+      - openjdk-17
+      - openjdk-17-default-jdk
```
EyeCantCU, line 32: This is fine, and will work, but `-default-jdk` pulls in `-openjdk-17` anyway so we don't need to include it twice

logstash-8.yaml
```diff
@@ -43,16 +44,24 @@ environment:
...
+      - glibc-locales
       - gradle
+      - grep
       - jq
       - jruby-9.4
       - logstash-filter-xml
       - logstash-integration-jdbc
       - logstash-output-opensearch
+      - net-tools
       - openjdk-17-default-jdk
```
EyeCantCU, line 61: In the future, we should use a toplevel variables, `java-version` and re-use it

EyeCantCU (dismissed):

EyeCantCU (approved):

PR #39889, opened 2025-01-17
Add flyway symlinks

mjramirez-dev:
Add symlinks to match `/flyway` and `/opt/java/openjdk` dirs and drop unneeded deps from `/flyway/lib/aad` and `/flyway/drivers/gcp` to match upstream flyway-11.2.0.

### DIFF
--- a/flyway.yaml
+++ b/flyway.yaml
@@ -1,7 +1,7 @@
 package:
   name: flyway
   version: "11.2.0"
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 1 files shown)


EyeCantCU (approved):

flyway.yaml
EyeCantCU, line 1: Needs epoch bump

EyeCantCU (changes requested):

flyway.yaml
mjramirez-dev: bumped

EyeCantCU (approved):

PR #39897, opened 2025-01-17
batch 7: go epoch bump

imjasonh:


ajayk (dismissed):

ajayk (approved):

stevebeattie:
Malcontent was just triggering on itself, need to dig into why it's not honoring IgnoreSelf.

imjasonh:
Screenshot 2025-01-17 at 9 06 32 PM

All the pods seem to have passed, but the watcher job seemed to be having trouble noticing that, and was hung. Rebased to retry.

PR #39902, opened 2025-01-17
ruby3.x-rails-8.0: new packages

MickeyPvX:
Creates Ruby on Rails package for Ruby 3.4, 3.3, and 3.2

Related: https://github.com/chainguard-dev/image-requests/issues/5160
EOL: https://endoflife.date/rails

#### For new package PRs only


imjasonh (dismissed):

ruby3.3-rails-8.0.yaml
```diff
@@ -0,0 +1,157 @@
...
+  copyright:
+    - license: MIT
+  dependencies:
+    provides:
+      - ruby${{vars.rubyMM}}-rails=${{package.full-version}}
+    runtime:
+      - gcc
+      - git
+      - glibc-dev
+      - make
```
jamonation, line 15: These shouldn't be runtime depends should they?

ruby3.3-rails-8.0.yaml
```diff
@@ -0,0 +1,157 @@
...
+      - ruby${{vars.rubyMM}}-bundler
+      - ruby-${{vars.rubyMM}}-dev
```
jamonation, line 17: Shouldn't be needed for runtime?

jamonation:
I think the runtime and build depends are mixed up here. Runtime shouldn't need much more than `ruby-<version>` and `tzdata`, whereas building needs all the dev packages.

ruby3.3-rails-8.0.yaml
MickeyPvX: I commented them out locally just to triple-check and rebuilt it...I could've sworn it was failing without them earlier, but it works without them now so...I'll drop those off 🤷‍♂️

MickeyPvX: Ok - new day, new testing after the CI tests failed (missing runtime dep `git`) - `rails` does in fact need `gcc` `ruby-dev` and some others in order for the rails server to start.

jamonation: Rails is gonna rails 🤷

jamonation (approved):
Nice tests!

PR #39912, opened 2025-01-18
Add liblmdb.so and liblmdb.so.0 symlinks

murraybd:
This is part of resolving https://github.com/chainguard-dev/customer-issues/issues/1975 where installing samba-client complains that liblmdb.so.0 is not a symlink:

```
# apk -q update && apk add -q samba-client
/sbin/ldconfig: /usr/lib/liblmdb.so.0 is not a symbolic link
```

### DIFF
--- a/lmdb.yaml
+++ b/lmdb.yaml
@@ -1,7 +1,7 @@
 package:
   name: lmdb
   version: 0.9.31
-  epoch: 2
+  epoch: 3
   description: Lightning Memory-Mapped Database
   copyright:
     - license: OLDAP-2.8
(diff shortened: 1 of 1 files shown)


sil2100 (approved):
Thanks Brian! So I think the **proper** fix for this would be to either modify the `0001-Debian-quirks-to-make-shared-library-versioned.patch` patch or add another one next to it - because the patch is IMO incomplete. This patch adds the multi-versioned-so bits, but forgets to touch the `make install` target to not follow symlinks, resulting in this issue.

However, since it's part of an escalation and basically adding a new patch is as invasive as just doing it via melange - let me merge this.

PR #39914, opened 2025-01-18
knative-serving package build

developer-guy:
Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/knative-serving-1.16.yaml
+++ b/knative-serving-1.16.yaml
@@ -1,0 +1,99 @@
+package:
+  name: knative-serving-1.16
+  version: 1.16.0
+  epoch: 0
+  description: Kubernetes-based, scale-to-zero, request-driven compute
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    provides:
+      - knative-serving=${{package.full-version}}
+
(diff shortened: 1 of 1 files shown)


kranurag7 (changes requested):
great work!!
We need a symlink to match the entrypoint at `/ko-app/queue` and `tzdata` at runtime for `${{package.name}}-queue` subpackage.

kranurag7 (approved):
thank you!! Looks good.

kranurag7 (changes requested):
knative supports version streams so we'll have to rename the file and package here with `foo-x.y` format.

knative-serving-1.16.yaml
```diff
@@ -0,0 +1,90 @@
+package:
+  name: knative-serving-1.16
+  version: 1.16.0
+  epoch: 0
+  description: Kubernetes-based, scale-to-zero, request-driven compute
```
kranurag7, line 5: I think we should consider adding a provides here? `knative-serving=${{package.full-version}}` 💭

kranurag7 (approved):
great work!!

PR #39980, opened 2025-01-20
samba: use the internal libldb library.

sil2100:
As per https://lists.samba.org/archive/samba-announce/2024/000674.html , libldb is now shipped only as part of the samba source. Samba builds ldb by default during regular build, and with the current samba version we already had the library binaries in the `samba` binary package. Let's stop depending on the external ldb package - the only reverse-dependencies are samba and ldb itself.

For this, make sure we install libldb as part of `samba-libs` and not `samba` itself. This way packages such as `samba-client` will pull it in instantly and finally work as expected.

Was thinking of adding some provides, but I don't think those are necessary? The resolver knows how to resolve the dependency as-is. We'll have to think about what to do with the ldb package later (this can wait).
This is fixing the second part of: https://github.com/chainguard-dev/customer-issues/issues/1975

### DIFF
--- a/samba.yaml
+++ b/samba.yaml
@@ -1,7 +1,7 @@
 package:
   name: samba
   version: 4.21.3
-  epoch: 0
+  epoch: 1
   description: "Tools to access a server's filespace and printers via SMB"
   copyright:
     - license: GPL-3.0-or-later AND LGPL-3.0-or-later
@@ -24,7 +24,6 @@ environment:
       - gnutls-dev
       - iniparser-dev
       - jansson-dev
-      - ldb-dev
       - libarchive-dev
       - libcap-dev
       - libtirpc-dev
@@ -41,7 +40,6 @@ environment:
       - pkgconf-dev
       - popt-dev
       - py3-dnspython
-      - py3-ldb
       - py3-markdown
       - py3-tdb
       - py3-tevent
@@ -466,6 +464,7 @@ subpackages:
           mv "${{targets.destdir}}"/usr/lib/libsmbconf.so.* "${{targets.contextdir}}"/usr/lib
           mv "${{targets.destdir}}"/usr/lib/libsmbldap.so.* "${{targets.contextdir}}"/usr/lib
           mv "${{targets.destdir}}"/usr/lib/libtevent-util.so.* "${{targets.contextdir}}"/usr/lib
+          mv "${{targets.destdir}}"/usr/lib/libldb.so.* "${{targets.contextdir}}"/usr/lib
 
 update:
   enabled: true


max-allan:
CVE-2011-2411 :  on HPE non-stop servers.
CVE-2022-32743 & CVE-2022-1615 look real, and affect "Samba 4.1+" according to NVD.
But 32743 was fixedin 4.15 : https://bugzilla.samba.org/show_bug.cgi?id=14833
And 1615 was fixed in 4.16.2 https://bugzilla.samba.org/show_bug.cgi?id=15103

It might be worth double checking the fixes are in the code we're building and then marking these issues as False pos/fixed.

dannf (commented):
Ideally we'd still produce the ldb/ldb-dev packages from this source of course, but I get that we don't currently have a way to version subpackages separately from the main package, so this seems like the best available option. It would be nice if melange supported that.

Should we drop `!ldb` from the `--private-libraries` configure option? We are essentially making it private here, so it *sounds* like the correct thing to do, but I'm not sure what  that changes in the build.

Seems like we should remove ldb.yaml as well, WDYT?

Do we need to copy any test cases over from ldb.yaml?

I currently agree with not adding provides, since this isn't something we'd want existing ldb users to migrate to. Side note - I've wondered if we shouldn't create some dummy package - e.g. "package-no-longer-exists" - that provides all no-longer-updated packages as a stop-gap until we have an expunge process, to try to prevent anyone installing unmaintained things.

sil2100:
I think the ldb.yaml removal + test bits should be a separate thread to this PR here. Current samba doesn't work at all and is anyway essentially linking against the built-in libldb anyway. I'd prefer it as a separate PR? Maybe one thing to check is if any images pull in both samba AND libldb directly? Since other than that , things should just continue to work.
I think the ldb package's only relevant tests are the ldd-check ones, which I think should be added at some point - but since those would take care of all samba-libs libraries, not only this single one, I'd also do it in a separate PR.

The !ldb part hm, I agree it's something we should look into. Do you want me to experiment with it in this PR?

sil2100:
I tried removing the `!ldb` part and it changed nothing: library still had the provides and shipped the library normally.

But I'm thinking it's a good thing. I don't think it's wrong for `samba-libs` to provide so:libldb, since that's what it does. And if some future binary is looking for libldb, it will pull in the right package as a dependency.
We could look into maybe using `--builtin-libraries` for samba and only bundling libdb, but I don't know if it makes sense. I see that samba-libs has a lot of various samba-related libraries, so the placement seems correct.

WDYT?

dannf:
> I tried removing the `!ldb` part and it changed nothing: library still had the provides and shipped the library normally.
> 
> But I'm thinking it's a good thing. I don't think it's wrong for `samba-libs` to provide so:libldb, since that's what it does. And if some future binary is looking for libldb, it will pull in the right package as a dependency. We could look into maybe using `--builtin-libraries` for samba and only bundling libdb, but I don't know if it makes sense. I see that samba-libs has a lot of various samba-related libraries, so the placement seems correct.
> 
> WDYT?

Thanks for checking! My concern was that the resolver would prefer `samba-libs` over `libldb` when resolving `so:libldb.so.2` because `samba-libs` has a greater package version. Of course, that is what we'd want right now, but it would make it difficult to split `libldb` back out again say, if we added `sssd` to wolfi. I experimented locally and found my  concern to be unfounded - the package version doesn't appear to come into play when dealing with `so:` depends.

However, AFAICT it is _undefined_ as to whether or not you get `samba-libs` or the old `ldb` from the archive when requesting `so:libldb.so.2`.  Locally, `ldb` wins for me, but that may just be the first one it comes across when scanning. If we want to ensure `samba-libs` wins today, we'll need to give `samba-libs` a bumped `provider-priority:`. And should we split it back out, we'd need to give the new `ldb` an even higher `provider-priority`.

Since I feel I now understand the impact, and how we can deal with it moving forward, I'll go ahead and approve. I'll leave it to you to decide if you want to bump the provider-priority for samba-libs now to make the resolution deterministic. But since nothing else depends on libldb - this is likely completely theoretical.

dannf (approved):

sil2100:
Oh! That is indeed interesting to know about our resolver logic. It is certainly something different from what I am used to originally - but it makes sense. Thanks!

PR #39981, opened 2025-01-20
apache2: update epoch to 6 and add curl dependency; download httpd-fo…

utieyin:
Updated the Apache2 package to allow the dynamic download of the entrypoint script called httpd-foreground and place it in a similar location to the upstream docker image here https://github.com/docker-library/httpd/blob/master/2.4/alpine/Dockerfile#L231

### DIFF
--- a/apache2.yaml
+++ b/apache2.yaml
@@ -28,6 +28,12 @@ environment:
       - wolfi-base
       - zlib-dev
 
+var-transforms:
+  - from: ${{package.version}}
+    match: ^(\d+\.\d+)\.\d+$
+    replace: "$1"
+    to: build-version
+
 pipeline:
   - uses: git-checkout
     with:
@@ -92,7 +98,7 @@ pipeline:
 
   - uses: fetch
     with:
-      uri: https://raw.githubusercontent.com/docker-library/httpd/master/2.4/alpine/httpd-foreground
+      uri: https://raw.githubusercontent.com/docker-library/httpd/master/${{vars.build-version}}/alpine/httpd-foreground
       expected-sha256: 003f4ca165cf0daa9721c907259d3ef6a7fd9017453079780c1ae91de96fa777
       extract: false
 
@@ -148,7 +154,6 @@ subpackages:
     pipeline:
       - runs: |
           set -x
-          export short_version=$(echo ${{package.version}} | cut -d. -f1-2)
 
           # Create necessary folders
           mkdir -p "${{targets.subpkgdir}}"/usr/local/apache2/
@@ -225,8 +230,9 @@ subpackages:
             grep -E '^\s*CustomLog /proc/self/fd/1' /usr/local/apache2/conf/httpd.conf;
             grep -E '^\s*CustomLog /proc/self/fd/1' /usr/local/apache2/conf/extra/httpd-ssl.conf;
             grep -E '^\s*TransferLog /proc/self/fd/1$' /usr/local/apache2/conf/extra/httpd-ssl.conf;
-
-            adduser -D -H www-data
+            if ! id -u "www-data" > /dev/null 2>&1; then
+              adduser -S -G "www-data" "www-data"
+            fi
 
             httpd -M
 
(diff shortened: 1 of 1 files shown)


apache2.yaml
```diff
@@ -142,21 +143,23 @@ subpackages:
...
           mkdir -p "${{targets.subpkgdir}}"/usr/local/apache2/logs
+          mkdir -p "${{targets.contextdir}}"/usr/local/bin
 
           # Install necessary config files
           mkdir -p "${{targets.subpkgdir}}"/etc/apache2
           cp "${{targets.destdir}}"/etc/apache2/original/httpd.conf "${{targets.subpkgdir}}"/etc/apache2
           cp -r "${{targets.destdir}}"/etc/apache2/original/extra/ "${{targets.subpkgdir}}"/etc/apache2
 
+          # Download the `httpd-foreground` script from the official Docker image
+          curl -vk -L https://raw.githubusercontent.com/docker-library/httpd/master/${short_version}/alpine/httpd-foreground -o ${{targets.contextdir}}/usr/local/bin/httpd-foreground
```
xnox, line 161: can this at all use git-checkout pipeline into a subdirectory and then install from there?
Or use fetch pipeline and assert a checksum?

This likely can change, without being possible to detect change, and thus will not have any life-cycle around it for updates.

jamonation: I'd just put it into our tree under `apache2/httpd-foreground`. The contents are not worth the effort here:

```
#!/bin/sh
set -e

# Apache gets grumpy about PID files pre-existing
rm -f /usr/local/apache2/logs/httpd.pid

exec httpd -DFOREGROUND "$@"
```

xnox: i guess fetch with checksum would be better; cause we don't expect it to change ever; but will notice when it does change if we checkout from a branch; and assert checksum; and checksum will fail when this gets updated.

utieyin: Moved it to fetch with checksum

apache2.yaml
```diff
@@ -142,9 +148,7 @@ subpackages:
     pipeline:
       - runs: |
           set -x
-
-          # Install the entrypoint script
-          install -Dm755 entrypoint.sh "${{targets.subpkgdir}}"/usr/bin/entrypoint.sh
+          export short_version=$(echo ${{package.version}} | cut -d. -f1-2)
```
xnox, line 152: is this intentionally dropping entrypoint.sh? is it no longer needed?
do we need short_version? seems unused.

apache2.yaml
```diff
@@ -90,6 +90,12 @@ pipeline:
 
   - uses: strip
 
+  - uses: fetch
+    with:
+      uri: https://raw.githubusercontent.com/docker-library/httpd/master/2.4/alpine/httpd-foreground
+      expected-sha256: 003f4ca165cf0daa9721c907259d3ef6a7fd9017453079780c1ae91de96fa777
```
xnox, line 96: for 2.4 you could use var-transform to compute 2.4.

apache2.yaml
utieyin: Yea, unintentional. Removed

apache2.yaml
utieyin: Done, thanks

apache2.yaml
```diff
@@ -143,9 +155,6 @@ subpackages:
       - runs: |
           set -x
 
-          # Install the entrypoint script
-          install -Dm755 entrypoint.sh "${{targets.subpkgdir}}"/usr/bin/entrypoint.sh
-
```
xnox, line 148: why is entrypoint script being removed?

if this is unintentional, it should be added back in.

utieyin: The file httpd-foreground is what we referred to as entrypoint.sh. In the upstream image, there is no reference to entrypoint but httpd-foreground which is why it was removed. The content is however thesame for entrypoint.sh and httpd-foreground

jamonation (approved):

PR #40003, opened 2025-01-20
rpcbind: add missing required dependencies

sil2100:
rpcbind already pulls in libtirpc through its so dependency, but it also requires the `/etc/netconfig` file that comes from `libtirpc-conf`. Let's try to pull in both explicitly so it's clear.

...then again question is: should we explicitly add config package dependencies? Or do we expect users to either provide their own configuration and/or install the package themselves?

Another thing that rpcbind also requires is the `rpc` user. I was thinking of adding scriptlets to create those, but looking at other packages, most of the time we don't explicitly create the required users and let the users do it themselves. Something to discuss for sure.

Addresses: https://github.com/chainguard-dev/customer-issues/issues/1980


sil2100:
Wanted to add a test for this, but I think I need to learn a bit more about the test environment for this to work as expected.

sil2100:
Some commentary here:
The post-install creation of the /var/run symlink is rather temporary. I really think we should move this to wolfi-baselayout, and I'm considering starting a separate PR for that. However, that requires making sure that we then remove all the /var/run creation instances from existing packages, as this can cause conflicts (and breakage). Which is why I don't want to do it here just yet. Especially that we already self-create /var/run for multiple packages.

There was a discussion about the user creation, and this is mostly just for apk usage and demonstration of what is needed to get this working. In most cases the scriptlets are not executed - not on test runs, not in image builds. But I still think that it might be useful to do it here.
BUT! I am open to comments. This might give people the wrong idea that everything just works out-of-the-box, which is non ideal. Now it's a matter of policy which way we prefer.

dannf:
> Some commentary here: The post-install creation of the /var/run symlink is rather temporary. I really think we should move this to wolfi-baselayout, and I'm considering starting a separate PR for that. However, that requires making sure that we then remove all the /var/run creation instances from existing packages, as this can cause conflicts (and breakage). Which is why I don't want to do it here just yet. Especially that we already self-create /var/run for multiple packages.

Understood, thanks for adding that comment.

> There was a discussion about the user creation, and this is mostly just for apk usage and demonstration of what is needed to get this working. In most cases the scriptlets are not executed - not on test runs, not in image builds. But I still think that it might be useful to do it here. BUT! I am open to comments. This might give people the wrong idea that everything just works out-of-the-box, which is non ideal. Now it's a matter of policy which way we prefer.

I think it makes sense the way you have it. Thanks for researching this, I learned a lot.

dannf (approved):

PR #40036, opened 2025-01-20
firefox: build from source tarball instead of hg

OddBloke:
The Mercurial `_RELEASE` tag seems to be created some time after the release has been made public on ftp.mozilla.org. release-monitoring.org uses ftp.mozilla.org as its source of truth, which generally means that we detect the new version and attempt to build the release before the tag has been published, resulting in build failures.

Switching to the release tarball should avoid manual intervention for every Firefox release, thereby decreasing the latency from upstream release to availability in Wolfi.  It has the additional advantage of future-proofing us against the planned migration to GitHub for development (though once that's occurred, we should examine whether consuming from GitHub directly would be appropriate).  (`wget` is probably also faster than `hg clone`, though I haven't confirmed this.)

### DIFF
--- a/firefox.yaml
+++ b/firefox.yaml
@@ -1,7 +1,7 @@
 package:
   name: firefox
   version: "134.0.2"
-  epoch: 0
+  epoch: 1
   description: Firefox web browser
   copyright:
     - license: GPL-3.0-only AND LGPL-2.1-only AND LGPL-3.0-only AND MPL-2.0
@@ -57,7 +57,6 @@ environment:
       - nodejs
       - pango-dev
       - pulseaudio-dev
-      - py3.11-mercurial
       - py3.11-pip # Matches Mach's recommended maximum Python version
       - python-3.11-dev # (3.12 definitely didn't work at the time of setting this)
       - rust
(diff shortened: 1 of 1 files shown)


EyeCantCU (commented):
This looks like a significant improvement over cloning the source with mercurial. May need to bump the epoch here though (elastic never ran)

firefox.yaml
```diff
@@ -94,11 +78,8 @@ pipeline:
       # with release repositories, so just extract what it does and do it
       # manually.
       # https://firefox-source-docs.mozilla.org/setup/linux_build.html
-      # Clone the repository. If the major version is changed, this needs to
-      # be updated.
-      echo "RUNNING: hg clone https://hg.mozilla.org/releases/mozilla-release/ -u ${RELEASE_TAG}"
-      hg clone https://hg.mozilla.org/releases/mozilla-release/ -u ${RELEASE_TAG}
-      cd mozilla-release
+      wget https://ftp.mozilla.org/pub/firefox/releases/${{package.version}}/source/firefox-${{package.version}}.source.tar.xz -O- | tar Jx
+      cd firefox-${{package.version}}
```
EyeCantCU, line 82: Instead of using `wget` directly, we can use the fetch pipeline here

OddBloke: Done, thank you!

EyeCantCU (approved):

PR #40068, opened 2025-01-21
add kubectl-images package

kranurag7:
for doing something like this on workstations and general troubleshooting. 
```bash
 $ kubectl images -n kube-system
[Summary]: 1 namespaces, 2 pods, 2 containers and 2 different images
+-----------------------------------------+------------------------+-----------------------------------------+
|                   Pod                   |       Container        |                  Image                  |
+-----------------------------------------+------------------------+-----------------------------------------+
| coredns-ccb96694c-hzg9g                 | coredns                | rancher/mirrored-coredns-coredns:1.12.0 |
+-----------------------------------------+------------------------+-----------------------------------------+
| local-path-provisioner-5cf85fd84d-tf2rv | local-path-provisioner | rancher/local-path-provisioner:v0.0.30  |
+-----------------------------------------+------------------------+-----------------------------------------+
```

### DIFF
--- a/kubectl-images.yaml
+++ b/kubectl-images.yaml
@@ -13,6 +13,7 @@ pipeline:
       tag: v${{package.version}}
       expected-commit: 582899624f93be6e8728125396ab567ae914bd0d
 
+  # NOTE: there's is no active releases for past 2 years but it's a single go file with small deps and given we always build using latest go, this should be good.
   - uses: go/build
     with:
       packages: ./cmd/main.go
--- a/kubectl-images.yaml
+++ b/kubectl-images.yaml
@@ -1,0 +1,32 @@
+package:
+  name: kubectl-images
+  version: 0.6.3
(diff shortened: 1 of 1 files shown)


mritunjaysharma394 (changes requested):
I see it has no active commits/releases since 2 years, are we sure it is maintained?

mritunjaysharma394:
Ack, thanks, @kranurag7, It is probably a good idea to comment on this somewhere for future reference of sustaining team, or if any custome in future requests etc

EyeCantCU (changes requested):
If we need this, we should fork and maintain it. At least implement automation for dependency bumps. But as-is, this is not maintained and should not be packaged in Wolfi

kranurag7:
closing due to maintenance concerns. I'll use `jsonpath` in image tests and for local usage build from source.

PR #40096, opened 2025-01-21
openjdk: Re-add JDK jre symlink

pnasrat:


openjdk-22.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: openjdk-22
   version: 22.0.2
-  epoch: 3
+  epoch: 4
```
jamonation, line 4: Does this need the symlink added too?

jamonation (approved):

Issue #34322, opened 2024-11-16
libbpf will fail pkgconf test pipeline


murraybd:
If the pkgconf test pipeline is added to libbpf it will fail in the following way:

```
2024/11/12 23:11:50 INFO running step "test/pkgconf"
2024/11/12 23:11:50 WARN + '[' -d /home/build ] uses=test/pkgconf
2024/11/12 23:11:50 WARN + cd /home/build uses=test/pkgconf
2024/11/12 23:11:50 WARN + exit 0 uses=test/pkgconf
2024/11/12 23:11:50 INFO running step "pkgconf build dependency check" uses=test/pkgconf
2024/11/12 23:11:50 WARN + '[' -d /home/build ] uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:11:50 WARN + cd /home/build uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:11:50 WARN + basename /home/build/melange-out/libbpf uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:11:50 WARN + dev_pkg=libbpf uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:11:50 WARN + cd / uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:11:50 WARN + apk info -L libbpf uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:11:50 WARN + grep '\.pc$' uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:11:50 WARN WARNING: opening /home/user/src/wolfi-os/packages: No such file or directory uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:11:50 WARN WARNING: opening from cache https://packages.wolfi.dev/os: No such file or directory uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:11:50 WARN + grep -q ^Name: usr/lib64/pkgconfig/libbpf.pc uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:11:50 WARN + basename usr/lib64/pkgconfig/libbpf.pc .pc uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:11:50 WARN + lib_name=libbpf uses=test/pkgconf name="pkgconf build dependency check"
[...]

PR #40164, opened 2025-01-22
Build libbpf with LIBDIR so pkgconf passes

murraybd:
If `LIBDIR` is not set then the .pc files end up in `/usr/lib64/pkgconfig/`, as seen in https://github.com/wolfi-dev/os/issues/34322, which then causes the pkgconf test pipeline to fail. This resolves that issue and additionally fixes some failures with apk audit and the current version of the package.

### DIFF
--- a/libbpf.yaml
+++ b/libbpf.yaml
@@ -1,7 +1,7 @@
 package:
   name: libbpf
   version: 1.5.0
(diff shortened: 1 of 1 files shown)


dannf (approved):
nice.

PR #40194, opened 2025-01-22
package(kbd): add kbd package, add as dependency of systemd-init

89luca89:
kbd is a dependency of systemd when used as pid1 to load keymaps

### DIFF
--- a/kbd.yaml
+++ b/kbd.yaml
@@ -37,6 +37,8 @@ pipeline:
 
   - uses: autoconf/make-install
 
+  - uses: strip
+
 update:
   enabled: true
   release-monitor:
--- a/kbd.yaml
+++ b/kbd.yaml
@@ -5,9 +5,6 @@ package:
   description: Linux keyboard tools
   copyright:
     - license: GPL-2.0-or-later
-  resources:
-    cpu: 4
-    memory: 12Gi
 
 environment:
   contents:
--- a/kbd.yaml
+++ b/kbd.yaml
@@ -1,0 +1,52 @@
+package:
+  name: kbd
+  version: "2.7.1"
+  epoch: 0
+  description: Linux keyboard tools
+  copyright:
+    - license: GPL-2.0-or-later
+  resources:
+    cpu: 4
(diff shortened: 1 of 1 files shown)


kbd.yaml
```diff
@@ -0,0 +1,52 @@
...
+        --disable-nls \
+        --prefix=/usr \
+        --sysconfdir=/etc \
+        --datadir=/usr/share \
+        --htmldir=/usr/share/html/${{package.name}} \
+        --disable-static
+
+  - uses: autoconf/make
+
+  - uses: autoconf/make-install
```
mattmoor, line 38: Do we need to strip?

xnox: Yes and no. We don't add -g anywhere, and all binaries are built without debug symbols, so most of the time our strip pipeline is a no-op.

I sort of want to move away with having strip pipeline everywhere, and instead have melange force inject -g0 compiler flag (or be tunable, such that if you want a debug rebuild you can do that without changing .yaml; and for software to opt-in into debug symbols if they need them - like openssl packaging does it).

smoser:
Does this work? My understanding was that there is a database of config files needed too?

89luca89:
> Does this work? My understanding was that there is a database of config files needed too?

yes the list of mappings is in /usr/share

kbd.yaml
89luca89: Done

kbd.yaml
```diff
@@ -0,0 +1,54 @@
+package:
+  name: kbd
+  version: "2.7.1"
+  epoch: 0
+  description: Linux keyboard tools
+  copyright:
+    - license: GPL-2.0-or-later
+  resources:
+    cpu: 4
+    memory: 12Gi
```
smoser, line 10: really? or copy-pasta?

89luca89: copy-pasta I can remove if you prefer

89luca89: removed

xnox (approved):

PR #40197, opened 2025-01-22
feat(services): add service auto-enable in multiuser target

89luca89:
enable by default some services linking to multiuser.target
add /etc/machine-id to systemd-init, it's needed to bootstrap some services
add kbd dependency for systemd-init

### DIFF
--- a/systemd.yaml
+++ b/systemd.yaml
@@ -370,11 +370,17 @@ subpackages:
         - ${{package.name}}
         - tzdata
         - mount
+        - kbd
     pipeline:
       - runs: |
           mkdir -p ${{targets.subpkgdir}}/etc ${{targets.subpkgdir}}/sbin
-          ln -s ../usr/lib/systemd/systemd ${{targets.subpkgdir}}/sbin/init
+          mkdir -p "${{targets.subpkgdir}}/usr/lib/systemd/system/multi-user.target.wants/"
+          ln -sf ../usr/lib/systemd/systemd ${{targets.subpkgdir}}/sbin/init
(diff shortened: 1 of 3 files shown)


xnox (changes requested):
can we chat about this

systemd.yaml
```diff
@@ -369,11 +369,14 @@ subpackages:
...
         - mount
+        - kbd
+        - agetty
     pipeline:
       - runs: |
           mkdir -p ${{targets.subpkgdir}}/etc ${{targets.subpkgdir}}/sbin
-          ln -s ../usr/lib/systemd/systemd ${{targets.subpkgdir}}/sbin/init
-          ln -sf ../usr/share/zoneinfo/UTC ${{targets.subpkgdir}}/etc/localtime
+          ln -sf ../usr/lib/systemd/systemd ${{targets.subpkgdir}}/sbin/init
+          ln -sf ../run/systemd/resolve/stub-resolv.conf "${{targets.subpkgdir}}/etc/resolv.conf"
```
smoser, line 378: just a comment. '-s' rather than '-sf' is nicer in that it will fail if something else changes and this is already there.  ie, if something above moved this around or something produced content there, then the build would fail rather than finding out later that we masked some content unintentionally.

xnox (dismissed):

xnox:
@89luca89 other prs got merged; so this needs a rebase now.

89luca89:
> @89luca89 other prs got merged; so this needs a rebase now.

done :+1:

xnox (dismissed):

justinvreeland (dismissed):

xnox (dismissed):

xnox (approved):

PR #40202, opened 2025-01-22
add compat package for rclone

kranurag7:
adding a compat to rclone package.


mamccorm:
We're working on this in separate PRs so we'll take care of this in another commit

PR #40204, opened 2025-01-22
chore: Expand test coverage and add a compat

8none1:
Add a compat to be compatible with the upstream paths and add new functional tests to confirm functionality and binary version.
Existing package.


kranurag7 (approved):
awesome work!! 🚀 

looks good, thank you for adding nice functional tests and explicit ldflags at build time.

PR #40225, opened 2025-01-22
fix(airflow): update dependencies, build UI files before installing the package

wojciechka:


kranurag7 (approved):
thank you!! looks good. The UI part is tested and we've static assets inside the package now and we don't get warning because of missing binaries.

PR #40245, opened 2025-01-22
fix(udev): fix build of empty packages for udev

89luca89:


justinvreeland (changes requested):
need an epoch bump.

89luca89:
> need an epoch bump.

Done now :+1:

justinvreeland (approved):
Thanks for fixing this!

PR #40247, opened 2025-01-22
pre-commit: add more checks

imjasonh:


imjasonh:
@dannf I'm not sure what we want to do about the Ruby patch. I believe we can configure an exemption for it, or we can remove the check entirely for this repo, or we can update the patch (which I'm less comfortable with)

dannf:
> @dannf I'm not sure what we want to do about the Ruby patch. I believe we can configure an exemption for it, or we can remove the check entirely for this repo, or we can update the patch (which I'm less comfortable with)

+1 on an exemption - I think we can do something like:

```
  - id: detect-private-key
    exclude: ^ruby-3\.0/0001-ruby-3\.0\.6-openssl-patch\.patch$
```

xnox (approved):

PR #40248, opened 2025-01-22
audit: remove openrc support, fix systemd integration

xnox:
Ensure pidfile is expected at /var/run location. Note that right now
systemd unit expects pidfile at an invalid location at /usr/var/run
which. It is created var.conf tmpfiles creates that symlink, and
source code writes it there, so yeah that.

Drop openrc integration, unused.


justinvreeland (dismissed):

audit.yaml
```diff
@@ -63,6 +55,7 @@ pipeline:
           --with-libcap-ng=yes
           --without-golang
           --with-io_uring
+          --runstatedir=/var/run
```
89luca89, line 58: `--sbindir=/usr/sbin`

89luca89 (changes requested):

audit.yaml
89luca89, line 58:
```suggestion
          --runstatedir=/var/run
          --sbindir=/usr/sbin
```

89luca89 (changes requested):

audit.yaml
89luca89: with this I've verified it boots without issues and errors now

89luca89 (approved):

PR #40273, opened 2025-01-22
Implement usrmerge for binaries

xnox:
This implements usrmerge for binaries. See:
- https://github.com/orgs/wolfi-dev/discussions/40270

### DIFF
--- a/busybox.yaml
+++ b/busybox.yaml
@@ -1,12 +1,14 @@
 package:
   name: busybox
   version: 1.37.0
-  epoch: 0
+  epoch: 1
   description: "swiss-army knife for embedded systems"
   copyright:
     - license: GPL-2.0-only
   dependencies:
     provider-priority: 10
+    runtime:
+      - wolfi-baselayout>=20230201-r17
   scriptlets:
     trigger:
       paths:
@@ -15,8 +17,13 @@ package:
         - /usr/bin
         - /usr/sbin
       script: |
-        #!/bin/busybox sh
-        /bin/busybox --install -s
+        #!/usr/bin/busybox sh
+        for i in `/usr/bin/busybox --list`; do
+          [ -e /usr/bin/$i ] && continue
+          [ -h /usr/bin/$i ] && continue
+          # Apko currently creates it like this
+          /usr/bin/busybox ln -s /bin/busybox /usr/bin/$i
+        done
 
 var-transforms:
   - from: ${{package.version}}
@@ -58,20 +65,17 @@ pipeline:
 
   - name: Install
     runs: |
-      mkdir -p "${{targets.destdir}}"/usr/sbin
       mkdir -p "${{targets.destdir}}"/usr/bin
       mkdir -p "${{targets.destdir}}"/tmp
       mkdir -p "${{targets.destdir}}"/var/cache/misc
-      mkdir -p "${{targets.destdir}}"/bin
-      mkdir -p "${{targets.destdir}}"/sbin
       mkdir -p "${{targets.destdir}}"/etc
       mkdir -p "${{targets.destdir}}"/usr/share/man/man1
       chmod 1777 "${{targets.destdir}}"/tmp
-      install -m755 busybox "${{targets.destdir}}"/bin/busybox
+      install -m755 busybox "${{targets.destdir}}"/usr/bin/busybox
       install -m644 securetty "${{targets.destdir}}"/etc/securetty
 
       mkdir -p "${{targets.destdir}}"/etc/busybox-paths.d
-      ./busybox --list-path > "${{targets.destdir}}"/etc/busybox-paths.d/busybox
+      ./busybox --list-path | sed 's|^bin|usr/bin|;s|^sbin|usr/bin|;s|^usr/sbin|usr/bin|' > "${{targets.destdir}}"/etc/busybox-paths.d/busybox
 
 subpackages:
   - name: busybox-full
@@ -89,20 +93,17 @@ subpackages:
           make CC="${{host.triplet.gnu}}-gcc" V=1 -j$(nproc)
       - name: Install
         runs: |
-          mkdir -p "${{targets.subpkgdir}}"/usr/sbin
           mkdir -p "${{targets.subpkgdir}}"/usr/bin
           mkdir -p "${{targets.subpkgdir}}"/tmp
           mkdir -p "${{targets.subpkgdir}}"/var/cache/misc
-          mkdir -p "${{targets.subpkgdir}}"/bin
-          mkdir -p "${{targets.subpkgdir}}"/sbin
           mkdir -p "${{targets.subpkgdir}}"/etc
           mkdir -p "${{targets.subpkgdir}}"/usr/share/man/man1
           chmod 1777 "${{targets.subpkgdir}}"/tmp
-          install -m755 busybox "${{targets.subpkgdir}}"/bin/busybox
+          install -m755 busybox "${{targets.subpkgdir}}"/usr/bin/busybox
           install -m644 securetty "${{targets.subpkgdir}}"/etc/securetty
 
           mkdir -p "${{targets.subpkgdir}}"/etc/busybox-paths.d
-          ./busybox --list-path > "${{targets.subpkgdir}}"/etc/busybox-paths.d/busybox-full
+          ./busybox --list-path | sed 's|^bin|usr/bin|;s|^sbin|usr/bin|;s|^usr/sbin|usr/bin|' > "${{targets.subpkgdir}}"/etc/busybox-paths.d/busybox-full
     scriptlets:
       trigger:
         paths:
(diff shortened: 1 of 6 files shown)


busybox.yaml
```diff
@@ -58,20 +58,17 @@ pipeline:
...
       mkdir -p "${{targets.destdir}}"/etc
       mkdir -p "${{targets.destdir}}"/usr/share/man/man1
       chmod 1777 "${{targets.destdir}}"/tmp
-      install -m755 busybox "${{targets.destdir}}"/bin/busybox
+      install -m755 busybox "${{targets.destdir}}"/usr/bin/busybox
       install -m644 securetty "${{targets.destdir}}"/etc/securetty
 
       mkdir -p "${{targets.destdir}}"/etc/busybox-paths.d
-      ./busybox --list-path > "${{targets.destdir}}"/etc/busybox-paths.d/busybox
+      ./busybox --list-path | sed 's|^bin|usr/bin|;s|^sbin|usr/bin|;s|^usr/sbin|usr/bin|' > "${{targets.destdir}}"/etc/busybox-paths.d/busybox
```
kaniini, line 71: This will be inconsistent if apko is not used to install busybox.  It would be better to fix the paths in busybox itself.

A simple way of building a wolfi chroot with strictly apk:

```
apk add —root /path/to/wherever —initdb —allow-untrusted -X https://packages.wolfi.dev/os wolfi-baselayout busybox
```

xnox: @kaniini yeah it is sad. Also looking at all the busybox code; i couldn't quickly find where/if paths are set as to which applet goes into bin-vs-sbin and into usr-vs-root. There is `INSTALL_NO_USR` config key, but I guess I need/want `INSTALL_ALWAYS_USR` which is not a thing (yet)?

kaniini: I think I agree with your conclusion, though the install path should be set in each applet definition.  There is probably a way to hack around this though.  I’m happy to look into it since we need to do it in Alpine too…

xnox: @kaniini https://gist.github.com/xnox/5d522f187b817ca8f76f126c5c39ef96 this is ugly and probably very redundant........ and not sure it does the right stuff for setuid things, but maybe that's sort of what we need here?

xnox:
Also currently hitting:

> 2025/01/22 23:43:46 ERRO failed to build package: unable to build guest: unable to generate image: installing apk packages: installing packages: installing coreutils (ver:9.6-r0 arch:x86_64): unable to install files for pkg coreutils: unable to install symlink from usr/bin/env -> coreutils: symlink coreutils /tmp/melange-guest-618158088/usr/bin/env: file exists

busybox.yaml
xnox: another approach; instead of patching busybox; hack the scriplets:

```
# wget https://apk.cgr.dev/wolfi-presubmit/b9ae84f89b7a9bf253aaff9ed7d5e37dcb718ffa/x86_64/busybox-1.37.0-r1.apk
# apk add --root /tmp/install-test --initdb --allow-untrusted -X https://packages.wolfi.dev/os wolfi-baselayout ./busybox-1.37.0-r1.apk
# cd /tmp/install-test-2/

# ls bin/ sbin/ usr/sbin/
bin/:

sbin/:
ldconfig

usr/sbin/:

# ls -l usr/bin/busybox usr/bin/ash 
lrwxrwxrwx    1 root     root            16 Jan 23 11:13 usr/bin/ash -> /usr/bin/busybox
-rwxr-xr-x    1 root     root        694808 Jan 23 11:01 usr/bin/busybox
```

Better but I do not like absolute symlink, instead of a relative one.

xnox: ah but it is defined as a symlink to /bin/busybox for now in apko. sadness, but ok.

busybox.yaml
```diff
@@ -15,8 +15,12 @@ package:
...
         - /usr/sbin
       script: |
-        #!/bin/busybox sh
-        /bin/busybox --install -s
+        #!/usr/bin/busybox sh
+        for i in `/usr/bin/busybox --list`; do
+          [ -e /usr/bin/$i ] && continue
+          [ -h /usr/bin/$i ] && continue
+          /usr/bin/busybox ln -s /bin/busybox /usr/bin/$i
+        done
```
xnox, line 23: @kaniini so with this, apk --initdb and apko both create identical images for busybox only. Without needing to patch busybox.

still need to make this work in concert with coreutils.

busybox.yaml
```diff
@@ -15,8 +17,12 @@ package:
...
```
smoser, line 25: Why /bin/busybox here?

Also, relative would be nice.

busybox.yaml
```diff
@@ -58,20 +64,17 @@ pipeline:
...
       mkdir -p "${{targets.destdir}}"/etc
       mkdir -p "${{targets.destdir}}"/usr/share/man/man1
       chmod 1777 "${{targets.destdir}}"/tmp
-      install -m755 busybox "${{targets.destdir}}"/bin/busybox
+      install -m755 busybox "${{targets.destdir}}"/usr/bin/busybox
       install -m644 securetty "${{targets.destdir}}"/etc/securetty
 
       mkdir -p "${{targets.destdir}}"/etc/busybox-paths.d
-      ./busybox --list-path > "${{targets.destdir}}"/etc/busybox-paths.d/busybox
+      ./busybox --list-path | sed 's|^bin|usr/bin|;s|^sbin|usr/bin|;s|^usr/sbin|usr/bin|' > "${{targets.destdir}}"/etc/busybox-paths.d/busybox
```
smoser, line 77: Everything into /usr/bin?

Then our links can be just 'ln -s busybox other-thing'

wolfi-baselayout.yaml
```diff
@@ -1,13 +1,64 @@
...
   description: "baselayout data for Wolfi"
   copyright:
     - license: MIT
   dependencies:
     runtime:
       - ca-certificates-bundle
+  scriptlets:
+    # Must work with busybox, coreutils, any $shell-binsh
+    pre-upgrade: |
+      #!/bin/sh
```
smoser, line 14: When can we get rid of this?

busybox.yaml
xnox: Because that is what currently apko generates See https://github.com/search?q=repo%3Achainguard-dev%2Fapko%20bin%2Fbusybox&type=code

Once this lands; we can work on fixing apko to start generating usr-merged layout; and likely relative links too. For now, I'm trying to ensure these upgrades get one the same layout & contents as the fresh apko build as much as possible - including these choices of paths.

busybox.yaml
xnox: Yes, but after this lands; dust settles; and apko is changed to produce usr-merged layout by default.

Because freshly built images for now will continue to use /bin/busybox for the symlink target.

wolfi-baselayout.yaml
xnox: As in the shebang? one must specify it; or the whole thing?

Apk will attempt the upgrade regardless; replacing symlinks will fail; and one ends up in a system with broken shell most likely. If one ends up in such state without backup one will have to do all of the below steps regardless; thus may as well document / attempt them anyway. Depending on what users have places in those locations it can still fail, but at least we ensure they will still have a working shell and utilities that we provided for them to fix up the container further if they have such a need.

I am only doing this, because there is no other way to stop the install or upgrade; as apk continues with unpacking broken things regardless.

I guess I can or should specify conflicts maybe? But not sure what I can conflict on.

busybox.yaml
dannf: makes sense, but isn't easily discoverable. Could you note that in a comment?

wolfi-baselayout.yaml
dannf: Added https://github.com/chainguard-dev/internal-dev/issues/8673 to track the post-migration code removal

wolfi-baselayout.yaml
```diff
@@ -1,13 +1,64 @@
...
+      [ -h /bin ] && exit 0
+      echo Performing usr-merge...
```
dannf, line 16: Should we `set -ex` this during the migration period to help with debugging?

busybox.yaml
xnox: We could instead of writing our own implementation ship suse's bash converter - https://build.opensuse.org/projects/openSUSE:Factory/packages/compat-usrmerge/files/convertfs

or debian's perl converter - https://salsa.debian.org/md/usrmerge/-/blob/master/convert-usrmerge?ref_type=heads

smoser: I just want to point a tool at a tree (${{targets.contextdir}}) and say "merge the usr here!".
it should be able to always pass.  but if it fails it should fail the build.

I'm sure I'm missing something and it is more complicated than i'm imagining.

then we run that from a pipeline as the last step of a package (and each subpackage).

i really want to be able to tell melange "always run these pipelines".

xnox:
We will be going with the pinball plan, but will keep this PR open for now because of other needs.

The pinball plan is:
- https://github.com/wolfi-dev/os/pull/42522
- https://github.com/wolfi-dev/os/pull/42525
- https://github.com/wolfi-dev/os/pull/42527

smoser:
I have several /sbin merge things https://github.com/wolfi-dev/os/pull/42723

PR #40317, opened 2025-01-23
add kn package

kranurag7:
knative client

### DIFF
--- a/kn.yaml
+++ b/kn.yaml
@@ -26,8 +26,7 @@ update:
   enabled: true
   github:
     identifier: knative/client
-    use-tag: true
-    strip-prefix: v
+    strip-prefix: knative-v
 
 test:
   pipeline:
--- a/kn.yaml
+++ b/kn.yaml
@@ -1,0 +1,37 @@
+package:
+  name: kn
+  version: 1.17.0
+  epoch: 0
+  description: Knative CLI implementation
+  copyright:
+    - license: Apache-2.0
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/knative/client
+      tag: knative-v${{package.version}}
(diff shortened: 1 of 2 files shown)


developer-guy:
soooo, we are managing multiple versions of knative releases, so, making knative client tool as version streamed would be great if we could do it? I'm curios about what you think? 🙄

kranurag7:
> soooo, we are managing multiple versions of knative releases, so, making knative client tool as version streamed would be great if we could do it? I'm curios about what you think? 🙄

that's a great question, I didn't create version stream for client side tooling given I considered them to a little mature and can work across version especially for commonly used features something like kubectl. 
We'll also have to maintain a separate version stream for this for updates because the release cadence for this is different. 

This is the main reason to go with latest for now, if you think we should version stream then happy to follow that path. :)

kn.yaml
```diff
@@ -0,0 +1,36 @@
+package:
+  name: kn
```
Dentrax, line 2: Would the package name `knative-client` be better?

kn.yaml
```diff
@@ -0,0 +1,36 @@
...
+  version: 1.17.0
+  epoch: 0
+  description: Knative CLI implementation
+  copyright:
+    - license: Apache-2.0
+
```
Dentrax, line 8: Should we add `provides:` here for `knative=${{package.full-version}}` and `kn=${{package.full-version}}`

kn.yaml
kranurag7: Yes, org/repo naming convention is better. :)

kn.yaml
kranurag7: added a provides for `kn` but not for `knative` mostly thinking that in future maybe we'll have a meta `knative` package someday. 💭

developer-guy:
let's merge this PR

developer-guy (approved):

PR #40329, opened 2025-01-23
glibc: switch to git and cherry-pick fix for CVE-2025-0395

sil2100:
Fix for CVE-2025-0395. PR for the advisory: https://github.com/wolfi-dev/advisories/pull/11351

#### For security-related PRs

### DIFF
--- a/glibc.yaml
+++ b/glibc.yaml
@@ -1,7 +1,7 @@
 package:
   name: glibc
   version: 2.40
-  epoch: 4
+  epoch: 5
   description: "the GNU C library"
   copyright:
     - license: LGPL-2.1-or-later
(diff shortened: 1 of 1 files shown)


sil2100:
Since for this PR I am also switching from fetch to git-checkout (which ideally should be a separate PR), I wanted to make sure that the 2.40 tarball and the 2.40 release tag on git have the same contents. So I git cloned and checked out + fetched the tarball:

```
$ diff -qr glibc glibc-2.40
Only in glibc: .git
```

The sources are identical.

sil2100:
I will modify the detect advisory to a fix one once this lands.

justinvreeland (approved):

sil2100:
Ran tests of 50 random reverse-dependency packages and they seemed to be good. Landing.

PR #40366, opened 2025-01-23
systemd: move systemd-firstboot to separate package

89luca89:


xnox (dismissed):

systemd.yaml
```diff
@@ -406,6 +406,17 @@ subpackages:
...
+    description: "Fake login service to boot directly into a shell"
+    dependencies:
+      runtime:
+        - ${{package.name}}
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/usr/lib/systemd/system/
+          cp ${{targets.destdir}}/usr/lib/systemd/system/serial-getty@.service ${{targets.subpkgdir}}/usr/lib/systemd/system/
+          sed -i "s|^ExecStart=.*agetty.*|ExecStart=-/bin/sh -l|" ${{targets.subpkgdir}}/usr/lib/systemd/system/serial-getty@.service
+
```
xnox, line 419: this should never exists; so i'm glad we dropped it.

also no epoch bump. also you should be able to get the same effect with debug-shell & emergency-shell and set it to the appropriate locations, no?!

smoser (approved):

PR #40380, opened 2025-01-23
audit - Split out the library from the daemon.

smoser:
systemd has a runtime dependency on libaudit, but does not need all of the other things.  So, split the libraries out into libaudit package.  Also split python3 bindings out of the "main" package as the audit deamon does not need them.


roaksoax (commented):
lgtm!

xnox (approved):

Issue #34371, opened 2024-11-16
re2 will fail pkgconf test pipeline


murraybd:
If the pkgconf test pipeline is added to re2 it will fail in the following way:

```
2024/11/12 23:14:59 INFO running step "test/pkgconf"
2024/11/12 23:14:59 WARN + '[' -d /home/build ] uses=test/pkgconf
2024/11/12 23:14:59 WARN + cd /home/build uses=test/pkgconf
2024/11/12 23:14:59 WARN + exit 0 uses=test/pkgconf
2024/11/12 23:14:59 INFO running step "pkgconf build dependency check" uses=test/pkgconf
2024/11/12 23:14:59 WARN + '[' -d /home/build ] uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:14:59 WARN + cd /home/build uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:14:59 WARN + basename /home/build/melange-out/re2-dev uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:14:59 WARN + dev_pkg=re2-dev uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:14:59 WARN + cd / uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:14:59 WARN + apk info -L re2-dev uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:14:59 WARN + grep '\.pc$' uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:14:59 WARN WARNING: opening /home/user/src/wolfi-os/packages: No such file or directory uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:14:59 WARN WARNING: opening from cache https://packages.wolfi.dev/os: No such file or directory uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:14:59 WARN + grep -q ^Name: usr/lib/pkgconfig/re2.pc uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:14:59 WARN + basename usr/lib/pkgconfig/re2.pc .pc uses=test/pkgconf name="pkgconf build dependency check"
2024/11/12 23:14:59 WARN + lib_name=re2 uses=test/pkgconf name="pkgconf build dependency check"
[...]

PR #40382, opened 2025-01-23
Add a disabled pkgconf test for re2

murraybd:
I was able to manually get the pkgconf test to run by using `debug -i` and installing the packages added as runtime dependencies, however getting the pkgconf test to pass in infra requires rebuilding the package which it currently fails to do. Instead of losing my work let's add the dependencies and the test as a comment.

### DIFF
--- a/re2.yaml
+++ b/re2.yaml
@@ -1,6 +1,6 @@
 package:
   name: re2
-  version: "2024.02.01"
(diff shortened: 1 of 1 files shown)


murraybd:
This will, once its enabled, fix https://github.com/wolfi-dev/os/issues/34371.

dannf (approved):

PR #40394, opened 2025-01-23
jellyfin-web: add test

OddBloke:


powersj (approved):
Don't love the sleep and would love to have a retry mechanism, but as a simple test for now thank you!

PR #40432, opened 2025-01-24
Add umount as runtime dependency of secrets-store-csi-driver

max-allan-cgr:


wojciechka (dismissed):

kranurag7 (changes requested):
can we get a confirmation if we need more utilities like `mount` or maybe some more coming from `util-linux` or `util-linux-misc`?

max-allan-cgr:
@kranurag7 
OTHER mounter functions  used: List, Mount, IsLikelyNotMountPount
Mount uses mount which is already installed
isLikely... uses stat, err := os.Stat(file) I am assuming stat is built in to Go and no bash binary needed.
List uses: https://github.com/kubernetes/mount-utils/blob/954afbfb05e3c9338ec8253854e90cf0eadbfa0b/mount_helper_unix.go#L250 which uses an os.ReadFile and utilio.ConsistentRead (I assume both internal functions)

There's a possibility of other Exec.Commands hidden away in dependencies, but I could be here all week searching....

secrets-store-csi-driver.yaml
```diff
@@ -2,13 +2,14 @@ package:
...
   version: 1.4.7
-  epoch: 3
+  epoch: 4
   description: Secrets Store CSI driver for Kubernetes secrets
   copyright:
     - license: Apache-2.0
   dependencies:
     runtime:
       - busybox # the csi driver needs the 'mount' applet
+      - umount
```
kranurag7, line 12: @max-allan-cgr upstream is containing these binaries so we should include more runtime deps here. 

```bash
$ apk search cmd:wdctl
util-linux-misc-2.40.4-r0
$ apk search cmd:utmpdump
util-linux-misc-2.40.4-r0
$ apk search cmd:uclampset
util-linux-misc-2.40.4-r0
$ apk search cmd:taskset
util-linux-misc-2.40.4-r0
$ apk search cmd:wall
util-linux-misc-2.40.4-r0
 $ apk search cmd:partx
partx-2.40.4-r0
```
multiple binaries are coming from `util-linux-misc` so I'll probably include that inside this image. 
I'll also include `mount` to the list of runtime deps. 

some more storage related binaries I find in upstream ones are `findmnt` comes from `findmnt` apk in wolfi, `lsblk` comes from `lsblk` package in wolfi and probably more which can be seen using `crane export <image> foo.tar` and followed by `tar -tf  foo.tar`

wojciechka: I updated it to only include `mount` and `umount` and validates that this works when the secret is mounted and then unmounted.

wojciechka:
I do not want approve it now as I'm co-author, @kranurag7 PTAL

secrets-store-csi-driver.yaml
kranurag7: should we include `util-linux-misc` which brings `mountpoint` 

maybe that was the reason to include `busybox` here but looks like it's not required doing a quick search on upstream library https://github.com/search?q=repo%3Akubernetes%2Fmount-utils+mountpoint&type=code 💭

developer-guy: `syft registry.k8s.io/csi-secrets-store/driver:v1.4.7 -ojson | jq '.artifacts[] | select(.type == "deb") | {name: .name, version: .version}'
`
this will output all the deb packages included into the image.

EyeCantCU (approved):

kranurag7 (approved):

PR #40519, opened 2025-01-25
libsolv: Fix ftbfs due to split/dev now installing /usr/share/cmake

dannf:
We no longer need to manually install /usr/share/cmake because the split/dev pipeline now does this for us. But since libsolv tries to anyway, it now fails to build. This is fallout from:
  https://github.com/chainguard-dev/melange/pull/1693

Our manual install was moving it into /usr/lib instead of /usr/share. /usr/share is where upstream installs it. /usr/share seems to be an OK choice with split/dev now explicitly supporting it. So let's avoid unnecessarily diverging from upstream and leave it under /usr/share.


murraybd:
Could you also add the ldd-check test pipeline to the libsolv-dev packaage?

```
7cf381a45b54:/work/packages# ldd /usr/lib/libsolv.so
        linux-vdso.so.1 (0x00007ffd10fad000)
        libc.so.6 => /lib/libc.so.6 (0x00007e1f6e2f2000)
        /lib64/ld-linux-x86-64.so.2 (0x00007e1f6e595000)
```

murraybd (approved):

PR #40529, opened 2025-01-26
move xvfb-run to a separate package

kranurag7:
move xvfb-run to a separate package
    
    separate out xvfb-run subpackage into it's own package because it was
    using a separate source and upon new commit the sha validation was
    failing leading to build time errors.
    
    one solution was to stop validation of commit sha and clone using branches.
    
    this moves to a package package and we now poll the branches weekly
    using git backend. bot will keep on bumping it weekly if upstream
    receives any new commit.

### DIFF
--- a/xorg-server.yaml
+++ b/xorg-server.yaml
@@ -1,7 +1,7 @@
 package:
   name: xorg-server
   version: 21.1.15
-  epoch: 0
+  epoch: 1
   description: "X Server"
   copyright:
     - license: SGI-B-2.0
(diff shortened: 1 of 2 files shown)


dannf (commented):
I think splitting this out is reasonable - thanks for doing this! At the same time, I'd also like to do away with having an expected-commit on a development branch. That was only intended to be temporary, and requires human intervention anytime Debian touches this package - which is the big problem I see here.

Debian does seem to be pushing tags now, so IMO, we should follow those. But since they could always backslide, we could take a trust-but-verify approach. What do you think of this?
  https://github.com/dannf/os/tree/xvfb-run-watch

It would of course need to be ported to your split-out xvfb-run.

Other than the occasional version bump PR, the only time this should then require human intervention is if Debian makes changes to this subdirectory and releases them without pushing a tag.

PR #40549, opened 2025-01-26
add a basic docs test, with a few working, tested sample usages

dustinkirkland:
This branch adds a `test/docs` pipeline, and a couple of working, tested usages against a somewhat random set of packages (7zip, apache2, keyutils, and zstd), that all currently pass.

There should be roughly another ~600 docs (sub)packages that we can add these tests against, especially the ones that use the `split/manpages` (~557) and `split/infodir` (~45) pipelines.

HOWEVER -- some of these will fail, and possibly for reasons that we should go and fix.  We have a few dozen docs packages which are totally empty (ie, don't ship any documentation, symlinks, or act as virtual packages depending on other packages).  e.g.: `utmps-docs`, `fio-docs`, `libcmis-doc`, `qt6-qtbase-doc`, etc.  In each of those cases, there are subpackages using split/manpages or split/infodir, but nothing is actually happening.  That seems to be indicative of a real problem in the packaging.


pipelines/test/docs.yaml
```diff
@@ -0,0 +1,63 @@
+name: docs
+
+needs:
+  packages:
+    - man-db
+      texinfo
```
mattmoor, line 6: is a `-` missing here? 🤔

mattmoor: It may be worth including `apk-tools` and `grep` in the dependencies as well?

dustinkirkland: Sure, fixed all of these.  (Note that info was installed, even without the hyphen, but I added it anyway)

mattmoor: Yeah, I assumed as much.  This seems really odd.  tagging @imjasonh @jonjohnsonjr for 👀

pipelines/test/docs.yaml
```diff
@@ -0,0 +1,65 @@
...
+          man -l /"$doc_file" >/dev/null
+          doc_files=true
+        fi
+      done
+      # Test info pages
+      for doc_file in $(apk info -L "$doc_pkg" | grep "^usr/share/info/"); do
+        if [ -f /"$doc_file" ]; then
+          # Ensure that info can read and render
+          # NOTE: info will dutifully print pretty much any file
+          info -f /"$doc_file" -o /dev/null
```
OddBloke, line 44: I've just done some testing locally and it looks like `info` won't _complain_ about being given pretty much any file, but it doesn't seem to write anything to the output file if it isn't a valid input:

```
# info -f /usr/share/info/texinfo.info -o test.txt
# wc -l test.txt
621 test.txt
# cat foo.txt 
lol ol ol ol k lkandw adw a
# info -f /foo.txt -o test.txt
# wc -l test.txt 
0 test.txt
# info -f /var/lib/db/sbom/texinfo-7.2-r0.spdx.json -o test.txt
# wc -l test.txt 
0 test.txt
```
So I think we could potentially do something like:
```suggestion
          # NOTE: info will only write output for valid input
          info -f /"$doc_file" -o /tmp/test.txt
          [ $(wc -l < /tmp/test.txt) != 0 ]
```

This might not catch truncated files or the like, but is a somewhat stronger test.

(I do see some `.png` files in my Ubuntu laptop's `/usr/share/info`, but we can cross that bridge when we come to it.)

OddBloke (dismissed):
Hey Dustin, thanks for cooking this up! Apologies it's taken a few days to get to.

I have a couple of inline comments: one forward-looking, and one which I think would strengthen the `info` testing a little. I'm happy for us to land this and I can PR my proposed `info` change afterwards, or you can take care of it: I'll leave it up to you!

Once landed (including the `info` fix), the plan is for @debasishbsws and I to batch up a few updates whenever we've exhausted our queues: starting small until we get a sense of the pass/fail rate and can expand the batch size. This should make a good impact on the quality of our doc packages, and on our test coverage!

pipelines/test/docs.yaml
```diff
@@ -0,0 +1,65 @@
...
+      # Get our doc package name, which is usually a sub package
+      doc_pkg=$(basename ${{targets.contextdir}})
+      # We seem to have a lot of "empty" doc packages...
+      if [ $(apk info -L "$doc_pkg" | wc -l) -le 3 ]; then
+        echo "See:"
+        apk info -L "$doc_pkg"
+        echo "This package [$doc_pkg] is completely empty (i.e. installs no files)."
+        echo "Please check the package build for proper docs installation, and either:"
+        echo "  (a) fix the docs subpackage build to actually include documentation (check the split/manpages and split/infodir pipelines), or"
+        echo "  (b) remove the test/docs pipeline (if for some reason we want an empty docs package), or"
```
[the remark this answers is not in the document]
dustinkirkland: Sure, I think that's reasonable!  I don't think I know where or how to go about adding that, though...  Show me?

pipelines/test/docs.yaml
dustinkirkland: Nice!  I just pushed a commit inspired by this suggestion that does improve the test.  Thanks!

pipelines/test/docs.yaml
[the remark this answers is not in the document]
OddBloke: I'm not sure either, I haven't delved into pipelines too much yet. This can definitely come later!

OddBloke (approved):

pipelines/test/docs.yaml
```diff
@@ -0,0 +1,63 @@
...
+    - apk-tools
+    - grep
+    - man-db
+    - texinfo
+
+pipeline:
+  - name: docs readability check
+    runs: |
+      # Get our doc package name, which is usually a sub package
+      doc_pkg=$(basename ${{targets.contextdir}})
```
murraybd, line 14: We should be able to replace this with ${{subpkg.name}} which is in melange 19.4.

murraybd (approved):

pipelines/test/docs.yaml
dustinkirkland: Sounds like there's an effort to fix these en masse, with a bunch of others using something similar.

Issue #40550, opened 2025-01-26
postfix fails to build due to multiple linting errors


murraybd:
I was trying to rebuild postfix so that a runtime dependency could be added to a library package so that the ldd-check test pipeline would pass but unfortunately postfix fails to build from source.

```
2025/01/26 16:41:07 WARN + exit 0 subpackage=postfix-sqlite
2025/01/26 16:41:07 INFO retrieving workspace from builder: 2d267b76087c5635c77f25505a4ef81b32add2087fc334dc0edc41975565ba61
2025/01/26 16:41:07 INFO retrieved and wrote post-build workspace to: /home/user/tmp/melange-workspace-1841499973
2025/01/26 16:41:07 INFO running package linters for postfix
2025/01/26 16:41:07 WARN linter "object" failed on package "postfix": open var/spool/postfix/active: permission denied; suggest: This package contains intermediate obj
ect files
linter "opt" failed on package "postfix": open var/spool/postfix/active: permission denied; suggest: This package should be a -compat package
linter "setuidgid" failed on package "postfix": file is setgid; suggest: Unset the setuid/setgid bit on the relevant files, or remove this linter
linter "srv" failed on package "postfix": open var/spool/postfix/active: permission denied; suggest: This package should be a -compat package
linter "strip" failed on package "postfix": open var/spool/postfix/active: permission denied; suggest: Properly strip all binaries in the pipeline
linter "usrlocal" failed on package "postfix": open var/spool/postfix/active: permission denied; suggest: This package should be a -compat package
linter "worldwrite" failed on package "postfix": open var/spool/postfix/active: permission denied; suggest: Change the permissions of any world-writeable files in the 
package, disable the linter, or make this a -compat package
2025/01/26 16:41:07 INFO linting apk: postfix
2025/01/26 16:41:07 INFO pod 2d267b76087c5635c77f25505a4ef81b32add2087fc334dc0edc41975565ba61 terminated
2025/01/26 16:41:07 INFO deleting image apko.local/cache:ad6f457bd28452b2a0eda8fd49d7a16b8bdfd7ab6921740dd05f2f352c38a18e
2025/01/26 16:41:07 INFO untagged apko.local/cache:ad6f457bd28452b2a0eda8fd49d7a16b8bdfd7ab6921740dd05f2f352c38a18e
[...]

PR #40553, opened 2025-01-26
Add the ldd-check test pipeline to some packages

murraybd:
This adds the ldd-check test pipeline to some more packages. mesa requires a rebuild to pick up a dependency, as do both nginx packages. The postfix test is disabled because it fails to build from source - https://github.com/wolfi-dev/os/issues/40550.

### DIFF
--- a/postfix.yaml
+++ b/postfix.yaml
@@ -207,6 +207,9 @@ subpackages:
   - range: _subpackages
     name: postfix-${{range.key}}
     description: "${{range.key}} map support for postfix"
+    dependencies:
+      runtime:
+        - mariadb-dev
     pipeline:
       - runs: |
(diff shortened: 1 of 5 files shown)


postfix.yaml
```diff
@@ -207,7 +207,15 @@ subpackages:
...
     description: "${{range.key}} map support for postfix"
+    dependencies:
+      runtime:
+        - mariadb-dev
     pipeline:
+      # test:
+      #   pipeline:
+      #     - uses: test/ldd-check
+      #       with:
+      #         packages: postfix-${{range.key}}:
```
joshrwolf, line 218: uncomment?

murraybd: The package currently fails to build from source so the test will fail until the package is rebuilt and the runtime dependency is picked up.

dannf (approved):

PR #40563, opened 2025-01-27
rename tensorflow-cpu to tensorflow-cpu-jupyter

kranurag7:
rename tensorflow-cpu to tensorflow-cpu-jupyter
    
    this commit renames tensorflow-cpu package to tensorflow-cpu-jupyter to
    be more explicit that this package provides jupyter support.
    
    this way we're future proof in case we require another tensorflow-cpu
    package.
    
    also withdraws tensorflow-cpu packages from index


maxgio92 (dismissed):
Thank you @kranurag7. I agree

maxgio92 (approved):

PR #40603, opened 2025-01-27
dmidecode: Add test that decodes a saved file

dannf:


smoser (approved):
I like this , thank you.
The only thing I don't _love_ is the binary file in git. But, I don't love my suggestion of base64 encoding the file so that it now "text" any more.

So I'm just going to approve. If you thought a base64 encoded binary file stored in git was any better than a binary file stored in git, then feel free to do that and I'll approve it too.

Thanks.

dannf:
> I like this , thank you. The only thing I don't _love_ is the binary file in git. But, I don't love my suggestion of base64 encoding the file so that it now "text" any more.
> 
> So I'm just going to approve. If you thought a base64 encoded binary file stored in git was any better than a binary file stored in git, then feel free to do that and I'll approve it too.
> 
> Thanks.

Thanks Scott. I don't have an issue with file being binary myself, though I understand why one would. The concern that crossed my mind was setting a precedent of storing test files here, and that could easily end up blowing up the repo size. The idea of a separate repo for this crossed my mind, but seemed overkill for this case. Perhaps we need some kind of external object store - that could be valuable for some of the large files that we use for testing AI things.

PR #40650, opened 2025-01-28
Add the ldd-check test pipeline to more packages

murraybd:
I tested all of these locally and they passed.


tcl.yaml
```diff
@@ -1,10 +1,13 @@
...
   name: tcl
   version: 9.0.1
-  epoch: 0
+  epoch: 1
   description: The Tcl scripting language
   copyright:
     - license: TCL
+  dependencies:
+    runtime:
+      - sqlite-libs
```
smoser, line 10: Unrelated?

If you found this as a result of ldd-check, then you shouldn't need to add it (because melange sca would find it, right?)

murraybd: This was definitely missing.

smoser (approved):
OK, I'll approve this, as it absolutely _does_ fix a missing dependency.
I've opened https://github.com/chainguard-dev/melange/issues/1764 to track why that dependency is not added by melange SCA.

PR #40654, opened 2025-01-28
ignore v3 and v2 branch tags by upstream which seems unstable for mockery

debasishbsws:
avoid picking unstable v3 redundant branch tags.

The upstream has started tagging v2 and v3 branches twice a week with the same tag names. The automation is picking up v3 releases, which are still in alpha, while upstream continues to release stable v2 versions (latest being v2.51.1). https://github.com/vektra/mockery/tags

### DIFF
--- a/mockery.yaml
+++ b/mockery.yaml
@@ -1,7 +1,7 @@
 package:
   name: mockery
-  version: "3"
-  epoch: 6
+  version: "2.51.1"
+  epoch: 0
   description: A mock code autogenerator for Go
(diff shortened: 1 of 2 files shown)


kranurag7 (approved):
thank you, looks good. Verified that we don't have any package for 2.51.1 tag.

withdrawing looks good here given `v3` tag is unstable upstream, here onwards, we only intend to update to {2,3}.x or {2,3}.x.x tags.

PR #40677, opened 2025-01-28
feat(package update and remove manual: true): gitlab-runner-17.8

debasishbsws:
removing the manual:true now so that automation can attempt to update some time which will result in some failures but most should pass. Instead of currently with manual:true where all updates need to be done manually

Package update to 17.8.3

### DIFF
--- a/gitlab-runner-17.8.yaml
+++ b/gitlab-runner-17.8.yaml
@@ -30,8 +30,8 @@ var-transforms:
 package:
   name: gitlab-runner-17.8
   # ---Additional updates required--- Review 'vars' section (above), when reviewing version bumps.
-  version: 17.8.0
-  epoch: 2
+  version: 17.8.3
+  epoch: 0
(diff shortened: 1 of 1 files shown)


gitlab-runner-17.8.yaml
```diff
@@ -135,8 +135,6 @@ subpackages:
 
 update:
   enabled: true
-  # Requires manual steps when updating
-  manual: true
```
OddBloke, line 139: I think these lines should be removed in your follow-up PR: until that lands, I think we will still need to manually verify any new versions that come in.

debasishbsws: I have added the check for this, I will also include this check for all the other package when I will add the entrypoint subpackages

debasishbsws: I think we can remove the manual:true with this check

joshrwolf (approved):

PR #40679, opened 2025-01-28
openssh: add snippets include for sshd as well

xnox:
- **openssh: add test case that snippets incude made it into openssh-client**
  

- **openssh: add snippets include for sshd as well**
  Add include configuration option to support loading configuration
  snippets in sshd_config (server) as well, similar to ssh_config
  (client).
  
  Also add tests that ensure that sed lines did in-fact work.
  
  This brings parity with Debian/Ubuntu systems.

### DIFF
--- a/openssh.yaml
+++ b/openssh.yaml
@@ -1,7 +1,7 @@
 package:
   name: openssh
   version: 9.9_p1
-  epoch: 4
+  epoch: 5
   description: "the OpenBSD SSH implementation"
   copyright:
     - license: ISC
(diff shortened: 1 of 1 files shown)


smoser (commented):
Can we make a placeholder directories for /etc/ssh_config.d and /etc/sshd_config.d ?

dannf (dismissed):

xnox:
> Can we make a placeholder directories for /etc/ssh_config.d and /etc/sshd_config.d ?

added.

@smoser @dannf please re-review

xnox:
LGTM! But I am tainted now.

imjasonh (approved):

smoser (commented):
this is really nice. thank you.

PR #40708, opened 2025-01-28
chore(postgres): set update to manual plus tests to catch entrypoint

mritunjaysharma394:
Follow up to #40671


kranurag7 (approved):
thank you adding the tests.

PR #40827, opened 2025-01-29
Revert "glibc: Re-enable OpenSSF hardening"

dannf:
We're experiencing segfaults when loading certain binaries. Let's revert this while we root cause.

This reverts commit df86be4fa504a6ffc618ff13aff62ac21489cc38.


murraybd (approved):

sil2100 (commented):
LGTM!

PR #40828, opened 2025-01-29
feat: add a default config for systemd-repart

89luca89:
Add a default config for systemd repart

### DIFF
--- a/systemd.yaml
+++ b/systemd.yaml
@@ -1,7 +1,7 @@
 package:
   name: systemd
   version: "257.2"
-  epoch: 10
+  epoch: 11
   description: The systemd System and Service Manager
   copyright:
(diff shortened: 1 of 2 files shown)


smoser (commented):
i wouldn't do this by default. It should be added in cases that it is desired.

89luca89:
> i would _not_ do this "by default" anywhere. this should be added as part of cloud image. (maybe a package for it or somethign... especially if you have to say the filesystem type)

it's indeed a separate subpackage, or you mean something external from here completely?

xnox:
> i wouldn't do this by default. It should be added in cases that it is desired.

even if it is installed in the image, there is a kernel cmdline opt-out from starting this

xnox (approved):

PR #40832, opened 2025-01-29
pipelines: Add compiler/filter-openssf-flags

dannf:
There are some compiler flags set by openssf.spec that can't be trivially removed. `-Wl,-z,now` is one such example, for which a copy and edit hack has been propagating. Let's create a pipeline for this so that we only have to maintain that code in one place.

Use of this should be discouraged if other options are available. But if the decision is between using this or disabling all openssf hardening flags, this is the better choice.

Make `datadog-agent` our first customer.

### DIFF
--- a/datadog-agent.yaml
+++ b/datadog-agent.yaml
@@ -3,7 +3,7 @@ package:
   # This package has two git checkouts. For each new release, the commit SHA for
   # DataDog/integrations-core must also be updated.
   version: 7.60.1
-  epoch: 3
+  epoch: 4
   description: "Collect events and metrics from your hosts that send data to Datadog."
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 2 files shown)


justinvreeland:
This looks neat!

xnox:
most ossf compiler flags have opt-outs => if one passes in FORTIFY_SOURCE=0 it gets honored. The bind now/lazy => is a bug in gcc compiler; there is a patch for it but not getting fixed/merged.

And also it doesn't help with things that fail to pass environment variables (looking at bazel), as in it may ignore GCC_SPEC_FILE.

I do think that bind lazy is common enough; and our customers might need it too in the gcc image (for example to compile datadog).

So yeah, we do need to fix gcc bug; we do need this spec filter; and we may need to provide stock openssf-lazy.spec file as well.

very nice overall!

justinvreeland:
> And also it doesn't help with things that fail to pass environment variables (looking at bazel), as in it may ignore GCC_SPEC_FILE.

Could write out the filtered file to `/etc/openssf-compiler-options` when that comes up. I haven't found anything that bypasses that check yet.

cmwilson21:
Thanks for the contribution!

PR #40844, opened 2025-01-29
gcc, binutils, stress-ng, busybox: no-change rebuild against new glibc

sil2100:
Added some old gcc packages as well. We need this to land after 2.41.


xnox (changes requested):
please rebuild busybox

dannf (dismissed):

xnox (dismissed):

sil2100:
Let's revisit this once we decide 2.41 is ready to land again.

PR #40847, opened 2025-01-29
glibc: no change rebuild

sil2100:
Somehow x86 glibc-2.41 was not published.


xnox (approved):

OddBloke (approved):

xnox (changes requested):
2.41-r0 seems to be published; and i fear building glibc again will fail the world again.

PR #40853, opened 2025-01-29
gcc rebuild with old glibc, if it let us do that

xnox:
Check if gcc bootstrap works with old glibc.


sil2100:
We will return to this once we decide to re-land 2.41.

PR #40855, opened 2025-01-29
Withdraw glibc-2.41-r0

dannf:


xnox (changes requested):
also glibc.yaml should be reverted back. as otherwise it will build again =)

withdrawn-packages.txt
```diff
@@ -1,92 +1,227 @@
...
-mockery-3-r0.apk
-mockery-3-r1.apk
-mockery-3-r2.apk
-mockery-3-r3.apk
-mockery-3-r4.apk
-mockery-3-r5.apk
-mockery-3-r6.apk
-tensorflow-cpu-0_git20250108-r0.apk
-tensorflow-cpu-2.18.0-r0.apk
+glibc-2.41-r0
```
xnox, line 1: file names must end in .apk

dannf:
> also glibc.yaml should be reverted back. as otherwise it will build again =)

got it, thanks for catching!

xnox:
gcc bootstrap agaisnt this older glibc-2.40-r8 is being tested at:
- https://github.com/wolfi-dev/os/pull/40853

to ensure that glibc we are reverting to is considered good.

xnox (approved):

PR #40873, opened 2025-01-30
github: no more digestabot

xnox:
- **github: remove unused action**
  Remove unused docker-run action, last user remove on 5th of December in d133af765da9bb4df57b706040341b17cc0229f8:
  - #35880
  

- **makefile: switch sdk image from digest to pull=always**
  Back in the day, sdk image was used for production builds github
  actions builds, and local builds. Digestabot used to maintain updating
  it in lockstep, such that local dev-container matches production
  github actions.
  
  As builds moved to elastic, the dev-container target has become
  disconnected from the production builds. It is thus best to just
  always pull it rather than rely on automatic digest bumps which are no
  longer critical. These digests have stopped being
  infrastructure-as-code.
  

- **devcontainer: update digest from March 2024 to latest tag**
  Not sure this is in use, as digestabot didn't update this digest, and
  thus this was stuck on March 2024 digest.
  
  Update to just latest tag.
  

- **github: drop digestabot, unused**
  With all other minor developer experience changes, digestabot is no
  longer needed in this repository.


imjasonh (approved):
🔥

sil2100 (approved):
Yep, less busywork at least. And I don't mind pulling new images more often when using dev-container!

PR #40875, opened 2025-01-30
apache2: Add dep on busybox and remove unused file

dannf:
The new entry point is a script that runs `rm`. We had an image fail to test because the `rm` command did not exist, so add a dep on `busybox`.
    
`busybox` also provides a shell, so we can drop the `dash-binsh` dependency. The `dash-binsh` dependency was added along with the previous entrypoint script in commit 8dad232d60bca144b357cb652cb0865d6e351be7, so I conclude it is just there to handle the entrypoint, which busybox's shell handles fine.

### DIFF
--- a/apache2.yaml
+++ b/apache2.yaml
@@ -1,7 +1,7 @@
 package:
   name: apache2
   version: "2.4.63"
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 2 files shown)


sil2100 (approved):
This makes sense, thanks for the explanation!

PR #40887, opened 2025-01-30
Test rebuild gcc against the glibc snapshot to try and determine if g…

sil2100:
This is just for testing.

### DIFF
--- a/gcc.yaml
+++ b/gcc.yaml
@@ -1,7 +1,7 @@
 package:
   name: gcc
   version: 14.2.0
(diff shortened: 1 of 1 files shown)


sil2100:
Using old https://github.com/wolfi-dev/os/pull/39853 closed PR presubmit for this, since it had the same snapshot-ish that I used for the world rebuild.

sil2100:
As seen, it all works. Per other threads: the issue has been identified.

PR #40952, opened 2025-01-30
clamav-docker-1.4: init and entry scripts package

kbsteere:
Docker specific clamav package to add in init scripts

#### For new package PRs only

#### For new version streams


clamav-1.4-docker.yaml
```diff
@@ -0,0 +1,41 @@
...
+  epoch: 0
+  description: An anti-virus toolkit for UNIX eis-ng backport
+  copyright:
+    - license: GPL-2.0-only
+  dependencies:
+    runtime:
+      - clamav-1.4
+      - libstdc++
+      - tini
+      - tzdata
```
kranurag7, line 13: there's more runtime deps https://github.com/Cisco-Talos/clamav-docker/blob/658ba84dfb16f0a70ea6588c5da208e36962e678/clamav/1.4/alpine/Dockerfile#L97-L109 not sure if all of them are required 💭

clamav-1.4-docker.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - ca-certificates-bundle
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/Cisco-Talos/clamav-docker
+      expected-commit: 658ba84dfb16f0a70ea6588c5da208e36962e678
+      tag: ""
+
+  - runs: |
```
kranurag7, line 28: I think we can use a `working-directory` here something like following:

<p>

```yaml
  - working-directory: clamav/1.4/alpine/scripts
    runs: |
      addgroup -S "clamav"
      adduser -D -G "clamav" -h "/var/lib/clamav" -s "/bin/false" -u 100 -S "clamav"

      mkdir -p "${{targets.destdir}}"/usr/bin
      mv docker-entrypoint-unprivileged.sh "${{targets.destdir}}"/usr/bin/init-unprivileged
      mv docker-entrypoint.sh "${{targets.destdir}}"/usr/bin/init
      mv clamdcheck.sh "${{targets.destdir}}"/usr/bin/clamdcheck.sh

      chmod -R 755 "${{targets.destdir}}"/usr/bin/
      chown -R clamav:clamav "${{targets.destdir}}"/usr/bin/
```

</p>
</details>

clamav-1.4-docker.yaml
```diff
@@ -0,0 +1,41 @@
...
+      addgroup -S "clamav"
+      adduser -D -G "clamav" -h "/var/lib/clamav" -s "/bin/false" -u 100 -S "clamav"
+
+      mkdir -p "${{targets.destdir}}"/usr/bin
+      mv /home/build/clamav/1.4/alpine/scripts/docker-entrypoint-unprivileged.sh "${{targets.destdir}}"/usr/bin/init-unprivileged
+      mv /home/build/clamav/1.4/alpine/scripts/docker-entrypoint.sh "${{targets.destdir}}"/usr/bin/init
+      mv /home/build/clamav/1.4/alpine/scripts/clamdcheck.sh "${{targets.destdir}}"/usr/bin/clamdcheck.sh
+
+      chmod -R 755 "${{targets.destdir}}"/usr/bin/
+      chown -R clamav:clamav "${{targets.destdir}}"/usr/bin/
```
kranurag7, line 38: `apk add` will hold `user` and `group` info during install. If we want to `apk` to hold this info then we can use `scriptlets` (ref: https://github.com/wolfi-dev/os/blob/bdb75b0b236710bf47a93fc3bd425cd7069687ba/rpcbind.yaml#L12-L22) but that won't work well with apko so we'll do this at image level.

clamav-1.4-docker.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+update:
+  enabled: false
```
kranurag7, line 41: we can poll it weekly using git backend. 💭 
```suggestion
update:
  enabled: true
  git: {}
  schedule:
    period: weekly
    reason: Upstream does not maintain tags or releases
```

clamav-1.4-docker.yaml
```diff
@@ -0,0 +1,41 @@
...
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates-bundle
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/Cisco-Talos/clamav-docker
```
kranurag7, line 24: we can use main branch and poll it weekly using git backend. 
```bash
      repository: https://github.com/Cisco-Talos/clamav-docker
      branch: main
```

clamav-1.4-docker.yaml
```diff
@@ -0,0 +1,41 @@
+package:
+  name: clamav-1.4-docker
+  version: "1.4.2"
```
kranurag7, line 3: In here, we should consider using the git specific version so that it's tracked on a weekly basis at least. 
```suggestion
  version: 0.0.0_git20250131
```

clamav-1.4-docker.yaml
```diff
@@ -0,0 +1,41 @@
...
+      adduser -D -G "clamav" -h "/var/lib/clamav" -s "/bin/false" -u 100 -S "clamav"
+
+      mkdir -p "${{targets.destdir}}"/usr/bin
+      mv /home/build/clamav/1.4/alpine/scripts/docker-entrypoint-unprivileged.sh "${{targets.destdir}}"/usr/bin/init-unprivileged
+      mv /home/build/clamav/1.4/alpine/scripts/docker-entrypoint.sh "${{targets.destdir}}"/usr/bin/init
+      mv /home/build/clamav/1.4/alpine/scripts/clamdcheck.sh "${{targets.destdir}}"/usr/bin/clamdcheck.sh
+
+      chmod -R 755 "${{targets.destdir}}"/usr/bin/
+      chown -R clamav:clamav "${{targets.destdir}}"/usr/bin/
+
```
kranurag7, line 39: we can consider adding some tests in here. 
```bash
test:
  pipeline:
    - runs: |
        stat /usr/bin/init-privileged
        stat /usr/bin/init
        stat /usr/bin/clamdcheck.sh
```
stat tests to start with but can run the scripts as well by adding the required dependencies.

kranurag7 (commented):
great work!! requesting some change inline.

clamav-1.4-docker.yaml
```diff
@@ -0,0 +1,41 @@
+package:
+  name: clamav-1.4-docker
```
mamccorm, line 2: If this was a version streamed application, we would use `clamav-docker-1.4` (version at the end), but couple of questions here:
 1. The repo isn't version streamed: https://github.com/Cisco-Talos/clamav-docker
 2. How would this be maintained by sustaining? it doesn't look to have any tags or releases, only a main commit. Whats the relationship between the application and the scripts? do they pin to a commit, or do their builds simply just download the latest all the time?

clamav-1.4-docker.yaml
mamccorm: Why are we adding runtime dependencies here if this is just a package to fetch / copy install scripts? wouldn't those be defined in the clamav package, and the clamav package would just add this package to install the scripts? (I'm guessing thats the aim here)

clamav-1.4-docker.yaml
mamccorm: Need some more info here on the relationship between clamav and the scripts - do they pin to git commit sha, or do they always fetch the latest from this repo? is there anywhere we can take a peak?

mamccorm (changes requested):
Thanks for looking into this - i'm not fully clear on this approach, but left some inline comments

clamav-1.4-docker.yaml
kbsteere: My assumption was that if I used clama-1.4 as a runtime dependency then it would pull in all of it's required runtime dependencies. The other dependencies are required based on the upstream maintainers. Correct me if my assumption is wrong here.

clamav-1.4-docker.yaml
kbsteere: Good point, I'll make this change

clamav-1.4-docker.yaml
kbsteere: My other thought was to deal with it at the image level so I'll do it there instead and remove the user and group stuff

clamav-1.4-docker.yaml
kbsteere: There are no versions for the clamav-docker repo, only for the source repo of clamav. Don't know if that version your recommending is coming from some where or if it's just and example

clamav-1.4-docker.yaml
kbsteere: Agreed, I wanted some eyes on this first to make sure it was the correct direction for this. I'll add tests now.

clamav-1.4-docker.yaml
kbsteere: So yeah this is the issue that was discussed here in slack thread. The clamav-docker repo appears to only track in the mainline and then occasionally in creates a branch randomly with some tag specifying versions. I could ask for an upstream change but don't know if they would use it. 

Basically they create commits that remove specific versions. Example, removes the [1.3 directory](https://github.com/Cisco-Talos/clamav-docker/tree/3e6108fb1c24525d2064a27559fb9dceffbdf4e1). I haven't compared the scripts between the different versions in the alpine directory but they appear to be identical ATM.

clamav-1.4-docker.yaml
kbsteere: The only other way I was attempting to do this would be to fetch the specific script file directly and then get the has with it. Only issue I ran into with that is the sha256 kept changing for the script files themselves which was rather odd. But that's the reason I abandoned that method.

clamav-1.4-docker.yaml
kbsteere: Also my thought has been the underlying package is version streamed. So that means that this also has to be.

jamonation (dismissed):

kbsteere:
@mamccorm need your review

clamav-docker-1.4.yaml
```diff
@@ -0,0 +1,97 @@
...
+  - uses: fetch
+    with:
+      expected-sha256: f99ed46fb8cd72f899709b652276944ebc5c8d0132890db35154e6dae365814e
+      uri: https://raw.githubusercontent.com/Cisco-Talos/clamav-docker/refs/heads/main/clamav/1.4/alpine/scripts/docker-entrypoint-unprivileged.sh
+      extract: false
+  - uses: fetch
+    with:
+      expected-sha256: 4034f6d63ee6c1d1ed3686733b5722f4b19055b623c82843927613f4e2f7c641
+      uri: https://raw.githubusercontent.com/Cisco-Talos/clamav-docker/refs/heads/main/clamav/1.4/alpine/scripts/docker-entrypoint.sh
+      extract: false
```
Dentrax, line 43: I've read the comment above but not still sure if i'm following what's going on here.

Create a transform above:

```
var-transforms:
  - from: ${{package.version}}
    match: ^(\d+\.\d+)\.\d+$
    replace: "$1"
    to: major-minor-version
```

Then use working dir:

```
- working-directory: clamav/${{vars.major-minor-version}}/alpine/scripts
  runs: ...
```

clamav-docker-1.4.yaml
```diff
@@ -0,0 +1,97 @@
...
+  - runs: |
+      mkdir -p "${{targets.destdir}}"/usr/bin
+      mv docker-entrypoint-unprivileged.sh "${{targets.destdir}}"/usr/bin/init-unprivileged
+      mv docker-entrypoint.sh "${{targets.destdir}}"/usr/bin/init
+      mv clamdcheck.sh "${{targets.destdir}}"/usr/bin/clamdcheck.sh
+      chmod -R 755 "${{targets.destdir}}"/usr/bin/
```
Dentrax, line 49:
```suggestion

```

clamav-docker-1.4.yaml
```diff
@@ -0,0 +1,97 @@
...
+  - uses: fetch
+    with:
+      expected-sha256: 4034f6d63ee6c1d1ed3686733b5722f4b19055b623c82843927613f4e2f7c641
+      uri: https://raw.githubusercontent.com/Cisco-Talos/clamav-docker/refs/heads/main/clamav/1.4/alpine/scripts/docker-entrypoint.sh
+      extract: false
+  - runs: |
+      mkdir -p "${{targets.destdir}}"/usr/bin
+      mv docker-entrypoint-unprivileged.sh "${{targets.destdir}}"/usr/bin/init-unprivileged
+      mv docker-entrypoint.sh "${{targets.destdir}}"/usr/bin/init
+      mv clamdcheck.sh "${{targets.destdir}}"/usr/bin/clamdcheck.sh
```
Dentrax, line 48:
```suggestion
      install -Dm755 docker-entrypoint-unprivileged.sh "${{targets.destdir}}"/usr/bin/init-unprivileged
      install -Dm755 docker-entrypoint.sh "${{targets.destdir}}"/usr/bin/init
      install -Dm755 clamdcheck.sh "${{targets.destdir}}"/usr/bin/clamdcheck.sh
```

clamav-docker-1.4.yaml
```diff
@@ -0,0 +1,97 @@
...
+      groups:
+        - groupname: clamav
+          gid: 100
+      users:
+        - username: clamav
+          gid: 100
+          uid: 100
+      run-as: 0
+    environment:
+      ENV TINI_SUBREAPER: 1
```
Dentrax, line 84: There is ` ` white space in the key, probably wrong?

clamav-docker-1.4.yaml
```diff
@@ -0,0 +1,97 @@
...
+    - name: "Failure test: clamdcheck.sh"
+      runs: |
+        bash /usr/bin/clamdcheck.sh && exit 1;
+    # this test will fail locally with run-as set to 0, but pass pipeline.
+    # switch to run-as: 100 if running locally.
+    - name: "Init test"
+      runs: |
+        sed -i 's/^#TCPSocket 3310/TCPSocket 3310/' /etc/clamav/clamd.conf
+        sed -i 's/^#TCPAddr localhost/TCPAddr localhost/' /etc/clamav/clamd.conf
+        sed -i '/^set -eu/a sleep 20' /usr/bin/clamdcheck.sh
```
Dentrax, line 95: If we need to do some `sed` stuff here, should we have a compat subpackage here that will tackle this? I was wondering the image-side about what to do while using `clamdcheck.sh`.

clamav-docker-1.4.yaml
```diff
@@ -0,0 +1,97 @@
...
+        - username: clamav
+          gid: 100
+          uid: 100
+      run-as: 0
+    environment:
+      ENV TINI_SUBREAPER: 1
+  pipeline:
+    - name: "Failure test: clamdcheck.sh"
+      runs: |
+        bash /usr/bin/clamdcheck.sh && exit 1;
```
Dentrax, line 88: Can we `grep` what the expected failure is?

clamav-docker-1.4.yaml
```diff
@@ -0,0 +1,97 @@
...
+# upstream also doesn't create tags to track any of these changes, they are tracked mostly in mainline but
+# occasionally tracked on their own branches.
+# If version 1.4 is removed or the has is changed then this build will fail requring manual intervention.
+pipeline:
+  # this is only used for tracking the clamav-docker repo
+  - uses: git-checkout
+    with:
+      repository: https://github.com/Cisco-Talos/clamav-docker
+      expected-commit: 658ba84dfb16f0a70ea6588c5da208e36962e678
+      tag: ""
```
Dentrax, line 27: This should be empty by default.
```suggestion

```

Dentrax (changes requested):

clamav-docker-1.4.yaml
```diff
@@ -0,0 +1,97 @@
...
+      expected-sha256: 4034f6d63ee6c1d1ed3686733b5722f4b19055b623c82843927613f4e2f7c641
+      uri: https://raw.githubusercontent.com/Cisco-Talos/clamav-docker/refs/heads/main/clamav/1.4/alpine/scripts/docker-entrypoint.sh
+      extract: false
+  - runs: |
+      mkdir -p "${{targets.destdir}}"/usr/bin
+      mv docker-entrypoint-unprivileged.sh "${{targets.destdir}}"/usr/bin/init-unprivileged
+      mv docker-entrypoint.sh "${{targets.destdir}}"/usr/bin/init
+      mv clamdcheck.sh "${{targets.destdir}}"/usr/bin/clamdcheck.sh
+      chmod -R 755 "${{targets.destdir}}"/usr/bin/
+
```
Dentrax, line 50: We may need to have compat subpackage to symlink some files: https://github.com/Cisco-Talos/clamav-docker/blob/658ba84dfb16f0a70ea6588c5da208e36962e678/clamav/1.4/debian/Dockerfile#L128-L130

clamav-docker-1.4.yaml
```diff
@@ -0,0 +1,97 @@
...
+      expected-sha256: f99ed46fb8cd72f899709b652276944ebc5c8d0132890db35154e6dae365814e
+      uri: https://raw.githubusercontent.com/Cisco-Talos/clamav-docker/refs/heads/main/clamav/1.4/alpine/scripts/docker-entrypoint-unprivileged.sh
+      extract: false
+  - uses: fetch
+    with:
+      expected-sha256: 4034f6d63ee6c1d1ed3686733b5722f4b19055b623c82843927613f4e2f7c641
+      uri: https://raw.githubusercontent.com/Cisco-Talos/clamav-docker/refs/heads/main/clamav/1.4/alpine/scripts/docker-entrypoint.sh
+      extract: false
+  - runs: |
+      mkdir -p "${{targets.destdir}}"/usr/bin
```
Dentrax, line 45: Let's also create `"${{targets.destdir}}"/var/lib/clamav` and `"${{targets.destdir}}"/var/log/clamav` as [they did already](https://github.com/Cisco-Talos/clamav-docker/blob/658ba84dfb16f0a70ea6588c5da208e36962e678/clamav/1.4/debian/Dockerfile#L123C48-L123C63)

clamav-docker-1.4.yaml
```diff
@@ -0,0 +1,97 @@
...
+        - clamav-daemon
+        - clamav-db
+        - clamav-doc
+        - clamav-libunrar
+        - clamav-milter
+        - freshclam
+        - libstdc++
+        - netcat-openbsd
+        - tini
+        - tzdata
```
Dentrax, line 73: Can we move those to runtime deps? As its included in the FINAL layer of the container image and expected by the entrypoint scripts? https://github.com/Cisco-Talos/clamav-docker/blob/658ba84dfb16f0a70ea6588c5da208e36962e678/clamav/1.4/debian/Dockerfile#L106-L117

Dentrax (changes requested):

Dentrax:
Thanks for working on this @kbsteere, I just checked the Dockerfile and left a few comments with suggestions. Apologies if my comments are making you go through this again!

clamav-docker-1.4.yaml
Dentrax: Please ensure we have each needed runtime deps here.

clamav-docker-1.4.yaml
Dentrax: And also `"${{targets.destdir}}"/run/clamav` and `"${{targets.destdir}}"/tmp` maybe? https://github.com/Cisco-Talos/clamav-docker/blob/658ba84dfb16f0a70ea6588c5da208e36962e678/clamav/1.4/alpine/scripts/docker-entrypoint.sh#L79-L85

clamav-docker-1.4.yaml
kbsteere: The linter throws an error if tag isn't present. Might be able to leak it empty but I believe the field is required.

clamav-docker-1.4.yaml
kbsteere: Yeah this makes sense. Should switch to this method.

clamav-docker-1.4.yaml
kbsteere: On install these scripts should end up in /usr/bin so I don't believe we need any symbolic links. During testing it didn't have any issues.

clamav-docker-1.4.yaml
kbsteere: This was discussed separately and was told we should be putting the runtime deps in the image itself and not included in the package. My original commit had these as runtime deps.

clamav-docker-1.4.yaml
kbsteere: should be removed

clamav-docker-1.4.yaml
kbsteere: yes good idea.

clamav-docker-1.4.yaml
Dentrax: Thats correct, but their Helm Charts, deployment manifest, etc. may use those endpoints, so we should provide those files as symlinks to make compat image.

PR #41021, opened 2025-01-31
Add a missing dependency and the ldd-check

murraybd:
This fixes an issue with jwt-tool's own tests by adding a dependency:

```
2025/01/31 13:24:29 WARN + jwt-tool --help
2025/01/31 13:24:29 INFO WARNING: Cryptodome libraries not imported - these are needed for asymmetric crypto signing and verifying
2025/01/31 13:24:29 INFO On most Linux systems you can run the following command to install:
2025/01/31 13:24:29 INFO python3 -m pip install pycryptodomex
```

It also adds the ldd-check test which passes of course.


sil2100 (approved):
Thanks!

PR #41022, opened 2025-01-31
Running an unnecessary test is not a failure

murraybd:
The pkgconf test pipeline currently exits 1 if the test is not necessary. The test not being needed is certainly not a failure and we shouldn't block PRs for running a test that is unnecessary.

This also makes the pkgconf test pipeline consistent with the ldd-check one which doesn't exit 1 if the test is not needed.


sil2100 (approved):
Okay. I'm all happy if there's consistency - regardless which way. Thanks!

PR #41033, opened 2025-01-31
gcc: remove some unneeded include-fixed headers

sil2100:
This was causing the 2.41 glibc to FTBFS. Not sure why pthread.h fixincludes is still included - others seem to have been dropped. Anyway, right now it's so outdated re: glibc that it's causing issues. We certainly don't need it.


sil2100:
Let's not merge it before I run a test rebuild against it + glibc.

sil2100:
FYI the tests were failing due to being run before we disabled the extra hardening tests (which now fail for gcc).

dannf (approved):

PR #41047, opened 2025-01-31
openssf-compiler-options: Allow -fno-omit-frame-pointer override

dannf:


xnox (dismissed):

xnox:
Please add a test case.

Even if it means doing "uses: git-checkout; uses: autotools/configure; ..." of the offending project.

Because our pipelines are usable in test section :-)

xnox (approved):
I mean this is obvious and thus should just go in.

PR #41058, opened 2025-01-31
ruby3.4-sinatra: Update deps

justinvreeland:
Because we don't have basic require tests for ruby packages we somtimes miss changed deps during updates. This is a good example. rack-2.2 is no longer needed but was never updated. This goes through fixes up the sinatra deps as well as adding a test pipeline.


xnox (approved):
Looks ruby

PR #41070, opened 2025-01-31
systemd: split off userdb

justinvreeland:


systemd.yaml
```diff
@@ -478,11 +478,30 @@ subpackages:
...
 
+  - name: "systemd-userdb"
+    description: "dynamic user/group manager"
+    dependencies:
+      runtime:
+        - ${{package.name}}
+    pipeline:
+      - runs: |
+          cd ${{targets.destdir}}/
+          mkdir -p ${{targets.subpkgdir}}/etc/systemd
```
xnox, line 495: Does this also move the /etc/ssh/sshd_config.d/10-userdb-something-file?

Cannot check on my phone.

xnox (changes requested):

systemd.yaml
justinvreeland: Yes full list for you:
```
etc/ssh/sshd_config.d/20-systemd-userdb.conf
usr/bin/userdbctl
usr/lib/systemd/sshd_config.d/20-systemd-userdb.conf
usr/lib/systemd/system/systemd-userdbd.service
usr/lib/systemd/system/systemd-userdbd.socket
usr/lib/systemd/systemd-userdbd
usr/lib/tmpfiles.d/20-systemd-userdb.conf
```

justinvreeland: Do we want to keep the userdb config file. It does look like Debian doesn't split it out.

justinvreeland: Seems to rely on the command to function though:
```
5   │ AuthorizedKeysCommand /usr/bin/userdbctl ssh-authorized-keys %u
```

xnox: But that's the bit that conflicts for me with other software that also wants to install authorized keys command.

And unlike systemd, in sshd it is not an accumulated value, the first command wins and other configs of the key are ignored :-(

xnox (approved):

PR #41193, opened 2025-02-03
python-3.13: add CVE-2025-0938 fixes

ajayk:
https://mail.python.org/archives/list/security-announce@python.org/thread/K4EUG6EKV6JYFIC24BASYOZS4M5XOQIB/
https://github.com/python/cpython/pull/129526

### DIFF
--- a/python-3.13.yaml
+++ b/python-3.13.yaml
@@ -1,7 +1,7 @@
 package:
   name: python-3.13
   version: 3.13.1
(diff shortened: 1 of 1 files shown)


smoser (approved):
Thanks.
top level python issue https://github.com/python/cpython/issues/105704
python 3.13 pr landed https://github.com/python/cpython/pull/129526

PR #41207, opened 2025-02-03
python-3.12: add CVE-2025-0938 fix

techalchemy:
https://mail.python.org/archives/list/security-announce@python.org/thread/K4EUG6EKV6JYFIC24BASYOZS4M5XOQIB/ python/cpython#129526

### DIFF
--- a/python-3.12.yaml
+++ b/python-3.12.yaml
@@ -1,7 +1,7 @@
 package:
   name: python-3.12
   version: 3.12.8
-  epoch: 2
+  epoch: 3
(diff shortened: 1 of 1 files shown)


smoser (approved):
upstream 3.12 pr is https://github.com/python/cpython/pull/129527
That PR says:
> merged commit [a7084f6](https://github.com/python/cpython/commit/a7084f6075c9595ba60119ce8c62f1496f50c568) into python:3.12 


The commit referenced is the same as listed here.
So :+1: from me.

PR #41322, opened 2025-02-04
sudo-rs: Include /etc/sudoers.d by default

justinvreeland:


roaksoax (dismissed):
lgtm!

sudo-rs.yaml
```diff
@@ -48,7 +48,10 @@ pipeline:
       EOF
       cp ${{targets.destdir}}/etc/pam.d/sudo ${{targets.destdir}}/etc/pam.d/sudo-i
       chmod 644 ${{targets.destdir}}/etc/pam.d/sudo*
-      echo "root ALL = (ALL:ALL) NOPASSWD:ALL" > ${{targets.destdir}}/etc/sudoers
+      cat <<EOF >${{targets.destdir}}/etc/sudoers
+      root ALL = (ALL:ALL) NOPASSWD:ALL
+      @includedir /etc/sudoers.d
+      EOF
```
xnox, line 54: I think @smoser here would say => can you please create and ship that empty /etc/sudoers.d directory?

note we also make ssh ship the ssh_config.d and sshd_config.d directories.

because in general it is pain to bind mount stuff there if it read-only and doesn't already exists, and one is forced to bindmount all of /etc in that case.

xnox (approved):

Issue #41300, opened 2025-02-04
freerdp-libs fails ldd-check test pipeline


murraybd:
The package freerdp will fail the ldd-check test pipeline with the following error:

```
2025/02/03 18:59:38 INFO FAIL[ldd-check]: /usr/lib/freerdp2/liburbdrc-client-libusb.so: missing liburbdrc-client.so
2025/02/03 18:59:38 INFO FAIL[ldd-check]: /usr/lib/freerdp2/libvideo-client.so: missing librdpgfx-client.so
2025/02/03 18:59:38 INFO INFO[ldd-check]: tested 38 files with ldd. 36 passes. 2 fails.
```

murraybd:
This specific failure makes me think something is wrong with ldd or the test itself as they are clearly there:

```
2025/02/04 13:27:42 INFO PASS[ldd-check]: /usr/lib/freerdp2/libainput-client.so
2025/02/04 13:27:42 INFO PASS[ldd-check]: /usr/lib/freerdp2/libaudin-client-alsa.so
2025/02/04 13:27:42 INFO PASS[ldd-check]: /usr/lib/freerdp2/libaudin-client-oss.so
2025/02/04 13:27:42 INFO PASS[ldd-check]: /usr/lib/freerdp2/libaudin-client.so
2025/02/04 13:27:42 INFO PASS[ldd-check]: /usr/lib/freerdp2/libdisp-client.so
2025/02/04 13:27:42 INFO PASS[ldd-check]: /usr/lib/freerdp2/libdrive-client.so
2025/02/04 13:27:42 INFO PASS[ldd-check]: /usr/lib/freerdp2/libecho-client.so
2025/02/04 13:27:42 INFO PASS[ldd-check]: /usr/lib/freerdp2/libgeometry-client.so
2025/02/04 13:27:42 INFO PASS[ldd-check]: /usr/lib/freerdp2/libparallel-client.so
2025/02/04 13:27:42 INFO PASS[ldd-check]: /usr/lib/freerdp2/libprinter-client-cups.so
2025/02/04 13:27:42 INFO PASS[ldd-check]: /usr/lib/freerdp2/libprinter-client.so
2025/02/04 13:27:42 INFO PASS[ldd-check]: /usr/lib/freerdp2/librdp2tcp-client.so
2025/02/04 13:27:42 INFO PASS[ldd-check]: /usr/lib/freerdp2/librdpei-client.so
[...]

PR #41324, opened 2025-02-04
Symlink some shared objects for ldd

murraybd:
As documented in https://github.com/wolfi-dev/os/issues/41300 the ldd-check test pipeline, which just uses ldd, would fail for freerdp because these two objects were not found.

### DIFF
--- a/freerdp.yaml
+++ b/freerdp.yaml
@@ -1,7 +1,7 @@
 package:
   name: freerdp
   version: 2.11.7
-  epoch: 4
+  epoch: 5
   description: FreeRDP client
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


dannf (commented):
Based on nothing but what I see here, it seems like the this package has chosen to use private, unversioned shared libraries. Assuming the thing that is linking against them works, I assume it is internally setting LD_LIBRARY_PATH or similar to access them? In which case, I'd suggest we do the same in the test environment instead of exposing them in /usr/lib.

It could also be that our install commands are broken, and these should've gone into /usr/lib instead of /usr/lib/freerdp2 in the first place - is that what other distros do?

dannf (approved):
This will make an otherwise-private library public, which I don't think we want to do. Presumably, whatever is using these libraries is able to find them in the private location already. For the purposes of the test, I think we should just set `LD_LIBRARY_PATH: /usr/lib/freerdp2` in the test environment block. That should allow `ldd-check` to find them for the purposes of the test.

sergiodj:
I intend to revisit #43370 now that the `ld.so.conf.d` changes have been merged.  It should provide another way to modify library search paths.

PR #41325, opened 2025-02-04
Add the test/docs pipeline to everything that passes

murraybd:
I tested these all on a workstation and the tests passed for every package.

### DIFF
--- a/ack.yaml
+++ b/ack.yaml
@@ -39,6 +39,9 @@ subpackages:
     pipeline:
       - uses: split/manpages
     description: ack manpages
+    test:
+      pipeline:
+        - uses: test/docs
 
 update:
   enabled: true
--- a/acl.yaml
+++ b/acl.yaml
@@ -73,6 +73,9 @@ subpackages:
     description: acl docs
     pipeline:
       - uses: split/manpages
+    test:
+      pipeline:
+        - uses: test/docs
 
 update:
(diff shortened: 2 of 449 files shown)


dustinkirkland:
Nice!!

debasishbsws:
Since there is no epoch bump in any package, the test has not been triggered in CI. Our initial plan was to batch update these changes along with an epoch bump.
@murraybd, if you confirm that everything looks good, merging this should also work.
@OddBloke, any thoughts on this?

OddBloke (approved):

sil2100 (approved):

smoser:
this regressed https://github.com/wolfi-dev/os/issues/43212 .

smoser:
> > Since there is no epoch bump in any package, the test has not been triggered in CI. Our initial plan was to batch update these changes along with an epoch bump. @murraybd, if you confirm that everything looks good, merging this should also work. @OddBloke, any thoughts on this?
> 
> We had an internal discussion and decided this was an appropriate path forward: @murraybd has only proposed changes which already pass in local Docker runner testing, so this shouldn't result in any regressions once landed.

It feels to me like we should not be able to make changes that don't have c-i run on them.
That leaves at least 2 options
1. bump epoch is required, you build and test.
2. fix c-i to run test any time a package changes.
3.

PR #41342, opened 2025-02-05
Icu package update

debasishbsws:
This require handling it differently even though the ABI check pass. there is a ABI diffrence in this version and result in fail of downstream packages.

### DIFF
--- a/R.yaml
+++ b/R.yaml
@@ -2,7 +2,7 @@
 package:
   name: R
   version: 4.4.2
-  epoch: 0
+  epoch: 1
   description: Language and environment for statistical computing
   copyright:
     - license: ( GPL-2.0-only OR GPL-3.0-only ) AND LGPL-2.1-or-later
--- a/apache-arrow.yaml
+++ b/apache-arrow.yaml
@@ -1,7 +1,7 @@
 package:
   name: apache-arrow
   version: "19.0.0"
-  epoch: 1
+  epoch: 2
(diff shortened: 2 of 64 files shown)


xnox (dismissed):
@debasishbsws @imjasonh @jonjohnsonjr @ajayk 

Please correct me if anything below is wrong:

Bumping all the things in a single PR does not guarantee that the new icu will be used to compile these dependants.

One most likely needs to land this separately - as in land icu in one PR into main; wait for it to build and publish; then land pr(s) with lots of wolfictl bumps. However such PRs will not do testing of this.

To pre-test such rebuilds, you can use staging to do rebuild with the icu rebuild draft PR. But there is another option with pure github:
* open icu pr (1), keep it draft wait for it to build
* open a new pr with wolfictl bumps AND also add the wolfi-presubmit repository from (1) above as build_repositories in the environment of all of these files

Such PR is not ok to merge into main; but it will correctly retest all builds against the new ICU abi; and will be summarised with build/test failures.

EDIT: wait, I think I am wrong. Because looking at successful builds:
https://apk.chaindag.dev/https/apk.cgr.dev/wolfi-presubmit/9818cc3e9849f05c67c8ab2bce6b32cc92faec1d/aarch64/libarrow-19.0.0-r2.apk@etag:a4012c49af1ceed2275030aa50a50ce0/.PKGINFO

the rebuild of apache did pick up the new icu 76 abi.

PR #41348, opened 2025-02-05
feat: Validate vars correctness in GitLab CNG package

debasishbsws:
- Added a validation pipeline to check if all variable tags (`registry-tag`, `indexer-tag`, `exporter-tag`, `logger-tag`, `mailroom-tag`, and `shell-tag`) match the expected upstream versions.
- Ensures that the automated package update fails if the vars do not align with upstream values, preventing mismatches and inconsistencies.
- Bumped epoch to `1` to reflect this change in package updates.


OddBloke (approved):
LGTM, thank you!

The CVE check failure is due to a `yamux` CVE which is pending an upstream fix. This does not represent a regression against the packages already in repo.

PR #41381, opened 2025-02-05
fix(python): Support binding to ports less than 1024

EyeCantCU:
Introduces subpackage with capabilities set to bind to privileged ports

Currently needed by pgAdmin

### DIFF
--- a/python-3.12.yaml
+++ b/python-3.12.yaml
@@ -1,7 +1,7 @@
 package:
   name: python-3.12
   version: "3.12.9"
-  epoch: 1
+  epoch: 2
   description: "the Python programming language"
   copyright:
     - license: PSF-2.0
@@ -24,6 +24,7 @@ environment:
       - ca-certificates-bundle
       - expat-dev
       - gdbm-dev
+      - libcap-utils
       - libffi-dev
       - libx11-dev
       - linux-headers
@@ -40,12 +41,12 @@ environment:
 # creates helpfull python3.M and 3.M variables
 var-transforms:
   - from: ${{package.name}}
-    match: '-'
-    replace: ''
+    match: "-"
+    replace: ""
     to: python
   - from: ${{package.version}}
     match: (\d).(\d+).(\d+)
-    replace: '$1.$2'
+    replace: "$1.$2"
     to: pyversion
 
 pipeline:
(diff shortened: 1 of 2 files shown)


Dentrax (dismissed):

python-3.12.yaml
```diff
@@ -195,6 +196,33 @@ subpackages:
...
+          cp "${{targets.outdir}}/${{package.name}}-base/usr/bin/${{vars.python}}" "${{targets.contextdir}}/usr/bin/"
+          setcap cap_net_bind_service=+eip "${{targets.contextdir}}/usr/bin/${{vars.python}}"
+    test:
+      environment:
+        contents:
+          packages:
+            - libcap-utils
+      pipeline:
+        - runs: |
+            getcap "/usr/bin/${{vars.python}}" | grep "cap_net_bind_service=eip"
```
maxgio92, line 224: Love this @EyeCantCU

kranurag7 (approved):
thank you RJ, do we want to the changes around var-transforms of using `""` inside python-3.13 as well 💭

python-3.12.yaml
```diff
@@ -195,6 +196,29 @@ subpackages:
...
+        - ${{package.name}}-base=${{package.full-version}}
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.contextdir}}/usr/bin"
+          cp "${{targets.outdir}}/${{package.name}}-base/usr/bin/${{vars.python}}" "${{targets.contextdir}}/usr/bin/"
+          setcap cap_net_bind_service=+eip "${{targets.contextdir}}/usr/bin/${{vars.python}}"
+    test:
+      pipeline:
+        - runs: |
+            stat "/usr/bin/${{vars.python}}"
```
maxgio92, line 220: Notes for the future. We can't test with `getcap` here because file capabilities are not persisted in the Melange's test pipeline container rootfs.

python-3.13.yaml
```diff
@@ -195,6 +196,29 @@ subpackages:
...
+        - ${{package.name}}-base=${{package.full-version}}
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.contextdir}}/usr/bin"
+          cp "${{targets.outdir}}/${{package.name}}-base/usr/bin/${{vars.python}}" "${{targets.contextdir}}/usr/bin/"
+          setcap cap_net_bind_service=+eip "${{targets.contextdir}}/usr/bin/${{vars.python}}"
+    test:
+      pipeline:
+        - runs: |
+            stat /usr/bin/${{vars.python}}
```
maxgio92, line 220: Notes for the future. We can't test with `getcap` here because file capabilities are not persisted in the Melange's test pipeline container rootfs.

maxgio92 (approved):

PR #41389, opened 2025-02-05
withdraw package prometheus-jmx-exporter-httpserver

Aditevil:
Withdraw package prometheus-jmx-exporter-httpserver:
1. There is no manifest for it.
2. [REF commit](https://github.com/wolfi-dev/os/commit/b7328e0c6685f8366516234bb5f2ff447fa3e1c3) as it is no longer needed as a separate apk.
3. [prometheus-jmx-exporter ](https://github.com/wolfi-dev/os/blob/main/prometheus-jmx-exporter.yaml#L49)still uses it in tests so it's confusing for future travellers.


Aditvil-Dev:
Closing in favor of https://github.com/wolfi-dev/os/pull/41466

PR #41405, opened 2025-02-05
Introduce libcurl-abi virtual package

dannf:
Use this to allow only `libcurl` libraries built from the same upstream source version instead of only relying on SONAMEs. `curl` may pick up new symbols from a minor libcurl bump.
    
https://github.com/chainguard-dev/internal-dev/issues/10381


xnox (dismissed):
This would break using alternative RustTLS library provider.

xnox:
I don't know how conflicts are processed, and if something like

```
- !curl>=8.13 # where 8.13 is package.version + 1
```

would work.

dannf:
> This would break using alternative RustTLS library provider.

Good catch. Could we just build them both from the same YAML? They appear to originate from the same source tree/tag.

xnox:
> > This would break using alternative RustTLS library provider.
> 
> Good catch. Could we just build them both from the same YAML? They appear to originate from the same source tree/tag.

no, as it will cause bootstrap cycle. Also they sometimes are broken with new versions of rust/rustls/rustls-ffi/curl.

jamonation:
https://github.com/wolfi-dev/os/compare/main...jamonation:wolfi-os:libcurl-depends-provides is my approach adding the virtual `libcurl` provides, and forcing package.full-version depends across both curl/curl-rustls

sergiodj:
I'm taking a look at this one.  I'd like to build both packages and experiment with adding/removing different versions of curl to see what happens.  But overall, this looks like it's the right approach.

My workstation is busy with other stuff now, but I'll get to this PR ASAP.

sergiodj:
OK, I ran some tests here.

I built `curl` and `curl-rustls` from this PR.  Then I built a new `curl` 8.13.0 package, and tried to do some install/upgrade dance.  The following works as expected:

- When `libcurl-openssl4-8.12.1-r1` is installed, `curl-8.12.1-r1` is also installed.
- When we try `apk add curl` in this case, nothing happens.
- When we try `apk add curl=8.13.0-r0`, `libcurl-openssl4-8.13.0-r0` is updated alongside it.

Running `apk add curl=8.12.1-r1` (effectively downgrading the package) doesn't do the expected thing: `libcurl-openssl4-8.13.0-r0` is kept installed.  But I don't think we support downgrades anyway.

I'm approving this PR based on my tests and on previous discussions with Dann.  Thanks a lot!

sergiodj (approved):

dannf:
Verify that an `apk add curl` does not pull in the updated curl, but rather the current wolfi `curl`:
```
user@dann-frazier:~$ docker run --user root --rm -it cgr.dev/chainguard-private/glibc:latest-dev /bin/bash
Unable to find image 'cgr.dev/chainguard-private/glibc:latest-dev' locally
latest-dev: Pulling from chainguard-private/glibc
973305f4ac8b: Pull complete 
Digest: sha256:ebb71668c912b0830a85e2416ce1b5c56e6a01fd50187b09c987c43f675bb313
Status: Downloaded newer image for cgr.dev/chainguard-private/glibc:latest-dev
bash-5.2# echo https://apk.cgr.dev/wolfi-presubmit/81c874a44d3cf5f8dddcc8ddaaa305b5e781d608 >> /etc/apk/repositories 
bash-5.2# apk update
fetch https://packages.cgr.dev/extras/x86_64/APKINDEX.tar.gz
fetch https://apk.cgr.dev/wolfi-presubmit/81c874a44d3cf5f8dddcc8ddaaa305b5e781d608/x86_64/APKINDEX.tar.gz
 [https://packages.cgr.dev/extras]
 [https://packages.wolfi.dev/os]
 [https://apk.cgr.dev/wolfi-presubmit/81c874a44d3cf5f8dddcc8ddaaa305b5e781d608]
OK: 155728 distinct packages available
bash-5.2# apk add curl
(1/1) Installing curl (8.12.1-r0)
Executing busybox-1.37.0-r40.trigger
OK: 52 MiB in 41 packages
```
Verify that an apk add `curl=<updatedver>` is refused because `libcurl-openssl4` is pinned in the image:
```
bash-5.2# apk add curl=8.12.1-r1
  libcurl-rustls4-8.12.1-r1:
    conflicts: libcurl-openssl4-8.12.1-r0[so:libcurl.so.4=4]
    satisfies: curl-8.12.1-r1[libcurl-abi=8.12.1] curl-8.12.1-r1[so:libcurl.so.4] git-2.49.0-r0[so:libcurl.so.4]
  libcurl-openssl4-8.12.1-r0:
    conflicts: libcurl-rustls4-8.12.1-r1[so:libcurl.so.4=4]
    satisfies: world[libcurl-openssl4=8.12.1-r0] curl-8.12.1-r1[so:libcurl.so.4] git-2.49.0-r0[so:libcurl.so.4]
```
Verify that an `apk add curl=<updatever> libcurl-openssl4=<updatever>` works, and smoke test curl:
```
bash-5.2# apk add curl=8.12.1-r1 libcurl-openssl4=8.12.1-r1
(1/2) Upgrading libcurl-openssl4 (8.12.1-r0 -> 8.12.1-r1)
(2/2) Upgrading curl (8.12.1-r0 -> 8.12.1-r1)
Executing busybox-1.37.0-r40.trigger
OK: 52 MiB in 41 packages
bash-5.2# curl https://www.nmt.edu -o foo.html
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 66772    0 66772    0     0  88813      0 --:--:-- --:--:-- --:--:-- 88792`
bash-5.2# ls -l foo.html 
-rw-r--r--    1 root     root         66772 Apr  3 17:18 foo.html
```

dannf:
> Running `apk add curl=8.12.1-r1` (effectively downgrading the package) doesn't do the expected thing: `libcurl-openssl4-8.13.0-r0` is kept installed. But I don't think we support downgrades anyway.

So you end up with a `curl` that depends on `libcurl-abi=8.12.1`, and a `libcurl-openssl4` that provides `libcurl-abi=8.13.0`? Weird, I wonder why?

sergiodj:
> > Running `apk add curl=8.12.1-r1` (effectively downgrading the package) doesn't do the expected thing: `libcurl-openssl4-8.13.0-r0` is kept installed. But I don't think we support downgrades anyway.
> 
> So you end up with a `curl` that depends on `libcurl-abi=8.12.1`, and a `libcurl-openssl4` that provides `libcurl-abi=8.13.0`? Weird, I wonder why?

This must have been a local inconsistency in my `local-wolfi` container, because I can't reproduce it anymore.  Please ignore that part of my comment.

PR #41428, opened 2025-02-05
Remove a bunch of packages not needed for public images.

mattmoor:


imjasonh (approved):

james-mchugh:
Hi. Out of curiosity, why were these packages removed from Wolfi? Many of these packages are relied on for production systems. Were they removed so they're only accessible behind the CG paywall?

PR #41429, opened 2025-02-05
Fix keyutils-dev

jonjohnsonjr:
There is an absolute symlink that is tripping up melange's SCA when run with bubblewrap.

https://github.com/chainguard-dev/melange/issues/1775


murraybd:
This looks good to me but just to be extra cautious let's wait for somebody's start of day to approve and merge this.

sil2100 (approved):

PR #41491, opened 2025-02-06
s6: usrmerge

dannf:


sil2100 (changes requested):
In general this feels okay, but I think it's not entirely complete. When you look at `s6`, there is also a subpackage `s6-ipcserver` which basically moves some of the /bin binaries from `destdir` to  `subpkgdir`. In this case it would just move the symlinks. So we need to adjust it to copy both `/usr/bin` and `/bin`.

dannf:
Pressing pause while we consider the best path forward.

PR #41493, opened 2025-02-06
net-tools: usrmerge

dannf:


sil2100 (approved):
Ok, at the beginning this was a bit confusing to me, since we only mention usrmerge here. But this is basically usrmerge AND sbin-merge that's happening here (per https://github.com/orgs/wolfi-dev/discussions/40270 ). And with this in mind, all good!

 I think we need to put a bit more context in the PR descriptions.

PR #41510, opened 2025-02-06
Add hydra to pkgs

anushkamittal20:
Fixes: New package

Related: N/A

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches


hydra.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/ory/hydra
+      tag: v${{package.version}}
+      expected-commit: ee8c339ddada3a42529c0416897abc32bad03bbb
+
+  - runs: |
+      go install -tags sqlite,json1,hsm \
```
anushkamittal20, line 27: need to change to use pipeline

mritunjaysharma394: any links to thread where the issue of not using the pipeline is discussed?

hydra.yaml
```diff
@@ -0,0 +1,63 @@
...
+  # - uses: go/install
+  #   with:
+  #     package: .
+  #     ldflags: -s -w -X 'github.com/ory/hydra/cmd.Version=${{package.version}}' -X 'github.com/ory/hydra/cmd.Commit=${EXPECTED_COMMIT}' -X 'github.com/ory/hydra/cmd.Date=$(date -u +%Y-%m-%dT%H:%M:%SZ)'
+  #     tags: sqlite,json1,hsm
+  - runs: |
+      go install -tags sqlite,json1,hsm \
+        -ldflags "-s -w \
+        -X github.com/ory/hydra/cmd.Version=${PACKAGE_VERSION} \
+        -X github.com/ory/hydra/cmd.Commit=${EXPECTED_COMMIT} \
```
mritunjaysharma394, line 35: Are these env variables defined somewhere? Example: `${EXPECTED_COMMIT}` and `${PACKAGE_VERSION}`? I think you probably will need this:
 
```suggestion
        -X github.com/ory/hydra/cmd.Version=v${{package.version}} \
        -X github.com/ory/hydra/cmd.Commit=$(git rev-parse HEAD) \
```

mritunjaysharma394 (changes requested):
Thanks for the work on this, just couple of questions and suggestions

hydra.yaml
anushkamittal20: I just realised these flags dont end up doing what I intended for them to do, that is set the version information. Saw same behavious in upstream build so eliminating flags altogether.

mritunjaysharma394: Or probably you were looking for this one? https://github.com/ory/hydra/blob/96d4ffe2cf3c6fce40d968f343088be39f8de165/Makefile#L184
My other question is what was the error that we see if using go pipeline, I suppose the ideal way is to use that and if we can't we should comment on why

mritunjaysharma394:
Possibly only this needs a little addressing https://github.com/wolfi-dev/os/pull/41510#discussion_r1961610094

hydra.yaml
anushkamittal20: THANK YOU FOR POINTING ME TO THAT
the binary doesnt quite get placed/built using the pipeline, let me try once again and add a comment.

hydra.yaml
anushkamittal20: I think I got this this time

anushkamittal20:
I missed that comment @mritunjaysharma394. Now with the last few commits I think I was able to resolve them all. Thank you for nudge in the right direction.

hydra.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+  - uses: go/install
+    with:
+      package: .
+      ldflags: -X 'github.com/ory/hydra/v2/driver/config.Version=$(git describe --abbrev=0 --tags)' -X 'github.com/ory/hydra/v2/driver/config.Date=$(TZ=UTC date -u "+%Y-%m-%dT%H:%M:%SZ")' -X 'github.com/ory/hydra/v2/driver/config.Commit=$(git rev-parse HEAD)'
+      tags: sqlite,json1,hsm
+
+  - runs: |
+      mkdir -p "${{targets.destdir}}/usr/local/bin"
+      cp /home/build/melange-out/${{package.name}}/usr/bin/hydra "${{targets.destdir}}"/usr/local/bin/hydra
```
mritunjaysharma394, line 34: I believe by default we are in the working directory of `/home/build/melange-out/${{package.name}}` so we don't need to add it again, we should do probably something like this:

```suggestion
      cp /usr/bin/hydra "${{targets.destdir}}"/usr/bin/hydra
```

I don't think we should have `local` here too as I don't see them in upstream https://github.com/ory/hydra/blob/96d4ffe2cf3c6fce40d968f343088be39f8de165/.docker/Dockerfile-alpine#L17, any reference from where you got it?

mritunjaysharma394:
Thanks @anushkamittal20 , just some small final suggestions hopefully

hydra.yaml
anushkamittal20: I remember picking it from somewhere, let me find it

hydra.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+update:
+  enabled: true
```
mritunjaysharma394, line 37: We can also have probably ignoring regex pattern for releases that have `-rc` or `-pre` suffix

```bash
  ignore-regex-patterns:
    - rc*
    - pre*
  ```
  
  As it seems from their releases https://github.com/ory/hydra/releases

hydra.yaml
anushkamittal20: cant recall now :(
I have removed that block (default in bin), updated tests and pushed the commit 🤞

mritunjaysharma394 (approved):

mritunjaysharma394:
Thanks for being patient with the reviews, lgtm, great work!

PR #41568, opened 2025-02-06
rsyslog: add other plugins and expand tests

Dentrax:
See: https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/main/rsyslog/APKBUILD

rsyslog supports lots of other systems so in order to enable those, some build-time deps are needed. Also introduce a new `plugins` subpackages so those can be `apk add`able individually. `rsyslog-plugins-all` may needed for the image. (Similar structure as `openldap` package building)

### DIFF
--- a/rsyslog.yaml
+++ b/rsyslog.yaml
@@ -23,10 +23,6 @@ environment:
       - flex
       - flex-dev
       - gnutls-dev
-      # - liblognorm-dev
-      # - czmq-dev
-      # - byacc
-      # - bsd-compat-headers
       - hiredis
       - hiredis-dev
       - iproute2
@@ -76,7 +72,6 @@ data:
       mmdblookup: mmdblookup
       mmfields: mmfields
       mmjsonparse: mmjsonparse
-      # mmnormalize: mmnormalize
       mmpstrucdata: mmpstrucdata
       mmrm1stspace: mmrm1stspace
       mmsequence: mmsequence
@@ -94,7 +89,6 @@ data:
       tls: lmnsd_gtls
       udpspoof: omudpspoof
       uxsock: omuxsock
-      # zmq: imczmq,omczmq
       imdocker: imdocker
       clickhouse: omclickhouse
 
@@ -202,7 +196,6 @@ subpackages:
         - rsyslog-mmdblookup
         - rsyslog-mmfields
         - rsyslog-mmjsonparse
-        # - rsyslog-mmnormalize
         - rsyslog-mmpstrucdata
         - rsyslog-mmrm1stspace
         - rsyslog-mmsequence
@@ -220,7 +213,6 @@ subpackages:
         - rsyslog-tls
         - rsyslog-udpspoof
         - rsyslog-uxsock
-        # - rsyslog-zmq
         - rsyslog-imdocker
         - rsyslog-clickhouse
     pipeline:
--- a/rsyslog.yaml
+++ b/rsyslog.yaml
@@ -1,10 +1,12 @@
 package:
   name: rsyslog
   version: 8.2412.0
-  epoch: 3
+  epoch: 2
   description: a Rocket-fast SYStem for LOG processing
   copyright:
-    - license: Apache-2.0 AND GPL-3.0-or-later AND LGPL-3.0-or-later
+    - license: Apache-2.0
+    - license: GPL-3.0-or-later
+    - license: LGPL-3.0-or-later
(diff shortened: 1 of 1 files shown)


rsyslog.yaml
```diff
@@ -1,54 +1,154 @@
...
       - bison
       - build-base
       - busybox
       - ca-certificates-bundle
       - curl-dev
+      - file
       - flex
+      - flex-dev
+      - gnutls-dev
+      # - liblognorm-dev
```
developer-guy, line 24: if these aren't needed anymore maybe we could remove them

rsyslog.yaml
```diff
@@ -1,54 +1,154 @@
...
+      clickhouse: omclickhouse
+
 pipeline:
   - uses: git-checkout
     with:
       expected-commit: 999a1b63d65f2b2f8484fe9e97bdc6ba0edb182d
       repository: https://github.com/rsyslog/rsyslog
       tag: v${{package.version}}
 
+  # TODO: Add more options
```
developer-guy, line 106: same

rsyslog.yaml
Dentrax: Yeah actually commented out intentionally for the next person we're going to add other missing plugins below. Not sure if should keep these or not 🤔

rsyslog.yaml
```diff
@@ -1,54 +1,154 @@
 package:
   name: rsyslog
   version: 8.2412.0
-  epoch: 1
+  epoch: 3
```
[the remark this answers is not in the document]
Dentrax: Thanks, done!

rsyslog.yaml
wojciechka: Perhaps group them at the end and leave a comment why they are commented out?

wojciechka (commented):
I think this is good, modulo some more comments

Dentrax:
@wojciechka - for some reason I couldn't reply in the discussion, but as per your comment:

> Perhaps group them at the end and leave a comment why they are commented out?

I've tried that but `yam` complains it and rearrange by alphabetical order.

rsyslog.yaml
Dentrax: https://github.com/wolfi-dev/os/pull/41568#issuecomment-2643730468

Dentrax: Done, thanks.

rsyslog.yaml
```diff
@@ -63,6 +159,66 @@ pipeline:
...
+            mv ${{targets.destdir}}/usr/lib/rsyslog/${plugin}.so ${{targets.contextdir}}/usr/lib/rsyslog/
+          done
+    test:
+      pipeline:
+        - runs: |
+            PLUGINS="$(echo "${{range.value}}" | tr ',' ' ')"
+            for plugin in $PLUGINS; do
+              stat /usr/lib/rsyslog/${plugin}.so
+            done
+
```
kranurag7, line 182: We can consider doing a split/doc and split/dev here well. 💭

kranurag7 (commented):
great work!! I think the behavior before and after refactoring is not changing so I think this looks good. 

Some minor comments/suggestions inline.

rsyslog.yaml
Dentrax: `split/dev` might be a breaking change, so i'd keep that for another discussion tbh.

rsyslog.yaml
```diff
@@ -63,6 +159,66 @@ pipeline:
...
+          for plugin in $PLUGINS; do
+            mv ${{targets.destdir}}/usr/lib/rsyslog/${plugin}.so ${{targets.contextdir}}/usr/lib/rsyslog/
+          done
+    test:
+      pipeline:
+        - runs: |
+            PLUGINS="$(echo "${{range.value}}" | tr ',' ' ')"
+            for plugin in $PLUGINS; do
+              stat /usr/lib/rsyslog/${plugin}.so
+            done
```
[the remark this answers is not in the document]
Dentrax: Notice that some range values has multiple items, so need to iterate it over.

We already have an `ldd-check` pipeline below but we can't run it per _range_ item since it would require adding multiple new runtime dependencies to fulfill the each plugins' requirement.

kranurag7 (approved):
thank you!! looks good to me, mostly looks non-breaking because before and after package is same and we're introducing more plugins via this commit. Will keep an eye on dependents.

Issue #41305, opened 2025-02-04
kubeflow-katib fails ldd-check test pipeline


murraybd:
The package kubeflow-katib will fail the ldd-check test pipeline with the following error:

```
2025/02/03 19:36:47 INFO FAIL[ldd-check]: /opt/katib/cmd/suggestion/hyperband/v1beta1/lib/python3.10/site-packages/scipy.libs/libgfortran-040039e1.so.5.0.0: missing libquadmath-96973f99.so.0.0.0
2025/02/03 19:36:59 INFO INFO[ldd-check]: tested 229 files with ldd. 228 passes. 1 fails.
```

Issue #41304, opened 2025-02-04
kserve fails ldd-check test pipeline


murraybd:
The package kserve will fail the ldd-check test pipeline with the following error:

```
2025/02/03 19:45:37 INFO FAIL[ldd-check]: /usr/lib/python3.11/site-packages/numpy.libs/libgfortran-040039e1.so.5.0.0: missing libquadmath-96973f99.so.0.0.0
2025/02/03 19:45:37 INFO FAIL[ldd-check]: /usr/lib/python3.11/site-packages/numpy.libs/libopenblas64_p-r0-0cf96a72.3.23.dev.so: missing libgfortran-040039e1.so.5.0.0
2025/02/03 19:46:18 INFO INFO[ldd-check]: tested 112 files with ldd. 110 passes. 2 fails.
```

a-crate:
Should be fixed, I can't close this though.

Issue #40843, opened 2025-01-29
kubeflow-pipelines-visualization-server fails ldd-check test pipeline


murraybd:
The package kubeflow-pipelines-visualization-server will fail the ldd-check test pipeline with the following error:

```
2025/01/28 01:31:39 INFO FAIL[ldd-check]: /usr/lib/python3.11/site-packages/h5py.libs/libhdf5-a6e30693.so.310.4.0: missing libsz-b66d1717.so.2.0.1
2025/01/28 01:31:39 INFO FAIL[ldd-check]: /usr/lib/python3.11/site-packages/h5py.libs/libhdf5_hl-e82549de.so.310.0.4: missing libhdf5-a6e30693.so.310.4.0 libsz-b66d1717.so.2.0.1
2025/01/28 01:31:39 INFO FAIL[ldd-check]: /usr/lib/python3.11/site-packages/h5py.libs/libsz-b66d1717.so.2.0.1: missing libaec-001fb5f0.so.0.0.12
2025/01/28 01:32:12 INFO FAIL[ldd-check]: /usr/lib/python3.11/site-packages/numpy.libs/libgfortran-040039e1.so.5.0.0: missing libquadmath-96973f99.so.0.0.0
2025/01/28 01:32:12 INFO FAIL[ldd-check]: /usr/lib/python3.11/site-packages/numpy.libs/libopenblas64_p-r0-0cf96a72.3.23.dev.so: missing libgfortran-040039e1.so.5.0.0
2025/01/28 01:32:23 INFO FAIL[ldd-check]: /usr/lib/python3.11/site-packages/pillow.libs/libbrotlidec-a621e7f2.so.1.1.0: missing libbrotlicommon-5b2eba61.so.1.1.0
2025/01/28 01:32:23 INFO FAIL[ldd-check]: /usr/lib/python3.11/site-packages/pillow.libs/libfreetype-edd71024.so.6.20.1: missing libpng16-2a828d05.so.16.44.0 libbrotlidec-a621e7f2.so.1.1.0
2025/01/28 01:32:23 INFO FAIL[ldd-check]: /usr/lib/python3.11/site-packages/pillow.libs/libharfbuzz-07d0ad17.so.0.61010.0: missing libfreetype-edd71024.so.6.20.1
[...]

Issue #40842, opened 2025-01-29
superset fails ldd-check test pipeline


murraybd:
The package superset will fail the ldd-check test pipeline with the following error:

```
2025/01/27 23:11:29 INFO running step "test/ldd-check"
2025/01/27 23:11:30 INFO [ldd-check] Testing binaries in package superset
2025/01/27 23:11:42 INFO PASS[ldd-check]: /usr/share/superset/venv/bin/python
2025/01/27 23:11:42 INFO PASS[ldd-check]: /usr/share/superset/venv/bin/python3
2025/01/27 23:11:42 INFO PASS[ldd-check]: /usr/share/superset/venv/bin/python3.11
2025/01/27 23:11:42 INFO PASS[ldd-check]: /usr/share/superset/venv/lib/python3.11/site-packages/_brotli.cpython-311-x86_64-linux-gnu.so
2025/01/27 23:11:42 INFO PASS[ldd-check]: /usr/share/superset/venv/lib/python3.11/site-packages/_cffi_backend.cpython-311-x86_64-linux-gnu.so
2025/01/27 23:11:42 INFO PASS[ldd-check]: /usr/share/superset/venv/lib/python3.11/site-packages/_geohash.cpython-311-x86_64-linux-gnu.so
2025/01/27 23:11:42 INFO PASS[ldd-check]: /usr/share/superset/venv/lib/python3.11/site-packages/pvectorc.cpython-311-x86_64-linux-gnu.so
2025/01/27 23:11:44 INFO PASS[ldd-check]: /usr/share/superset/venv/lib/python3.11/site-packages/PIL/_imaging.cpython-311-x86_64-linux-gnu.so
2025/01/27 23:11:44 INFO PASS[ldd-check]: /usr/share/superset/venv/lib/python3.11/site-packages/PIL/_imagingcms.cpython-311-x86_64-linux-gnu.so
2025/01/27 23:11:44 INFO PASS[ldd-check]: /usr/share/superset/venv/lib/python3.11/site-packages/PIL/_imagingft.cpython-311-x86_64-linux-gnu.so
2025/01/27 23:11:44 INFO PASS[ldd-check]: /usr/share/superset/venv/lib/python3.11/site-packages/PIL/_imagingmath.cpython-311-x86_64-linux-gnu.so
[...]

PR #41625, opened 2025-02-07
Allow test/ldd-check to resolve shared objects in the same python module, run newly passing tests

### LINKED ISSUE #41002: checkov fails ldd-check test pipeline
The package checkov will fail the ldd-check test pipeline with the following error:

```
2025/01/30 20:53:51 INFO FAIL[ldd-check]: /usr/share/app/checkov/.venv/lib/python3.11/site-packages/igraph.libs/libxml2-3998bec4.so.2.9.1: missing liblzma-004595ca.so.5.2.2
2025/01/30 20:54:42 INFO INFO[ldd-check]: tested 50 files with ldd. 49 passes. 1 fails.
```

a-crate:
Fixes: https://github.com/wolfi-dev/os/issues/41002, https://github.com/wolfi-dev/os/issues/40838, https://github.com/wolfi-dev/os/issues/41304, https://github.com/wolfi-dev/os/issues/41305

These are cases where a python module brings a shared object with it, and then that shared object links against a 2nd shared object brought by the same module.
In this case SO1 links against SO2 using just the library name (instead of a relative path), which the python interpreter allows by loading objects in the same directory before looking elsewhere. LDD doesn't do this by default.

Related: https://github.com/wolfi-dev/os/issues/40831, https://github.com/wolfi-dev/os/issues/40843, https://github.com/wolfi-dev/os/issues/40842

These have cases of this but they also have some other ldd failures unrelated to python stuff.

### DIFF
--- a/kubeflow-katib.yaml
+++ b/kubeflow-katib.yaml
@@ -86,6 +86,11 @@ subpackages:
           tar c ./pkg | tar xvf - -C ${{targets.subpkgdir}}/opt/katib/
           mv .venv-earlystopping/* ${{targets.subpkgdir}}/opt/katib/$SUGGESTION_DIR
           mv ./$SUGGESTION_DIR/* ${{targets.subpkgdir}}/opt/katib/$SUGGESTION_DIR
+    test:
+      pipeline:
+        - uses: test/ldd-check
+          with:
+            packages: $(basename ${{targets.contextdir}})
 
   - name: katib-tfevent-metricscollector
     description: Kubeflow Katib tfevent-metricscollector
@@ -105,6 +110,11 @@ subpackages:
           cp -r ./pkg ${{targets.subpkgdir}}/opt/katib/pkg
           mv ./$SUGGESTION_DIR/* ${{targets.subpkgdir}}/opt/katib/$SUGGESTION_DIR
           cd ${{targets.subpkgdir}}/opt/katib/$SUGGESTION_DIR
+    test:
+      pipeline:
+        - uses: test/ldd-check
+          with:
+            packages: $(basename ${{targets.contextdir}})
 
   - name: katib-suggestion-hyperband
     description: Kubeflow Katib suggestion-hyperband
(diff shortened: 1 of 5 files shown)


pipelines/test/ldd-check.yaml
```diff
@@ -88,7 +90,13 @@ pipeline:
           return 0
         fi
 
-        ldd "$f" >$outf 2>$errf || rc=$?
+        if [[ "$f" =~ python.*site-packages ]]; then
```
smoser, line 93: i dont' actually think you're guaranteed bash here.

same posix equivalent:

```bash
case "$f" in
    python*site-packages)
        ...
esac

File: kubeflow-katib.yaml [hunk 2]
```diff
@@ -86,6 +86,11 @@ subpackages:
           tar c ./pkg | tar xvf - -C ${{targets.subpkgdir}}/opt/katib/
           mv .venv-earlystopping/* ${{targets.subpkgdir}}/opt/katib/$SUGGESTION_DIR
           mv ./$SUGGESTION_DIR/* ${{targets.subpkgdir}}/opt/katib/$SUGGESTION_DIR
+    test:
+      pipeline:
+        - uses: test/ldd-check
+          with:
+            packages: $(basename ${{targets.contextdir}})
```
murraybd, line 93: You could actually use `${{subpkg.name}}` instead of the above.

pipelines/test/ldd-check.yaml
a-crate: Fixed.

kubeflow-katib.yaml
a-crate: Fixed.

murraybd (dismissed):

kubeflow-katib.yaml
```diff
@@ -105,6 +110,11 @@ subpackages:
           cp -r ./pkg ${{targets.subpkgdir}}/opt/katib/pkg
           mv ./$SUGGESTION_DIR/* ${{targets.subpkgdir}}/opt/katib/$SUGGESTION_DIR
           cd ${{targets.subpkgdir}}/opt/katib/$SUGGESTION_DIR
+    test:
+      pipeline:
+        - uses: test/ldd-check
+          with:
+            packages: $(basename ${{targets.contextdir}})
```
smoser, line 117: Is there a reason you wouldn't use subpkg.name here?

a-crate: Nope, that was a mistake in my find+replace.

murraybd (approved):

PR #41772, opened 2025-02-08
add nats-top package

kranurag7:
nats-top package

### DIFF
--- a/nats-top.yaml
+++ b/nats-top.yaml
@@ -17,6 +17,7 @@ pipeline:
     with:
       packages: ./nats-top.go
       output: nats-top
+      extra-args: -cover
       ldflags: |
         -X main.version=${{package.version}}
         -X main.commit=$(git rev-parse HEAD)
@@ -38,6 +39,8 @@ update:
 
 test:
   environment:
+    environment:
+      GOCOVERDIR: /home/build
     contents:
       packages:
         - nats-server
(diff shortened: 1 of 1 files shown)


nats-top.yaml
```diff
@@ -51,3 +51,4 @@ test:
         sleep 2
         nats-top -o nats-top-output
         cat nats-top-output | grep -Ei 'server|bytes|sec'
+    - uses: go/covdata
```
maxgio92, line 54: It requires:
- binary to be built with `go build -cover`
- the `GOCOVERDIR` set (unfortunate, as currently pipelines allows to inject packages but not environment to the test environment)

nats-top.yaml
```diff
@@ -0,0 +1,57 @@
...
+    with:
+      repository: https://github.com/nats-io/nats-top
+      tag: v${{package.version}}
+      expected-commit: 812d35cc991232d7606816a15b2ea0b12fb129e7
+
+  - uses: go/build
+    with:
+      packages: ./nats-top.go
+      output: nats-top
+      extra-args: -cover
```
maxgio92, line 20: @kranurag7 I don't think we want to publish the main package with the instrumented binary, but having it as as subpackage, because it has binary size and performance consequnces

kranurag7:
closing given we've this package on main already

PR #41928, opened 2025-02-09
http-echo: go/build pipeline, fix ldflags and expand tests

kranurag7:
this commit does the following for http-echo package:
    - uses go/build pipeline now
    - fixes ldflags
    - expand tests where we do a curl test.


http-echo.yaml
```diff
@@ -49,8 +32,14 @@ update:
...
+      packages:
+        - curl
   pipeline:
-    # AUTOGENERATED
     - runs: |
-        http-echo --version
+        http-echo --version | grep ${{package.version}}
         http-echo --help
+    - runs: |
+        http-echo -listen=:8080 -text="hello world" &
```
EyeCantCU, line 44: I'd prefer we use `test/daemon-check-output` here and used `wait-for-it` afterwards to guarantee port availability before using curl

http-echo.yaml
```diff
@@ -49,8 +32,14 @@ update:
...
+        curl -s localhost:8080 | grep hello
```
EyeCantCU, line 45: Or at least retired a few times just in case

kranurag7:
closing out; the tests on main is refactored via https://github.com/wolfi-dev/os/pull/51656

PR #42122, opened 2025-02-10
add xeol package

kranurag7:
adds xeol package (https://github.com/xeol-io/xeol)

### DIFF
--- a/xeol.yaml
+++ b/xeol.yaml
@@ -67,4 +67,4 @@ test:
         xeol --help
     - name: Run an EOL check on curl image
       runs: |
-        xeol cgr.dev/chainguard/curl:latest
+        xeol cgr.dev/chainguard/curl:latest --output json --quiet | jq | grep -E 'chainguard|curl'
--- a/xeol.yaml
+++ b/xeol.yaml
@@ -16,6 +16,18 @@ pipeline:
       tag: v${{package.version}}
       expected-commit: 76ef5b862f988a162bd4c9cdda2c144a3178efc3
 
+  - uses: go/bump
+    with:
+      deps: |-
+        github.com/go-git/go-git/v5@v5.13.0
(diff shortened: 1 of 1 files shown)


xeol.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+test:
+  pipeline:
+    - name: ${{package.name}} version and help
+      runs: |
+        xeol --version | grep ${{package.version}}
+        xeol --help
+    - name: Run an EOL check on curl image
+      runs: |
+        xeol cgr.dev/chainguard/curl:latest
```
Dentrax, line 70: Just a sanity check, can we ensure it prints something to stdout|stderr? Simple `grep` check would work here. (My concern was that tool might just exits with zero without writing anything)

Dentrax (changes requested):

xeol.yaml
```diff
@@ -0,0 +1,74 @@
...
+      packages:
+        - jq
+  pipeline:
+    - name: ${{package.name}} version and help
+      runs: |
+        xeol --version | grep ${{package.version}}
+        xeol --help
+    - name: Run an EOL check on curl image
+      runs: |
+        xeol cgr.dev/chainguard/curl:latest --output json --quiet | jq | grep -E 'chainguard|curl'
```
developer-guy, line 74: this looks pretty much ok to me but how about doing something like this? https://github.com/wolfi-dev/os/blob/ccf360e63f4a3e28e1e296a7d4112b5e2bde2e0f/distribution.yaml#L141-L142

or we could just simply use "crane registry serve &" then we can use localhost:5000 as a registry host to pull image and use it for xeol.

I'm looking forward to hear from you!

developer-guy: sounds good to me!

Dentrax (approved):

Dentrax (approved):

PR #42185, opened 2025-02-10
ingress-nginx-controller: build from specific commit of opentelemetry-cpp-contrib

kranurag7:
build using specific commit of opentelemetry-cpp-contrib.

### DIFF
--- a/ingress-nginx-controller-1.12.yaml
+++ b/ingress-nginx-controller-1.12.yaml
@@ -3,7 +3,7 @@ package:
   name: ingress-nginx-controller-1.12
   version: 1.12.0
   # There are manual changes to review between each package update. See 'vars:' section.
-  epoch: 6
+  epoch: 7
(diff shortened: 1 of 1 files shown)


EyeCantCU (changes requested):
It doesn't look like our config tests OTel anymore

kranurag7:
> It doesn't look like our config tests OTel anymore

thank you RJ, I've include few more opentelemetry specific keys in there, it's changing b/w updates. I wanted to do this but this will be a little unstable for now given the config might be not updated upstream as well. 

```bash
    - runs: |
        git clone https://github.com/open-telemetry/opentelemetry-cpp-contrib
        cd opentelemetry-cpp-contrib
        git reset --hard ${{vars.OTEL_SHA}}
        nginx -T -c instrumentation/nginx/test/conf/nginx.conf
```

EyeCantCU (dismissed):

EyeCantCU (approved):

PR #42235, opened 2025-02-10
add rpm2cpio package

kranurag7:
adds rpm2cpio 

(similar to wait-for-it script) 

coming from https://github.com/freebsd/freebsd-ports/blob/main/archivers/rpm2cpio/files/rpm2cpio and uses bsdtar because gnutar doesn't recognized the flag.

### DIFF
--- a/rpm2cpio.yaml
+++ b/rpm2cpio.yaml
@@ -34,8 +34,10 @@ test:
     contents:
       packages:
         - curl
+        - jq
   pipeline:
     - runs: |
         rpm2cpio
-        curl -L -o yq.rpm https://kojipkgs.fedoraproject.org//packages/yq/4.43.1/5.fc40/src/yq-4.43.1-5.fc40.src.rpm
-        rpm2cpio yq.rpm | grep -i fedora
+        export COSIGN_VERSION=$(curl -s --retry 5 "https://api.github.com/repos/sigstore/cosign/releases/latest" | jq -r ".tag_name" | sed 's/v//')
+        curl -fsSL --retry 5 -o cosign.rpm https://github.com/sigstore/cosign/releases/download/v${COSIGN_VERSION}/cosign-${COSIGN_VERSION}-1.${{build.arch}}.rpm
+        rpm2cpio cosign.rpm | cpio -idmv
--- a/rpm2cpio.yaml
+++ b/rpm2cpio.yaml
@@ -30,6 +30,12 @@ update:
   exclude-reason: single script file which hasn't been updated in last 10 years.
 
 test:
+  environment:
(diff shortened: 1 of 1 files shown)


rpm2cpio.yaml
```diff
@@ -0,0 +1,35 @@
...
+      sed -i 's/tar/bsdtar/g' ${{targets.destdir}}/usr/bin/rpm2cpio
+
+update:
+  enabled: false
+  exclude-reason: single script file which hasn't been updated in last 10 years.
+
+test:
+  pipeline:
+    - runs: |
+        rpm2cpio
```
EyeCantCU, line 35: We should be able to fetch any RPM and then use cpio to parse

EyeCantCU (changes requested):

rpm2cpio.yaml
developer-guy: instead fetching the yq rpm from an internet which may be removed in the future don't you think that we could add this file into the package folder? but other than that this PR looks so cool!

kranurag7: I had this fear :homerdisappear: 

but the rpm size looks not so big so I think we can put it inside the package directory, I'll do that in the subsequent commit. 

```bash
 $ ls -lah yq.rpm
-rw-r--r--    1 user     docker    307.7K Feb 12 07:03 yq.rpm
```
I also tried if we can package a `*.rpm` package easily with some dummy files and turns out that's not easy (phew). :(

Dentrax: It seems you've added test case! Resolving this discussion.

Dentrax (dismissed):

rpm2cpio.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+test:
+  environment:
+    contents:
+      packages:
+        - curl
+  pipeline:
+    - runs: |
+        rpm2cpio
+        curl -L -o yq.rpm https://kojipkgs.fedoraproject.org//packages/yq/4.43.1/5.fc40/src/yq-4.43.1-5.fc40.src.rpm
```
EyeCantCU, line 40: We shouldn't fetch a pinned version of yq. Let's retrieve the latest release of cosign instead (that version may be pulled from koji in the future): https://github.com/sigstore/cosign/releaseshttps://github.com/sigstore/cosign/releases

EyeCantCU (changes requested):

rpm2cpio.yaml
Dentrax: Thanks, done!

EyeCantCU (approved):

PR #42242, opened 2025-02-10
glibc: add test for compatibility with gcc

sil2100:
Add the test for the issue we missed when trying to release 2.41 prematurely. There is a very very very low chance this will happen again, but I still think it's a good idea to add a test. In case something similar is broken with pthreads once again.

This can land as-is, doesn't actually require the new glibc. Confirmed to be failing with a straight rebuild of glibc without fixing gcc fix-includes.

### DIFF
--- a/glibc.yaml
+++ b/glibc.yaml
@@ -615,6 +615,11 @@ subpackages:
         - empty
 
 test:
+  environment:
+    contents:
+      packages:
+        - gcc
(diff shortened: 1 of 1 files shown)


sil2100:
Test succeeded in my manual run, though only ran x86.

dannf (approved):

xnox (approved):

PR #42250, opened 2025-02-10
py3-* misc deps

justinvreeland:


py3-netifaces.yaml
```diff
@@ -0,0 +1,85 @@
...
+data:
+  - name: py-versions
+    items:
+      "3.10": "310"
+      "3.11": "311"
+      "3.12": "312"
+      "3.13": "313"
+
+vars:
+  module_name: netifaces
```
dannf, line 72: nit - we use `import` in most other places

dannf (dismissed):

py3-netifaces.yaml
```diff
@@ -0,0 +1,85 @@
...
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}
+    pipeline:
+      - name: Python Build
+        uses: py/pip-build-install
+        with:
+          python: python${{range.key}}
+    dependencies:
+      provides:
+        - py3-${{vars.pypi-package}}
```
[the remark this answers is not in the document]
justinvreeland: The initial intent behind it was to ensure a packages was installed that worked when `py3-*` was installed. Now that everything is converted it probably does more harm than good. People should be specifying a python version to depend on explicitly and that should resolve the deps imo.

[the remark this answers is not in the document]
smoser: I agree. Hope do we stop these things from being added? Hope do we rip off the bandaid?

Thoughts?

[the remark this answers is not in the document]
justinvreeland: Probably something like:

Add a melange linter as a warning.

Give guidance to people to used the versioned versions for packages and images.

Take a pass at sed/yq fixing some large number of them.

Make the melange linter an error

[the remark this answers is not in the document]
justinvreeland: Opened a ticket to track this but it shouldn't stop this PR.

https://github.com/wolfi-dev/os/issues/42965

dannf (dismissed):
Can you add the py3-supported-* metapackages? Apologies for not noticing their absence last time. I think that is a convention we want to keep.

justinvreeland:
> Can you add the py3-supported-* metapackages? Apologies for not noticing their absence last time. I think that is a convention we want to keep.

Oh I def should. I'm not sure why that didn't make it into my template tbh.

py3-netifaces.yaml
```diff
@@ -0,0 +1,99 @@
...
+            import: ${{vars.import}}
+            python: python${{range.key}}
+
+  - name: py3-supported-${{vars.pypi-package}}
+    description: meta package providing ${{vars.pypi-package}} for supported python versions.
+    dependencies:
+      runtime:
+        - py3.10-${{vars.pypi-package}}
+        - py3.11-${{vars.pypi-package}}
+        - py3.12-${{vars.pypi-package}}
```
dannf, line 55: no 3.13?

py3-pyserial.yaml
```diff
@@ -0,0 +1,89 @@
...
+            import: ${{vars.import}}
+            python: python${{range.key}}
+
+  - name: py3-supported-${{vars.pypi-package}}
+    description: meta package providing ${{vars.pypi-package}} for supported python versions.
+    dependencies:
+      runtime:
+        - py3.10-${{vars.pypi-package}}
+        - py3.11-${{vars.pypi-package}}
+        - py3.12-${{vars.pypi-package}}
```
dannf, line 55: no 3.13?

justinvreeland: uggg old snippet I had lying around.

dannf (dismissed):
*chef's kiss*

justinvreeland:
> _chef's kiss_

So close! As I designed them the `py3-supported` tests caught that all of the packages were installing binaries that still need to be split out : (

py3-pyserial.yaml
```diff
@@ -0,0 +1,110 @@
...
+
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}-bin
+    description: Executable binaries for ${{vars.pypi-package}} installed for python${{range.key}}
+    dependencies:
+      runtime:
+        - py${{range.key}}-${{vars.pypi-package}}
+      provides:
+        - py3-${{vars.pypi-package}}-bin
+        - py3-${{vars.pypi-package}}
```
dannf, line 72: I don't think we want the `-bin` packages providing this. We've done that in cases where we are migrating from a single `py3-foo` that provided both lib and bin in one, so that installing `py3-foo` still brought in commands like it did before . But this package is new, so that's not an issue here.

But also - weren't we going to drop the `py3-*` provides for  *new* `py3.x-*` packages anyway? Not a blocker if you want to keep them for now, thanks for opening the issue on that.

py3-pyserial.yaml
```diff
@@ -0,0 +1,110 @@
...
+    description: meta package providing ${{vars.pypi-package}} for supported python versions.
+    dependencies:
+      runtime:
+        - py3.10-${{vars.pypi-package}}
+        - py3.11-${{vars.pypi-package}}
+        - py3.12-${{vars.pypi-package}}
+        - py3.13-${{vars.pypi-package}}
+    test:
+      pipeline:
+        - uses: python/import
```
dannf, line 59: oh, we could add the new `tests/emptypackage` pipeline here

dannf (commented):
yay, tests work!

dannf (approved):

PR #42270, opened 2025-02-10
Add preset for containerd.service

xnox:
Such that it starts when needed. Sort of expected it to have a .socket
unit tbh.


roaksoax (approved):
LGTM!

PR #39066, opened 2025-01-08
ca-certs: re-add support for Entrust certs

imjasonh:

PR #42291, opened 2025-02-11
java-cacerts: update to 20241121 bundle

xnox:
Update java-cacerts to 20241121 bundle, also attempt to enable updates
to it, with lockstep against ca-certs.

If this doesn't work, we should just rebuild/refetch ca-certificates
in java-cacerts from scratch. Due to build-dependency cycles we can't
do it all in ca-certificates itself, but we should be able to redo it
again.

### DIFF
--- a/java-cacerts.yaml
+++ b/java-cacerts.yaml
@@ -1,8 +1,8 @@
 package:
   name: java-cacerts
   # Update this when ca-certificates is updated.
-  version: "20240705"
-  epoch: 0
+  version: "20241121"
+  epoch: 1
   description: "default certificate authorities for Java"
   copyright:
     - license: MIT
@@ -14,7 +14,7 @@ package:
 environment:
   contents:
     packages:
-      - ca-certificates
+      - ca-certificates=${{package.full-version}}
       - p11-kit-trust
       - wolfi-base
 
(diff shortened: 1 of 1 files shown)


java-cacerts.yaml
```diff
@@ -36,4 +36,6 @@ pipeline:
       chmod +x "${{targets.destdir}}"/etc/ca-certificates/update.d/java-cacerts
 
 update:
-  enabled: false
+  enabled: true
```
justinvreeland, line 39: 👍

justinvreeland (approved):

java-cacerts.yaml
```diff
@@ -14,7 +14,7 @@ package:
 environment:
   contents:
     packages:
-      - ca-certificates
+      - ca-certificates=${{package.full-version}}
```
smoser, line 17: Why did you use full-version here rather than just version?
full version includes epoch.  So you can't change this package at _all_ without bumping ca-certificates.
Maybe thats what you meant by ' lockstep against ca-certs', but it feels overly lockstep.

xnox: @smoser we had to revert certificate removal in an empoch. thus yes this is intentional to propagate. See for example https://github.com/wolfi-dev/os/commit/a674f13e78d98d7e33e38f82acb04450c25bb43e : 
- https://github.com/wolfi-dev/os/pull/39066

Which broke production deployments.

PR #42363, opened 2025-02-11
Add tests to wolfi-baselayout

mattmoor:


wolfi-baselayout.yaml
```diff
@@ -63,5 +63,31 @@ pipeline:
...
+    - name: "Scanner expectations"
+      runs: |
+        # This tells scanners which feed to look at.
+        grep 'ID=wolfi' /etc/os-release
+
+        # This tells most scanners which of a vendor's feeds to look at,
+        # but as a rolling distro we should never be changing this, so
+        # instead of using ${{package.version}} we hardcode this here
+        # because only the epoch should really ever change.
+        grep 'VERSION_ID="20230201"' /etc/os-release
```
mattmoor, line 90: @luhring LMK if there is anything inaccurate here!

luhring: Yeah that sounds reasonable to me! ([linking related section of scanner impl docs](https://github.com/chainguard-dev/vulnerability-scanner-support/blob/main/docs/scanning_implementation.md#well-then-whats-that-version_id-field-doing-there))

Given this test, do you think we should make `VERSION_ID="${{package.version}}"` hard coded in the `Generate /etc/os-release` step?

mattmoor: @luhring Then I think the test would pass even if the version changes, so I think I kind of like the current setup even if it's a little weird.  We could add a test that `${{package.version}}` is some fixed string too, but this covers that so 🤷‍♂️

luhring: Yeah — it's little implicit the way it is now, but I think that's okay

luhring (approved):

PR #42384, opened 2025-02-11
add tests for wolfi-keys

dustinkirkland:
Add both positive and negative tests for wolfi-keys

Fixes: https://github.com/chainguard-dev/internal-dev/issues/9151


OddBloke (approved):
LGTM, and I've confirmed the tests pass for me locally.

Issue #30234, opened 2024-10-07
Invalid fetch


xnox:
By clobbing together apkrane json output, doing curl --head request to get content size and comparing the results on per apk package and finding those that are different.

then mapping back to Origin to get list of yaml files, and checking those that exists, there are 433 yaml files which have at least one apk on x86_64 that is failing to `apk fetch` with bad archive.

Can these please be no change bumped and rebuilt?

[exists.txt](https://github.com/user-attachments/files/17285308/exists.txt)

ZiTAL:
```
apk fetch -R git rsync unzip wget zip
Downloading bash-5.2.37-r2
Downloading ca-certificates-bundle-20241010-r1
Downloading cyrus-sasl-2.1.28-r4
Downloading gdbm-1.24-r0
Downloading git-2.47.0-r0
Downloading glibc-2.40-r3
Downloading glibc-locale-posix-2.40-r3
Downloading heimdal-7.8.0-r6
Downloading keyutils-libs-1.6.3-r4
Downloading krb5-conf-1.0-r3
Downloading krb5-libs-1.21.3-r1
Downloading ld-linux-2.40-r3
Downloading libacl1-2.3.2-r3
Downloading libattr1-2.5.2-r3
Downloading libbrotlicommon1-1.1.0-r4
Downloading libbrotlidec1-1.1.0-r4
Downloading libcom_err-1.47.1-r0
ERROR: libcom_err-1.47.1-r0: BAD archive
Downloading libcrypt1-2.40-r3
Downloading libcrypto3-3.3.2-r2
Downloading libcurl-openssl4-8.10.1-r2
Downloading libevent-2.1.12-r6
Downloading libexpat1-2.6.3-r1
Downloading libgcc-14.2.0-r4
Downloading libidn2-2.3.7-r2
ERROR: libidn2-2.3.7-r2: BAD archive
Downloading libldap-2.6.8-r5
Downloading liblz4-1-1.10.0-r2
Downloading libnghttp2-14-1.64.0-r0
Downloading libpcre2-8-0-10.44-r1
ERROR: libpcre2-8-0-10.44-r1: BAD archive
Downloading libpsl-0.21.5-r3
ERROR: libpsl-0.21.5-r3: BAD archive
Downloading libssl3-3.3.2-r2
Downloading libunistring-1.3-r1
Downloading libverto-0.3.2-r4
ERROR: libverto-0.3.2-r4: BAD archive
[...]

PR #42414, opened 2025-02-11
wolfictl bump for apk fetch BAD archive

dannf:
Due to past bugs there is invalid metadata in the apkindex for the
size attribute. Rebuild and reindex affected packages.
See: #30234

### DIFF
--- a/libxcomposite.yaml
+++ b/libxcomposite.yaml
@@ -1,7 +1,7 @@
 package:
   name: libxcomposite
   version: 0.4.6
-  epoch: 3
+  epoch: 4
   description: X11 Composite extension library
   copyright:
     - license: MIT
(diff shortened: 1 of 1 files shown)


murraybd (approved):

murraybd (approved):

murraybd (dismissed):

dannf:
The bump-only change failed to test because the docs test failed. I've replaced the bump-only change w/ one that fixes the docs issue.

murraybd (approved):

murraybd:
> The bump-only change failed to test because the docs test failed. I've replaced the bump-only change w/ one that fixes the docs issue.

Hmm, I shouldn't have committed anything that didn't pass.

dannf:
> > The bump-only change failed to test because the docs test failed. I've replaced the bump-only change w/ one that fixes the docs issue.
> 
> Hmm, I shouldn't have committed anything that didn't pass.

`0.4.6-r2` had a non-empty docs package, while `0.4.6-r3` had an empty one. I've no idea why - maybe we used to install more build-deps by default? At any rate, because of you, we noticed.

PR #42441, opened 2025-02-11
fix: gobump packages is not required field, so make it so it is not.

vaikas:
gobump tool does not require `packages` to be a specified field, so remove a requirement from the go bump pipeline.
https://github.com/chainguard-dev/gobump/blob/main/cmd/gobump/root.go#L34

### DIFF
--- a/pipelines/go/bump.yaml
+++ b/pipelines/go/bump.yaml
@@ -49,16 +49,12 @@ pipeline:
         go mod graph
       fi
 
-      # Packages and replaces are optional, so if not given, we don't pass them.
-      packagesString=""
-      if [ ! -z "${{inputs.deps}}" ]; then
-        packagesString="--packages ${{inputs.deps}}"
-      fi
-
-      replacesString=""
-      if [ ! -z "${{inputs.replaces}}" ]; then
-        replacesString="--replaces ${{inputs.replaces}}"
-      fi
-
       # We use the --tidy flag to run go mod tidy before and after in some cases (if old versions of go are used, we need to update the go.mod format)
-      gobump ${packagesString} ${replacesString} --tidy=${{inputs.tidy}} --skip-initial-tidy=${{inputs.skip-initial-tidy}} --show-diff=${{inputs.show-diff}} --go-version=${GO_VERSION} --compat=${{inputs.tidy-compat}}
+      gobump \
+        --packages="${{inputs.deps}}" \
+        --replaces="${{inputs.replaces}}" \
+        --tidy="${{inputs.tidy}}" \
+        --skip-initial-tidy="${{inputs.skip-initial-tidy}}" \
+        --show-diff="${{inputs.show-diff}}" \
+        --go-version="${GO_VERSION}" \
+        --compat="${{inputs.tidy-compat}}"
--- a/pipelines/go/bump.yaml
+++ b/pipelines/go/bump.yaml
@@ -1,66 +1,54 @@
 name: Bump go deps to a certain version
-
-# Leave go off here, some packages pin to an older version. Let's just assume this gets injected elsewhere.
+# do not include 'go' in 'needs.packages'.
+# some packages pin to an older version; assume it gets injected elsewhere.
 needs:
   packages:
     - git
     - gobump
 
 inputs:
-  deps:
-    description: The deps to bump, space separated
   modroot:
     description: The root of the module
     default: .
+  deps:
+    description: "A space-separated list of packages to update (go bump --packages)"
+  replaces:
+    description: "A space-separated list of packages to replace"
+  bump-file:
+    description: "Filename containing the list of packages to update / replace"
   go-version:
     description: "The go version to set the go.mod syntax to"
     default: ""
-  replaces:
-    description: "The replaces to add to the go.mod file"
   tidy:
     default: true
     description: Run go mod tidy command before and after the bump
+  tidy-compat:
+    description: "Set the go version for which the tidied go.mod and go.sum files should be compatible"
+    default: ""
   skip-initial-tidy:
     default: false
     description: Skip the initial go mod tidy command
   show-diff:
     default: false
     description: Show the difference between the go.mod file before and after the bump
-  tidy-compat:
-    description: "Set the go version for which the tidied go.mod and go.sum files should be compatible"
-    default: ""
   show-dependency-graph:
     default: false
     description: Display a dependency graph of the go.mod file under the modroot
 
 pipeline:
   - runs: |
-      deps="${{inputs.deps}}"
-      replaces="${{inputs.replaces}}"
-      if [ -z "$deps" ] && [ -z "$replaces" ]; then
-        echo "Both 'deps' and 'replaces' are empty.  At least one is required."
-        exit 1
-      fi
-
       cd "${{inputs.modroot}}"
-      # todo rawlingsj: Remove this once we have a new gobump version that does this for us
-      # if ${{inputs.go-version}} is not set, we use the go version of the current environment
-      if [ -z "${{inputs.go-version}}" ]; then
-        GO_VERSION=$(go version | awk '{print $3}' | sed 's/go//')
-      else
-        GO_VERSION=${{inputs.go-version}}
-      fi
-      
+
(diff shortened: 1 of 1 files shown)


pipelines/go/bump.yaml
```diff
@@ -50,5 +49,16 @@ pipeline:
...
+      fi
+
+      replacesString=""
+      if [ ! -z "${{inputs.replaces}}" ]; then
+        replacesString="--replaces ${{inputs.replaces}}"
+      fi
+
       # We use the --tidy flag to run go mod tidy before and after in some cases (if old versions of go are used, we need to update the go.mod format)
-      gobump --packages "${{inputs.deps}}" --replaces "${{inputs.replaces}}" --tidy=${{inputs.tidy}} --skip-initial-tidy=${{inputs.skip-initial-tidy}} --show-diff=${{inputs.show-diff}} --go-version=${GO_VERSION} --compat=${{inputs.tidy-compat}}
+      gobump ${packagesString} ${replacesString} --tidy=${{inputs.tidy}} --skip-initial-tidy=${{inputs.skip-initial-tidy}} --show-diff=${{inputs.show-diff}} --go-version=${GO_VERSION} --compat=${{inputs.tidy-compat}}
```
smoser, line 64: This will have quoting issues if there are more than one "deps" won't it?

smoser (dismissed):

pipelines/go/bump.yaml
```diff
@@ -50,5 +49,16 @@ pipeline:
         go mod graph
       fi
 
+      # Packages and replaces are optional, so if not given, we don't pass them.
+      packagesString=""
```
vaikas, line 53: packagesString = "${{inputs.deps}}"

pipelines/go/bump.yaml
vaikas: go bump ${packageString:+"--packages=${{packageString}"}

smoser:
@vaikas I put 2 implementations in here.
I'm pretty sure either one will work.
1. https://github.com/wolfi-dev/os/blob/4b2464234fcb0b47eb969e16b6045d43a0117ad2/pipelines/go/bump.yaml - this one just _always_ passes both `--packages=` and `--replaces=` to gobump. The expectation is that gobump will read `--packages=` (empty string) the same way it would when no `--packages` was provided at all.
2. https://github.com/wolfi-dev/os/blob/47e064a54f06136a0664dbd22dfa52444fd2c549/pipelines/go/bump.yaml - this version will only pass '--packages=<value>' if 'inputs.deps' has been provided and only pass `--replaces=<value>` if `inputs.replaces` has been passed.  It also checks to make sure that one of them has been provided a non-empty value.

vaikas:
> @vaikas I put 2 implementations in here. I'm pretty sure either one will work.
> 
> 1. https://github.com/wolfi-dev/os/blob/4b2464234fcb0b47eb969e16b6045d43a0117ad2/pipelines/go/bump.yaml - this one just _always_ passes both `--packages=` and `--replaces=` to gobump. The expectation is that gobump will read `--packages=` (empty string) the same way it would when no `--packages` was provided at all.
> 2. https://github.com/wolfi-dev/os/blob/47e064a54f06136a0664dbd22dfa52444fd2c549/pipelines/go/bump.yaml - this version will only pass '--packages=' if 'inputs.deps' has been provided and only pass `--replaces=<value>` if `inputs.replaces` has been passed.  It also checks to make sure that one of them has been provided a non-empty value.

Thank you so much!! Let's go with the first approach. It handles the empty string correctly:
https://github.com/chainguard-dev/gobump/blob/main/cmd/gobump/root.go#L54
There's an extra check for empty string for replaces being empty here:
https://github.com/chainguard-dev/gobump/blob/main/cmd/gobump/root.go#L67

But it actually should not be necessary:
https://pkg.go.dev/strings#Fields

pipelines/go/bump.yaml
```diff
@@ -36,8 +35,14 @@ inputs:
 
 pipeline:
   - runs: |
-      cd "${{inputs.modroot}}"
+      deps="${{inputs.deps}}"
+      replaces=${{inputs.replaces}}
```
sil2100, line 39:
```suggestion
      replaces="${{inputs.replaces}}"
```

For consistency I'd propose quoting both the same way - safer this way.

sil2100 (commented):
I see this version has a bit more checks than the ones I reviewed for enterprise-packages and extra-packages. I think both are fine, since passing in empty `--packages=""` etc. works, but I do like that here we get an error message when none of those arguments are present.

Can we get the same check added to enterprise-packages and extra-packages? Thanks!
One inline comment and we're good to go.

sil2100 (dismissed):
LGTM!

pipelines/go/bump.yaml
```diff
@@ -36,8 +35,14 @@ inputs:
 
 pipeline:
   - runs: |
-      cd "${{inputs.modroot}}"
+      deps="${{inputs.deps}}"
```
vaikas, line 38: I think this is not right:
https://github.com/chainguard-dev/gobump/blob/main/cmd/gobump/root.go#L34
You can give a bumpfile or deps or packages.

But, looks like that input has not been wired through. Just if that gets wired through, need to modify this tool.

smoser: ugh.  
My reading of that code says that 
1. you must give packages, deps (replaces) or bump file
2. you cannot combine packages or deps (replaces) with bump file
3. you can combine packages and deps (replaces)

If you agree that is the correct reading of the code then we can make that change here.

However... I want to get away from logic here. Its kind of nice to have melange catch bad usage, but it is _not_ nice to sync two implementations. I would rather just call `gobump` with `--this=<value> --that=<value>` and expect it to check inputs and give good error messages.

smoser: To further explain , this is what I think the file would look like.
Note I dropped the 'go-version' stuff inline as it looks like it is now supported in gobump ([link](https://github.com/chainguard-dev/gobump/blob/acb8ac8afb89004698b32759a62e5a7161f15785/pkg/update/update.go#L82)).

```yaml
name: Bump go deps to a certain version
# do not include 'go' in 'needs.packages'.
# some packages pin to an older version; assume it gets injected elsewhere.
needs:
  packages:
    - git
    - gobump

inputs:
  modroot:
    description: The root of the module
    default: .
  deps:
    description: "A space-separated list of packages to update (go bump --packages)"
  replaces:
    description: "A space-separated list of packages to replace"
  bump-file:
    description: "Filename containing the list of packages to update / replace"
  go-version:
    description: "The go version to set the go.mod syntax to"
    default: ""
  tidy:
    default: true
    description: Run go mod tidy command before and after the bump
  tidy-compat:
    description: "Set the go version for which the tidied go.mod and go.sum files should be compatible"
    default: ""
  skip-initial-tidy:
    default: false
    description: Skip the initial go mod tidy command
  show-diff:
    default: false
    description: Show the difference between the go.mod file before and after the bump
  show-dependency-graph:
    default: false
    description: Display a dependency graph of the go.mod file under the modroot

pipeline:
  - runs: |
      cd "${{inputs.modroot}}"

      if [ "${{inputs.show-dependency-graph}}" = "true" ]; then
        go mod graph
      fi

      # We use the --tidy flag to run go mod tidy before and after in some cases
      # (if old versions of go are used, we need to update the go.mod format)
      gobump \
        --packages="${{inputs.deps}}" \
        --replaces="${{inputs.replaces}}" \
        --bump-file="${{inputs.bump-file}}" \
        --tidy="${{inputs.tidy}}" \
        --compat="${{inputs.tidy-compat}}" \
        --skip-initial-tidy="${{inputs.skip-initial-tidy}}" \
        --show-diff="${{inputs.show-diff}}" \
        --go-version="${{inputs.go-version}}"
```

vaikas: +1 for not adding the logic here (because it can get out of date). I would prefer this being simpler.

sil2100 (approved):
This looks good. It's actually the same thing I already reviewed and merged for the other repos, minus the addition of the `bump-file` input.

Let's make sure the `bump-file` changes also get backported to enterprise and extra.

PR #42451, opened 2025-02-11
extend the wolfi-base tests a bit

dustinkirkland:
Extend @mattmoor tests of wolfi-base a bit.  Use all of which, type, and command to check for apk and sh (each of which could fail in various ways, such as broken symlinks, etc).  Also, add a few other things to negative check for (e.g. wget and git).

### DIFF
--- a/wolfi-base.yaml
+++ b/wolfi-base.yaml
@@ -25,17 +25,23 @@ test:
   pipeline:
     - name: "Commands exist"
       runs: |
-        for x in apk sh ; do
-          which $x
-          type $x
-          command -v $x
+        for x in "which" "type" "command -v"; do
+          for y in apk sh ; do
+            if ! $x $y; then
+              echo "ERROR: $y missing"
+              exit 1
+            fi
+          done
         done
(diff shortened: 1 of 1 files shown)


wolfi-base.yaml
```diff
@@ -26,18 +26,16 @@ test:
...
       runs: |
-        for x in curl ; do
-          if which $x ; then
-            echo "$x not found"
-            exit 1
-          fi
+        for x in curl git wget; do
+          which $x && echo "$x found but should not be installed" && exit 1
+          type $x && echo "$x found but should not be installed" && exit 1
+          command -v $x && echo "$x found but should not be installed" && exit 1
```
mattmoor, line 38: Are you sure this does what you want?  Unfortunately these won't run presubmit unless there is an epoch bump

dustinkirkland: I reworked these, as a nested loop, with more clear error messages and exit criteria...see what you think in the newer push....

murraybd (approved):

PR #42461, opened 2025-02-11
Adding a provides key to the spark compat

utieyin:
Add provide for spark compat

Fixes:

Related:

#### For new package PRs only

The spark-operator relies on the spark compat package which has now changed name since the refactoring of spark-3.5 into separate scala versions. This change adds a provide key so that when used the package name doesn't lead to issues with advisories

### DIFF
--- a/spark-3.5-scala-2.13.yaml
+++ b/spark-3.5-scala-2.13.yaml
@@ -78,7 +78,7 @@ subpackages:
     description: "Compatibility package to place binaries in the location expected by upstream image"
     dependencies:
       provides:
-        - spark-compat
+        - spark-compat=${{package.full-version}}
     pipeline:
       - runs: |
           mkdir -p "${{targets.subpkgdir}}"/usr/bin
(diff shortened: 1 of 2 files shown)


spark-3.5-scala-2.13.yaml
```diff
@@ -76,6 +76,9 @@ subpackages:
 
   - name: ${{package.name}}-compat
     description: "Compatibility package to place binaries in the location expected by upstream image"
+    dependencies:
+      provides:
+        - spark-compat
```
developer-guy, line 81: is this a valid syntax in melange? should we add this ${{package.full-version}} ? like `spark-compat=${{package.full-version}}`

developer-guy: it looks like its a valid syntax, nice!

developer-guy (dismissed):

EyeCantCU (changes requested):
This will conflict as they're provided at the same version. 2.12 will always be resolved

developer-guy (approved):

PR #42470, opened 2025-02-12
add a font test pipeline and have ~25 pacakges use it

dustinkirkland:
Added a new pipeline for testing fonts.  Ensure that some files get installed in the /usr/share/fonts directory, and use the file command to check their magic file type.  Also, use fonttest to display some information about them.  Fail whenever any of that doesn't work.  Added a bunch of invocations of this to some font packages where this currently works.

### DIFF
--- a/pipelines/test/fonts.yaml
+++ b/pipelines/test/fonts.yaml
@@ -1,0 +1,50 @@
+name: fonts
+
+needs:
+  packages:
+    - file
+    - py3-fonttools
+
+pipeline:
+  - name: font usability test
+    runs: |
+      # Get our font package name
+      font_pkg=$(basename ${{targets.contextdir}})
+      # Find all font files installed by this font_pkg
+      cd /
+      font_files=false
+      # Test font files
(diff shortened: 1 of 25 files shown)


pipelines/test/fonts.yaml
```diff
@@ -0,0 +1,50 @@
...
+needs:
+  packages:
+    - file
+    - py3-fonttools
+
+pipeline:
+  - name: font usability test
+    runs: |
+      # Get our font package name
+      font_pkg=$(basename ${{targets.contextdir}})
```
dannf, line 12: ${{subpkg.name}} FTW

dannf (approved):
thanks!

pipelines/test/fonts.yaml
murraybd: I think there is enough usage of `$(basename ${{targets.contextdir}})` that we should replace these en masse and not hold up this particular PR until this is fixed in it.

PR #42514, opened 2025-02-12
add emptypackage test pipeline

dustinkirkland:
We actually have a couple hundred "empty" packages in wolfi, where "empty" means that the only think shipped in the APK is the spdx json sbom file.

There are some good reasons to have "empty" packages -- such as source packages that ship all of their goodness in subpackages, or virtual packages that just depend on others, etc.  For these, let's at least test that a package that's expected to be empty, stays empty.  Thus, we'll fail if an empty package starts shipping files, unexpectedly.

In a separate PR, I'll submit some tests for our current empty packages, to use this pipeline.  This will help tremendously toward our goal of 100% package test coverage.  (In that, what do you test, in an empty package?)

### DIFF
--- a/font-samyak.yaml
+++ b/font-samyak.yaml
@@ -39,7 +39,14 @@ subpackages:
           cd ${{range.key}}
           mkdir -p "${{targets.subpkgdir}}"/usr/share/fonts/samyak
           install -D -m644 *.ttf -t "${{targets.subpkgdir}}"/usr/share/fonts/samyak
+    test:
+      pipeline:
+        - uses: test/fonts
 
 update:
   enabled: false
   exclude-reason: No releases or tags
+
+test:
+  pipeline:
+    - uses: test/emptypackage
--- a/pipelines/test/emptypackage.yaml
+++ b/pipelines/test/emptypackage.yaml
@@ -6,7 +6,9 @@ pipeline:
       # Get our empty package name
       pkg=$(basename ${{targets.contextdir}})
       # Make sure this is an empty package
-      if [ $(apk info -L "$pkg" | wc -l) -le 3 ]; then
+      # Skip the "contains:" line and any empty lines
+      # We're only really expecting a single spdx.json file in an empty package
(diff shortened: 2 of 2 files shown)


dustinkirkland:
FYI, the following 1-liner will reverse sort all binary packages, based on their package size in bytes:

```$ zcat /var/cache/apk/APKINDEX.a7c09383.tar.gz | grep "^P:\|^V:\|^S:" | tr '\n' ' ' | sed -e "s/ P:/\n/g" -e "s/ [VS]:/ /g" | sort -nr -k3```

Most of the "empty packages", are under 2KB or so.

xnox:
@dustinkirkland we already have a melange linter for this, but maybe it is not working out? do you want to progressively make it an enforced check? and mark up all existing packages as empty as needed? And do you want to be able to run the melange lint against existing packages rather than during build only? Should this replace the melange lint check?

dustinkirkland:
Yeah, I know about the lint check, but don't yet see where it's doing
anything....

First, I want to be able to get to the point where we have 100% test
coverage, on all packages and sub packages.  What does that mean for empty
packages?  What do we test there?  I think we should declare in the package
yaml that we expect it to be empty, and then verify that.  And I think it
should be a breaking change, if something has long been empty, but suddenly
starts including files (without a corresponding declaration in the package
yaml).  I think that should trigger a human to look at it, a la a package
build failure.  For virtual packages (that depend on others), I think we
should also check that the depended upon packages get installed (see
yesterday's wolfi-base tests).

Now, I also suspect that we have some empty packages which are not supposed
to be empty.  They might not be installing stuff correctly, etc.  And this
work could shake those out.  We might also have some useless empty
packages, that should be purged. Again, this work should shake those out.

xnox:
yeah, melange lint doesn't do that today. it is opt-out, and it is non-asserting to be a ratchet the other way around.

murraybd (approved):

murraybd:
> Yeah, I know about the lint check, but don't yet see where it's doing anything.... 

It's not doing anything because the default behavior is set to `Ignore` and subsequently you can't use it with `melange lint`.

PR #42534, opened 2025-02-12
feat: build openssh with pam support

zmarano:
- Compiles openssh with pam support.
- Adds a base line Wolfi sshd pam config and optional subpackage to use it.

TODO: there's an issue with mkhomedir_helper.so where it won't create home directories correctly.

### DIFF
--- a/openssh.yaml
+++ b/openssh.yaml
@@ -1,7 +1,7 @@
 package:
   name: openssh
   version: "9.9_p2"
-  epoch: 0
+  epoch: 1
   description: "the OpenBSD SSH implementation"
   copyright:
     - license: ISC
@@ -19,6 +19,7 @@ environment:
       - busybox
       - ca-certificates-bundle
       - libedit-dev
+      - linux-pam-dev
       - openssl-dev
       - wolfi-baselayout
       - zlib-dev
@@ -55,6 +56,7 @@ pipeline:
          --with-default-path='/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin' \
          --with-privsep-user=sshd \
          --with-ssl-engine \
+         --with-pam \
          --with-libedit
 
   - uses: autoconf/make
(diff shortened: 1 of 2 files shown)


openssh/sshd.pam
a-crate, line 1: Needs `session    required    pam_mkhomedir.so`

https://linux.die.net/man/8/pam_mkhomedir

wolif-baselayout might need to provide /etc/skel for this to work, not sure what happens when it doesn't exist

openssh.yaml
```diff
@@ -194,6 +196,29 @@ subpackages:
...
 
+  - name: "openssh-pam-config"
+    description: "OpenSSH server pam configuration"
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/etc/pam.d
+          mkdir -p "${{targets.subpkgdir}}"/etc/ssh/sshd_config.d
+          mkdir -p "${{targets.subpkgdir}}"/etc/skel # may be needed for the pam_mkhomedir module to work, TBD
+          install -Dm644 sshd.pam "${{targets.subpkgdir}}"/etc/pam.d/sshd
+          echo "UsePAM yes" >> "${{targets.subpkgdir}}"/etc/ssh/sshd_config.d/pam_config
```
a-crate, line 207: Filename needs to match `*.conf` to be included.

zmarano: Done.

openssh.yaml
```diff
@@ -194,6 +196,29 @@ subpackages:
...
+    description: "OpenSSH server pam configuration"
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/etc/pam.d
+          mkdir -p "${{targets.subpkgdir}}"/etc/ssh/sshd_config.d
+          mkdir -p "${{targets.subpkgdir}}"/etc/skel # may be needed for the pam_mkhomedir module to work, TBD
+          install -Dm644 sshd.pam "${{targets.subpkgdir}}"/etc/pam.d/sshd
+          echo "UsePAM yes" >> "${{targets.subpkgdir}}"/etc/ssh/sshd_config.d/pam.conf
+    dependencies:
+      runtime:
```
a-crate, line 209: Should this be here? "openssh-server-config" has no such dependency

zmarano: I was wondering about that too. Both config subpackages are useless without the openssh-server package. Is there a reason we wouldn't have that dependency?

zmarano: Ack

a-crate (approved):

xnox (approved):

PR #42552, opened 2025-02-12
Feature/add melange helpers

smoser:
* Move ldd-check pipeline to use ldd-check
 * Add melange-helpers package

This adds 'melange-helpers' package built from https://github.com/smoser/melange-helpers .
The idea is to move much of the pipelines out of inline shell yaml
and also to add the 'usrmerge-tool' that can help do usrmerge changes.


smoser:
closing in favor of https://github.com/wolfi-dev/os/pull/43220

PR #42560, opened 2025-02-12
symlink java dir instead of copy -r

Aditevil:
Symlink java17 directory instead of copying it over.


kranurag7 (approved):
thank you!! this looks good to me.

PR #42569, opened 2025-02-13
spark-3.5-scala-2.13 invalid dependency bump removal and CVE Fix

jamie-albert:
Spark is a project with very [particular dependency specifications](https://issues.apache.org/jira/browse/SPARK-44114), automation has [added incompatible pombump-dep](https://github.com/wolfi-dev/os/commit/a49a02e057fb2c23fa8612319ec37f8122893c60) and property changes that compile but changes functionality that is required by upstream. Worse, it is not updating the dependencies in a ubiquitous manner. For example updating a specific dependency in a group instead of updating the[ dependency.version structure ](https://github.com/apache/spark/blob/496245314a7d284ed94594051b9f60185a131a67/pom.xml#L135)that is used in this pom.xml. Due to this issue, we should remove spark from the auto cve remediation attempts. This PR needs to wait on this to merge: https://github.com/chainguard-dev/infra-images/pull/2086. Also remediating netty CVE by bumping version. Checked this by going through every change to this version in spark from 96 to 118 and there is no detectible reason why this is being saved for the 4.0.0 release


jamie-albert:
Should wait until this is merged as this PR depends on it: https://github.com/chainguard-dev/infra-images/pull/2086

utieyin (approved):
LGTM

PR #42627, opened 2025-02-13
Complete 3.13 transition

xnox:
Finish bumping priority from 300 to 313 for all remaining py3.13
packages.


dannf (approved):

xnox:
I wonder where elastic builds app is 🥺

PR #42729, opened 2025-02-14
make keycloak-26.1 a version stream package

hectorj2f:
We want to make our current keycloak package a version stream one.


hectorj2f:
Closing this proposed PR for now! We are not including keycloak into our version stream packages.

PR #42758, opened 2025-02-14
fix mime.types issue

developer-guy:
there was an issue while building an image for opencost-ui:

```
2025/02/14 17:53:11 [emerg] 1#1: open() "/etc/nginx/mime.types" failed (2: No such file or directory) in /etc/nginx/nginx.conf:19
```

so this PR aims to fix it.

### DIFF
--- a/opencost-ui.yaml
+++ b/opencost-ui.yaml
@@ -1,7 +1,7 @@
 package:
   name: opencost-ui
   version: 1.114.0
-  epoch: 0
+  epoch: 1
   dependencies:
     runtime:
       - nginx
@@ -14,6 +14,7 @@ environment:
   contents:
     packages:
       - busybox
+      - nginx-package-config
       - nodejs
       - npm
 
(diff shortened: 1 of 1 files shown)


kranurag7 (approved):

opencost-ui.yaml
```diff
@@ -41,12 +42,12 @@ subpackages:
     dependencies:
       provides:
         - nginx-config=0.${{package.full-version}}
```
EyeCantCU, line 44: We need to remove this provides and keep the replaces

EyeCantCU (changes requested):

opencost-ui.yaml
EyeCantCU: Oh wait a moment... I could be wrong. Will follow up in the AM

developer-guy: I can tell you this if I remove the provides its starting to install harbor-2.9-nginx-config during image building :D

EyeCantCU: That's unfortunate and very weird behavior. I'll approve for now, but will be taking a closer look at this

EyeCantCU (approved):

PR #42762, opened 2025-02-14
chore(dotnet-9): Update to 9.0.2

Dentrax:
Update .NET and .NET bootstrap to 9.0.2 (as we were not able to successfully build 9.0.2 with 9.0.101 or 9.0.0 bootstrap)

Microsoft never withdrew the tag for 9.0.101 and never published a tag for 9.0.103. To avoid a withdraw of 9.0.101, mangle the version as 9.0.200

### DIFF
--- a/dotnet-9.yaml
+++ b/dotnet-9.yaml
@@ -1,7 +1,8 @@
+#nolint:git-checkout-must-use-github-updates
 package:
   name: dotnet-9
-  version: 9.0.101
-  epoch: 2
+  version: 9.0.200
+  epoch: 0
   description: ".NET SDK"
   copyright:
     - license: MIT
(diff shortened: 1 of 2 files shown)


dotnet-9.yaml
```diff
@@ -1,7 +1,8 @@
+#nolint:git-checkout-must-use-github-updates
 package:
   name: dotnet-9
-  version: 9.0.101
-  epoch: 2
+  version: 9.0.200
```
mamccorm, line 4: Whats the reason we can't use 9.0.2 here and start fresh with the correct version? it'll be newer than 9.0.101, and means we can continue to use GH to poll for updates?

EyeCantCU: We can't use GitHub to pill for updates because 101 > 2, regardless of the age of the tag. Even if we withdrew 101, we'd still be stuck using release monitor

mamccorm (approved):

PR #42873, opened 2025-02-17
feat(jupyter-base-notebook): add required packages

wojciechka:


maxgio92 (dismissed):

kranurag7 (dismissed):
thanks for extending singleuser test.

jamonation:
How about removing the `/usr/lib/python3.12/site-packages/psutil/tests/__pycache__/*.pyc` files @wojciechka? Those are triggering the scanner - are there other pyc files that are getting copied into the final APK as well?

egibs:
> How about removing the `/usr/lib/python3.12/site-packages/psutil/tests/__pycache__/*.pyc` files @wojciechka? Those are triggering the scanner - are there other pyc files that are getting copied into the final APK as well?

Good call -- the findings are false-positives but this would definitely be a cleaner approach.

egibs:
The malcontent finding is from here: https://github.com/giampaolo/psutil/blob/94e66113c52d6480580d77ce25271032737e0d6b/psutil/tests/test_windows.py#L59-L71

Specifically:
```py
cmdline = (
        "powershell.exe -ExecutionPolicy Bypass -NoLogo -NonInteractive "
        f"-NoProfile -WindowStyle Hidden -Command \"{cmd}\""  # noqa: Q003
    )
```

wojciechka:
> The malcontent finding is from here: https://github.com/giampaolo/psutil/blob/94e66113c52d6480580d77ce25271032737e0d6b/psutil/tests/test_windows.py#L59-L71

@egibs @stevebeattie I added steps that remove the .py and .pyc file to avoid any false positives. PTAL and let me know if this is what you had in mind.

Feel free to approve it if you did, @maxgio92 approved it before, but the fix for malcontent had to be added.

egibs:
> > The malcontent finding is from here: [giampaolo/psutil@`94e6611`/psutil/tests/test_windows.py#L59-L71](https://github.com/giampaolo/psutil/blob/94e66113c52d6480580d77ce25271032737e0d6b/psutil/tests/test_windows.py#L59-L71)
> 
> @egibs @stevebeattie I added steps that remove the .py and .pyc file to avoid any false positives. PTAL and let me know if this is what you had in mind.

LGTM -- I think we get get by with only removing all of the `pycache/` files. I ran a local scan and the actual `.py` file wasn't detected (but I'll double check).

maxgio92 (approved):

PR #42905, opened 2025-02-18
expand contour test coverage in melange

mamccorm:
Expands melange tests to ensure Contour can be launched in a kwok cluster, the process is running, and logging as expected. epoch bump needed, otherwise CI won't run the tests against a binary.


mritunjaysharma394 (approved):
Thanks for doing it!

PR #43096, opened 2025-02-19
rpm-sequoia: introduce new package required by rpm

OddBloke:
Upstream RPM previously shipped their own GPG implementation, which they no longer consider secure or supported.  They've moved to depending on the separate rpm-sequoia project, so we need to package it up.

Related: https://github.com/wolfi-dev/os/pull/43070

#### For new package PRs only

### DIFF
--- a/rpm-sequoia.yaml
+++ b/rpm-sequoia.yaml
@@ -1,0 +1,55 @@
+package:
+  name: rpm-sequoia
+  version: "1.7.0"
+  epoch: 0
+  description: "An OpenPGP backend for rpm using Sequoia PGP"
+  copyright:
+    - license: LGPL-2.0-or-later
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - cargo-auditable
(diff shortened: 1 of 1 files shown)


eslerm (commented):
I do not have any objections and trust that the upstream devs are implementing these security features for the right reason.

I would note the three issues listed on https://fedoraproject.org/wiki/Changes/RpmSequoia#User_Experience and ask if this could affect us.

Nice work @OddBloke !

OddBloke:
> I would note the three issues listed on https://fedoraproject.org/wiki/Changes/RpmSequoia#User_Experience and ask if this could affect us.

I think these could _potentially_ affect users, but I do think it's unlikely. Furthermore, the only option such users would have is to continue using the described-by-its-own-devs-as-insecure PGP backend, so they should really be addressing any issues this might cause already.

> Nice work @OddBloke !

Thanks, and thanks for the review!

kbsteere (approved):

PR #43098, opened 2025-02-19
Fix JDK libs eg freetype

pnasrat:
lost during jre split


EyeCantCU (approved):

dakaneye (commented):
lgtm as well

PR #43114, opened 2025-02-19
postgresql-*: reenable automatic updates

OddBloke:
The reason given for requiring manual updates is that package builds might fail if a given value isn't kept up-to-date.  However, it's preferable to only have to manually intervene when that value requires updating, rather than for every new upstream version.

### DIFF
--- a/postgresql-16.yaml
+++ b/postgresql-16.yaml
@@ -279,10 +279,6 @@ subpackages:
 update:
   version-separator: _
   enabled: true
-  # Requires manual steps when updating because of usage of ALPINE_VERSION: "3.21" while pulling docker-entrypoint scripts
-  # which gets updated in https://github.com/docker-library/postgres.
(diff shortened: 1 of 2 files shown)


mritunjaysharma394:
Cc @mamccorm as I think you suggested to have manual as true for this one but honestly it might be one off a change among many version updates so I also agree but that also leads to either verifying everytime the alpine version or a possible risk of honk

OddBloke:
> Cc @mamccorm as I think you suggested to have manual as true for this one but honestly it might be one off a change among many version updates so I also agree but that also leads to either verifying everytime the alpine version or a possible risk of honk

Thanks for the quick review! I think the testing that you added when making it `manual: true` means we should see package build failures if the entrypoint scripts don't get installed properly, which should hopefully avoid the risk of honk.

kbsteere (approved):

PR #43139, opened 2025-02-20
py3-netifaces: Use ${{vars.import}} for module name in import test

dannf:


justinvreeland (approved):

dannf:
I was using this as my "new best example" for a new python package, and noticed this nit

PR #43163, opened 2025-02-20
Complete 3.13 transition

xnox:
Finish bumping priority from 300 to 313 for all remaining py3.13
packages, apart from supported-python.


sil2100 (approved):
I can't think of a reason not to do this. Thanks!

PR #43165, opened 2025-02-20
Complete 3.13 transition for py3-supported-python

xnox:
Bumping priority from 300 to 313 on py3-supported-python.

### DIFF
--- a/py3-supported-python.yaml
+++ b/py3-supported-python.yaml
@@ -1,7 +1,7 @@
 package:
   name: py3-supported-python
   version: 1
-  epoch: 4
+  epoch: 5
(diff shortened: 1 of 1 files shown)


murraybd:
@xnox Do you know why this is building in staging but not production? I'd think the necessary changes would be in production by now...

xnox:
> @xnox Do you know why this is building in staging but not production? I'd think the necessary changes would be in production by now...

I would have thought so too... But we are possibly in deployment freeze.

xnox:
I will reopen internal escalation.

xnox:
Let's try again in:
- https://github.com/wolfi-dev/os/pull/47572

PR #43209, opened 2025-02-20
groff - split groff-base package out like debian and avoid perl dep.

smoser:
installation of groff pulls in perl.
aws-cli-v2 uses groff to render help, and thus using it requires perl which is less than ideal.

Debian has a 'groff-base' package.  I've taken the list of files in their groff-base package and moved them into a groff-base package here, with a couple exceptions
1. our groff package did not (and do not) ship 3 symlinks (geqn gpic and gtbl). Each points to the non 'g' version (geqn -> eqn). I did not start creating those.
2. grog is a perl script and I wanted to avoid perl dependency, so I left it in the full package (groff).

### DIFF
--- a/aws-cli-2.yaml
+++ b/aws-cli-2.yaml
@@ -3,7 +3,7 @@
 package:
   name: aws-cli-2
   version: "2.24.10"
-  epoch: 0
+  epoch: 1
   description: "Universal Command Line Interface for Amazon Web Services (v2)"
   copyright:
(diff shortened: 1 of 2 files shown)


dannf (dismissed):

smoser:
rebased to pick up test fix in #43222

dannf (dismissed):

dustinkirkland (approved):

PR #43210, opened 2025-02-20
Package the licenseclassifier tool

sil2100:
New package. We most likely won't be using this tool for license linting, but it's good to have it around for testing purposes. To compare results.


licenseclassifier.yaml
```diff
@@ -0,0 +1,42 @@
...
+    with:
+      modroot: .
+      packages: ./tools/identify_license
+      output: licenseclassifier
+
+update:
+  enabled: true
+  github:
+    identifier: google/licenseclassifier
+    strip-prefix: v
```
dannf, line 33: I see the repo has tags w/ "pre" and "alpha" that we probably want to ignore

dannf (changes requested):

licenseclassifier.yaml
sil2100: Good catch! Thanks!

dannf (approved):

PR #43219, opened 2025-02-20
Libffi changes

utieyin:
Recompiling libffi with `fPIC`

Fixes:

Related:


utieyin:
The blast radius for this change is large and it was agreed to keep it to where it is currently needed which is the pypy bootstrap

PR #43257, opened 2025-02-20
feat: add polkit and duktape packages to wolfi

zmarano:
polkit is needed for certain functionality in systemd to correctly function. duktape is a dependency to build polkit. We are not establishing further polkit policies other than the default ones from upstream in this package.

### DIFF
--- a/duktape.yaml
+++ b/duktape.yaml
@@ -1,0 +1,36 @@
+package:
+  name: duktape
+  version: "2.7.0"
+  epoch: 0
+  description: "Embeddable Javascript engine with a focus on portability and compact footprint"
+  copyright:
+    - license: MIT
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+
+pipeline:
+  - uses: fetch
+    with:
+      expected-sha256: 90f8d2fa8b5567c6899830ddef2c03f3c27960b11aca222fa17aa7ac613c2890
+      uri: https://github.com/svaarala/duktape/releases/download/v${{package.version}}/duktape-${{package.version}}.tar.xz
+
+  - runs: |
+      DESTDIR=${{targets.destdir}} INSTALL_PREFIX=/usr make -f Makefile.sharedlibrary install
+
+update:
+  enabled: true
+  github:
+    identifier: svaarala/duktape
+    strip-prefix: v
+
+test:
+  pipeline:
+    - uses: test/ldd-check
+      with:
+        packages: ${{package.name}}
+    - uses: test/pkgconf
--- a/polkit.yaml
+++ b/polkit.yaml
@@ -1,0 +1,102 @@
+package:
+  name: polkit
+  version: "126"
+  epoch: 0
+  description: "A toolkit for defining and handling authorizations."
+  copyright:
+    - license: LGPL-2.0-or-later
+  dependencies:
+    runtime:
+      - dbus
+      - linux-pam
+      - systemd
+      - wolfi-base
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - cmake
+      - dbus-dev
+      - duktape
+      - expat-dev
+      - gettext-dev
+      - glib-dev
+      - glib-gir
+      - gobject-introspection-dev
+      - linux-pam-dev
+      - meson
+      - perl
+      - systemd-dev
+
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: d627b0d1e1108563658dabe3fb8d2a065e64df10
+      repository: https://github.com/polkit-org/polkit.git
+      tag: ${{package.version}}
+
+  - uses: meson/configure
+    with:
+      opts: |
+        -Dman=true \
+        -Dpam_prefix=/etc/pam.d \
+        -Dpam_module_dir=/usr/lib/security
+
+  - uses: meson/compile
+
+  - uses: meson/install
+
+subpackages:
+  - name: polkit-doc
+    pipeline:
+      - uses: split/manpages
+    dependencies:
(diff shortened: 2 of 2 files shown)


duktape.yaml
```diff
@@ -0,0 +1,35 @@
...
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+
+pipeline:
+  - uses: fetch
+    with:
+      expected-sha256: 90f8d2fa8b5567c6899830ddef2c03f3c27960b11aca222fa17aa7ac613c2890
+      uri: https://github.com/svaarala/duktape/releases/download/v${{package.version}}/duktape-${{package.version}}.tar.xz
```
xnox, line 20: Is this manually created download artifact with generated things or stuff not in the git repo, or a straight up GitHub tag export?

If it just git tag export, we prefer to build from a git-checkout.

polkit.yaml
```diff
@@ -0,0 +1,76 @@
...
+      - dbus
+      - linux-pam
+      - shadow
+      - systemd
+      - wolfi-base
+  scriptlets:
+    post-install: |
+      #!/bin/sh
+      systemd-sysusers
+      chgrp -R polkitd /etc/polkit-1/rules.d
```
xnox, line 19: Scripts are not executed during image builds, thus ensure there is tmfiles.d snippet to the chgrp effect. Didn't check if there is one already.

Also we try not to use scriplets at all. Despite the interactive testing convenience.

polkit.yaml
```diff
@@ -0,0 +1,76 @@
...
+      - shadow-dev
+      - systemd
+      - systemd-dev
+
+pipeline:
+  - uses: fetch
+    with:
+      expected-sha256: 2814a7281989f6baa9e57bd33bbc5e148827e2721ccef22aaf28ab2b376068e8
+      uri: https://github.com/polkit-org/polkit/archive/refs/tags/${{package.version}}.tar.gz
+
```
xnox, line 48: Please switch to git-checkout if possible.

polkit.yaml
```diff
@@ -0,0 +1,76 @@
...
+  - uses: meson/configure
+    with:
+      opts: |
+        -Dman=true \
```
xnox, line 52: Please add docs subpackage and split/docs into it.

polkit.yaml
zmarano: Hmm nevermind, it somehow seems to just work without it. So yeah, removed the post-install scriptlet.

justinvreeland (dismissed):

polkit.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+  - uses: meson/compile
+
+  - uses: meson/install
+
+subpackages:
+  - name: polkit-doc
+    pipeline:
+      - uses: split/manpages
+    description: polkit manpages
```
xnox, line 56: we need a few more subpackage splits here.

we need to do split-dev such that all the .pc files are split.
Also we need to split locales/translations too.
As otherwise the package is heavy.

To ensure polkit starts in priviledged containers with systemd, preset snippet is needed too please.

something to ponder if we want to split runtime libraries out as well, or not. Likely not worth it.

xnox (dismissed):
it works, can merge as is.

would be nice to get locales & -dev subpackage splits though. Up to you if you want to dissmiss that comment and merge as is; or to do it now.

polkit.yaml
zmarano: Thanks! Done.

zmarano:
> it works, can merge as is.
> 
> would be nice to get locales & -dev subpackage splits though. Up to you if you want to dissmiss that comment and merge as is; or to do it now.

Split out locales and dev- much better. (still learning what these things do- thanks for the pointer)

polkit.yaml
```diff
@@ -0,0 +1,93 @@
...
+        -Dpam_module_dir=/usr/lib/security
+
+  - uses: meson/compile
+
+  - uses: meson/install
+
+subpackages:
+  - name: polkit-doc
+    pipeline:
+      - uses: split/manpages
```
murraybd, line 55: Can you add the test/docs pipeline here too?

zmarano: Done!

polkit.yaml
```diff
@@ -0,0 +1,96 @@
...
+      pipeline:
+        - uses: test/docs
+
+  - name: polkit-dev
+    pipeline:
+      - uses: split/dev
+    dependencies:
+      runtime:
+        - polkit
+    description: polkit dev
```
murraybd, line 70: It looks like this could use a test:

```
2025-02-21T21:20:43Z WARN 🔗 linter "pkgconf" failed on package "polkit-dev": pkgconfig directory found: usr/lib/pkgconfig/polkit-agent-1.pc; suggest: This package provides files in a pkgconfig directory, please add the pkgconf test pipeline
```

duktape.yaml
```diff
@@ -0,0 +1,35 @@
...
+  enabled: true
+  github:
+    identifier: svaarala/duktape
+    strip-prefix: v
+
+test:
+  pipeline:
+    - uses: test/ldd-check
+      with:
+        packages: ${{package.name}}
```
murraybd, line 35: This could also use a pkgconf test:

```
2025-02-21T21:20:09Z WARN 🔗 linter "pkgconf" failed on package "duktape": pkgconfig directory found: usr/lib/pkgconfig/duktape.pc; suggest: This package provides files in a pkgconfig directory, please add the pkgconf test pipeline
```

polkit.yaml
zmarano: Thanks, done.

duktape.yaml
zmarano: Thanks, done.

xnox (approved):

PR #43316, opened 2025-02-21
fix(rsyslog): do not ship files that may clash at build time

89luca89:
Move them to a scriptlet


smoser (changes requested):
better to just change rsyslog to keep state in /run ?

smoser:
https://github.com/wolfi-dev/os/pull/43318  is what i'm suggesting.

89luca89:
Let's do your way

PR #43362, opened 2025-02-21
chore(llvm): Build the whole LLVM stack together

EyeCantCU:
This PR refactors LLVM 19 and all of its components to build together within the same stack increasing maintainability

Other runtimes we previously built have been added as well (sans OpenMP for now due to issues with OpenSSF flags - we need to build Clang with those). This also introduces Clang's compiler runtime

Finally, libraries that are ABI compatible and share a consistent version between releases of LLVM are now built in the stack. Virtual provides have been added for compatibility, but will be progressively phased out


llvm.yaml
```diff
@@ -0,0 +1,830 @@
...
+          mkdir -p ${{targets.contextdir}}/${{vars.llvm-prefix}}/lib/
+          mv ${{targets.destdir}}/${{vars.llvm-prefix}}/lib/$soname ${{targets.contextdir}}/usr/lib/
+          ln -s $soname ${{targets.contextdir}}/usr/lib/$sonamefull
+          ln -sf ../../$soname ${{targets.destdir}}/${{vars.llvm-prefix}}/lib/$soname
+          ln -sf ../../$soname ${{targets.destdir}}/${{vars.llvm-prefix}}/lib/$sonamefull
+    test:
+      pipeline:
+        - uses: test/ldd-check
+          with:
+            packages: $(basename ${{targets.contextdir}})
```
murraybd, line 137: You could use ${{subpkg.name}} here.

EyeCantCU: Thank you, done!

jamonation:
@murraybd - is there anything preventing this from being merged from the OS side of things? 

e.g. I see rust 1.8x needs `libLLVM-19` at runtime, which this provides. Likewise clang-19, llvm-19 and llvm-19-dev.

llvm.yaml
```diff
@@ -0,0 +1,830 @@
...
+
+  - name: libclang-cpp
+    description: Clang ${{vars.major-version}} runtime library
+    dependencies:
+      provides:
+        - libclang-cpp-${{vars.major-version}}=${{package.full-version}}
+      runtime:
+        - libLLVM
```
murraybd, line 145: We've seen some issues with abseil-cpp that leads us to believe depending on the specific version would be the safest thing e.g. `${{package.full-version}}` . See https://github.com/chainguard-dev/internal-dev/issues/9394#issuecomment-2688648434 for an example.

llvm.yaml
```diff
@@ -0,0 +1,830 @@
...
+          mkdir -p ${{targets.contextdir}}/${{vars.llvm-prefix}}/share
+          mv ${{targets.destdir}}/${{vars.llvm-prefix}}/share/scan-* ${{targets.contextdir}}/${{vars.llvm-prefix}}/share/
+
+          mkdir -p ${{targets.contextdir}}/${{vars.llvm-prefix}}/lib
+          mv ${{targets.destdir}}/${{vars.llvm-prefix}}/lib/libear* ${{targets.contextdir}}/${{vars.llvm-prefix}}/lib/
+    test:
+      pipeline:
+        - runs: |
+            scan-build --help
+            scan-view --help
```
murraybd, line 211: It looks like there are also binaries for `analyze-build` and `intercept-build` do those have arguments like `--help` or `--version` that we could test?

llvm.yaml
```diff
@@ -0,0 +1,830 @@
...
+            pp-trace --version
+            pp-trace --help
+
+  - name: py3-clang
+    description: Clang ${{vars.major-version}} Python bindings
+    dependencies:
+      provides:
+        - py3-clang-${{vars.major-version}}=${{package.full-version}}
+      runtime:
+        - libLLVM
```
murraybd, line 315: Again, I think specifying the version here would be safest.

llvm.yaml
EyeCantCU: Will do!

murraybd (dismissed):

murraybd (dismissed):

murraybd (approved):

PR #43487, opened 2025-02-23
expand tests using dex oidc with oauth2-proxy

developer-guy:
Expand the tests by providing a OIDC integration with dex


oauth2-proxy.yaml
```diff
@@ -46,7 +46,51 @@ update:
...
+        staticClients:
+        - id: oauth2-proxy
+          redirectURIs:
+          - 'http://127.0.0.1:4180/oauth2/callback'
+          name: 'oauth2-proxy'
+          secret: proxy
+        EOF
+
+        dex serve dex.yaml &
+        sleep 5
```
EyeCantCU, line 81: This seems like a good place to use `test/daemon-check-output` so that package tests don't fail if we time things wrong

oauth2-proxy.yaml
```diff
@@ -46,7 +46,51 @@ update:
...
+            --client-id oauth2-proxy \
+            --client-secret proxy \
+            --redirect-url http://127.0.0.1:4180/oauth2/callback \
+            --oidc-issuer-url http://127.0.0.1:5556/dex \
+            --cookie-secure=false \
+            --cookie-secret=$(head -c 16 /dev/urandom | base64) \
+            --email-domain kilgore.trout \
+            --upstream file://$PWD/#/static/ &
+
+         sleep 3
```
EyeCantCU, line 94: Here as well

EyeCantCU (changes requested):

EyeCantCU (approved):

PR #43498, opened 2025-02-24
fix(ffmpeg-dev): Always pull in ffmpeg

EyeCantCU:
Its possible to resolve to ffmpeg-cuda instead of ffmpeg when library versions align so explicitly depend on ffmpeg

Also set provider priority going forward


kranurag7 (approved):
thank you!!

PR #43503, opened 2025-02-24
libffi: move pic build into its own package

dannf:


utieyin (approved):
Confirmed that this works for building pypy.
LGTM

PR #43509, opened 2025-02-24
Refactor hydra; use go/build

anushkamittal20:
refactors hydra pkg

### DIFF
--- a/hydra.yaml
+++ b/hydra.yaml
@@ -39,6 +39,8 @@ update:
 
 test:
   environment:
+    environment:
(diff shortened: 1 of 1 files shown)


kranurag7 (commented):
I think we can also consider removing explicit `CGO_ENABLED` via this patch, that is not required.

kranurag7:
there's also a conflict coming up, please rebase with main.

kranurag7 (approved):
thanks!!

PR #43554, opened 2025-02-24
tw: Fix typo in comment block

dannf:


smoser (approved):
i would also have approved of deleting the comment block. its not really useful.

PR #43577, opened 2025-02-24
add tzdata

chainkuti:
update kube-state-metrics to include tzdata

### DIFF
--- a/kube-state-metrics.yaml
+++ b/kube-state-metrics.yaml
@@ -1,7 +1,7 @@
 package:
   name: kube-state-metrics
   version: "2.15.0"
-  epoch: 1
+  epoch: 2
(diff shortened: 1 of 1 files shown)


developer-guy (commented):
This is just a build time dependency you defined and its not gonna be included into our final package here, instead you should define tzdata as a runtime dependency to make it available in the final image

kube-state-metrics.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: kube-state-metrics
   version: "2.15.0"
-  epoch: 1
+  epoch: 2
```
developer-guy, line 4:
```
dependencies:
  runtime:
    - tzdata
```

jamonation (approved):

PR #43594, opened 2025-02-24
fix(py3-cassandra-medusa): Vendor poetry in virtual environment

EyeCantCU:
Medusa depends on Poetry 1.x, and our Poetry has moved on to 2.x. Vendor 1.x in the virtual environment

https://github.com/thelastpickle/cassandra-medusa/pull/844

### DIFF
--- a/py3-cassandra-medusa.yaml
+++ b/py3-cassandra-medusa.yaml
@@ -2,7 +2,7 @@
 package:
   name: py3-cassandra-medusa
   version: "0.23.0"
-  epoch: 2
+  epoch: 3
   description: Apache Cassandra backup and restore tool
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


py3-cassandra-medusa.yaml
```diff
@@ -48,35 +45,34 @@ pipeline:
       POETRY_VIRTUALENVS_IN_PROJECT=true poetry install
       poetry build
 
-  - runs: |
       # Setup the virtualenv
-      python${{vars.py-version}} -m venv .venv --system-site-packages
+      python${{vars.py-version}} -m venv .venv
```
mamccorm, line 49: I've reviewed and can't think of any case we might need '--system-site-packages' for this package ,but checking this change was intentional cleanup?

EyeCantCU: @mamccorm it's intentional. We install poetry above as a precursor to building medusa. If we enable site packages, Python picks up on Poetry and doesn't install it inside the virtual environment when we reinstall below. It's just a difference in the config file, and we enable system site packages after installing Poetry in the virtual environment

mamccorm (approved):

PR #43604, opened 2025-02-25
expand tests for descheduler package

mamccorm:
Small improvement to package tests. Previously always downloaded an example policy file from the 'master' branch, now pulls in the example policy from the GitHub release we're building.

Tests now attempt to create a descheduler policy.


kranurag7 (approved):
thank you!! looks good to me.

PR #43370, opened 2025-02-21
Allow tests to pass extra library paths to `ldd-check` & fix neon

sergiodj:
neon is currently failing `ldd-check` because it installs libraries under non standard locations (`/usr/libexec`).  We'll need to pass these paths to `LD_LIBRARY_PATH` in order to properly check these libs.

I decided to implement a new parameter to `ldd-check` called `extra_library_paths`, which takes a colon-separated list of paths that will be added to `LD_LIBRARY_PATH` before running tests.

This PR also includes a commit to enable `ldd-check` on neon and make sure that it passes the correct paths via `extra_library_paths`.

Fixes: #40839 
Relates: #40629

PR #43608, opened 2025-02-25
Improve neon by adjusting `rpath` while installing PostgreSQL libs/bins

### LINKED ISSUE #43607: neon installs binaries/libs that need `rpath` adjustment
neon currently installs PostgreSQL binaries/libraries (for several pgsql versions) under `/usr/libexec/neon`.  This requires `rpath` adjustment during link time in order to make these artifacts usable, otherwise the user is supposed to fiddle with environment variables to make everything work.

Related to #43370

sergiodj:
This PR improves the neon package and make its PostgreSQL binaries and libraries to be usable out-of-the-box by adjust their `rpath`.

Unfortunately I had to use the full path while adjust `rpath` because of the many files being installed under nested directories.  In order to be able to use `$ORIGIN` I'd have to take note of the subdir level and adjust `rpath` accordingly, which is too much work for too little gain.

Fixes: #43607 
Relates: #43370

### DIFF
--- a/neon.yaml
+++ b/neon.yaml
@@ -1,7 +1,7 @@
 package:
   name: neon
   version: "8005"
-  epoch: 0
+  epoch: 1
   description: "Serverless Postgres. We separated storage and compute to offer autoscaling, branching, and bottomless storage."
   copyright:
     - license: Apache-2.0
--- a/neon.yaml
+++ b/neon.yaml
@@ -29,6 +29,7 @@ environment:
       - libseccomp-dev
       - libtool
       - openssl-dev
+      - patchelf
       - perl
       - pkgconf
       - poetry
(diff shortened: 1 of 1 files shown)


dannf:
Just curious, would you still want to use this mechanism if https://github.com/wolfi-dev/os/pull/43700 was available?

sergiodj:
> Just curious, would you still want to use this mechanism if #43700 was available?

I don't believe so, no.  I left a comment on #43700 and mentioned that I was thinking about `ld.so.conf.d` when I wrote this neon hack, but noticed that we didn't support it.  Thanks for the PR.

Once that lands I will withdrawal this PR.

sergiodj:
> > Just curious, would you still want to use this mechanism if #43700 was available?
> 
> I don't believe so, no. I left a comment on #43700 and mentioned that I was thinking about `ld.so.conf.d` when I wrote this neon hack, but noticed that we didn't support it. Thanks for the PR.
> 
> Once that lands I will withdrawal this PR.

@dannf I spoke too soon.

Unfortunately using `/etc/ld.so.conf.d` snippets won't work here, because `ld` will happily load the first `libpq.so.5` library it finds, which ends up always being `v14`.  This leads to situations like:

```
# ldd /usr/libexec/neon/v15/bin/psql
        linux-vdso.so.1 (0x00007ffcb6945000)
        libpq.so.5 => /usr/libexec/neon/v14/lib/libpq.so.5 (0x00007a8163d5e000)
        libreadline.so.8 => /usr/lib/libreadline.so.8 (0x00007a8163d05000)
        libm.so.6 => /lib/libm.so.6 (0x00007a8163c22000)
        libc.so.6 => /lib/libc.so.6 (0x00007a8163a30000)
        libssl.so.3 => /usr/lib/libssl.so.3 (0x00007a8163933000)
        libcrypto.so.3 => /usr/lib/libcrypto.so.3 (0x00007a8163400000)
        libncursesw.so.6 => /usr/lib/libncursesw.so.6 (0x00007a81633c2000)
        /lib64/ld-linux-x86-64.so.2 (0x00007a8163e6c000)
        libtinfo.so.6 => /usr/lib/libtinfo.so.6 (0x00007a816337e000)
```

Note how `v15` `psql` loads `v14` `libpq.so.5`.

So yeah, this PR is still valid.

dannf (dismissed):
Makes sense. It is rather annoying to have to do this when, I presume, the application DTRT, and we're just working around a test. But I don't have a better idea, and the way you've done it is easy to understand.

One suggestion - somewhat orthogonal to your changes. I see you notice that 17 was missing, and added it. How will we catch this when 18 is added? Perhaps we could could `exit 1` if we find a version we don't recognize? Or, maybe we just run it for all `pg_install/v*` ?

sergiodj:
> Makes sense. It is rather annoying to have to do this when, I presume, the application DTRT, and we're just working around a test. But I don't have a better idea, and the way you've done it is easy to understand.

Thanks.

I believe the actual reason why we're doing this ugly dance here is because we want the package to be completely usable out of the box.  While it is possible (and even recommended by upstream) for users to initialize neon with `POSTGRES_DISTRIB_DIR` set to the location where the PostgreSQL installation is, it should still be possible to invoke the standalone binaries without getting an error.

I just wanted to make sure that we're on the same page w.r.t. the reasoning behind these changes, but please let me know what you think.

> One suggestion - somewhat orthogonal to your changes. I see you notice that 17 was missing, and added it. How will we catch this when 18 is added? Perhaps we could could `exit 1` if we find a version we don't recognize? Or, maybe we just run it for all `pg_install/v*` ?

Good idea.  I think I prefer the latter.  Let me adjust the package.

dannf (approved):

PR #43612, opened 2025-02-25
glibc: Run `make check`

dannf:
https://github.com/chainguard-dev/internal-dev/issues/9039

### DIFF
--- a/glibc.yaml
+++ b/glibc.yaml
@@ -37,10 +37,12 @@ environment:
     keyring:
       - https://packages.wolfi.dev/bootstrap/stage3/wolfi-signing.rsa.pub
     packages:
+      - bash # needed by `make check`
       - bison
       - build-base
       - busybox
       - ca-certificates-bundle
+      - coreutils # needed by `make check`
       - flex
       - gawk
       - grep
@@ -112,6 +114,9 @@ pipeline:
 
   - runs: |
       make -C build -j$(nproc) install DESTDIR="${{targets.destdir}}"
+
+  - runs: |
+      make -C build -j$(nproc) check
 
   - name: "Set up ldconfig"
     runs: |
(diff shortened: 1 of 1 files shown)


dannf:
Current state when running on an x86 workstation instance:

```
2025/02/25 05:37:04 INFO         === Summary of results ===
2025/02/25 05:37:04 INFO     104 FAIL
2025/02/25 05:37:04 INFO    5242 PASS
2025/02/25 05:37:04 INFO      33 UNSUPPORTED
2025/02/25 05:37:04 INFO      16 XFAIL
2025/02/25 05:37:04 INFO       4 XPASS
```

sil2100 (dismissed):
Thank you dann! Pushed a change on top of yours that runs the testsuite but not blocking on failures. I know this is not very useful right now, but I treat it as a temporary solution before we get a proper baseline set. The pipeline step is named so it will be easy to grep in the logs, and I want to add a manual step to go through the failures for now. It's like a non-blocking lint right now.

glibc.yaml
```diff
@@ -113,6 +115,12 @@ pipeline:
   - runs: |
       make -C build -j$(nproc) install DESTDIR="${{targets.destdir}}"
 
+  - name: "Run the glibc testsuite, not blocking on failures for now"
+    runs: |
+      # For now we run the glibc testsuite for manual investigation.
+      # Later on we'll establish a baseline and make regressions fail the build.
+      make -C build -j$(nproc) check || true
```
xnox, line 122: can we please store the testsuite.logs and package them into an apk? can be into -dev or a new one like -testresults?

As the builds are long, and it is customary to submit those to upstream with bug reports. Or just in general when testing RC builds.

Similarly for gcc, if we start running test-suite there, package it such that one can email it to upstream to submit test results.

sil2100: I like that idea. Let me create a `glibc-testresults` package for this!

sil2100 (approved):
Already used this during 2.41 testing! Let's get it landed today.

xnox (approved):

xnox:
@sil2100 approved. But choose when to land it.

PR #43700, opened 2025-02-25
Add /etc/ld.so.conf.d support

dannf:
We have several packaged libraries that install in non-standard paths. The CUDA libs are the primary example. Today, we have to propagate these paths up through various levels of packaging, in both build and test sections, and ultimately into images by way of `LD_LIBRARY_PATH` settings. This is error-prone.

Packages that rely on non-standard library paths can drop `.conf` snippets into this directory. This relies on the existing support in `apk` for running `ldconfig` after package installation, and similar support in `apko` when preparing an image filesystem.

### DIFF
--- a/glibc.yaml
+++ b/glibc.yaml
@@ -1,7 +1,7 @@
 package:
   name: glibc
   version: 2.40
-  epoch: 9
+  epoch: 10
   description: "the GNU C library"
   copyright:
     - license: LGPL-2.1-or-later
--- a/glibc.yaml
+++ b/glibc.yaml
@@ -123,7 +123,10 @@ pipeline:
 
   - name: "Set up ldconfig"
     runs: |
-      cp vendor/ld.so.conf "${{targets.destdir}}"/etc/ld.so.conf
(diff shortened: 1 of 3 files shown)


sergiodj (changes requested):
This is great.  Thanks.

I was thinking about `ld.so.conf.d` when I was doing #43370 .  It will certainly help.

I'm leaving a few minor suggestions, but otherwise LGTM.

sil2100 (dismissed):
This looks great, feels to work as expected. Thank you for the test case!

sergiodj (dismissed):
LGTM, thanks!

dannf:
`ingress-nginx-controller-*` seems to be the principle modifier of `/etc/ld.so.conf`. I prepared PRs to bump that package to build against this glibc:
https://github.com/wolfi-dev/os/pull/43875
https://github.com/chainguard-dev/enterprise-packages/pull/15489

And I manually built/tested both image sets on x86. They succeeded w/o error.

sergiodj (approved):

PR #43795, opened 2025-02-25
New package: liquibase-package-manager

mjramirez-dev:
Adding Liquibase-package-manager tool. Additional package needed to match upstream liquibase.
Related:
- https://github.com/chainguard-dev/image-requests/issues/5181

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/liquibase-package-manager.yaml
+++ b/liquibase-package-manager.yaml
@@ -13,6 +13,12 @@ pipeline:
       tag: v${{package.version}}
       expected-commit: 1cd39ecef00dee2b6cb800acd2d2410f06179231
 
+  - name: Prebuild setup
+    runs: |
+      # The version in file: liquibase-package-manager/internal/app/VERSION is statically set to 0.2.3
+      # Need to run 'make updateVersion' before performing go build to set proper lmp release version
+      make updateVersion
+
   - uses: go/build
     with:
       packages: ./cmd/lpm/darwin.go
@@ -30,21 +36,40 @@ update:
 
 test:
   environment:
+    environment:
+      JAVA_HOME: /usr/lib/jvm/java-17-openjdk
+      REPO: liquibase/liquibase
+      LIQUIBASE_HOME: /liquibase
     contents:
       packages:
         - go
         - curl
         - git
+        - openjdk-17
   pipeline:
-    - name: Run test suite
+    - name: Setup
       runs: |
         # Clone lpm repo for test suite access
         git clone -b v${{package.version}} https://github.com/liquibase/liquibase-package-manager
+
+        # Setup liquibase home dir
+        mkdir -p /liquibase
+
+        # Get liquibase upstream source
+        export VERSION=$(curl -s https://api.github.com/repos/$REPO/releases/latest | grep '"tag_name"' | cut -d '"' -f 4)
+        export LATEST_RELEASE_URL=$(curl -s https://api.github.com/repos/$REPO/releases/latest | grep "browser_download_url" | grep ".tar.gz" | cut -d '"' -f 4)
+        curl -Ls $LATEST_RELEASE_URL -o liquibase-$VERSION.tar.gz
+
+        # Extract liquibase
+        tar -xzvf liquibase-$VERSION.tar.gz -C /liquibase
+    - name: Run test suite
+      runs: |
+        # Navigate to lpm dir
         cd ${{package.name}}
 
         # Install prereqs
-        go install honnef.co/go/tools/cmd/staticcheck@latest
+        make test-setup
 
         # Run test suite
-        export PATH=$PATH:$(go env GOPATH)/bin
-        make test
+        export PATH=$PATH:$JAVA_HOME/bin
+        make e2e
--- a/liquibase-package-manager.yaml
+++ b/liquibase-package-manager.yaml
@@ -6,11 +6,6 @@ package:
   copyright:
     - license: Apache-2.0 AND MPL-2.0
 
-environment:
-  contents:
-    packages:
-      - go
-
 pipeline:
   - uses: git-checkout
     with:
--- a/liquibase-package-manager.yaml
+++ b/liquibase-package-manager.yaml
@@ -24,14 +24,6 @@ pipeline:
       packages: ./cmd/lpm/darwin.go
       output: lpm
 
-subpackages:
-  - name: ${{package.name}}-compat
-    description: "Compatibility package to place binaries in the location expected by lpm"
-    pipeline:
-      - runs: |
-          mkdir -p "${{targets.contextdir}}/bin"
-          ln -sf /usr/bin/lpm ${{targets.contextdir}}/bin/lpm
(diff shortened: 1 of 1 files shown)


liquibase-package-manager.yaml
```diff
@@ -0,0 +1,56 @@
...
+  epoch: 0
+  description: "Easily manage external dependencies for Database Development. Search for, install, and uninstall liquibase drivers, extensions, and utilities."
+  copyright:
+    - license: Apache-2.0 AND MPL-2.0
+
+environment:
+  contents:
+    packages:
+      - go
+
```
jamonation, line 13:
```suggestion
```

Don't need to be explicit here, the `go/build` pipeline will do the right thing

jamonation (dismissed):

liquibase-package-manager.yaml
```diff
@@ -0,0 +1,51 @@
...
+      expected-commit: 1cd39ecef00dee2b6cb800acd2d2410f06179231
+
+  - uses: go/build
+    with:
+      packages: ./cmd/lpm/darwin.go
+      output: lpm
+
+  - runs: |
+      mkdir -p "${{targets.contextdir}}/bin"
+      ln -sf /usr/bin/lpm ${{targets.contextdir}}/bin/lpm
```
kranurag7, line 23: non-blocking comment:
I expected this to be part of `${{package.name}}-compat` package but this works inside of main package as well. We carry symlinks inside a compat package looking at existing packages. 💭

liquibase-package-manager.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+  - uses: strip
```
kranurag7, line 25: nit: we can omit this given go/build pipeline already does the stripping. ref: https://github.com/chainguard-dev/melange/blob/a46420071cfd3beb3a72327bd4254535bc8c43b2/pkg/build/pipelines/go/build.yaml#L58-L62

liquibase-package-manager.yaml
```diff
@@ -0,0 +1,51 @@
...
+    contents:
+      packages:
+        - go
+        - curl
+  pipeline:
+    - uses: git-checkout
+      with:
+        repository: https://github.com/liquibase/liquibase-package-manager
+        tag: v${{package.version}}
+        expected-commit: 1cd39ecef00dee2b6cb800acd2d2410f06179231
```
kranurag7, line 44: this looks good but this brings will fail package update on every update. As of now, the bot only updates the git commit inside main pipeline.

we can consider using the following in here after including `git` inside packages above. 
```bash
git clone -b v${{package.version}} https://github.com/liquibase/liquibase-package-manager
```

kranurag7 (commented):
great work!! I've left some non-blocking suggestions inline.

liquibase-package-manager.yaml
```diff
@@ -0,0 +1,50 @@
...
+        # Clone lpm repo for test suite access
+        git clone -b v${{package.version}} https://github.com/liquibase/liquibase-package-manager
+        cd ${{package.name}}
+
+        # Install prereqs
+        go install honnef.co/go/tools/cmd/staticcheck@latest
+
+        # Run test suite
+        export PATH=$PATH:$(go env GOPATH)/bin
+        make test
```
kranurag7, line 50: this is testing things on go side and not the final binary. I think we can do something like following:
```bash
/usr/bin/lpm 2>&1 | grep -i liquibase
```

kranurag7 (commented):
thank you for addressing the reviews, this looks good to me.
I am still curious about the decision around not have have the symlinks inside compat package here.

liquibase-package-manager.yaml
mjramirez-dev: Looks like I was using the wrong test in their makefile.

They have another set of tests that are actually testing the binary here:
- https://github.com/liquibase/liquibase-package-manager/blob/master/Makefile#L54

Working on getting that test suite working in my test pipeline.

mjramirez-dev:
> thank you for addressing the reviews, this looks good to me. I am still curious about the decision around not have have the symlinks inside compat package here.

Looks like I missed that. Went ahead and moved the symlinks inside the compat package.

liquibase-package-manager.yaml
```diff
@@ -0,0 +1,82 @@
...
+    with:
+      packages: ./cmd/lpm/darwin.go
+      output: lpm
+
+subpackages:
+  - name: ${{package.name}}-compat
+    description: "Compatibility package to place binaries in the location expected by lpm"
+    dependencies:
+      runtime:
+        - ${{package.name}}
```
kranurag7, line 32: we can consider omitting this dependencies here.

liquibase-package-manager.yaml
```diff
@@ -0,0 +1,82 @@
...
+      runs: |
+        # Navigate to lpm dir
+        cd ${{package.name}}
+
+        # Install prereqs
+        make test-setup
+
+        # Run test suite
+        export PATH=$PATH:$JAVA_HOME/bin
+        make e2e
```
kranurag7, line 82: I maybe afraid that all this is not touching the binary but I think we're covering that somewhere in the loop as well. 

VEXRUN is pointing to `java -jar $(VEXRUN_FILE)` so maybe it's all doing testing for that one jar file but I maybe wrong here. 

I'm confident on the test here because we're running the upstream e2e test here. 

If we also include `/usr/bin/lpm 2>&1 | grep -i liquibase` in one step then it'll make sure that the binary doesn't crash in between updates.

mjramirez-dev: The tests being run for the 'e2e' tests can be found here:
- https://github.com/liquibase/liquibase-package-manager/tree/master/tests/endtoend

Which look like their invoking the binary. But it also doesn't hurt to add that additional check you mentioned. I'll go ahead and add another step to include that check, so that we stay on the safe side. Appreciate the caution.

kranurag7 (dismissed):
thanks a lot for the work and addressing the reviews.

kranurag7:
We're undergoing usrmerge so the bin compat will be removed in coming days as part of that work. We can consider using the /usr/bin/ binary unless the image really looks for the binary at /bin/lpm path. 💭

mjramirez-dev:
> We're undergoing usrmerge so the bin compat will be removed in coming days as part of that work. We can consider using the /usr/bin/ binary unless the image really looks for the binary at /bin/lpm path. 💭

I'm fine with taking out the bin compat and just using /usr/bin. Less work for the usrmerge maintenance window.

kranurag7:
> I'm fine with taking out the bin compat and just using /usr/bin. Less work for the usrmerge maintenance window.

I think that would be great, I can give a ✅ after your next push. :)

kranurag7 (approved):
thank you!! looks good, verified that project LICENSE file contains both Apache-2.0 and MPL.

PR #43796, opened 2025-02-25
Add the pkgconf test to oath-toolkit

murraybd:
The pkgconf linter told me to add the test and I did and it passed.


oath-toolkit.yaml
```diff
@@ -1,6 +1,7 @@
 package:
   name: oath-toolkit
   version: 2.6.12
+  epoch: 1
```
gaughen, line 4: I did not expect to see you needing to add epoch, I expected it to be a diff. interesting.

gaughen (approved):
LGTM

PR #43810, opened 2025-02-26
fix(datadog-agent): Don't remove object files

EyeCantCU:
Leveraged by the OOM killer and other features


kranurag7 (approved):
thank you!! reading the flags we're removing and description, this looks good.

PR #43879, opened 2025-02-26
kubernetes-csi-driver-nfs package build

developer-guy:
Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/kubernetes-csi-driver-nfs.yaml
+++ b/kubernetes-csi-driver-nfs.yaml
@@ -1,0 +1,58 @@
+package:
+  name: kubernetes-csi-driver-nfs
+  version: 4.10.0
+  epoch: 0
+  description: This driver allows Kubernetes to access NFS server on Linux node
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - coreutils
+      - efs-utils
+      - libcap-dev
+      # https://github.com/kubernetes-csi/csi-driver-nfs/blob/b68ee7181160b12717c80c53667e2d33bf1edc7b/Dockerfile#L21
+      - mount
+      - umount
+      - util-linux
+      - util-linux-misc
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/kubernetes-csi/csi-driver-nfs
+      tag: v${{package.version}}
+      expected-commit: 360cff75fa82720bbcc1eaba1194215d2937e750
+
+  - uses: go/build
+    with:
+      go-package: go
+      packages: ./cmd/nfsplugin
+      ldflags: |
+        -X github.com/kubernetes-csi/csi-driver-nfs/pkg/nfs.driverVersion=${{package.version}}
+        -X github.com/kubernetes-csi/csi-driver-nfs/pkg/nfs.gitCommit=$(git rev-parse --short=8 HEAD)
+        -X github.com/kubernetes-csi/csi-driver-nfs/pkg/nfs.buildDate=$(date -u -d "@${SOURCE_DATE_EPOCH:-$(date +%s)}" "+%Y-%m-%dT%H:%M:%SZ")
+      output: nfsplugin
+
+subpackages:
+  - name: ${{package.name}}-compat
+    description: "Compatibility package for matching with the upstream Dockerfile"
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.contextdir}}"/
+          ln -sf /usr/bin/nfsplugin "${{targets.contextdir}}"/nfsplugin
+
+update:
+  enabled: true
+  github:
+    identifier: kubernetes-csi/csi-driver-nfs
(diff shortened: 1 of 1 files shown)


kubernetes-csi-driver-nfs.yaml
```diff
@@ -0,0 +1,39 @@
...
+      packages: ./cmd/nfsplugin
+      output: nfsplugin
+
+subpackages:
+  - name: ${{package.name}}-compat
+    description: "Compatibility package for matching with the upstream Dockerfile"
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.contextdir}}"/
+          ln -sf /usr/bin/nfsplugin "${{targets.contextdir}}"/nfsplugin
```
kranurag7, line 28: we can consider adding tests here. (stat or /nfsplugin --help)

kubernetes-csi-driver-nfs.yaml
```diff
@@ -0,0 +1,39 @@
...
+update:
+  enabled: true
+  github:
+    identifier: kubernetes-csi/csi-driver-nfs
+    strip-prefix: v
+
+test:
+  pipeline:
+    - runs: |
+        nfsplugin --help
```
kranurag7, line 39: non-blocking: if we can extend something more here that would be great. I think it will require a nfs up so not sure if we can do it here or not. maybe start nfs-server using test/daemon-check-output and do some tests around it. 💭

kubernetes-csi-driver-nfs.yaml
```diff
@@ -0,0 +1,39 @@
+package:
+  name: kubernetes-csi-driver-nfs
+  version: 4.10.0
+  epoch: 0
+  description: This driver allows Kubernetes to access NFS server on Linux node
+  copyright:
+    - license: Apache-2.0
```
kranurag7, line 7: I think we'll require some runtime deps here (reading: https://github.com/kubernetes-csi/csi-driver-nfs/blob/b68ee7181160b12717c80c53667e2d33bf1edc7b/Dockerfile#L21 and checking `crane export gcr.io/k8s-staging-sig-storage/nfsplugin:canary nfsplugin.tar` and `tar -tf nfsplugin.tar`) 💭

kranurag7 (commented):
great work!! I've left some suggestions inline.

kubernetes-csi-driver-nfs.yaml
developer-guy: That'd be great if we could do that but this is CSI Plugin and we need a real Kubernetes cluster to be able to test its functionality because we have to create a PersistentVolume like [this](https://github.com/kubernetes-csi/csi-driver-nfs/blob/b68ee7181160b12717c80c53667e2d33bf1edc7b/deploy/example/nfs-provisioner/nginx-pod.yaml#L17-L24) and then we should mount that PV to the Pod to see if we can write files to that directory and check NFS server that the file is there too.

developer-guy: Then I saw this: https://github.com/kubernetes-csi/csi-driver-nfs/blob/b68ee71/test/sanity/run-test.sh then tried to setup an NFS server by following this guide: https://wiki.alpinelinux.org/wiki/Setting_up_an_NFS_server
but failed due to following error:

```bash
Exporting File System...
exporting *:/data
exportfs: /data does not support NFS export
Export validation failed, exiting...
```

so I added a simple test with grep.

wojciechka (approved):

kranurag7 (approved):

kubernetes-csi-driver-nfs.yaml
```diff
@@ -0,0 +1,58 @@
...
+  dependencies:
+    runtime:
+      - coreutils
+      - efs-utils
+      - libcap-dev
```
jamonation, line 12: Should this be plain the libcap package here? Why does this package need development headers at runtime if it is a compiled go binary?

kranurag7: good catch Jamon, I missed this. 

We only need libcap here (not the headers) 
```bash
# dpkg -L libcap2
/.
/lib
/lib/x86_64-linux-gnu
/lib/x86_64-linux-gnu/libcap.so.2.66
/lib/x86_64-linux-gnu/libpsx.so.2.66
/usr
/usr/share
/usr/share/doc
/usr/share/doc/libcap2
/usr/share/doc/libcap2/changelog.Debian.gz
/usr/share/doc/libcap2/changelog.gz
/usr/share/doc/libcap2/copyright
/lib/x86_64-linux-gnu/libcap.so.2
/lib/x86_64-linux-gnu/libpsx.so.2
```
opened a follow-up here https://github.com/wolfi-dev/os/pull/43994

developer-guy: We should fix this @kranurag7, could you please send a PR for fixing it? 🙏

PR #43884, opened 2025-02-26
ingress-nginx-controller-1.12: Drop ld.so.conf modification

sergiodj:
We already cover /usr/local/lib in our /etc/ld.so.conf, so there's no need to replicate it there.


dannf (approved):
thanks!

PR #43893, opened 2025-02-26
Fix: usrmerge /bin

zmarano:


cloudnative-pg.yaml
```diff
@@ -1,10 +1,14 @@
 package:
   name: cloudnative-pg
   version: "1.25.1"
-  epoch: 0
+  epoch: 30
```
xnox, line 4: Note to self, figure out why this package needs changing.

openbao.yaml
```diff
@@ -46,11 +50,9 @@ subpackages:
     description: "Compatibility package to allow replacing vault image"
     pipeline:
       - runs: |
-          mkdir -p "${{targets.subpkgdir}}/bin"
-          ln -sf /usr/bin/bao "${{targets.subpkgdir}}/bin/bao"
-          ln -sf /bin/bao "${{targets.subpkgdir}}/bin/vault"
-
           mkdir -p "${{targets.subpkgdir}}/usr/bin"
+          ln -sf /usr/bin/bao "${{targets.subpkgdir}}/usr/bin/vault"
```
xnox, line 54: Relative would be better. But also inconsequential.

shadow.yaml
```diff
@@ -2,12 +2,13 @@
 package:
   name: shadow
   version: "4.17.3"
-  epoch: 20
+  epoch: 30
```
xnox, line 5: Note to self, check why.

zmarano: Whoops- forgot to include the actual changes! Fixed.

openbao.yaml
zmarano: Agreed- fixed.

shadow.yaml
```diff
@@ -111,11 +113,11 @@ subpackages:
     pipeline:
       - runs: |
           cd ${{targets.destdir}}
+          mkdir -p ${{targets.subpkgdir}}/usr/bin
           for bin in ${{vars.login-utils}}; do
             for dir in bin usr/bin usr/sbin; do
```
xnox, line 118: this seems to be doing merged-usrsbin too. but ok.

xnox (approved):

PR #43924, opened 2025-02-26
Add the pkgconf test to libxdamage

murraybd:
The runtime deps were need to resolve some errors.


gaughen (approved):
approving.  Per Brian the runtime dependencies are being added because the testing is revealing a failure in our package build and adding the dependency fixes it.

PR #44011, opened 2025-02-27
cassandra: use vars and add jemalloc

Dentrax:
Add `vars:` and `jemalloc` runtime dep.


jamonation (approved):
built this into an image and all tests pass - thanks for adding jemalloc!

PR #44061, opened 2025-02-27
remove yq from kubectl-bitnami-compat

rakshitgondwal:
Removes `yq` from the `kubectl-bitnami-compat` sub package.
`yq` is being removed from the package level and being added to the image level.


utieyin (approved):
LGTM

mamccorm (approved):

PR #42723, opened 2025-02-14
Fix/usrmerge sbin

smoser:
These are change to packages that ship files in /sbin, moving them to /usr/bin/

PR #44083, opened 2025-02-27
fix: busybox - usrmerge /sbin

zmarano:


smoser:
pulled this into https://github.com/wolfi-dev/os/pull/42723

PR #44088, opened 2025-02-27
fix: busybox - usrmerge /bin

zmarano:


zmarano:
Pulled into https://github.com/wolfi-dev/os/pull/43893

PR #44107, opened 2025-02-28
feat(chromium): Add support for arm64

EyeCantCU:
These changes implement support for arm64 in our builds of Chromium. Here are the changes at a high level:
* Build Chromium with our toolchain. Use our Clang, our Rust, our Java, etc to produce builds of Chromium independent of Chromium's toolchain. Chromium's toolchain only supports x64 systems. Ours supports x64 and arm64
* Use Gentoo's Chromium tarballs. Google's depot tools do not work on arm64. Dependencies must be synced before we start building Chromium, and Gentoo actively maintains tarballs containing Chromium's source and dependencies after they've been synced with gclient. In the future, we should be able to perform syncs using qemu

### DIFF
--- a/chromium.yaml
+++ b/chromium.yaml
@@ -103,9 +103,7 @@ environment:
       - libxtst
       - libxtst-dev
       - linux-headers
-      - mesa
       - mesa-dev
-      - mesa-gbm
       - minizip
       - nghttp2-dev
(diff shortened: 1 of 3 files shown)


jamonation (approved):
🤞

PR #44112, opened 2025-02-28
ksm: use go/build, add bitnami-compat and expand tests

kranurag7:
ksm: use go/build, add bitnami-compat and expand tests

### DIFF
--- a/kube-state-metrics.yaml
+++ b/kube-state-metrics.yaml
@@ -41,9 +41,6 @@ subpackages:
         with:
           image: ${{package.name}}
           version-path: 2/debian-12
-      - runs: |
-          chmod g+rwX ${{targets.contextdir}}/opt/bitnami
-          find / -perm /6000 -type f -exec chmod a-s {} \; || true
 
 update:
   enabled: true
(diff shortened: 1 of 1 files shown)


kube-state-metrics.yaml
```diff
@@ -30,12 +22,28 @@ pipeline:
...
+  - name: ${{package.name}}-bitnami-compat
+    description: "compat package for ${{package.name}} image"
+    pipeline:
+      - uses: bitnami/compat
+        with:
+          image: ${{package.name}}
+          version-path: 2/debian-12
+      - runs: |
+          chmod g+rwX ${{targets.contextdir}}/opt/bitnami
+          find / -perm /6000 -type f -exec chmod a-s {} \; || true
```
wojciechka, line 46: Wouldn't this also try to modify the entire build environment? Should find be done inside `${{targets.contextdir}}/opt/bitnami` instead?

kranurag7: copy-paste from other configs, I think we don't need it. let me remove it or use `find ${{targets.contextdir}}/ -perm /6000 -type f -exec chmod a-s {} \; || true`

utieyin (dismissed):
LGTM

wojciechka (approved):

PR #44118, opened 2025-02-28
daemon-check-output: expand error cases

Dentrax:
Expands test cases of `daemon-check-output` pipeline.

### DIFF
--- a/pipelines/test/daemon-check-output.yaml
+++ b/pipelines/test/daemon-check-output.yaml
@@ -38,10 +38,15 @@ inputs:
       FATAL
       Traceback.*most.recent.call
       Exception in thread
-      java.lang.NullPointerException
-      java.lang.RuntimeException
+      java.lang.*Exception
       Gem::MissingSpecError
       command not found
+      connection refused
+      unable to load
+      does not exist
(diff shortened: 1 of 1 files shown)


smoser (commented):
We want this list to never false-positive on failure (to only fail when something is an actual failure).  NullPointerException or RuntimeException seem pretty safe from that.  But any exception in output seems like it might just be noise, especially in debug output.

this one seems the most likely to cause false positive, on output like:

Each of 'connection refused', 'does not exist' and the wide 'java.lang.*Exception' look like they could false positive on debug output.  The problem with casting a wide net here is that if something is hit by a false positive, the only option is to then declare the list in that package's melange yaml, effectively forking it.

I'm not 100% sure that i'm right here and I don't know that there is an obvious "right" answer.  thoughts?

Second comment, I do hope that we're going to re-write daemon-check-output in go sometime in [tw](https://github.com/chainguard-dev/tw). interested?

Dentrax:
100% agree on your concerns @smoser, maybe we can introduce a new field like `extra_error_strings`? That way we wouldn't override the default values. And also, I like the idea of rewriting this in Go in `tw`.

PR #44119, opened 2025-02-28
redis-operator-0.18.1 : new package

rahulduvedi:
Related:

#### For new package PRs only


redis-operator.yaml
```diff
@@ -0,0 +1,45 @@
+package:
+  name: redis-operator
+  version: "0.18.1"
+  epoch: 1
```
kranurag7, line 4:
```suggestion
  epoch: 0
```
we start with epoch o and bump it on every rebuild.

redis-operator.yaml
```diff
@@ -0,0 +1,45 @@
...
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - ca-certificates-bundle
+      - wolfi-base
+      - go
```
kranurag7, line 15: we can consider omitting this because go/build pipeline already pulls in all the necessary dependencies. ref: https://github.com/chainguard-dev/melange/blob/3337f81591f7bcd9aa5d9d09d032f2a97e86d67d/pkg/build/pipelines/go/build.yaml#L3-L7

redis-operator.yaml
```diff
@@ -0,0 +1,45 @@
...
+    with:
+      repository: "https://github.com/OT-CONTAINER-KIT/redis-operator.git"
+      tag: "v${{package.version}}"
+      expected-commit: "c48c9c5b0d09fb3e3438e269c7b5457585d062c2"
+
+  - uses: go/build
+    with:
+      packages: "."
+      output: "manager"
+      ldflags: "-s -w"
```
kranurag7, line 28: we can consider omitting this because we don't want to strip for better sbom discoverability and govulncheck also relies on this.

redis-operator.yaml
```diff
@@ -0,0 +1,45 @@
...
+      prefix: "usr"
+      install-dir: "bin"
+      tidy: true
```
kranurag7, line 31: `tidy: true` might not be needed and other two are default so we can consider omitting this. 💭

redis-operator.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+  - uses: strip
```
kranurag7, line 33: go/build pipeline already does the stripping so we can consider omitting this. ref: https://github.com/chainguard-dev/melange/blob/3337f81591f7bcd9aa5d9d09d032f2a97e86d67d/pkg/build/pipelines/go/build.yaml#L58-L62

redis-operator.yaml
```diff
@@ -0,0 +1,45 @@
...
+  enabled: true
+  github:
+    identifier: "OT-CONTAINER-KIT/redis-operator"
+    strip-prefix: "v"
+
+test:
+  pipeline:
+    - name: "Verify redis-operator binary"
+      runs: |
+        /usr/bin/manager -help
```
kranurag7, line 45: can we extend the tests a little bit more if possible?

kranurag7 (commented):
great work!! I've put some comments inline. Please feel free to reach out if you've any questions w.r.t. any of this. We can sync about it as well.

redis-operator.yaml
rahulduvedi: @kranurag7 can you please help me with these for this time?
I worked on your other comments, This is the one which I'm quite not able to figure out.

kranurag7: we can run the manager binary in a k8s environment to then grep some logs when the controller starts. This is what happens when you deploy the image in k8s cluster. You've the binary running in a k8s environment so it has access to apiserver to interact with. To set the kubernetes environment we can use `test/kwok/cluster` pipeline and to grep some logs we use `test/daemon-check-output` pipeline which runs the binary with a timeout and look for logs. 

<details><summary> tests </summary>
<p>

```yaml
test:
  environment:
    environment:
      KUBERNETES_SERVICE_HOST: "127.0.0.1"
      KUBERNETES_SERVICE_PORT: 32764
  pipeline:
    - name: "Verify redis-operator binary"
      runs: |
        /usr/bin/manager -help
    - uses: test/kwok/cluster
      with:
        timeout: 30
        start: /usr/bin/manager
        expected_output: |
          Registering webhook
          Kind=Redis
          Kind=RedisCluster
          Kind=RedisReplication
          starting manager
          Starting metrics server
          Starting webhook server
          Starting Controller
          Starting EventSource
```

</p>
</details>

redis-operator.yaml
rahulduvedi: Thanks for the suggestion @kranurag7,  I have resolved this.

redis-operator.yaml
rahulduvedi: Thanks for suggesting, This was helpful.

redis-operator.yaml
rahulduvedi: Thanks for suggesting and providing reference, I have removed this

redis-operator.yaml
rahulduvedi: Thanks for suggesting and providing reference, I have removed this

kranurag7: we can also omit the following given this is default. 
```yaml
      prefix: "usr"
      install-dir: "bin"
```

redis-operator.yaml
kranurag7: We can consider omitting `ldflags: "-s -w"` because we don't want to strip everything from the go binary.

kranurag7 (commented):
we'll also require a compat package for this given upstream has the binary at `/manager` 

<p>

```yaml
subpackages:
    pipeline:
      - runs: |
          mkdir -p ${{targets.contextdir}}
          ln -sf /usr/bin/manager ${{targets.contextdir}}/manager
    description: Compat package for ${{package.name}}
    test:
      pipeline:
        - runs: |
            stat /manager
```

</p>
</details>

redis-operator.yaml
rahulduvedi: Removed.

redis-operator.yaml
rahulduvedi: Removed

redis-operator.yaml
```diff
@@ -0,0 +1,53 @@
...
+  version: "0.19.1"
+  epoch: 0
+  description: "A Kubernetes operator to manage Redis clusters"
+  copyright:
+    - license: Apache-2.0
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: "https://github.com/OT-CONTAINER-KIT/redis-operator.git"
```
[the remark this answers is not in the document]
kranurag7: yes Mark, I think this is correct reading the issue and given this is the most popular operator (from previous work exp) 

I don't have a hard opinion on naming for now, if we want to be specific from now then maybe we can call it `ot-redis-operator` but it's good as well and we can be specific in the next ones that come up. 💭 (both works imho)

kranurag7 (approved):
thank you the changes looks good to me.

redis-operator.yaml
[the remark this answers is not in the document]
rahulduvedi: Hello @mamccorm  @kranurag7 can we please resolve this comment so we can go ahead and close this PR, if any 
other changes are not required.

mamccorm (approved):

PR #44133, opened 2025-02-28
feat(mysql-9.1): add bitnami-compat package

wojciechka:


kranurag7 (dismissed):

mysql-9.1.yaml
```diff
@@ -132,6 +132,59 @@ subpackages:
...
+
+          mkdir -p "${{targets.contextdir}}/usr/share/${{package.name}}"
+          ln -s /usr/share/${{package.name}} "${{targets.contextdir}}/usr/share/${{package.name}}/${{package.name}}"
+
+          mkdir -p "${{targets.contextdir}}/opt/bitnami/mysql/conf/bitnami"
+          install -m644 bitnami-my.cnf "${{targets.contextdir}}/opt/bitnami/mysql/conf/my.cnf"
+          cp -pfR "${{targets.contextdir}}/opt/bitnami/mysql/conf" "${{targets.contextdir}}/opt/bitnami/mysql/conf.default"
+
+          mkdir -p "${{targets.contextdir}}/opt/bitnami/mysql/logs"
+          ln -sf /proc/1/fd/1 "${{targets.contextdir}}/opt/bitnami/mysql/logs/mysqld.log"
```
kranurag7, line 171: nit: for better readability.
```suggestion
          ln -sf /dev/stdout "${{targets.contextdir}}/opt/bitnami/mysql/logs/mysqld.log"
```

mysql-9.1.yaml
```diff
@@ -132,6 +132,59 @@ subpackages:
...
+    test:
+      environment:
+        contents:
+          packages:
+            - ${{package.name}}
+      pipeline:
+        - runs: |
+            test -f /opt/bitnami/mysql/bin/mysql
```
kranurag7, line 179: I think we can consider passing in `BITNAMI_APP_NAME=mysql` as env and then running `/opt/bitnami/scripts/mysql/entrypoint.sh /opt/bitnami/scripts/mysql/run.sh` to see what comes up and then grepping some logs. this will also validate that all symlinks are functional and we're not missing something here.

kranurag7 (commented):
great work!! so many symlinks :melting: I'm leaving some suggestions inline.

mysql-9.1.yaml
wojciechka: I added the exact same symlink that upstream is doing. I'd rather keep this as is unless you're strong about this change?

mysql-9.1.yaml
wojciechka: I was thinking of this as next step. I want to get proper testing at image level to make sure the compat package is complete before adding more package tests.

mysql-9.1.yaml
```diff
@@ -132,6 +132,93 @@ subpackages:
...
 
+  - name: ${{package.name}}-bitnami-compat
+    dependencies:
+      provides:
+        - mysql-bitnami-compat=${{package.full-version}}
+      runtime:
+        - bash
+        - busybox
+        - coreutils
+        - ini-file
```
kranurag7, line 143: I think somewhere down the road we'll move `ini-file` (it is using go toolchain as of now) to image level, not a blocker for now but we can start it today. :)

kranurag7 (approved):
thank you, looks good!!

mysql-9.1.yaml
wojciechka: Let's keep it as is for now. We can refactor it later, it's a minor change if we need it for any reason.

PR #44134, opened 2025-02-28
Correct symlink path in subpackage, and add additional link

mamccorm:
Earlier commit: https://github.com/wolfi-dev/os/commit/962bc3d52d764bb882e9e27c6ae34f8b27e9bf0e, didn't set the correct path for the metrics-server symlink in the -compat package. This is now corrected.

Also adding additional directory expected for compatibility with an upstream helm chart.

### DIFF
--- a/metrics-server.yaml
+++ b/metrics-server.yaml
@@ -1,7 +1,7 @@
 package:
   name: metrics-server
   version: 0.7.2
-  epoch: 6
+  epoch: 7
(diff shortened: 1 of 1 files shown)


kranurag7 (approved):
thank you for the follow-up, looks good. Excited for what's coming up after merging this. 

I thought of symlinking `/opt/bitnami/metrics-server/certificates/` to `/tmp` similar to match the regular image but I changed my mind. An empty directory looks good and easier.

PR #44156, opened 2025-02-28
license linting: fix licenses for some wolfi packages.

sil2100:
Experimenting with license linting tools. Got an extract of all licenses for wolfi, wrote a quick program that runs though those and checks which licenses are not matching. Hacked it up to fix the 'obvious' cases - aka. one license file, over 95% match confidence, and ran automation to fix those up.

There rest are a bit more tricky, like multi-licensing or some fuzzy matches. We'll get to those.


sil2100:
I'd recommend not merging this just yet as automation as I'd still like to check those results before continuing.

dannf (commented):
fwiw, I think it is worth bumping the epochs here.

sil2100:
Yeah, was thinking about that. Certainly the epoch bumps are needed to get the right info into the SBOMs. I guess I'll bump them right before this is ready for review, as epoch bumps can start conflicting very easily

dannf (dismissed):

javacruft (dismissed):

javacruft (approved):

PR #44159, opened 2025-02-28
Build prism with ttypescript again

paulgibert:
Prism is built with [ttypescript](https://github.com/cevek/ttypescript), a depreciated project. [ts-patch](https://github.com/nonara/ts-patch) is the recommended drop-in replacement but does not support the version of typescript used by prism. Originally we wrote some patches that allowed us to build prism with ts-patch, but recent builds are breaking. This is an example of what it would look like if we were to go back to building with ttypescript.

Note: patches are removed. We call `yarn && yarn build` invoking `ttsc` as defined in the top-level `package.json`.


paulgibert:
@philroche @kbsteere for your consideration

dnegreira (approved):

philroche (approved):

PR #44175, opened 2025-02-28
py3-pydantic: pin pydantic-core at version expected by upstream

OddBloke:
Using the most recent pydantic-core results in exceptions in downstream packages like `py3-fromager` as the API expected by pydantic has changed since.

This also adds execution of `tests/test_version.py`: upstream have introduced a test that the version of `pydantic-core` installed is the one expected by `pydantic` itself: this will force us to update the version pin when upstream changes require it.

### DIFF
--- a/py3-pydantic.yaml
+++ b/py3-pydantic.yaml
@@ -2,7 +2,7 @@
 package:
   name: py3-pydantic
   version: "2.10.6"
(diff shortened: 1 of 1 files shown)


powersj (approved):
Awesome test for catching when to update

PR #44186, opened 2025-02-28
enable elf notes by default

xnox:
- **Add gcc-14 spec files that opt-out of melange spec file include**

- **openssf-compiler-options: gcc-14 enable melange spec file include by default**
  Revert this commit to back out the defaults change.
 
- **epoch bump**

After three rounds of canaries, enable packaging ELF notes by default, this will affect most C, C++ and rust binaries.

Go binaries will need a separate changes in melange pipelines to activate elf notes, due to need to specify external linking mode by default.

If needed, provide the opt-out file, to be able to build with hardening, but without the elf note. If all goes well, it will be dropped in 4 weeks time.

### DIFF
--- a/openssf-compiler-options.yaml
+++ b/openssf-compiler-options.yaml
@@ -1,7 +1,7 @@
 package:
   name: openssf-compiler-options
   version: 20240627
-  epoch: 12
+  epoch: 13
   description: "Compiler Options Hardening Guide for C and C++"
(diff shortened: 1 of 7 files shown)


xnox:
something is odd with omit pointers.

sergiodj (approved):
AFAIU, LGTM.

PR #44214, opened 2025-03-01
add bitnami-compat package to alertmanager

kranurag7:
adds compat package to alertmanager

### DIFF
--- a/prometheus-alertmanager.yaml
+++ b/prometheus-alertmanager.yaml
@@ -55,7 +55,7 @@ subpackages:
           mkdir -p ${{targets.contextdir}}/opt/bitnami/alertmanager/data/
           mkdir -p ${{targets.contextdir}}/etc/alertmanager
           mkdir -p ${{targets.contextdir}}/alertmanager
-          chmod g+rwX ${{targets.contextdir}}/opt/bitnami
+          chmod g+rwX ${{targets.contextdir}}/opt/bitnami/alertmanager/data/
           ln -sf /usr/bin/alertmanager ${{targets.contextdir}}/opt/bitnami/alertmanager/bin/alertmanager
           ln -sf /usr/bin/amtool ${{targets.contextdir}}/opt/bitnami/alertmanager/bin/amtool
           ln -sf /etc/alertmanager/alertmanager.yml ${{targets.contextdir}}/opt/bitnami/alertmanager/conf/config.yml
--- a/prometheus-alertmanager.yaml
+++ b/prometheus-alertmanager.yaml
@@ -86,6 +86,9 @@ subpackages:
               # Test health endpoints
               curl -s http://127.0.0.1:9093/-/ready | grep -i OK
               curl -s http://127.0.0.1:9093/-/healthy | grep -i OK
+              # Create a test alert
+              curl -XPOST -H "Content-Type: application/json" -d '[{"labels":{"alertname":"TestAlertFromMelangeTest","severity":"critical"}}]' http://127.0.0.1:9093/api/v2/alerts
+
 
 update:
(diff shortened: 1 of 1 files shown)


jamonation (dismissed):

prometheus-alertmanager.yaml
```diff
@@ -45,6 +44,35 @@ pipeline:
...
+        contents:
+          packages:
+            - ${{package.name}}
+      pipeline:
+        - runs: |
+            /opt/bitnami/alertmanager/bin/alertmanager --version | grep ${{package.version}}
+            /opt/bitnami/alertmanager/bin/alertmanager --help
+            /opt/bitnami/alertmanager/bin/amtool --version | grep ${{package.version}}
+            /opt/bitnami/alertmanager/bin/amtool --help
+
```
Dentrax, line 75: Can we have a `test/daemon-check-output` pipeline here to spin up with defined env vars and its entrypoint/cmd?

prometheus-alertmanager.yaml
```diff
@@ -45,6 +44,35 @@ pipeline:
 
   - uses: strip
 
+subpackages:
+  - name: ${{package.name}}-bitnami-compat
+    dependencies:
```
Dentrax, line 49: Should we add runtime deps? https://eng.inky.wtf/docs/container-images/playbooks/bitnami/#runtime-dependencies

prometheus-alertmanager.yaml
Dentrax: Then simple `curl` test for `/api/v2/alerts` endpoint? Or status / health endpoints.

prometheus-alertmanager.yaml
```diff
@@ -45,6 +44,35 @@ pipeline:
...
+          ln -sf /usr/bin/alertmanager ${{targets.contextdir}}/opt/bitnami/alertmanager/bin/alertmanager
+          ln -sf /usr/bin/amtool ${{targets.contextdir}}/opt/bitnami/alertmanager/bin/amtool
+          ln -sf /etc/alertmanager/alertmanager.yml ${{targets.contextdir}}/opt/bitnami/alertmanager/conf/config.yml
+          ln -sf /opt/bitnami/alertmanager/conf ${{targets.contextdir}}/etc/alertmanager
+          ln -sf /opt/bitnami/alertmanager/data ${{targets.contextdir}}/alertmanager
+    test:
+      environment:
+        contents:
+          packages:
+            - ${{package.name}}
```
Dentrax, line 68: We shouldn't add this as melange populates it automatically.
```suggestion
```

Dentrax (changes requested):

prometheus-alertmanager.yaml
kranurag7: thank you for looking into it, given we're invoking the binary, including the main package in here is required. When we do sub-package test, it doesn't pull in the main package so we'll have to be explicit here.

prometheus-alertmanager.yaml
kranurag7: given the entrypoint is `/opt/bitnami/alertmanager/bin/alertmanager` we don't need any runtime dependencies. If it was scripts, then we needed some but in this case we can omit this.

prometheus-alertmanager.yaml
Dentrax: Oh, right :D

prometheus-alertmanager.yaml
Dentrax: I didn't notice that, thanks!

Dentrax (approved):
Thanks!

PR #44216, opened 2025-03-01
add bitnami compats to cert-manager

kranurag7:
adds compat and a simple stat test to verify the symlink, testing is minimal in here given there's no scripts and entrypoint is the binary itself which we're already testing.

### DIFF
--- a/cert-manager-1.17.yaml
+++ b/cert-manager-1.17.yaml
@@ -2,7 +2,7 @@ package:
   name: cert-manager-1.17
   # See https://cert-manager.io/docs/installation/supported-releases/ for upstream-supported versions
   version: "1.17.1"
-  epoch: 1
+  epoch: 2
   description: Automatically provision and manage TLS certificates in Kubernetes
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


cert-manager-1.17.yaml
```diff
@@ -69,6 +69,22 @@ subpackages:
       provides:
         - cert-manager-controller=${{package.full-version}}
 
+  - name: cert-manager-controller-${{vars.major-minor-version}}-bitnami-compat
+    dependencies:
+      provides:
+        - cert-manager-controller-bitnami-compat=${{package.full-version}}
+    pipeline:
+      - uses: bitnami/compat
```
mamccorm, line 77: We should consider dropping the bitnami/compat part as it doesn't look to need any of the files being pulled in

wojciechka (dismissed):

cert-manager-1.17.yaml
kranurag7: I'm keeping it here to keep the licenses file intact that's coming along but no harm in removing as well.

mamccorm: If it's just a symlink we're creating, i'd drop this, as we're copying in stuff we don't need / aren't used, but not blocking

mamccorm (dismissed):

mamccorm (approved):

PR #44219, opened 2025-03-01
add sealed-secrets package

kranurag7:
adds sealed-secrets and sealed-secrets-kubeseal with provides:kubeseal via this patch.


sealed-secrets.yaml
```diff
@@ -0,0 +1,111 @@
...
+    description: "compat package with bitnami/sealed-secrets-controller image"
+    pipeline:
+      - uses: bitnami/compat
+        with:
+          image: sealed-secrets-controller
+          version-path: 0/debian-12
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/opt/bitnami/sealed-secrets-controller/bin/
+          # this is conflicting with ca-certificates-bundle
+          rm -f ${{targets.contextdir}}/etc/ssl/certs/ca-certificates.crt
```
developer-guy, line 67: First time, I'm seeing this do we know that the files the one we add by ca-certificates-bundle package and ca-certificates.crt in this case? I just wanted to make sure that if the file ca-certificates.crt is something required for Bitnami version of sealed-secrets-controller

kranurag7: thank you for looking into this, this is vendored by upstream here https://github.com/bitnami/containers/blob/main/bitnami/sealed-secrets-kubeseal/0/debian-12/rootfs/etc/ssl/certs/ca-certificates.crt 

our package provides the following:
```bash
$ apk info -L ca-certificates-bundle
ca-certificates-bundle-20241121-r2 contains:
etc/pki/tls/certs/ca-bundle.crt
etc/ssl/cert.pem
etc/ssl/certs/ca-bundle.crt
etc/ssl/certs/ca-certificates.crt
var/lib/db/sbom/ca-certificates-bundle-20241121-r2.spdx.json
```
removed the upstream once here so that we can use our package.

Dentrax: +1, can we diff to compare with ours whether this is bitnami-specific thing? https://github.com/bitnami/containers/blob/main/bitnami/sealed-secrets-controller/0/debian-12/rootfs/etc/ssl/certs/ca-certificates.crt

Dentrax: Then we might need to run `update-ca-certificates`

sealed-secrets.yaml
```diff
@@ -0,0 +1,118 @@
...
+        - runs: |
+            kubeseal --version | grep ${{package.version}}
+
+  - name: ${{package.name}}-kubeseal-bitnami-compat
+    description: "compat package with bitnami/sealed-secrets-kubeseal image"
+    pipeline:
+      - uses: bitnami/compat
+        with:
+          image: sealed-secrets-kubeseal
+          version-path: 0/debian-12
```
EyeCantCU, line 51: Let's use variables above for the path. `major-version` and `debian-version`

sealed-secrets.yaml
```diff
@@ -0,0 +1,118 @@
...
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/opt/bitnami/sealed-secrets-kubeseal/bin/
+          # this is conflicting with ca-certificates-bundle
+          rm -f ${{targets.contextdir}}/etc/ssl/certs/ca-certificates.crt
+          ln -sf /usr/bin/kubeseal ${{targets.contextdir}}/opt/bitnami/sealed-secrets-kubeseal/bin/kubeseal
```
EyeCantCU, line 56: nit: A soft link is fine (omit `-f`)

sealed-secrets.yaml
```diff
@@ -0,0 +1,118 @@
...
+          ln -sf /usr/bin/kubeseal ${{targets.contextdir}}/kubeseal
+    test:
+      pipeline:
+        - runs: |
+            stat /opt/bitnami/sealed-secrets-kubeseal/bin/kubeseal
+            stat /kubeseal
```
EyeCantCU, line 62: Ideally, we'd install both binaries and run the same tests ran below to validate the links work as intended. We know `stat` will always pass

sealed-secrets.yaml
```diff
@@ -0,0 +1,118 @@
...
+    description: "compat package with bitnami/sealed-secrets-controller image"
+    pipeline:
+      - uses: bitnami/compat
+        with:
+          image: sealed-secrets-controller
+          version-path: 0/debian-12
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/opt/bitnami/sealed-secrets-controller/bin/
+          # this is conflicting with ca-certificates-bundle
+          mv ${{targets.contextdir}}/etc/ssl/certs/ca-certificates.crt ${{targets.contextdir}}/etc/ssl/certs/bitnami-ca-certificates.crt
```
EyeCantCU, line 74: We should just remove it like we've done above

sealed-secrets.yaml
```diff
@@ -0,0 +1,118 @@
...
+    description: "compat package with bitnami/sealed-secrets-kubeseal image"
+    pipeline:
+      - uses: bitnami/compat
+        with:
+          image: sealed-secrets-kubeseal
+          version-path: 0/debian-12
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/opt/bitnami/sealed-secrets-kubeseal/bin/
+          # this is conflicting with ca-certificates-bundle
+          rm -f ${{targets.contextdir}}/etc/ssl/certs/ca-certificates.crt
```
EyeCantCU, line 55: I wouldn't force commands we run as a part of the pipeline. If it fails, we'd want to know

EyeCantCU (dismissed):

sealed-secrets.yaml
wojciechka: We are not doing this in other places. Should we start doing this everywhere now?

It may be a good idea, if so, why not add it everywhere and require for new packages?

wojciechka (dismissed):

wojciechka (approved):

sealed-secrets.yaml
wojciechka: Agreed on another approach for now

PR #44236, opened 2025-03-02
add missing /etc/ssh/ssh_known_hosts file

developer-guy:
There was an issue like the upstream image includes ssh_known_hosts where ours does not so we should add that.


kranurag7 (approved):
upstream contains an empty file and looks like it's required.

jamonation (approved):

PR #44295, opened 2025-03-03
Nats-top package addition

AnmolVirdi:
**Nats-top package** `Current version: v.0.6.3`
- top-like tool for monitoring NATS servers.

**Type**: Go binary
**REF**: https://github.com/nats-io/nats-top

**Tests:**
- nats-top, nats server and 2 client mode
- delimiter, refresh delays, subscriptions, sort restuls
- TLS Mode


nats-top.yaml
```diff
@@ -0,0 +1,91 @@
...
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/nats-io/nats-top
+      tag: v${{package.version}}
+      expected-commit: 812d35cc991232d7606816a15b2ea0b12fb129e7
+
+  - uses: go/build
+    with:
+      packages: ./nats-top.go
```
mamccorm, line 18: We should drop this and have it default to building go.mod, which imports nats-top.go:
 - https://github.com/nats-io/nats-top/blob/main/go.mod

Otherwise we may lose important dependencies which are included in go.mod

AnmolVirdi: Noted, I've updated default to building go.mod

AnmolVirdi:
- Improved quality of tests.
- Added more validations.
- Replaced `openssl` with `mkcert` for certs creation for test in TLS mode. But I'm still using openssl in the test environment to verify these certs.

nats-top.yaml
```diff
@@ -0,0 +1,91 @@
...
+        - util-linux-misc
+        - mkcert
+        - openssl
+  pipeline:
+    - name: "Version test"
+      runs: |
+        nats-top --version 2>&1 | grep ${{package.version}}
+    - name: Test nats-top with minimal nats server and 2 clients
+      runs: |
+        nats-server -p 4222 -m 8222 & sleep 2 &
```
maxgio92, line 51: Can we move it into a dedicated step, named something like "Run nats-server"?

nats-top.yaml
```diff
@@ -0,0 +1,91 @@
...
+        nats sub client1 > /dev/null & sleep 2 & nats sub client2 > /dev/null & sleep 2
```
maxgio92, line 52: Can we move these subscriber runs into a separate step?

nats-top.yaml
```diff
@@ -0,0 +1,91 @@
...
+        nats-top -s localhost -b -u -o nats-output-${{package.version}}.txt
+        cat nats-output-${{package.version}}.txt
```
maxgio92, line 54: This will never fail. What's the expected output from nats-top?

I would check that the output contains the exact client names and that connections polled is 2 in the output.

Something like:
```yml
    - name: Test correct clients are in nats-top output
      runs: |
        nats-top -s localhost -b -u -o output.txt
        grep -q client1 output.txt
        grep -q client2 output.txt
        grep -q "Connections Polled: 2" output.txt
```

nats-top.yaml
AnmolVirdi: Yes, noted. Making changes accordingly

nats-top.yaml
```diff
@@ -0,0 +1,91 @@
...
+    - name: test connection limit (-n) and sorting options (-sort)
+      runs: |
+        nats-top -n 1 -sort uptime -o nats-output-${{package.version}}.txt
```
maxgio92, line 57: I would test those options separately as with one connection the sorting result wouldn't change.

Something like

```yml
    - name: test connection limit (-n)
      runs: |
        nats-top -n 1 -o output.txt
        grep -q "Connections Polled: 1" output.txt
    - name: test sorting options (-sort)
      runs: |
        nats-top -sort uptime -u -o output.txt
        grep -q client2 <(tail -1 output.txt)
```

nats-top.yaml
```diff
@@ -0,0 +1,91 @@
...
+        export TERM=xterm && timeout 3 script -q -c "nats-top -d 1 -r 3" /dev/null
+    - name: Test Nats-top in TLS mode
+      runs: |
+        mkdir -p /tmp/certs
+        cd /tmp/certs
+        # Using mkcert to generate certs
+        mkcert -install
+        mkcert -cert-file server-cert.pem -key-file server-key.pem localhost 127.0.0.1
+        mkcert -client -cert-file client-cert.pem -key-file client-key.pem localhost 127.0.0.1
+        cp "$(mkcert -CAROOT)/rootCA.pem" ca-cert.pem
```
maxgio92, line 78: Can we move the PKI preparation into a separate step?

nats-top.yaml
```diff
@@ -0,0 +1,91 @@
...
+        cd ../..
+        # Killing previous nats-server process
+        pkill nats-server & sleep 2
+        # Certs validation before injecting them into NATS
+        openssl verify -CAfile ./tmp/certs/ca-cert.pem ./tmp/certs/server-cert.pem
+        openssl verify -CAfile ./tmp/certs/ca-cert.pem ./tmp/certs/client-cert.pem
```
maxgio92, line 84: Can we move the PKI preparation into a separate step? (all of the above commands, basically, until this one included).
I think we can skip the certificate validation, as we're basically testing `mkcert` by doing so, which is out of scope from this test suite I would say.

AnmolVirdi: - I tried adding certs validation to another step, but somehow the certs generated were not found in the next step. Even though I used docker runner

- I believe if in future TLS mode issue occurs, it could be due to 2 reasons: Nats-top tls issue or certs issue. Having a separate verification step would allow us to discover the issue quicker, i.e., if step fails, it's only because of nats-top tls issue and never certs since they're validated.

nats-top.yaml
AnmolVirdi: I'm adding nats server and nats sub together in a separate step since both of them are needed to set up server environment for `nats-top` tests

nats-top.yaml
AnmolVirdi: Thanks for pointing this, making changes accordingly.

nats-top.yaml
maxgio92: Also with the server is fine, I'd just remove it from the tests as it's part of the fixtures setup

nats-top.yaml
```diff
@@ -0,0 +1,98 @@
...
+        nats-top -sort uptime -u -o nats-output-${{package.version}}.txt
+        grep -q client2 <(tail -1 nats-output-${{package.version}}.txt)
+    - name: test output with delimiter (-l)
+      runs: |
+        nats-top -o nats-output-${{package.version}}.csv -l ","
+        if [ ! -s "nats-output-${{package.version}}.csv" ]; then
+          echo "Error: CSV File is empty!"
+          exit 1
+        fi
+        head -n 5 "nats-output-${{package.version}}.csv"
```
maxgio92, line 74: `head` doesn't exit 1 when the number of lines are greater than the ones of the file being read.

Instead, I'd validate via Python `csv` module to make it easier, like
```yml
    - name: test output with delimiter (-l)
      runs: |
        nats-top -o output.csv -l ","
        if [ ! -s "output.csv" ]; then
          echo "Error: CSV File is empty!"
          exit 1
        fi
        # Sanitize the input
        sed -i '/^$/d' output.csv
        # Ensure at least one column for each rows
        python -c "import csv; reader = csv.reader(open('output.csv', 'r')); [print(row[0]) for row in reader]"
```

nats-top.yaml
AnmolVirdi: I'm not sure if moving it out of tests will work since we're re-setting up nats server later in the TLS mode test

nats-top.yaml
AnmolVirdi: Thanks for pointing this, i've updated the test.

nats-top.yaml
```diff
@@ -0,0 +1,98 @@
...
+    - name: test setting refresh delay (-d) and refresh count (-r)
+      runs: |
+        export TERM=xterm && timeout 3 script -q -c "nats-top -d 1 -r 3" /dev/null
```
maxgio92, line 81: Is it required to set $TERM and `script`?
It seems that with `script` doesn't propagate the exit code:

```sh
/home/build # export TERM=xterm && timeout 3 script -c "nats-top -d 1 -non-existent-flag 3"
Script started, output log file is 'typescript'.
flag provided but not defined: -non-existent-flag

usage: nats-top [-s server] [-m http_port] [-ms https_port] [-n num_connections] [-d delay_secs] [-r max] [-o FILE] [-l DELIMITER] [-sort by]
                [-cert FILE] [-key FILE] [-cacert FILE] [-k] [-b] [-v|--version] [-u|--display-subscriptions-column]

Script done.
/home/build # echo $?
0
/home/build # export TERM=xterm && timeout 3 nats-top -d 1 -non-existent-flag 3
flag provided but not defined: -non-existent-flag

usage: nats-top [-s server] [-m http_port] [-ms https_port] [-n num_connections] [-d delay_secs] [-r max] [-o FILE] [-l DELIMITER] [-sort by]
                [-cert FILE] [-key FILE] [-cacert FILE] [-k] [-b] [-v|--version] [-u|--display-subscriptions-column]

/home/build # echo $?
1
```

nats-top.yaml
```diff
@@ -0,0 +1,102 @@
...
+          echo "Error: CSV File is empty!"
+          exit 1
+        fi
+        # Sanitize the input
+        sed -i '/^$/d' nats-output-${{package.version}}.csv
+        # Ensure at least one column for each rows
+        python -c "import csv; reader = csv.reader(open('nats-output-${{package.version}}.csv', 'r')); [print(row[0]) for row in reader]"
+    - name: test setting refresh delay (-d) and refresh count (-r)
+      runs: |
+        export TERM=xterm && timeout 3 script -q -c "nats-top -d 1 -r 3" /dev/null
```
maxgio92, line 81:
```suggestion
        export TERM=xterm
        script --quiet --return --command "timeout 3 nats-top -d 1 -r 3"
```
`--return` propagates the exit code from timeout that will propagate the exit code from `nats-top`, so test will fail when expected

nats-top.yaml
maxgio92: I tested the below locally, which works
```yml
    - name: Run NATS server
      runs: |
        nats-server -p 4222 -m 8222 & sleep 2 &
    - name: Run NATS subscribers
      runs: |
        nats sub client2 > /dev/null &
        sleep 2
        nats sub client1 > /dev/null &
        sleep 2
    - name: Test correct clients are in nats-top output
      runs: |
        nats-top -s localhost -b -u -o output.txt
        grep -q client1 output.txt
        grep -q client2 output.txt
        grep -q "Connections Polled: 2" output.txt
[...]
```

nats-top.yaml
maxgio92: > I tried adding certs validation to another step, but somehow the certs generated were not found in the next step. Even though I used docker runner
Yeah we need to move them out of /tmp

> I believe if in future TLS mode issue occurs, it could be due to 2 reasons: Nats-top tls issue or certs issue. Having a separate verification step would allow us to discover the issue quicker, i.e., if step fails, it's only because of nats-top tls issue and never certs since they're validated.

I see your point, makes sense. It's harmless anyway, fine for me.

nats-top.yaml
```diff
@@ -0,0 +1,102 @@
...
+        # Killing previous nats-server process
+        pkill nats-server & sleep 2
+        # Certs validation before injecting them into NATS
+        openssl verify -CAfile ./tmp/certs/ca-cert.pem ./tmp/certs/server-cert.pem
+        openssl verify -CAfile ./tmp/certs/ca-cert.pem ./tmp/certs/client-cert.pem
+        # Setting up NATS server
+        nats-server -p 4223 -m 8223 --tls --tlscert ./tmp/certs/server-cert.pem --tlskey ./tmp/certs/server-key.pem --tlscacert ./tmp/certs/ca-cert.pem & sleep 2
+        # Creating a NATS client and publishing sample message
+        nats sub -s "localhost:4223" --tlscert ./tmp/certs/client-cert.pem --tlskey ./tmp/certs/client-key.pem --tlsca ./tmp/certs/ca-cert.pem "test.subject" > /dev/null & sleep 1
+        nats-top -s localhost -m 8223 -cert ./tmp/certs/client-cert.pem -key ./tmp/certs/client-key.pem -cacert ./tmp/certs/ca-cert.pem -o nats-output-${{package.version}}.txt
```
maxgio92, line 101: Surprisingly it seems that `-cert`/`-key`/`-cacert` are no-op:

```shell
/home/build # nats-top -s localhost -m 8223 -cert pippo -key pluto -cacert paperino -o /dev/null
/home/build # echo $?
0
```

nats-top.yaml
```diff
@@ -0,0 +1,102 @@
...
+        grep -q "Connections Polled: 1" nats-output-${{package.version}}.txt >/dev/null 2>&1
```
maxgio92, line 102:
```suggestion
    - name: Run nats server in TLS mode
      runs: |
        # Killing previous nats-server process
        pkill nats-server & sleep 2
        # Setting up NATS server
        nats-server -p 4223 -m 8223 --tls --tlscert ./certs/server-cert.pem --tlskey ./certs/server-key.pem --tlscacert ./certs/ca-cert.pem & sleep 2
    - name: Run NATS subscribers in TLS mode
      runs: |
        nats sub -s "localhost:4223" --tlscert ./certs/client-cert.pem --tlskey ./certs/client-key.pem --tlsca ./certs/ca-cert.pem "test.subject" > /dev/null & sleep 1
    - name: Test NATS top in TLS mode
      runs: |
        nats-top -s localhost -m 8223 -cert ./certs/server-cert.txt -key ./certs/client-key.pem -cacert ./certs/ca-cert.pem -o output.txt
        grep -q "Connections Polled: 1" output.txt
```

nats-top.yaml
maxgio92: Reproduced also with upstream via Go, on a Ubuntu container, so it doesn't seem our issue:
```shell
docker run --rm -it --entrypoint bash ubuntu:latest
$ apt update && apt install -y golang >/dev/null 2>&1
$ go install github.com/nats-io/nats-server/v2@main
$ go install github.com/nats-io/natscli/nats@latest
$ go install github.com/nats-io/nats-top@latest
# generate certs
$ nats-server -p 4223 -m 8223 --tls --tlscert server-cert.pem --tlskey server-key.pem --tlscacert ca-cert.pem >/dev/null &
$ nats sub -s "localhost:4223" --tlscert client-cert.pem --tlskey client-key.pem --tlsca ca-cert.pem "test.subject" >/dev/null &
$ nats-top -s localhost -m 8223 -cert non-existent-cert.txt -key non-existent-key.pem -cacert non-existent-ca-cert.pem -o /dev/null
$ echo $?
0
```

maxgio92:
Sorry for being pedantic, once addressed all the point this test suite is going to be very good! Thanks!

nats-top.yaml
```diff
@@ -0,0 +1,102 @@
...
+    - name: Set up nats server and 2 client
+      runs: |
+        nats-server -p 4222 -m 8222 & sleep 2 &
+        nats sub client1 > /dev/null & sleep 2 & nats sub client2 > /dev/null & sleep 2
+    - name: Test and validate nats-top (basic)
+      runs: |
+        nats-top -s localhost -b -u -o nats-output-${{package.version}}.txt
+        grep -q "client1" nats-output-${{package.version}}.txt >/dev/null 2>&1
+        grep -q "client2" nats-output-${{package.version}}.txt >/dev/null 2>&1
+        grep -q "Connections Polled: 2" nats-output-${{package.version}}.txt >/dev/null 2>&1
```
maxgio92, line 59: `>/dev/null 2>&1` not required with `grep -q`

nats-top.yaml
```diff
@@ -0,0 +1,102 @@
...
+    - name: test connection limit (-n)
+      runs: |
+        nats-top -n 1 -o nats-output-${{package.version}}.txt
+        grep -q "Connections Polled: 1" nats-output-${{package.version}}.txt >/dev/null 2>&1
```
maxgio92, line 63: `>/dev/null 2>&1` not required with `grep -q`

nats-top.yaml
AnmolVirdi: I've added changes accordingly. Thanks

nats-top.yaml
AnmolVirdi: Thank you for suggesting this! We can use `script` for running any interactive tests now, for future cases.

nats-top.yaml
```diff
@@ -0,0 +1,102 @@
...
+        mkcert -install
+        mkcert -cert-file server-cert.pem -key-file server-key.pem localhost 127.0.0.1
+        mkcert -client -cert-file client-cert.pem -key-file client-key.pem localhost 127.0.0.1
+        cp "$(mkcert -CAROOT)/rootCA.pem" ca-cert.pem
+        cd ../..
+        # Killing previous nats-server process
+        pkill nats-server & sleep 2
+        # Certs validation before injecting them into NATS
+        openssl verify -CAfile ./tmp/certs/ca-cert.pem ./tmp/certs/server-cert.pem
+        openssl verify -CAfile ./tmp/certs/ca-cert.pem ./tmp/certs/client-cert.pem
```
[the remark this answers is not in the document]
AnmolVirdi: - Removing the tmp directory and moving certs generation to separate step.

nats-top.yaml
AnmolVirdi: I've added these changes in the next commit.

nats-top.yaml
AnmolVirdi: Removed, Thanks

nats-top.yaml
AnmolVirdi: Thanks for pointing this out @maxgio92.
This is because in the current test we were accidentally trying to access http_port and not `https`. I've pushed another commit which makes sure the certs are being used properly.
- Added `tlsverify` flag to make sure we use certs for client verifications
- Accessing nats-top using tls port

nats-top.yaml
AnmolVirdi: Thanks for investigating it was `tmp` dir issue. The tests work fine now!

nats-top.yaml
maxgio92: Yes, that's awesome indeed @AnmolVirdi, thanks

nats-top.yaml
```diff
@@ -0,0 +1,112 @@
...
+        pkill nats-server & sleep 2
+        nats-server -p 4223 --https_port 8223 --tlsverify --tlscert ./certs/server-cert.pem --tlskey ./certs/server-key.pem --tlscacert ./certs/ca-cert.pem &
+        sleep 2
+    - name: Set up one NATS subscriber in TLS mode
+      runs: |
+        nats sub -s "localhost:4223" --tlscert ./certs/client-cert.pem --tlskey ./certs/client-key.pem --tlsca ./certs/ca-cert.pem "test.subject" > /dev/null &
+        sleep 1
+    - name: Test NATS top in TLS mode
+      runs: |
+        nats-top -s "localhost" -ms 8223 -u -cert ./certs/client-cert.pem -key ./certs/client-key.pem -cacert ./certs/ca-cert.pem -o nats-output-${{package.version}}.txt
```
maxgio92, line 110: @AnmolVirdi  So as discussed it seems that we need to explicitely set the URL to use TLS (https) otherwise -cert/-key/-cacert are no op (it strange nats-top doesn't complain BTW)

AnmolVirdi: The issue was with passing tls mode via `-ms` flag for https port. Earlier I was using '-m' which is default http port.

maxgio92 (approved):

PR #44323, opened 2025-03-03
fix pyspark-scala-2.13 GHSA-r978-9m6m-6gm6

dnegreira:


dnegreira:
Not the right action to take here, closing.

PR #44380, opened 2025-03-03
pipelines: test/tw/ldd-check: Default to packages: ${{context.name}}

dannf:
If no inputs are provided default to behaving as if `packages: ${{context.name}}` was provided, as that's by far the most common mode in-use.

### DIFF
--- a/pipelines/test/tw/ldd-check.yaml
+++ b/pipelines/test/tw/ldd-check.yaml
@@ -22,7 +22,12 @@ inputs:
 pipeline:
   - name: "check for missing library dependencies using ldd"
     runs: |
+      if [ -z "${{inputs.files}}" ] && [ -z "${{inputs.packages}}" ]; then
+          packages="${{context.name}}"
+      else
+          packages="${{inputs.packages}}"
+      fi
       ldd-check \
           --files="${{inputs.files}}" \
-          --packages="${{inputs.packages}}" \
+          --packages="$packages" \
           --verbose="${{inputs.verbose}}"


pipelines/test/tw/ldd-check.yaml
```diff
@@ -22,7 +22,12 @@ inputs:
 pipeline:
   - name: "check for missing library dependencies using ldd"
     runs: |
+      if [ -z "${{inputs.files}}" ] && [ -z "${{inputs.packages}}" ]; then
+          packages="${{context.name}}"
```
sergiodj, line 26: It seems like `context.name` isn't defined:

```
Error: compiling build: compiling subpackage "acl-dev" tests: compiling Pipeline[1]: compiling Pipeline[0]: mutating runs: variable context.name not defined
```

(This is from the regular pipeline build)

sergiodj (changes requested):

pipelines/test/tw/ldd-check.yaml
sergiodj: Dann just told me that `context.name` is new.  Marking this discussion as resolved.

smoser, line 26: This probably right. 
If we're only ever passing the single package, then we can just make the default be context.name and even drop support for files. 

If needed, we can at another pipeline that calls ldd-check with files. Then the usage is super short. 

Uses: tw/ldd-check

Or

Uses:ldd-check-files
  With:
    Files: ...

(Forgive the phone added capitalization)

dannf:
superseded by https://github.com/chainguard-dev/tw/pull/20

dannf:
superseded by #45745

PR #44398, opened 2025-03-04
Fix druid pombump and improve test

smoser:
*  Move bump of netty
    
    PR #37343 (da1effa048283f) explains why this change is made.
    
    The druid service was showing stacktrace like:
    
     > Exception in thread "main" java.lang.NoClassDefFoundError: io/netty/handler/ssl/SslContext
     >  at org.apache.druid.server.emitter.EmitterModule.configure(EmitterModule.java:78)
     >  at com.google.inject.spi.Elements$RecordingBinder.install(Elements.java:409)
     >  at com.google.inject.spi.Elements.getElements(Elements.java:108)
     >  at com.google.inject.util.Modules$OverrideModule.configure(Modules.java:213)
     >  at com.google.inject.AbstractModule.configure(AbstractModule.java:66)
     >  at com.google.inject.spi.Elements$RecordingBinder.install(Elements.java:409)
    
    Such can be seen in the log dir in the *.stdout files such as 'router.log.stdout'.

 *  improve druid test
    
    This more carefully looks at the druid output and looks for
    successfull http service running.

### DIFF
--- a/druid.yaml
+++ b/druid.yaml
@@ -1,7 +1,7 @@
 package:
   name: druid
   version: "32.0.0"
-  epoch: 5
+  epoch: 6
   description: Apache Druid is a high performance real-time analytics database.
   copyright:
     - license: Apache-2.0
--- a/druid.yaml
+++ b/druid.yaml
@@ -96,12 +96,37 @@ test:
       packages:
         - python3
         - perl
+        - curl
   pipeline:
     - uses: test/daemon-check-output
       with:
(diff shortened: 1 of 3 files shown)


dakaneye (approved):

dakaneye:
@rawlingsj it looks like (from git-blame) that something in automation added that netty-handler thing to the pombump-deps.yaml, when it should've updated the properties file instead. I absolutely have no idea how it should've known the right thing to do but wanted to call this out (no action needed) as the reason for this specific issue.

PR #44400, opened 2025-03-04
Paulgibert/chromium typo

paulgibert:
Fixed typo in chromium.yaml description


paulgibert:
Some weirdness with prism somehow got wrapped up in this commit.

PR #44487, opened 2025-03-04
add soci-snapshotter

joshrwolf:


soci-snapshotter.yaml
```diff
@@ -0,0 +1,59 @@
...
+  - uses: strip
+
+subpackages:
+  - name: soci-snapshotter-static
+    dependencies:
+      runtime:
+        - soci-snapshotter
+    options:
+      # Do not provide any embedded components
+      no-provides: true
```
imjasonh, line 41: What embedded components are we filtering out with this?

soci-snapshotter.yaml
```diff
@@ -0,0 +1,59 @@
...
+    pipeline:
+      - runs: |
+          make STATIC=1
+          make CMD_DESTDIR=${{targets.contextdir}} install
+      - uses: strip
```
imjasonh, line 46: IIRC certain compile-time settings and binary stripping makes govulncheck analysis less reliable. Can we remove these, and re-add them later when we're sure they're needed?

soci-snapshotter.yaml
joshrwolf: providing zlib: https://github.com/awslabs/soci-snapshotter/blob/main/docs/build.md#dependencies

PR #44564, opened 2025-03-04
Usrmerge usrsbin 4

a-crate:


nsd.yaml
```diff
@@ -39,7 +43,9 @@ pipeline:
 
   - uses: autoconf/configure
     with:
-      opts: CC=/usr/bin/gcc
+      opts: |
+        CC=/usr/bin/gcc \
```
justinvreeland, line 47: Do configure scripts take variables like this? Make does but, I suspect this doesn't do anything but if it did it would block the openssf compiler options from being added by the wrapper and should probably be removed.

a-crate: I think so, it's in the [manual](https://www.gnu.org/software/autoconf/manual/autoconf-2.60/autoconf.html#Manual-Configuration) as an example. Section 13.1

justinvreeland: Oh neat, thanks I'll put up a PR to remove this.

zmarano (approved):

PR #44576, opened 2025-03-04
git: add signing and verification tools as runtime dependencies.

wlynch:
These tools are exec-ed out to by git at run time to perform signing and verification for git commit signatures.

This is a bit of a grey area w.r.t. packaging - technically git can function without these, but since these are configured by default these feel like they're worth including? The alternative is we can bundle these at the image layer.
Open to suggestions here!

### DIFF
--- a/git.yaml
+++ b/git.yaml
@@ -1,12 +1,17 @@
 package:
   name: git
   version: "2.48.1"
-  epoch: 1
+  epoch: 2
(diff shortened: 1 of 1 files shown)


wlynch:
Closing in favor of adding to the image directly since this is an optional dependency. https://github.com/chainguard-images/images-private/pull/8036

PR #44369, opened 2025-03-03
Supress 'apk info' noise when testing docs

murraybd:
`apk info` will produce some WARNING messages when being run and this supresses those.  Here's an example:

```
2025/03/03 13:44:21 WARN + apk info -L gdbm-doc
2025/03/03 13:44:21 WARN + grep ^usr/share/info/
2025/03/03 13:44:21 WARN WARNING: opening /home/brian-murray/Working/wolfi-os/packages: No such file or directory
2025/03/03 13:44:21 WARN WARNING: opening from cache https://packages.wolfi.dev/os: No such file or directory
2025/03/03 13:44:21 WARN + '[' -f /usr/share/info/gdbm.info ]
2025/03/03 13:44:21 WARN + info -f /usr/share/info/gdbm.info -o -
2025/03/03 13:44:21 WARN + wc -l
2025/03/03 13:44:21 WARN + '[' 74 -gt 0 ]
2025/03/03 13:44:21 WARN + doc_files=true
2025/03/03 13:44:21 WARN + apk info -L gdbm-doc
2025/03/03 13:44:21 WARN + grep ^usr/share/
2025/03/03 13:44:21 WARN WARNING: opening /home/brian-murray/Working/wolfi-os/packages: No such file or directory
2025/03/03 13:44:21 WARN WARNING: opening from cache https://packages.wolfi.dev/os: No such file or directory
```

PR #44595, opened 2025-03-05
Revert "Supress 'apk info' noise when testing docs"

zmarano:
Reverts wolfi-dev/os#44369

Breaks some doc test cases.


gaughen (approved):

murraybd:
> Reverts #44369
> 
> Breaks some doc test cases.

It'd be helpful to know what broke. Could you share?

murraybd:
The problematic line follows:

```
       # We seem to have a lot of "empty" doc packages...
-      if [ $(apk info -qL "$doc_pkg" | wc -l) -le 3 ]; then
+      if [ $(apk info -qL "$doc_pkg" | wc -l) -le 2 ]; then
```

So the count just needed to change.

PR #44644, opened 2025-03-05
python: Use more generic march for rosetta compatibiity

pnasrat:
See: https://github.com/wolfi-dev/os/pull/38076

Related: chainguard-dev/customer-issues#2094


python-3.11.yaml
```diff
@@ -79,6 +94,7 @@ pipeline:
          --with-builtin-hashlib-hashes=blake2 \
          --with-lto \
          --with-wheel-pkg-dir=/usr/share/python-wheels
+      exit 1
```
techalchemy, line 97: debugging?

pnasrat: doh

techalchemy (approved):
lgtm, thanks!

justinvreeland (dismissed):

pnasrat:
Test errors on 3.12 and 3.13 will investigate

OddBloke (approved):

dannf:
Is there any additional testing we should do to make sure this isn't going to break anything (abi-checker maybe?)

Will this noticeably regress performance for any customers?

We may very well want to do this anyway of course - it would just be good to understand the trade-offs and do what we can to avoid other customer escalations.

PR #44098, opened 2025-02-28
coreutils (and dependent packages) - usrmerge usrsbin

a-crate:
Same as #43122, but these packages had prior usr merges (either /bin or /sbin)

PR #44665, opened 2025-03-05
glibc - usrmerge usr/sbin

smoser:


smoser:
folding back into https://github.com/wolfi-dev/os/pull/44098 . it needs util-linux which needs to be sbin-merged...

PR #44704, opened 2025-03-05
use 9.0.103 to correct scanners prevents false positives

developer-guy:
There is an issue reported related to the vuln scanners that scanner reports false positive vulnerabilities addressed in 9.0.101 to be flagged as vulnerable in 9.0.2.

```bash
dotnet-9          9.0.2-r0   9.0.101-r0  apk   CVE-2024-43498       Critical  
```

-->

#### For new package PRs only

### DIFF
--- a/dotnet-9.yaml
+++ b/dotnet-9.yaml
@@ -1,6 +1,6 @@
 package:
   name: dotnet-9
-  version: 9.0.2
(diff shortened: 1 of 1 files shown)


pdeslaur:
Would it make more sense to update our advisories with the correct fixed version?

PR #44728, opened 2025-03-05
repmgr-17 - usrmerge /usr/sbin

smoser:
repmgr-17 has coreutils in its runtime deps. So pull it out so that it can get that.


a-crate:
Needs /usr/sbin-merged coreutils to build.

zmarano (approved):

PR #44783, opened 2025-03-06
pypy3: remove pypy-bootstrap and build pypy3 standalone.

sil2100:
Currently, whenever we build pypy3.10 or pypy3.11 we rely on the pypy-bootstrap package, which is a special 'bootstrapping' package that should not be used for anything else. However, in theory, pypy-bootstrap does ship the pypy-c python2.7-based interpreter in a custom directory, and in theory could be used in images - which is not ideal.

Let's get rid of pypy-bootstrap completely as we don't really need it. Instead, let's do what Debian does: build a throwaway cpython2.7 binary just to execute rpython for pypy build purposes. Since sadly python2.7 is required per: https://doc.pypy.org/en/latest/build.html and there's no way around it. But this way we at least have no 2.7 interpreter lying around in the repos.

The build is a bit longer, but not that much longer actually.


dannf (commented):
LGTM! @utieyin @jamonation  - WDYT? If deliver is good with it, I'll approve.

jamonation:
Makes sense to me, keeps things contained to just the build env as opposed to touching the package repo, so +1 here.

dannf (approved):

sergiodj (approved):

PR #44834, opened 2025-03-06
fix(selenium): Use correct versions of things

EyeCantCU:
Docker Selenium should always be built with the versions of gRPC, Netty, and OTel consumed by upstream (unless we've a compelling reason not to, i.e. vulnerability remediation). This addresses issues encountered when tracing is enabled. Likewise, the version of DS should always be consistent with the version of Selenium. The two should never be mismatched

Also land Selenium 4.29.0 and move on to Java 21 per upstream


jamonation (approved):
Matches what I found when I was poking at these packages - and thanks for tracking down that OTEL fix - not an easy one to spot

PR #44902, opened 2025-03-06
Copy NGINX.CONF from bitnami/containers repo

Aditevil:


Aditvil-Dev:
separated out the bitnami package here, https://github.com/chainguard-dev/enterprise-packages/pull/16124

PR #44905, opened 2025-03-06
Fix: openssh pam config subpackage name.

zmarano:


dannf:
Thanks! Do you mind withdrawing the files from the last epoch, so no `-configuration` package remains that could accidentally get installed? I'd suggest removing all of the files from that epoch, so there's also not a weird incomplete build.

zmarano:
> Thanks! Do you mind withdrawing the files from the last epoch, so no `-configuration` package remains that could accidentally get installed? I'd suggest removing all of the files from that epoch, so there's also not a weird incomplete build.

Absolutely, done.

dannf (approved):

PR #44912, opened 2025-03-06
Add `.dir-locals.el`

sergiodj:
Emacs users also contribute to Wolfi!

Add a `.dir-locals.el` file containing some eglot configuration that allows [yaml-language-server](https://github.com/redhat-developer/yaml-language-server) to be used.


dannf (approved):
yas.

PR #44947, opened 2025-03-06
llvm: fix /usr/bin/llvm-config symlink

OddBloke:
And add it to testing to avoid future regressions.

Previously this package was named `llvm-19`, which would have produced the correct path, so this was a minor miss in the package rename.


roaksoax (dismissed):
Lgtm

debasishbsws (approved):

PR #45009, opened 2025-03-07
mariadb-11.7 - Include a configuration file with appropriate defaults

a-crate:
Post `/bin` usrmerge, `/bin/mariadbd` now exists because of the `/bin` symlink to `/usr/bin`, where `mariadbd` actually resides. The mariadb docker entrypoint startup scripts check for the existence of `/bin/mariadbd`, and guess that the installation root is `/` when it is actually `/usr`, causing a failure to find some startup SQL scripts.

### DIFF
--- a/mariadb-11.7.yaml
+++ b/mariadb-11.7.yaml
@@ -1,7 +1,7 @@
 package:
   name: mariadb-11.7
   version: "11.7.2"
-  epoch: 1
+  epoch: 2
   description: "The MariaDB open source relational database"
   copyright:
     - license: GPL-3.0-or-later
@@ -10,6 +10,7 @@ package:
     memory: 32Gi
   dependencies:
     runtime:
+      - mariadb-config
       - pwgen
     provides:
       - mariadb=${{package.full-version}}
@@ -152,6 +153,17 @@ pipeline:
       mkdir -p "${{targets.destdir}}"/var/lib/mysql
 
 subpackages:
+  - name: "${{package.name}}-config"
+    description: Mariadb configuration file with appropriate defaults
+    dependencies:
+      provides:
+        - mariadb-config=${{package.full-version}}
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/etc/
+          echo "[mariadbd]" >> "${{targets.subpkgdir}}"/etc/my.cnf
+          echo "basedir = /usr" >> "${{targets.subpkgdir}}"/etc/my.cnf
+
   - name: "${{package.name}}-dev"
     description: "headers for mariadb"
     dependencies:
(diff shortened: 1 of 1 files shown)


xnox:
Please check if similar change needs to happen in all the other forks of this and all other versions streams. E.g. mysql and mariadb version streams.

mariadb-11.7.yaml
```diff
@@ -152,6 +153,17 @@ pipeline:
...
+  - name: "${{package.name}}-config"
+    description: Mariadb configuration file with appropriate defaults
+    dependencies:
+      provides:
+        - mariadb-config=${{package.full-version}}
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/etc/
+          echo "[mariadbd]" >> "${{targets.destdir}}"/etc/my.cnf
+          echo "basedir = /usr" >> "${{targets.destdir}}"/etc/my.cnf
```
jamonation, line 165:
```suggestion
          mkdir -p ${{targets.subpkgdir}}/etc/
          echo "[mariadbd]" >> "${{targets.subpkgdir}}"/etc/my.cnf
          echo "basedir = /usr" >> "${{targets.subpkgdir}}"/etc/my.cnf
```

mariadb-11.7.yaml
```diff
@@ -201,7 +213,7 @@ subpackages:
           mv "${{targets.destdir}}"/usr/bin/mariadb "${{targets.subpkgdir}}"/usr/bin/
 
   - name: "${{package.name}}-oci-entrypoint"
-    description: Entrypoint for using HAProxy in OCI containers
+    description: Entrypoint for using MariaDB in OCI containers
```
jamonation, line 216: Thank you! That was bugging me all day

a-crate:
> Please check if similar change needs to happen in all the other forks of this and all other versions streams. E.g. mysql and mariadb version streams.

Does need to happen to all mariadb version streams, but does not affect mysql.

FWIW I really tried to write a test that can reproduce this issue in a melange docker container (as opposed to a chainguard image docker container) but could not manage to do so, something about this is specific to that environment and I have not discovered what exactly it is.

mariadb-11.7.yaml
a-crate: Oops, thank you.

mariadb-11.7.yaml
a-crate: Me too!

xnox (approved):

xnox:
hm, i do wonder if life would be better if we simply our $PATH to exclude the usr-merged locations.

roaksoax:
@xnox maybe when we finish usrmerge?

xnox:
> @xnox maybe when we finish usrmerge?

Ack

Let me try to prototype a draft only.

I fear creating config file where there was none previously, and people maybe bind mounting their own config there and still observing breakage.

roaksoax:
@xnox ack. And yes, was thinking about the same thing of people bind mounting their own config. Which is why I wonder if the entrypoint should just hardcode /usr/?

But @jamonation you probably know better.

a-crate:
Closed in favor of the approach in #45189 to avoid breaking users mounting configs.

jamonation:
Thanks for checking @roaksoax. @a-crate and I discussed, and yes, without our my.cnf allowing for conf.d by default, whatever users are doing with our image means we're pretty stuck having to let them provide their own my.cnf files at runtime.

PR #45154, opened 2025-03-07
glibc: better support ldconfig trigger with different shells

xnox:
apk-tools orders triggers in runtime dependency order, meaning glibc
trigger is guaranteed to execute before busybox trigger sets up "sh"
symlink.

Move glibc trigger to apk-tools post-commit hook, such that it is
executed after all triggers have been fired, to ensure shell is
accessible.

Separately, will submit glibc upstream feature change to make it
possible to use "#!/usr/bin/ldconfig.trigger" as a trigger which
ignores all arguments.

Add test cases to verify that bootstraps work with various shells.

### DIFF
--- a/glibc.yaml
+++ b/glibc.yaml
@@ -736,7 +736,6 @@ test:
   environment:
     contents:
       packages:
-        - apk-tools
         - gcc
         - glibc-dev
   pipeline:
--- a/glibc.yaml
+++ b/glibc.yaml
@@ -1,7 +1,7 @@
 package:
   name: glibc
   version: 2.40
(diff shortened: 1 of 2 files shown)


dannf (dismissed):
thanks!

xnox:
@dannf there was lint failure, which is now fixed. Please re-review.

dannf (approved):

dannf:
approved :) but maybe wait until Monday to merge?

xnox:
> approved :) but maybe wait until Monday to merge?

we do release to prod on Fridays =)

PR #45242, opened 2025-03-07
add a test/headers pipeline and a handful of working examples

dustinkirkland:
With some help from @xnox, I have a test/headers pipeline that attempts to verify that installed header files can be included.

It will try both c and c++, in sequence, succeeding if either succeed.

It will fail for a variety of reasons -- and the test itself could certainly use more refinement -- please contribute!  (Or, if this is better suited to land in tw, I'm cool with that too....)

Note that this test did in fact find one real issue, with two particular header files in libb64 that are unusable on Wolfi as is.  So, within this PR is also a patch, bufsiz.patch for that library.  With that patch (which Debian and Ubuntu also carry) -- the test succeeds and the header is in fact useful.

It also contains successful usage with several other packages (with epoch bumps), that previously had no tests.


libb64-1.2/bufsiz.patch
```diff
@@ -0,0 +1,26 @@
+diff --git a/include/b64/decode.h b/include/b64/decode.h
+index 12b16ea..6d79ac5 100644
+--- a/include/b64/decode.h
++++ b/include/b64/decode.h
+@@ -22,7 +22,7 @@ namespace base64
+ 		base64_decodestate _state;
+ 		int _buffersize;
+ 
+-		decoder(int buffersize_in = BUFFERSIZE)
++		decoder(int buffersize_in = BUFSIZ)
```
dannf, line 10: Note that this is a functional change. Upstream passes `-DBUFFERSIZE=16777216` by default when building the library:

https://github.com/libb64/libb64/blob/ce864b17ea0e24a91e77c7dd3eb2d1ac4175b3f0/base64/Makefile#L16

Where glibc sets this to 8192:

https://github.com/bminor/glibc/blob/9b91484bee8f6f1bf1e2d26a8df461b553784528/libio/stdio.h#L100

It's not going to break anything - this is intended to be configurable. But what do you think of this?

https://github.com/libb64/libb64/pull/14

If you're good with that, I can try swapping this patch out with a cherry-pick of that.

dannf (dismissed):
Thanks Dustin, great addition!

What do you think about moving this into https://github.com/chainguard-dev/tw like we did w/ ldd-check? That should make it easier to iterate on the code w/o having to resync pipelines across repos (as much). Happy to take a shot at that if you like.

Other than that, a few comments inline.

pipelines/test/headers.yaml
```diff
@@ -0,0 +1,76 @@
...
+        AC_INIT([example],[1.0],[example@example.com])
+        AC_PREREQ(2.69)
+        AC_PROG_CXX
+        AC_LANG(C++)
+        AC_CHECK_HEADER($f,[cxx_header_found=yes],,)
+        AS_IF([test "x\$cxx_header_found" != "xyes"], [AC_MSG_ERROR([Unable to check header])])
+        AC_OUTPUT()
+      EOF
+      }
+      pkg=$(basename ${{targets.contextdir}})
```
smoser, line 61: You can use context.name now.

dustinkirkland:
Yeah, fantastic, I'll work on moving it into tw this weekend!

pipelines/test/headers.yaml
```diff
@@ -0,0 +1,76 @@
+name: headers
+
+needs:
+  packages:
+    - autoconf
+    - build-base
+    - file
```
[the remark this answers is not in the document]
dustinkirkland: Good catch.  I was using it previously, but dropped that logic.  Will remove.

pipelines/test/headers.yaml
dustinkirkland: Thanks, will do.  I'm moving the vast majority of this logic to tw.

pipelines/test/headers.yaml
```diff
@@ -0,0 +1,76 @@
...
+        AC_PREREQ(2.69)
+        AC_PROG_CXX
+        AC_LANG(C++)
+        AC_CHECK_HEADER($f,[cxx_header_found=yes],,)
+        AS_IF([test "x\$cxx_header_found" != "xyes"], [AC_MSG_ERROR([Unable to check header])])
+        AC_OUTPUT()
+      EOF
+      }
+      pkg=$(basename ${{targets.contextdir}})
+      for f in $(apk info -L "$pkg" 2>/dev/null | grep "^usr/include/.*\.h[p]*$" | sed -e "s:^usr/include/::"); do
```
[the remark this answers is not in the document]
dustinkirkland: Thanks!  Will use.

libb64-1.2/bufsiz.patch
dustinkirkland: Thanks for the review -- yes please, way better.  You can go ahead and make that fix / change to libb64-1.2 separately, and land that, while I'm navigating landing the header-check script in tw, and then I'll rework this pipeline and branch accordingly....

dannf: Went on a ride and, with a cleared head, realized this is probably not the most-wolfi approach. I think what upstream had in mind is that `-DBUFFERSIZE` should be explicitly set by any app that uses it. Instead of bending upstream to have a default, I think the better solution here is to let the pipeline pass arbitrary flags. How about we use the generic `opts:` convention that `autoconf/configure` provides?

dustinkirkland: Okay, so you're suggesting that in the test: section, add a cflags variable, or some such, that the header-check pipeline would use, if present?  If so, I think that's reasonable, I guess...

dannf: To be clear, I don't want to block the addition of this test on this interface, but perhaps we hold off on adding the test to this package until we decide how to do this, since it will set a pattern.

The generic `opts:` input for `autoconf/configure` seems to just let you append arbitrary strings to the pipeline. That would work - we could have:

```yaml
  pipeline:
    - uses: test/headers
       with:
          opts: -DBUFFERSIZE=123456
```

I like the thought of a specific `CFLAGS` better and, well, we already have that interface!

```yaml
test:
  environment:
    environment:
      CFLAGS: -DBUFFERSIZE=123456
  pipeline:
    - uses: test/tw/header-check
```

pipelines/test/headers.yaml
dustinkirkland: Moved all of this to tw.

libb64-1.2/bufsiz.patch
dustinkirkland: Have a look now, please @dannf ...  Note there there's a separate PR for tw that will have to land, in order to catch and use the new configure-opts option.

pipelines/test/headers.yaml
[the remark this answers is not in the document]
dustinkirkland: Done!  I used this exactly in the tw/header-check script.

PR #45433, opened 2025-03-09
Add ldd-check test to several packages

murraybd:


apache-activemq-artemis/pombump-deps.yaml
gaughen, line 1: why is this included in the ldd-check test pr?

murraybd: It's just a white space change due to running `yam` on apache-activemq-artemis directory in addition to the .yaml for that package.

smoser (commented):
How willing would you be to wait for a ldd-check update that included https://github.com/chainguard-dev/tw/pull/20 ?

If we have the ability to use `${{context.name}}` , then that means all of these:

```yaml
    - uses: test/tw/ldd-check
      with:
        packages: abseil-cpp
```

Turn into:
```yaml
    - uses: test/tw/ldd-check
```

It _would_ require 
1. an included change `pipelines/test/tw/ldd-check.yaml` to use `${{context.name}}`
2. support to be in all necessary places for ${{context.name}} use , which landed https://github.com/chainguard-dev/melange/pull/1810

@dannf do you know if 1810 made it into all necesary places?

murraybd:
> How willing would you be to wait for a ldd-check update that included [chainguard-dev/tw#20](https://github.com/chainguard-dev/tw/pull/20) ?

While I'm not in a hurry I've changes similar to this PR to 225 yaml files and would prefer not to have to redo the work.

gaughen (approved):

PR #45713, opened 2025-03-10
Nats box adder

AnmolVirdi:
**Nats-box** _Current version: 0.16.0
- Virtual package containing NATS utility tools.

**Type:** Virtual package
**REF:** [https://github.com/nats-io/nats-box](https://github.com/nats-io/nats-box)
[https://github.com/nats-io/nats-box/blob/main/Dockerfile](https://github.com/nats-io/nats-box/blob/main/Dockerfile)

**Includes:**
      - bash
      - ca-certificates
      - curl
      - figlet
      - jq


AnmolVirdi:
**Notes:**
- This package includes `bash` because `nats-box` is mostly used in interactive mode only.

nats-box.yaml
```diff
@@ -0,0 +1,52 @@
...
+  pipeline:
+    - name: "Verify basic components"
+      runs: |
+        nats --version
+        nats-top --version
+        nsc --version
+        update-ca-certificates --version
+        curl --version
+        jq --version
+        figlet -v
```
Dentrax: Not sure if we should have those since each package have its own test but... 🤔

nats-box.yaml
```diff
@@ -0,0 +1,47 @@
...
+  github:
+    identifier: nats-io/nats-box
+    strip-prefix: v
+
+test:
+  pipeline:
+    - name: "Verify entrypoint.sh and profile.sh files"
+      runs: |
+        test -s "../../etc/profile.d/profile.sh"
+        test -s "../../entrypoint.sh"
```
Dentrax, line 47: Is is possible to run the entrypoint here?

AnmolVirdi: It would just start a login shell. Both entrypoint as well as profile scripts have minimal functionality.
REF: [https://github.com/nats-io/nats-box/blob/main/entrypoint.sh](https://github.com/nats-io/nats-box/blob/main/entrypoint.sh)


Please let me know if you think it's a good idea to run entrypoint

nats-box.yaml
```diff
@@ -0,0 +1,49 @@
...
+      repository: https://github.com/nats-io/nats-box
+      tag: v${{package.version}}
+      expected-commit: b73821692fc596b210816de93c5011047298646b
+
+  - runs: |
+      mkdir -p ${{targets.destdir}}/etc/profile.d
+      sed -i 's/^.*MARKER-REPLACE-LINE-IN-CI.*/echo "nats-box v${{package.version}}" >\&2/' profile.sh
+      mv profile.sh ${{targets.destdir}}/etc/profile.d
+      mv entrypoint.sh ${{targets.destdir}}/entrypoint.sh
+      chmod 777 ${{targets.destdir}}/entrypoint.sh
```
Dentrax, line 36:
```suggestion
      install -Dm755 profile.sh ${{targets.destdir}}/etc/profile.d
      install -Dm755 entrypoint.sh ${{targets.destdir}}/entrypoint.sh
```

nats-box.yaml
Dentrax: ack - can we expand the tests a bit here, if possible?

AnmolVirdi: @Dentrax this his a virtual package which consists of NATS packages which have their own test suites.

mritunjaysharma394: Thats great but I think it would be great to see these utilities being referenced in some form of test to have higher confidence in our package? For example

`nats pub -s demo.nats.io test 'Hello World'` as stated in upstream too? https://github.com/nats-io/nats-box?tab=readme-ov-file#getting-started

AnmolVirdi: Thanks for the feedback @mritunjaysharma394, I've actually added these tests at image level in the Image PR
REF: [NATS-Box PR](https://github.com/chainguard-images/images-private/pull/8185/files)

nats-box.yaml
```diff
@@ -0,0 +1,45 @@
+package:
+  name: nats-box
+  version: 0.17.0
+  epoch: 0
+  description: NATS utilities toolbox.
+  dependencies:
+    runtime:
```
Dentrax, line 7: Needs additional runtime deps: https://github.com/nats-io/nats-box/blob/6cdf3ae5d1e4499137bc650ea1b8d8fc3adb9197/Dockerfile#L20-L21

nats-box.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/nats-io/nats-box
+      tag: v${{package.version}}
+      expected-commit: b73821692fc596b210816de93c5011047298646b
+
+  - runs: |
+      mkdir -p ${{targets.destdir}}/etc/profile.d
```
Dentrax, line 29: Should we also need to create additional dirs? `/nsc`, `/home/nats`, `/nsc/nkeys`, `/nsc/.config`

nats-box.yaml
Dentrax: Also see: https://github.com/nats-io/nats-box/blob/6cdf3ae5d1e4499137bc650ea1b8d8fc3adb9197/profile.sh#L3

Dentrax (changes requested):

nats-box.yaml
AnmolVirdi: I've created them at Image level since they need certain permissions as well

nats-box.yaml
Dentrax: Will we have a compat package for `/usr/local/bin/nats-top`?

AnmolVirdi: Yes, already added it in nats-box image addition PR.

Dentrax (approved):

kranurag7 (approved):

PR #45750, opened 2025-03-10
tw - bump to v0.0.4, add header-check and utilize in uthash

smoser:
tw 0.0.4 contains 'header-check' which can be run to do a autoconf like test on header files.

This:
1. bumps tw to v0.0.4
2. adds pipelines/test/tw/header-check.yaml
3. uses `uses: test/tw/header-check` from uthash.yaml

### DIFF
--- a/tw.yaml
+++ b/tw.yaml
@@ -1,14 +1,16 @@
 package:
   name: tw
-  version: 0.0.3
+  version: 0.0.4
   epoch: 0
-  description:
+  description: Testing tools
   options:
     no-provides: true
   copyright:
     - license: Apache-2.0
 
 environment:
+  environment:
+    GOMODCACHE: /var/cache/melange
   contents:
     packages:
       - build-base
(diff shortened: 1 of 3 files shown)


dustinkirkland (approved):

tw.yaml
```diff
@@ -47,6 +57,43 @@ subpackages:
...
+    pipeline:
+      - runs: |
+          make -C header-check MELANGE_CONTEXTDIR=${{targets.contextdir}} melange-install
+    test:
+      pipeline:
+        - runs: |
+            [ -f /usr/bin/header-check ]
+            [ -x /usr/bin/header-check ]
+
+  - name: syspeek
```
dannf, line 78: I don't see this tool mentioned anywhere - did you mean to add it? If so, should it come with a sample pipeline?

joshrwolf: its relatively recent addition to `tw`: https://github.com/chainguard-dev/tw/tree/main/syspeek-tool

cc @maxgio92

joshrwolf (approved):

tw.yaml
dannf: OK, upstream added a new tool and we're now shipping it, but we haven't added a pipeline for it. No reason to block on that. Thanks!

dannf (approved):

PR #45792, opened 2025-03-11
add tests for bazelisk, fix runtime dependency problem

dustinkirkland:
Add a couple of tests for bazelisk, which actually identified a missing runtime dependency on libstdc++.

Why didn't SCA catch this?  Perhaps because the first thing this binary does is reach out to the internet and fetch a newer copy of itself?  Is that what we want?  Not sure...  If we want this to work out of the box in this manner, then we need to add the runtime dependency on libstdc++.  On the other hand, a binary that reaches out to the internet and downloads a different thing to run doesn't really feel like the secure source for open source....

Definitely need a review on this one....

### DIFF
--- a/bazelisk.yaml
+++ b/bazelisk.yaml
@@ -1,13 +1,14 @@
 package:
   name: bazelisk
   version: 1.25.0
-  epoch: 2
+  epoch: 3
   description: A user-friendly launcher for Bazel.
   copyright:
     - license: Apache-2.0
   dependencies:
     runtime:
       - ca-certificates-bundle
+      - libstdc++
 
 environment:
   contents:
@@ -47,3 +48,10 @@ update:
   github:
     identifier: bazelbuild/bazelisk
     strip-prefix: v
+
+test:
+  pipeline:
+    - runs: |
+        bazelisk help
+        bazelisk version
+        bazelisk license >/dev/null


smoser:
To clarify, it doesn't reach out to the internet for a newer version of _itself_.  It reaches out for bazel.

```
> # apk add bazelisk
> fetch https://packages.wolfi.dev/os/x86_64/APKINDEX.tar.gz
> (1/1) Installing bazelisk (1.25.0-r2)
> Executing busybox-1.37.0-r30.trigger
> OK: 20 MiB in 16 packages
> 05dffff3b4be:/# bazelisk 
> 2025/03/11 11:33:42 Downloading https://releases.bazel.build/8.1.1/release/bazel-8.1.1-linux-x86_64...
> Downloading: 58 MB out of 58 MB (100%) 
> /root/.cache/bazelisk/downloads/sha256/a2a095d...dae681def/bin/bazel:
> error while loading shared libraries:
> libstdc++.so.6: cannot open shared object file: No such file or directory
```

bazel _does_ have a dependency on libstdc++.  but in my quick test, even if I had a 8.1.1 bazel in the path bazelisk won't use it.

There is not a useful message on the original commit to bazelisk.  I might suggest the right answer is to drop it until we have a request for it.

smoser:
Feel free to comment on my fix here https://github.com/wolfi-dev/os/pull/45800

OddBloke (approved):

mattmoor:
> Perhaps because the first thing this binary does is reach out to the internet and fetch a newer copy of itself?

This is fundamentally what bazelisk is, it's a wrapper around bazel that does that... 🙃 

What needs this?  Could we make it just use Bazel and kill this package?

dustinkirkland:
Looks like:
 - selenium: https://github.com/wolfi-dev/os/blob/dc0b995296ecdbe7b86bad7d0f0dcf0f8b2005e5/selenium.yaml#L21
 - tensorflow-core: https://github.com/wolfi-dev/os/blob/dc0b995296ecdbe7b86bad7d0f0dcf0f8b2005e5/tensorflow-core.yaml#L29

PR #45800, opened 2025-03-11
Drop bazelisk until there is identified need.

smoser:
As pointed out in https://github.com/wolfi-dev/os/pull/45792, bazelisk has some properties that make it difficult to deliver in a safe way.

The 'About' at https://github.com/bazelbuild/bazelisk says:

> Bazelisk is a wrapper for Bazel written in Go. It automatically picks a
> good version of Bazel given your current working directory, downloads it
> from the official server (if required) and then transparently passes
> through all command-line arguments to the real Bazel binary. You can
> call it just like you would call Bazel.


dustinkirkland (changes requested):
I don't think this is going to fly.  We have two other package builds that depend on bazelisk...  See: selenium and tensorflow...

PR #45833, opened 2025-03-11
Change to using tw/ldd-check and its default packages when available.

smoser:


smoser:
closing this mega-pr. replacing with https://github.com/wolfi-dev/os/labels/smoser%2Ftw-ldd-check

PR #45873, opened 2025-03-11
cleaned up the libb64 test now that we have configure-opts

dustinkirkland:
Cleaner, simpler test now that we have our configure-opts variable in the test pipeline and header-check tool


dannf (approved):
kick ass

PR #45910, opened 2025-03-11
Make dependencies between scala variants consistent

utieyin:
`spark-3.5-scala-2.13` was recently compiled with `-Pkubernetes` to allow deploying spark in a kubernetes cluster. The jars needed for this are available in the package and therefore in the image, the absence of a shell makes it difficult to verify this in a non dev docker image. This change adds busybox as a runtime dependency to the package. To avoid potential issues with the functioning of the different scala variants, we have decided to make the runtime dependencies between `spark-3.5-scala-2.12` and `spark-3.5-scala-2.13` consistent. This is to keep the spark experience consistent across variants.


xnox (changes requested):
Can you please update the PR title & description, to match the contents of the proposed changes?

Since the PR was opened, the scope of changes has increased and no longer matches the previously stated goal.

utieyin:
> Can you please update the PR title & description, to match the contents of the proposed changes?
> 
> Since the PR was opened, the scope of changes has increased and no longer matches the previously stated goal.

@xnox Both have been updated. We will have a longer conversation if those packages are all needed in our ongoing work with spark-fips

mamccorm:
if this change is solely so users can verify scala versions, that wouldn't justify dropping busybox onto the image

mamccorm (approved):

xnox (approved):

PR #46006, opened 2025-03-12
kyverno-policy-reporter-ui: embed static frontend assets at correct path

kranurag7:
`crane export ghcr.io/kyverno/policy-reporter-ui:2.2.0 - | tar -tvf - var/run`


kyverno-policy-reporter-ui.yaml
```diff
@@ -34,25 +38,28 @@ pipeline:
...
   - name: kyverno-policy-reporter-ui-compat
     description: compatibility symlinks package for policyreporter-ui Dockerfile
     pipeline:
       - runs: |
-          # Symlink the binary from usr/bin to /app/
-          mkdir -p ${{targets.subpkgdir}}/app
-          ln -sf /usr/bin/policyreporter-ui ${{targets.subpkgdir}}/app/policyreporter-ui
+          # Symlink the binary from usr/bin to /ko-app/
+          mkdir -p ${{targets.contextdir}}/ko-app
+          ln -sf /usr/bin/policyreporter-ui ${{targets.contextdir}}/ko-app/policyreporter-ui
```
Dentrax, line 63: In tests, can we spin up this on kwok and port-forward to assert that it returns 2xx response code?

Dentrax (changes requested):

kyverno-policy-reporter-ui.yaml
kranurag7: thank you very much for the review, extended tests via https://github.com/wolfi-dev/os/pull/46006/commits/1ea680340bdbe45e2a8130e5427d22831b7ffede

kyverno-policy-reporter-ui.yaml
```diff
@@ -65,7 +72,30 @@ update:
...
+    - name: Run and test ${{package.name}}
+      uses: test/daemon-check-output
+      with:
+        start: /usr/bin/policyreporter-ui run
+        timeout: 30
+        expected_output: |
+          register UI
+          /var/run/ko/ui
+          Server starts
+        # Can't use default error string set because of this log line `[ERROR] failed to load config file: open config.yaml: no such file or directory`.
```
Dentrax, line 93: Let's `curl -sLO https://raw.githubusercontent.com/kyverno/policy-reporter-ui/refs/heads/main/backend/config.example.yaml` this in the `setup:` and try to omit the `error_strings`.

kyverno-policy-reporter-ui.yaml
```diff
@@ -34,25 +38,28 @@ pipeline:
...
   - name: kyverno-policy-reporter-ui-compat
     description: compatibility symlinks package for policyreporter-ui Dockerfile
     pipeline:
       - runs: |
-          # Symlink the binary from usr/bin to /app/
-          mkdir -p ${{targets.subpkgdir}}/app
-          ln -sf /usr/bin/policyreporter-ui ${{targets.subpkgdir}}/app/policyreporter-ui
+          # Symlink the binary from usr/bin to /ko-app/
+          mkdir -p ${{targets.contextdir}}/ko-app
+          ln -sf /usr/bin/policyreporter-ui ${{targets.contextdir}}/ko-app/policyreporter-ui
```
Dentrax, line 62: Can we have a `readlink` test here?
Is this a breaking change?

Maybe we should keep the old link, just in case?

kyverno-policy-reporter-ui.yaml
kranurag7: thank you for the suggestion, I've adopted this approach.

kyverno-policy-reporter-ui.yaml
kranurag7: thank you, I think this was probably an oversight in the first go itself with `/ko-app` being `/app` but keeping it here to make it more backwards compatible. 
this is a breaking change per se but it's not working otherwise so I'll say it's good that we're now more closely inline with the project.

Dentrax (approved):
LGTM! 🚀

PR #46301, opened 2025-03-13
added gem-check to a bunch of packages

dustinkirkland:
Added a test pipeline that uses a new tw function called gem-check.

Also, use it against ~32 gems that can use it successfully currently.

This PR depends on first merging: https://github.com/chainguard-dev/tw/pull/29


ruby3.2-json-jwt.yaml
```diff
@@ -59,3 +59,8 @@ var-transforms:
     match: ^ruby(\d\.\d+)-.*
     replace: $1
     to: rubyMM
+
+test:
+  pipeline:
+    - runs:
+        gem-check
```
smoser, line 66: I don't know how this works.
I think you want 
```yaml
test:
  pipeline:
    - uses: gem-check
```

dustinkirkland: Crap, you're right.  Was working on this way too late last night...

javacruft (approved):
Scanned the build logs to check tests running :+1:

PR #46374, opened 2025-03-13
opensearch-dashboards-2: remove unneeded remediations, fix CVE-2025-27789

OddBloke:


OddBloke:
The CVE reported in ci-cve-scan requires a major version bump of the dependency, so I'd like to handle that separately

powersj (approved):

PR #46475, opened 2025-03-13
Use split/manpages and test/docs with multiple packages

murraybd:


dannf (dismissed):

nodejs-20.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: nodejs-20
-  version: "20.19.0"
-  epoch: 0
+  version: "20.18.3"
```
dannf, line 3: this went back in time

dustinkirkland (dismissed):

nodejs-20.yaml
```diff
@@ -41,7 +41,7 @@ pipeline:
     with:
       repository: https://github.com/nodejs/node.git
       tag: v${{package.version}}
-      expected-commit: bb1a61d8737feff534bb85368dab3b7c554c863d
+      expected-commit: 4819c99baa28bf2c1baf411ba100c467fec3d486
```
javacruft, line 44: I think this is hangover from the back in time versioning and can be dropped.

javacruft (dismissed):

dannf (approved):

PR #46502, opened 2025-03-14
openssl: add ELF notes

xnox:
Add package ELF notes to OpenSSL. This is the last canary affecting
many runtime images (previously git with ELF notes was already added
to all dev tags).

Once this lands, we shall land elf notes by default in openssf
compiler options.

### DIFF
--- a/openssl.yaml
+++ b/openssl.yaml
@@ -2,7 +2,7 @@
 package:
   name: openssl
   version: 3.4.1
-  epoch: 1
+  epoch: 2
   description: "the OpenSSL cryptography suite"
   copyright:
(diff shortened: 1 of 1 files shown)


sil2100 (approved):
The same change was done to the provider package in enterprise. I have confirmed that this properly adds the ELF notes metadata to the openssl binary:

```
Displaying notes found in: .note.package
  Owner                Data size 	Description
  FDO                  0x0000005c	FDO_PACKAGING_METADATA
    Packaging Metadata: {"type":"apk","os":"wolfi","name":"openssl","version":"3.4.1-r2","architecture":"x86_64"}
```

PR #46504, opened 2025-03-14
Ensure 'Chainguard welcome' message is displayed for -compat packages

mamccorm:
We provide images which are compatible with various upstream deployment methods, including helm charts. Some of thse deployment methods may require additional packaging tweaks. For example, a helm chart may expect a binary in a different location, or a specific upstream entrypoint script.

Some of these upstream entrypoint scripts, may print out a welcome message. The changes in this PR introduce a custom 'Chainguard welcome' message, which re-enforce these are Chainguard packages/images, providing compatibility with upstream.

**What's not included in this PR? -** You'll notice I added bash in a few places to runtime dependencies of tests. These runtime deps are not new and defined in images. Ideally we'd remove these from image level and define here, but that will be handed in a separate work item.


clickhouse-25.2.yaml
```diff
@@ -169,6 +169,15 @@ subpackages:
           ln -s /dev/stderr ${{targets.contextdir}}/var/log/clickhouse-server/clickhouse_error.log
           mkdir -p ${{targets.contextdir}}/var/lib
           ln -s /bitnami/clickhouse/data ${{targets.contextdir}}/var/lib/clickhouse
+    test:
+      environment:
+        contents:
+          packages:
+            - ${{package.name}}
+      pipeline:
+        - uses: bitnami/validate-welcome-message
```
mattmoor, line 178: Is this missing an epoch bump?

mattmoor (dismissed):
LGTM, but there are several packages that we need epoch bumps for.

clickhouse-25.2.yaml
mamccorm: Yep! Resolved a few merge conflicts and in the process, reset the epoch bump. I'll get on that, thank-you for spotting and calling out

murraybd:
Why is `bash` added to the test environment for only some packages?

mamccorm:
> Why is `bash` added to the test environment for only some packages?

Hey @murraybd, see note in the PR description field above. Some already have bash, others define at the image level - this is something we should look into in future

mamccorm:
Cannot get a green CI build for this. Errors are different on each attempt. Retrying a couple more times

mamccorm:
Splitting up pipeline and test updates to separate PRs in the aid of getting them through the system. i'm reasonably confident now in the tests

PR #46690, opened 2025-03-14
datadog-agent: do not build with BPF probes in the go embedded FS

maxgio92:
fix(datadog-agent): do not build agent and system-probe with CO-RE BPF probes in embed.FS

The flag --bundle-ebpf for the agent.build and system-probe.build Invoke tasks, makes oom-kill.o to fail because it's expected to be embedded in the Go binary embed.FS at the path 'build', which is not [0].

When the --bundle-ebpf invoke flag is specified the go binaries are built with the build flag ebpf_bindata [1], which makes pkg/ebpf/bytecode.GetReader [2] look for the CO-RE BPF probes in the embed.FS binary-embedded filesystem at the overridden path 'build' [3], instead of one specified by argument on the host FS [4] in case the agent is compiled without the go build flag ebpf_bindata.

The problem comes for probes that are not included in the embedded FS [0], which are copied over build/ at build time by the system-probe.build invoke task [5], ready to be embedded during the go build.

oom-kill.o BPF probe is not included, which makes the loading of it, when the agent's system probe is configured with the oom kill probe enabled [6] and the agent Go binary compiled with the ebpf_bindata build flag, to fail, logging indeed:
```
(/home/build/cmd/system-probe/modules/oom_kill_probe.go:31 in func13) | Starting the OOM Kill probe 
(/home/build/pkg/collector/corechecks/ebpf/probe/oomkill/oom_kill.go:51 in NewProbe) | error loading CO-RE oom-kill probe: error reading oom-kill.o: open build/oom-kill.o: file does not exist. falling back to runtime compiled probe 
(/home/build/pkg/util/kernel/find_headers.go:159 in GetKernelHeaders) | Unable to find kernel headers: no valid matching kernel header directories found. To download kernel headers, set system_probe_config.enable_kernel_header_download to true (/home/build/cmd/system-probe/api/module/loader.go:95 in Register) | error creating module oom_kill_probe: unable to start the OOM kill probe: unable to find kernel headers (/home/build/cmd/system-probe/subcommands/run/command.go:364 in startSystemProbe) | error while starting api server, exiting: failed to create system probe: no module could be loaded 
Error: error while starting api server, exiting: failed to create system probe: no module could be loaded
```

This seems to be a potential bug in upstream, but we don't have a reason why
we should need some BPF probes embedded in the Go binary besides in the package.

Finally, this is made possible thanks to this commit [7] which makes our package to include the probe object files.

0. https://github.com/DataDog/datadog-agent/blob/eab078f2e713852de8e4f4c168ea009065841903/pkg/ebpf/bytecode/asset_reader_bindata.go#L21
1. https://github.com/DataDog/datadog-agent/blob/17329b350f3e31ef8bb7906e842e1989486ec097/tasks/system_probe.py#L830
2. https://github.com/DataDog/datadog-agent/blob/eab078f2e713852de8e4f4c168ea009065841903/pkg/ebpf/co_re.go#L71
3. https://github.com/DataDog/datadog-agent/blob/eab078f2e713852de8e4f4c168ea009065841903/pkg/ebpf/bytecode/asset_reader_bindata.go#L29
4. https://github.com/DataDog/datadog-agent/blob/eab078f2e713852de8e4f4c168ea009065841903/pkg/ebpf/bytecode/asset_reader_nobindata.go#L18
5. https://github.com/DataDog/datadog-agent/blob/eab078f2e713852de8e4f4c168ea009065841903/tasks/system_probe.py#L1554
6. https://github.com/DataDog/helm-charts/blob/5e1d863df9725270aeb52959606e08b713a3cd38/charts/datadog/templates/system-probe-configmap.yaml#L31
7. https://github.com/wolfi-dev/os/pull/43810/commits/79d060f1693adf17d55ba359581da4389b606d28


datadog-agent/datadog-debug-001.diff
```diff
@@ -0,0 +1,98 @@
+diff --git a/pkg/collector/corechecks/ebpf/probe/oomkill/oom_kill.go b/pkg/collector/corechecks/ebpf/probe/oomkill/oom_kill.go
```
maxgio92, line 1: This is not necessary. It's just for debug purposes.

datadog-agent.yaml
```diff
@@ -102,7 +106,7 @@ pipeline:
   # https://github.com/DataDog/datadog-agent/pull/31426
   - uses: patch
     with:
-      patches: revert-import-python-relatively-instead-of-using-the-patch.patch remove-removed-godebug.patch
+      patches: revert-import-python-relatively-instead-of-using-the-patch.patch remove-removed-godebug.patch datadog-debug-001.diff
```
maxgio92, line 109: This is only for debug purposes (see below).

datadog-agent.yaml
```diff
@@ -17,6 +17,10 @@ package:
       - libseccomp
       - shadow
 
+capabilities:
```
maxgio92, line 20: This avoids having to run the build with sudo.

datadog-agent.yaml
```diff
@@ -122,6 +126,7 @@ pipeline:
     with:
       deps: |-
         golang.org/x/oauth2@v0.27.0
+        golang.org/x/net@v0.36.0
```
maxgio92, line 129: Upstream already bumped it to 0.37.0 here:
* https://github.com/DataDog/datadog-agent/commit/3eb1856391bab682859c429dc9509529fd639813#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6

EyeCantCU (approved):

PR #46791, opened 2025-03-15
added a few more ruby tests, these with specific requires

dustinkirkland:
Here are some more successful ruby tests.  All of these include specific libraries to test requiring...


dannf (approved):
LGTM :)

PR #46869, opened 2025-03-15
step-ca: mitigate CVEs, migrate to go/build and test/daemon-check-output to prevent …

Dentrax:
Migrate to `go/build` and `test/daemon-check-output` to prevent CI flakes in tests. Mitigates:

```
└── 📄 /usr/bin/step-ca
        📦 golang.org/x/net v0.35.0 (go-module)
            Unknown CVE-2025-22870 fixed in 0.36.0, 0.36.0
            Unknown CVE-2025-22870 fixed in 0.36.0
            Medium CVE-2025-22870 GHSA-qxp5-gwg8-xv66 fixed in 0.36.0
```

### DIFF
--- a/step-ca.yaml
+++ b/step-ca.yaml
@@ -10,7 +10,6 @@ environment:
   contents:
     packages:
       - pcsc-lite-dev
-      - pkgconf-dev
       - step
 
 pipeline:
@@ -36,7 +35,7 @@ pipeline:
       output: step-ca
       ldflags: |
         -X main.Version=${{package.version}}
-        -X main.BuildTime=$(date -u '+%Y-%m-%d %H:%M UTC')
+        -X main.BuildTime=$(date -u +"%Y-%m-%dT%H:%M:%SZ")
 
 update:
   enabled: true
--- a/step-ca.yaml
+++ b/step-ca.yaml
@@ -1,7 +1,7 @@
 package:
   name: step-ca
   version: "0.28.2"
-  epoch: 4
+  epoch: 5
   description: A private certificate authority (X.509 & SSH) & ACME server for secure automated certificate management, so you can use TLS everywhere & SSO for SSH.
   copyright:
     - license: Apache-2.0
@@ -9,9 +9,9 @@ package:
 environment:
   contents:
     packages:
-      - busybox
-      - ca-certificates-bundle
-      - go
+      - pcsc-lite-dev
+      - pkgconf-dev
+      - step
 
 pipeline:
   - uses: git-checkout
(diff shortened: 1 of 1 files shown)


step-ca.yaml
```diff
@@ -27,13 +27,15 @@ pipeline:
...
-        -o "${{targets.destdir}}/usr/bin/step-ca" ./cmd/step-ca
-
-  - uses: strip
+  - uses: go/build
+    with:
+      modroot: .
+      packages: ./cmd/step-ca
+      output: step-ca
+      ldflags: |
+        -X main.Version=${{package.version}}
```
kranurag7, line 38: we can consider including buildtime in here as well. ref: https://github.com/smallstep/certificates/blob/6ce27d3fdbfe2eacc25e48a52267a50579bb93b3/.goreleaser.yml#L42 💭

Dentrax: Makes sense, done!

step-ca.yaml
```diff
@@ -27,13 +27,16 @@ pipeline:
...
+        -X main.BuildTime=$(date +%F-%T)
```
kranurag7, line 39: Is this same as `date -u '+%Y-%m-%d %H:%M UTC'` 💭 ref: https://github.com/smallstep/certificates/blob/6ce27d3fdbfe2eacc25e48a52267a50579bb93b3/Makefile#L62

step-ca.yaml
```diff
@@ -27,13 +27,16 @@ pipeline:
...
-
-  - uses: strip
+  - uses: go/build
+    with:
+      modroot: .
+      packages: ./cmd/step-ca
+      output: step-ca
+      ldflags: |
+        -X main.Version=${{package.version}}
+        -X main.BuildTime=$(date -u '+%Y-%m-%d %H:%M UTC')
```
kranurag7, line 39:
```suggestion
        -X main.BuildTime=$(TZ=UTC date -u +"%Y-%m-%dT%H:%M:%SZ")
```
UTC and matches what upstream is using.

Dentrax: `-u` already forces time zone to UTC so setting `TZ=UTC` would be redundant in this case.

kranurag7: Yes that is true :)

step-ca.yaml
```diff
@@ -1,17 +1,17 @@
...
 
 environment:
   contents:
     packages:
-      - busybox
-      - ca-certificates-bundle
-      - go
+      - pcsc-lite-dev
+      - pkgconf-dev
+      - step
```
kranurag7, line 14: we can consider omitting all this, go toolchain coming via go/build is enough here (build and tests are passing locally). Is there a reason to include these build time deps here? 💭

Dentrax: Weird, build was failing on me without having those deps, also step-ca requires step package as a build time dep.

Dentrax: I just omit `pkgconf-dev`, thanks!

kranurag7 (approved):
thank you!!

PR #47069, opened 2025-03-17
python: rebuild with gcc without SAHF instructions

xnox:
Rebuild python against gcc without sahf instructions.

### DIFF
--- a/python-3.13.yaml
+++ b/python-3.13.yaml
@@ -1,7 +1,7 @@
 package:
   name: python-3.13
   version: "3.13.2"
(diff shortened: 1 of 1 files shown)


justinvreeland:
Seems to be working

```
python-3.13-3.13.2-r5 x86_64 {python-3.13} (PSF-2.0) [installed]
python-3.13-base-3.13.2-r5 x86_64 {python-3.13} (PSF-2.0) [installed]
/ # python3 -c  'print((-0.01)**2);'
0.0001
/ #
```

PR #47173, opened 2025-03-17
tw/ldd-check cleanup batch 15

smoser:


p11-kit.yaml
```diff
@@ -63,14 +63,10 @@ subpackages:
...
-        - uses: test/ldd-check
-          with:
-            packages: $(basename ${{targets.contextdir}})
+        - uses: test/tw/ldd-check
         - runs: |
             trust --help
-        - uses: test/ldd-check
-          with:
-            packages: p11-kit-trust
+        - uses: test/tw/ldd-check
```
murraybd, line 69: This is the second usage of the ldd-check test for this package.

p11-kit.yaml
```diff
@@ -85,12 +81,8 @@ subpackages:
...
     test:
       pipeline:
-        - uses: test/ldd-check
-          with:
-            packages: $(basename ${{targets.contextdir}})
-        - uses: test/ldd-check
-          with:
-            packages: p11-kit-server
+        - uses: test/tw/ldd-check
+        - uses: test/tw/ldd-check
```
murraybd, line 86: Once should be enough right? ;-)

smoser:
tests passed at https://github.com/wolfi-dev/os/pull/47173/checks?check_run_id=38920549306

p11-kit.yaml
smoser: I didn't add the second one. but I've removed the dupes here now.

p11-kit.yaml
smoser: These were already there. I've removed the dupe.

murraybd (approved):

PR #47273, opened 2025-03-18
fix(libcxx*): Drop provides for older LLVM

javacruft:
The libcxx headers and libraries provided by LLVM 19 are not compatible with LLVM < 17:

```
/usr/bin/../include/c++/v1/__configuration/compiler.h:37:8: error: "Libc++ only supports Clang 17 and later" [-Werror,-W#warnings]
       warning "Libc++ only supports Clang 17 and later"
```

Drop provides for older LLVM releases which provide their own compatible versions of these packages.


xnox:
For this to take full effect, one would also need to withdraw all past full-version of APKs that did provide those provides. As both apko and apk-tools will continue to helpfully use -r0 and any older providers.

EyeCantCU (approved):

PR #47308, opened 2025-03-18
libarchive: backport CVE fixes

OddBloke:
These patches are self-contained and apply cleanly, so instead of waiting for upstream to cut a release (which hasn't happened since October), apply the fixes directly ourselves.


OddBloke:
We'll need to manually flag the advisories as fixed.

powersj (approved):

PR #47324, opened 2025-03-18
apk-tools: disable cache by default

xnox:
apk-tools fails to work correctly with missing cache, corrupted cache,
incompete cache, or on filesystems mounted without incorrect
timestamps. As cache appears to be based on lstat timestamps, rather
than on etags.

Also, apkindex cache in containers use up disk space and create
non-minimised layers. Thus in general for containers operating by
default without cache, is preffered mode of operation.

With this change, apkindex will be downloaded upon each invocation apk
update. Meaning to change in performance for single apk invocation,
but may cause additional apkindex downloads (potential of additional
10MB network traffic, from second invocation onwards), but on the flip
side the resulting layer will be minimised. Slightly increased
dockerfile build time, with a guranteed smaller end image.

### DIFF
--- a/apk-tools.yaml
+++ b/apk-tools.yaml
@@ -1,7 +1,7 @@
 package:
   name: apk-tools
   version: "2.14.10"
-  epoch: 1
+  epoch: 2
   description: "apk-tools (Wolfi package manager)"
   copyright:
     - license: GPL-2.0-only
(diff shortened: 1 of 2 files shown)


xnox:
fail.

xnox:
this does not work for apk update as expected

PR #47063, opened 2025-03-17
gcc: block SAHF instructions by default for compatability with Apple Rosetta 2

xnox:
Apple Rosetta 2 implements many accelerated vector instructions, but
not SAHF which affects a small minority of binaries. Block emitting
SAHF instructions to maintain compatibility with Apple Rosetta 2.

Similar changes may need to be applied to all older gcc; as well as to the clang toolchains if possible (or added to the openssf-compiler-options).

PR #47344, opened 2025-03-18
Rebuild nodejs with -march=x86-64-v2 -mno-sahf

xnox:
Merging a previous PR, largely did nothing[1], as epoch bumps were
superseeded by manpage split prs, and github let this be merged
without a conflict, but also not actually building anything.

Rebuild all nodejs, drop any -march/-mno-sahf settings, which will use
gcc-14 defaults, which are now set to be -march=x86-64-v2 -mno-sahf.

[1] See https://github.com/wolfi-dev/os/commit/c38130cc3f7869652376a837e96946e36163a99f

### DIFF
--- a/nodejs-16.yaml
+++ b/nodejs-16.yaml
@@ -1,7 +1,7 @@
 package:
   name: nodejs-16
   version: 16.20.2
-  epoch: 10
+  epoch: 11
   description: "JavaScript runtime built on V8 engine"
   dependencies:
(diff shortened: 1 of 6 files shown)


murraybd:
The gcc-14 changes can be found in  https://github.com/wolfi-dev/os/pull/47063.

murraybd (approved):

PR #47361, opened 2025-03-18
apk-tools: cherry-pick MR 292 to fix apk cache update on Rosetta 2

xnox:
Cherry-pick merge request from @justinvreeland to fix apk tools
failing to correctly use or update cache files.

For more details see:
- https://gitlab.alpinelinux.org/alpine/apk-tools/-/merge_requests/292

### DIFF
--- a/apk-tools.yaml
+++ b/apk-tools.yaml
@@ -1,7 +1,7 @@
 package:
   name: apk-tools
   version: "2.14.10"
-  epoch: 1
+  epoch: 2
   description: "apk-tools (Wolfi package manager)"
   copyright:
     - license: GPL-2.0-only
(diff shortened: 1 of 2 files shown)


justinvreeland:
Ahhh right as I gave up on trying to get the `cherry-pick` option to work.

apk-tools.yaml
```diff
@@ -31,6 +31,9 @@ pipeline:
       tag: v${{package.version}}
       expected-commit: 9d074efdc12bc41b5d24190595a5269a770e852a
 
+  - runs: |
```
justinvreeland, line 34: ahh I guess this keeps the message instead of just patch applying it?

justinvreeland (approved):
I approve even though I wrote it so we might want someone else but it's not like GitHub Is here to enforce this one!

PR #47377, opened 2025-03-18
Call corepack enable fixing a FTBFS

murraybd:
Adding a separate call to `corepack enable` resolves a hang when building the teleport package. From what I saw during the build process one of the later commands either `pnpm config set...` or `make ensure-js-deps` ended up unsuccessfully calling `corepack enable` and for some reason putting in a separate step works. I'm not familiar with these languages so this could be wrong.

This fixes https://github.com/chainguard-dev/internal-dev/issues/10583

### DIFF
--- a/teleport.yaml
+++ b/teleport.yaml
@@ -1,7 +1,7 @@
 package:
   name: teleport
   version: "17.3.2"
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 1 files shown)


murraybd:
This was superseded by https://github.com/wolfi-dev/os/pull/44922

PR #47459, opened 2025-03-19
ruby-yajl-ruby git deep clone so commit checkout works.

smoser:
These were all failing to build from source because the shallow clone that git does by default could not find the provided hash.

hopefully developer will create a tag as requested
 https://github.com/brianmario/yajl-ruby/issues/216


roaksoax (approved):
lgtm!

PR #47460, opened 2025-03-19
Flannel package enhancements misconfigured version setting + tests  

rahulduvedi:
`flannel-0.26.5 `: test expansion and version bug correction
Fixes: 
The version for the package was coming as `dev` due to incorrect setup of `-ldflags` for version.


flannel.yaml
```diff
@@ -44,8 +44,42 @@ update:
...
+        - etcd
   pipeline:
-    # AUTOGENERATED
     - runs: |
+        # Check flanneld version
         flanneld --version
-        flanneld --help
+
+        # Start etcd (default v3)
+        mkdir -p /var/lib/etcd
```
kranurag7, line 58: We can consider using test/daemon-check-output pipeline here.

rahulduvedi: Thanks for suggesting, I have updated the code to use test/demon-check-output pipeline.

wojciechka (dismissed):

maxgio92:
This is awesome

flannel.yaml
```diff
@@ -44,8 +44,62 @@ update:
...
+          done
+
+          # Store flannel network config in etcd
+          printf "{\"Network\": \"10.244.0.0/16\", \"Backend\": {\"Type\": \"host-gw\"}}" > /tmp/flannel-config.json
+          etcdctl --endpoints=http://127.0.0.1:2379 put /coreos.com/network/config "$(cat /tmp/flannel-config.json)"
+
+          # Save PID for cleanup
+          echo "export ETCD_PID=$ETCD_PID" > /tmp/etcd_pid.sh
+        # Disable iptables to avoid permission issues in unprivileged test environments and focus testing on core networking capabilities
+        start: flanneld --etcd-endpoints=http://127.0.0.1:2379 --iface=eth0 --iptables-forward-rules=false
```
maxgio92, line 80:
```suggestion
        start: |
          sleep 3
          IFACE=$(ip -json route show default | jq -r '.[0].dev')
          echo "export IFACE=$IFACE" >> /tmp/env.sh
          flanneld --etcd-endpoints=http://127.0.0.1:2379 --iface=$IFACE --iptables-forward-rules=false
```

flannel.yaml
```diff
@@ -44,8 +44,62 @@ update:
...
+            fi
+            sleep 1
+          done
+
+          # Store flannel network config in etcd
+          printf "{\"Network\": \"10.244.0.0/16\", \"Backend\": {\"Type\": \"host-gw\"}}" > /tmp/flannel-config.json
+          etcdctl --endpoints=http://127.0.0.1:2379 put /coreos.com/network/config "$(cat /tmp/flannel-config.json)"
+
+          # Save PID for cleanup
+          echo "export ETCD_PID=$ETCD_PID" > /tmp/etcd_pid.sh
```
maxgio92, line 78:
```suggestion
          echo "export ETCD_PID=$ETCD_PID" > /tmp/env.sh
```

flannel.yaml
```diff
@@ -44,8 +44,62 @@ update:
...
+          # Load etcd PID from setup phase
+          . /tmp/etcd_pid.sh
+
+          # Verify flannel created the proper subnet configuration
+          if [ ! -f /run/flannel/subnet.env ]; then
+            echo "Subnet file not found"
+            exit 1
+          fi
+          grep -q "FLANNEL_SUBNET=" /run/flannel/subnet.env
+
```
maxgio92, line 103:
```suggestion
+          # Check leases.
+          LEASE_SUBNET_ETCD_KEY=$(etcdctl get /coreos.com/network/subnets --prefix --keys-only)
+          LEASE_PUBLIC_IP=$(etcdctl get $LEASE_SUBNET_ETCD_KEY --print-value-only=true | jq '.PublicIP' | tr -d "\"")
+
+          LEASE_ETCD_LEASE_ID=$(etcdctl lease list | tail -1)
+          etcdctl lease timetolive --keys $LEASE_ETCD_LEASE_ID | grep $LEASE_SUBNET_ETCD_KEY
+          ip address show eth0 | grep $LEASE_PUBLIC_IP
+          ip route get $LEASE_PUBLIC_IP
```

This is an amazing test! Than ks @rahulduvedi!
I'd just add a small check on the IP leases it creates on etcd as etcd leases as well.

flannel.yaml
```diff
@@ -44,8 +44,62 @@ update:
     strip-prefix: v
 
 test:
+  environment:
+    contents:
+      packages:
+        - ${{package.name}}
+        - etcd
```
maxgio92, line 51:
```suggestion
        - etcd
        - jq
        - iproute2
```

flannel.yaml
```diff
@@ -28,14 +28,18 @@ pipeline:
...
+      ldflags: "-extldflags '-static' -X github.com/flannel-io/flannel/pkg/version.Version=${{package.version}}"
+
   - runs: |
-      CGO_ENABLED=0
       if [ "$(go env GOARCH)" = "amd64" ]; then
-        CGO_ENABLED=1
+        export CGO_ENABLED=1
+      else
+        export CGO_ENABLED=0
       fi
```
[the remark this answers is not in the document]
rahulduvedi: I think we cannot set the architecture specific env specs in the go/build pipeline, so we have to skip using the go/build pipeline in this one.

flannel.yaml
```diff
@@ -28,14 +30,13 @@ pipeline:
         golang.org/x/crypto@v0.35.0
         golang.org/x/net@v0.36.0
 
-  # Build with CGO_ENABLED=1 to include the UDP backend
   - runs: |
-      CGO_ENABLED=0
+      export CGO_ENABLED=0
```
maxgio92, line 34: Why we're disabling CGO here?

flannel.yaml
```diff
@@ -28,14 +30,13 @@ pipeline:
...
       if [ "$(go env GOARCH)" = "amd64" ]; then
-        CGO_ENABLED=1
+        export CGO_ENABLED=1
       fi
-      CGO_ENABLED=$(CGO_ENABLED) go build -o "${{targets.destdir}}"/usr/bin/flanneld \
-        -ldflags '-extldflags "-static" -X github.com/flannel-io/flannel/pkg/version.version=${{package.version}}'
+      go build -o "${{targets.destdir}}"/usr/bin/flanneld \
```
maxgio92, line 38: Can we use the [go/build](https://github.com/chainguard-dev/melange/tree/main/pkg/build/pipelines/go#gobuild) pipeline here?

flannel.yaml
```diff
@@ -44,8 +45,85 @@ update:
...
+    contents:
+      packages:
+        - ${{package.name}}
+        - etcd
+        - jq
+        - iproute2
   pipeline:
-    # AUTOGENERATED
-    - runs: |
+    - name: "Custom test for flanneld"
```
maxgio92, line 56: Can we split some tasks into individual steps?

flannel.yaml
rahulduvedi: We cannot use the `go/build` because there is no way to pass environments on the basis of architecture in our current melange build system

flannel.yaml
rahulduvedi: We're disabling CGO (CGO_ENABLED=0) in some cases because Flannel's UDP backend explicitly doesn't support architectures other than amd64, and enabling CGO there would result in build-time or runtime errors ([source](https://github.com/flannel-io/flannel/blob/master/backend/udp/udp.go#L65-L67)).

flannel.yaml
rahulduvedi: Sure, Working on it.

flannel.yaml
rahulduvedi: The reason we're disabling CGO for non-AMD64 architectures is to maintain compatibility with upstream's architecture-specific implementations. Looking at the Flannel Makefile, you can see they explicitly distinguish CGO settings by architecture:
https://github.com/flannel-io/flannel/blob/master/Makefile#L12-L15
```
ifeq ($(ARCH),amd64)
    CGO_ENABLED=1
else
    CGO_ENABLED=0
endif
```
This is intentional - the UDP backend has architecture-specific code that's designed to work with CGO enabled on AMD64 and disabled elsewhere. Following this pattern ensures we maintain compatibility with upstream's design choices and avoid undefined symbols in the UDP networking code.
Adding to this here
For our FIPS `[package](https://github.com/chainguard-dev/enterprise-packages/pull/19481/)`, we make a different trade-off (enabling CGO everywhere for FIPS compliance), but for the standard package, we should follow upstream's architecture-specific approach to ensure full functionality across all backends.

maxgio92 (approved):
LGTM

PR #47463, opened 2025-03-19
Remove duplicate ldd-check tests

murraybd:
A couple of packages had the `test/tw/ldd-check` test in them multiple times. This removes the duplicate tests.


smoser (approved):
thanks.

PR #47517, opened 2025-03-20
New Python packages for CherryPy + Dependencies

javacruft:
New packages for CherryPy with supporting dependency chain including:

- tempora
- portend
- autocommand
- path
- jaraco.text
- jaraco.collections
- cheroot

This supports packaging work for Ceph (which uses CherryPy for the Ceph Manager Dashboard Module).

Related: #43573

#### For new package PRs only

#### For PRs that add patches

Patch cherry picked for cherrypy to switch from an unmaintained lockfile dependency to filelock.


py3-cherrypy.yaml
```diff
@@ -0,0 +1,138 @@
...
+        - uses: python/import
+          with:
+            python: python${{range.key}}
+            import: ${{vars.pypi-package}}
+        - uses: git-checkout
+          with:
+            repository: https://github.com/cherrypy/cherrypy
+            tag: v${{package.version}}
+            expected-commit: d1eb64e27ac2dbb761b3cf96eb0c883d5bd87786
+        - uses: patch
```
sil2100, line 79: Any reason why you created a hand-made patch instead of using `git-checkout`'s 'cherry-picks:`? It's a cleaner alternative, if the patch is part of the upstream repo. Might be the case here, right?

sil2100 (dismissed):
One inline comment, but not necessary to fix - more like an improvement. Other than that all looking good. Thanks!

py3-cherrypy.yaml
javacruft: It is - I discovered this feature this morning! I'll re-do with cherry-picks instead plus the pytest-xdist package should be available shortly to reduce test durations

sil2100 (approved):

PR #47533, opened 2025-03-20
cifs-utils - cleanup empty sbin/ dir

smoser:
cifs-utils is failing the usrmerge lint test because it contains an empty sbin/ directory.

So get rid of that.


smoser:
this failure is blocking https://github.com/wolfi-dev/os/pull/47193

zmarano (approved):

PR #47576, opened 2025-03-20
jq: remediate CVE-2024-53427

xnox:
Cherrypick patches from master branch, to remediate CVE-2024-53427.

Also add testcase that checks that CVE is patched, as it changes
behaviour: NaN1 fromjson used to be 'true' and now is invalid.

CVE-2024-53427

Also see:
- https://github.com/jqlang/jq/issues/3196
- https://github.com/advisories/GHSA-8mxc-vqrq-gcm8
- https://github.com/jqlang/jq/pull/3245
- https://github.com/jqlang/jq/commit/b86ff49f46a4a37e5a8e75a140cb5fd6e1331384
- https://github.com/jqlang/jq/commit/a09a4dfd55e6c24d04b35062ccfe4509748b1dd3

### DIFF
--- a/jq.yaml
+++ b/jq.yaml
@@ -1,7 +1,7 @@
 package:
   name: jq
   version: 1.7.1
-  epoch: 3
+  epoch: 4
   description: "a lightweight and flexible JSON processor"
   copyright:
     - license: MIT
@@ -23,6 +23,12 @@ pipeline:
       repository: https://github.com/jqlang/jq.git
       tag: jq-${{package.version}}
       expected-commit: 71c2ab509a8628dbbad4bc7b3f98a64aa90d3297
+
+  - uses: patch
+    with:
+      patches: |
+        0001-fix-jv_number_value-should-cache-the-double-value-of.patch
+        0002-Reject-NaN-with-payload-while-parsing-JSON.patch
 
   - runs: |
       autoreconf -vfi
@@ -79,3 +85,7 @@ test:
       runs: |
         echo '[{"id":1,"name":"John"},{"id":2,"name":"Jane"}]' | jq '{users: map({userId: .id, userName: .name})}' | grep -E '"userId": 1|"userId": 2' || exit 1
     - uses: test/tw/ldd-check
+    - name: CVE-2024-53427
+      runs: |
+        output=$(echo '["NaN1"]' | jq -r '.[] | try (fromjson | isnan) catch .')
+        [ "$output" = "Invalid numeric literal at EOF at line 1, column 4 (while parsing 'NaN1')" ]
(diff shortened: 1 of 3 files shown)


IdlePhysicist (dismissed):

murraybd (approved):

jq.yaml
```diff
@@ -79,3 +85,7 @@ test:
       runs: |
         echo '[{"id":1,"name":"John"},{"id":2,"name":"Jane"}]' | jq '{users: map({userId: .id, userName: .name})}' | grep -E '"userId": 1|"userId": 2' || exit 1
     - uses: test/tw/ldd-check
+    - name: CVE-2024-53427
+      runs: |
+        output=$(echo '["NaN1"]' | jq -r '.[] | try (fromjson | isnan) catch .')
+        [ "$output" = "Invalid numeric literal at EOF at line 1, column 4 (while parsing 'NaN1')" ]
```
stevebeattie, line 91: It might be useful to have a positive test here for the "Nan" case, just to make sure we haven't broken NaN parsing completely with the cherry-pick.

jq/0002-Reject-NaN-with-payload-while-parsing-JSON.patch
```diff
@@ -0,0 +1,33 @@
+From 7a7cea283237cd24557c0b39954ebcd260802f34 Mon Sep 17 00:00:00 2001
+From: itchyny <itchyny@cybozu.co.jp>
+Date: Wed, 5 Mar 2025 07:43:54 +0900
+Subject: [PATCH 2/2] Reject NaN with payload while parsing JSON
+
+This commit drops support for parsing NaN with payload in JSON like
+`NaN123` and fixes CVE-2024-53427. Other JSON extensions like `NaN` and
+`Infinity` are still supported. Fixes #3023, fixes #3196, fixes #3246.
+
+(cherry picked from commit a09a4dfd55e6c24d04b35062ccfe4509748b1dd3)
```
stevebeattie, line 10: Would be useful to mention the dropped parts of the patch (the added tests) and why (because they don't apply cleanly) in the patch header here.

A nice feature to add to melange would be a filterdiff like capacity to the cherry-picks mechanism, that let you filter out a conflicting part of a commit (like an entry to a changelog, or, in this case, testcases that don't apply and that we wouldn't run :disappointed: ) ; basically a way to express the equivalent of:

```
git show a09a4dfd55e6c24d04b35062ccfe4509748b1dd3  | filterdiff -x '?/tests/*'
```

But maybe that lets one tweak a cherry-picked commit too much.

xnox: yes, when we implemented the cherry-picks arg to the git-checkout pipeline, the very first thing we did was like "oh, but what i really want is filterdiff". I do wonder if we could do something like git format-patch; followed by filterdiff; followed by git am with extra comments / headers added on things that are filtered out. but yeah, that's a wishlist in the melange project.

jq.yaml
xnox: agreed, let me add that too in a follow up.

xnox: https://github.com/wolfi-dev/os/pull/47589

jq/0002-Reject-NaN-with-payload-while-parsing-JSON.patch
smoser: @stevebeattie wrt filterdiff - https://github.com/chainguard-dev/melange/issues/1422

PR #47664, opened 2025-03-21
openjdk-24: enable linkable runtime modules

xnox:
This new jdk feature should enable jlink feature, without need to have
a separate copy of jmods. This will reduce jdk:24-dev tag size
significantly.

See https://openjdk.org/jeps/493


xnox:
Merged upon 3rd attempt:

PR #47665, opened 2025-03-21
Update Python version to 3.11 and add early stopping simulation script

utieyin:
Adding python dependencies to some components of kubeflow-katib and a quick test to earlystopping

Fixes:
Python dependencies for kubeflow-katib
Related:
https://github.com/orgs/chainguard-dev/projects/45/views/62?pane=issue&itemId=102755766&issue=chainguard-dev%7Ccustomer-issues%7C2132
kubeflow-katib has several components that depends on python. That was previously missing from the packages. This PR adds the python package to some of these components


kubeflow-katib.yaml
```diff
@@ -6,6 +6,9 @@ package:
   copyright:
```
mritunjaysharma394, line 6: I think we should bump the epoch?

utieyin: Thanks Mri, epoch bumped

kubeflow-katib.yaml
```diff
@@ -95,13 +99,18 @@ subpackages:
         - uses: test/ldd-check
           with:
             packages: ${{subpkg.name}}
+        - uses: test/daemon-check-output
+          with:
+            start: python earlystopping.py --max_epochs 20 --patience 2 > output.log
```
jamonation, line 104: This output.log doesn't look like it is used anywhere?

kubeflow-katib/earlystopping.py
jamonation, line 1: This runs in a plain python interpreter with no kubeflow-katib library installed

utieyin: Yea, this was to test regular python as that was the reported issue. I will remove this test

jamonation: Hmm I wouldn't just not test things - is there an import or some module withing kubeflow-katib that can parse the output you generated?

kubeflow-katib.yaml
```diff
@@ -1,11 +1,14 @@
...
   name: kubeflow-katib
-  epoch: 13
+  epoch: 14
   version: 0.17.0
   description: Kubeflow Katib services
   copyright:
     - license: Apache-2.0
 
+vars:
+  py-version: 3.11
```
EyeCantCU, line 10: Solution looks good to me. Any reason we've moved to 3.11 here though?

utieyin: I looked at the upstream and they had python 11 and thought it might be appropriate to do so for our packages

EyeCantCU (approved):

PR #47670, opened 2025-03-21
shadow: set bindir to /usr/bin

OddBloke:
This avoids having to explicitly clean up the empty `/bin` directory (which we do to avoid tripping up the usrmerge linter).

Also add smoke tests for the package I'm changing, to make sure I got it right.


roaksoax (approved):
lgtm!

PR #47673, opened 2025-03-21
feat(spark): Refactor

EyeCantCU:
- Remove PySpark Scala 2.13 subpackage and instead fix the symlink to PySpark in our compat subpackage. Add temp virtual provides for PySpark to Spark w/ Scala 2.13
- Only build with OpenJDK 8 for serial compat with Hadoop and gRPC. Add temp virtual provides until we move over
- Drop several unnecessary patches
- Test TLS


spark-3.5.yaml
```diff
@@ -20,370 +33,240 @@ environment:
...
-                cat <<EOF > SQLTest.scala
-                import org.apache.spark.sql.SparkSession
-                val spark = SparkSession.builder.appName("SQLTest").getOrCreate()
-                import spark.implicits._
-                val df = Seq((1, "Alice"), (2, "Bob")).toDF("id", "name")
-                df.createOrReplaceTempView("people")
-                val result = spark.sql("SELECT name FROM people WHERE id = 2")
-                assert(result.count() == 1 && result.first().getString(0) == "Bob")
-                EOF
-                cat SQLTest.scala | /usr/lib/spark/bin/spark-shell --conf spark.jars.ivy=/tmp/.ivy --master local[*]
```
jamonation, line 340: I don't see a scala test anymore, is that intentional? The SQL, R, and Python tests look good

EyeCantCU: Intentional, the Spark shell itself is already leveraging Scala and we can only have one running at a given time. Unless we want to track the PID, kill the running shell, and re-instantiate it, but I don't believe that's as good of a test as the shell already actively running every app we submit

jamonation (approved):

PR #47689, opened 2025-03-21
etcd: Expand compat package tests. Define runtime dependencies.

mamccorm:


etcd-3.5.yaml
```diff
@@ -41,6 +41,9 @@ pipeline:
 subpackages:
   - name: etcd-3.5-bitnami-compat
     description: "compat package with bitnami/etcd image"
+    dependencies:
+      runtime:
+        - bash
```
mamccorm, line 46: Was defined as a runtime dep at the image level - moving to package level, aligning with other similar examples.

jamonation:
Looks like a conflict to resolve for the package version @mamccorm - then +1 from me

etcd-3.5.yaml
jamonation: Thanks for moving these over

jamonation (approved):

PR #47692, opened 2025-03-21
Logstash: Expand compat package tests

mamccorm:


logstash-8.yaml
```diff
@@ -171,6 +171,10 @@ subpackages:
   - name: ${{package.name}}-compat
     description: Compatibility with the upstream image
     dependencies:
+      runtime:
```
mamccorm, line 174: These were defined as runtime dependencies at the image level. Moving to package level runtimes for consistency.

mamccorm:
Similar log change already added

PR #47693, opened 2025-03-21
minio-client: Expand compat package tests

mamccorm:


mc.yaml
```diff
@@ -73,10 +73,22 @@ subpackages:
           # Restore path
           find ${{targets.subpkgdir}}/opt/bitnami -type f -exec sed 's#${{targets.subpkgdir}}##g' -i {} \;
     test:
+      environment:
+        contents:
+          packages:
```
mamccorm, line 78: These where previously defined at the image level as runtime dependencies - moving to package level for consistency with other similar packages.

jamonation (approved):

PR #47705, opened 2025-03-21
py3-*: use versioned numpy runtime dependency where appropriate

OddBloke:
These packages currently resolve their numpy dependency to the correct version because another of their dependencies correctly depends on the versioned numpy package, meaning the resolver selects a numpy-providing package which satisfies both those constraints.

Without those other transitive constraints (which could potentially be removed in future), installing these packages for any Python version will incorrectly pull in numpy for the latest Python version.

### DIFF
--- a/py3-cmaes.yaml
+++ b/py3-cmaes.yaml
@@ -1,7 +1,7 @@
 package:
   name: py3-cmaes
   version: 0.11.1
-  epoch: 3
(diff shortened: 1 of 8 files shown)


philroche (approved):

smoser:
Very nice, thank you, @OddBloke

PR #47706, opened 2025-03-21
chore(openjdk): Strip the JDK and JRE

EyeCantCU:
Previously we hadn't been using our strip pipeline, resulting in much larger installations of the JRE and JDK


EyeCantCU:
Drafting until next week as it's a Friday, but reviews are appreciated :)

xnox (approved):

PR #48077, opened 2025-03-24
aws-network-policy-agent v1.2.0 ftbfs

murraybd:
This fixes a failure to build with the new version of aws-network-policy-agent which was discovered in https://github.com/wolfi-dev/os/pull/43390.


murraybd:
I think I should add some headers to the patches though...

sil2100 (changes requested):
Looks good, but I'd like to understand a bit more about where the patches come from and what is their rationale - either in the patches themselves or a comment on the patch pipeline. Thanks!

murraybd:
> Looks good, but I'd like to understand a bit more about where the patches come from and what is their rationale - either in the patches themselves or a comment on the patch pipeline. Thanks!

I've done a much better job of documenting things in a PR which supersedes this https://github.com/wolfi-dev/os/pull/48199

PR #48092, opened 2025-03-25
add docker runner specific make targets

kranurag7:
some packages both during build and test times behave different when we use docker runner and when we use bubblewrap runner.

We use bubblewrap on ws and mac and docker in CI. Adding these make targets so that it's easier to have access to docker specific runners.

less typing and I continue to be lazy :)


wojciechka (approved):

maxgio92 (approved):

murraybd:
Could you announce this new targets somewhere?

murraybd (approved):

PR #48110, opened 2025-03-25
Fix ingress-nginx-controller broken build and CVEs

hectorj2f:
This is related to this list of CVEs: https://thehackernews.com/2025/03/critical-ingress-nginx-controller.html

### DIFF
--- a/ingress-nginx-controller-1.12.yaml
+++ b/ingress-nginx-controller-1.12.yaml
@@ -169,8 +169,8 @@ vars:
   # Instrumentation for nginx plugin: https://github.com/open-telemetry/opentelemetry-cpp-contrib
   # NOTE: this is really really critical and the config can break in backwards compatible way!!
   # We should always use the SHA available in release-x.y branch of upstream project (e.g. https://github.com/kubernetes/ingress-nginx/blob/release-1.12/images/nginx/rootfs/build.sh#L532)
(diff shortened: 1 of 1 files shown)


grpc-1.70.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: grpc-1.70
   version: "1.70.1"
-  epoch: 2
+  epoch: 3
```
hectorj2f, line 4: We need to bump this epoch to build grpc using the latest protobuf version otherwise there are problems resolving the libprotobuf deps.

kbsteere (approved):

ingress-nginx-controller-1.12.yaml
```diff
@@ -1,7 +1,7 @@
 #nolint:valid-pipeline-fetch-digest,valid-pipeline-git-checkout-commit,valid-pipeline-git-checkout-tag
 package:
   name: ingress-nginx-controller-1.12
-  version: 1.12.0
+  version: 1.12.1
   # There are manual changes to review between each package update. See 'vars:' section
   epoch: 14
```
smoser, line 6: didn't zero the epoch. oh well.

hectorj2f: Good catch!

hectorj2f: We might need to get the epoch 0, and withdraw 14, 🤦🏻

PR #48116, opened 2025-03-25
feat(external-dns): add bitnami-compat package

wojciechka:


external-dns.yaml
```diff
@@ -37,6 +37,26 @@ pipeline:
 
   - uses: strip
 
+subpackages:
+  - name: ${{package.name}}-bitnami-compat
+    dependencies:
+      provides:
+        - mysql-bitnami-compat=${{package.full-version}}
```
kranurag7, line 44:
```suggestion

```
I think we can remove this given it's not version streamed. also mysql (hehe, get over it now)

kranurag7 (approved):

PR #48124, opened 2025-03-25
nginx-prometheus-exporter: adds missing symlink

Dentrax:
Adds missing `exporter` symlink: https://github.com/bitnami/charts/blob/3521264684c5a1b09a50a5c8c50453cc9af62f02/bitnami/nginx/templates/deployment.yaml#L351C15-L351C23


kranurag7 (approved):
we agreed that one symlink is better than multiple nested one.

PR #48164, opened 2025-03-25
Add cluster-api

utieyin:
Add cluster-api meta package version 1.9.6 with subpackages and pipelines
The package requires a full blown kubernetes cluster for complete testing so most of the tests are just to test the execution of the binaries for each component

cluster-api-1.9: new package

Fixes:

Related:
#### For new package PRs only

### DIFF
--- a/cluster-api-1.9.yaml
+++ b/cluster-api-1.9.yaml
@@ -1,0 +1,144 @@
+package:
+  name: cluster-api-1.9
+  version: "1.9.6"
+  epoch: 0
+  description: Cluster API meta package
+  dependencies:
+    provides:
+      - cluster-api=${{package.full-version}}
+  copyright:
+    - license: Apache-2.0
+
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: 813e90d8daef60b715764ce4729b40d82f811cc1
+      repository: https://github.com/kubernetes-sigs/cluster-api
+      tag: v${{package.version}}
+
+subpackages:
+  - name: ${{package.name}}-controller
+    description: cluster-api core controller
+    dependencies:
+      provides:
+        - cluster-api-controller=${{package.full-version}}
+    pipeline:
+      - uses: go/build
+        with:
+          output: manager
+          packages: .
+          ldflags: |
+            -X sigs.k8s.io/cluster-api/version.gitVersion=$(git describe --dirty --tags --always) \
+            -X sigs.k8s.io/cluster-api/version.gitCommit=$(git rev-parse HEAD) \
+            -X sigs.k8s.io/cluster-api/version.gitTreeState=$(shell git diff --quiet >/dev/null 2>&1 || echo "dirty" || echo "clean") \
+            -X sigs.k8s.io/cluster-api/version.buildDate=$(date -u +'%Y-%m-%dT%H:%M:%SZ')
+
+  - name: ${{package.name}}-kubeadm-bootstrap-controller
+    description: cluster-api core controller
+    dependencies:
+      provides:
+        - cluster-api-kubeadm-bootstrap-controller=${{package.full-version}}
+    pipeline:
+      - uses: go/build
+        with:
+          output: kubeadm-bootstrap-controller
+          packages: ./bootstrap/kubeadm
+          ldflags: |
+            -X sigs.k8s.io/cluster-api/version.gitVersion=$(git describe --dirty --tags --always) \
+            -X sigs.k8s.io/cluster-api/version.gitCommit=$(git rev-parse HEAD) \
+            -X sigs.k8s.io/cluster-api/version.gitTreeState=$(shell git diff --quiet >/dev/null 2>&1 || echo "dirty" || echo "clean") \
+            -X sigs.k8s.io/cluster-api/version.buildDate=$(date -u +'%Y-%m-%dT%H:%M:%SZ')
+
+  - name: ${{package.name}}-kubeadm-controlplane-controller
+    description: cluster-api core controller
+    dependencies:
+      provides:
+        - cluster-api-kubeadm-controlplane-controller=${{package.full-version}}
+    pipeline:
+      - uses: go/build
+        with:
+          output: kubeadm-controlplane-controller
+          packages: ./controlplane/kubeadm
+          ldflags: |
+            -X sigs.k8s.io/cluster-api/version.gitVersion=$(git describe --dirty --tags --always) \
+            -X sigs.k8s.io/cluster-api/version.gitCommit=$(git rev-parse HEAD) \
+            -X sigs.k8s.io/cluster-api/version.gitTreeState=$(shell git diff --quiet >/dev/null 2>&1 || echo "dirty" || echo "clean") \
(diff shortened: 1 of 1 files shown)


cluster-api-1.9.yaml
```diff
@@ -0,0 +1,143 @@
...
+      tag: v${{package.version}}
+
+  - uses: go/bump
+    with:
+      deps: |-
+        golang.org/x/crypto@v0.35.0
+        golang.org/x/oauth2@v0.27.0
+        golang.org/x/net@v0.36.0
+
+subpackages:
```
Dentrax, line 26: Can we use `item:` pipeline here? And iterate over packages in the single subpkg. `nvidia-container-toolkit` comes mind as an example.

cluster-api-1.9.yaml
```diff
@@ -0,0 +1,143 @@
...
+    - uses: git-checkout
+      with:
+        expected-commit: 813e90d8daef60b715764ce4729b40d82f811cc1
+        repository: https://github.com/kubernetes-sigs/cluster-api
+        tag: v${{package.version}}
+    - uses: test/kwok/cluster
+    - runs: |
+        kubectl apply -f config/crd/bases/
+    - uses: test/daemon-check-output
+      with:
```
maxgio92, line 116:
```suggestion
    - uses: test/daemon-check-output
      with:
        setup: kubectl apply -f config/crd/bases/
```

cluster-api-1.9.yaml
```diff
@@ -0,0 +1,143 @@
...
+            -X sigs.k8s.io/cluster-api/version.gitVersion=${{package.version}} \
+            -X sigs.k8s.io/cluster-api/version.gitCommit=${GIT_COMMIT} \
+            -X sigs.k8s.io/cluster-api/version.gitTreeState=$(shell git diff --quiet >/dev/null 2>&1 || echo "dirty" || echo "clean") \
+            -X sigs.k8s.io/cluster-api/version.buildDate=$(date -u +'%Y-%m-%dT%H:%M:%SZ')
+
+  - name: ${{package.name}}-clusterctl
+    description: cluster-api command line tool
+    dependencies:
+      provides:
+        - clusterctl=${{package.full-version}}
```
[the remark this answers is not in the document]
maxgio92: Yeah I think we should transform the package name in order to move the version as suffix, like `cluster-api-clusterctl-1.9`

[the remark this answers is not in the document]
maxgio92: Being version streamed I think it's correct, like `cluster-api-1.9-clusterctl`

cluster-api-1.9.yaml
utieyin: item doesn't work a 100% but added where useful

cluster-api-1.9.yaml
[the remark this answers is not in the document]
utieyin: There are existing packages for cluster-api, might revisit this after removing previous packages

cluster-api-1.9.yaml
utieyin: As the crd apply step is needed by the other daemon check output steps, we agreed to keep it separate

maxgio92 (dismissed):

cluster-api-1.9.yaml
```diff
@@ -0,0 +1,136 @@
...
+          Registering webhook
+          Conversion webhook enabled
+          Starting manager
+          problem running manager
+    - uses: test/daemon-check-output
+      with:
+        start: /kubeadm-bootstrap-controller
+        expected_output: |
+          Stopping and waiting for webhooks
+          Stopping and waiting for HTTP servers
```
Dentrax, line 125: Should we drop a comment above why we are doing negative-tests?

cluster-api-1.9.yaml
```diff
@@ -0,0 +1,136 @@
...
+          Starting EventSource
+          All workers finished
+    - uses: test/daemon-check-output
+      with:
+        start: /kubeadm-controlplane-controller
+        expected_output: |
+          Registering webhook
+          Conversion webhook enabled
+          Starting manager
+          problem running manager
```
Dentrax, line 119: Same here, comment would be nice.

cluster-api-1.9.yaml
```diff
@@ -0,0 +1,136 @@
...
+        repository: https://github.com/kubernetes-sigs/cluster-api
+        tag: v${{package.version}}
+    - uses: test/kwok/cluster
+    - runs: |
+        kwokctl scale node --replicas 3
+    - uses: test/daemon-check-output
+      with:
+        setup: kubectl apply -f config/crd/bases/
+        start: /manager --profiler-address=127.0.0.1:8080 --health-addr=127.0.0.1:8081
+        expected_output: |
```
Dentrax, line 107: Should we have a `post:` step to test with `wait-for-it` the `8081` port and then `curl | grep` tests, if possible?

cluster-api-1.9.yaml
```diff
@@ -0,0 +1,136 @@
...
+        - ${{package.name}}-clusterctl-compat
+  pipeline:
+    - uses: git-checkout
+      with:
+        expected-commit: 813e90d8daef60b715764ce4729b40d82f811cc1
+        repository: https://github.com/kubernetes-sigs/cluster-api
+        tag: v${{package.version}}
+    - uses: test/kwok/cluster
+    - runs: |
+        kwokctl scale node --replicas 3
```
Dentrax, line 102: Should we drop a comment why we need to scale the Nodes?

cluster-api-1.9.yaml
```diff
@@ -0,0 +1,136 @@
...
+        /clusterctl version | grep -q ${{package.version}}
+        /clusterctl --help | grep -q "Get started with Cluster API using clusterctl"
+        /clusterctl config repositories | grep helm
+
+update:
+  enabled: true
+  github:
+    identifier: kubernetes-sigs/cluster-api
+    strip-prefix: v
+    tag-filter: v
```
Dentrax, line 136: If we have `-1.9` version suffix, we should have the filter that version too. Otherwise this would pick up the latest.

cluster-api-1.9.yaml
utieyin: Added a comment to the PR notes but have also included same at the beginning of the tests.
We cannot execute the binaries outside a fake kubernetes cluster and without access to one, the next best thing is a fail test to show the binary itself is executable

cluster-api-1.9.yaml
utieyin: See above

cluster-api-1.9.yaml
utieyin: See above

cluster-api-1.9.yaml
utieyin: Not needed, removed

cluster-api-1.9.yaml
utieyin: Done, I think?

Dentrax (dismissed):
Thanks for the work! 💜

maxgio92 (approved):

PR #48199, opened 2025-03-25
Upgrade aws-network-policy-agent to version 1.20

murraybd:
This updates the package to version 1.20 which a bot had tried to due but it failed to build from source due to upstream's usage of `inline` with a function. This is resolved by patching the upstream code to use libbpf's macro `_always_inline`. 

Additionally, while I was here I added a patch to drop `-march=bpf` from the Makefile because while it is a warning with clang-16 it will cause a failure with clang-18.

Finally, I also resolved a warning from the linter:

```
2025/03/25 20:18:04 WARN linter "object" failed on package "aws-network-policy-agent": package contains intermediate object file "usr/bin/tc.v4egress.bpf.o". This is usually wrong. In most cases they should be removed: usr/bin/tc.v4egress.bpf.o; suggest: This package contains intermediate object files
```

For the record the original package update was in https://github.com/wolfi-dev/os/pull/43390.


sil2100 (approved):
Me likey. LGTM!

PR #48222, opened 2025-03-25
glibc - usrmerge /lib

zmarano:


xnox (changes requested):
This has triggered the known issue that moved ld.so on x86_64 in a way that breaks compatibility, elf scanners and parsers.

```
curl -L https://apk.cgr.dev/wolfi-presubmit/d60879ea2598c041d975a1361929d4eb14fc1002/x86_64/tzutils-2.41-r50.apk | tar -xz
file usr/bin/zdump 
usr/bin/zdump: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /usr/lib/ld-linux-x86-64.so.2, for GNU/Linux 4.9.0, not stripped
```

Note that interpreter is declared as /usr/lib/ld-linux-x86-64.so.2 which is a breakage, because that string constant is an ABI. 

Observe that on all other usrmerge systems the declared elf interpreter remains at the historical location.

Ubuntu:
```
$ file /usr/bin/zdump 
/usr/bin/zdump: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=027cd645e7468b51ee601536acc95f359fb03b22, for GNU/Linux 3.2.0, stripped
```
Fedora:
```
# file /usr/bin/zdump
/usr/bin/zdump: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=aaa77681e00b0df19b059ffc9deb9a112248f07c, for GNU/Linux 3.2.0, stripped
```

Despite that for example on ubuntu, it is a symlink to a multiarch location:
```
$ realpath /lib64/ld-linux-x86-64.so.2
/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
```

Hilariously, we already are not using the historical location.
```
# file /usr/bin/zdump 
/usr/bin/zdump: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-x86-64.so.2, for GNU/Linux 4.9.0, with debug_info, not stripped
```

I feel like we should correct this. and set whatever we need to set, for glibc to install `/lib64/ld-linux-x86-64.so.2` and use that as interpreter path for ELF binaries. And then fix up the install paths in the apk.

I am surprised that `/lib/ld-linux-x86-64.so.2` has been working for us so well for so long. But it might be a reason why some of our CNI plugins were failing to execute when copied out to various host OSes.

Please set rtlibdir to /lib64 on x86_64, and /lib on arm64.

Such that binaries are produced with ELF interpreter paths /lib64/ld-linux-x86-64.so.2 and /lib/ld-linux-aarch64.so.1.

zmarano:
> Please set rtlibdir to /lib64 on x86_64, and /lib on arm64.

I was wondering about this- now I see why that comment was there.

glibc.yaml
```diff
@@ -76,19 +71,24 @@ pipeline:
...
       echo "user-defined-trusted-dirs=/usr/local/lib:/usr/local/lib64:/usr/lib:/usr/lib64:/lib:/lib64" >> configparms.base
-      echo "rtlddir=/lib" >> configparms.base
-      echo "slibdir=/lib" >> configparms.base
+      echo "slibdir=/usr/lib" >> configparms.base
       echo "libdir=/usr/lib" >> configparms.base
       echo "rootsbindir=/usr/bin" >> configparms.base
-
       echo "build-programs=no" > configparms
+      # If x86_64, set to /lib64, /lib for aarch64
+      if [[ ${{build.arch}} == "x86_64" ]]; then
```
xnox, line 79: no to bashism, please use `[ "${{build.arch}} = "x86_64" ]`

zmarano: old habits.... fixed

glibc.yaml
```diff
@@ -128,19 +124,24 @@ pipeline:
...
       echo "user-defined-trusted-dirs=/usr/local/lib:/usr/local/lib64:/usr/lib:/usr/lib64:/lib:/lib64" >> configparms.base
-      echo "rtlddir=/lib" >> configparms.base
-      echo "slibdir=/lib" >> configparms.base
+      echo "slibdir=/usr/lib" >> configparms.base
       echo "libdir=/usr/lib" >> configparms.base
       echo "rootsbindir=/usr/bin" >> configparms.base
-
       echo "build-programs=no" > configparms
+      # If x86_64, set to /lib64, /lib for aarch64
+      if [ ${{build.arch}} == "x86_64" ]; then
```
sergiodj, line 132:
```suggestion
      if [ "${{build.arch}}" = "x86_64" ]; then
```

sergiodj (dismissed):

xnox (dismissed):

smoser (approved):

PR #48225, opened 2025-03-25
usrmerge - /lib - wolfi-baselayout

zmarano:


wolfi-baselayout.yaml
```diff
@@ -43,25 +43,17 @@ pipeline:
...
-      # ln -s bin ${{targets.destdir}}/usr/sbin
-      # ln -s usr/lib ${{targets.destdir}}/usr/lib
-
-      for i in lib64 usr/lib64 usr/local/lib64; do
-        ln -s lib "${{targets.destdir}}"/${i}
-      done
+      ln -s usr/bin ${{targets.destdir}}/usr/sbin
+      ln -s usr/lib ${{targets.destdir}}/lib
+      ln -s usr/lib ${{targets.destdir}}/lib64
+      ln -s usr/lib ${{targets.destdir}}/usr/lib64
```
a-crate, line 56: You're not doing anything with `usr/local/lib64` or `usr/local/lib`

zmarano: Yeah, it's not clear they need to even exist... I was going to bring that up as well. I think usr/local/lib64 was just "there" for no reason. And we aren't doing anything with usr/local/bin or usr/local/sbin anywhere else.

a-crate: I suspect it's there to make sure that `usr/local/lib64` is a symlink to `usr/local/lib`, like the other lib directories.

zmarano: I was looking around at other distros versions of the world. Ubuntu for example doesn't have a /usr/local/lib64 at all. Fedora and RHEL do but it's not a symlink- just a directory with nothing in it. I'm not sure if there is actually one model that is right or wrong though. Also, I don't see anyone not keeping /usr/lib64 around  instead of linking it to /usr/lib ... I haven't yet found the definitive reason why. But for example, Ubuntu puts ld-linux in /usr/lib and creates symlinks in /usr/lib64 which is then linked back to /lib64 so in effect there's only a symlink in /usr/lib64 and no real files. So ehh?

wolfi-baselayout.yaml
```diff
@@ -43,25 +43,17 @@ pipeline:
 
   - name: Install
     runs: |
-      for i in dev etc etc/profile.d etc/secfixes.d home proc root var/log usr/bin usr/local/lib sys tmp var/spool/cron opt run usr/lib; do
+      for i in dev etc etc/profile.d etc/secfixes.d home proc root var/log usr/bin sys tmp var/spool/cron opt run usr/lib; do
```
smoser, line 46: dropping /usr/local/lib is not necessarily related here, is it?

zmarano: While true it isn't directly related, wolfi-baselayout isn't making usr/local directories for anything else and shouldn't need to. Some packages do create these directories for their own purposes. This felt like a useful cleanup to make. We could however leave it to a future cleanup.

xnox: BREAKING CHANGE - please put it back.

xnox: Essentially, people who use read-only mounts and read-only containers cannot bind mount files to that destination unless it exists.

Also i  am surprised we are getting away without usr/local/bin

xnox: also further down the road usr/local/lib64 is also removed.

zmarano: gotcha .... I remember seeing usr/local/bin somewhere else (not in wolfi-baselayout) and I don't remember where now

smoser (approved):

blaine-arcjet:
This change seemed to have broken builds. It looks like it was responsible for https://github.com/wolfi-dev/os/issues/57413

PR #48242, opened 2025-03-26
include merged-bin as runtime dep for thanos

kranurag7:
something like bitnami compat


kranurag7:
explicit is better but we don't need this.

PR #48316, opened 2025-03-26
Libnvidia container use ldd check

murraybd:
I was looking for packages which called `ldd` directly to test `.so` files instead of using the `ldd-check` test from tw and found one and also added the test to spdlog.


dannf (approved):
well spotted

PR #48344, opened 2025-03-27
ingress-nginx-controller: Revert otel plugin version to the compatible version

max-allan-cgr:
Fixes:
The SHA of the OTEL plugin must match the upstream project!


maxgio92:
Sorry @max-allan-cgr, even this one came first, I'm going to close it in favor of https://github.com/wolfi-dev/os/pull/48349, as there's some convo already there.

PR #48345, opened 2025-03-27
unbound: add compat package for mailcow

javacruft:
Add unbound-mailcow-compat package to support use of unbound in mailcow compatible images.

Move default unbound.conf into new unbound-conf package and use a provides=unbound-conf with a lower priority to allow the mailcow configuration to be switched in when needed.

### DIFF
--- a/unbound.yaml
+++ b/unbound.yaml
@@ -1,10 +1,14 @@
 package:
   name: unbound
   version: 1.22.0
-  epoch: 1
+  epoch: 2
   description: "Unbound is a validating, recursive, and caching DNS resolver."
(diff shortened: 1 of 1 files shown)


sil2100 (approved):
This is interesting. Not sure if we've done things like this before, but it makes sense - this way the mailcow compat package is here where the context is. Might be a bit confusing from the SBOM POV but we'll have to revisit that at some point anyway.

Stylistic nitpick (but not needed): the `-mailcow-compat` subpackage description is capitalized, but all the other subpackage descriptions are lower-case.

PR #48355, opened 2025-03-27
remove `ini-file` from `mysql-bitnami`

rakshitgondwal:
This PR removes `ini-file` from the package level.


kranurag7 (approved):
given it's already at image level, looks good.

PR #48412, opened 2025-03-27
apk-tools: create /lib usrmerge patch

a-crate:
A small patch which moves `/lib/apk` to `/usr/lib/apk`. apk-tools upstream plans for handling usrmerge included auto-detecting the DB location and moving the lockfile to /run for FHS compliance. This does none of that, it's really just enough to make `--initdb` work for an already usrmerged system.

```
97cabd7f6480:~# apk add --allow-untrusted -X /work/packages --initdb --root tmp wolfi-baselayout
(1/2) Installing ca-certificates-bundle (20241121-r2)
(2/2) Installing wolfi-baselayout (20230201-r21)
OK: 0 MiB in 2 packages
97cabd7f6480:~# ls -l tmp
total 44
lrwxrwxrwx    1 root     root             7 Mar 27 20:07 bin -> usr/bin
drwxr-xr-x    2 root     root          4096 Mar 27 20:07 dev
drwxr-xr-x    7 root     root          4096 Mar 27 20:07 etc
drwxrwxr-x    2 root     root          4096 Mar 27 20:07 home
lrwxrwxrwx    1 root     root             7 Mar 27 20:07 lib -> usr/lib
lrwxrwxrwx    1 root     root             3 Mar 27 20:07 lib64 -> lib
drwxrwxr-x    2 root     root          4096 Mar 27 20:07 opt
dr-xr-xr-x    2 root     root          4096 Mar 27 20:07 proc
drwx------    2 root     root          4096 Mar 27 20:07 root
drwxrwxr-x    2 root     root          4096 Mar 27 20:07 run
lrwxrwxrwx    1 root     root             7 Mar 27 20:07 sbin -> usr/bin
drwxrwxr-x    2 root     root          4096 Mar 27 20:07 sys
drwxrwxrwt    2 root     root          4096 Mar 27 20:07 tmp
drwxr-xr-x    5 root     root          4096 Mar 27 20:07 usr
drwxr-xr-x    7 root     root          4096 Mar 27 20:07 var
97cabd7f6480:~#
```

### DIFF
--- a/apk-tools.yaml
+++ b/apk-tools.yaml
@@ -1,7 +1,7 @@
 package:
   name: apk-tools
   version: "2.14.10"
-  epoch: 2
+  epoch: 3
   description: "apk-tools (Wolfi package manager)"
   copyright:
     - license: GPL-2.0-only
@@ -31,8 +31,9 @@ pipeline:
       tag: v${{package.version}}
       expected-commit: 9d074efdc12bc41b5d24190595a5269a770e852a
 
-  - runs: |
-      git am 292.patch
+  - uses: patch
+    with:
+      patches: 292.patch usrmerge-lib.patch
 
   - runs: |
       sed -i -e 's:-Werror::' Make.rules
--- a/apk-tools/usrmerge-lib.patch
+++ b/apk-tools/usrmerge-lib.patch
@@ -1,0 +1,17 @@
+diff --git a/src/database.c b/src/database.c
+index dc5e4fd..f33ebbc 100644
+--- a/src/database.c
++++ b/src/database.c
+@@ -1299,6 +1299,12 @@ static int apk_db_create(struct apk_database *db)
+ 	mknodat(db->root_fd, "dev/console", S_IFCHR | 0600, makedev(5, 1));
+ 	mkdirat(db->root_fd, "etc", 0755);
+ 	mkdirat(db->root_fd, "etc/apk", 0755);
++	struct stat lib_stat;
++	if (lstat("/lib", &lib_stat) == 0 && S_ISLNK(lib_stat.st_mode)) {
++		mkdirat(db->root_fd, "usr", 0755);
++		mkdirat(db->root_fd, "usr/lib", 0755);
++		symlinkat("usr/lib", db->root_fd, "lib");
++	}
+ 	mkdirat(db->root_fd, "lib", 0755);
+ 	mkdirat(db->root_fd, "lib/apk", 0755);
+ 	mkdirat(db->root_fd, "lib/apk/db", 0755);


a-crate:
CI fails because `apko` creates `/lib` still, a version of apko with [this patch](https://github.com/chainguard-dev/apko/pull/1593) should build successfully. It can also build without the `merged-lib` dependency and `wolfi-baselayout` changes if you want to see it locally.

smoser:
@a-crate , thanks.
how easily could you test only a change to initdb like this:

```patch
commit 65c066c3db62b75cf5659e216aa4fb838db58ec9 (HEAD)
Author: Scott Moser <smoser@brickies.net>
Date:   Thu Mar 27 16:53:48 2025 -0400

    try initdb in usrmerge-lib world

diff --git a/src/database.c b/src/database.c
index 7eb75a1..3e59bfc 100644
--- a/src/database.c
+++ b/src/database.c
@@ -1290,7 +1290,7 @@ static int add_protected_paths_from_file(void *ctx, int dirfd, const char *file)
 
 static int apk_db_create(struct apk_database *db)
 {
-       int fd;
+       int fd, r;
 
        mkdirat(db->root_fd, "tmp", 01777);
        mkdirat(db->root_fd, "dev", 0755);
@@ -1301,7 +1301,14 @@ static int apk_db_create(struct apk_database *db)
        mknodat(db->root_fd, "dev/console", S_IFCHR | 0600, makedev(5, 1));
        mkdirat(db->root_fd, "etc", 0755);
        mkdirat(db->root_fd, "etc/apk", 0755);
-       mkdirat(db->root_fd, "lib", 0755);
+
+       r = symlinkat("usr/lib", db->root_fd, "lib");
+    if (r == -1) {
+        apk_error("failed to symlink lib -> usr/lib");
+        return -errno;
+    }
+       mkdirat(db->root_fd, "usr", 0755);
+       mkdirat(db->root_fd, "usr/lib", 0755);
        mkdirat(db->root_fd, "lib/apk", 0755);
        mkdirat(db->root_fd, "lib/apk/db", 0755);
        mkdirat(db->root_fd, "var", 0755);
```

That is, let it continue thinking its db is at /lib/db, but on initdb we create the linked path.

a-crate:
I can test pretty easily. That works as well, I get the same (correct) result with `apk --initdb` (`lib -> usr/lib` and it installs wolfi-baselayout without errors). Your patch doesn't apply exactly to the `v2.14.10` tag so I had to recreate it but the content of the changes is the same.

apk-tools/usrmerge-lib.patch
```diff
@@ -0,0 +1,33 @@
...
+ 	mknodat(db->root_fd, "dev/console", S_IFCHR | 0600, makedev(5, 1));
+ 	mkdirat(db->root_fd, "etc", 0755);
+ 	mkdirat(db->root_fd, "etc/apk", 0755);
++	r = symlinkat("usr/lib", db->root_fd, "lib");
++	if (r == -1) {
++	apk_error("failed to symlink lib -> usr/lib");
++		return -errno;
++	}
++	mkdirat(db->root_fd, "usr", 0755);
++	mkdirat(db->root_fd, "usr/lib", 0755);
```
smoser, line 30: Just a thought...I had. I'm not sure what is right.
They basically just ignore all errors here.  I think the intent is to not fail on subsequent apk runs that have '--initdb' in attempt to be more "user-friendly".  Ultimately here they just take the path that if they can open etc/apk/world and lib/apk/db/installed then they're happy.

I chose to error above if the symlink fails.  I think that isn't right.

As it is here
1. subsequent calls to '--initdb' will fail because the symlinkat will return EEXIST.
2. if lib exists as a directory it will fail.

I'm pretty sure we to ignore the error in 1.
I'm not sure what we want to do in 2.

Do we want '--initdb' to work on a non-usrmerged lib?  thoughts?

a-crate: I think we should make case 1 succeed, I am indifferent on making case 2 succeed but if ignoring the return value will make both work I think that's fine.

a-crate:
Pushed a version which works on DBs in existing chroots regardless of whether it's usrmerged or not. Does not do any conversion.
```
62611f88ef33:~# mkdir -p tmp-libmerge
62611f88ef33:~# mkdir -p tmp-nolibmerge/lib
62611f88ef33:~# apk add --allow-untrusted -X /work/packages --initdb --root tmp-libmerge wolfi-baselayout
(1/2) Installing ca-certificates-bundle (20241121-r2)
(2/2) Installing wolfi-baselayout (20230201-r21)
OK: 0 MiB in 2 packages
62611f88ef33:~# ls -l tmp-libmerge
total 44
lrwxrwxrwx    1 root     root             7 Mar 31 21:17 bin -> usr/bin
drwxr-xr-x    2 root     root          4096 Mar 31 21:17 dev
drwxr-xr-x    7 root     root          4096 Mar 31 21:17 etc
drwxrwxr-x    2 root     root          4096 Mar 31 21:17 home
lrwxrwxrwx    1 root     root             7 Mar 31 21:17 lib -> usr/lib
lrwxrwxrwx    1 root     root             3 Mar 31 21:17 lib64 -> lib
drwxrwxr-x    2 root     root          4096 Mar 31 21:17 opt
dr-xr-xr-x    2 root     root          4096 Mar 31 21:17 proc
drwx------    2 root     root          4096 Mar 31 21:17 root
drwxrwxr-x    2 root     root          4096 Mar 31 21:17 run
lrwxrwxrwx    1 root     root             7 Mar 31 21:17 sbin -> usr/bin
drwxrwxr-x    2 root     root          4096 Mar 31 21:17 sys
drwxrwxrwt    2 root     root          4096 Mar 31 21:17 tmp
drwxr-xr-x    5 root     root          4096 Mar 31 21:17 usr
drwxr-xr-x    7 root     root          4096 Mar 31 21:17 var
62611f88ef33:~# apk add --allow-untrusted -X /work/packages --initdb --root tmp-nolibmerge wolfi-baselayout # unpack error is a wolfi-baselayout thing not an apk-tools thing
(1/2) Installing ca-certificates-bundle (20241121-r2)
(2/2) Installing wolfi-baselayout (20230201-r21)
ERROR: wolfi-baselayout-20230201-r21: failed to rename .apk.27182ee91faffcfc8d2bb0f6fc8e5d6a1c64151607c91750 to lib.
1 error; 0 MiB in 2 packages
62611f88ef33:~# ls -l tmp-nolibmerge/
total 48
lrwxrwxrwx    1 root     root             7 Mar 31 21:17 bin -> usr/bin
drwxr-xr-x    2 root     root          4096 Mar 31 21:17 dev
drwxr-xr-x    7 root     root          4096 Mar 31 21:17 etc
drwxrwxr-x    2 root     root          4096 Mar 31 21:17 home
drwxr-xr-x    3 root     root          4096 Mar 31 21:17 lib
lrwxrwxrwx    1 root     root             3 Mar 31 21:17 lib64 -> lib
drwxrwxr-x    2 root     root          4096 Mar 31 21:17 opt
dr-xr-xr-x    2 root     root          4096 Mar 31 21:17 proc
drwx------    2 root     root          4096 Mar 31 21:17 root
drwxrwxr-x    2 root     root          4096 Mar 31 21:17 run
lrwxrwxrwx    1 root     root             7 Mar 31 21:17 sbin -> usr/bin
drwxrwxr-x    2 root     root          4096 Mar 31 21:17 sys
drwxrwxrwt    2 root     root          4096 Mar 31 21:17 tmp
drwxr-xr-x    5 root     root          4096 Mar 31 21:17 usr
drwxr-xr-x    7 root     root          4096 Mar 31 21:17 var
62611f88ef33:~#
```

xnox:
I like how suicinically small this change is, making this patch trivial to maintain.

Even if creating /lib directory call will always fail, and so be it.

apk-tools/usrmerge-lib.patch
```diff
@@ -0,0 +1,15 @@
...
+--- a/src/database.c
++++ b/src/database.c
+@@ -1299,7 +1299,10 @@ static int apk_db_create(struct apk_database *db)
+ 	mknodat(db->root_fd, "dev/console", S_IFCHR | 0600, makedev(5, 1));
+ 	mkdirat(db->root_fd, "etc", 0755);
+ 	mkdirat(db->root_fd, "etc/apk", 0755);
++	mkdirat(db->root_fd, "usr", 0755);
++	mkdirat(db->root_fd, "usr/lib", 0755);
++	symlinkat("usr/lib", db->root_fd, "lib");
+ 	mkdirat(db->root_fd, "lib", 0755);
```
smoser, line 12: maybe one nit. if symlink above succeeds, then mkdirat lib is guaranteed to fail. so maybe only try it if the symlink failed?

smoser (commented):
I have one nit in line, and i'm not stuck on it.

I think we could actually land the apk change here.
it will only be used when someone does `apk --initdb` and that operation should be backwards and forwards compatible, right?

apk-tools.yaml
```diff
@@ -1,13 +1,14 @@
...
   version: "2.14.10"
-  epoch: 2
+  epoch: 50
   description: "apk-tools (Wolfi package manager)"
   copyright:
     - license: GPL-2.0-only
   dependencies:
     runtime:
       - ca-certificates-bundle
+      - merged-lib
```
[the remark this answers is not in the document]
a-crate: I was thinking that it might still be desirable even though its not true, that old systems should stay in a totally non-libmerged world until they upgrade to an image that has merged-lib. But that won't exactly hold true I suppose, if they do `--initdb` and pull a version of wolfi-baselayout that provides merged-lib things will get a bit unreliable no matter what.

a-crate:
It is backwards compatible with existing chroots but whenever it's used to create new ones it will do `lib -> usr/lib` by default, and IMO we should not merge that until it has a wolfi-baselayout with merged-lib to pull into that new chroot.

apk-tools/usrmerge-lib.patch
a-crate: SGTM. I also moved mkdir usr/lib stuff to only happen if it succeeds. So we only create the directory structure that is needed.

smoser (dismissed):
I like it.

I think we should land apk-tools. its one less thing to worry about later, and one less thing that might cause revert during that migration.

If it causes problems we can back it out.

kbsteere (dismissed):

kbsteere:
Ignore my approval

a-crate:
Pushed a version which only does /lib usrmerge if the host running apk is usrmerged (based on discussion with @xnox today).

I think this is backwards and forwards compatible and reasonable to land ahead of time, as it won't start merging things until we land wolfi-baselayout and will stop if we retract it.
It will behave strangely on non-wolfi systems but I don't think that matters.

I didn't add the flag we discussed (`--no-usrmerge`) to skip doing the `/lib` symlink. I'm comfortable merging without it, but can also make time later to implement that. It will definitely increase patch complexity though, and will no longer be so simple to carry.

zmarano (approved):

xnox (approved):

PR #48420, opened 2025-03-27
Add the ldd-check test to samba and add runtime deps

murraybd:
Multiple subpackages required the addition of runtime dependencies for the ldd-check to pass because those subpackages were missing other shared objects.

### DIFF
--- a/samba.yaml
+++ b/samba.yaml
@@ -1,7 +1,7 @@
 package:
   name: samba
   version: "4.22.0"
-  epoch: 0
+  epoch: 1
   description: "Tools to access a server's filespace and printers via SMB"
   copyright:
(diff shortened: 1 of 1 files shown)


sergiodj (approved):
Thanks, @murraybd .

This LGTM.

Samba ships a lot of private libraries (which are installed under `/usr/lib/samba` in our case).  In this case, it's not desirable for us to use `ld.so.conf.d` snippets because they would incorrectly make those libraries available to external programs.  For that reason, there's not much we can do during build time (`melange`-wise) to create the dependency graph between the packages the provide and depend those libraries.

All that is to say that we're stuck with manually listing the runtime dependencies, which is what you did here.

+1

PR #48430, opened 2025-03-28
Add ldd-check test to several packages

murraybd:
More packages could use testing.


murraybd:
The package builds, except for soxr, passed here https://github.com/wolfi-dev/os/pull/48430/checks?check_run_id=39593757490. Soxr and everything else passed when I used "update branch" but I don't know where to find that.

gaughen (approved):

PR #48497, opened 2025-03-28
py3-altair: new package

MickeyPvX:
Related: https://github.com/chainguard-dev/image-requests/issues/5554

Building dependencies for installed Python libraries on the public sagemaker-distribution image.

#### For new package PRs only


py3-altair.yaml
```diff
@@ -0,0 +1,112 @@
...
+  - name: py3-supported-${{vars.pypi-package}}
+    description: meta package providing ${{vars.pypi-package}} for supported python versions.
+    dependencies:
+      runtime:
+        - py3.10-${{vars.pypi-package}}
+        - py3.11-${{vars.pypi-package}}
+        - py3.12-${{vars.pypi-package}}
+        - py3.13-${{vars.pypi-package}}
+
+test:
```
EyeCantCU, line 73: We can move this test to the subpackage above and run for all versions instead of just latest

py3-altair.yaml
```diff
@@ -0,0 +1,112 @@
...
+  environment:
+    contents:
+      packages:
+        - jq
+        - py3-pandas
```
EyeCantCU, line 78:
```suggestion
        - py${{range.key}}-pandas
```

EyeCantCU (changes requested):

py3-altair.yaml
MickeyPvX: Done!

py3-altair.yaml
MickeyPvX: Also done!

EyeCantCU (approved):

PR #48621, opened 2025-03-31
unbound: misc updates for unbound-mailcow-compat

javacruft:
Expand support in mailcow-compat subpackage for supervisord and syslog-ng configuration and a service healthcheck all sourced from the mailcow-dockerized repository.

Rename -conf -> -config (which seems to be more of a standard) and adjust priority around default of 0 so  that alternative configuration files can be switched in without  broader impact.

Drop unbound-mailcow-compat 1.22.0-r2  - r3 sets the provider-priority to -1 to ensure compatibility with  -config packages which don't set a provider-priority - drop r2  version of package to avoid this being used with new unbound    package versions.

### DIFF
--- a/unbound.yaml
+++ b/unbound.yaml
@@ -1,13 +1,13 @@
 package:
   name: unbound
   version: 1.22.0
-  epoch: 2
+  epoch: 3
   description: "Unbound is a validating, recursive, and caching DNS resolver."
   copyright:
     - license: BSD-3-Clause
   dependencies:
     runtime:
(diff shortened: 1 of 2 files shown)


sil2100 (approved):
OK, this makes sense!

Unrelated to these changes (thing for a separate PR): I think with the recent additions of `-mailcow-compat` we'll have to re-visit the licencing metadata here. It looks like the mailcow-dockerized repo is GPLv3...

PR #48670, opened 2025-03-31
exim: set BIN_DIRECTORY to usr/bin

xnox:
make install is called multiple times, thus it is best to set correct
destination for all invocations, instead of fixing them one by one.


roaksoax (approved):
lgtm!

PR #48676, opened 2025-03-31
shadow: usrmerge usrsbin

a-crate:


zmarano (dismissed):

a-crate:
Test changes are because shadow's nologin always exits 1 no matter what, but busybox noglogin exits 0 with --help. 

Before this PR shadow-login was not overwriting busybox nologin so it was actually testing busybox nologin. After this change it overwrites correctly, and now exits 1 even with --help.

justinvreeland (approved):

PR #48686, opened 2025-03-31
tests: add service file test pipeline and use it in in a few packages

justinvreeland:
The tar lines, while I don't like fix an issues that systemd-analyze
verify found where symlinks are turned into real files. Test was moved
forward in the pipeline because we don't have install via file lists
and otherwise other packages will pickup the test files when they shoulnd't.
    
All the service files aren't 100% clean but they're not longer failing.

This should probably also have a warning linter like ldd or docs do.

### DIFF
--- a/pipelines/test/verify-service.yaml
+++ b/pipelines/test/verify-service.yaml
@@ -1,0 +1,54 @@
+name: verify-service
+
+needs:
+  packages:
+    - apk-tools
+    - grep
+    - systemd
+
+inputs:
+  man:
+    description: Run verify without skipping doc tests
+    default: 'false'
+
+pipeline:
+  - name: Verify Service files
+    runs: |
+      set -xe
+
+      # If there's a templated service file systemd-analyze verify
+      # will fill in  test_instance for the template. If the
+      # test resource doesn't exist verify won't pass.
+      # The quaota service need a mountpoint to verify the quota service against.
+      # Which casued this but any service that depends on a template will probably
+      # need something.
+      cat << EOF > /usr/lib/systemd/system/test_instance.mount
+      [Unit]
+      Description=Test Mount
+      DefaultDependencies=no
+      Conflicts=umount.target
+      Before=local-fs.target umount.target
+
+      [Mount]
+      What=/dev/null
+      Where=/test_instance
+      EOF
+
+      # ${{package.name}} when run inside of a test pipeline is always the name of the top-level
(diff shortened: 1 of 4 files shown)


justinvreeland:
huh, compression is probably a waste of resources since I just want it to track the links and attrs for me.

systemd.yaml
```diff
@@ -324,22 +346,35 @@ subpackages:
...
           # Remove those files from the main package
           find . -name '*machine*' -not -name 'systemd-machine-id*' -exec rm -f '{}' \;
+
           find . -name '*nspawn*' -exec rm -f '{}' \;
           find . -name '*vmspawn*' -exec rm -f '{}' \;
           find . -name '*import*' -exec rm -f '{}' \;
           find . -name '*export*' -exec rm -f '{}' \;
-          find . -name '*portable*' -type f -exec rm -f '{}' \;
+          find . -name '*portable*' -exec rm -f '{}' \;
           find . -name '*portable*' -type d -exec rmdir '{}' \;
```
a-crate, line 378: This line is a noop now right?

pipelines/test/verify-service.yaml
a-crate, line 1: Failing when there are no service files in the package would be useful I think, both as a check to prevent this test from being added where it isn't needed and to make sure the packager didn't forget an install command.

a-crate (commented):
It would be nice to optionally install `-doc` or `-docs` and enable man checking if those exist but I think that would require melange changes to do in a clean way.

systemd.yaml
justinvreeland: Not from the rm change, rm -f still won't remove directories. This just unrestricted it from regular files since It wasn't removing symlinks.  The directory still won't be removed unless it's empty in the next line.

justinvreeland:
> It would be nice to optionally install `-doc` or `-docs` and enable man checking if those exist but I think that would require melange changes to do in a clean way.

Ohhh that I wanted those tests to run but forgot I could just install packages.

This should be doable by taking another input variable. We can then  pop the tests on the package with `withdocs: true` and then install the packages that have the service file as part of the environment for the docs tests. Could probably even check if the name of the package is `doc` and then enable the flag and assume the tester added the packages with the service files to the environment.  Failure on empty helps there as well in case you installed no service files in the doc env. Not as idea as a system where melange can manage it all but should be doable.

a-crate (approved):

PR #48729, opened 2025-04-01
util-linux: fix mount regression

xnox:
Which causes many helm chart deployment failures.

Thanks to @kranurag7 for report

Related:
- https://github.com/util-linux/util-linux/issues/3474
- https://github.com/k3s-io/k3s/issues/12006


kranurag7 (approved):
you're sonic fast, I was late by few seconds :) (kidding) 

thank you for fixing this.

PR #48809, opened 2025-04-02
py3-pybind11 correct name of cmake function

dannf:
There's a case error in the function name. It has been reported upstream:
      https://github.com/pybind/pybind11/issues/3996


sil2100 (approved):
LGTM! Thanks!

PR #48908, opened 2025-04-03
Set PYTHONHASHSEED for reproducible .pycs

imjasonh:


techalchemy (approved):
Gentoo seems to set `PYTHONHASHSEED` to `SOURCE_DATE_EPOCH`: https://wiki.gentoo.org/wiki/User:OstCollector/Reproducible_Build#The_use_of_SOURCE_DATE_EPOCH_and_PYTHONHASHSEED

OpenSUSE uses a static value of `0`: https://lists.reproducible-builds.org/pipermail/rb-general/2024-June/003436.html and https://github.com/openSUSE/post-build-checks/blob/master/suse-buildsystem.sh#L8

and it looks like arch has also fixed this by pinning to `0`: https://archlinux.org/todo/unreproducible-python-bytecode/

Note that I don't think it really matters what we set this to, as long as it is deterministic.

Relevant CPython Bug: https://github.com/python/cpython/issues/73894

cc @pnasrat

pnasrat (approved):
LGTM

PR #48916, opened 2025-04-03
Flannel

rahulduvedi:
Added compat package for flannel


Dentrax (approved):
https://github.com/flannel-io/flannel/blob/21e82fd13c5dd15525a410869b6c3f35a280fd23/images/Dockerfile#L32

PR #48928, opened 2025-04-03
add missing scripts to whereabouts package

developer-guy:
Signed-off-by: Batuhan Apaydin <batuhan.apaydin@chainguard.dev>

Upstream added two new scripts to the image so we have to do the same for package:
- https://github.com/k8snetworkplumbingwg/whereabouts/blame/d33f526057a47746ce345eec938a5bc3f0414e85/Dockerfile#L14

### DIFF
--- a/whereabouts.yaml
+++ b/whereabouts.yaml
@@ -1,7 +1,7 @@
 package:
   name: whereabouts
   version: "0.9.0"
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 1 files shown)


jamonation (approved):
Any tests that can check for missing scripts? Looks good to merge, but if you can think of a way to prevent this happening again it wouldn't hurt to add a check or two

developer-guy:
they have just added these scripts so I'm not sure if this is something we prevent from happening in that case :(

PR #48967, opened 2025-04-03
Disable DNSSEC by default in systemd-resolved

zmarano:


xnox (approved):
This is the only sane default. dnssec is fragile; and also allowing downgrade is fragile (poor performance, missed dns resolutions).

Will propose upstream change too.

PR #49002, opened 2025-04-03
Add build deps and extra-library-paths so ldd-checks pass

murraybd:
The ldd-check testing of all these openjdk packages was failing until I made these changes.

### DIFF
--- a/openjdk-12.yaml
+++ b/openjdk-12.yaml
@@ -121,15 +121,15 @@ subpackages:
     description: "OpenJDK 12 Java Runtime Environment"
     dependencies:
       runtime:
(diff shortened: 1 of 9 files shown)


dannf (commented):
It doesn't seem like the `-jre` packages should need `-dev` packages. Would the library packages be sufficient?

murraybd:
> It doesn't seem like the `-jre` packages should need `-dev` packages. Would the library packages be sufficient?

It seems like it.

dannf (approved):
ta! I assume (hope) a lot of this can get droped when @sergiodj 's melange work is merged.

PR #49017, opened 2025-04-03
pipelines/test/verify-service: test sockets, allow skipping files

a-crate:
```
amelia-crate@amelia-crate-16 ~/w/os (test-analyze-socket)> for pkg in (grep -ls "uses: test/verify-service" *); make test/(echo $pkg | sed s/.yaml//) >/dev/null 2>&1 || echo "Broke $pkg's tests!"; end
Broke openssh.yaml's tests!
Broke systemd.yaml's tests!
amelia-crate@amelia-crate-16 ~/w/os (test-analyze-socket)>
```

### DIFF
--- a/pipelines/test/verify-service.yaml
+++ b/pipelines/test/verify-service.yaml
@@ -7,6 +7,9 @@ needs:
     - systemd
 
 inputs:
+  skip-files:
+    description: Files to skip testing
+    default: ''
   man:
     description: Run verify without skipping doc tests
     default: 'false'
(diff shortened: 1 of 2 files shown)


a-crate:
Fixed systemd, openssh isn't actually broken an unrelated test just fails locally. See passing tests in de2b3fe87dc3bc7eb7fbd35c64fd5b6be90096de, I'm dropping the epoch bumps now.

justinvreeland (commented):
Why is syslog intentionally shipped without the service? It's a socket unit for socket activated services isn't it useless without the service? Seem like we should just move it into the same package as the service.

a-crate:
It's intended to activate the user's chosen syslog implementation (if any) [:link:](https://github.com/systemd/systemd/blob/a0ca911d633690a0c2097a4319daf1d3a803fb70/units/syslog.socket#L32). There's no applicable subpackage to move it to.

justinvreeland:
> It's intended to activate the user's chosen syslog implementation (if any) [🔗](https://github.com/systemd/systemd/blob/a0ca911d633690a0c2097a4319daf1d3a803fb70/units/syslog.socket#L32). There's no applicable subpackage to move it to.


Ohh nice I thought journald would've plugged into it but it looks like it sets up it's own thing.

justinvreeland (approved):

PR #49037, opened 2025-04-04
Yamale package addition

AnmolVirdi:
**Yamale** package `Current version: v.6.0.0`
- A schema and validator for YAML.

**Type:** Python
**REF:** https://github.com/23andMe/Yamale

**Tests:**
- schema tests
- Test using ruamel.yaml parser included.


Dentrax (changes requested):
Thanks, great tests! Let's support all python versions and add `provides: yamale` block.

yamale.yaml
```diff
@@ -0,0 +1,149 @@
...
+      expected-commit: c989fabfb2813885c5355f5f8bbf06ca02847e40
+
+subpackages:
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}
+    description: python${{range.key}} version of ${{vars.pypi-package}}
+    dependencies:
+      provider-priority: ${{range.value}}
+      provides:
+        - py3-${{vars.pypi-package}}
```
Dentrax, line 42: Should this also provide `yamale`? to make the `apk add yamale` work?

MickeyPvX: I was going to ask something similar - shouldn't this package be prefixed with `py3-` and the filename as well?

MickeyPvX: It can be used as a lib, which is why the binary needs to be split out.  Their docs show both usages: https://github.com/23andMe/Yamale?tab=readme-ov-file#api

AnmolVirdi: `py3-yamale` does not include `yamale` binary at this level, adding `yamale` in the providers list would function the same, which is not expected.

MickeyPvX: I would say yes - it aligns with all of our other python packages, which we prefix with `py3-`

Dentrax (approved):
LGTM!

PR #49098, opened 2025-04-04
Add libcurl-abi virtual package (take #2)

dannf:


curl.yaml
```diff
@@ -1,10 +1,16 @@
...
+  epoch: 3
   description: "URL retrieval utility and library"
   copyright:
     - license: MIT
+  dependencies:
+    runtime:
+      # Allow only libcurl libraries built from the same upstream
+      # source version instead of only relying on SONAMEs.
+      # https://github.com/chainguard-dev/internal-dev/issues/10381
+      - libcurl-abi=${{package.version}}
```
xnox, line 13: do you want to use

```
options:
  no-depends: true
```
+ specify all runtime dependencies
+ use test/ldd-check pipeline

Such that this package stops using `so:` dependencies?

dannf: I think we still want to use the `so:` dependencies, including the libcurl one. Here, we're adding an additional, tighter constraint.

javacruft (approved):

PR #49309, opened 2025-04-07
curl: bump

dannf:


justinvreeland:
Could you add a note about why it's being bumped? Is it related to the glibc stuff?

dannf:
> Could you add a note about why it's being bumped? Is it related to the glibc stuff?

Oh, sorry - yes, this was just to check what would happen if we rebuilt a random package against the new glibc but installed it in an image using the old glibc. Test ran successfully. I'll close this PR, no reason to merge.

Issue #42965, opened 2025-02-18
Strategy to stop providing unversioned `py3*` packages for install


justinvreeland:
Ideally the end user for those library requirements will specify the correct version python that they need and these packages should not be used.

To do this we need a way to warn someone that they're depending on an unversioned python for runtime.

cmwilson21:
Thanks for the contribution!

We’re closing this as part of a repo-wide reset to support upcoming changes in how contributions are validated and merged. You can find more context in our discussion post here: [Upcoming changes to wolfi-dev/os](https://github.com/orgs/wolfi-dev/discussions/77550).

If you’d like to continue with this work, please feel free to reopen or resubmit your issue or PR after reviewing the updated [Contribution Guidelines](https://github.com/wolfi-dev/os/blob/main/CONTRIBUTING.md).

We appreciate your understanding and your continued contributions to Wolfi.

PR #49321, opened 2025-04-07
py3-langsmith: new package

MickeyPvX:
Adding `py3-langsmith` and `py3-langsmith-pyo3` as dependencies to `langchain` (coming soon)

#### For new package PRs only


py3-langsmith.yaml
```diff
@@ -0,0 +1,120 @@
...
+      expected-commit: a25a1c4a2eb400cd5db43eb61e80a14a84170cad
+      repository: https://github.com/langchain-ai/langsmith-sdk
+      tag: v${{package.version}}
+
+subpackages:
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}
+    description: python${{range.key}} version of ${{vars.pypi-package}}
+    dependencies:
+      provider-priority: ${{range.value}}
```
justinvreeland, line 43: Hi, since this is a new package would you mind removing these provides probably also means removing the test from the top level packager I'd think.  We shouldn't have un-versioned python3 packages at all and the best way to catch them is when they're first added.

context: https://github.com/wolfi-dev/os/issues/42965

justinvreeland (changes requested):

py3-langsmith.yaml
MickeyPvX: So just dropping all these?
```yaml
dependencies:
  provides: py3-${{vars.pypi-package}}
```

I'll be neck deep in Python stuff for a while, so if there's any other quality guidelines I can help with, let me know.

justinvreeland: Yeap! the provider priorities can go with them too.

I think the biggest quality improvement going forward would be when you make the image or have runtime python deps it'd be ideal if you can specify the python version in the dependency otherwise packages or images can pickup multiple version of the python deps especially during python upgrades due to drift in the requirements. For example. something depending on packages like `py3-<module>` and getting py3.14 for some module packages and py3.13 other modules while py3.14 modules are being rebuilt.  Fortunately the subpackages in her are all fully qualified so if you do `py3.13-langsmith` it should resolve all the deps down to py3.13 correctly through the other multi-versioned python packages.

justinvreeland (approved):
Thanks!

PR #49343, opened 2025-04-07
linkerd-extension-init: use test/daemon-check-output pipeline

kranurag7:
passing with `make docker-test/linkerd-extension-init`


dnegreira:
TIL about `docker-test` thanks for the fix.

dnegreira (approved):

PR #49386, opened 2025-04-08
ceph: Add missing runtime dependencies

javacruft:
The Ceph OSD block device detection and initialization process uses a load of additional runtime tools; add these to the package so block devices are correctly detected.


utieyin (dismissed):
Looks good

javacruft:
As the PyO3 is current not tractable to fix I've removed the restful and cephadm modules from the installed package - these don't appear to be used by rook but we should add to compatibility notes where required to say these modules are not enable.

I'll raise an issue to track as well.

utieyin (approved):
Looks good

PR #49432, opened 2025-04-08
py3-jsonpatch: split jsonpatch binaries from jsonpatch lib / adding additional testing

MickeyPvX:
Splitting `jsonpatch` binaries from the lib, deconflicts situations like these: https://github.com/wolfi-dev/os/pull/49326/checks?check_run_id=40176300241

Packages that depend on `jsonpatch`:
- `ceph` - lib only, doesn't use binaries
- `cloud-init` - lib only, doesn't use binaries

### DIFF
--- a/py3-jsonpatch.yaml
+++ b/py3-jsonpatch.yaml
@@ -1,22 +1,28 @@
-# Generated from https://pypi.org/project/jsonpatch/
 package:
   name: py3-jsonpatch
-  version: "1.33"
-  epoch: 0
+  version: 1.33
+  epoch: 1
   description: Apply JSON-Patches (RFC 6902)
   copyright:
     - license: BSD-3-Clause
   dependencies:
     provider-priority: "0"
 
+vars:
+  pypi-package: jsonpatch
+
+data:
+  - name: py-versions
+    items:
+      3.10: "310"
+      3.11: "311"
+      3.12: "312"
+      3.13: "313"
+
 environment:
   contents:
     packages:
-      - build-base
-      - busybox
-      - ca-certificates-bundle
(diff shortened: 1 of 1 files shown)


py3-jsonpatch.yaml
```diff
@@ -33,27 +39,59 @@ subpackages:
...
-  - name: py-versions
-    items:
-      "3.10": "310"
-      "3.11": "311"
-      "3.12": "312"
-      "3.13": "313"
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}-bin
+    description: Executable binaries for ${{vars.pypi-package}} installed for python${{range.key}}
+    dependencies:
```
AnmolVirdi, line 61: Do we need to set `provides priority` and `provides` here?
Ref: https://github.com/wolfi-dev/os/blob/83b18dbbe03f156395d617ed0049ffde0f459467/py3-jupyterlab.yaml#L86-L120

MickeyPvX: Not anymore - @justinvreeland may want to elaborate, but we talked about it a bit on another PR: https://github.com/wolfi-dev/os/pull/49321#discussion_r2031925690

AnmolVirdi (approved):

justinvreeland (commented):
Ahhh you beat me to it this should look like this not what's here: https://github.com/wolfi-dev/os/pull/49461

justinvreeland (changes requested):
whoops

justinvreeland (approved):
Oh sorry, your linter making changes confused me a bit when I first looked this works!

PR #49465, opened 2025-04-08
resurrect some patches that were used after all

imjasonh:


lint.sh
```diff
@@ -42,3 +42,15 @@ for file in $(find .github/chainguard -type f); do
     exit 1
   fi
 done
+
+err=0
+for f in $(find . -type f -name "*.patch" -depth 2 ); do
+  if ! grep -q $(basename $f) $(dirname $f).yaml; then
+   echo "Found unused patch file: $f"
```
murraybd, line 49: Could you also check for the presence of a series file and if the .patch files are in that?

murraybd (changes requested):

lint.sh
imjasonh: Whooops, that wasn't even meant to be included. I'll make sure to do that, when/if I propose this in the future.

murraybd: The lint.sh file has now been deleted and I don't think that's what you wanted.

murraybd (approved):

PR #49470, opened 2025-04-08
py3-pybind11: Update patch to allow using Python3_add_library()

dannf:
Even after fixing the casing, I'm seeing a build failure because
Python_FOUND is set but Python_add_library() is not available.
Refactor the code to use Python3_add_library(), if available,
in that case.


sergiodj (approved):
LGTM, thanks Dann.

Do we want to submit this fix upstream?

dannf:
> LGTM, thanks Dann.
> 
> Do we want to submit this fix upstream?

I did :) https://github.com/pybind/pybind11/pull/5606

PR #49526, opened 2025-04-09
py3-kubernetes: support multiple Pythons

javacruft:
Add support for multiple Python versions.


javacruft:
I double checked that installing 'py3-kubernetes' results in the same actions before and after this update.

xnox (approved):

PR #49535, opened 2025-04-09
New package auditwheel

xnox:
Dependencies are already packaged.

Not really needed, but could be useful soon:
- auditwheel repair can be used to fixup wheels during collect-wheels pipeline
- auditwheel show is a useful debugging / test pipeline tool to determine compat

### DIFF
--- a/py3-auditwheel.yaml
+++ b/py3-auditwheel.yaml
@@ -1,0 +1,102 @@
+package:
+  name: py3-auditwheel
+  version: "6.3.0"
+  epoch: 1
+  description: auditing and relabeling of PEP 600 Linux wheels
+  copyright:
+    - license: MIT
+  dependencies:
+    provider-priority: 0
+
+vars:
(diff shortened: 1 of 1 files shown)


py3-auditwheel.yaml
```diff
@@ -0,0 +1,104 @@
...
+subpackages:
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}
+    description: ${{vars.pypi-package}} installed for python${{range.key}}
+    dependencies:
+      runtime:
+        - py${{range.key}}-packaging
+        - py${{range.key}}-pyelftools
+      provider-priority: ${{range.value}}
+      provides:
```
justinvreeland, line 50: We shouldn't be adding new version less python provides since it can cause mixed python versions during upgrades.

justinvreeland (dismissed):

kbsteere (approved):

PR #49540, opened 2025-04-09
feat: add apache-kvrocks package

wojciechka:
Fixes:

Related:

#### For new package PRs only


apache-kvrocks.yaml
```diff
@@ -0,0 +1,91 @@
...
+
+test:
+  environment:
+    contents:
+      packages:
+        - valkey-cli
+        - wait-for-port
+  pipeline:
+    - runs: |
+        kvrocks &
```
kranurag7, line 84: looks like here we can use test/daemon-check-output pipeline, see some logs in action and then do the valkey-cli invocations in post: block.

kranurag7 (approved):
default timeout is 30s seconds, looks good, thanks for using daemon-check-output pipeline.

PR #49620, opened 2025-04-10
Builds opencv libraries

srishtih:


opencv.yaml
```diff
@@ -0,0 +1,136 @@
...
+      opts: |
+        -DENABLE_BUILD_HARDENING=ON \
+        -DENABLE_THIN_LTO=ON \
+        -DWITH_OPENCL=ON \
+        -DWITH_OPENGL=ON \
+        -DWITH_TBB=ON \
+        -DWITH_VULKAN=ON \
+        -DWITH_QT=ON \
+        -DWITH_ADE=OFF \
+        -DWITH_opencv_gapi=OFF \
```
mritunjaysharma394, line 72: Probably, before anyone points in future, we can write a comment somewhere that Graph API volatile and we are disabling it 
https://docs.opencv.org/4.x/d0/d1e/gapi.html

opencv.yaml
```diff
@@ -0,0 +1,136 @@
+package:
+  name: opencv
+  version: 4.11.0
+  epoch: 3
```
mritunjaysharma394, line 4: we can set this to epoch to 0

opencv.yaml
```diff
@@ -0,0 +1,136 @@
...
+  environment:
+    contents:
+      packages:
+        - gcc
+        - opencv-dev
+        - pkgconf
+  pipeline:
+    - name: Verify opencv installation, please improve the test as needed
+      runs: |
+        g++ test_opencv.cpp `pkg-config --cflags --libs opencv4` -o test_opencv
```
mritunjaysharma394, line 129: Thank you for the great test

opencv.yaml
```diff
@@ -0,0 +1,136 @@
...
+    pipeline:
+      - uses: split/dev
+
+  - range: libs
+    name: libopencv-${{range.key}}
+    description: opencv's ${{range.value}} library
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/usr/lib
+          mv ${{targets.destdir}}/usr/lib/libopencv_${{range.key}}.so.* ${{targets.subpkgdir}}/usr/lib
```
mritunjaysharma394, line 117: Can we have some of sub package based test to ensure that these libs exist in the expected directory, if it’s too complex due to range thing, then we can discuss alternatives but it will be good to have it

jamonation (approved):
LGTM

mritunjaysharma394 (approved):

PR #49664, opened 2025-04-10
package update: Bump to stable v3.0.0 and enable automated updates

debasishbsws:
Previously, we were building the v3 beta-1 version due to issues in the v2 release, as discussed here: https://github.com/distribution/distribution/discussions/4404.

Now that the upstream has released the stable v3.0.0 version, we are:
- Updating the package to version 3.0.0
- Removing `manual: true` to re-enable automated updates
- Withdrawing all non-stable (beta) versions

This aligns us with the latest upstream stable release and simplifies future maintenance.


debasishbsws:
Ok the build will not gonna work as we already have this version

debasishbsws:
We should not withdrawn all the older package

AmberArcadia (approved):

PR #49673, opened 2025-04-10
WIP: ldconfig.trigger test

xnox:


glibc/ldconfig-Add-ldconfig.trigger-no-argument-execution-.patch
```diff
@@ -0,0 +1,46 @@
+From fb07152b0e2065b9c5607eaa6013a8d31a30a481 Mon Sep 17 00:00:00 2001
+From: Dimitri John Ledkov <dimitri.ledkov@surgut.co.uk>
+Date: Tue, 8 Apr 2025 12:04:24 +0100
+Subject: [PATCH] ldconfig: Add ldconfig.trigger no argument execution mode
+
+Typically package management invocation passes in additional arguments
+(e.g. "postinst" "postinstall" "post-commit") to mainstainer scripts,
+hooks, scriplets, triggers. And to invoke ldconfig without any
+arguments, one typically needs to create a shell script calls ldconfig
+without arguments.
```
smoser, line 10: This sentence does not make a lot of sense to me.
> "to invoke ldconfig without any arguments, one typically needs to create a shell script [that] calls ldconfig without arguments".

Why do you need a shell script to do that? versus changing whatever _was_ calling 'ldconfig arg1...' to call 'ldconfig'?

smoser: ah.... because the distro moved ldconfig to ldconfig.real and then put a shell script to call ldconfig.real in /usr/bin/ldconfig.

ie, /usr/sbin/ldconfig says (on ubuntu), but _not_ on wolfi:
```bash
#!/bin/sh

if  test $# = 0							\
    && test x"$LDCONFIG_NOTRIGGER" = x				\
 && test x"$DPKG_MAINTSCRIPT_PACKAGE" != x			\
 && dpkg-trigger --check-supported 2>/dev/null
then
	if dpkg-trigger --no-await ldconfig; then
		if test x"$LDCONFIG_TRIGGER_DEBUG" != x; then
			echo "ldconfig: wrapper deferring update (trigger activated)"
		fi
		exit 0
	fi	
fi

exec /sbin/ldconfig.real "$@"
```

But since wolfi does _not_ do that, why would I not just invoke 'ldconfig' or invoke ldconfig.real ?

smoser: > Thus instead of symlinking ldconfig to the commit_hooks.d directory i need to do #!/bin/sh ldconfig instead, cause that works.

this didn't work in my testing, see below.  Calling 'my-ldconfig' with parameter 'pre-commit' gets the parameter passed through to 'my-ldconfig.real'.

```
$ cat /tmp/my-ldconfig
#!/bin/sh /tmp/my-ldconfig.real

$ cat /tmp/my-ldconfig.real
#!/bin/sh
echo "$0 got $#: $*"

$ /tmp/my-ldconfig pre-commit
/tmp/my-ldconfig.real got 2: /tmp/my-ldconfig pre-commit
```

Thanks for the explanation. the /etc/apk/commit_hooks.d was the bit i was missing.
I understand the problem better.

this  is an odd solution, but i don't have a better one.

My only remaining comment is that 'ldconfig.trigger' is a weird name to mean 'ignore command line arguments'

xnox: > this didn't work in my testing

well, you symlinked a wrapper script =) rather than a real ldconfig binary as shipped by glibc upstream. I was talking through the point of view of upstream ldconfig binary as shipped by upstream, wolfi, and most distributions.

ldconfig.real and wrapper was precisely done because debian wanted to move to triggers and couldn't agree to make all maintainers to stop calling it randomly in their maintainer scripts.

xnox: Note trigger is typically a file-path based trigger that apk-tools, rpm and dpkg support, thus the name was supposed to be intuitive for distribution package maintainers that are very familiar with their respective packaging trigger implementations =/

PR #49709, opened 2025-04-10
npm: add test for issue that should block updates, re-enable updates

OddBloke:
Specifically https://github.com/npm/cli/issues/8216 breaks using `npm`
against a registry hosted at any path but `/`.  The test added here
passes on 11.2.0, but fails on a (locally-built) 11.3.0.
    
As this will result in package updates failing until upstream release a
fix for the issue, we can re-enable automated updates.  We won't have
to come back later and re-enable them manually, and the first fixed
version should land automatically.

### DIFF
--- a/npm.yaml
+++ b/npm.yaml
@@ -2,7 +2,7 @@
 package:
   name: npm
   version: "11.2.0"
-  epoch: 1
+  epoch: 2
   description: "the npm package manager for javascript, mainline"
   copyright:
     - license: Artistic-2.0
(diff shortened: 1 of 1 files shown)


OddBloke:
This test correctly failed for 11.3.0 [here](https://github.com/wolfi-dev/os/pull/49709/checks?check_run_id=40335836390), and is passing for 11.2.0 in this PR. This is ready to land.

kbsteere (approved):

PR #49716, opened 2025-04-10
feat: add systemd-repart for ext4

89luca89:
So we can choose which one

### DIFF
--- a/systemd.yaml
+++ b/systemd.yaml
@@ -1,7 +1,7 @@
 package:
   name: systemd
   version: "257.5"
-  epoch: 0
+  epoch: 1
   description: The systemd System and Service Manager
   copyright:
     - license: LGPL-2.1-or-later AND GPL-2.0-or-later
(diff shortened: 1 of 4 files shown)


systemd.yaml
```diff
@@ -558,8 +558,8 @@ subpackages:
             systemd-firstboot --version
             systemd-firstboot --help
 
-  - name: "systemd-repart-rootfs"
-    description: "Config file to auto grow root filesystem"
+  - name: "systemd-repart-rootfs-xfs"
```
EyeCantCU, line 561: Non-blocking, but we could use a range here if we plan to support other filesystems outside of ext4 and xfs

EyeCantCU (dismissed):

systemd.yaml
89luca89: Right! Done that now, thanks :)

EyeCantCU (approved):

PR #49765, opened 2025-04-10
Fix the python shbang for oscap-docker which has /usr/sbin/python3

murraybd:
The oscap-docker python script has `/usr/sbin/python3` in its shbang line so the fix shbang step needs to be expanded to include sbin. I opted to do this as a second line to ease readability.

The symlink was only causing the test to pass so drop that.

### DIFF
--- a/openscap.yaml
+++ b/openscap.yaml
@@ -1,7 +1,7 @@
 package:
   name: openscap
   version: "1.4.2"
-  epoch: 0
+  epoch: 1
   description: NIST Certified SCAP 1.2 toolkit
   copyright:
     - license: LGPL-2.1-or-later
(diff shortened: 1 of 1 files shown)


dannf (approved):
LGTM. Non-blocking suggestions follow...

I didn't follow why it is now /usr/sbin instead of /usr/bin - something usrmerge related?

If it's likely to be inconsistent going forward, maybe something like:

 sed -i '1s,^#!.*/python3$,#!/usr/bin/env python${{vars.py-version}},' "$f"

Alternatively, we could add a check for the shebang before we modify it, so the build fails if it turns out our `sed` is going to be a no-op:

head -1 "$f" | grep '^#!/usr/sbin/python3$'

PR #49777, opened 2025-04-10
iptables - split iptables-apply out to a subpackage.

smoser:
iptables-apply is a bash script (per its #!).

As an unintended side effect of usrmerge, melange began noticing this recently and now correctly added the dependency on bash to iptables.

Many users will not use iptables-apply and would rather not have the unused depends on bash.  Evidence for this is the fact that melange previously did not identify the dependency and no one screamed.

The solution here is to just split out iptables into its own package. If a user needs it, they can get it there.

### DIFF
--- a/iptables.yaml
+++ b/iptables.yaml
@@ -1,7 +1,7 @@
 package:
   name: iptables
   version: 1.8.11
-  epoch: 21
+  epoch: 22
   description: Linux kernel firewall, NAT and packet mangling tools
   copyright:
     - license: GPL-2.0-or-later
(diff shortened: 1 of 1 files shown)


smoser:
this _will_ break people that are using iptables-save. 
those people already were broken, but in a different way - they had to `apk add bash` to make it work.

stevebeattie (approved):
This LGTM, and confirmed that it causes the new dependency on bash in the main `iptables` package to be dropped. The only subpackage that depends on bash is `iptables-xtables-privileged` and that dependency existed prior to the sbin-merge change (i.e. it existed in `iptables-xtables-privileged=1.8.11-r0`).

This change does keep the copy of `iptables-apply` in the `iptables-xtables-privileged` package.

PR #49838, opened 2025-04-10
bug: fix systemd-logind-service subpackage

zmarano:


systemd.yaml
```diff
@@ -512,11 +512,11 @@ subpackages:
         - wolfi-baselayout
     pipeline:
```
a-crate, line 513: Could use `test/verify-service` in the test pipeline to avoid this

zmarano: Hmm not sure it will work correctly. Because the subpackage name is `systemd-logind-service `and not `systemd-logind`- let's find out though.

zmarano: Yeah - seems to work right. Thanks!

a-crate (approved):

PR #49988, opened 2025-04-11
adding mediainfo package, this is commonly available in debian and us…

pete-the-great:
Fixes:

Related:

#### For new package PRs only


mediainfo.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+pipeline:
+  - name: Clone ZenLib
+    runs: |
+      git clone https://github.com/MediaArea/ZenLib.git
+      cd ZenLib/Project/GNU/Library
+      ./autogen.sh
+      ./configure --prefix=/usr
+      make
+      make DESTDIR="${{targets.destdir}}" install
```
dannf, line 33: please use pipelines where possible. here we could be using `git-checkout` and `autoconf/{configure,make,make-install}`. Instead of cd'ing, you can use `working-directory`.

mediainfo.yaml
```diff
@@ -0,0 +1,74 @@
...
+      - automake
+      - libtool
+      - pkgconf
+      - pkgconf-dev
+      - glibc
+      - wolfi-baselayout
+      - zlib-dev
+
+pipeline:
+  - name: Clone ZenLib
```
dannf, line 26: you're doing more than cloning here :)

mediainfo.yaml
```diff
@@ -0,0 +1,74 @@
+package:
+  name: mediainfo
+  version: 25.03
```
dannf, line 3: should zenlib and mediainfo lib be separate packages, so they can each be part of our update automation?

mediainfo.yaml
```diff
@@ -0,0 +1,74 @@
...
+      - pkgconf
+      - pkgconf-dev
+      - glibc
+      - wolfi-baselayout
+      - zlib-dev
+
+pipeline:
+  - name: Clone ZenLib
+    runs: |
+      git clone https://github.com/MediaArea/ZenLib.git
```
dannf, line 28: We should pin to a tag and check against a known commit, otherwise we're subject to random changes on this tree, which has a security risk if this repo were to become compromised.

dannf (changes requested):
Thanks for the contribution! Comments inline.

mediainfo.yaml
pete-the-great: updated

mediainfo.yaml
pete-the-great: updated comments

mediainfo.yaml
pete-the-great: separated into mediainfolib / libzen

mediainfo.yaml
pete-the-great: fixed against the known version from github with hash

mediainfo.yaml
dannf: Thanks for switching to the `git-checkout` pipeline. Could you also use the `autoconf/*` pipelines?

libzen.yaml
```diff
@@ -0,0 +1,55 @@
...
+    working-directory: Project/GNU/Library
+
+subpackages:
+  - name: "libzen-dev"
+    description: "Development files for libzen"
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.destdir}}/usr/share/doc/libzen"
+          cp -r ZenLib/Documentation/* "${{targets.destdir}}/usr/share/doc/libzen/" || true
+        working-directory: ZenLib/Project/GNU/Library
```
dannf, line 50: This looks like a pipeline for installing a docs package, not a -dev package. For the `-dev`,  could you use the `split/dev` pipeline?

mediainfo.yaml
```diff
@@ -0,0 +1,63 @@
...
+  description: "Utility for displaying technical and tag data for media files"
+  copyright:
+    - license: BSD-3-Clause
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates
+      - git
```
dannf, line 14: not necessary since you're using the `git-checkout` pipeline

mediainfo.yaml
```diff
@@ -0,0 +1,63 @@
...
+      - build-base
+      - autoconf
+      - automake
+      - libtool
+      - pkgconf
+      - pkgconf-dev
+      - glibc
+      - wolfi-baselayout
+      - zlib-dev
+      - libzen
```
dannf, line 24: i suspect you'll really need `libzen-dev` once you `split/dev` that one

mediainfo.yaml
```diff
@@ -0,0 +1,63 @@
...
+      - busybox
+      - ca-certificates
+      - git
+      - build-base
+      - autoconf
+      - automake
+      - libtool
+      - pkgconf
+      - pkgconf-dev
+      - glibc
```
dannf, line 21: I wouldn't explicitly build-dep on glibc. The things that need it will already depend on it.

mediainfo.yaml
```diff
@@ -0,0 +1,63 @@
...
+    uses: git-checkout
+    with:
+      repository: https://github.com/MediaArea/MediaInfo.git
+      tag: "v${{package.version}}"
+      expected-commit: 06f1b659f57c081fabdc91d3f3546b73206e171c
+
+  - name: Autogen and Install MediaInfo
+    runs: |
+      pwd
+      ls -la
```
dannf, line 38: I suggest dropping this debug code

mediainfo.yaml
pete-the-great: everything works with libzen package

pete-the-great:
made requested changes with cleanup and pipelines

mediainfo.yaml
```diff
@@ -0,0 +1,65 @@
...
+  name: mediainfo
+  version: 25.03
+  epoch: 0
+  description: "Utility for displaying technical and tag data for media files"
+  copyright:
+    - license: BSD-3-Clause
+  dependencies:
+    runtime:
+      - libzen
+      - mediainfolib
```
dannf, line 11: these dependencies should be automatically detected and added by melange, and that's preferred. if they are not being automatically added, we should figure out why.

libzen.yaml
```diff
@@ -0,0 +1,54 @@
+package:
+  name: libzen
```
dannf, line 2: This includes header files:

https://apk.chaindag.dev/https/apk.cgr.dev/wolfi-presubmit/733c11d163a1493486f396aff2ee5c1d4a014cc1/x86_64/libzen-0.4.41-r0.apk@etag:e7e1936dae8c53fc0f9dfb7c71c2d721/usr/include/ZenLib/

Those should be in a `libzen-dev` package. See the `split/dev` pipeline. Packages that build against libzen should then depend on `libzen-dev` instead of `libzen`.

mediainfolib.yaml
```diff
@@ -0,0 +1,57 @@
+package:
+  name: mediainfolib
```
dannf, line 2: similar to libzen, this should also have a `-dev` package

dannf (changes requested):
almost there!

mediainfolib.yaml
```diff
@@ -0,0 +1,57 @@
...
+environment:
+  contents:
+    packages:
+      - busybox
+      - ca-certificates
+      - git
+      - build-base
+      - libtool
+      - pkgconf
+      - pkgconf-dev
```
dannf, line 18: I'm not sure why you'd need `pkgconf-dev` - isn't that just for building tools against `libpkgconf`?

mediainfo.yaml
pg-cgr: cleaned these up, as they are not needed.

mediainfolib.yaml
pg-cgr: fixed this, and fixed deps in mediainfo package.

dannf (changes requested):
Awesome! Just one last thing. I wondered why I hadn't noticed `autogen` in other YAML files, so I took a look at the melange pipelines:

https://github.com/chainguard-dev/melange/blob/main/pkg/build/pipelines/autoconf/configure.yaml

aha! it does an `autoreconf` for us. So I think you can just remove that pipeline step.

EDIT: I see that we do run `./autogen.sh` in a lot of places actually, I just hadn't noticed. I don't know why. If you believe the `autogen.sh` step is needed, I'm happy to approve this as-is.

pete-the-great:
> Awesome! Just one last thing. I wondered why I hadn't noticed `autogen` in other YAML files, so I took a look at the melange pipelines:
> 
> https://github.com/chainguard-dev/melange/blob/main/pkg/build/pipelines/autoconf/configure.yaml
> 
> aha! it does an `autoreconf` for us. So I think you can just remove that pipeline step.
> 
> EDIT: I see that we do run `./autogen.sh` in a lot of places actually, I just hadn't noticed. I don't know why. If you believe the `autogen.sh` step is needed, I'm happy to approve this as-is.

Indeed, this is an unneeded step. It will be removed.

mediainfolib.yaml
```diff
@@ -26,10 +26,6 @@ pipeline:
       expected-commit: b1369cf34d6567aceac4f0c4208bc5caa0d4d2f7
 
   - name: Build and Install MediaInfoLib
```
dannf, line 28: mind removing this line as well ^

pete-the-great: removed unneeded name fields in pipelines

dannf: Sorry, I wasn't complaining about the `name` field in general. It just appeared as though you had an orphaned `-name` field for a step that you'd removed.

dannf (dismissed):

dannf (approved):

PR #50008, opened 2025-04-11
ceph: add more missing dependencies

javacruft:
- **ceph: add py3-scikit-learn to runtime deps**
- **ceph: drop runtime dependency on ceph from -dev.**
- **ceph: add runtime depends on smartmontools and sgdisk.**


dannf (approved):
LGTM

dannf:
While this clearly didn't introduce them, the cve scan did detect issues. Is there a plan for those?

PR #50242, opened 2025-04-14
add more tests to grpc-health-probe package

kranurag7:
mostly using openfga to mock a grpc health endpoint


Aditevil (approved):
Nice work!

PR #50254, opened 2025-04-14
Trivy operator

rahulduvedi:
trivy-operator-0.25.0: new package

### DIFF
--- a/py3-pystache.yaml
+++ b/py3-pystache.yaml
@@ -1,0 +1,92 @@
+package:
+  name: py3-pystache
+  version: "0.6.8"
+  epoch: 0
+  description: "A mustache template renderer for Python"
+  copyright:
+    - license: MIT
+
+vars:
+  pypi-package: pystache
+
+data:
+  - name: py-versions
+    items:
+      3.10: "310"
+      3.11: "311"
+      3.12: "312"
+      3.13: "313"
+
+environment:
+  contents:
+    packages:
+      - py3-supported-build-base-dev
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/PennyDreadfulMTG/pystache.git
+      tag: v${{package.version}}
+      expected-commit: 531a0ecea00a696423175ef46c319b29408dc3e2
+
+subpackages:
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}
+    description: python${{range.key}} version of ${{vars.pypi-package}}
+    dependencies:
+      provides:
+        - py3-${{vars.pypi-package}}
+    pipeline:
+      - uses: py/pip-build-install
+        with:
+          python: python${{range.key}}
+      - uses: strip
+    test:
+      environment:
+        contents:
+          packages:
+            - python-${{range.key}}
+      pipeline:
+        - uses: python/import
+          with:
+            python: python${{range.key}}
+            import: ${{vars.pypi-package}}
+        - runs: |
+            # Test the package
+            cat <<EOF > test.py
+            import pystache
+            print(pystache.render('Hi {{person}}!', {'person': 'World'}))
+            EOF
+            python test.py | grep -q 'Hi World!'
+
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}-bin
+    description: Executable binaries for ${{vars.pypi-package}} installed for python${{range.key}}
+    dependencies:
+      runtime:
+        - py${{range.key}}-${{vars.pypi-package}}
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/usr
+          mv ${{targets.contextdir}}/../py${{range.key}}-${{vars.pypi-package}}/usr/bin ${{targets.contextdir}}/usr
+    test:
+      pipeline:
+        - runs: |
+            pystache -h | grep "Usage"
+
+  - name: py3-supported-${{vars.pypi-package}}
+    description: meta package providing ${{vars.pypi-package}} for supported python versions.
+    dependencies:
+      runtime:
+        - py3.10-${{vars.pypi-package}}
+        - py3.11-${{vars.pypi-package}}
+        - py3.12-${{vars.pypi-package}}
+        - py3.13-${{vars.pypi-package}}
+
+update:
+  enabled: true
+  manual: false
+  github:
+    identifier: PennyDreadfulMTG/pystache
+    strip-prefix: v
+    use-tag: true


trivy-operator.yaml
```diff
@@ -0,0 +1,77 @@
...
+          echo "Error: trivy-operator binary not found or not executable"
+          exit 1
+        fi
+
+        if ! /usr/bin/trivy-operator version 2>&1 | grep -q "Starting operator"; then
+          echo "Error: Binary doesn't produce expected startup message"
+          exit 1
+        fi
+
+        echo "Basic binary verification passed"
```
kranurag7, line 68: we can simplify this by doing `/usr/bin/trivy-operator version 2>&1 | grep -q "Starting operator`

trivy-operator.yaml
```diff
@@ -0,0 +1,77 @@
...
+  version: "0.25.0"
+  epoch: 1
+  description: "Kubernetes-native security toolkit that finds and reports vulnerabilities and misconfigurations"
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - ca-certificates
+      - trivy
+      - tzdata
```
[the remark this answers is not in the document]
kranurag7: we can omit all three here.

trivy-operator.yaml
```diff
@@ -0,0 +1,70 @@
...
+        - ${{package.name}}
+  pipeline:
+    - name: "Basic binary verification"
+      runs: |
+        if [ ! -x /usr/bin/trivy-operator ]; then
+          echo "Error: trivy-operator binary not found or not executable"
+          exit 1
+        fi
+
+        # Check if binary produces expected startup message (script exits if grep fails)
```
kranurag7, line 58: can be omitted, given the line below will error out if binary is missing.

trivy-operator.yaml
```diff
@@ -0,0 +1,70 @@
...
+  github:
+    identifier: aquasecurity/trivy-operator
+    strip-prefix: v
+
+test:
+  environment:
+    environment:
+      OPERATOR_NAMESPACE: "trivy-system"
+      KUBERNETES_SERVICE_HOST: "127.0.0.1"
+      KUBERNETES_SERVICE_PORT: "32764"
```
kranurag7, line 46: these two environment variables should not be required, `make docker-test/trivy-operator`

trivy-operator.yaml
```diff
@@ -0,0 +1,69 @@
...
+        - ${{package.name}}
+  pipeline:
+    - name: "Basic binary verification"
+      runs: |
+        if [ ! -x /usr/bin/trivy-operator ]; then
+          echo "Error: trivy-operator binary not found or not executable"
+          exit 1
+        fi
+
+        # Check if binary produces expected startup message (script exits if grep fails)
```
kranurag7, line 57: I think we can consider removing this. 

If the binary is not found then `/usr/bin/trivy-operator version 2>&1 | grep -q "Starting operator"` command will fail with `No such file or directory`

AnmolVirdi (changes requested):

trivy-operator.yaml
```diff
@@ -0,0 +1,69 @@
...
+        github.com/golang-jwt/jwt/v4@v4.5.2
+        github.com/golang-jwt/jwt/v5@v5.2.2
+        golang.org/x/oauth2@v0.27.0
+        helm.sh/helm/v3@v3.17.3
+        golang.org/x/net@v0.38.0
+
+  - uses: go/build
+    with:
+      packages: "./cmd/trivy-operator"
+      output: trivy-operator
```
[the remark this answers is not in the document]
rahulduvedi: No need to set modroot: . explicitly—Melange’s Go build plugin defaults to your repo root (where go.mod lives). Only override modroot if your module root is in a different subdirectory.
@AnmolVirdi

trivy-operator.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+        # Check if binary produces expected startup message (script exits if grep fails)
+        /usr/bin/trivy-operator version 2>&1 | grep -q "Starting operator"
+
+        echo "Basic binary verification passed"
+    - uses: test/kwok/cluster
+    - uses: test/daemon-check-output
+      with:
+        timeout: 60
+        start: /usr/bin/trivy-operator
```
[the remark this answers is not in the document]
rahulduvedi: IMO it should be `/usr/bin/trivy-operator` as it is the default location where it is installed by melange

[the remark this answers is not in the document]
AnmolVirdi: Both are same, by default all binaries are installed in `/usr/bin` only, and they are all accessible through default root dir since `/usr/bin` is in `PATH`.

trivy-operator.yaml
```diff
@@ -0,0 +1,69 @@
...
+        expected_output: |
+          Starting operator
+          Resolved install mode
+          Watching all namespaces
```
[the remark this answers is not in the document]
rahulduvedi: I think the output is a good enough test to check if the binary is building up correctly or not as this may require us to install some CRDs which is not possible with out current setup sadly.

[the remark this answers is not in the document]
AnmolVirdi: Use `uses: test/kwok/cluster` and it would add cluster to your melange test environment

trivy-operator.yaml
```diff
@@ -0,0 +1,96 @@
...
+            --from-literal=trivy.insecure=false \
+            --from-literal=trivy.offlineScan=false \
+            --from-literal=trivy.securityChecks=vuln,secret \
+            --from-literal=trivy.severity=UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL \
+            --from-literal=trivy.timeout=5m0s
+        start: /usr/bin/trivy-operator
+        expected_output: |
+          Starting operator
+          Resolved install mode
+          Watching all namespaces
```
AnmolVirdi, line 96: This deployment test is incomplete. 

Re-sharing the conversation with context here: https://github.com/wolfi-dev/os/pull/50254#discussion_r2052101768

We have a cluster set up using `uses: test/kwok/cluster` . I don't see any problem in adding CRDs to the setup.

trivy-operator.yaml
```diff
@@ -0,0 +1,96 @@
...
+          kubectl create configmap trivy-operator -n trivy-system \
+            --from-literal=trivy.ignoreUnfixed=true \
+            --from-literal=trivy.skipFiles=/etc/ssl/certs/ca-certificates.crt \
+            --from-literal=trivy.skipDirs=/usr/lib/ruby/gems/2.7.0/cache \
+            --from-literal=trivy.insecure=false \
+            --from-literal=trivy.offlineScan=false \
+            --from-literal=trivy.securityChecks=vuln,secret \
+            --from-literal=trivy.severity=UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL \
+            --from-literal=trivy.timeout=5m0s
+        start: /usr/bin/trivy-operator
```
AnmolVirdi, line 92:
```suggestion
        start: trivy-operator
```
Upstream image contents:
```
            "WorkingDir": "/",
            "Entrypoint": [
                "trivy-operator"
            ]
```
More context here: https://github.com/wolfi-dev/os/pull/50254/files#r2068524111

AnmolVirdi (changes requested):

trivy-operator.yaml
rahulduvedi: Under the hood, Melange’s go/build step puts your compiled binary into usr/bin/<output> by default (that is, it uses prefix=usr and install-dir=bin), so our trivy-operator lands at /usr/bin/trivy-operator. When you run melange test, the harness drops you into a shell whose $PATH already includes /usr/bin, which means typing trivy-operator automatically runs the /usr/bin version, so we can write anything that makes the tests clearer,

check out the go/build pipeline definition here for more understanding.
https://github.com/chainguard-dev/melange/blob/main/pkg/build/pipelines/go/build.yaml

trivy-operator.yaml
```diff
@@ -0,0 +1,96 @@
...
+        # Check if binary produces expected startup message (script exits if grep fails)
+        trivy-operator version 2>&1 | grep -q "Starting operator"
+
+        # Verify package version in ldflags
+        if ! strings /usr/bin/trivy-operator | grep -q "version=${{package.version}}"; then
+          echo "Error: Package version ${{package.version}} not found in binary"
+          exit 1
+        fi
+
+        echo "Basic binary verification passed"
```
Dentrax, line 73: This binary doesn't have any [subcmds](https://github.com/aquasecurity/trivy-operator/blob/main/cmd/trivy-operator/main.go) at all, let's omit this test as we already cover below.
```suggestion
```

trivy-operator.yaml
AnmolVirdi: @rahulduvedi  I'm aware and already mentioned the same here: https://github.com/wolfi-dev/os/pull/50254/files#r2068524111

It's generally recommended to go match tests for upstream image. I'm fine with either approach.

trivy-operator.yaml
```diff
@@ -0,0 +1,69 @@
...
+  pipeline:
+    - name: "Basic binary verification"
+      runs: |
+        if [ ! -x /usr/bin/trivy-operator ]; then
+          echo "Error: trivy-operator binary not found or not executable"
+          exit 1
+        fi
+
+        # Check if binary produces expected startup message (script exits if grep fails)
+        /usr/bin/trivy-operator version 2>&1 | grep -q "Starting operator"
```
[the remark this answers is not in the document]
rahulduvedi: There is no need, there is no version subcmd in trivy operator.
Resolving this for now

PR #50306, opened 2025-04-14
py3-faiss-cpu | faiss: new packages

MickeyPvX:
Adds `faiss` package and the Python package for `faiss-cpu`.

Dependency of `py3-jupyter-ai`

#### For new package PRs only


py3-faiss-cpu.yaml
```diff
@@ -0,0 +1,114 @@
...
+            if [ ! -d "sift1M" ]
+            then
+              curl -O ftp://ftp.irisa.fr/local/texmex/corpus/sift.tar.gz
+              echo "b23d1b3b2ee8469d819b61ca900ef0ed  sift.tar.gz" > MD5.txt
+
+              md5sum -c MD5.txt
+
+              mkdir -p sift1M
+              tar -xzf sift.tar.gz -C sift1M --strip-components=1
+            fi
```
Dentrax, line 90: Let's use `fetch` pipeline here.
```suggestion
```

faiss.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - uses: cmake/build
+        working-directory: ${{package.name}}
+      - uses: cmake/install
+        working-directory: ${{package.name}}
+      - uses: strip
+    test:
+      pipeline:
+        - uses: test/pkgconf
+        - uses: test/tw/ldd-check
+
```
Dentrax, line 48: Let's have simple test here, if possible.

faiss.yaml
```diff
@@ -0,0 +1,54 @@
...
+          opts: |
+            -DBUILD_SHARED_LIBS=ON \
+            -DBUILD_TESTING=OFF \
+            -DFAISS_ENABLE_C_API=ON \
+            -DFAISS_ENABLE_GPU=OFF \
+            -DFAISS_ENABLE_PYTHON=OFF
+      - uses: cmake/build
+        working-directory: ${{package.name}}
+      - uses: cmake/install
+        working-directory: ${{package.name}}
```
Dentrax, line 42: nit: not-a-blocker: you should be able to do something like:
```suggestion
      - working-directory: ${{package.name}}
        pipeline:
        - uses: cmake/configure
          with:
            opts: |
              -DBUILD_SHARED_LIBS=ON \
              -DBUILD_TESTING=OFF \
              -DFAISS_ENABLE_C_API=ON \
              -DFAISS_ENABLE_GPU=OFF \
              -DFAISS_ENABLE_PYTHON=OFF
      - uses: cmake/build
      - uses: cmake/install
```

Dentrax (changes requested):

py3-faiss-cpu.yaml
MickeyPvX: I keep forgetting about that one 😅

faiss.yaml
MickeyPvX: Not sure this is practical - the `py3-faiss-cpu` package (which is tested by running some sample code) is what's building the wheel and installing it, this package is just providing the underlying libraries.

Dentrax (dismissed):

faiss.yaml
MickeyPvX: The `faiss` package doesn't provide a binary - just a shared lib, so we can't even do that 🤷

jamonation (approved):

PR #50395, opened 2025-04-14
Steampipe package addition (turbot/steampipe)

AnmolVirdi:
**Steampipe** `Current version: 1.1.0`
- Steampipe is the zero-ETL way to query APIs and services, used to expose data sources to SQL.

**Type:** Go Binary
**REF:** [https://github.com/turbot/steampipe](https://github.com/turbot/steampipe)

**Tests:**
- Complete test suite including core plugin and query funtionalities.

### DIFF
--- a/steampipe.yaml
+++ b/steampipe.yaml
@@ -1,7 +1,7 @@
 package:
   name: steampipe
   version: 1.1.0
-  epoch: 0
+  epoch: 6
   description: Steampipe is the zero-ETL way to query APIs and services, used to expose data sources to SQL.
   dependencies:
     runtime:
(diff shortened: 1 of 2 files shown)


steampipe.yaml
```diff
@@ -0,0 +1,93 @@
+package:
+  name: steampipe
+  version: 1.1.0
+  epoch: 0
+  description: Steampipe is the zero-ETL way to query APIs and services, used to expose data sources to SQL.
+  dependencies:
+    runtime:
+      - bash
+      - ca-certificates-bundle
+      - gcc
```
kranurag7, line 10: gcc is not required as a runtime dependency here.

kranurag7 (commented):
this as of now is missing static assets for UI inside /home/steampipe. Is that not required for this package?

steampipe.yaml
AnmolVirdi: steampipe installs and sets up postges, fdw as a part of initial setup.

AnmolVirdi:
@kranurag7 the UI element of steampipe was moved to turbot/powerpipe. As of now, steampipe server just launches local postgres service that listens on 9193 by default.

kranurag7 (approved):

kranurag7:
the scripts omits the chown part but that'll be handled at image level. The package doesn't embed UI as of now and that'll be done in a separate package.

PR #50413, opened 2025-04-15
build gdal with numpy bindings

kranurag7:
we always want to build using numpy bindings else we run into errors like following:

```bash
ImportError: cannot import name ‘gdal_array’ from ‘osgeo’
```


mritunjaysharma394 (approved):
thanks!

PR #50425, opened 2025-04-15
add more defaults to test/kwok/cluster pipeline

kranurag7:
we duplicate these parameters over and over again in package tests. Adding these here given these are good defaults which always remains the same.

some packages where these are getting repeated are:
```bash
$ rg 'KUBERNETES_SERVICE_PORT' -l
mongodb-kubernetes-operator.yaml
kube-state-metrics.yaml
cluster-proportional-autoscaler.yaml
sealed-secrets.yaml
kyverno-1.13.yaml
flux-source-controller.yaml
lvm-driver.yaml
ruby3.3-fluentd-kubernetes-daemonset-1.18.yaml
kubernetes-event-exporter.yaml
ruby3.2-fluentd-kubernetes-daemonset-1.18.yaml
ruby3.1-fluentd-kubernetes-daemonset-1.18.yaml
redis-operator.yaml
ruby3.4-fluentd-kubernetes-daemonset-1.18.yaml
metrics-server.yaml
```

### DIFF
--- a/pipelines/test/kwok/cluster.yaml
+++ b/pipelines/test/kwok/cluster.yaml
@@ -8,6 +8,12 @@ needs:
     - kubernetes # has a runtime dependency on kubectl
     - etcd
 
+inputs:
+  namespace:
+    description: namespace in which we want to run the operator/controller/manager
+    required: true
+    default: default
+
 pipeline:
   - runs: |
       # ensure we can create a cluster using our own components instead of
@@ -45,3 +51,11 @@ pipeline:
 
       kubectl wait --for=condition=Ready nodes --all
       kubectl cluster-info
+      export KUBERNETES_SERVICE_HOST="127.0.0.1"
+      export KUBERNETES_SERVICE_PORT="32764"
+      mkdir -p /var/run/secrets/kubernetes.io/serviceaccount
+      CA=$(kubectl config view --raw -o jsonpath='{.clusters[0].cluster.certificate-authority}')
+      cp $CA /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+      echo ${{inputs.namespace}} > /var/run/secrets/kubernetes.io/serviceaccount/namespace
+      kubectl create sa ${{inputs.namespace}}
+      kubectl create token ${{inputs.namespace}} > /var/run/secrets/kubernetes.io/serviceaccount/token
--- a/pipelines/test/kwok/cluster.yaml
+++ b/pipelines/test/kwok/cluster.yaml
@@ -51,11 +51,7 @@ pipeline:
 
       kubectl wait --for=condition=Ready nodes --all
       kubectl cluster-info
-      export KUBERNETES_SERVICE_HOST="127.0.0.1"
-      export KUBERNETES_SERVICE_PORT="32764"
       mkdir -p /var/run/secrets/kubernetes.io/serviceaccount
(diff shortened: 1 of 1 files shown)


wojciechka (dismissed):

pipelines/test/kwok/cluster.yaml
```diff
@@ -45,3 +51,11 @@ pipeline:
 
       kubectl wait --for=condition=Ready nodes --all
       kubectl cluster-info
+      export KUBERNETES_SERVICE_HOST="127.0.0.1"
+      export KUBERNETES_SERVICE_PORT="32764"
+      mkdir -p /var/run/secrets/kubernetes.io/serviceaccount
+      CA=$(kubectl config view --raw -o jsonpath='{.clusters[0].cluster.certificate-authority}')
+      cp $CA /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+      echo ${{inputs.namespace}} > /var/run/secrets/kubernetes.io/serviceaccount/namespace
+      kubectl create sa ${{inputs.namespace}}
```
maxgio92, line 60:
```suggestion
      kubectl create sa default
```

pipelines/test/kwok/cluster.yaml
```diff
@@ -45,3 +51,11 @@ pipeline:
...
+      kubectl create token ${{inputs.namespace}} > /var/run/secrets/kubernetes.io/serviceaccount/token
```
maxgio92, line 61: I'd not create the token by default following k8s. For security reason, token are not created by default since 1.23, and they're not always required in our tests IMHO

pipelines/test/kwok/cluster.yaml
```diff
@@ -45,3 +51,11 @@ pipeline:
 
       kubectl wait --for=condition=Ready nodes --all
       kubectl cluster-info
+      export KUBERNETES_SERVICE_HOST="127.0.0.1"
+      export KUBERNETES_SERVICE_PORT="32764"
```
maxgio92, line 55: These will be set only for the current pipeline step, which is only this shell process and children that run kwok.
Something like this will fail:
```
...
  - uses: test/kwok/cluster
  - runs: |
      set -u
      echo $KUBERNETES_SERVICE_HOST
```

with
```
...
2025/04/15 10:56:53 WARN /bin/sh: KUBERNETES_SERVICE_HOST: parameter not set
```

maxgio92 (approved):

wojciechka (approved):

a-crate:
Do all the packages which use this pipeline still pass testing after these changes?

pipelines/test/kwok/cluster.yaml
```diff
@@ -45,3 +51,7 @@ pipeline:
 
       kubectl wait --for=condition=Ready nodes --all
       kubectl cluster-info
+      mkdir -p /var/run/secrets/kubernetes.io/serviceaccount
+      CA=$(kubectl config view --raw -o jsonpath='{.clusters[0].cluster.certificate-authority}')
+      cp $CA /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
```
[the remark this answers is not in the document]
kranurag7: thanks Jamon for reviews, I think if we hardcode this to `/root` then that should work but we might run into issues when we use `accounts:` block inside tests. 

but it'll fail anyways if we use any other UID here because that uid won't have enough permissions to create dirs in `/var/` directory.

So, Yes I think we can do this change.

PR #50434, opened 2025-04-15
openssf-compiler-options: make it easier to integrate with oldglibc

xnox:
Sometimes binaries need to build against older glibc. Currently it is
achieved via build-depends on oldglibc and environment/buildsystem
specific settings. But those at times fail to propagate to the
toolchain. Instead of relying on the environment/buildsystem settings
create integration point in openssf-compiler-options to allow oldglibc
package to inject the relevant linker & include flags. Note that lots
of software is starting to omit glibc libraries that got merged into
libc.so, and thus oldglibc can also specify linking with old split-out
dl, pthread libraries to resolve a class of failures to build from
source. These builds can also then make different hardening choices,
and optimisations in terms of march/mtune.

This has been end to end tested in:
- https://github.com/chainguard-dev/enterprise-packages/pull/20345

But will be cleaned up and split out into individual pull
requests. This is one of them.

### DIFF
--- a/openssf-compiler-options.yaml
+++ b/openssf-compiler-options.yaml
@@ -1,7 +1,7 @@
 package:
   name: openssf-compiler-options
   version: 20240627
-  epoch: 17
+  epoch: 18
   description: "Compiler Options Hardening Guide for C and C++"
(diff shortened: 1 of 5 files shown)


techalchemy (approved):
lgtm, thanks!

PR #50441, opened 2025-04-15
haproxy-3.1: remediate CVE-2025-3246

dnegreira:
Apply patch from upstream to remediate CVE-2025-3246.
Patch: https://github.com/haproxy/haproxy/commit/3e3b9eebf871510aee36c3a3336faac2f38c9559
Fixes: GHSA-frg5-h47x-75j9

### DIFF
--- a/haproxy-3.1.yaml
+++ b/haproxy-3.1.yaml
@@ -1,7 +1,7 @@
 package:
   name: haproxy-3.1
   version: "3.1.6"
-  epoch: 40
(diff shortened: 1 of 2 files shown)


dnegreira:
This patch is fixing the alert on the CVE scan, but since there is no version information upstream, we will need to land a patch and create an advisory for when the package is built.

jamie-albert (approved):

PR #50475, opened 2025-04-15
chore(world): Pin numpy to 1.x

EyeCantCU:
All packages will need to be rebuilt with 2.x. Pin to 1.x for now to avoid disruption


AmberArcadia:
This is super stale now and we're fixing numpy issues in fresh PRs :)

PR #50513, opened 2025-04-15
xfsprogs - split into xfsprogs-core and xfsprogs-extra

smoser:
The vast majority of filesystem use is covered with 'fsck' and 'mkfs'.
The other tools provided by xfs are largely debug or recovery tools.

This splits out that core functionality into its own package in
order to reduce the dependency footprint of xfsprogs.

The -extra package is added for all other xfsprogs function.

Things to note here:
 * systemd-fsck (and probably other tools that run fsck)
   will invoke xfs.fsck.
 * xfs.fsck is a shell script wrapper around
   xfs_repair, thus incuring a /bin/sh dependency.

   Unfortunately, the /bin/sh dependency is not recorded here
   and will not be found by SCA in melange.
   See https://github.com/chainguard-dev/go-apk/issues/264 for more
   info.

 * Runtime use of xfs does not imply filesystem creation, so
   so mkfs.xfs could be further split out here into its own package.

   The current reverse dependencies of xfsprogs are
    aws-ebs-csi-driver
    azuredisk-csi
    docker-dind
    gcp-compute-persistent-disk-csi-driver-fips
    lvm-driver

   In the interest of caution, and the fact that mkfs does not incur
   any additional dependencies I have left mkfs.xfs as part of xfs-core
   so that those packages will have access to it.

### DIFF
--- a/xfsprogs.yaml
+++ b/xfsprogs.yaml
@@ -1,15 +1,15 @@
 package:
   name: xfsprogs
   version: "6.14.0"
-  epoch: 1
+  epoch: 2
   description: XFS filesystem utilities
   copyright:
     - license: LGPL-2.1-or-later
   dependencies:
     runtime:
       - merged-usrsbin
-      - python3
       - wolfi-baselayout
+      - xfsprogs-core
 
 environment:
   contents:
@@ -57,12 +57,27 @@ pipeline:
 subpackages:
   - name: xfsprogs-dev
     description: "headers for xfsprogs"
-    pipeline:
-      - uses: split/dev
     dependencies:
       runtime:
         - merged-usrsbin
         - wolfi-baselayout
+        - xfsprogs-libs
+    pipeline:
+      - uses: split/dev
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/usr/lib64
+          mv ${{targets.destdir}}/usr/lib64/libhandle.so ${{targets.subpkgdir}}/usr/lib64
+
+  - name: xfsprogs-doc
+    description: xfsprogs docs
+    pipeline:
+      - uses: split/manpages
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/usr/share/doc
+          mv ${{targets.destdir}}/usr/share/doc/xfsprogs ${{targets.subpkgdir}}/usr/share/doc
+    test:
+      pipeline:
+        - uses: test/docs
 
   - name: xfsprogs-libs
     description: "libraries for xfsprogs"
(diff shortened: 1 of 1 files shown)


stevebeattie:
You need to also split out `/usr/bin/xfs_property` to get rid of the dependency on `cmd:bash` in the main xfsprogs package.

smoser:
> You need to also split out /usr/bin/xfs_property to get rid of the dependency on cmd:bash in the main xfsprogs package.

bleh. you're right.  I really want to just make a '-extra' package and put anything non-essential into it.


these are the shbangs in the curren package
```
$ cat check
apk info -L $1 | while read f; do
  [ -f "$f" ] || continue;
  out=$(head -n1 "$f")  || continue;
  [ "${out#*#!}" = "$out" ] && echo "$f" && continue;
  echo "$f: $out"; done
$ ./check | grep usr.bin
usr/bin/fsck.xfs: #!/bin/sh -f
usr/bin/mkfs.xfs
usr/bin/xfs_admin: #!/bin/sh -f
usr/bin/xfs_bmap: #!/bin/sh -f
usr/bin/xfs_copy
usr/bin/xfs_db
usr/bin/xfs_estimate
usr/bin/xfs_freeze: #!/bin/sh -f
usr/bin/xfs_fsr
usr/bin/xfs_growfs
usr/bin/xfs_info: #!/bin/sh -f
usr/bin/xfs_io
usr/bin/xfs_logprint
usr/bin/xfs_mdrestore
usr/bin/xfs_metadump: #!/bin/sh -f
usr/bin/xfs_mkfile: #!/bin/sh -f
usr/bin/xfs_ncheck: #!/bin/sh -f
usr/bin/xfs_property: #!/bin/bash -f
usr/bin/xfs_protofile: #!/usr/bin/python3
usr/bin/xfs_quota
usr/bin/xfs_repair
usr/bin/xfs_rtcp
usr/bin/xfs_spaceman
```

my feeling is that everything is "extra" other than mkfs fsck (and xfs_repair as a dep) and maybe xfs_freeze (which calls xfs_io). (growfs, systemd just makes an ioctl to grow [link](https://github.com/systemd/systemd/blob/5d902cc21f896a7f2c764982ce88083e77287cee/src/shared/resize-fs.c#L78)).

stevebeattie:
The `fsck.xfs` thing is really weird, it by default tells you to run xfs_repair directly, despite having a bunch of code to invoke xfs_repair. It looks like it's just a placeholder for fsck to call that does nothing: https://linux.die.net/man/8/fsck.xfs


> fsck.xfs is called by the generic Linux [fsck](https://linux.die.net/man/8/fsck)(8) program at startup to check and repair an XFS filesystem. XFS is a journaling filesystem and performs recovery at [mount](https://linux.die.net/man/8/mount)(8) time if necessary, so fsck.xfs simply exits with a zero exit status.

smoser:
> The fsck.xfs thing is really weird, it by default tells you to run xfs_repair directly,

as far as I can tell, [systemd-fsck](https://manpages.ubuntu.com/manpages/plucky/man8/systemd-fsck-root.service.8.html) literally invokes fsck.<type>.  And I'm guessing it is not the only thing that does that. so the shell script provides the caller "api".

smoser:
I'm interested in feedback from anyone on this question: do we think that xfsprogs having (indirectly) mkfs.xfs and fsck.xfs is all we really need?

while some things _could_ be using more than that, I don't know that they are and it seems less likely.

dannf (approved):
I haven't used XFS in ages, but this seems to make sense to me.

89luca89 (approved):

PR #50522, opened 2025-04-15
sudo: Update to 1.9.16p2

sergiodj:
sudo is not using `${{package.version}}` in its `fetch` pipeline, which causes the automatic update to actually *not* update anything.

### DIFF
--- a/sudo.yaml
+++ b/sudo.yaml
@@ -1,7 +1,7 @@
 package:
   name: sudo
-  version: 1.9.16
-  epoch: 40
+  version: 1.9.16_p2
+  epoch: 0
   description: Give certain users the ability to run some commands as root
   copyright:
     - license: ISC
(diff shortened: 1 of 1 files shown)


sergiodj:
I believe this is ready for review.  Tests are still failing, but apparently that's expected with sudo because the build pipeline requires permission to setsuid binaries (which apparently doesn't happen on GH CI?)

dannf (dismissed):

dannf (dismissed):

sergiodj:
This proved to be more complex than initially thought.

See https://github.com/chainguard-dev/melange/pull/1926 for more information.  Meanwhile, I'll revert the test skip hunk from this PR.

egibs:
@sergiodj -- #52093 has updated comments for parsing the version when we land support for patch levels.

PR #50557, opened 2025-04-16
package update: Bump libmspack to 1.11 and switch to git-based updates

debasishbsws:
Fixes: https://github.com/chainguard-dev/internal-dev/issues/11237

- Updated `libmspack` version from `0.11_alpha` to `1.11`.
- Switched from `release-monitor` to `git` backend for version tracking.
- The release monitor was already referencing the GitHub repo: https://github.com/kyz/libmspack.git
- This repo hosts both `libmspack` and `cabextract` and follows a unified versioning scheme.
- Replaced `fetch` with `git-checkout` to track upstream tags directly.

Even though their [official website](https://www.cabextract.org.uk/libmspack/) says:
> "The latest release of libmspack is [libmspack 0.11alpha](https://www.cabextract.org.uk/libmspack/libmspack-0.11alpha.tar.gz), released on 5 February 2023. In-development code can be obtained from the [libmspack Git repository](https://github.com/kyz/libmspack/tree/master/libmspack)."

Since both `libmspack` and `cabextract` are closely tied and live in the same repository, this update improves maintainability and allows for automated updates going forward.

Arch Linux uses the same version format ref: https://gitlab.archlinux.org/archlinux/packaging/packages/libmspack/-/blob/main/PKGBUILD

### DIFF
--- a/libmspack.yaml
+++ b/libmspack.yaml
@@ -1,8 +1,8 @@
 # Generated from https://git.alpinelinux.org/aports/plain/community/libmspack/APKBUILD
 package:
   name: libmspack
-  version: 0.11_alpha
-  epoch: 3
+  version: 1.11
+  epoch: 0
   description: Library for Microsoft CAB compression formats
   copyright:
     - license: LGPL-2.1-only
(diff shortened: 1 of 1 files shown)


OddBloke (changes requested):
I don't believe we should move away from upstream's versioning scheme: I think Arch are wrong here. Debian, Ubuntu and Fedora all use the 0.x versioning.

I think moving to git is potentially a good improvement: have you confirmed that the content of the tarball and of the git checkout are the same? Some projects will include generated files in their tarball, so this isn't always the case.

If they are the same (or the differences are immaterial), then I think we could filter tags to a prefix of `v0.` to avoid picking up cabextract tags. (I would suggest leaving it `manual: true` though, with an explanatory comment, as this isn't entirely reliable. Given the infrequency of releases, it won't cause much additional work.)

PR #50559, opened 2025-04-16
Azure workload identity

rahulduvedi:


azure-workload-identity-webhook.yaml
```diff
@@ -0,0 +1,56 @@
+package:
+  name: azure-workload-identity-webhook
+  version: 1.5.0
+  epoch: 1
```
kranurag7:
```suggestion
  epoch: 0
```

azure-workload-identity-webhook.yaml
```diff
@@ -0,0 +1,56 @@
...
+  description: Mutating webhook for Azure Workload Identity
+  copyright:
+    - license: MIT
+
+environment:
+  environment:
+    CGO_ENABLED: "0"
+    GOFLAGS: "-trimpath"
```
kranurag7, line 12: we should consider omitting this, defaults of go/build pipeline is good. We want to use dynamic linking and avoid trimming the path.

azure-workload-identity-webhook.yaml
```diff
@@ -0,0 +1,56 @@
...
+      tag: v${{package.version}}
+      expected-commit: 350cb28ceab55d0696bef2077a1cc9dff368e982
+
+  - uses: go/build
+    with:
+      packages: ./cmd/webhook
+      output: manager
+      ldflags: |-
+        -X github.com/Azure/azure-workload-identity/pkg/version.BuildVersion=v${{package.version}}
+        -X github.com/Azure/azure-workload-identity/pkg/version.BuildTime=$(date -u +%Y-%m-%d-%H:%M)
```
kranurag7, line 27: include `-d "@$SOURCE_DATE_EPOCH"` as well for reproducibility reasons.

azure-workload-identity-webhook.yaml
```diff
@@ -0,0 +1,56 @@
...
+        -X github.com/Azure/azure-workload-identity/pkg/version.VCSVersion=$(git rev-parse --short HEAD)
+        -X github.com/Azure/azure-workload-identity/pkg/webhook.ProxyImageVersion=v${{package.version}}
+        -X github.com/Azure/azure-workload-identity/pkg/webhook.ProxyImageRegistry=mcr.microsoft.com/oss/azure/workload-identity
```
kranurag7, line 30: looks good to fallback to as a proxy registry.

azure-workload-identity-webhook.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+  - uses: strip
```
kranurag7, line 32:
```suggestion

```
can be omitted go/build pipeline does the stripping for us.

azure-workload-identity-webhook.yaml
```diff
@@ -0,0 +1,56 @@
...
+  enabled: true
+  github:
+    identifier: Azure/azure-workload-identity
+    strip-prefix: v
+
+test:
+  pipeline:
+    - name: "Run manager and verify help output"
+      runs: |
+        manager --help
```
kranurag7, line 56: can we expand on tests here, maybe by using `test/kwok/cluster` pipeline and then running the binary in a pseudo-kubernetes environment.

kranurag7 (commented):
great work!! some comments inline.

azure-workload-identity-webhook.yaml
kranurag7: this is resolved via https://github.com/wolfi-dev/os/pull/50559/commits/375840479d3e90616a26e196033d959633011e8a

azure-workload-identity-webhook.yaml
rahulduvedi: Thanks for reminding me about this again. I missed it again

azure-workload-identity-webhook.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+        kubectl get namespace azure-workload-identity-test
+
+        # Create dummy certificates in /tmp
+        echo "Creating dummy certificates..."
+        openssl req -x509 -newkey rsa:2048 -keyout /tmp/tls.key -out /tmp/tls.crt -days 365 -nodes -subj "/CN=webhook.azure-workload-identity-test.svc"
+
+        echo "Starting manager with cert rotation disabled..."
+        export AZURE_TENANT_ID="00000000-0000-0000-0000-000000000000"
+        manager --disable-cert-rotation --metrics-addr=:8082 --health-addr=:8081 --webhook-cert-dir=/tmp > /tmp/webhook.log 2>&1 &
```
kranurag7, line 70: we can use `test/daemon-check-output` pipeline here to start this with a timeout and then verify the logs.

In the post: block we can check for `/metrics` or `/healthz` endpoints which are likely exposes looking at the command above. 

the namespace creation and openssl commands can go in `setup:` block. 

metrics-server is a good example to refer: https://github.com/wolfi-dev/os/blob/0822c81573d7c9ae86f46e48c76d2b0434a8f03b/metrics-server.yaml#L85-L117

rahulduvedi: Thanks for suggesting these, I made the required changes

azure-workload-identity-webhook.yaml
```diff
@@ -0,0 +1,100 @@
...
+          Starting metrics server
+          Serving metrics server
+          Starting webhook server
+          Serving webhook server
+          Registering webhook
+        error_strings: |
+          failed
+          Failed
+          error
+          Error
```
kranurag7, line 88: is this required? I think we can omit this probably.

azure-workload-identity-webhook.yaml
```diff
@@ -0,0 +1,100 @@
...
+        post: |
+          # Verify health endpoint
+          curl -s http://localhost:8081/healthz || exit 1
```
kranurag7, line 90: we can consider grepping for `grep -i "OK"` if it's returning ok.

azure-workload-identity-webhook.yaml
```diff
@@ -0,0 +1,100 @@
...
+          echo "Health endpoint check passed"
+
+          # Check prometheus metrics endpoint
+          curl -s http://localhost:8082/metrics || exit 1
```
kranurag7, line 93: same here, we can consider grepping for metrics.

rahulduvedi: Made the requested changes, thanks for suggesting

kranurag7 (approved):

PR #50565, opened 2025-04-16
fix(update): Use simple PyPI URI for amazon-q-developer-jupyterlab-ext and bump to version 3.4.8

debasishbsws:
Fix https://github.com/chainguard-dev/internal-dev/issues/11297

- Updated package to version `3.4.8` and reset epoch to `0`.
- Replaced direct PyPI file URI with simplified source URI using the `/packages/source` endpoint.
- This avoids hash-based path issues that break update automation.
- Fixes update bot failure https://github.com/chainguard-dev/internal-dev/issues/11297 caused by dynamic PyPI URLs.


py3-amazon-q-developer-jupyterlab-ext.yaml
```diff
@@ -31,8 +31,8 @@ environment:
 pipeline:
   - uses: fetch
     with:
-      expected-sha256: 2537744335ac2a1d470ecc8673b5ecccbcec0f104b2703076e20f4e4f9ff6704
-      uri: https://files.pythonhosted.org/packages/61/0c/fc475fe80cb645ecc5324d0a551f8f33929d1560f70cf812c62677cc8e23/amazon-q-developer-jupyterlab-ext-${{package.version}}.tar.gz
+      expected-sha256: 3bcf811804dd679a5cf6eb4cfc0e860cdbe34ea5d4f3092b8288cc1f4269fd05
+      uri: https://pypi.io/packages/source/a/amazon-q-developer-jupyterlab-ext/amazon-q-developer-jupyterlab-ext-${{package.version}}.tar.gz
```
OddBloke, line 35: Good fix! Unless you've seen an indication that pypi.io is preferred by PyPI, I would stick with files.pythonhosted.org, though, for consistency with our other packages.

debasishbsws: Done.

OddBloke (approved):

PR #50581, opened 2025-04-16
openjdk-24: reduce jdk-dev image by 34% by enabling linkable runtime modules

xnox:
Build OpenJDK 24 with [JEP 493](https://openjdk.org/jeps/493): Linking Run-Time Images without JMODs.

This new jdk feature enables use of jlink feature, without need to have a separate copy of jmods. This reduces jdk:24-dev tag size significantly, without compromising on features.

This does change packaging in following ways:

- one cannot modify the builds, as otherwise jlink detects changes to
  JAVA_HOME and refuses to perform jlink
- this means no debug build, no stripping, no splitting dbg build
- all debug symbols are turned off, but relocations are still present,
  thus there is binary size increase of ~0.01% for jre, and ~1% for jdk without jmods.
- But the jdk-dev image which usually ships with jmods becomes 34% smaller at 164M instead of 473M 
- java-cacerts is no longer a symlink, but a vendored copy of our build of it.
- it means when java-cacerts are upgraded, openjdk-24+ must be
  rebuilt, to revendor change. This might be actually a good thing, as
  certificate changes have caused connectivity issues in the past, and
  with this change java epoch will change, which may be easier to spot
  and debug (separate version tag too).

Also cleaned build which was setting empty variables, since the move to implicit spec files.

### DIFF
--- a/openjdk-24.yaml
+++ b/openjdk-24.yaml
@@ -1,7 +1,7 @@
 package:
   name: openjdk-24
   version: "24.0.1"
-  epoch: 1
+  epoch: 2
   description: OpenJDK 24
   copyright:
     - license: GPL-2.0-with-classpath-exception
@@ -10,7 +10,6 @@ package:
       - alsa-lib
       - freetype
       - giflib
-      - java-cacerts
       - lcms2
       - libfontconfig1
       - libjpeg-turbo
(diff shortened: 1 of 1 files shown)


kranurag7 (approved):
Having read JEP 493, I was looking forward to this one. thank you very much Dimitri for this.

EyeCantCU (approved):

PR #50606, opened 2025-04-16
Update OpenJDK licenses to use GPL-2.0+Classpath

jamonation:
Tidying licenses for OpenJDK packages

### DIFF
--- a/openjdk-10.yaml
+++ b/openjdk-10.yaml
@@ -1,7 +1,7 @@
 package:
   name: openjdk-10
   version: 10.0.2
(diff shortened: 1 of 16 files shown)


openjdk-10.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: openjdk-10
   version: 10.0.2
-  epoch: 6
+  epoch: 7
```
ajayk, line 4: No changes , this can be reverted back

ajayk (approved):

PR #50661, opened 2025-04-16
ld linux lib64

xnox:
- **glibc: install ld-linux to real /lib, instead of symlink location /lib64**
  

- **glibc: drop CXXFLAGS and use package elf notes**
  Full openssf hardening is not possible, but at least add package ELF notes.
  
  Also CXXFLAGS are unset since completed implementation of openssf
  specf file, thus remove obsolete comment and empty override of an
  unset variable.

### DIFF
--- a/glibc.yaml
+++ b/glibc.yaml
@@ -44,11 +44,10 @@ environment:
       - texinfo
       - wolfi-baselayout
       - zlib
-  # glibc manages FORTIFY_SOURCE on per-file basis
   environment:
-    CPPFLAGS: ""
     # https://github.com/chainguard-dev/internal-dev/issues/7756
-    GCC_SPEC_FILE: /dev/null
+    # Use at least package elf notes
+    GCC_SPEC_FILE: /home/build/.melange.gcc.spec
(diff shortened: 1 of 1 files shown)


xnox:
Hm....

```
$ readelf --notes libc.so.6

Displaying notes found in: .note.ABI-tag
  Owner                Data size 	Description
  GNU                  0x00000010	NT_GNU_ABI_TAG (ABI version tag)
    OS: Linux, ABI: 4.9.0

Displaying notes found in: .note.package
  Owner                Data size 	Description
  FDO                  0x00000058	FDO_PACKAGING_METADATA
    Packaging Metadata: {"type":"apk","os":"wolfi","name":"glibc","version":"2.41-r2","architecture":"aarch64"}
  FDO                  0x00000058	FDO_PACKAGING_METADATA
    Packaging Metadata: {"type":"apk","os":"wolfi","name":"glibc","version":"2.41-r2","architecture":"aarch64"}
```

That's not nice, we use append to link, but then somehow loading it twice?!

PR #50710, opened 2025-04-17
glibc: install ld-linux to real /lib, instead of symlink location /lib64

xnox:
Note that currently apk audit fails on ld-linux, because it installs
ld-linux into /lib64 on x86_64, which is a symlinked location as per
wolfi-baselayout.

Make the .apk ship it in "/lib" on x86_64, which remains LSB
compliant, but helps with unpacking, as now ld-linux can be unpacked
either before or after wolfi-baselayout.

### DIFF
--- a/glibc.yaml
+++ b/glibc.yaml
@@ -54,7 +54,9 @@ pipeline:
   - uses: git-checkout
     with:
       expected-commit: 74f59e9271cbb4071671e5a474e7d4f1622b186f
-      repository: https://sourceware.org/git/glibc.git
+      # sourceware.org often has clone issues, switch to mirror
+      # recommended at https://sourceware.org/glibc/wiki/GlibcGit
+      repository: https://repo.or.cz/glibc.git
       tag: glibc-${{package.version}}
 
   - uses: patch
--- a/glibc.yaml
+++ b/glibc.yaml
@@ -1,7 +1,7 @@
 package:
   name: glibc
   version: "2.41"
-  epoch: 1
+  epoch: 2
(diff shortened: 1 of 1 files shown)


smoser (approved):

eslerm:
@xnox I do not love the change to use `https://repo.or.cz/glibc.git` instead of upstream.

Even if upstream has pull issues, switching to an external downstream mirror carries significantly more risk and liability. Even when [upstream mentions the downstream in documentation](https://sourceware.org/glibc/wiki/GlibcGit), it introduces high risk at a distro level. Are the security controls as tight on or.cz? It would be easier to explain lapses in upstream's security if an incident were to occur.

An alternative would be to mirror our own copy, similar to [OpenSSL](https://github.com/chainguard-dev/openssl).

xnox:
> @xnox I do not love the change to use `https://repo.or.cz/glibc.git` instead of upstream.
> 
> Even if upstream has pull issues, switching to an external downstream mirror carries significantly more risk and liability. Even when [upstream mentions the downstream in documentation](https://sourceware.org/glibc/wiki/GlibcGit), it introduces high risk at a distro level. Are the security controls as tight on or.cz? It would be easier to explain lapses in upstream's security if an incident were to occur.
> 

Ok. Will attempt to switch back. If we had a pull through mirror or cache that would help.


> An alternative would be to mirror our own copy, similar to [OpenSSL](https://github.com/chainguard-dev/openssl).

We do not mirror our own openssl. No idea what you are talking about.

PR #50973, opened 2025-04-18
Makefile: remove --generate-index setting

dannf:
As the TODO comment implies, this setting doesn't actually do anything.
If it did, that would break many developer's work flows.

Let's remove the line to avoid confusion.

See: https://github.com/wolfi-dev/os/pull/39831


xnox (dismissed):

xnox:
@dannf somehow there is a conflict; please rebase.

xnox (approved):

PR #51272, opened 2025-04-21
Use the Chainguard kernel for QEMU runner.

mattmoor:
This also adds a `CLAUDE.md` to enable folks to more readily use `claude` with Wolfi (this uses the QEMU runner by default).


dannf (approved):
Thanks @mattmoor !

PR #51291, opened 2025-04-21
quality: add kubernetes-ingress-defaultbackend package tests

yashpimple:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches


kubernetes-ingress-defaultbackend.yaml
```diff
@@ -38,7 +38,19 @@ update:
...
+        - wait-for-it
   pipeline:
     - name: help message
       runs: |
         defaultbackend --help
+    - name: Basic functional test for endpoints
+      runs: |
+        defaultbackend &
+        wait-for-it localhost:8080 -t 10
+        wget -q --server-response http://localhost:8080/ -O /dev/null 2>&1 | grep "404 Not Found"
```
EyeCantCU, line 54: This seems redundant

kubernetes-ingress-defaultbackend.yaml
```diff
@@ -38,7 +38,19 @@ update:
...
+
+        wget -q --server-response http://localhost:8080/healthz -O /dev/null 2>&1 | grep "200 OK"
```
EyeCantCU, line 56: I'd use curl to check the response instead of wget + grep

EyeCantCU (changes requested):

kubernetes-ingress-defaultbackend.yaml
```diff
@@ -38,7 +38,19 @@ update:
...
+      packages:
+        - wget
+        - wait-for-it
   pipeline:
     - name: help message
       runs: |
         defaultbackend --help
+    - name: Basic functional test for endpoints
+      runs: |
+        defaultbackend &
```
kranurag7, line 52: thank you, the tests look good.
Can we use `test/daemon-check-output` pipeline in here? If possible then we should default to using that and then check some logs and then in the post: block make use curl to make some request to the server.

yashpimple: @kranurag7 the reason i didn't use `test/daemon-check-output` is because after running `defaultbackend -port=9091` we don't see any output logs in the terminal and due to which i didn't made use of it and went straight ahead with `defaultbackend &`
```
yash-pimple:~/os$ defaultbackend -port=9091


yash-pimple:~/os$^C
```

kubernetes-ingress-defaultbackend.yaml
```diff
@@ -38,7 +38,18 @@ update:
...
+        - curl
   pipeline:
     - name: help message
       runs: |
         defaultbackend --help
+    - name: Basic functional test for endpoints
+      runs: |
+        defaultbackend &
+        wait-for-it localhost:8080 -t 10
+        curl -sI --retry 5 http://localhost:8080/ | grep "404 Not Found"
```
EyeCantCU, line 54: We don't need to validate the response was what we wanted with grep. Do `curl -f -LI...`

yashpimple: @EyeCantCU if we use `curl -f -LI` the -f option causes curl to fail on HTTP errors. However, in our case, the default backend is designed to return a 404 Not Found for the root endpoint (/), which is the expected behavior using `curl -f` here cause the test to fail upon receiving the expected 404 response. Therefore, I've opted to use `curl -LI` without the -f option for the root endpoint, allowing the test to pass when a 404 is correctly returned.

Dentrax (dismissed):

Dentrax (approved):

EyeCantCU (approved):

PR #51292, opened 2025-04-21
add py3-graph-tool package

kranurag7:
to build: 
```bash
make package/cgal
make package/mm-common
make package/libsigcplusplus
make package/cairomm
make package/sparsehash
make package/py3-graph-tool
```

### DIFF
--- a/py3-graph-tool.yaml
+++ b/py3-graph-tool.yaml
@@ -68,6 +68,7 @@ subpackages:
       provides:
         - graph-tool=${{package.full-version}}
       runtime:
+        - cairomm-dev
         - py${{range.key}}-cairo
         - py${{range.key}}-matplotlib
         - py${{range.key}}-numpy
--- a/py3-graph-tool.yaml
+++ b/py3-graph-tool.yaml
@@ -45,6 +45,7 @@ environment:
       - mpfr-dev
       - pkgconf-dev
       - py3-supported-cairo
+      - py3-supported-gobject3
       - py3-supported-numpy
       - py3-supported-python-dev
       - py3-supported-scipy
@@ -68,6 +69,7 @@ subpackages:
       runtime:
         - py${{range.key}}-cairo
         - py${{range.key}}-matplotlib
+        - py${{range.key}}-gobject3
         - py${{range.key}}-numpy
         - py${{range.key}}-scipy
     pipeline:
--- a/py3-graph-tool.yaml
+++ b/py3-graph-tool.yaml
@@ -1,0 +1,177 @@
+package:
+  name: py3-graph-tool
+  version: 2.97
+  epoch: 0
+  description: graph-tool is an efficient Python module for manipulation and statistical analysis of graphs (a.k.a. networks).
+  resources:
+    cpu: 65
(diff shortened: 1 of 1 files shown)


kranurag7:
malcontent was once review here: https://github.com/wolfi-dev/os/pull/51446

pinning to x86 only because it's not building on arm64 architecture for now. I think because of resource constraint, I continue to investigate that.

py3-graph-tool.yaml
```diff
@@ -0,0 +1,124 @@
...
+          opts: |
+            --enable-openmp \
+            --includedir=/usr/lib/cairomm-1.16 \
+            PYTHON=python${{range.key}}
+      - runs: |
+          ln -s /usr/lib/cairomm-1.16/include/cairommconfig.h /usr/include/cairommconfig.h
+          ln -s /usr/include/cairomm-1.16/cairomm/ /usr/include/cairomm
+          ln -s /usr/lib/sigc++-3.0/include/sigc++config.h /usr/include/sigc++config.h
+          ln -s /usr/include/sigc++-3.0/sigc++/ /usr/include/sigc++
+          export CPPFLAGS="$CPPFLAGS -I/usr/lib/cairomm-1.16/include -I/usr/include/cairomm-1.16/cairomm/ -I/usr/include/sigc++-3.0/sigc++/"
```
kranurag7, line 89: not liking this very much but something to do with lots of configure flags and environment variables. I tried using `-I` and `--includedir` and exporting environment variables again and again but all of them were not able to find headers so ended up symlinking them to `/usr/include`

xnox: this indicates broken packaging; of those includes; and theri pkgconfig file. Instead of doing this every time, it is best to fix those pc files to do the right thing out of the box.

py3-graph-tool.yaml
```diff
@@ -0,0 +1,184 @@
...
+    memory: 128Gi
+  copyright:
+    - license: GPL-3.0-only
+
+vars:
+  pypi-package: graph-tool
+  import: graph_tool
+
+data:
+  # TODO: add other supported versions
```
jamonation, line 17: What about other Python versions @kranurag7? Any reason to not add them here?

kranurag7: Jamon, the reason yesterday was that it takes ~45 minutes to build. 

the reason today is that I'm running into linker issues with older versions.

```bash
/usr/lib/gcc/x86_64-pc-linux-gnu/14/../../../../x86_64-pc-linux-gnu/bin/ld: cannot find -lboost_python312: No such file or directory
```
I'm still looking into this as of now.

jamonation:
Can this be merged as is, and refactored to support Python 3.12/3.11 and arm64 at a later date?

kranurag7:
> Can this be merged as is, and refactored to support Python 3.12/3.11 and arm64 at a later date?

Yes Jamon, this is ready to go now I think. mostly it's working and if there are some hiccups; I'll do a followup. From manual testing; this is mostly looking good now.

Dentrax (approved):

jamonation (approved):

PR #51301, opened 2025-04-21
haproxy-3.1: convert to git-checkout and git update backend

OddBloke:
This avoids using release-monitoring.org which, as it doesn't distinguish between version streams of HAProxy, could result in us publishing, e.g., 3.2.x versions to the haproxy-3.1 package.

I've compared the release tarball and the tagged tree, and this is the only difference between the two:

```diff
diff --color -r -u haproxy-3.1/SUBVERS haproxy-3.1.7/SUBVERS
--- haproxy-3.1/SUBVERS	2025-04-21 09:51:58.634952910 -0400
+++ haproxy-3.1.7/SUBVERS	2025-04-17 12:14:57.000000000 -0400
@@ -1,2 +1,2 @@
--$Format:%h$
+-c3f4089

diff --color -r -u haproxy-3.1/VERDATE haproxy-3.1.7/VERDATE
--- haproxy-3.1/VERDATE	2025-04-21 09:51:58.634952910 -0400
+++ haproxy-3.1.7/VERDATE	2025-04-17 12:14:57.000000000 -0400
@@ -1,2 +1,2 @@
-$Format:%ci$
+2025-04-17 18:14:57 +0200
 2025/04/17
```

### DIFF
--- a/haproxy-3.1.yaml
+++ b/haproxy-3.1.yaml
@@ -37,6 +37,7 @@ pipeline:
       tag: v${{package.version}}
       expected-commit: c3f40894532072afb380f1779bce1ef85974df2f
       repository: https://git.haproxy.org/git/haproxy-3.1.git
+
   - uses: autoconf/make
     with:
       opts: |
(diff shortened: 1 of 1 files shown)


debasishbsws (commented):
@OddBloke Its small but still the source is different. Shouldn't we update the epoch?

PR #51313, opened 2025-04-21
add sed and find to etcd-bitnami-compat

sschubertchainguard:
add sed and find to etcd-bitnami-compat


developer-guy (approved):
LGTM, thanks!

kranurag7 (approved):
thank you Scott, given we're adding these deps to -compat package, this looks good to me.

PR #51333, opened 2025-04-21
Claude-improved `berg` tests

mattmoor:


berg.yaml
```diff
@@ -41,7 +41,29 @@ update:
...
+          echo "Invalid command should have failed"
+          exit 1
+        fi
+
+        # Verify help command lists available commands
+        berg help | grep -q "Commands:" || exit 1
+    - name: Test version output format
+      runs: |
+        # Check version output format and value
+        berg --version | grep -q "codeberg-cli ${{package.version}}" || exit 1
```
dustinkirkland, line 61: These are very, very brittle, and ill-advised.  We've long had the ability to test version strings in this way (including in the original non-AI generate tests), but we've found that these break for simple, dumb, and benign reasons, and create toil on the sustaining team....

berg.yaml
```diff
@@ -41,7 +41,29 @@ update:
...
+    - name: Test command documentation
+      runs: |
+        # Verify that subcommands have help information
+        berg repo --help | grep -q "Repository subcommands" || exit 1
+        berg user --help | grep -q "User subcommands" || exit 1
+        berg issue --help | grep -q "Issue subcommands" || exit 1
```
dustinkirkland, line 67: These, too, can be brittle, and break for little to no reason...

mattmoor:
That's the bulk of this one, so will close and revisit with better prompting.

PR #51352, opened 2025-04-21
Add cue schema and vet make target

justinvreeland:
This schema file is generated from the schema.json file in melange:
cue import -f schema.json

The vet makefile target will check the melange file against the schema in the cue file.

I've also included an example fix for the 7zip error as an example but we shouldn't try to fix these by hand.

```
cue vet 7zip.yaml schema.cue
package.version: conflicting values 2301 and string (mismatched types int and string):
    ./7zip.yaml:3:12
    ./schema.cue:3:1
    ./schema.cue:51:12
    ./schema.cue:281:12
pipeline.0.with."strip-components": conflicting values 0 and string (mismatched types int and string):
    ./7zip.yaml:21:25
    ./schema.cue:3:1
    ./schema.cue:62:14
    ./schema.cue:62:17
    ./schema.cue:379:13
make: *** [cue_vet/7zip.yaml] Error 1
```

See also:
https://github.com/chainguard-dev/melange/pull/1929
https://github.com/wolfi-dev/os/pull/50967

### DIFF
--- a/7zip.yaml
+++ b/7zip.yaml
@@ -1,6 +1,6 @@
 package:
   name: 7zip
-  version: 2301
+  version: "2301"
   epoch: 3
   description: "File archiver with a high compression ratio"
   copyright:
@@ -18,7 +18,7 @@ pipeline:
     with:
       uri: https://7-zip.org/a/7z${{package.version}}-src.tar.xz
       expected-sha512: e39f660c023aa65e55388be225b5591fe2a5c9138693f3c9107e2eb4ce97fafde118d3375e01ada99d29de9633f56221b5b3d640c982178884670cd84c8aa986
-      strip-components: 0
+      strip-components: "0"
 
   - name: Configure and build
     runs: |
--- a/Makefile
+++ b/Makefile
@@ -282,4 +282,6 @@ authget = tok=$$(chainctl auth token --audience=$(1)) || \
 
 cue_targets = $(foreach name,$(yamls),cue_vet/$(name))
 $(cue_targets): cue_vet/%:
-	cue vet $* schema.cue
+	$(MELANGE) compile --arch ${ARCH} --pipeline-dir ./pipelines/ $* > /tmp/$*.json
+	cue vet /tmp/$*.json schema.cue
+	rm /tmp/$*.json
--- a/Makefile
+++ b/Makefile
@@ -279,3 +279,7 @@ authget = tok=$$(chainctl auth token --audience=$(1)) || \
   echo "auth-download[$(1)] to $(2) from $(3)" && \
   curl -LS --silent -o $(2).tmp --user "user:$$tok" $(3) && \
 	mv "$(2).tmp" "$(2)"
+
+cue_targets = $(foreach name,$(yamls),cue_vet/$(name))
+$(cue_targets): cue_vet/%:
+	cue vet $* schema.cue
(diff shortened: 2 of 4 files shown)


joshrwolf (dismissed):

7zip.yaml
```diff
@@ -18,7 +18,7 @@ pipeline:
     with:
       uri: https://7-zip.org/a/7z${{package.version}}-src.tar.xz
       expected-sha512: e39f660c023aa65e55388be225b5591fe2a5c9138693f3c9107e2eb4ce97fafde118d3375e01ada99d29de9633f56221b5b3d640c982178884670cd84c8aa986
-      strip-components: 0
+      strip-components: "0"
```
smoser, line 21: This is a big "oh well". But it works be nice to use correct types where possible. 

(Ie, I expect that to be loaded by a yaml lib as a string, when its value should be an integer)

justinvreeland: Yeah, there's a few examples of this having the potential to over-quote things. But Floats are surprisingly sticky https://github.com/chainguard-dev/yam/issues/100 which would be nice to fix.

xnox: this did come up in external security reviews that our 'with' dictionary is always strings (as processed) and for now we do not have ability for pipelines to declare types of the input vars. And inlining type info `!!str` or whatever the syntax for that is, is much uglier than just over-quoting.

smoser: Ah. right. I was thinking that the thing that edited this was "schema aware".  And this is a pipeline and we don't have schema for pipelines (even built-in ones). 

I haven't paid enough attention to cue work.  I saw schema and expected that it would enforce types, and that this type was wrong.

justinvreeland: I can generate schemas for the pipelines, and In an ideal world this will be wrapped up into an LSP so I my editor can format/validate this for me. But small steps. Most everything in our melange schema is really a string. or dictionaries strings.

justinvreeland:
Wrapping back around to this after spending some more time reading the melange code.   I think there's a better first step than I have here.   The melange files we have on disk are different than the final config object that's created. A good example of this is the priorities which are defined in the schema have here as strings but are in reality always used as integers.   So we end up in a situation where someone adds an integer as the provider priority and cue parses it as an integer but melange knows it's a string and someone else will use provider priority as a string so they can have melange do the replacement.  The `py3-` packages do this a lot.

What we can do I think addresses some of the concerns about schema types and over quoting is to compile the melange first. I would've thought would be a 100% pass rate but when we do this we drop down from a few hundred failures to one legit failure.  In this way our schema validation still works but doesn't fight with the melange replacement steps it doesn't know about or have to try and mess with conversion. I think ideally we would have something that reformats the file to match an exiting schema somewhere.

```
/usr/bin/melange compile --arch x86_64 --pipeline-dir ./pipelines/ exim.yaml > /tmp/exim.yaml.json
2025/05/08 20:53:20 WARN unable to detect commit for build config file: determining HEAD: reference not found
2025/05/08 20:53:20 WARN git repository URL for build config not provided
cue vet /tmp/exim.yaml.json schema.cue
subpackages.5.test.pipeline: conflicting values null and [...#Pipeline] (mismatched types null and list):
    ./schema.cue:3:1
    ./schema.cue:66:17
    ./schema.cue:66:20
    ./schema.cue:541:9
    ./schema.cue:554:13
    ../../../../tmp/exim.yaml.json:1:10290
make: *** [Makefile:254: cue_vet/exim.yaml] Error 1
```

You can run this yourself over every file with this nice little fish snippet takes me about 5mins.

```
❯ time for yaml_file in *.yaml
      set OUTPUT (make cue_vet/$yaml_file &> $yaml_file.vet)
      set result $status
      if [ $result = 0 ]
          rm $yaml_file.vet
      end
  end
```

PR #51354, opened 2025-04-21
Improve Trino package

mosabua:
Trino package update. Trino 474 uses Java 24 in the container so our package should do so as well.

Also removed the need for python since the Trino launcher has been using a Go-based binary for a while and Python is no longer needed.

Also remove the Kudu and Phoenix connector plugin comment lines since we remove the connectors in upstream.

Fyi .. I am a Trino maintainer but not a wolfi expert.. ping me on Chainguard or Trino slack for discussion or comment in review.

### DIFF
--- a/trino.yaml
+++ b/trino.yaml
@@ -1,7 +1,7 @@
 package:
   name: trino
-  version: "474"
-  epoch: 2
+  version: "475"
+  epoch: 3
   description: The distributed SQL query engine for big data, formerly known as PrestoSQL
   copyright:
     - license: Apache-2.0
@@ -10,8 +10,7 @@ package:
     memory: 64Gi
   dependencies:
     runtime:
-      - openjdk-23-default-jvm
-      - python3 # required by trino launcher script
+      - openjdk-24-default-jvm
       - trino-config
 
 environment:
@@ -26,7 +25,7 @@ environment:
       - go
       - jvmkill
       - maven
-      - openjdk-23
+      - openjdk-24-default-jdk
       - upx
       - wolfi-base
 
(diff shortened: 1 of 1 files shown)


trino.yaml
```diff
@@ -26,7 +25,7 @@ environment:
       - go
       - jvmkill
       - maven
-      - openjdk-23
+      - openjdk-24
```
kranurag7, line 28:
```suggestion
      - openjdk-24-default-jdk
```
nit: to be explicit with full jdk.

trino.yaml
```diff
@@ -195,7 +192,7 @@ test:
         - ${{package.name}}-oci-entrypoint # need to setup basic
         - bash
     environment:
-      JAVA_HOME: /usr/lib/jvm/java-23-openjdk
+      JAVA_HOME: /usr/lib/jvm/java-24-openjdk
```
kranurag7, line 195:
```suggestion
      JAVA_HOME: /usr/lib/jvm/default-jvm
```
this way we won't have to update it on jdk updates.

kranurag7 (commented):
thank you for updating it. this is awesome. 

we need to bump the epoch field here `package.epoch` to ensure that we get a new revision out here.

trino.yaml
mosabua: updated

mosabua:
Thank you @kranurag7 and @ajayk - I addressed the comments now, maybe good to go...

mosabua:
Thanks for the spelling fix @kranurag7 .. should I squash and rebase or something and is there something else I should do?

ajayk (dismissed):

mosabua:
Updated commit hash now .. 🤞 this turns green now

ajayk (approved):

mosabua:
Awesome .. thank you @ajayk

PR #51367, opened 2025-04-22
package vm with ui

anushkamittal20:
UI was previously missing from vm, added now

### DIFF
--- a/victoriametrics.yaml
+++ b/victoriametrics.yaml
@@ -94,13 +94,3 @@ test:
         victoria-metrics -version
         victoria-metrics --version
         victoria-metrics --help
-    - name: "Test victoria-metrics-prod"
-      uses: test/daemon-check-output
-      with:
-        start: victoria-metrics -httpListenAddr=:8428
-        post: |
-          fail() { echo "FATAL: $*"; exit 1; }
-          curl -L http://localhost:8428/vmui | grep -q "<title>vmui</title>" || fail "vmui page missing or unexpected"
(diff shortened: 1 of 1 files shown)


victoriametrics.yaml
```diff
@@ -13,13 +13,31 @@ data:
...
+
 pipeline:
   - uses: git-checkout
     with:
       expected-commit: 123f373537323f4e7278db5601dab4e67759af1c
       repository: https://github.com/VictoriaMetrics/VictoriaMetrics
       tag: v${{package.version}}
 
+  - runs: |
+      cp docs/victoriametrics/MetricsQL.md app/vmui/packages/vmui/src/assets/MetricsQL.md
```
kranurag7, line 30: If we intend to copy this to the main package then we should create `app/vmui/packages/vmui/src/assets/` directory and cp into `${{targets.contextdir}}` 

or this is expected this way only, as of now, this seems build-time only.

anushkamittal20: so here's me understanding of this:
https://github.com/VictoriaMetrics/VictoriaMetrics/blob/master/app/vmui/README.md#static-build I am getting the static build files in the right place i.e. under `app/` so when we finally build vm its embedded.

In the steps added, I am replicating the same things (the make targets) in our build. We dont need them in the final binary.

kranurag7: thank you for the references, cross verified it again and this looks good.

kranurag7 (approved):
thank you, looks good to me. upstream project is using nodejs-20 so we're using the same here in build environment.

PR #51445, opened 2025-04-22
makefile kernel improvements

dannf:
- Avoid using predictable paths in `/tmp`
- Split up the target into multiple chunks to ease debugging
- Add kernel dep to targets that call melange build when using qemu runner
- Fix `QEMU_KERNEL_IMAGE` being set in melange's environment
- Cleanup kernel w/ `clean` target

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,9 +117,10 @@ fetch-kernel:
 kernel/APKINDEX.tar.gz:
 	$(MAKE) apk-token
 	mkdir -p kernel
-	curl -LS --silent -o $@ \
+	curl -LS --silent -o $@.tmp \
 	  --user user:$$(chainctl auth token --audience apk.cgr.dev) \
 	  https://apk.cgr.dev/chainguard-private/$(ARCH)/APKINDEX.tar.gz
+	mv $@.tmp $@
 
 kernel/APKINDEX: kernel/APKINDEX.tar.gz
 	tar -x -C kernel -f $< $(notdir $@)
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ WOLFICTL ?= $(shell which wolfictl)
 KEY ?= local-melange.rsa
 REPO ?= $(shell pwd)/packages
 SOURCES ?= gs://wolfi-sources/
+QEMU_KERNEL_REPO := https://apk.cgr.dev/chainguard-private/
 
 ifneq (${MELANGE_RUNNER},)
 	MELANGE_OPTS += --runner ${MELANGE_RUNNER}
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -98,22 +105,43 @@ ${CACHEDIR}/.libraries_token.txt: cache
...
+	$(MAKE) apk-token
+	kver=$(file < kernel/chosen); \
+	  curl -LS --silent -o $@.tmp \
+	  --user user:$$(chainctl auth token --audience apk.cgr.dev) \
+	  https://apk.cgr.dev/chainguard-private/$(ARCH)/linux-$$kver.apk
+	mv $@.tmp $@
+
+kernel/boot/vmlinuz: kernel/linux.apk
+	tar -x -C kernel -f $< boot/ 2> /dev/null
+	touch $@
```
mattmoor, line 144: JFYI I copied the original block from enterprise, so let's update there as well!

Makefile
```diff
@@ -19,6 +19,12 @@ SOURCES ?= gs://wolfi-sources/
 ifneq (${MELANGE_RUNNER},)
 	MELANGE_OPTS += --runner ${MELANGE_RUNNER}
 endif
+QEMU_KERNEL_IMAGE ?= kernel/boot/vmlinuz
+ifeq (${MELANGE_RUNNER},qemu)
+	QEMU_KERNEL_DEP = ${QEMU_KERNEL_IMAGE}
+	export QEMU_KERNEL_IMAGE
+endif
```
mattmoor, line 26: Does this magic make it so that we fetch the kernel when the melange runner is set to QEMU?  Amazing 🤩

mattmoor (commented):
I'm a Makefile noob, so going to defer to @smoser here, but this LGTM!

Makefile
```diff
@@ -98,22 +105,44 @@ ${CACHEDIR}/.libraries_token.txt: cache
...
+
+kernel/APKINDEX: kernel/APKINDEX.tar.gz
+	tar -x -C kernel -f $< $(notdir $@)
+	touch $@
+
+kernel/chosen: kernel/APKINDEX
+	grep -e '^P:' -e '^V:' $< | sed -n '/^P:linux$$/{n;p}' > \
+	  kernel/available
+	grep '^V:' kernel/available | sed 's/V://' | \
+	  sort -V | tail -n1 > kernel/chosen
```
smoser, line 125: 2 things here. when you redirect like this, you create the file and then it will be up to date even if it fails.
it'd be good to verify that there is content you expected and then move it into place.

```
out=$$(grep '^V:' kernel/available | sed 's/V://' | sort -V | tail -n1) && [ -n "$$out" ] && echo "$$out" > $@)

smoser: something like that, then you it doesnt get updated unless its sane.

smoser (approved):
love it. thanks.

PR #51468, opened 2025-04-22
ruby3.2/.3/.4/-rails-8.0/GHSA-5w6v-399v-w3cc fix

jamie-albert:
Epoch bump brings in new nokogiri version which remediates the CVE


stevebeattie:
The malcontent scan is hitting a false positive that is flagging an embedded copy of the libev C library in the nio4r gem, based on some aggressive javascript rules that were recently added.

AmberArcadia (approved):

PR #51479, opened 2025-04-23
Pgpool2-4.6 package addition

AnmolVirdi:
**Pgpool2-4.6 package** `Current version: v4.6.0`
- Middleware that works between PostgreSQL servers and a PostgreSQL database client.

**Type**: C Binaries
**REF**: https://github.com/pgpool/pgpool2

**Tests:**
- Complete setup and validation with local `postgresql` server
- `pcp binaries` tests
- `Healthchecks`

### DIFF
--- a/pgpool2-4.6.yaml
+++ b/pgpool2-4.6.yaml
@@ -242,7 +242,7 @@ test:
         pcp_health_check_stats -U "$PG_USER" -h "$PGHOST" -p "$PCP_PORT" -w -v 0 | grep -q "primary"
     - name: "Display and validate Pgpool childprocesses"
       runs: |
-        pcp_proc_count -U "$PG_USER" -h "$PGHOST" -p "$PCP_PORT" 
+        pcp_proc_count -U "$PG_USER" -h "$PGHOST" -p "$PCP_PORT"
         psql -U $PG_USER -h localhost -p $PGPOOL_PORT -d ${PG_DB} -c "SELECT pg_sleep(60);" & sleep 1
         pcp_proc_info -U "$PG_USER" -h "$PGHOST" -p "$PCP_PORT" --all | grep -q "Execute command"
     - name: "Check and validate pcp_pool_status"
--- a/pgpool2-4.6.yaml
+++ b/pgpool2-4.6.yaml
@@ -156,6 +156,7 @@ update:
     identifier: pgpool/pgpool2
     strip-prefix: V
     use-tag: true
+    tag-filter-prefix: V4_6
 
 test:
   environment:
--- a/pgpool2-4.6.yaml
+++ b/pgpool2-4.6.yaml
@@ -244,5 +244,3 @@ test:
       runs: |
         pcp_stop_pgpool -U "$PG_USER" -h "$PGHOST" -p "$PCP_PORT" --mode=fast | grep -q "Command Successful"
     - uses: test/tw/ldd-check
-      with:
-        packages: ${{package.name}}
--- a/pgpool2-4.6.yaml
+++ b/pgpool2-4.6.yaml
@@ -1,0 +1,213 @@
+package:
+  name: pgpool2-4.6
+  version: 4.6.0
+  epoch: 0
+  description: Middleware that works between PostgreSQL servers and a PostgreSQL database client.
+  copyright:
+    - license: BSD-3-Clause AND MIT
+  dependencies:
+    provides:
+      - pgpool2=${{package.full-version}}
+    runtime:
+      - merged-usrsbin
+      - wolfi-baselayout
+
+environment:
+  contents:
+    packages:
+      - autoconf
+      - automake
+      - bison
+      - build-base
+      - busybox
+      - flex
+      - libtool
+      - openssl-dev
+      - postgresql-dev
+
+var-transforms:
+  - from: ${{package.version}}
+    match: \.
+    replace: _
+    to: mangled-package-version
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/pgpool/pgpool2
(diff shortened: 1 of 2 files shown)


pgpool2-4.6.yaml
```diff
@@ -0,0 +1,259 @@
...
+  enabled: true
+  ignore-regex-patterns:
+    - _RPM
+  version-transform:
+    - match: "_"
+      replace: .
+  github:
+    identifier: pgpool/pgpool2
+    strip-prefix: V
+    use-tag: true
```
debasishbsws, line 158: Since this is a version stream for `4.6`, you should add `tag-filter-prefix:` to ensure the automation only picks updates relevant to this version stream:
```suggestion
  github:
    identifier: pgpool/pgpool2
    strip-prefix: V
    use-tag: true
    tag-filter-prefix: V4_6
```
This helps avoid updates from other major versions (e.g., `V4_7`) being incorrectly applied.

pgpool2-4.6.yaml
```diff
@@ -0,0 +1,259 @@
...
+      # copy the patch file to patch the output form bitnami/compat and apply it
+      - runs: |
+          cp bitnami-compat-remove-os-flavour.patch "${{targets.contextdir}}/"
+      - uses: patch
+        working-directory: "${{targets.contextdir}}"
+        with:
+          patches: bitnami-compat-remove-os-flavour.patch
+      # remove patch file to prevent it from being packaged
+      - runs: |
+          rm "${{targets.contextdir}}/bitnami-compat-remove-os-flavour.patch"
```
debasishbsws, line 97: You don’t need to copy the patch file into the working directory before applying it. You can directly use the absolute path, which simplifies the pipeline and removes unnecessary steps:
```suggestion
      - uses: patch
        working-directory: "${{targets.contextdir}}"
        with:
          patches: /home/build/bitnami-compat-remove-os-flavour.patch
```

Also, please make sure to include context for the patch—why it's needed and when it can be removed. Refer to our [Patch-creation-guideline](https://eng.inky.wtf/docs/packages/guides/cve-remediations-guides/patch-creation-guide/)

debasishbsws (changes requested):
Some suggestions

pgpool2-4.6.yaml
AnmolVirdi: Thanks @debasishbsws, I've pushed suggested changes and added more context for the patch.

pgpool2-4.6.yaml
```diff
@@ -0,0 +1,255 @@
...
+  epoch: 0
+  description: Middleware that works between PostgreSQL servers and a PostgreSQL database client.
+  copyright:
+    - license: BSD-3-Clause AND MIT
+  dependencies:
+    provides:
+      - pgpool2=${{package.full-version}}
+    runtime:
+      - merged-usrsbin
+      - wolfi-baselayout
```
kranurag7, line 13: I think we can omit this given we're already usrmerged.

pgpool2-4.6.yaml
```diff
@@ -0,0 +1,255 @@
...
+  - name: ${{package.name}}-dev
+    pipeline:
+      - uses: split/dev
+    dependencies:
+      provides:
+        - pgpool2-dev=${{package.full-version}}
+      runtime:
+        - ${{package.name}}
+        - merged-usrsbin
+        - wolfi-baselayout
```
kranurag7, line 69: same for these two runtime deps.

pgpool2-4.6.yaml
```diff
@@ -0,0 +1,255 @@
...
+    dependencies:
+      provides:
+        - pgpool2-bitnami-compat=${{package.full-version}}
+      runtime:
+        - busybox
+        - coreutils
+        - merged-usrsbin
+        - openldap
+        - postgresql-client
+        - wolfi-baselayout
```
kranurag7, line 82: we can omit merged-usrsbin and wolfi-baselayout here.

pgpool2-4.6.yaml
```diff
@@ -0,0 +1,255 @@
...
+        pcp_pool_status -U "$PG_USER" -h "$PGHOST" -p "$PCP_PORT" | grep -q "wolfi" #used sr_check_user
+    - name: "Detach primary node"
+      runs: |
+        pcp_detach_node -U "$PG_USER" -h "$PGHOST" -p "$PCP_PORT" -n 0 --gracefully | grep -q "Command Successful"
+    - name: "Stop pgpool using pcp_stop_pgpool"
+      runs: |
+        pcp_stop_pgpool -U "$PG_USER" -h "$PGHOST" -p "$PCP_PORT" --mode=fast | grep -q "Command Successful"
+    - uses: test/tw/ldd-check
+      with:
+        packages: ${{package.name}}
```
kranurag7, line 255: we can consider omitting this because it defaults to using the same.

kranurag7 (commented):
great tests addition, leaving some comments inline.

pgpool2-4.6.yaml
AnmolVirdi: Returns this error in case I omit it:
```
ERRO failed to test package: compiling pgpool2-4.6.yaml tests: compiling "" test pipelines: compiling Pipeline[0]: mutating with: variable context.name not defined
```

pgpool2-4.6.yaml
AnmolVirdi: Noted, I added these packages since they were recently merged into older `pgpool2` melange configs.

I've updated changes to omit them.

pgpool2-4.6.yaml
AnmolVirdi: Update, thanks

kranurag7 (approved):

debasishbsws (approved):

PR #51494, opened 2025-04-23
readline: fix ncurses linking

javacruft:
Ensure that --as-needed is disabled during linking of ncurses otherwise
the resulting readline so is underlinked and causes issues further
downstream with missing symbols.


sil2100 (approved):
Nice!

PR #51496, opened 2025-04-23
build boost with gcc-14

kranurag7:
as of now this is pulling in gcc-13

### DIFF
--- a/boost.yaml
+++ b/boost.yaml
@@ -1,7 +1,7 @@
 package:
   name: boost
   version: "1.88.0"
-  epoch: 0
+  epoch: 1
   description: "Free peer-reviewed portable C++ source libraries"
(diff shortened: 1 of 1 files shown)


justinvreeland (approved):

dannf:
Do we know what is causing the `ci-mal-report` failures?

justinvreeland:
Looks like it thinks we're trying to hide javascript values or code in our returns.  Since they're not javascript we don't need to worry about it.  There's a long list of things that can hide in javascript code.

eslerm:
Cheers @justinvreeland. malcontent's recent c094e73a ("Improve detection of malicious obfuscated Javascript, PowerShell and Lua (#866)") is a bit noisy atm.

PR #51516, opened 2025-04-23
New package crosstool-ng

xnox:
Add a new package that is able to build toolchains. This will help
packaging native & cross toolchains, with a declarative config without
need to use out of date boostrap-stageN archives.


dannf (dismissed):
Could we use `py3.13-build-base-dev` or similar to avoid installing a `python3`? It doesn't seem to need it.

dannf (approved):

PR #51528, opened 2025-04-23
openjdk-24 fix update-ca-certificates integration

xnox:
Removal of /etc/ssl/certs/java/cacerts is causing integration failure with update-ca-certificates and previous ways of updating cacerts.

It was done to support jlink without jmods for java 24 only in:

Create inverse symlink for the time being, and also correct incorrect symlink override in the jre packaging.

This should keep jlink working, if cacerts are not modified.

This should also restore support for adding custom cacerts to the keystore, using the same paths as on previous jdk versions and other distributions.

### DIFF
--- a/java-cacerts.yaml
+++ b/java-cacerts.yaml
@@ -2,7 +2,7 @@ package:
   name: java-cacerts
   # Update this when ca-certificates is updated.
   version: "20241121"
-  epoch: 2
+  epoch: 40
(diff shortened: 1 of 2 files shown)


EyeCantCU (approved):

xnox:
Note that once https://bugs.openjdk.org/browse/JDK-8353185 24.0.2 is released we can actually revert back to the symlinks how they were before.

PR #51573, opened 2025-04-23
New version stream: airflow-3

AmberArcadia:
Added a version stream for airflow-3, based on the previous package and adapted for the new version.

#### For new version streams

- [ ] If non-streamed package names no longer built, open PR to withdraw them (see [WITHDRAWING PACKAGES](https://github.com/chainguard-dev/enterprise-packages/blob/main/withdrawn-packages.txt))


airflow-3.yaml
```diff
@@ -0,0 +1,267 @@
+package:
+  name: airflow-3
+  version: "3.0.0"
+  epoch: 0
+  description: Platform to programmatically author, schedule, and monitor workflows
+  options:
+    #  There is a dependency on libarrow.so although it
+    #  is provided in the virtual environment. Enabling no-depends
+    #  works around this
+    no-depends: true
```
OddBloke, line 10: Have you confirmed this is still required?

airflow-3.yaml
```diff
@@ -0,0 +1,267 @@
...
+  dependencies:
+    runtime:
+      - coreutils # numfmt is required
+      - eudev-dev # for libudev.so.1
```
OddBloke, line 14: This should probably be either `eudev-libs` or `libudev`: https://apk.chaindag.dev/https/packages.wolfi.dev/os/x86_64/APKINDEX.tar.gz@etag:d215204967fe1ce42b8d441dc1a1f283/APKINDEX?full=false&search=&depend=&provide=so%3Alibudev.so.1

airflow-3.yaml
```diff
@@ -0,0 +1,267 @@
...
+  - uses: patch
+    with:
+      patches: urllib3.patch
+
+  - working-directory: ./airflow-core/src/airflow/ui
+    runs: |
+      # front-end build
+      yarn install --frozen-lockfile
+      # Update BrowsersList regularly: https://github.com/browserslist/update-db#readme
+      npx update-browserslist-db@latest
```
OddBloke, line 78: If I'm reading that README right, this will update the version of a particular package used for the build. Do we think there's enough benefit from doing this here to diverge from the version upstream ships?

airflow-3.yaml
```diff
@@ -0,0 +1,267 @@
...
+      yarn run build
+      rm -rf node_modules
+
+  - name: "Build and install the main package"
+    runs: |
+      # requires EPOCH to be later that 1980
```
OddBloke, line 84: What does?

airflow-3.yaml
```diff
@@ -0,0 +1,267 @@
...
+      export SOURCE_DATE_EPOCH=315532800
+
+      # To install mysqlclient wheel
+      export MYSQLCLIENT_CFLAGS=`mysql_config --cflags`
+      export MYSQLCLIENT_CFLAGS=`mysql_config --cflags`
```
OddBloke, line 89:
```suggestion
      export MYSQLCLIENT_CFLAGS=`mysql_config --cflags`
```

airflow-3.yaml
```diff
@@ -0,0 +1,267 @@
...
+
+  - name: "Build and install the additional provider packages"
+    runs: |
+      # By default, the airflow celery provider is not built, but running the upstream helm chart requires it
+      # There are a few others we want to include by default as well
+
+      cd providers
+      # If changing this list, be sure to update the test below too
+      # Use / as a separator
+      for pkg in amazon celery cncf/kubernetes docker elasticsearch google grpc hashicorp microsoft/azure mysql odbc openlineage postgres redis sendgrid sftp slack snowflake ssh; do
```
OddBloke, line 103: Should this list be a `var` so we don't have to keep it in sync manually?

airflow-3.yaml
```diff
@@ -0,0 +1,267 @@
...
+      wget https://www.openssl.org/source/old/1.1.1/openssl-1.1.1u.tar.gz
+      tar xzf openssl-1.1.1u.tar.gz
+      cd openssl-1.1.1u
+      ./config --prefix=${{targets.destdir}}/usr shared
+      make -j$(nproc)
+      make install_sw # Install only libraries & headers (skip docs)
+
+      # Remove extra links so things still default to the more modern version
+      rm ${{targets.destdir}}/usr/lib/libcrypto.so
+      rm ${{targets.destdir}}/usr/lib/libssl.so
```
OddBloke, line 147: OpenSSL 1.1.1 isn't receiving security support any longer, so I don't think we can do this. We either need to figure out a way to get it using a modern OpenSSL, or we'll have to drop support for this provider on ARM.

The MySQL database driver situation in Python is confusing (there are a bunch of different ones), but we can dig into it together.

OddBloke (changes requested):
Thanks for taking this on, I can tell it hasn't been easy! I have a few inline comments, mostly minor, but the biggest one is that we will need to figure out a path forward that doesn't involve OpenSSL 1.1.1.

airflow-3.yaml
AmberArcadia: Yes, without it, we get this when testing:
```
failed to test package: unable to build guest: unable to generate image: installing apk packages: error getting package dependencies: solving "airflow-3" constraint: resolving "airflow-3-3.0.0-r0.apk" deps:
solving "so:libarrow.so.2000" constraint: nothing provides "so:libarrow.so.2000"
```

airflow-3.yaml
AmberArcadia: I've seen this around a bunch over my years of development. My understanding is it's harmless but can have a bit of a benefit to do. I know we've got it in some of our pipelines already (including in airflow.yaml) and it's been fine so far.

airflow-3.yaml
AmberArcadia: I have no idea, that was from the previous version. It's present from the first commit to introduce the package, so I assume it's airflow itself.

airflow-3.yaml
AmberArcadia: Yes, but we need to use a / in one place and a - in another. I suppose some fancy transforms could handle that though...

airflow-3.yaml
AmberArcadia: `libudev` works so I'll switch to that.

airflow-3.yaml
AmberArcadia: Yep, got it working with a transform!

OddBloke (approved):

PR #51704, opened 2025-04-25
Pgpool2-4.6: Compat package addition and build fixes

AnmolVirdi:
- Added compile_error patch for `pgproto` build. [REF](https://github.com/pgpool/pgpool2_on_k8s/blob/master/pgpool.docker/fix_compile_error.patch)

- Added symlinks as well as startup files ([start.sh, entrypoint.sh](https://github.com/pgpool/pgpool2_on_k8s/tree/master/pgpool.docker)) for `pgpool2`


pgpool2-4.6.yaml
```diff
@@ -63,6 +68,45 @@ subpackages:
...
+          # copy etc to pgpool/etc.default
+          mkdir -p "${{targets.contextdir}}/opt/pgpool-II/etc.default"
+          cp -pr "${{targets.destdir}}/etc/pgpool2/." "${{targets.contextdir}}/opt/pgpool-II/etc.default"
+      - name: "Fetch and add startup scripts"
+        runs: |
+          # Add start.sh and entrypoint scripts same as upstream
+          git clone --depth 1 --branch master https://github.com/pgpool/pgpool2_on_k8s
+          install -Dm755 pgpool2_on_k8s/pgpool.docker/start.sh  ${{targets.contextdir}}/opt/pgpool-II/bin/start.sh
+          install -Dm755 pgpool2_on_k8s/pgpool.docker/entrypoint.sh ${{targets.contextdir}}/opt/pgpool-II/bin/entrypoint.sh
+
```
kranurag7, line 109: we should move this to a separate package and call it something like `${{package.name}}-oci-entrypoint` if it's entrypoint script. 

One package should ideally have only one source checkout.

pgpool2-4.6.yaml
```diff
@@ -63,6 +68,45 @@ subpackages:
...
 
+  - name: ${{package.name}}-compat
+    description: "Compatibility package for Upstream pgpool image"
+    dependencies:
+      provides:
+        - pgpool2-compat=${{package.full-version}}
+      runtime:
+        - bash
+        - coreutils
+        - busybox
```
kranurag7, line 79: I'll remove busybox in favor of coreutils, findutils, grep, gnutar, gawk etc.

if coreutils and bash are required for entrypoint script, this should be the runtime dependencies of the entrypoint package.

pgpool2-4.6.yaml
```diff
@@ -63,6 +68,45 @@ subpackages:
...
+            pcp_node_info pcp_pool_status pcp_proc_count pcp_proc_info pcp_promote_node \
+            pcp_recovery_node pcp_stop_pgpool pcp_watchdog_info \
+            pg_enc pg_md5 pgpool pgpool_setup pgproto watchdog_setup wd_cli \
+          ; do
+            ln -s "/usr/bin/${bin}" "${{targets.subpkgdir}}/opt/pgpool-II/bin/${bin}"
+          done
+
+          # copy etc to pgpool/etc
+          mkdir -p "${{targets.contextdir}}/opt/pgpool-II/etc"
+          cp -pr "${{targets.destdir}}/etc/pgpool2/." "${{targets.contextdir}}/opt/pgpool-II/etc/"
```
kranurag7, line 98: given at image level, we'll have both main and compat package, we should use ln here over cp.

pgpool2-4.6.yaml
```diff
@@ -63,6 +68,45 @@ subpackages:
...
+
+          # copy etc to pgpool/etc.default
+          mkdir -p "${{targets.contextdir}}/opt/pgpool-II/etc.default"
+          cp -pr "${{targets.destdir}}/etc/pgpool2/." "${{targets.contextdir}}/opt/pgpool-II/etc.default"
```
kranurag7, line 102: don't see this in the upstream one, is this required? 

```bash
$ docker run --rm -it --entrypoint=bash pgpool/pgpool
9ae25c3b6074:/$ ls  /opt/pgpool-II/
bin      etc      include  lib      share
```

pgpool2-4.6.yaml
```diff
@@ -35,6 +35,11 @@ pipeline:
       tag: V${{vars.mangled-package-version}}
       expected-commit: 7bcfd1ffec63105c177d2e42202a520373960b2f
 
+  # Patch to fix compilation errors for pgproto
+  - uses: patch
+    with:
+      patches: /home/build/fix_compile_error.patch
```
kranurag7, line 41: has something changed here? it passed recently when we merged it.

pgpool2-4.6.yaml
AnmolVirdi:
```
docker run --rm -it --entrypoint=bash pgpool/pgpool:4.0.23                                             
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
eda621c0c8d1:/$ ls opt/pgpool-II/
bin/     etc/     include/ lib/     share/

File: pgpool2-4.6.yaml (hunk 1 above)
Comment (AnmolVirdi) [reply pgpool2-4.6.yaml:109]: Noted, moving this to separate subpackage

File: pgpool2-4.6.yaml (hunk 2 above)
Comment (AnmolVirdi) [reply pgpool2-4.6.yaml:79]: Noted, moving these packages to entrypoint subpackage.

File: pgpool2-4.6.yaml (hunk 4 above)
Comment (AnmolVirdi) [reply pgpool2-4.6.yaml:102]: Upstream dockerfile:
https://github.com/pgpool/pgpool2_on_k8s/blob/master/pgpool.docker/Dockerfile.pgpool

File: pgpool2-4.6.yaml [hunk 6]
```diff
@@ -63,6 +68,59 @@ subpackages:
...
 
+  - name: ${{package.name}}-compat
+    description: "Compatibility package for Upstream pgpool image"
+    dependencies:
+      provides:
+        - pgpool2-compat=${{package.full-version}}
+      runtime:
+        - postgresql
+        - postgresql-client
+        - postgresql-dev
```
EyeCantCU, line 79: Concerning that the Postgres version will just change under people's feet, but upstream appears to be okay with that as they're doing the same.

EyeCantCU, line 79: Not convinced we need development headers for Postgres for pgpool. I'd omit these (even if upstream includes them) as they're useless outside of a build env

pgpool2-4.6.yaml
```diff
@@ -63,6 +68,59 @@ subpackages:
...
+    pipeline:
+      - name: "Symlink all pgpool binaries to /opt/pgpool-II/bin"
+        runs: |
+          mkdir -p ${{targets.subpkgdir}}/opt/pgpool-II/bin
+          for bin in \
+            pcp_attach_node pcp_detach_node pcp_health_check_stats pcp_node_count \
+            pcp_node_info pcp_pool_status pcp_proc_count pcp_proc_info pcp_promote_node \
+            pcp_recovery_node pcp_stop_pgpool pcp_watchdog_info \
+            pg_enc pg_md5 pgpool pgpool_setup pgproto watchdog_setup wd_cli \
+          ; do
```
EyeCantCU, line 89: We could do this instead:

```
for bin in ${{targets.destdir}}/usr/bin/; do....
```

pgpool2-4.6.yaml
```diff
@@ -63,6 +68,59 @@ subpackages:
...
+
+          # Link sample files
+          mkdir -p "${{targets.contextdir}}/opt/pgpool-II/etc"
+          # Create symlinks from /etc/pgpool2/* to /opt/pgpool-II/etc/
+          for bin in \
+            aws_eip_if_cmd.sh.sample follow_primary.sh.sample pool_hba.conf.sample \
+            aws_rtb_if_cmd.sh.sample pcp.conf.sample recovery_1st_stage.sample escalation.sh.sample \
+            pgpool_remote_start.sample replication_mode_recovery_1st_stage.sample failover.sh.sample \
+            pgpool.conf.sample replication_mode_recovery_2nd_stage.sample \
+          ; do
```
EyeCantCU, line 101: `for sample in ${{targets.destdir}}/etc/pgpool2/; do...`

pgpool2-4.6.yaml
```diff
@@ -63,6 +68,59 @@ subpackages:
...
+        - bash
+        - coreutils
+        - grep
+        - gawk
+        - gnutar
+    pipeline:
+      - name: "Fetch and add startup scripts"
+        runs: |
+          # Add start.sh and entrypoint scripts same as upstream
+          git clone --depth 1 --branch master https://github.com/pgpool/pgpool2_on_k8s
```
EyeCantCU, line 120: We should be using `git-checkout` here, even if this is a subpackage, so that our packages are reproducible. I would split this into a separate package manifest personally so we can track updates

EyeCantCU (changes requested):

pgpool2-4.6.yaml
AnmolVirdi: Noted, updated PR accordingly.

pgpool2-4.6.yaml
AnmolVirdi: Thanks @EyeCantCU, updated PR accordingly.

pgpool2-4.6.yaml
AnmolVirdi: I've updated the PR with git-checkout
The upstream image is no longer maintained and these entrypoint files got last update 3 years ago. 
https://github.com/pgpool/pgpool2_on_k8s/blob/master/pgpool.docker/entrypoint.sh
Fortunately, `pgpool2` behaves the same and these scripts work well.

pgpool2-4.6.yaml
AnmolVirdi: Agreed, I've moved `postgres` (17) to main package's runtime deps as per your suggestion.

EyeCantCU (approved):

pgpool2-4.6.yaml
AnmolVirdi: Updated the PR, thanks

PR #51718, opened 2025-04-25
fix: crossplane-provider-* repository path is under the crossplane-contrib org

hectorj2f:
Fixes: The repository path of the crossplane provider repos are now under crossplane-contrib organization.

### DIFF
--- a/crossplane-provider-aws.yaml
+++ b/crossplane-provider-aws.yaml
@@ -1,7 +1,7 @@
 package:
   name: crossplane-provider-aws
   version: "1.21.1"
-  epoch: 1
+  epoch: 2
   description: Official AWS Provider for Crossplane by Upbound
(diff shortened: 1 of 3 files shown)


antitree:
Malcontent fp. This appears to be triggered from https://github.com/crossplane-contrib/provider-upjet-azure/blob/78491d4e53681d712a460b9b849b67bf078ae0ea/examples/insights/v1beta1/monitorautoscalesetting.yaml#L87 or the other example files

powersj (approved):

Aasrishcflt:
Hey Team,
Probably this should have also been updated along with the above change, https://github.com/wolfi-dev/os/blob/main/crossplane-provider-aws.yaml#L92C17-L92C37 
Thanks!

PR #51763, opened 2025-04-25
add nextcloud-server

Dentrax:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/nextcloud-server.yaml
+++ b/nextcloud-server.yaml
@@ -1,7 +1,7 @@
 #nolint:valid-pipeline-git-checkout-tag
 package:
   name: nextcloud-server
-  version: 31.0.4
+  version: 31.0.5
   epoch: 0
   description: "Nextcloud server, a safe home for all your data"
   copyright:
@@ -118,7 +118,7 @@ pipeline:
   - uses: fetch
     with:
       uri: https://download.nextcloud.com/server/releases/nextcloud-${{package.version}}.tar.bz2
-      expected-sha256: a47541566d5c6ac6f63e4f617e27da295156da47daa2cd22eee3400fd2ad1251
+      expected-sha256: 2228b8f524dcd87f08a2a9281d41b04f5cb500b7624bc8c8e07b8e3039061503
       directory: ${{targets.destdir}}/usr/src/nextcloud
       delete: true
 
--- a/nextcloud-server.yaml
+++ b/nextcloud-server.yaml
@@ -140,6 +140,7 @@ pipeline:
     runs: |
       # Upstream image uses /usr/bin/apache2, but we have /usr/bin/httpd instead, both are the same
       # https://github.com/docker-library/php/blob/d21ab07e7f4014a5443f47c8ff8e292f9c703a58/apache2-foreground#L40
+      mkdir -p ${{targets.destdir}}/usr/bin
       ln -sf /usr/bin/httpd ${{targets.destdir}}/usr/bin/apache2
 
 update:
--- a/nextcloud-server.yaml
+++ b/nextcloud-server.yaml
@@ -34,6 +34,7 @@ package:
       - php-${{vars.php-version}}-bcmath
       - php-${{vars.php-version}}-exif
       - php-${{vars.php-version}}-ftp
+      - php-${{vars.php-version}}-fpm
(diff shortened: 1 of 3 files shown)


nextcloud-server.yaml
```diff
@@ -0,0 +1,388 @@
...
+      runs: |
+        for f in /cron.sh /entrypoint.sh /upgrade.exclude; do
+          stat ${f}
+          [ -x "$f" ]
+        done
+        find /usr/src/nextcloud/config/ -type f -name "*.php" | grep -q .
+    - name: Check http.conf
+      runs: |
+        # Check if the httpd.conf file is valid
+        cat /etc/apache2/httpd.conf | grep "ServerRoot"
```
jamonation, line 287: This isn't a valid check:

```
2a5c21bd78a2:/work/packages# grep ServerRoot /etc/apache2/httpd.conf
# with "/", the value of ServerRoot is prepended -- so "logs/access_log"
# with ServerRoot set to "/usr/local/apache2" will be interpreted by the
# ServerRoot: The top of the directory tree under which the server's
# ServerRoot at a non-local disk, be sure to specify a local disk on the
# same ServerRoot for multiple httpd daemons, you will need to change at
ServerRoot "/"
```

Skip using cat and use grep directly is one thing that will simplify looking for strings.

But if testing the config file's validity is the goal, use `apachectl -S` to parse the entire config, with includes, enabled modules etc. e.g.

```
2a5c21bd78a2:/work/packages# apachectl -S
httpd: Syntax error on line 31 of /etc/apache2/httpd.conf: ServerRoot must be a valid directory
```

nextcloud-server.yaml
```diff
@@ -0,0 +1,388 @@
...
+    PHP_UPLOAD_LIMIT: "512M"
+    PHP_OPCACHE_MEMORY_CONSUMPTION: "128"
+    PHP_INI_DIR: "etc/php"
+
+pipeline:
+  - runs: mkdir -p ${{targets.destdir}}/usr/src/nextcloud
+
+  - uses: fetch
+    with:
+      uri: https://download.nextcloud.com/server/releases/nextcloud-${{package.version}}.tar.bz2
```
jamonation, line 120: Curious why this is using fetch instead of git clone from https://github.com/nextcloud/server ?

Dentrax: IIRC they release their source code by omitting other lots of unnecessary stuff (that we don't need), and since PHP is not a compiled lang. The tarball should be exactly same in the GitHub, so this is OK.

Dentrax: done!

nextcloud-server.yaml
Dentrax: Thanks so much! I've updated the test and added a validation step with `apachectl`!

nextcloud-server.yaml
```diff
@@ -0,0 +1,401 @@
...
+    post-install: |
+      #!/bin/sh
+      # As wolfi-baselayout does provide /var/spool/cron already, and we can not create this
+      # directory in the package, we need to create the cron file in the post-install scriptlet.
+      # Since scriptlets don't run in apko, this is for the `apk add` command only.
+      mkdir -p /var/spool/cron/crontabs || true
+      echo '*/5 * * * * php -f /var/www/html/cron.php' > /var/spool/cron/crontabs/www-data
+
+vars:
+  php-version: 8.1
```
jamonation, line 85: Their docs note that PHP 8.1 is deprecated: Can this be built with with 8.3 as recommended? https://docs.nextcloud.com/server/latest/admin_manual/installation/system_requirements.html

Dentrax: Nice catch! https://github.com/wolfi-dev/os/pull/51763/commits/608fc0a7877c3d1728eb9a6c7cf5c3ef23a4fbfe

eslerm:
@Dentrax what input are you looking for from Security?

Dentrax:
> @Dentrax what input are you looking for from Security?

Pushed a new commit, so malcontent/reviewed label dismissed, PTAL - thanks.

eslerm (approved):

PR #51764, opened 2025-04-25
kernel follow up fixes

dannf:


smoser (approved):
can you add kernel/ to .gitignore also?

smoser:
> can you add kernel/ to .gitignore also?

doh. you did. thanks.

PR #35681, opened 2024-12-03
Berkeley GenAI Club: Add Autogenerated Test Pipelines for Kubernetes Ingress Defaultbackend, Terraform Provider Google, and Hugo-Extended

Vinay-Vinod:
### Summary
This PR introduces autogenerated test pipelines for the following packages:

kubernetes-ingress-defaultbackend: Adds tests to validate the functionality of the 404-server, including server startup, health check endpoint, and 404 response validation.
terraform-provider-google: Includes tests for provider installation and basic Terraform initialization using the Google provider.
hugo-extended: Incorporates tests to check the Hugo version, create a new Hugo site, and validate successful site builds.
The tests were generated using a Go gptscript, ensuring comprehensive coverage and consistent behavior validation across these packages.

Fixes:
Used gptscript to generate test pipelines

Related:

#### For new package PRs only

- [ ] This PR is marked as fixing a pre-existing package request bug
  - [ ] Alternatively, the PR is marked as related to a pre-existing package request bug, such as a dependency
- [ ] REQUIRED - The package is available under an OSI-approved or FSF-approved license
- [ ] REQUIRED - The version of the package is still receiving security updates
- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`)

#### For new version streams

- [ ] The upstream project actually supports multiple concurrent versions.
- [ ] Any subpackages include the version string in their package name (e.g. `name: ${{package.name}}-compat`)
- [ ] The package (and subpackages) `provides:` logical unversioned forms of the package (e.g. `nodejs`, `nodejs-lts`)
- [ ] If non-streamed package names no longer built, open PR to withdraw them (see [WITHDRAWING PACKAGES](https://github.com/wolfi-dev/os/blob/main/WITHDRAWING_PACKAGES.md))

#### For package updates (renames) in the base images

When updating packages part of base images (i.e. cgr.dev/chainguard/wolfi-base or ghcr.io/wolfi-dev/sdk)
- [ ] REQUIRED cgr.dev/chainguard/wolfi-base and ghcr.io/wolfi-dev/sdk images successfully build
- [ ] REQUIRED cgr.dev/chainguard/wolfi-base and ghcr.io/wolfi-dev/sdk contain no obsolete (no longer built) packages
- [ ] Upon launch, does `apk upgrade --latest` successfully upgrades packages or performs no actions

#### For security-related PRs

- [ ] The security fix is recorded in the [advisories](https://github.com/wolfi-dev/advisories) repo

#### For version bump PRs

[...]

PR #51782, opened 2025-04-25
Fix some tests that were merged but didn't run

murraybd:
The PR in https://github.com/wolfi-dev/os/pull/35681 never actually ran the tests because it was submitted before we started running tests on pre-submit. After I ran them locally (to switch from using OK to check for 200) I discovered a couple of them failed. This fixes that.

### DIFF
--- a/kubernetes-ingress-defaultbackend.yaml
+++ b/kubernetes-ingress-defaultbackend.yaml
@@ -1,7 +1,7 @@
 package:
   name: kubernetes-ingress-defaultbackend
   version: "1.34.2"
-  epoch: 0
+  epoch: 1
   description: 'A simple web server that respond 404 common used in kubernetes ingress, serve pages 404 at root and 200 at /healthz'
   copyright:
     - license: Apache-2.0
--- a/kubernetes-ingress-defaultbackend.yaml
+++ b/kubernetes-ingress-defaultbackend.yaml
@@ -51,7 +51,7 @@ test:
         sleep 5
 
         # Use a retry mechanism to check if the health endpoint is available
-        for i in {1..5}; do
+        for i in $(seq 1 5); do
           if curl -s -o /dev/null -w "%{http_code}" http://localhost:8080/healthz | grep -q "^200$"; then
             echo "Server is up"
             break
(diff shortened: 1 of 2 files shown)


kubernetes-ingress-defaultbackend.yaml
```diff
@@ -52,7 +52,7 @@ test:
 
         # Use a retry mechanism to check if the health endpoint is available
         for i in {1..5}; do
-          if curl -s http://localhost:8080/healthz | grep "OK"; then
+          if curl -s -o /dev/null -w "%{http_code}" http://localhost:8080/healthz | grep -q "^200$"; then
```
dannf, line 55: nice, i didn't know you could get the exit code that way

kubernetes-ingress-defaultbackend.yaml
```diff
@@ -62,7 +62,16 @@ test:
...
-        killall defaultbackend
+        defaultbackend > defaultbackend.log 2>&1 &
+
+        sleep 5
+
+        for i in {1..5}; do
+          if curl -s -o /dev/null -w "%{http_code}" http://localhost:8080/ | grep -q "^404$"; then
+            echo "Server 404ed"
+            break
+          else
```
dannf, line 73: (nit - feel free to ignore) -  I think you could `exit 0` instead of `break` above to simplify the logic. That would let you close the if here and hoist the "Server not ready" bit out w/o the need for an else.

dannf (changes requested):

kubernetes-ingress-defaultbackend.yaml
```diff
@@ -62,7 +62,16 @@ test:
...
-      runs: |
-        killall defaultbackend
+        defaultbackend > defaultbackend.log 2>&1 &
+
+        sleep 5
+
+        for i in {1..5}; do
+          if curl -s -o /dev/null -w "%{http_code}" http://localhost:8080/ | grep -q "^404$"; then
+            echo "Server 404ed"
+            break
```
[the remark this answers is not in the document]
murraybd: The service is supposed to return a 404 for anything but healthz. https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/default-backend.md

kubernetes-ingress-defaultbackend.yaml
murraybd: The original PR was AI generated but let's make it better.

kubernetes-ingress-defaultbackend.yaml
murraybd: Really, its in the help for `retry-autopkgtest-regressions`. ;-)

dannf (approved):

PR #51794, opened 2025-04-25
open-webui: new package

MickeyPvX:
Building open-webui package.

Remaining CVEs will need advisories; the latest package versions in `pyodide` are shown here: https://pyodide.org/en/stable/usage/packages-in-pyodide.html - note `requests` and `pillow` versions.

#### For new package PRs only


egibs:
Malcontent findings look good. The findings are all from comments.

open-webui.yaml
```diff
@@ -0,0 +1,105 @@
...
+      - libxext
+      - python-${{vars.py-version}}-base
+  options:
+    # Don't depend on external libraries included in virtual environments
+    no-depends: true
+    # Don't resolve libraries installed in virtual environments as providers
+    no-provides: true
+
+vars:
+  py-version: 3.11
```
Dentrax, line 30: Can we have a comment above why we pinned 3.11 here?

open-webui.yaml
```diff
@@ -0,0 +1,105 @@
...
+  pipeline:
+    - uses: test/daemon-check-output
+      with:
+        expected_output: |
+          Creating knowledge table
+          v${{package.version}} - building the best AI user interface.
+          Started server process
+          Waiting for application startup.
+        start: open-webui serve
+        timeout: 300
```
Dentrax, line 171: Let's have a `post:` field here, ensure port is up with `wait-for-it` and with a simple `curl` test to `grep` web content?

open-webui.yaml
```diff
@@ -0,0 +1,105 @@
...
+      cp -r venv/* ${{targets.contextdir}}/usr/share/${{package.name}}/
+
+      # Symlink binaries
+      mkdir -p ${{targets.contextdir}}/usr/bin
+      for bin in ${{targets.contextdir}}/usr/share/${{package.name}}/bin/*
+      do
+        # Don't link venv python - clashes with runtime python dep
+        if [[ $bin != *"python"* ]]
+        then
+          ln -sf "/usr/share/${{package.name}}/bin/${bin##*/}" "${{targets.contextdir}}/usr/bin/${bin##*/}"
```
Dentrax, line 103: Can we have a `readlink` test to ensure we properly symlink-ed each file in the `${{targets.contextdir}}/usr/share/${{package.name}}/bin/*`.

open-webui.yaml
```diff
@@ -0,0 +1,105 @@
...
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: 53764fe64884da147359e54ed6d9607fe57f1600
+      repository: https://github.com/open-webui/open-webui
+      tag: v${{package.version}}
+
+  - uses: patch
+    with:
+      patches: open-webui-dependency-update.patch
```
Dentrax: Should we have a comment above why we have this patch file?

Dentrax (dismissed):

open-webui.yaml
MickeyPvX: It was intended to try and mitigate the now remaining 2 CVEs by upgrading `pyodide` but even the latest version still has them.  I'm removing the patch as it seems pointless, and will have to file some advisories it looks like.

egibs:
> Malcontent findings look good. The findings are all from comments.

The new findings are also false positives and I have a PR open to address them.

jamonation (approved):

PR #51851, opened 2025-04-27
Add and expand test coverage for various ruby packages

mamccorm:
Adds and expand test coverage for various ruby packages

### DIFF
--- a/ruby3.2-openid_connect-1.1.8.yaml
+++ b/ruby3.2-openid_connect-1.1.8.yaml
@@ -1,7 +1,7 @@
 package:
   name: ruby3.2-openid_connect-1.1.8
   version: 1.1.8
-  epoch: 7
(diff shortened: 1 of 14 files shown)


ruby3.2-openid_connect-1.1.8.yaml
kranurag7, line 1: looks like got skipped on this one. copy/pasting 3.3 one should work (thinking...) but not a blocker.

kranurag7 (approved):

ruby3.2-openid_connect-1.1.8.yaml
mamccorm: Intersting, thanks good spot. I'll see if I can tune claude.md to make sure it doesn't skip, in the next batch of updates :)

PR #51872, opened 2025-04-28
Libqcow package addition - Includes subpackage for python bindings

AnmolVirdi:
**Libqcow package** `Current version: 20240308`
- Library to access the QEMU Copy-On-Write (QCOW) image file format

**Type**: C library
**REF**: https://github.com/libyal/libqcow

**Tests:**
- ldd-check

### DIFF
--- a/libqcow.yaml
+++ b/libqcow.yaml
@@ -36,7 +36,7 @@ pipeline:
   - uses: autoconf/configure
 
   - uses: autoconf/make
-  
+
   - uses: autoconf/make-install
 
   - uses: strip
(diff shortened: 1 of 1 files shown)


kranurag7 (approved):

libqcow.yaml
```diff
@@ -0,0 +1,62 @@
+package:
+  name: libqcow
+  version: "20240308"
+  epoch: 1
```
kranurag7, line 4: epoch should be 0 for new packages but given it's in now, we can keep it as it is.

PR #51892, opened 2025-04-28
ceph: add jq to runtime dependencies

javacruft:
Aligns with upstream packages and use by rook for liveness
checks of the Ceph MDS daemon.


dannf (approved):
I'm pretty sure I've seen this `ci-cve-scan` issue before and it was a no-op. Really would be nice to have a way to override these.

sil2100 (approved):

PR #51921, opened 2025-04-28
fix python fips detection and blocking of blake2

techalchemy:
- fixes chainguard-dev/internal-dev#11886
- backports from wolfi-dev/os#36503 (see python/cpython#127301)

### DIFF
--- a/python-3.11.yaml
+++ b/python-3.11.yaml
@@ -1,7 +1,7 @@
 package:
   name: python-3.11
   version: "3.11.12"
-  epoch: 0
+  epoch: 1
   description: "the Python programming language"
   copyright:
     - license: PSF-2.0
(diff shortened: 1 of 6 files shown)


python-3.11.yaml
```diff
@@ -57,7 +57,7 @@ pipeline:
 
   - uses: patch
     with:
```
xnox, line 59: please rebase and do an epoch bump, as python 3.11 did not get built.

xnox (dismissed):
3.10 & 3.12 work correctly as described now.
3.11 needs an epoch bump.

techalchemy:
should be good to go @xnox

xnox (approved):

PR #51956, opened 2025-04-28
kubeflow-pipelines-visualization-server/GHSA-vqfr-h8mv-ghfj fix

jamie-albert:
bumping both the parent dependency httpcore and h11 as h11 v0.16.0 is not compatible with httpcore 1.0.7


jamie-albert:
kubeflow-pipelines-visualization-server v2.5.0 is currently in package build failure queue, this will require merging still to remediate CVEs

stevebeattie:
Malcontent failure is a whole bunch of false positives.

dnegreira (approved):

PR #51959, opened 2025-04-29
feat(pipelines): Configure gradle to use Libraries

EyeCantCU:
Creates init script that injects the repository for Libraries, and disables mavenCentral

### DIFF
--- a/pdftk.yaml
+++ b/pdftk.yaml
@@ -2,7 +2,7 @@
 package:
   name: pdftk
   version: 3.3.3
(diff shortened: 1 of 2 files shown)


dannf (approved):
LGTM. One (non-blocking) question though - should we be using `libraries` instead of `ecosystems`?

EyeCantCU:
@dannf very good point. The only reason I've used ecosystems here is for consistency, but I feel that would make more sense overall for pipelines like these

PR #52093, opened 2025-04-30
Remove extraneous setuid bits from man-db and sudo

### LINKED ISSUE #52036: sudo: gratuitous setuid permissions on /etc/sudo*
I can't think of any exploit for this, but while poking around I noticed some gratuitous setuid permissions in the sudo package:

* /etc/sudo.conf is 4644, should be 0644
* /etc/sudoers is 4440, should be 0440

I believe these were errantly added in https://github.com/wolfi-dev/os/blame/45b75807a8191c4f2cbe7d2a132ad82d161c7c6c/sudo.yaml#L54 - I believe you can just remove the two lines adding `u+s` on the config files.

I did confirm that `sudo` still works if you change the permissions appropriately.

cc @EyeCantCU

### LINKED ISSUE #52037: man-db: gratuitous setuid/setgid on /bin/man
While poking around Wolfi I noticed that /bin/man was setuid/setgid root (6755).

While I've seen `setgid` to a `man` group before, I've never seen `man` setuid root. While I don't have a specific security vulnerability in mind, this may make the `man-db` package unnecessarily susceptible to a buffer-overflow or logic flaw. 

For comparison, in Alpine and Debian, `/bin/man` is 0755. I tried setting the same in my Wolfi environment and the man command still appeared to function.

I'm also not sure where this permission is coming in.

egibs:
Closes: #52036
Closes: #52037

`man-db` can be built without `setuid` and we can tweak the `/etc/sudo*` permissions.

### DIFF
--- a/man-db.yaml
+++ b/man-db.yaml
@@ -2,7 +2,7 @@
 package:
   name: man-db
   version: 2.13.0
-  epoch: 40
+  epoch: 41
   description: The man command and related utilities for examining on-line help files
   copyright:
     - license: GPL-3.0
@@ -36,7 +36,8 @@ pipeline:
   - uses: autoconf/configure
     with:
       opts: |
-        --sbindir=/usr/bin
+        # By default, man will be installed as a setuid program
+        --sbindir=/usr/bin --disable-setuid
 
   - uses: autoconf/make
 
@@ -71,6 +72,11 @@ test:
       packages:
         - man-db-doc
   pipeline:
+    - name: Ensure that setuid bits are not set
+      runs: |
+        set -x
+        if [ -u /usr/bin/man ]; then exit 1; fi
+        if [ -u /bin/man ]; then exit 1; fi
     - runs: |
         find /usr/share/man/ -type f -name "*.[0-9]" | xargs -i man -Tdvi -l {} >/dev/null
         apropos --version
--- a/man-db.yaml
+++ b/man-db.yaml
@@ -36,7 +36,6 @@ pipeline:
   - uses: autoconf/configure
     with:
       opts: |
-        # By default, man will be installed as a setuid program
         --sbindir=/usr/bin --disable-setuid
 
   - uses: autoconf/make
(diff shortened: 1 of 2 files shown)


man-db.yaml
```diff
@@ -71,6 +71,11 @@ test:
       packages:
         - man-db-doc
   pipeline:
+    - name: Ensure that setuid bits are not set
+      runs: |
+        set -x
+        if [ -u /usr/bin/man ]; then exit 1; fi
+        if [ -u /bin/man ]; then exit 1; fi
```
stevebeattie, line 78: Also, please double check the `/usr/bin/mandb` binary:

```
2a4cfbc6f0f4:/work/packages# ls -la /usr/bin/ | grep -v -E '^(l|d|-rwx)'
total 7536
-rwsr-sr-x    1 man      man         148792 Mar 31 16:07 man
-rwsr-sr-x    1 man      man         176712 Mar 31 16:07 mandb
```

stevebeattie (changes requested):

man-db.yaml
egibs: Added in [`dce3443` (#52093)](https://github.com/wolfi-dev/os/pull/52093/commits/dce3443e5501643c3f15c76dc91def12e6902048).

sudo.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: sudo
   version: 1.9.16
-  epoch: 40
+  epoch: 41
```
stevebeattie, line 4: Given that we're incorporating a new version, the `epoch` should be reset to 0.

egibs: That's a bit tricky. We can't go back to 0 for the same version and `1.9.16_p2` isn't valid [yet].

stevebeattie: ohhh I see. Weird. It looks like the wolfi version got bumped to 1.9.16 but it was still pulling the 1.9.15p5 tarball. :(

stevebeattie (approved):
LGTM, thanks.

PR #52125, opened 2025-04-30
chore(ceph): bump ceph epoch to verify arrow reliance

kwmonroe:
check ceph ci with apache-arrow 0.19


kwmonroe:
killing this; we've reproducibly confirmed ceph19 can't build as-is with arrow20 deps.

PR #52156, opened 2025-04-30
config(cve-pr-closer): adding octosts policy for cve-pr-closer bot

tdunlap607:
Octosts policy for cve-pr-closer bot to close stale request-cve-remediation PRs.

### DIFF
--- a/.github/chainguard/lifecycle-cve-pr-closer.sts.yaml
+++ b/.github/chainguard/lifecycle-cve-pr-closer.sts.yaml
@@ -1,0 +1,9 @@
+issuer: https://accounts.google.com
+
+# lc-cve-pr-closer-cron@prod-enforce-fabc.iam.gserviceaccount.com
+claim_pattern:
+  email: 'lc-cve-pr-closer-cron@prod-enforce-fabc\.iam\.gserviceaccount\.com'
+
+permissions:
+  contents: read
+  pull_requests: write


vaikas (approved):

.github/chainguard/lifecycle-cve-pr-closer.sts.yaml
```diff
@@ -0,0 +1,9 @@
+issuer: https://accounts.google.com
+
+# lc-cve-pr-closer-cron@prod-enforce-fabc.iam.gserviceaccount.com
+claim_pattern:
+  email: 'lc-cve-pr-closer-cron@prod-enforce-fabc\.iam\.gserviceaccount\.com'
```
imjasonh, line 5: It's a tiny thing, but elsewhere we require the exact SA ID, and comment what its name is.

https://github.com/wolfi-dev/os/blob/main/.github/chainguard/ci-diff-report.sts.yaml

I don't think this was for any strong technical/ideological reason, but it would be good to be consistent at leaest.

imjasonh (approved):

.github/chainguard/lifecycle-cve-pr-closer.sts.yaml
tdunlap607: I see, I'll create a second PR for the SA ID once I get it. Thank you!

vaikas: I'm fine with that being the protocol but curious about the reasons for it, as above is self documenting. Just a note.

PR #52253, opened 2025-05-01
fix(vector): add openssl fixes to Cargo.toml

mjramirez-dev:
Fixes:

Some inline fixes are added via sed to the `Cargo.toml` file during package build, to allow vector to properly load the system provided openssl instead of vendored version.

### DIFF
--- a/vector.yaml
+++ b/vector.yaml
@@ -36,10 +36,6 @@ pipeline:
   - runs: |
       sed -i '7s/deny/allow/' src/lib.rs
       export RUSTFLAGS="$RUSTFLAGS --cap-lints=warn"
-      # Needed for vector to load our own openssl properly w/ FIPS
-      # Vector was ignoring our openssl.cnf and fipsmodule.cnf
-      # even when configured with proper OPENSSL_CONF and OPENSSL_MODULE env variables
-      # REF: https://vector.dev/docs/reference/configuration/tls/#fips-provider-example
       # Force building with system provided openssl instead of vendored version
       sed -i -E 's|^(openssl)\s*=.*version\s*=\s*"([0-9]+\.[0-9]+\.[0-9]+).*|openssl = { version = "0.10.72", default-features = false }|' Cargo.toml
       sed -i -E 's|^(openssl-src)\s*=.*version\s*=\s*"([0-9]+).*|openssl-src = { version = "300", optional = true, default-features = false, features = ["force-engine", "legacy"] }|' Cargo.toml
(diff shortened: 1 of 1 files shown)


vector.yaml
jamonation, line 1: I think there's probably a cleaner way than sed across the whole lines - if/when there's a version bump then these won't match and replace the options. Likely just need a capture group or two with regex. Also the epoch should be 1

vector.yaml
```diff
@@ -35,7 +36,13 @@ pipeline:
   - runs: |
       sed -i '7s/deny/allow/' src/lib.rs
       export RUSTFLAGS="$RUSTFLAGS --cap-lints=warn"
-      make build
+      # Needed for vector to load our own openssl properly w/ FIPS
+      # Vector was ignoring our openssl.cnf and fipsmodule.cnf
+      # even when configured with proper OPENSSL_CONF and OPENSSL_MODULE env variables
+      # REF: https://vector.dev/docs/reference/configuration/tls/#fips-provider-example
```
jamonation, line 42:
```suggestion
      # Force building with system provided openssl instead of vendored version
```

vector.yaml
```diff
@@ -35,7 +36,14 @@ pipeline:
...
```
jamonation, line 42:
```suggestion
```

vector.yaml
```diff
@@ -35,7 +36,10 @@ pipeline:
   - runs: |
       sed -i '7s/deny/allow/' src/lib.rs
       export RUSTFLAGS="$RUSTFLAGS --cap-lints=warn"
-      make build
+      # Force building with system provided openssl instead of vendored version
+      sed -i -E 's|^(openssl)\s*=.*version\s*=\s*"([0-9]+\.[0-9]+\.[0-9]+).*|openssl = { version = "0.10.72", default-features = false }|' Cargo.toml
```
jamonation, line 40: This is pinning to 0.10.72, whereas the capture group should be used here to always write back the regex detected version

vector.yaml
```diff
@@ -35,7 +36,10 @@ pipeline:
...
+      sed -i -E 's|^(openssl-src)\s*=.*version\s*=\s*"([0-9]+).*|openssl-src = { version = "300", optional = true, default-features = false, features = ["force-engine", "legacy"] }|' Cargo.toml
```
jamonation, line 41: Likewise the version here is hard coded, and should be using the capture group

jamonation (approved):

PR #52335, opened 2025-05-02
man-db: use direct download link instead of mirrors

OddBloke:
Per the footer on https://download.savannah.nongnu.org/releases/man-db/, mirrors can lag up to 24 hours; this URL is given as the workaround for this lag.

(I'll let automation create the update PR, as package builds are currently broken pending a production deployment.)


AmberArcadia:
@OddBloke looks like things went a bit stale

powersj (approved):

PR #52430, opened 2025-05-05
golem package build

developer-guy:
Fixes:

Related:

#### For new package PRs only


golem.yaml
```diff
@@ -0,0 +1,86 @@
...
+      repository: https://github.com/golemcloud/golem
+      tag: v${{package.version}}
+      expected-commit: 51d5cbb1b9fcca357561c32577998ef74ce30f3d
+
+  - name: Configure and build
+    runs: |
+      rustup install stable
+      rustup target add wasm32-wasip1
+      rustup target add wasm32-unknown-unknown
+
```
EyeCantCU, line 37: Did we try our rust?

golem.yaml
```diff
@@ -0,0 +1,86 @@
...
+  ignore-regex-patterns:
+    - "-dev.*"
+  github:
+    identifier: golemcloud/golem
+    strip-prefix: v
+
+test:
+  pipeline:
+    # AUTOGENERATED
+    - runs: stat /usr/bin/golem-component-service
```
EyeCantCU, line 86: Can we expand tests to cover functional usage?

developer-guy: so this is not a complete golem package (yet) we just need to build one of the components of golem named golem-component-service so we will need them all to be able to provide some good functionality test so as soon as we get a request for other services of golem I'll enhance the tests for sure

golem.yaml
```diff
@@ -0,0 +1,87 @@
...
+      expected-commit: 51d5cbb1b9fcca357561c32577998ef74ce30f3d
+
+  - name: Configure and build
+    runs: |
+      # rustup install stable
+      # rustup target add wasm32-wasip1
+      # rustup target add wasm32-unknown-unknown
+
+      # # Put our rustup-installed toolchains into $PATH
+      # export PATH="$(dirname "$(rustup which cargo)"):$PATH"
```
Dentrax, line 40: Let's comment-out or omit entirely.

golem.yaml
```diff
@@ -0,0 +1,87 @@
...
+      environment:
+        contents:
+          packages:
+            - ${{package.name}}
+      pipeline:
+        - uses: test/tw/ldd-check
+        - name: Test symlinks
+          runs: |
+            set -e
+            test -f $(readlink -f /golem-component-service)
```
Dentrax, line 74:
```suggestion
            test "$(readlink /golem-component-service)" = "/usr/bin/golem-component-service"
```

golem.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+      cargo make --profile ci build-release
+
+      mkdir -p ${{targets.contextdir}}/usr/bin/
+      mv ./target/release/golem-component-service ${{targets.contextdir}}/usr/bin/golem-component-service
+
+  - uses: strip
+
+subpackages:
+  - name: ${{package.name}}-component-service
```
Dentrax, line 54: Should we have runtime deps here? https://github.com/golemcloud/golem/blob/0a04f54f954ec859b59b810d9184cadba3fafaec/golem-component-service/docker/Dockerfile#L20C42-L20C79

golem.yaml
```diff
@@ -0,0 +1,87 @@
...
+      # export PATH="$(dirname "$(rustup which cargo)"):$PATH"
+
+      export PATH=$PATH:$HOME/.cargo/bin
+
+      cargo install --force cargo-make
+
+      cargo make --profile ci build-release
+
+      mkdir -p ${{targets.contextdir}}/usr/bin/
+      mv ./target/release/golem-component-service ${{targets.contextdir}}/usr/bin/golem-component-service
```
Dentrax, line 49: Main package name is `golem` but it only provides `golem-component-service`, wondering if this should have a subpkg with a `ranges` pipeline, or we can drop a TODO block if other binaries needed in the future.

golem.yaml
developer-guy: let's remove these comments

golem.yaml
developer-guy: maybe we can do this refactoring later on, atm the only component that we need to build is component-service, that's it.

golem.yaml
developer-guy: this is a fair concern, let me add them, thanks for the catch!

developer-guy: but then I noticed that its using a ubuntu as base image https://github.com/golemcloud/golem/blob/0a04f54f954ec859b59b810d9184cadba3fafaec/golem-component-service/docker/Dockerfile#L1 so there are lots of utilities included into the image so we to have make sure that which are needed for runtime functionality so the only of knowing this is to create an image from it and test it at runtime so I think we might leave runtime dependencies empty at this moment.

Dentrax: yeah actually i have never seen a use-case why we add pkgconf at the runtime deps, and why they add `libssl-dev` to runtime? Looking at the [Cargo](https://github.com/golemcloud/golem/blob/0a04f54f954ec859b59b810d9184cadba3fafaec/golem-component-service/Cargo.toml) file, we may need to have `libssl`, `ca-certificates`? at the runtime anyway since there is gRPC dependency.

Dentrax: `cargo tree -i openssl` it seems some deps has transitive dependency on openssl, and since `reqwest` [doesn't use](https://github.com/golemcloud/golem/blob/0a04f54f954ec859b59b810d9184cadba3fafaec/Cargo.toml#L163) the `rustls` feature, let's have the `pkgconf`, `libssl-dev`, and `ca-certificates` packages in the runtime to match with upstreams.

Dentrax (changes requested):

golem.yaml
```diff
@@ -0,0 +1,93 @@
...
+      expected-commit: 3649cae14476dd534b6d51d12e686b1f020aa27b
+
+  - name: Configure and build
+    runs: |
+      # rustup install stable
+      # rustup target add wasm32-wasip1
+      # rustup target add wasm32-unknown-unknown
+
+      # # Put our rustup-installed toolchains into $PATH
+      # export PATH="$(dirname "$(rustup which cargo)"):$PATH"
```
Dentrax, line 40:
```suggestion
```

Dentrax (approved):

PR #52445, opened 2025-05-05
add cairomm-1.16 package

kranurag7:
adds cairomm-1.16 package; required by graph-tool


Dentrax (approved):
Confirmed that both dev and doc subpkgs does seem well, checked logs and we have needed build-time deps, so LGTM!

PR #52457, opened 2025-05-05
fix(kubeflow-pipelines-metadata-writer): Use Python 3.10

EyeCantCU:
ml-metadata doesn't support anything newer

### DIFF
--- a/kubeflow-pipelines.yaml
+++ b/kubeflow-pipelines.yaml
@@ -1,7 +1,7 @@
 package:
   name: kubeflow-pipelines
   version: "2.5.0"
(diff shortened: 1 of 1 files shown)


developer-guy:
I found a PR in the upstream repository where they applied the fix for the CVE here so I asked them for an update on that PR: https://github.com/kubeflow/pipelines/pull/11741#issuecomment-2854447639

I'm waiting for a response from them before proceeding further and also I'll try to apply the changes in their PR as a patch for sure to see if it works

kranurag7 (approved):

PR #52479, opened 2025-05-05
openjdk8: update to the latest GA(8u452)

ajayk:
https://mail.openjdk.org/pipermail/distro-pkg-dev/2025-May/042657.html


dakaneye (approved):
looks like 8u459 is security/bugfix patches and backporting some fixes

k4leung4 (approved):

PR #52514, opened 2025-05-05
fix(cve-pr-closer): switching to subject pattern for octosts

tdunlap607:
Follow-up to this [comment](https://github.com/wolfi-dev/os/pull/52156#discussion_r2070350671), switching to subject pattern for consistency with octosts pattern.

### DIFF
--- a/.github/chainguard/lifecycle-cve-pr-closer.sts.yaml
+++ b/.github/chainguard/lifecycle-cve-pr-closer.sts.yaml
@@ -1,8 +1,7 @@
 issuer: https://accounts.google.com
 
 # lc-cve-pr-closer-cron@prod-enforce-fabc.iam.gserviceaccount.com
-claim_pattern:
-  email: 'lc-cve-pr-closer-cron@prod-enforce-fabc\.iam\.gserviceaccount\.com'
+subject_pattern: "(101045395775641394837)"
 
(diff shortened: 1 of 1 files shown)


.github/chainguard/lifecycle-cve-pr-closer.sts.yaml
```diff
@@ -1,8 +1,7 @@
 issuer: https://accounts.google.com
 
 # lc-cve-pr-closer-cron@prod-enforce-fabc.iam.gserviceaccount.com
-claim_pattern:
-  email: 'lc-cve-pr-closer-cron@prod-enforce-fabc\.iam\.gserviceaccount\.com'
+subject_pattern: "(101045395775641394837)"
```
imjasonh, line 4:
```suggestion
subject: "101045395775641394837"
```

If we only expect one subject ever, we can dispense with the regex. Either way is fine though.

imjasonh (dismissed):
If

.github/chainguard/lifecycle-cve-pr-closer.sts.yaml
tdunlap607: I think a single subject is expected. Updated!

imjasonh (approved):

PR #52526, opened 2025-05-06
melange: add tmpfs /dev/shm to melange-init for VMs

stevebeattie:
Builds using bazel are failing with the qemu melange runner because it expects the presence of a tmpfs mounted at /dev/shm. Modify the init for melange VMs to create the mountpoint for it and populate it.

### DIFF
--- a/melange.yaml
+++ b/melange.yaml
@@ -1,7 +1,7 @@
 package:
   name: melange
   version: "0.23.15"
-  epoch: 1
+  epoch: 2
   description: build APKs from source code
   copyright:
     - license: Apache-2.0
--- a/melange/init
+++ b/melange/init
@@ -8,6 +8,8 @@ mount -t proc proc -o nodev,nosuid,hidepid=2 /proc
 mount -t devtmpfs -o nosuid,noexec devtmpfs /dev
 mount -t sysfs sys -o nodev,nosuid,noexec /sys
 mount -t tmpfs -o nodev,nosuid tmpfs /tmp
+mkdir -p -m 0755 /dev/shm
+mount -t tmpfs -o nodev,nosuid,mode=1777 tmpfs /dev/shm
 
 # Fix /dev/fd
 ln -s /proc/self/fd /dev/fd


dustinkirkland (commented):
Isn't /dev/shm usually 777?  It is on Ubuntu anyway....

dustinkirkland:
Oh, maybe the the 777 perms come when tmpfs is actually mounted?

stevebeattie:
Updates:
- made the permissions mode explicit in the mount call
- rebased on top of main to resolve the melange version conflict.

With the updated changes applied, this is what the relevant bits look like in a VM:

```
+ ls -dn /dev/shm
drwxrwxrwt    2 0        0               40 May  6 17:29 /dev/shm
+ mount
+ grep shm
tmpfs on /dev/shm type tmpfs (rw,nosuid,nodev,relatime)
```

xnox (dismissed):
For flags that systemd uses for these you can look it up at:

https://github.com/systemd/systemd/blob/dde0ed633510535db565cb0916282ddf937422d7/src/shared/mount-setup.c#L99

```
        { "tmpfs",       "/dev/shm",                  "tmpfs",      "mode=01777",                               MS_NOSUID|MS_NODEV|MS_STRICTATIME,
          MNT_FATAL|MNT_IN_CONTAINER|MNT_USRQUOTA_GRACEFUL },
```

Which is 01777 mode, nosuid, nodev which is correct. And yes it is premounted in containers, meaning it is assumed and expected to always be present in any runtime.

The purpose is shared memory objects, so tools normally use it to communicate with each other there.

xnox:
Side note, why are we attempting to reimplement systemd here? =)

stevebeattie:
rebased to bump the melange to avoid the conflict with https://github.com/wolfi-dev/os/pull/52646

@xnox yeah, I had that thought that we're re-implimenting systemd functionality here.

xnox:
> rebased to bump the melange epoch to avoid the conflict with #52646
> 
> @xnox yeah, I had that thought that we're re-implimenting systemd functionality here.

i think init predates us having a working systemd package.

xnox (approved):

PR #52546, opened 2025-05-06
php: fix installation conflict along with apache2

Dentrax:
Follow-up fix for: https://github.com/wolfi-dev/os/pull/52332

Fixes an issue where we can't have the `apache2` and `php-8.x` at the same time in the runtime.


kranurag7 (approved):
Looks good, don't think httpd.conf is used at runtime so removing it looks good. Just to be safe we can also consider renaming the same.

PR #52618, opened 2025-05-06
ceph: re-enable rgw parquet support

javacruft:
Add compatibility patch for Apache Arrow >= 20 and re-enable
build flags for feature.


ceph/arrow-20-compat.patch
```diff
@@ -0,0 +1,877 @@
+From 8da50fe3138e72831efd935d0d41c329974ce760 Mon Sep 17 00:00:00 2001
+From: James Page <james.page@chainguard.dev>
+Date: Tue, 6 May 2025 14:59:48 +0100
+Subject: [PATCH] Refactor crypto integration with arrow
```
xnox, line 4: has this been submitted upstream? Cause I don't think we would want to maintain this in the long term.

javacruft: It has - I can add a reference

javacruft: @xnox added an Origin header to the patch for future travellers.

xnox (approved):

PR #52642, opened 2025-05-06
add py3-pystache package configuration

utieyin:
py3-pystache

#### For new package PRs only

### DIFF
--- a/py3-pystache.yaml
+++ b/py3-pystache.yaml
@@ -5,12 +5,9 @@ package:
   description: "A mustache template renderer for Python"
   copyright:
     - license: MIT
-  dependencies:
-    provider-priority: 0
 
 vars:
   pypi-package: pystache
-  py-version: "3.11"
 
 data:
   - name: py-versions
@@ -50,30 +47,24 @@ subpackages:
           mv ${{targets.contextdir}}/usr/bin ./.cleanup/${{range.key}}/
       - uses: strip
     test:
+      environment:
+        contents:
+          packages:
+            - python-${{range.key}}
       pipeline:
+        - runs: |
+            pystache -h | grep "Usage"
         - uses: python/import
           with:
             python: python${{range.key}}
             import: ${{vars.pypi-package}}
-
-  - range: py-versions
-    name: py${{range.key}}-${{vars.pypi-package}}-bin
-    description: Executable binaries for ${{vars.pypi-package}} installed for python${{range.key}}
-    dependencies:
-      provider-priority: ${{range.value}}
-      provides:
-        - py3-${{vars.pypi-package}}
-        - py3-${{vars.pypi-package}}-bin
-      runtime:
-        - py${{range.key}}-${{vars.pypi-package}}
-    pipeline:
-      - runs: |
-          mkdir -p ${{targets.contextdir}}/usr/
-          mv ./.cleanup/${{range.key}}/bin ${{targets.contextdir}}/usr/
-    test:
-      pipeline:
         - runs: |
-            pystache -h | grep "Usage"
+            # Test the package
+            cat <<EOF > test.py
+            import pystache
+            print(pystache.render('Hi {{person}}!', {'person': 'World'}))
+            EOF
+            python test.py | grep -q 'Hi World!'
 
   - name: py3-supported-${{vars.pypi-package}}
     description: meta package providing ${{vars.pypi-package}} for supported python versions.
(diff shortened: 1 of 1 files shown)


py3-pystache.yaml
```diff
@@ -0,0 +1,119 @@
...
+      tag: v${{package.version}}
+      expected-commit: 531a0ecea00a696423175ef46c319b29408dc3e2
+
+subpackages:
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}
+    description: python${{range.key}} version of ${{vars.pypi-package}}
+    dependencies:
+      provides:
+        - py3-${{vars.pypi-package}}
```
justinvreeland, line 41: We shouldn't be creating new packages with the unversioned python3 it can cause the version to get mixed up when resolved at runtime.

py3-pystache.yaml
```diff
@@ -0,0 +1,119 @@
...
+  epoch: 0
+  description: "A mustache template renderer for Python"
+  copyright:
+    - license: MIT
+  dependencies:
+    provider-priority: 0
+
+vars:
+  pypi-package: pystache
+  py-version: "3.11"
```
justinvreeland, line 13: This is interesting, I've not seen this before curious hwy the version is here?

justinvreeland (dismissed):

py3-pystache.yaml
```diff
@@ -0,0 +1,119 @@
...
+        - uses: python/import
+          with:
+            python: python${{vars.py-version}}
+            import: ${{vars.pypi-package}}
+
+test:
+  environment:
+    contents:
+      packages:
+        - py${{vars.py-version}}-${{vars.pypi-package}}
```
justinvreeland, line 97: Installing a specific version of the package and then using it in the test kind of defeats the purpose of the test making sure the package is available and works. You can move that test up to the sub package if you want to run it for every version. I really the `py3-pystache` package would never be installed. But if it is the priorities should be installing the newest, 3.13 version not 3.11

py3-pystache.yaml
```diff
@@ -0,0 +1,119 @@
+package:
+  name: py3-pystache
+  version: "0.6.8"
+  epoch: 0
+  description: "A mustache template renderer for Python"
+  copyright:
+    - license: MIT
+  dependencies:
+    provider-priority: 0
```
EyeCantCU, line 9: We shouldn't set this

py3-pystache.yaml
```diff
@@ -0,0 +1,119 @@
...
+        - uses: python/import
+          with:
+            python: python${{range.key}}
+            import: ${{vars.pypi-package}}
+
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}-bin
+    description: Executable binaries for ${{vars.pypi-package}} installed for python${{range.key}}
+    dependencies:
+      provider-priority: ${{range.value}}
```
EyeCantCU, line 63: Exclude this too

py3-pystache.yaml
```diff
@@ -0,0 +1,119 @@
...
+      provides:
+        - py3-${{vars.pypi-package}}
+        - py3-${{vars.pypi-package}}-bin
```
EyeCantCU, line 66: Remove these

EyeCantCU (dismissed):

py3-pystache.yaml
```diff
@@ -0,0 +1,80 @@
...
+            python: python${{range.key}}
+            import: ${{vars.pypi-package}}
+        - runs: |
+            # Test the package
+            cat <<EOF > test.py
+            import pystache
+            print(pystache.render('Hi {{person}}!', {'person': 'World'}))
+            EOF
+            python test.py | grep -q 'Hi World!'
+
```
justinvreeland, line 64:
```suggestion
  - range: py-versions
    name: py${{range.key}}-${{vars.pypi-package}}-bin
    description: Executable binaries for ${{vars.pypi-package}} installed for python${{range.key}}
    dependencies:
      runtime:
        - py${{range.key}}-${{vars.pypi-package}}
    pipeline:
      - runs: |
          mkdir -p ${{targets.contextdir}}/usr
          mv ${{targets.contextdir}}/../py${{range.key}}-${{vars.pypi-package}}/usr/bin ${{targets.contextdir}}/usr
    test:
      pipeline:
        - runs: |

```

You do need a bin package so these packages don't overlap.  Something like this should work where the runs is doing `--help` or `--version` for the commands that are being added.

py3-pystache.yaml
justinvreeland: Looks like the provides and the priority are still her?

justinvreeland (dismissed):

justinvreeland (approved):
👍🏻

PR #52698, opened 2025-05-07
remove mention of nmap and withdraw existing packages

kranurag7:
nmap no longer exists so remove mentions of the same. 

also withdraws the existing package given we're no longer using it anywhere.


sil2100 (approved):

mamccorm (changes requested):
Do not merge

PR #52726, opened 2025-05-07
Add missing clang dependency

mattmoor:
These all failed due to clang missing in QEMU runner

### DIFF
--- a/pgvector-16.yaml
+++ b/pgvector-16.yaml
@@ -3,7 +3,7 @@
 package:
   name: pgvector-16
   version: 0.8.0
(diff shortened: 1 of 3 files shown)


mattmoor:
It took me forever to figure out what was going on here, and the problem is that my previous bug handling absolute paths is still in the melange used in prod, so we're getting `/home/build/tmp` instead of `/tmp`.

Jason is pushing to production ~now, and then I'll touch things here to trigger a rebuild and things should work. 🤞

imjasonh (approved):

PR #52773, opened 2025-05-07
Fix crossplane provider update identifiers to match repo

AmberArcadia:
Follow on from https://github.com/wolfi-dev/os/pull/51718


powersj (approved):
Thank you!

PR #52774, opened 2025-05-07
From wolfi repo from openssl

mattmoor:
`openssl` explicitly lists the Wolfi APK repository, and we're seeing failures due to the signing key being missing for it, so this lists that explicitly.

### DIFF
--- a/openssl.yaml
+++ b/openssl.yaml
@@ -2,7 +2,7 @@
 package:
   name: openssl
   version: "3.5.0"
-  epoch: 0
+  epoch: 1
   description: "the OpenSSL cryptography suite"
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


sil2100 (commented):
Currently this as-is causes `./lint.sh` to complain. I'm also wondering what is the reason for us adding wolfi as an explicit `build_repository`. Maybe I'm missing some context. It was added during 3.4.0 release when we pinned the version of `jitterentropy-library` - however, the build is working fine without the explicit build repositories addition.
If there's no real reason for this, I'd vote for removing build_repositories completely.

xnox:
We should fix lint, as forbidden repository and forbidden keyring are marked up with nolint at the top.

Separately, can we switch to "apk.cgr.dev/chainguard" here? then no keyring will need to be specified and it should then just all work?

mattmoor:
@xnox I assumed avoiding `apk.cgr.dev/chainguard` was intentional for some reason, but if we can drop that, then even better.

mattmoor:
Trying a local build to confirm we can just remove it 🤞

xnox:
Actually yes we can drop all of this as we have dropped doing build cycles detection or prevention.

murraybd (approved):

PR #52788, opened 2025-05-07
sts-policy: add staging scrape comments bot, read only access

rawlingsj:
This allows a staging bot read access to PR comments

### DIFF
--- a/.github/chainguard/scrape-error-comments.sts.yaml
+++ b/.github/chainguard/scrape-error-comments.sts.yaml
@@ -1,7 +1,8 @@
 issuer: https://accounts.google.com
 
-# detect@jrawlings-chainguard.iam.gserviceaccount.com
-subject: "114294174394359664841"
+# 114294174394359664841: detect@jrawlings-chainguard.iam.gserviceaccount.com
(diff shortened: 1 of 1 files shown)


.github/chainguard/scrape-error-comments.sts.yaml
```diff
@@ -1,7 +1,8 @@
 issuer: https://accounts.google.com
 
-# detect@jrawlings-chainguard.iam.gserviceaccount.com
-subject: "114294174394359664841"
+# 114294174394359664841: detect@jrawlings-chainguard.iam.gserviceaccount.com
```
vaikas, line 3: Do we still need this?

vaikas (approved):

.github/chainguard/scrape-error-comments.sts.yaml
rawlingsj: Possibly, if something doesn't work out and we need to rework some logic.  It's hard to mock all the comments we've had so this is the best test bed to pull from.  For early dev being able to use a dev cloud run env has been helpful.

dannf (approved):

PR #52796, opened 2025-05-08
jenkins-docker: move to version stream

kranurag7:
this was not version streamed properly and we look for this name inside the image build. let's version stream this properly so that this gets handled by automation from here on. 

have a provides for backwards compatibility as well so this is not a breaking change.


OddBloke (changes requested):
2.509 isn't an LTS, so shouldn't be version streamed.

kranurag7:
closing given this is no longer required.

PR #52818, opened 2025-05-08
added package: trivy-operator

rahulduvedi:


trivy-operator.yaml
```diff
@@ -0,0 +1,82 @@
...
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - glibc-dev
+
+environment:
+  contents:
+    packages:
+      - glibc
+      - glibc-dev
```
kranurag7, line 16: we can omit this; glibc runtime and build-time deps are not required here.

rahulduvedi: I was facing some version related issues with glibc that why I added these here. 
I am not able to replicate, so removing these now.

kranurag7 (approved):

PR #52882, opened 2025-05-08
Create expected cache directory

mattmoor:
I see this when running tests with the QEMU runner:
```
2025/05/08 14:20:12 INFO > 2025/05/08 21:20:11| ERROR: cannot change current directory to /var/cache/squid: (2) No such file or directory
```


xnox:
due to permissions we cannot ship it like this; we need it done separately. will propose fixups.

PR #52891, opened 2025-05-08
Add `cgroups` to the set of mounts melange init does.

mattmoor:
This is needed for the `cgroups` tests to pass under QEMU.

### DIFF
--- a/melange/init
+++ b/melange/init
@@ -10,6 +10,7 @@ mount -t sysfs sys -o nodev,nosuid,noexec /sys
 mount -t tmpfs -o nodev,nosuid tmpfs /tmp
 mkdir -p -m 0755 /dev/shm
 mount -t tmpfs -o nodev,nosuid,mode=1777 tmpfs /dev/shm
+mount -t cgroup -o all cgroup /sys/fs/cgroup
 
 # Fix /dev/fd
 ln -s /proc/kcore /dev/core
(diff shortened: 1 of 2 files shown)


melange/init
```diff
@@ -10,6 +10,7 @@ mount -t sysfs sys -o nodev,nosuid,noexec /sys
 mount -t tmpfs -o nodev,nosuid tmpfs /tmp
 mkdir -p -m 0755 /dev/shm
 mount -t tmpfs -o nodev,nosuid,mode=1777 tmpfs /dev/shm
+mount -t cgroup -o all cgroup /sys/fs/cgroup
```
xnox, line 13: please use these options and settings:

https://github.com/systemd/systemd/blob/df5dd05941da82b3ed291a85595098b168eabda8/src/shared/mount-setup.c#L109

please do not mount cgroup v1, only cgroup2

mattmoor:
```suggestion
mount -t cgroup2 -o nsdelegate,memory_recursiveprot,nosuid,nodev,noexec cgroup2 /sys/fs/cgroup
```

mattmoor, line 13:
```suggestion
mount -t cgroup2 -o nsdelegate,memory_recursiveprot,nosuid,nodev,noexec cgroup2 /sys/fs/cgroup
```

stevebeattie (approved):
Confirmed that this results in a cgroup2 fs mounted in the same way systemd mounts it.

PR #52912, opened 2025-05-09
grafana-11.6: add -buildvcs=false to upstream build system

dnegreira:
Since go1.24, the go build command now sets the main module’s version in the compiled binary based on the version control system tag and/or commit. A +dirty suffix will be appended if there are uncommitted changes.
Use the -buildvcs=false flag to omit version control information from the binary.

### DIFF
--- a/grafana-11.6.yaml
+++ b/grafana-11.6.yaml
@@ -1,7 +1,7 @@
 package:
   name: grafana-11.6
   version: "11.6.1"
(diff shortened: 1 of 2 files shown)


stevebeattie:
See also https://github.com/wolfi-dev/os/pull/52914, though, which makes the same epoch bump.

kbsteere (approved):

PR #52947, opened 2025-05-09
quality: improve prometheus-pushgateway package

sanjayk0508:
Improved the package by using `go/build` and extending the testing with functional tests


prometheus-pushgateway.yaml
```diff
@@ -15,10 +15,7 @@ environment:
     packages:
       - bash
       - build-base
-      - busybox
```
kranurag7, line 18: we can also remove all the remaining explicit deps here.

prometheus-pushgateway.yaml
```diff
@@ -72,8 +66,62 @@ update:
...
+      uses: test/daemon-check-output
+      with:
+        setup: |
+          mkdir -p /tmp/pushgateway
+        start: prometheus-pushgateway --web.listen-address=:9091 --persistence.file=/tmp/pushgateway/metrics --persistence.interval=5s --web.enable-admin-api
+        timeout: 30
+        expected_output: |
+          msg="starting pushgateway"
+          caller=tls_config.go:347 msg="Listening on"
+        error_strings: |
```
Dentrax, line 89: Any reason why we override the default `error_strings`? It'd be nice to have a comment above for reasoning.

prometheus-pushgateway.yaml
```diff
@@ -72,8 +66,62 @@ update:
...
+        - curl
+        - wait-for-it
   pipeline:
-    # AUTOGENERATED
-    - runs: |
-        pushgateway --version
-        pushgateway --help
+    - name: "Basic binary verification"
+      runs: |
+        prometheus-pushgateway --version
```
Dentrax, line 77: What about:
```suggestion
        prometheus-pushgateway --version | grep -i ${{package.version}}
```

prometheus-pushgateway.yaml
```diff
@@ -72,8 +66,62 @@ update:
...
+        timeout: 30
+        expected_output: |
+          msg="starting pushgateway"
+          caller=tls_config.go:347 msg="Listening on"
+        error_strings: |
+          ERROR
+          FATAL
+          panic
+          failed
+        post: |
```
Dentrax, line 94: Let's have the `promtool` dep and _vet_ the metrics:

```
curl -s http://localhost:9091 | promtool check metrics
```

Dentrax (changes requested):

prometheus-pushgateway.yaml
sanjayk0508: my bad, we can use default `error_strings` here

prometheus-pushgateway.yaml
```diff
@@ -72,8 +65,57 @@ update:
...
+              break
+            fi
+            if [ $i -eq 10 ]; then
+              exit 1
+            fi
+            sleep 1
+          done
+
+          # admin API wipe functionality
+          curl -s -X PUT http://localhost:9091/api/v1/admin/wipe
```
kranurag7, line 120: nit: we can consider using `-fsSL` everywhere when we're using `-s` which can have some false positives sometimes.

kranurag7 (dismissed):

prometheus-pushgateway.yaml
sanjayk0508: thanks for the review, have updated it to `-fsSL`, PTAL

prometheus-pushgateway.yaml
```diff
@@ -33,13 +23,16 @@ pipeline:
...
+  - uses: go/build
+    with:
+      packages: .
+      output: prometheus-pushgateway
+      ldflags: |
+        -X github.com/prometheus/common/version.Version=${{package.version}}
+        -X github.com/prometheus/common/version.Revision=aff0d834b730375fc965135e3e2d17d15f82a76f
+        -X github.com/prometheus/common/version.Branch=HEAD
+        -X github.com/prometheus/common/version.BuildUser=reproducible@reproducible
+        -X github.com/prometheus/common/version.BuildDate=$(date +%Y%m%d-%H:%M:%S)
```
developer-guy, line 35: we mostly set dates like this `$(date -u -d "@$SOURCE_DATE_EPOCH" +%Y-%m-%d-%H:%M)` for reproducibility perspective.

sanjayk0508: got it! thanks for the review, have updated the `ldflags`, PTAL

Dentrax (approved):

prometheus-pushgateway.yaml
```diff
@@ -33,13 +23,16 @@ pipeline:
...
-      make build
-
-  - runs: |
-      install -Dm755 pushgateway "${{targets.destdir}}"/usr/bin/pushgateway
-
-  - uses: strip
+  - uses: go/build
+    with:
+      packages: .
+      output: prometheus-pushgateway
```
Dentrax, line 29: This is going to be a BREAKING CHANGE, we were installing the package to `/usr/bin/pushgateway`. `prometheus-pushgateway` image would be going to break. Should we have a symlink for this below?

Dentrax (changes requested):

prometheus-pushgateway.yaml
sanjayk0508: thank you @Dentrax for heads up! have added a symlink. LMK if i'm overlooking anything

prometheus-pushgateway.yaml
```diff
@@ -33,13 +23,20 @@ pipeline:
...
+        -X github.com/prometheus/common/version.Branch=HEAD
+        -X github.com/prometheus/common/version.BuildUser=reproducible@reproducible
+        -X github.com/prometheus/common/version.BuildDate=$(date -u -d "@$SOURCE_DATE_EPOCH" +%Y-%m-%d-%H:%M)
 
+  # for backward compatibility with older installations expecting pushgateway in /usr/bin
   - runs: |
-      install -Dm755 pushgateway "${{targets.destdir}}"/usr/bin/pushgateway
-
-  - uses: strip
+      ln -sf prometheus-pushgateway ${{targets.destdir}}/usr/bin/pushgateway
```
Dentrax, line 39:
```suggestion
      ln -sf /usr/bin/prometheus-pushgateway ${{targets.destdir}}/usr/bin/pushgateway
```

And also, can we have a `readlink` test below? Like this:

```
test "$(readlink /usr/bin/pushgateway)" = "/usr/bin/prometheus-pushgateway"
```

Then

```
pushgateway --help
```

Dentrax (dismissed):

EyeCantCU (approved):

PR #53094, opened 2025-05-12
feat: add iamguarded compat pipelines

wojciechka:


sil2100 (changes requested):
Looks good in general, but I don't see the `.github-token` bits added to `.gitignore`. I think this was done for the other repositories? Since I'd really prefer if we play it safe.

sil2100 (approved):
LGTM now!

PR #53169, opened 2025-05-12
feat: microvm-init setup additional ports

89luca89:
Harden SSHD config a bit, and add an additional port to listen to that can be used to syphon back the workspace when done. This will allow us to avoid adding additional packages to build environments.

### DIFF
--- a/microvm-init/init
+++ b/microvm-init/init
@@ -82,20 +82,6 @@ if [ -e /dev/vda ]; then
 	mount --bind /mount/home/build /home/build
 fi
 
-# Set default hard limit of open file descriptors to match systemd init
-# behavior. The pam_limit module uses the settings in /proc/1/limit as
-# its default values.
-if [ -f /proc/sys/fs/nr_open ] ; then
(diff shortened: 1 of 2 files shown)


microvm-init/init
```diff
@@ -78,6 +78,8 @@ if [ -e /dev/vda ]; then
 
 	# Make /home/build world writeable under /mount
 	chmod 0777 /mount/home/build
+	# Make it accessible also from initramfs
+	mount --bind /mount/home/build /home/build
```
mattmoor, line 82: I'm worried about the combination of these commands.

When I had `/home/build` more permissively permissioned (in an iteration), sshd was rejecting SSH requests.

89luca89: here it did not give problems to me
I suspect the problem is more on the keys permissions than home itself probably

89luca89: Now I'm avoiding sharing it, and workspacetar will just directly tar from /mount/home/build

mattmoor (approved):

PR #53183, opened 2025-05-13
Don't use `$HOME` in `chromium.yaml`

mattmoor:


kranurag7 (approved):
looks good, nit: we can also consider using `${{package.srcdir}}` but not a blocker.

mattmoor:
I don't see any hits for `package.srcdir` in Wolfi, so going with this for now.

PR #53195, opened 2025-05-13
feat(CODEOWNERS): Add guardrails around changes to workflows and withdrawn-packages.txt which can remove packages from package repository

philroche:
There are workflows which alter the state of the package repository and
changes must be approved by the Foundations or acceleration team before merging.

Also, withdrawn-packages.txt is used in conjunction with the workflows/withdraw-packages.yaml
workflow to completely withdraw packages from the package repository. Any withdrawals must be
approved by the Foundations squad to avoid removals that would have unintended consequences.

Prompted by a discussion during onboarding about reproducible builds we realised that the
package versions in our .melange.yaml could be altered despite having the same version. This could
only be achieved via package withdrawal and subsequent re-upload but the conversation highlighted
the importance on ensuring that any package withdrawal is made with great care and thought.

In this commit I assign the responsibility of ensuring responsible use of withdrawn-packages.txt to
the foundations squad.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -16,6 +16,15 @@ CODEOWNERS @wolfi-dev/foundations-squad
 Makefile   @wolfi-dev/foundations-squad
 pipelines/ @wolfi-dev/foundations-squad
 
+# There are workflows which alter the state of the package repository and
+# changes must be approved by the Foundations or acceleration team before merging.
+workflows/  @wolfi-dev/foundations-squad @wolfi-dev/acceleration
+
+# withdrawn-packages.txt is used in conjunction with the workflows/withdraw-packages.yaml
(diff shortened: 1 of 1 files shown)


dannf:
Thanks @philroche. This feels right to me, but earns a discussion with the rest of the team. I feel like we need to define the criteria for a withdrawal approval.

The issue you highlight about duplicate packages is one that is won't actually fire until a YAML file is restored. If we want to catch that case, then we probably need to require review on new files. Perhaps we could automate a check for pre-existing epochs, though it's possible those will cross repository boundaries and need to be ran externally.

dustinkirkland:
I think I agree with @dannf -- is there a withdrawal process documented somewhere, that we can link in here?

cmwilson21:
Thanks for the contribution!

PR #53225, opened 2025-05-13
Add curl and move entrypoint script in opensearch-2

max-allan-cgr:
Fixes:
Problem with the opensearch operator needing curl and the entrypoint to be in /usr/share/opensearch
Added compatibility tests to the image.


utieyin (approved):
looks good

EyeCantCU (approved):

PR #53230, opened 2025-05-13
screen: apply oss-security patches and security recommendations

OddBloke:
Specifically, the recommendations in [0]:

* apply the series of patches provided for screen 5.0.0
* pass `--with-pty-mode=0620` to `configure`
* remove the setuid bit from the `screen` binary (which we ship as `screen-5.0.0`)

[0] https://www.openwall.com/lists/oss-security/2025/05/12/1

### DIFF
--- a/screen.yaml
+++ b/screen.yaml
@@ -2,7 +2,7 @@
 package:
   name: screen
   version: 5.0.0
-  epoch: 1
+  epoch: 2
   description: Window manager that multiplexes a physical terminal
   copyright:
     - license: GPL-3.0-or-later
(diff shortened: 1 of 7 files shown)


screen.yaml
```diff
@@ -27,15 +27,32 @@ pipeline:
       expected-sha256: f04a39d00a0e5c7c86a55338808903082ad5df4d73df1a2fd3425976aed94971
       uri: https://ftp.gnu.org/gnu/screen/screen-${{package.version}}.tar.gz
 
+  # These patches are taken from the oss-security mailing list:
+  # https://www.openwall.com/lists/oss-security/2025/05/12/1
+  - uses: patch
+    with:
+      patches: |
+        0001-logfile-reintroduce-lf_secreopen-to-fix-CVE-2025-233.patch 0002-default-PTY-mode-apply-safe-default-mode-of-0620-to-.patch 0003-attacher.c-fix-bad-strncpy-which-can-lead-to-a-buffe.patch 0004-attacher.c-prevent-temporary-0666-mode-on-PTYs-to-fi.patch 0005-Avoid-file-existence-test-information-leaks-to-fix-C.patch 0006-socket.c-don-t-send-signals-with-root-privileges-to-.patch
```
imjasonh, line 35:
```suggestion
        0001-logfile-reintroduce-lf_secreopen-to-fix-CVE-2025-233.patch
        0002-default-PTY-mode-apply-safe-default-mode-of-0620-to-.patch
        0003-attacher.c-fix-bad-strncpy-which-can-lead-to-a-buffe.patch
        0004-attacher.c-prevent-temporary-0666-mode-on-PTYs-to-fi.patch
        0005-Avoid-file-existence-test-information-leaks-to-fix-C.patch
        0006-socket.c-don-t-send-signals-with-root-privileges-to-.patch
```

(assuming our patch pipeline is capable of handling that...)

OddBloke: Huh, I thought I'd tried that in the past and it hadn't worked, but... it does. Fixed!

OddBloke:
> @OddBloke Can we use upstream's 5.0.1 instead ? https://git.savannah.gnu.org/cgit/screen.git/commit/?h=screen-v5&id=464c8d8f945f53f8cbb854517279349e09d74756

We currently use the release tarballs for builds, so it's not _trivial_ to do so until the tarball is uploaded; that said, provided there isn't a git->release tarball generation step which we haven't packaged the tooling for, switching shouldn't be too hard.

screen.yaml
```diff
@@ -27,15 +27,37 @@ pipeline:
       expected-sha256: f04a39d00a0e5c7c86a55338808903082ad5df4d73df1a2fd3425976aed94971
       uri: https://ftp.gnu.org/gnu/screen/screen-${{package.version}}.tar.gz
 
+  # These patches are taken from the oss-security mailing list:
```
eslerm, line 30: could we add "and are incorporated in future upstream releases" or similar?

eslerm (approved):
Thanks for your quick response and patience. Appreciate that cveless 0006 is included :)

OddBloke:
I didn't manage to get back to this yesterday, and the 5.0.1 tarball was uploaded this morning, so automation should pick it up, obviating the need for this PR. Thanks for the attention, @eslerm!

PR #53249, opened 2025-05-13
Remove the lean4 package definition

murraybd:
lean4 was failing to build from source with version 4.18 and 4.19 when libuv, which it build depends on, was updated to version 1.51 from version 1.50. While I fixed it in https://github.com/wolfi-dev/os/pull/53144 the patch is hackish because it modifies the stage0 files something which upstream says not to do. Additionally, upstream is building the package with clang, but we are not and likely should be. Given that the package is not used by any images[1] or other packages[2] I think we should remove it as it will cause more toil going forward.

[1] https://tf.chaindag.dev/14997225385/mega-module-*.tfplan.json/packages search for lean4
[2]
```
brian-murray:~/src$ for dir in wolfi-os enterprise-packages extra-packages; do grep lean4 $dir/*.yaml; done
wolfi-os/lean4.yaml:  name: lean4
wolfi-os/lean4.yaml:      repository: https://github.com/leanprover/lean4
wolfi-os/lean4.yaml:  - name: "lean4-static"
wolfi-os/lean4.yaml:    description: "lean4 static libraries"
wolfi-os/lean4.yaml:  - name: "lean4-dev"
wolfi-os/lean4.yaml:    description: "lean4 development headers"
wolfi-os/lean4.yaml:    identifier: leanprover/lean4
```


dannf (approved):
thanks for cleaning things up!

PR #53290, opened 2025-05-14
quality: extend test thanos package

yashpimple:
extend test coverage for thanos package by adding functional version checks and runtime validation using `daemon-check-output`


Dentrax (approved):
Nice work!

PR #53291, opened 2025-05-14
Rabbitmq default user credential bitnami compat

rahulduvedi:
added new bitami compatable subpackage to rabbitmq-default-user-credential-updater


Dentrax (changes requested):

mamccorm (changes requested):
Needs some more discussion around tests. Will loop back offline

PR #53302, opened 2025-05-14
quality: extend test thanos-operator package

yashpimple:
add tests for `thanos-operator` package by switching to `go/build` and adding functional runtime tests with CRD installation and controller startup validation.


thanos-operator.yaml
```diff
@@ -14,6 +14,9 @@ environment:
       - go
```
rakshitgondwal, line 14: we should not be requiring these packages if we use go/build right?

yashpimple: thanks for catching that, I had overlooked it. Updated the PR

thanos-operator.yaml
```diff
@@ -1,19 +1,15 @@
...
-  contents:
-    packages:
-      - busybox
-      - ca-certificates-bundle
-      - go
-
 pipeline:
+  # Address  CVE-2022-21698  GHSA-69ch-w2m2-3vjp  GHSA-69cg-p879-7622
+  # Mitigate CVE-2022-28948
+  # Mitigate CVE-2023-39325 and CVE-2023-3978
```
Dentrax, line 12: Not needed or wrongly placed, lets omit.
```suggestion
```

thanos-operator.yaml
```diff
@@ -62,7 +53,33 @@ update:
...
+      packages:
+        - kubectl
+        - git
   pipeline:
     # AUTOGENERATED
     - runs: |
         manager --help
+    - uses: test/kwok/cluster
+    - runs: |
+        git clone --depth=1 https://github.com/banzaicloud/thanos-operator.git
```
Dentrax, line 67:
```suggestion
        git clone --depth=1 --branch=${{package.version}} https://github.com/banzaicloud/thanos-operator.git
```

thanos-operator.yaml
```diff
@@ -62,7 +53,33 @@ update:
...
+          # Create required certificates directory
+          mkdir -p /tmp/certs
+          kubectl config view --minify --raw > /tmp/kwok-kubeconfig.yaml
+        start: |
+          manager -metrics-addr=:8080 -enable-leader-election=false -kubeconfig=/tmp/kwok-kubeconfig.yaml -verbosity=2
+        timeout: 30
+        expected_output: |
+          starting manager
+          starting metrics server
+          Starting Controller
```
Dentrax, line 85: Can we have a `post:` step here, and check for port with `wait-for-it` and simple `curl` request, if possible?

Dentrax (changes requested):

thanos-operator.yaml
yashpimple: @Dentrax updated package PR to include a `post` step that checks the metrics port 👍

Dentrax (approved):
Great work!

PR #53311, opened 2025-05-14
fix: format ext4 with news profile

89luca89:
This is useful to allow a greater number of inodes, which is ideal for source files and compilation


maxgio92 (approved):
LGTM

PR #53331, opened 2025-05-14
qemu - enable vnc support.

smoser:
Without this change, using 'qemu-system-x86 -vnc' pretends to not know what vnc is.


zmarano (approved):

89luca89:
Leaving this here maybe it can be done in this pr: 
https://github.com/chainguard-dev/wolfi-vm/pull/337#issuecomment-2880520013

PR #53339, opened 2025-05-14
add istioctl-zsh-completion package

sschubertchainguard:
- add istioctl-zsh-completion package
- add bash-completion dependency for istioctl-bash-completion


kranurag7 (commented):
this requires an epoch bump

sschubertchainguard:
Thank you! Bumped.

kranurag7 (approved):

PR #53344, opened 2025-05-14
chore(kubevela): add functional tests w/ KOWK + remove reduntant statement in definition

tulilirockz:
Improves on the test by actually testing if KubeVela works, and also removes a few options we don't need anymore on the manifest

### DIFF
--- a/kubevela.yaml
+++ b/kubevela.yaml
@@ -35,6 +35,8 @@ pipeline:
       output: vela
       packages: ./references/cmd/cli/main.go
 
+  - uses: strip
+
 update:
   enabled: true
   github:
--- a/kubevela.yaml
+++ b/kubevela.yaml
@@ -1,7 +1,7 @@
 package:
   name: kubevela
   version: "1.10.3"
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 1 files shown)


kubevela.yaml
```diff
@@ -16,6 +16,8 @@ environment:
       - busybox
       - ca-certificates-bundle
       - go
+  environment:
```
kranurag7, line 19: we always want to link dynamically so we can omit CGO_ENABLED and while we're here we can also remove all the explicit build-time dependencies because go/build pipeline already pulls in all of those. 

ref: https://github.com/chainguard-dev/melange/blob/83d4d3ec5dc419946ab0510833d6e006b6b29797/pkg/build/pipelines/go/build.yaml#L3-L7

kubevela.yaml
```diff
@@ -47,19 +35,23 @@ pipeline:
       output: vela
       packages: ./references/cmd/cli/main.go
 
-  - uses: strip
```
jamonation, line 50: Should be ok to leave this step in - it is fairly conservative and will preserve symbol names in go binaries

jamonation (approved):

kranurag7 (approved):

PR #53351, opened 2025-05-14
glibc: Enable openssf-compiler-options

sergiodj:
Restore openssf-compiler-options hardening on glibc.

This build is now passing on both amd64 and arm64.

Relates: https://github.com/chainguard-dev/internal-dev/issues/7756
Relates: https://github.com/wolfi-dev/os/pull/39303
Relates: https://github.com/chainguard-dev/internal-dev/issues/7940

### DIFF
--- a/glibc.yaml
+++ b/glibc.yaml
@@ -2,7 +2,7 @@ package:
   name: glibc
   version: "2.41"
   # Every glibc update causes build disruption; always announce on #eng-psa
-  epoch: 50
+  epoch: 51
   description: "the GNU C library"
   copyright:
     - license: LGPL-2.1-or-later
@@ -44,8 +44,6 @@ environment:
   # glibc manages FORTIFY_SOURCE on per-file basis
   environment:
     CPPFLAGS: ""
-    # https://github.com/chainguard-dev/internal-dev/issues/7756
-    GCC_SPEC_FILE: /dev/null
 
 pipeline:
   - uses: git-checkout
@@ -111,6 +109,7 @@ pipeline:
         release/2.41/master/624285af3bfc0d2a513cc921d21142a46e3cebb0: elf: Test case for bug 32976 (CVE-2025-4802)
         release/2.41/master/899dd3ab2fc2bff6b8d37345ee34b7b1ef23fa0c: x86_64: Fix typo in ifunc-impl-list.c.
         release/2.41/master/515d4166f4dbcf43b1568e3f63a19d9a92b2d50e: elf: Fix subprocess status handling for tst-dlopen-sgid (bug 32987)
+        master/81467d4b6168c7ce40d951d6b32e387109c0e5ae: elf: Add optimization barrier for __ehdr_start and _end
 
   - uses: patch
     with:


dannf:
We merged this before, but it resulted in an incident where a number of binaries were segfaulting.

dannf (dismissed):
Adding a blocking comment just to prevent an overly friendly foundation'er from approving until we've root caused the issue

xnox:
> We merged this before, but it resulted in an incident where a number of binaries were segfaulting.

Some of the hardening enablement requires rebuilding / rebootstrapping in order. We did it a few times before in debian and ubuntu. If we look through mailing list archives there should be messages from @doko42 about it. Possibly @stevebeattie remembers it too?

sergiodj:
@eslerm ^

sergiodj:
I'm working on this PR again and I have to say that I'm cautiously optimistic about the results.

With a hardened glibc built using the new GCC 15, I cannot trigger the segfaults that were happening when using a hardened glibc.  I haven't done extensive testing so far, thus the caution in saying that the problem has been addressed.

The tests I did were:

* Install `py3-matplotlib` and run `python3.13 -c 'import matplotlib'`.
* Run `make test/tensorflow-cpu-jupyter`.

These were two scenarios where we'd see segfaults happening.

I believe a better test would be to:

* Run `make test` against all packages in the archive, and take note of those that fail only with the hardened glibc.
* Rebuild our images with the hardened glibc and run their tests.

I'll report back when I have more concrete results.

sergiodj:
I spent some time yesterday doing a really intense debugging session (with a friend who was a glibc contributor) to examine the crash with a hardened glibc compiled with GCC 14, and although I wasn't able to exactly say *why* the crash occurs, I believe I can provide a much better explanation of what's going on behind the scenes.  I won't be very verbose here because writing everything up would take a lot of time, and I intend to publish a blog post about the investigation anyway.

The first thing I did was to come up with a simple reproducer, because that would make my life easier.  This is the backtrace I got when `python3.13 -c 'import matplotlib'` crashed (truncated for brevity):

```
#0  0x00007c43afe9972c in __pthread_kill_implementation () from /lib/libc.so.6
#1  0x00007c43afe3d8be in raise () from /lib/libc.so.6
#2  0x00007c43afe2531f in abort () from /lib/libc.so.6
#3  0x00007c43af84f79d in uw_init_context_1[cold] () from /usr/lib/libgcc_s.so.1
#4  0x00007c43af86d4d8 in _Unwind_RaiseException () from /usr/lib/libgcc_s.so.1
#5  0x00007c43acac9014 in __cxxabiv1::__cxa_throw (obj=0x5b7d7f52fab0, warning: could not convert 'std::type_info' from the host encoding (ISO-8859-1) to UTF-32.
This normally should not happen, please file a bug report.
tinfo=0x7c429b6fd218 <typeinfo for pybind11::attribute_error>, dest=0x7c429b5f7f70 <pybind11::reference_cast_error::~reference_cast_error() [clone .lto_priv.0]>)
    at ../../../../libstdc++-v3/libsupc++/eh_throw.cc:93
#6  0x00007c429b5ec3a7 in ft2font__getattr__(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) [clone .lto_priv.0] [clone .cold] () from /usr/lib/python3.13/site-packages/matplotlib/ft2font.cpython-313-x86_64-linux-gnu.so
#7  0x00007c429b62f086 in pybind11::cpp_function::initialize<pybind11::object (*&)(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >), pybind11::object, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, pybind11::name, pybind11::scope, pybind11::sibling>(pybind11::object (*&)(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >), pybind11::object (*)(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >), pybind11::name const&, pybind11::scope const&, pybind11::sibling const&)::{lambda(pybind11::detail::function_call&)#1}::_FUN(pybind11::detail::function_call&) [clone .lto_priv.0] ()
   from /usr/lib/python3.13/site-packages/matplotlib/ft2font.cpython-313-x86_64-linux-gnu.so
#8  0x00007c429b603886 in pybind11::cpp_function::dispatcher(_object*, _object*, _object*) () from /usr/lib/python3.13/site-packages/matplotlib/ft2font.cpython-313-x86_64-linux-gnu.so
...
```

I also found out that Emacs crashed sometimes.  This is the backtrace I got from it:

```
#0  0x00007eede329972c in __pthread_kill_implementation () from /lib/libc.so.6
#1  0x00007eede323d8be in raise () from /lib/libc.so.6
#2  0x00007eede322531f in abort () from /lib/libc.so.6
#3  0x00007eede262879d in uw_init_context_1[cold] () from /usr/lib/libgcc_s.so.1
#4  0x00007eede2646e7c in _Unwind_Backtrace () from /usr/lib/libgcc_s.so.1
#5  0x00007eede3327b11 in backtrace () from /lib/libc.so.6
#6  0x000059535963a8a1 in emacs_backtrace ()
#7  0x000059535956499a in main ()
```

Note how both crashes happen while libgcc is trying to unwind the stack (for different reasons, but still).  Unwinding a stack is a complex operation and I'm not surprised that we're seeing problems here, although I am surprised that we're still focusing on GCC while the problem seems to manifest because of glibc.  But let's keep going.

We can readily see that from `uw_init_context_1[cold]` upwards everything seems the same.  The `[cold]` part of the function name means that GCC considered a particular branch as not likely to be reached.  I examined the disassembly of `uw_init_context_1`, and it looked like a pattern that's seen in `assert` calls:

```asm
...
   0x00007eede2645e82 <+146>:   call   0x7eede2644a90 <uw_frame_state_for>
   0x00007eede2645e87 <+151>:   test   %eax,%eax
   0x00007eede2645e89 <+153>:   jne    0x7eede2628798 <uw_init_context_1.cold>
...
```

Upon inspecting libgcc's source code, we see the following excerpt:

https://github.com/gcc-mirror/gcc/blob/c0be0298a9553c13af642f45628a15d833473657/libgcc/unwind-dw2.c#L1340-L1341

I also checked the contents of `%eax`, and it indeed matches the value of `_URC_END_OF_STACK`.  So now we know that this is the cause of the abort.

I decided to proceed further and compile a new GCC with debuginfo so that I could analyse what `uw_frame_state_for` is doing.  This took me down a rabbit hole, but eventually I found that the problem is that `fde` is `NULL` here:

https://github.com/gcc-mirror/gcc/blob/c0be0298a9553c13af642f45628a15d833473657/libgcc/unwind-dw2.c#L1007-L1019

Because amd64 defines `MD_FALLBACK_FRAME_STATE_FOR`, we end up calling it inside the `if` statement, but that's not really important.  What's important is understanding that `_Unwind_Find_FDE` *never* returns `NULL` when glibc isn't hardened, but does return `NULL` *sometimes* when glibc is hardened.

Now, `_Unwind_Find_FDE` is complex and deals with CFI, `.eh_frame` and DWARF (*not* the DWARF that goes inside the `.debug*` sections) in order to find the frame description information and properly unwind the stack.  It eventually calls `find_fde_tail`, which does the hard job.

Here we get to the end of the line, at least for my debugging session.  I was able to track down the exact spot where `find_fde_tail` is returning `NULL`:

https://github.com/gcc-mirror/gcc/blob/c0be0298a9553c13af642f45628a15d833473657/libgcc/unwind-dw2-fde-dip.c#L454-L455

When I looked at the addresses for `pc` and `table[0].initial_loc + data_base`, I found that the latter is an address that falls within `/lib/ld-linux-x86-64.so.2`'s text section, while the former fell within libgcc's text.  Since the loader is mapped close to the end of the process virtual memory, the comparison is true and the function ends up returning `NULL`.  On a non-hardened glibc, I noticed that both addresses would fall within libgcc's sections, which wouldn't trigger the return.

As I said in the beginning, this is an explanation of what is happening behind the scenes up until libgcc decides to unwind the stack.  It doesn't really explain why we're seeing the discrepancy in the addresses handled by `find_fde_tail`.  Nonetheless, I believe this gives us a much clearer picture of the problem and, more importantly, allowed me to come up with a minimal reproducer for the problem.  The reproducer isn't as deterministic as running `python3.13 -c 'import matplotlib'`, but during my tests I've determined that it consistently crashes if I run it a dozen times or more.

The reproducer is simple.  We can trigger a stack unwinding by using `backtrace(3)`:

```c
#include <execinfo.h>

int
main(int argc, char *argv[])
{
  void *a[4096];
  backtrace (a, 100);
  return 0;
}
```

With that, it is already possible to confidently write a testcase for the glibc which exercises the problem and makes sure that the crash doesn't happen.

FWIW, I did several tests using a hardened glibc compiled with GCC 15, and was able to confirm that the problem does not manifest anymore.

dannf:
Thanks for the detailed investigation, and I love the simplicity of the reproducer you obtained, even if it isn't 100%. I suppose we could run it in a loop 32 times or something in our package test.

I feel like you're narrowing in on the problem - but that we're not at a point where we can say "we know what caused it / we know it is fixed", which leaves me a bit paranoid. What if it's just 95% better, but there's still some other binaries out there that can trigger the crash.

I'm curious, have you narrowed down which `openssf-compiler-flags` flag(s) is causal? Maybe that information, combined with what you've identified here, would help us pin point the problem. Or, as you said before, we could bisect GCC to see if we can determine what the relevant change in GCC 15 is that fixed it.

(btw, the morning this originally landed in wolfi, I noticed `emacs` crashing as well!)

xnox:
cet, ibt and shadowstack are likely the culprits no?

xnox:
and the reason why things work now is that gcc15's libgcc is now compiled with gcc14 which had hardening on and is now shadowstack aware? there was weird ordering in the builds one has to do when trying to get glibc doing shadowstack stuff.

sergiodj:
> Thanks for the detailed investigation, and I love the simplicity of the reproducer you obtained, even if it isn't 100%. I suppose we could run it in a loop 32 times or something in our package test.

I've added a new test to #54778 which iterates over the test 1000 times.  It's still pretty fast and should be able to catch any unwinding problems.

> I feel like you're narrowing in on the problem - but that we're not at a point where we can say "we know what caused it / we know it is fixed", which leaves me a bit paranoid. What if it's just 95% better, but there's still some other binaries out there that can trigger the crash.

I understand the paranoia.  OTOH, I feel like I have to point out that we may never really fully understand (in a timely manner) what change(s) on GCC impacted this issue and how exactly they fixed it.

> I'm curious, have you narrowed down which `openssf-compiler-flags` flag(s) is causal? Maybe that information, combined with what you've identified here, would help us pin point the problem. Or, as you said before, we could bisect GCC to see if we can determine what the relevant change in GCC 15 is that fixed it.

Let me come up with a bisecting strategy and leave it running.  Maybe we'll hit something obvious.

> (btw, the morning this originally landed in wolfi, I noticed `emacs` crashing as well!)

Ah, I actually learned about that because of your message!

sergiodj:
> and the reason why things work now is that gcc15's libgcc is now compiled with gcc14 which had hardening on and is now shadowstack aware? there was weird ordering in the builds one has to do when trying to get glibc doing shadowstack stuff.

It may very well be CET/etc, but the problem seems to be related only to glibc; even when I downgrade libgcc to 14.2.0-r12, it still works with a glibc that's been compiled (and hardened) against GCC 15.

xnox:
@sergiodj can we test things still work on Apple Rosetta 2 Docker Desktop VM => we have had interesting cases where their Jit expects assembly instructions to be in certain order and certain side-effects. Mostly out of paranoia.

cc @justinvreeland

sergiodj:
After a lengthy `git bisect` run, I found that the GCC commit that fixes the problem is:

```
commit 99b1daae18c095d6c94d32efb77442838e11cbfb
Author: Richard Biener <rguenther@suse.de>
Date:   Fri May 3 14:04:41 2024 +0200

    tree-optimization/114589 - remove profile based sink heuristics
    
    The following removes the profile based heuristic limiting sinking
    and instead uses post-dominators to avoid sinking to places that
    are executed under the same conditions as the earlier location which
    the profile based heuristic should have guaranteed as well.
    
    To avoid regressing this moves the empty-latch check to cover all
    sink cases.
    
    It also stream-lines the resulting select_best_block a bit but avoids
    adjusting heuristics more with this change.  gfortran.dg/streamio_9.f90
    starts execute failing with this on x86_64 with -m32 because the
    (float)i * 9.9999...e-7 compute is sunk across a STOP causing it
    to be no longer spilled and thus the compare failing due to excess
    precision.  The patch adds -ffloat-store to avoid this, following
    other similar testcases.
    
    This change fixes the testcase in the PR only when using -fno-ivopts
    as otherwise VRP is confused.
    
            PR tree-optimization/114589
            * tree-ssa-sink.cc (select_best_block): Remove profile-based
            heuristics.  Instead reject sink locations that sink
            to post-dominators.  Move empty latch check here from
            statement_sink_location.  Also consider early_bb for the
            loop depth check.
            (statement_sink_location): Remove superfluous check.  Remove
            empty latch check.
            (pass_sink_code::execute): Compute/release post-dominators.
    
            * gfortran.dg/streamio_9.f90: Use -ffloat-store to avoid
            excess precision when not spilling.
            * g++.dg/tree-ssa/pr114589.C: New testcase.
```

Unfortunately, my GCC-fu isn't enough to understand *why* this commit makes things work.  I will open an upstream bug against GCC 14 and report my findings there; maybe someone can help me make some sense out of this.

Meanwhile, I talked to @dannf and he suggested that I find the exact hardening flag that is making glibc crash.  As it turns out, it's `-z now`.  This actually corroborates Dann's own findings back when he first investigated this issue.

With all this new information, I believe the best way to move forward is to remove `-z now` from the list of flags being used during linking.  Even though the problem doesn't manifest with a hardened glibc compiled with GCC 15, I'd still prefer to err on the side of caution here, at least until we have more concrete information from upstream GCC.

Last but not least, my testcase is [now part of our glibc package](https://github.com/wolfi-dev/os/commit/c453aa6000890777c965ae2b5a7a4c36d54af363) which should give us more confidence moving forward too.

justinvreeland:
Built & tested this with docker rosetta on my mac and everything seems fine. I am on 15.5 and haven't been able to reproduce rosetta issues on that version.

sergiodj:
FTR, I opened https://gcc.gnu.org/bugzilla/show_bug.cgi?id=120653 which hopefully brings upstream attention to this issue and provides us with a better understanding of the fix.

glibc.yaml
```diff
@@ -48,8 +48,8 @@ environment:
   # glibc manages FORTIFY_SOURCE on per-file basis
   environment:
     CPPFLAGS: ""
-    # https://github.com/chainguard-dev/internal-dev/issues/7756
-    GCC_SPEC_FILE: /dev/null
+    # See https://github.com/wolfi-dev/os/pull/53351#issuecomment-2963753802
+    GCC_SPEC_FILE: /home/build/openssf.spec
```
xnox, line 47: if we are confident that gcc-15 fixed it, should we drop this hunk altogether and use the default build?

glibc.yaml
```diff
@@ -120,6 +120,10 @@ pipeline:
     with:
       patches: Disable-AVX512VL.patch
 
+  - runs: |
+      gccdir=$(GCC_SPEC_FILE=/dev/null gcc --print-search-dirs | grep ^install: | cut -d' ' -f2)
+      sed -r 's/-z now//' < "$gccdir/openssf.spec" > /home/build/openssf.spec
+
```
xnox: if we are confident that gcc-15 fixed it, should we drop this hunk altogether and use the default build?

glibc.yaml
```diff
@@ -181,6 +185,12 @@ pipeline:
       mkdir -p ${{targets.destdir}}/etc/apk/commit_hooks.d
       cp ldconfig-commit.sh -p ${{targets.destdir}}/etc/apk/commit_hooks.d/
 
+  - name: "Clean up /etc/ld.so.cache"
+    runs: |
+      # Having a package own a file that will be different in most
+      # every image poisons our layering mechanism
+      rm ${{targets.destdir}}/etc/ld.so.cache
+
```
xnox: I thought this got done in a different pr already.

xnox (dismissed):

glibc.yaml
sergiodj: I believe it's prudent to do this in steps.  We know for sure that the stack unwinding problem is caused by `-z now`, and I am personally convinced that GCC 15 fixes the problem, but I'd like to wait for upstream input on *why* the commit I found actually fixes the issue.  Once we know that, I think we'll be more confident to just always enable `-z now` and be sure that the problem won't creep back in.

glibc.yaml
sergiodj: There was a draft PR that @dannf was working on, but we decided to join efforts and make a single PR for everything.

sergiodj:
Converted to draft so that I can do a world rebuild (of packages and images) to make sure this doesn't break anything unexpectedly.

dannf (approved):

PR #53358, opened 2025-05-14
fix(kubevela): split vela-cli into its own subpackage

tulilirockz:
That way we dont need to have the entire vela-cli when installing the kubevela operator (vela-core).
Saves about 40MB when you just want the operator

image

I'm not how to approach this tho, should it be an entirely separate package or just a subpackage here?

### DIFF
--- a/kubevela.yaml
+++ b/kubevela.yaml
@@ -9,7 +9,6 @@ package:
     runtime:
       - ca-certificates-bundle
       - vela-cli
-      - vela-core
 
 pipeline:
   - uses: git-checkout
--- a/kubevela.yaml
+++ b/kubevela.yaml
@@ -29,13 +29,33 @@ pipeline:
       output: manager
       packages: ./cmd/core/main.go
 
-  - uses: go/build
-    with:
-      ldflags: -X github.com/oam-dev/kubevela/version.VelaVersion=${{package.version}}
-      output: vela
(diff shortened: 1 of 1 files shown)


jamonation (changes requested):
Let's figure out how to refactor the subpackage/package dependencies to ensure `apk add kubevela` gets the CLI version

tulilirockz:
Split the package into `vela-core` + `vela-cli` now and a depends for the kubevela, that way we have the same functionality for `kubevela` and we can install them separately if we want to. (As @jamonation suggested)

kubevela.yaml
```diff
@@ -1,13 +1,15 @@
...
-  epoch: 1
+  epoch: 2
   description: KubeVela is a modern application delivery platform that makes deploying and operating applications across today's hybrid, multi-cloud environments easier, faster and more reliable
   copyright:
     - license: Apache-2.0
   dependencies:
     runtime:
       - ca-certificates-bundle
+      - vela-cli
+      - vela-core
```
jamonation, line 12:
```suggestion
```

Keeping just `vela-cli` for now will mean anything using this main `kubevela` will be an effective no-op. Once things are updated to use `vela-cli`, then we can remove this entirely

jamonation (approved):

PR #53443, opened 2025-05-15
opensearch-3: Move docker entrypoint

IdlePhysicist:
This commit lifts part of Max's change from #53225 that wasn't applied
to this version stream (likely because the PR that landed this was
already in flight).  The opensearch-operator expects the entrypoint to
be in a specific place, this commit moves the script and symlinks it
back to the old location.


kwmonroe (approved):
:+1:

PR #53471, opened 2025-05-15
Add ldd-check tests to several packages

murraybd:
Using the power of melange and a local archive I found several more packages which would benefit from ldd-check testing.


renovate.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: renovate
   version: "40.14.4"
-  epoch: 0
+  epoch: 1
```
javacruft, line 4: epoch bump only?

javacruft (approved):

PR #53496, opened 2025-05-16
chore(kubevela): add `kubevela-` prefix for vela-{core,cli} subpackages

tulilirockz:
Kubevela has a [`vela-cli`](https://github.com/kubevela/kubevela/tree/master/references) reference implementation that differs from the [`vela-cli`](https://github.com/go-vela/cli) project. Adding the prefix would be nice in order to clear up any confusion around these or anything. Talked about this with @Dentrax and he seemed to agree.

### DIFF
--- a/kubevela.yaml
+++ b/kubevela.yaml
@@ -1,14 +1,14 @@
 package:
   name: kubevela
   version: "1.10.3"
-  epoch: 2
+  epoch: 3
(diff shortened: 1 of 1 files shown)


Dentrax (approved):

developer-guy (approved):
we haven't shared these packages or used in any of images so its safe to change them so I'm approving it.

PR #53546, opened 2025-05-16
Refreshed pre-commit config and brought the repo into compliance

AmberArcadia:
Applied the latest config from https://github.com/chainguard-dev/pre-commit-hooks/blob/main/example.pre-commit-config.yaml, and ensured all files are in compliance. This should be a technical non-op in that no packages / configs are functionally modified, just formatted and cleaned up to fit the rules. Some files were excluded from some checks due to significant hurdles, such as misspell on certain patch files.


gnupg/0320-gpg-default-to-aes256.patch
```diff
@@ -11,7 +11,7 @@ It's 2017, and pretty much everyone has AES-256 available.  Symmetric
 crypto is also rarely the bottleneck (asymmetric crypto is much more
 expensive).  AES-256 provides some level of protection against
 large-scale decryption efforts, and longer key lengths provide a hedge
-against unforseen cryptanalysis.
```
justinvreeland, line 14: It's a minor change but we should exclude patches from our pre commit formatting hooks if they're pulled from upstream we shouldn't be changing it.

.github/chainguard/lofo.sts.yaml
```diff
@@ -1,8 +1,6 @@
-# LOg FOrwarder GH Bot
-
 issuer: https://accounts.google.com
 
-
+# LOg FOrwarder GH Bot
```
justinvreeland, line 3: The yam formatter has a habit of moving comments around incorrectly. We should probably only apply it to things our tool consumes or split it out to double check the comments.

.github/dependabot.yml
```diff
@@ -1,4 +1,5 @@
 version: 2
+
```
justinvreeland, line 2: I think this is caused by the yam '.' directive. I'm not sure the fields yam uses for melange files will apply correctly to how we want to format github yaml files.

xnox (approved):

xnox:
Reviewed that all changes are inconsequential.

Now that this is merged, we will have to rebase a few lib-merge staged PRs which would then ship updates.

For some of the files, they were changed without epoch bumps. To maintain consistency will do a sweep, after lib-merge lands, to bump epochs on anything that will remain unrebuild still past then.

kubeflow-pipelines/metadata-envoy-config.yaml
```diff
@@ -1,13 +1,17 @@
 admin:
   access_log_path: /tmp/admin_access.log
   address:
-    socket_address: { address: 0.0.0.0, port_value: 9901 }
```
justinvreeland, line 4: Ther'es no reason for us to reformat files from what upstream has pretty sure this came from here: https://github.com/kubeflow/pipelines/blob/master/third_party/metadata_envoy/envoy.yaml

justinvreeland:
> Reviewed that all changes are inconsequential.

Oh it was merged while I was reviewing. I don't agree with this. I don't think we should be formatting with yam yaml files not meant for our tooling and I don't think we should be formatting patches or upstream yaml files at all. Especially since we know yam has problems with comments.

xnox:
> > Reviewed that all changes are inconsequential.
> 
> Oh it was merged while I was reviewing. I don't agree with this. I don't think we should be formatting with yam yaml files not meant for our tooling and I don't think we should be formatting patches or upstream yaml files at all. Especially since we know yam has problems with comments.

ACK! will revert things you point out.

kubeflow-pipelines/metadata-envoy-config.yaml
xnox: we likely want to make our yam config more tight.

AmberArcadia:
Thank you both for your thoughts! I was mulling over the idea of excluding patches and .github yaml and such, but it seemed like they could be made to conform without causing issues. I'm happy to exclude those going forward though! This is absolutely all subject to change. I still have the other two repos to do, but I'll wait until after the lib-merge process is done for simplicity.

gnupg/0320-gpg-default-to-aes256.patch
AmberArcadia: Gotcha, we can definitely exclude all patches going forward.

.github/chainguard/lofo.sts.yaml
AmberArcadia: Agreed, had to get very familiar with the quirks to make sure it didn't outright delete comments. That one in particular was getting deleted until I moved it iirc. I'm happy to exclude .github though if you'd prefer.

.github/dependabot.yml
AmberArcadia: Interesting. I did notice it alternated between wanting line breaks and not wanting them depending on the file, perhaps that's why.

.github/chainguard/lofo.sts.yaml
AmberArcadia: Happy to exclude this file instead.

kubeflow-pipelines/metadata-envoy-config.yaml
AmberArcadia: Happy to exclude this file instead.

.github/chainguard/lofo.sts.yaml
justinvreeland: Ahh yeah itt can be hard to get it to keep the top level comments for some reason.

.github/dependabot.yml
justinvreeland: It is a bit strange though tbh i'm not sure why it's happening that's just my guess.

justinvreeland:
Thanks for doing all the work here! I'm all in favor of formatting things. The only part I feel strongly about is the patches and excluding *.patch should fix that.   I think we probably don't want to change files we get from upstream in the `<package-name>/` directories but it's hard to distinguish upstream yaml files and file we've created to bump deps and stuff : (

The .github files we can do w/e we want with worth pointing out that our yam config file might end up with overlapping entries for those yaml files and our melange yaml files and we might want to tune it a bit to make sure we get the output we want. i'm 100% in favor of the spelling fixes other than patches as well. Surprised the whitespace and newline fixers weren't included though. Too much noise?

.github/dependabot.yml
AmberArcadia: I looked into this a bit further and it seems to stem from our `.yam.yaml` configuration. Perhaps we add some tweaks to it in the future.

PR #53630, opened 2025-05-18
rootlesskit switch to go/build

xnox:
Switch rootlesskit to go/build such that binary is built optimized,
and is easier to fipsify.


kranurag7 (approved):
thank you; looks good to land.

PR #53677, opened 2025-05-19
chore(kubevela): vela-core-entrypoint compatibility package for upstream image compat

tulilirockz:
Needs to be here so that we get the same entrypoint as the [upstream image](https://github.com/kubevela/kubevela/blob/ef9b6f3cc10a4b6871b5698ca41fea3f6b3bcaec/Dockerfile#L46).


kranurag7 (approved):
looks good; we're checking the existence of entrypoints in the compat package and the main package tests looks good.

PR #53708, opened 2025-05-19
glibc: no change rebuild to fix unstripped binaries

dannf:
There was a bug in `melange` introduced in v0.23.13[0] and subsequently
fixed in v0.23.15[1] that prevented the strip pipeline from working.

[0] https://github.com/chainguard-dev/melange/commit/6935178aa924131a38f7067927649c73c50dc1d3
[1] https://github.com/chainguard-dev/melange/commit/b489539855bdcb981e2ecac3b2de1280da8f0c71

### DIFF
--- a/glibc.yaml
+++ b/glibc.yaml
@@ -2,7 +2,7 @@ package:
   name: glibc
   version: "2.41"
   # Every glibc update causes build disruption; always announce on #eng-psa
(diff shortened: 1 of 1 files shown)


sergiodj (approved):
LGTM, thanks for finding the issue.

PR #53769, opened 2025-05-20
chore(pipelines/auth): Configure pip to retrieve guarded libraries 

EyeCantCU:
Introduce pipeline that configures pip to fetch guarded libraries


dannf (approved):
fwiw, LGTM

dannf (changes requested):
actually, +1 on the pipeline, but let's not use it for any packages yet. that's an active discussion. But it's useful for testing in the meantime... so, maybe we should require users to set an environment variable to enable it? e.g. `PIP_TESTING_DONT_MERGE_THIS=1` ?

PR #53792, opened 2025-05-20
Support systemd-logind in our sshd pam session

zmarano:
The user sessions via ssh needs to be registered with systemd-logind to support features like lingering sessions

### DIFF
--- a/openssh/sshd.pam
+++ b/openssh/sshd.pam
@@ -22,6 +22,9 @@ session required pam_mkhomedir.so
 # Set the loginuid process attribute.
 session required pam_loginuid.so
 
+# Use pam_systemd to register this session with logind
+session required pam_systemd.so
+
 # Standard base session setup and teardown.
 session include base-session-noninteractive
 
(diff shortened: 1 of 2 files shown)


openssh/sshd.pam
```diff
@@ -22,6 +22,9 @@ session required pam_mkhomedir.so
 # Set the loginuid process attribute.
 session required pam_loginuid.so
 
+# Use pam_systemd to register this session with logind
+session required pam_systemd.so
```
smoser, line 26: that is already there via linux-pam
```
$ apk info --who-owns /etc/pam.d/base-session-noninteractive
/etc/pam.d/base-session-noninteractive is owned by linux-pam-1.7.0-r42

$ sudo grep -r systemd /etc/pam.d/
/etc/pam.d/base-session-noninteractive:session optional pam_systemd.so
```

zmarano: Yup- you're right! Closing this PR.

PR #53839, opened 2025-05-21
added package: nifi-registry

rahulduvedi:
**added package: apache/nifi-registry:2.3.0**


nifi-registry.yaml
```diff
@@ -0,0 +1,149 @@
+package:
+  name: nifi-registry
+  version: "2.4.0"
+  epoch: 10
+  description: Apache NiFi Registry is a registry for storing and managing shared resources such as versioned flows across one or more instances of NiFi.
+  copyright:
+    - license: Apache-2.0
+  resources:
+    cpu: 5
+    memory: 2Gi
```
Dentrax, line 10: Not sure if we do need this.

nifi-registry.yaml
```diff
@@ -0,0 +1,149 @@
+package:
+  name: nifi-registry
```
Dentrax, line 2: FWIW package name should be `apache-nifi-registry` instead.

nifi-registry.yaml
```diff
@@ -0,0 +1,149 @@
...
+    use-tag: true
+    strip-prefix: rel/nifi-
+    tag-filter: rel/nifi-
+
+test:
+  environment:
+    contents:
+      packages:
+        - java-common-jre
+        - openjdk-21-jre
```
Dentrax, line 133: Provided by runtime already.
```suggestion
```

nifi-registry.yaml
```diff
@@ -0,0 +1,149 @@
...
+      NIFI_REGISTRY_LOG_DIR: /opt/nifi-registry/nifi-registry-current/logs
+  pipeline:
+    - uses: test/daemon-check-output
+      with:
+        start: /opt/nifi-registry/scripts/start.sh start
+        expected_output: |
+          Started Server on
+          Successfully initiated communication with Bootstrap
+          Started Application
+        timeout: 60
```
Dentrax, line 149: Can we have `post:` test? Like `wait-for-it` + `curl` + `grep` combination?

nifi-registry.yaml
```diff
@@ -0,0 +1,149 @@
...
+        - openjdk-21-jre
+    environment:
+      JAVA_HOME: /usr/lib/jvm/java-21-openjdk
+      NIFI_REGISTRY_BASE: /opt/nifi-registry
+      NIFI_REGISTRY_HOME: /opt/nifi-registry/nifi-registry-current
+      TOOLKIT_HOME: /opt/nifi-registry/nifi-toolkit-current
+      NIFI_REGISTRY_PID_DIR: /opt/nifi-registry/nifi-registry-current/run
+      NIFI_REGISTRY_LOG_DIR: /opt/nifi-registry/nifi-registry-current/logs
+  pipeline:
+    - uses: test/daemon-check-output
```
Dentrax, line 142: Lets have another simple test step to ensure that some important files, directories, entrypoints, etc. are exist

nifi-registry.yaml
```diff
@@ -0,0 +1,149 @@
...
+          unzip nifi-registry-toolkit-${{package.version}}-bin.zip \
+            -d "${{targets.subpkgdir}}/${TOOLKIT_HOME}"
+          rm -f "${{targets.subpkgdir}}/${TOOLKIT_HOME}/bin/"*.bat
+          ln -sf "${TOOLKIT_HOME}" \
+                 "${{targets.subpkgdir}}/${TOOLKIT_HOME%/*}/nifi-toolkit-registry-${{package.version}}"
+      - uses: strip
+    test:
+      pipeline:
+        - runs: |
+            test -d /opt/nifi-registry/nifi-toolkit-current
```
Dentrax, line 115: Can we also include the `readlink` here to ensure that the symlink is pointing valid file.

nifi-registry.yaml
```diff
@@ -0,0 +1,149 @@
...
+  copyright:
+    - license: Apache-2.0
+  resources:
+    cpu: 5
+    memory: 2Gi
+  dependencies:
+    runtime:
+      - bash
+      - busybox
+      - coreutils
```
Dentrax, line 15: Why we add `busybox` and `coreutils` at the same time?

nifi-registry.yaml
```diff
@@ -0,0 +1,149 @@
+package:
+  name: nifi-registry
+  version: "2.4.0"
+  epoch: 10
```
Dentrax, line 4:
```suggestion
  epoch: 0
```

Dentrax (changes requested):

nifi-registry.yaml
rahulduvedi: Sorry for the oversight - I added both without thinking clearly; was aiming for complete utility coverage but didn't consider the redundancy between busybox and coreutils.

nifi-registry.yaml
rahulduvedi: Apologies, I was testing it locally and was updating epoch
I should be more careful.

nifi-registry.yaml
rahulduvedi: Added `readlink` here

nifi-registry.yaml
rahulduvedi: added more tests

nifi-registry.yaml
rahulduvedi: Thanks for suggestions, I removed these

nifi-registry.yaml
rahulduvedi: Added tests to check these

apache-nifi-registry.yaml
```diff
@@ -0,0 +1,211 @@
...
+    runtime:
+      - bash
+      - busybox
+      - git
+      - java-common-jre
+      - openjdk-${{vars.java-version}}-jre
+      - xmlstarlet
+
+vars:
+  java-version: '21'
```
tulilirockz, line 18: Mind adding a little comment explaining when we should bump this?

Dentrax (dismissed):
LGTM, thanks so much for the great work and addressing the comments here! @rahulduvedi

apache-nifi-registry.yaml
rahulduvedi: Added inline comment! @tulilirockz

Dentrax (approved):

PR #53856, opened 2025-05-21
Pgpool package update: Add busybox to pgpool2-oci-entrypoint subpackage

AnmolVirdi:
- Added `busybox` to entrypoint subpackage and bump up epoch.

### DIFF
--- a/pgpool2-4.6.yaml
+++ b/pgpool2-4.6.yaml
@@ -1,7 +1,7 @@
 package:
   name: pgpool2-4.6
   version: "4.6.1"
-  epoch: 0
+  epoch: 1
   description: Middleware that works between PostgreSQL servers and a PostgreSQL database client.
   copyright:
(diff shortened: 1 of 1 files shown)


pgpool2-4.6.yaml
```diff
@@ -104,6 +104,7 @@ subpackages:
         - grep
         - gawk
         - gnutar
+        - busybox # needed by entrypoint script
```
kranurag7, line 107: Is full busybox required or only sh (that's available via dash-binsh)

AnmolVirdi: `sed`, `hostname`, etc are required.

AnmolVirdi: Dockerfile reference: https://github.com/pgpool/pgpool2_on_k8s/blob/master/pgpool.docker/entrypoint.sh

kranurag7 (approved):

PR #53894, opened 2025-05-21
feat(pipelines): add xcover pipelines

maxgio92:
This commit adds xcover pipelines, for starting a coverage profiling, stopping the profiling, and ensuring a minimum coverage threshold is respected, before completing tests.

The xcover/ensure pipeline will exits 1 if the threshold is not respected.

xcover/profile requires sys_admin to create BPF uprobes multi link, and sys_resource for setting memlock rlimit.


joshrwolf (approved):

maxgio92:
No longer needed since are now Melange built-in.

PR #53915, opened 2025-05-21
fix(update): Enable update for govulncheck using git

debasishbsws:
Package update and Enable update for govulncheck using git


kranurag7 (approved):
excellent!!

PR #53916, opened 2025-05-21
Makefile: make docker implementation configurable

kaniini:
e.g. make DOCKER=podman dev-container

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ else
 endif
 TARGETDIR = packages/${ARCH}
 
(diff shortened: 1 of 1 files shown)


xnox:
does melange know how to use nerdctl / podman instead of docker in the docker runner; or is that not needed yet?

kaniini:
I don't personally use the docker runner (I use bubblewrap), but it works through the docker socket, so isn't going to be compatible with those anyway.

PR #53969, opened 2025-05-22
calico-3.30: add goldmane and whisker

Dentrax:
https://www.tigera.io/blog/calico-open-source-3-30-exploring-the-goldmane-api-for-custom-kubernetes-network-observability/

3.30 introduced 2 new components: `goldmane` and `whisker`

### DIFF
--- a/calico-3.30.yaml
+++ b/calico-3.30.yaml
@@ -720,6 +720,7 @@ subpackages:
           packages: ./goldmane/cmd/health
           output: health
       - runs: |
+          # Upstream expects the executables to be in `/` dir: https://github.com/projectcalico/calico/blob/master/goldmane/docker/Dockerfile
           mkdir -p "${{targets.subpkgdir}}"/
           ln -sf /usr/bin/goldmane "${{targets.subpkgdir}}"/goldmane
           ln -sf /usr/bin/health "${{targets.subpkgdir}}"/health
@@ -788,10 +789,10 @@ subpackages:
               # https://github.com/projectcalico/calico/blob/master/whisker/docker-image/Dockerfile
               mkdir -p "${{targets.contextdir}}"/etc/nginx/
               mkdir -p "${{targets.contextdir}}"/etc/nginx/conf.d/
-              install -Dm755 docker-image/nginx-start.sh "${{targets.contextdir}}"/usr/bin/nginx-start.sh
-              install -Dm755 docker-image/nginx.conf ${{targets.contextdir}}/etc/nginx/
-              install -Dm755 docker-image/default.conf ${{targets.contextdir}}/etc/nginx/conf.d/
-              install -Dm755 LICENSE "${{targets.contextdir}}"/licenses/LICENSE
+              install -Dm644 docker-image/nginx-start.sh "${{targets.contextdir}}"/usr/bin/nginx-start.sh
+              install -Dm644 docker-image/nginx.conf ${{targets.contextdir}}/etc/nginx/
+              install -Dm644 docker-image/default.conf ${{targets.contextdir}}/etc/nginx/conf.d/
+              install -Dm644 LICENSE "${{targets.contextdir}}"/licenses/LICENSE
               mkdir -p "${{targets.contextdir}}"/usr/share/nginx/html
               cp -avR dist/* "${{targets.contextdir}}"/usr/share/nginx/html/
     test:
@@ -838,6 +839,7 @@ subpackages:
               start worker processes
             post: |
               wait-for-it -t 10 --strict localhost:8081 -- echo "NGINX is up"
+              curl -sfL -k "https://localhost:8081" | grep -qi "Welcome to nginx!"
 
   - name: "calico-whisker-backend-${{vars.short-package-version}}"
     description: "Whisker backend"
--- a/calico-3.30.yaml
+++ b/calico-3.30.yaml
@@ -686,6 +686,67 @@ subpackages:
           runs: |
             typha-client --help
 
+  - name: "calico-goldmane-${{vars.short-package-version}}"
+    dependencies:
+      provides:
+        - calico-goldmane=${{package.full-version}}
+      runtime:
+        - glibc
+        - merged-bin
+        - merged-usrsbin
+        - wolfi-baselayout
+    pipeline:
+      - uses: go/build
+        with:
+          modroot: .
+          packages: ./goldmane/cmd
+          output: goldmane
+      - uses: go/build
+        with:
+          modroot: .
+          packages: ./goldmane/cmd/flowgen
+          output: flowgen
+      - uses: go/build
+        with:
+          modroot: .
+          packages: ./goldmane/cmd/stream
+          output: stream
+      - uses: go/build
+        with:
+          modroot: .
+          packages: ./goldmane/cmd/flowgen
+          output: flowgen
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/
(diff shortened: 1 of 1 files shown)


calico-3.30.yaml
```diff
@@ -686,6 +688,193 @@ subpackages:
           runs: |
             typha-client --help
 
+  - name: "calico-goldmane-${{vars.short-package-version}}"
+    description: "Goldmane"
+    dependencies:
+      provides:
+        - calico-goldmane=${{package.full-version}}
+      runtime:
+        - glibc
```
tulilirockz, line 697: Doesn't the `go/build` pipeline bring this in?

calico-3.30.yaml
```diff
@@ -718,6 +910,8 @@ test:
         /usr/bin/start_runit 2>&1 | grep -q 'no configuration has been provided'
         /usr/bin/calico-typha --help
         /usr/bin/typha-client --help
+        /usr/bin/health --help
+        /usr/bin/stream --help
```
tulilirockz, line 914: Maybe a version check here also? - not sure about how the cli behaves

calico-3.30.yaml
```diff
@@ -718,6 +910,8 @@ test:
         /usr/bin/start_runit 2>&1 | grep -q 'no configuration has been provided'
         /usr/bin/calico-typha --help
         /usr/bin/typha-client --help
+        /usr/bin/health --help
```
tulilirockz, line 913: CI failure here, not sure why its not finding it tho

tulilirockz: Maybe we should move this to a `test:` under the actual subpackage?

calico-3.30.yaml
Dentrax: `stream` is a binary intended for internal usage and only has `--help` flag: https://github.com/projectcalico/calico/blob/1d3e5ce87449cc6cee6dcb604b2ef6d890ebb9fe/goldmane/cmd/stream/main.go

calico-3.30.yaml
Dentrax: Oh because we were not building this! Sorry for the oversight.

calico-3.30.yaml
Dentrax: That pipelines only provides the build-time dependencies!

tulilirockz: Oh LOL true

tulilirockz (dismissed):
Seems fine to me!

calico-3.30.yaml
```diff
@@ -686,6 +688,189 @@ subpackages:
...
+        with:
+          modroot: .
+          packages: ./goldmane/cmd/health
+          output: health
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/
+          ln -sf /usr/bin/goldmane "${{targets.subpkgdir}}"/goldmane
+          ln -sf /usr/bin/health "${{targets.subpkgdir}}"/health
+          ln -sf /usr/bin/flowgen "${{targets.subpkgdir}}"/flowgen
+          ln -sf /usr/bin/stream "${{targets.subpkgdir}}"/stream
```
OddBloke, line 727: Why are we linking into `/` here?

calico-3.30.yaml
```diff
@@ -686,6 +688,189 @@ subpackages:
...
+              yarn build
+          - name: Install the files
+            runs: |
+              # https://github.com/projectcalico/calico/blob/master/whisker/docker-image/Dockerfile
+              mkdir -p "${{targets.contextdir}}"/etc/nginx/
+              mkdir -p "${{targets.contextdir}}"/etc/nginx/conf.d/
+              install -Dm755 docker-image/nginx-start.sh "${{targets.contextdir}}"/usr/bin/nginx-start.sh
+              install -Dm755 docker-image/nginx.conf ${{targets.contextdir}}/etc/nginx/
+              install -Dm755 docker-image/default.conf ${{targets.contextdir}}/etc/nginx/conf.d/
+              install -Dm755 LICENSE "${{targets.contextdir}}"/licenses/LICENSE
```
OddBloke, line 794: Should these be `644`?

calico-3.30.yaml
```diff
@@ -686,6 +688,189 @@ subpackages:
...
+            - shadow
+      pipeline:
+        - name: Test important files
+          runs: |
+            #test "$(readlink /usr/sbin/nginx)" = "/usr/bin/nginx"
+            stat /usr/bin/nginx-start.sh
+            if [ -z "$(ls -A /usr/share/nginx/html/)" ]; then
+              echo "/usr/share/nginx/html/ is empty"
+              exit 1
+            fi
```
OddBloke, line 813: Instead of (or in addition to) this, could we `curl` something from the running server later, to confirm that (a) the files were installed, _and_ (b) nginx is correctly configured to serve them?

calico-3.30.yaml
```diff
@@ -686,6 +688,189 @@ subpackages:
...
+        - dash-binsh
+        - sed
+        - libacl1
+        - libattr1
+        - libcap
+        - libcrypt1
+        - libcrypto3
+        - libpcre2-8-0
+        - libselinux
+        - libssl3
```
OddBloke, line 772: Were these dependencies not correctly detected by SCA (and generated as `so:` dependencies)?

calico-3.30.yaml
Dentrax: In the upstream Dockerfiles, they also symlink the files to `/`.

calico-3.30.yaml
Dentrax: It actually doesn't matter as they will get overridden in the `apko` manifests anyway. It seems `0o755` is mostly used.

calico-3.30.yaml
Dentrax: Yep, actually in the next step, we run `/usr/bin/nginx-start.sh` to do further check.

calico-3.30.yaml
Dentrax: I suspect if it will, upstream Dockerfile particularly `apt-install` those maybe for a valid reason (i.e., runtime need etc.) Checked the final APK build here: https://apk.chaindag.dev/https/apk.cgr.dev/wolfi-presubmit/5c1e2f6cf28811954bb7df91f0036f1be5548438/aarch64/calico-whisker-3.30-3.30.0-r2.apk@etag:4cfdeebe73c5186ade0e71df0f90fa07/.PKGINFO

I able to see the `depends:` as expected. So having explicitly here would be more safer (as we already tests too!).

calico-3.30.yaml
OddBloke: We only check that the server comes up: I'm suggesting that we confirm it's serving up the content that we expect it to.

calico-3.30.yaml
OddBloke: I'd like to see at least a comment to that effect.

AFAICT nothing else in this package is symlinked to `/`: do we know why they do that?

calico-3.30.yaml
Dentrax: Added `curl` test! https://github.com/wolfi-dev/os/pull/53969/commits/2d42f20eb7098eda4491000d491dcb755e990c7a

calico-3.30.yaml
Dentrax: > we'll run into issues in the future due to the requirements upstream changing

Yeah that makes sense. I just tested by removing those `- lib*`s and test are still passing! Means SCA already adds those. But only needed to add `libcap`, test was complaining - and they also add this lib.

OddBloke (approved):

PR #53978, opened 2025-05-22
ollama{,-cpu}: better tests for ollama CPU, correct ldflags and deps for ollama

tulilirockz:
Honestly just a few miscelaneous improvements that were found with this package

### DIFF
--- a/ollama.yaml
+++ b/ollama.yaml
@@ -38,13 +38,11 @@ subpackages:
       - uses: cmake/build
       - uses: cmake/install
     test:
-      environment:
-        environment:
-          LD_LIBRARY_PATH: /usr/lib/ollama/
       pipeline:
         - uses: test/tw/ldd-check
           with:
             packages: ${{package.name}}-cpu
+            extra-library-paths: /usr/lib/ollama
 
 update:
   enabled: true
--- a/ollama.yaml
+++ b/ollama.yaml
@@ -1,7 +1,7 @@
 package:
   name: ollama
   version: "0.7.1"
-  epoch: 0
+  epoch: 1
   description: Get up and running with Llama 2 and other large language models locally
   copyright:
     - license: MIT
--- a/ollama.yaml
+++ b/ollama.yaml
@@ -3,9 +3,6 @@ package:
   version: "0.7.1"
   epoch: 0
   description: Get up and running with Llama 2 and other large language models locally
-  dependencies:
-    runtime:
-      - ${{package.name}}-cpu
   copyright:
     - license: MIT
 
--- a/ollama.yaml
+++ b/ollama.yaml
@@ -3,6 +3,9 @@ package:
   version: "0.7.1"
   epoch: 0
   description: Get up and running with Llama 2 and other large language models locally
+  dependencies:
+    runtime:
+      - ${{package.name}}-cpu
   copyright:
     - license: MIT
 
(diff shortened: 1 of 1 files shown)


ollama.yaml
```diff
@@ -18,32 +21,30 @@ pipeline:
...
     pipeline:
       - uses: cmake/configure
         with:
           opts: --preset "CPU"
       - uses: cmake/build
       - uses: cmake/install
-      - name: Install LibGGML CPU to main LD_LIBRARY_PATH
-        runs: |
-          set -e
-          install -Dpm0644 -t ${{ targets.subpkgdir }}/usr/lib/ ${{ targets.subpkgdir }}/usr/lib/ollama/libggml*.so
```
Dentrax, line 41: Why we are removing this? Would this be a breaking change?

tulilirockz: Nope, nothing uses it, and this was a workaround to make the `ldd` tests work. Ollama's official image uses `/usr/lib/ollama/libggml*` instead of `/usr/lib/libggml*`, too. Ended up figuring out that `LD_LIBRARY_PATH` works for the test so this isnt useful anymore! :)

ollama.yaml
```diff
@@ -18,32 +21,30 @@ pipeline:
...
     test:
+      environment:
+        environment:
+          LD_LIBRARY_PATH: /usr/lib/ollama/
```
kranurag7, line 43: instead of using this, we can use the following which is part of the pipeline options:

```yaml
  extra-library-paths:
    description: |
      Prepend the provided string to the LD_LIBRARY_PATH.
      For more than one path, use a ":" delimited string.
    required: false
    default: ""
```

tulilirockz: https://github.com/wolfi-dev/os/pull/53978/commits/f93e8bfd9184f4ded90c9ba940b3de6733c4803c

tulilirockz: Thanks! Didn't think about searching for this on the melange docs :P

ollama.yaml
```diff
@@ -1,8 +1,11 @@
 package:
   name: ollama
   version: "0.7.0"
-  epoch: 1
+  epoch: 2
   description: Get up and running with Llama 2 and other large language models locally
+  dependencies:
+    runtime:
+      - ${{package.name}}-cpu
```
EyeCantCU, line 8: If we need `ollama-gpu` and want to use the same binary, this breaks that use case

EyeCantCU (changes requested):

ollama.yaml
tulilirockz: We should maybe make `ollama-gpu` depend on `ollama-cpu` then. Having the `ollama-cpu` depend on `ollama` makes it so `ollama-fips` pulls in... `ollama` (non-fips)

tulilirockz: e8ce192aa5fb1e5d9dd3d131e787a46270f74df1

Dentrax (dismissed):

ollama.yaml
```diff
@@ -18,32 +18,28 @@ pipeline:
...
-  - uses: strip
+      ldflags: |
+        -X=github.com/ollama/ollama/version.Version=${{package.version}}
+        -X=github.com/ollama/ollama/server.mode=release
 
 subpackages:
   - name: ${{package.name}}-cpu
     description: Ollama CPU inferencing component
     dependencies:
       runtime:
```
Dentrax, line 29: Can we have `provides:` here as the APK provides, maybe something like:

```yaml
provides:
      - libggml-base=${{package.full-version}}
      - libggml-cpu=${{package.full-version}}
```

Or we should think to rename the subpackage to `libggml`.

APK only has those `SO` files.

tulilirockz: I dont think this is valid for this tho, I think that this libggml thing from ollama is ollama specific, we probably would have a separate package for libggml if we were to distribute it outside of `/usr/lib/ollama`... :think:

EyeCantCU (dismissed):

Dentrax (dismissed):

Dentrax (approved):

PR #54002, opened 2025-05-22
feat: add additonal tests for kong

rakshitgondwal:
This PR adds additional tests for the `kong` package


kong.yaml
```diff
@@ -86,9 +86,50 @@ update:
...
+                  paths: [ / ]
+          CFG
+
+          cat <<'SH' >/tmp/run-kong.sh
+          #!/usr/bin/env bash
+          set -euo pipefail
+          export KONG_DATABASE=off
+          export KONG_DECLARATIVE_CONFIG=/tmp/kong.yml
+          export KONG_ADMIN_LISTEN=127.0.0.1:8001
+          export KONG_PROXY_LISTEN=127.0.0.1:8000
```
Dentrax: lets move this env vars to above, `environment.environment`.

kong.yaml
```diff
@@ -86,9 +86,50 @@ update:
...
+          kong prepare -p /tmp/kong_prefix
+          exec kong start -v -p /tmp/kong_prefix
```
Dentrax, line 125: We can keep only `kong prepare -p /tmp/kong_prefix` in the `setup:`. And move the `kong start -v -p /tmp/kong_prefix` to `start:` field.

Dentrax (changes requested):

kong.yaml
rakshitgondwal: done

kong.yaml
rakshitgondwal: weirdly, that doesn't work
i am writing these commands in a script and then invoking that script in the `start` section

Dentrax (approved):

PR #54022, opened 2025-05-22
glibc - add test for locale packages

smoser:


glibc.yaml
```diff
@@ -753,6 +753,26 @@ subpackages:
...
         - merged-usrsbin
         - wolfi-baselayout
+    test:
+      environment:
+        contents:
+          packages:
+            - posix-libc-utils
+      pipeline:
+        - runs: |
+            set +x
```
a-crate, line 763: Doesn't melange --interactive set this? Do you need to do it explicitly?

a-crate (approved):
IMO you should bump epoch and push a commit to show it passes in CI, then revert and merge. But does not need to block if it passes locally.

smoser:
a draft pr with glibc bump at https://github.com/wolfi-dev/os/pull/54059 so we can see the tests run.

PR #54038, opened 2025-05-22
Chromium(Test): Add test for subpackages by checking the files and symlink

debasishbsws:
Basic test, running it locally passes, didn't bump epoch as it takes an hour to build.

### DIFF
--- a/chromium.yaml
+++ b/chromium.yaml
@@ -8,7 +8,7 @@
 package:
   name: chromium
   version: "136.0.7103.113"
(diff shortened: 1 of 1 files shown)


murraybd (changes requested):
Please bump the epoch so the tests will run in the infrastructure.

powersj (dismissed):

murraybd (approved):

powersj (approved):

PR #54039, opened 2025-05-22
quality: improve test for kubernetes-dashboard-auth package

sanjayk0508:
In package tests, we run the `dashboard-auth` service directly on localhost and cannot use `kubectl proxy`. This means we cannot test API endpoints like `/api`, `/livez`, or `/readyz`, as they are exposed via the k8s API server (on port 8001) and not directly by the service on port 8000.

### DIFF
--- a/kubernetes-dashboard-auth.yaml
+++ b/kubernetes-dashboard-auth.yaml
@@ -48,13 +48,14 @@ test:
   pipeline:
     - uses: test/kwok/cluster
     - name: "Setup test environment"
+    - name: Verify kubernetes-dashboard-auth installation
       runs: |
-        # Create service account and RBAC
-        kubectl create serviceaccount test-user
-        kubectl create clusterrolebinding test-user-admin \
-          --clusterrole=cluster-admin \
-          --serviceaccount=default:test-user
-        sleep 2
+        dashboard-auth --address=0.0.0.0 --apiserver-host="https://accounts.google.com" --kubeconfig=/root/.kube/config &> dashboard-auth.log &
+        sleep 5
+        grep "Starting Kubernetes Dashboard Auth" dashboard-auth.log || exit 1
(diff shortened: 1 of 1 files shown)


kranurag7:
we can consider using `kubectl get --raw /` if we're able to query the endpoints that way.

sanjayk0508:
thanks @kranurag7 for the suggestion, have update the test, works fine, LMK if we need anything else

kubernetes-dashboard-auth.yaml
```diff
@@ -51,6 +56,37 @@ test:
...
+            --apiserver-host="https://127.0.0.1:32764" \
+            --kubeconfig=/root/.kube/config \
+            --apiserver-skip-tls-verify \
+            -v=4
+        timeout: 30
+        expected_output: |
+          "Starting Kubernetes Dashboard Auth" version="${{package.version}}"
+          "Listening and serving insecurely on" address="0.0.0.0:8000"
+        post: |
+          wait-for-it -t 30 localhost:8000 || exit 1
```
Dentrax, line 80: Should exit by default.
```suggestion
          wait-for-it -t 30 localhost:8000
```

kubernetes-dashboard-auth.yaml
```diff
@@ -51,6 +56,37 @@ test:
...
+
+          # Check Kubernetes API health
+          for endpoint in / /readyz /livez /openapi/v2; do
+          echo "Checking $endpoint..."
+          kubectl get --raw "$endpoint" || {
+            echo "Failed to query $endpoint"
+            exit 1
+          }
+          done
```
Dentrax, line 89: Not sure if we need this as we are talking with API server, not with the dashboard.

Dentrax (changes requested):

PR #54047, opened 2025-05-22
mariadb-connector-c: fetch tarball from upstream directly

OddBloke:


OddBloke:
I've confirmed locally that the tarball SHAs are the same. Package Update Config Check is failing, but the latest release was today so I hope that's just due to tarball publication delay: if not, we can fix it further in the package update PR.

powersj (approved):

PR #54062, opened 2025-05-22
feat: add ghscan package

egibs:
This PR adds a package for our `ghscan` package. No CVEs and fairly straightforward to build 🙂


eslerm (approved):
Cheers.

Is ghscan still as functional as it once was? I thought GitHub made some breaking changes.

PR #54069, opened 2025-05-22
feat(cgal.yaml): add emptypackage test to cgal

philroche:
feat( cgal.yaml): add emptypackage test to cgal


philroche:

PR #54090, opened 2025-05-22
feat(gdal.yaml): add emptypackage test to gdal

philroche:
feat( gdal.yaml): add emptypackage test to gdal


philroche:

PR #54255, opened 2025-05-24
nftables: add json support

ajayk:


xnox:
This got fixed in:
- https://github.com/wolfi-dev/os/pull/73340

I wish we had landed this PR way back when.

PR #54283, opened 2025-05-25
nextcloud-server: version stream

Dentrax:
Version stream for nextcloud-server: https://endoflife.date/nextcloud


kranurag7 (approved):
looks good given it's the same package which we're moving to version stream now.

imho, if we bump the epoch then we're ignoring the possibility of a race b/w removal and addition on new package with `-r0`

kranurag7 (approved):

PR #54297, opened 2025-05-26
Fix(update): Use the mirror github repo with git-checkout and Update using GitMonitor

debasishbsws:
- package version update 
- change update to use GitMonitor
- use git-checkout to the github repo


OddBloke (approved):
I've confirmed that the 3.4.5 tarball and v3.4.5 tag have the same content. Thanks!

PR #54300, opened 2025-05-26
Fix(checkov): Build, Package Update & Enable auto Update

debasishbsws:
- Package Update to version 3.2.432
- python 3.12 using pip build install pipeline
- Enable Update for the package using GitMonitor


debasishbsws:
Why ci apk add run before elastic-build end

OddBloke (approved):

PR #54303, opened 2025-05-26
fix(etcd-3.6): use version path 3.6 for etcd bitnami compat

vishal-chdhry:
We are currently using `version-path: 3.5/debian-12` in etcd 3.6 bitnami, which might be causing an image build failure, this PR updates it to 3.6


debasishbsws (approved):
Looks good

PR #54375, opened 2025-05-27
add envoy-iamguarded package

kranurag7:


wojciechka (dismissed):
LGTM, but want to be sure you're ok with symlinking `envoy.yaml` instead of copying it?

I am fine either way.

wojciechka (approved):

PR #54402, opened 2025-05-27
stow: Add GNU Stow package

IdlePhysicist:
Here I add the GNU Stow package. https://www.gnu.org/software/stow/

### DIFF
--- a/stow.yaml
+++ b/stow.yaml
@@ -1,0 +1,61 @@
+package:
+  name: stow
+  version: 2.4.1
+  epoch: 0
+  description: "GNU stow"
+  copyright:
+    - license: GPL-3.0
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - perl
+      - texinfo
+
+pipeline:
+  - uses: fetch
+    with:
+      uri: https://ftpmirror.gnu.org/gnu/stow/stow-${{package.version}}.tar.gz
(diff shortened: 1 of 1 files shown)


stow.yaml
```diff
@@ -0,0 +1,62 @@
...
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - perl
+      - texinfo
+
+pipeline:
+  - uses: fetch
+    with:
+      uri: https://ftp.gnu.org/gnu/stow/stow-${{package.version}}.tar.gz
```
sergiodj: You might want to use `ftpmirror.gnu.org` here, as `ftp.gnu.org` is less stable.

stow.yaml
```diff
@@ -0,0 +1,62 @@
...
+    pipeline:
+      - uses: split/manpages
+    description: stow manpages
+    test:
+      pipeline:
+        - uses: test/docs
+
+update:
+  enabled: true
+  manual: true
```
sergiodj: OOC, any specific reason for enabling `manual: true` here?

IdlePhysicist: That was an oversight, I used make.yaml as my basis for this, and it uses `manual: true`

sergiodj (approved):

stow.yaml
```diff
@@ -0,0 +1,61 @@
+package:
+  name: stow
+  version: 2.4.1
+  epoch: 0
+  description: "GNU stow"
+  copyright:
+    - license: GPL-3.0
```
sergiodj, line 7: Just a comment that the license is `GPL-3.0-or-later`, but that can be fixed in a subsequent commit.

PR #54470, opened 2025-05-27
add py3-safetensors package

kranurag7:


Aditevil (approved):

py3-safetensors.yaml
```diff
@@ -0,0 +1,75 @@
...
+  pypi-package: safetensors
+  module-name: safetensors
+
+data:
+  - name: py-versions
+    items:
+      3.10: '310'
+      3.11: '311'
+      3.12: '312'
+      3.13: '313'
```
tulilirockz, line 21: (non-blocker): I feel like we could use `vars-transforms` here, not sure tho

tulilirockz (approved):

py3-safetensors.yaml
tulilirockz: Just an idea tbh, I don't know exactly how this would be done on melange

PR #54477, opened 2025-05-27
gcc: Update to GCC 14.3.0

sergiodj:
Update the current GCC to 14.3.0 (the latest release under the 14.x branch) so that we can unblock #54401.

This is part of the GCC 15 update process.


xnox (changes requested):
surely fetch expected-sha512 needs updating.

sergiodj:
Patches are not needed anymore (great).  Removing them...

sergiodj:
We'll hold off on upgrading to GCC 14.3 for now.

PR #54527, opened 2025-05-28
Add a link for modules to nginx package

max-allan-cgr:
Nginx does not use the --module-path when loading modules... It uses the --prefix directory.

Rather than move the modules from /usr/lib/nginx to /var/lib/nginx, I made a symlink. Maybe we should just move them but that feels like it could have knock on consequences.


mritunjaysharma394 (approved):
Thanks Max, like the symlink approach to not break things

PR #54528, opened 2025-05-28
No-change rebuild packages to ensure newer SBOM generation (1/2).

sil2100:
No-change rebuild bumps for newer SBOMs. We have recently added `downloadLocation` support, and we'd need some of those to include this ASAP.

Dividing this into smaller batches. Batch 1 out of 2.

### DIFF
--- a/curl.yaml
+++ b/curl.yaml
@@ -1,7 +1,7 @@
 package:
   name: curl
   version: "8.14.0"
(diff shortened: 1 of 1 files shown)


xnox:
looks like the big one got merged; and thus here only curl is left.

justinvreeland (approved):

PR #54546, opened 2025-05-28
withdraw py3-jsonpatch-1.33-r0

MickeyPvX:
Withdrawing `py3-jsonpatch` packages that have `provides: py3-jsonpatch`


smoser (commented):
I think we want to remove the "main" package also

smoser (approved):

PR #54561, opened 2025-05-28
Revert "src/6.3.1 package update (#53327)"

xnox:
Never built, and tag appears to have been removed upstream. Thus
simply revert this commit. Currently this is causing elasticbuild
failure on every push.

Note that 6.3.0 upstream tag exists, so automation should attempt
upgrade to that, once this is merged.

This reverts commit f089055c2a6a9cdc75fd4ffeda6b5fcdee88e75f.

### DIFF
--- a/src.yaml
+++ b/src.yaml
@@ -1,6 +1,6 @@
 package:
   name: src
-  version: "6.3.1"
+  version: "6.2.0"
   epoch: 0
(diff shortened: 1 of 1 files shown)


kranurag7 (approved):

philroche:
Thanks. Ah I misread the diff

PR #54602, opened 2025-05-28
chainguard-security-guide: update to CG stig 3.2.2, add tests

stevebeattie:
v3.2.2 release tightens up the package pattern match in the Remote Services check to avoid false positives.

Also add tests of individual rules and checks to ensure that we don't regress in the future:

- ensure the certificate bundle hash passes / commit matches (will fail on updates to ca-certificates as the stig will need to be updated)

- ensure the "no remote services" check passes even when python 3.12 with telnetlib.py in the standard python library is installed (telnetlib.py was removed in python 3.13, will need to come up with a different check then).

Ref: https://github.com/chainguard-dev/stigs/pull/14
Ref: https://github.com/chainguard-dev/prodsec/issues/220

### DIFF
--- a/chainguard-security-guide.yaml
+++ b/chainguard-security-guide.yaml
@@ -1,6 +1,6 @@
 package:
   name: chainguard-security-guide
-  version: "3.2.1"
+  version: "3.2.2"
   epoch: 0
   description: Security automation content for Chainguard Images
   copyright:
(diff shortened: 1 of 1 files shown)


stevebeattie:
Rebased to tip of main to get a new build attempt.

egibs (approved):

PR #54604, opened 2025-05-28
spdx-tools-java: fix schmea with spec compliant dashed operating-system

xnox:
Current schema upstream use underscore, where as the spec requires to
use a dash. Fix the constant in the schema to be with a dash.

Avoid a breaking change by accepting both dash and underscore.

Reference: https://github.com/spdx/tools-java/pull/211


sil2100 (approved):
Good idea with supporting both for now!

PR #54622, opened 2025-05-28
python-3.10, python-3.11, python-3.12, python-3.13: Add missing cmd provides (P0)

dakaneye:
## Summary
This PR fixes a P0 issue where APKO builds were resolving to ancient Python versions (e.g., python-3.12=3.12.2-r6) instead of current versions (e.g., 3.12.10-r1).

## Root Cause
The python-3.XX-base subpackages were missing critical `cmd:` provides entries that APKO uses for package resolution. When packages like numpy request `cmd:python3.12`, APKO couldn't find it in the current packages and fell back to very old versions that still had these provides.

This was an oversight from the March 2024 refactoring where the Python binaries were moved from the main package to the -base subpackage, but the provides metadata wasn't updated accordingly.

## Changes
Added the following provides to all python-3.XX-base subpackages (3.10, 3.11, 3.12, 3.13):
```yaml
provides:
  - cmd:python${{vars.pyversion}}=${{package.full-version}}
  - cmd:pydoc${{vars.pyversion}}=${{package.full-version}}
```

Also incremented the epoch for all affected packages to ensure updates are applied.

## Testing
- Verified that the binaries exist in the -base subpackages

Related to [#12590](https://github.com/chainguard-dev/internal-dev/issues/12590)

## Impact
This is a P0 issue affecting production images including:
- management-api-for-apache-cassandra-4.1
- py3-cassandra-medusa
- Other images that depend on specific Python versions via cmd: provides

### DIFF
--- a/python-3.10.yaml
+++ b/python-3.10.yaml
@@ -1,7 +1,7 @@
 package:
   name: python-3.10
   version: "3.10.17"
-  epoch: 1
+  epoch: 2
   description: "the Python programming language"
   copyright:
     - license: PSF-2.0
@@ -119,6 +119,9 @@ subpackages:
   - name: "${{package.name}}-base"
     description: "${{package.name}} without /usr/bin/python3"
     dependencies:
+      provides:
+        - cmd:python${{vars.pyversion}}=${{package.full-version}}
+        - cmd:pydoc${{vars.pyversion}}=${{package.full-version}}
       runtime:
         - py3-setuptools-wheel
         - py3-pip-wheel
--- a/python-3.11.yaml
+++ b/python-3.11.yaml
@@ -1,7 +1,7 @@
 package:
   name: python-3.11
   version: "3.11.12"
-  epoch: 1
+  epoch: 2
   description: "the Python programming language"
   copyright:
     - license: PSF-2.0
@@ -119,6 +119,9 @@ subpackages:
   - name: "${{package.name}}-base"
     description: "${{package.name}} without /usr/bin/python3"
     dependencies:
+      provides:
+        - cmd:python${{vars.pyversion}}=${{package.full-version}}
+        - cmd:pydoc${{vars.pyversion}}=${{package.full-version}}
       runtime:
         - py3-setuptools-wheel
         - py3-pip-wheel
--- a/python-3.12.yaml
+++ b/python-3.12.yaml
@@ -1,7 +1,7 @@
 package:
   name: python-3.12
   version: "3.12.10"
-  epoch: 1
+  epoch: 2
   description: "the Python programming language"
   copyright:
     - license: PSF-2.0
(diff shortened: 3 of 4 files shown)


javacruft:
Looking at python-3.11-base:

```
curl -sL https://packages.wolfi.dev/os/x86_64/python-3.11-base-3.11.12-r1.apk | tar -Oxz .PKGINFO
# Generated by melange
pkgname = python-3.11-base
pkgver = 3.11.12-r1
arch = x86_64
size = 39702011
origin = python-3.11
pkgdesc = python-3.11 without /usr/bin/python3
url = 
commit = 3ed62612508d432c8325799a6d55e0e3a3c9f1c7
maintainer = wolfi
builddate = 1745929999
license = PSF-2.0
depend = py3-pip-wheel
depend = py3-setuptools-wheel
depend = so:ld-linux-x86-64.so.2
depend = so:libbz2.so.1
depend = so:libc.so.6
depend = so:libcrypt.so.1
depend = so:libcrypto.so.3
depend = so:libexpat.so.1
depend = so:libffi.so.8
depend = so:libgdbm.so.6
depend = so:liblzma.so.5
depend = so:libm.so.6
depend = so:libmpdec.so.4
depend = so:libncursesw.so.6
depend = so:libpanelw.so.6
depend = so:libreadline.so.8
depend = so:libsqlite3.so.0
depend = so:libssl.so.3
depend = so:libz.so.1
provides = cmd:pydoc3.11=3.11.12-r1
provides = cmd:python3.11=3.11.12-r1
provides = so:libpython3.11.so.1.0=1.0
datahash = 973936a1ffcb706dede405cdeb69c5bda1b83bfecfaad2b4570e71295236e0e2
```

The existing version of this package already provides cmd:pythonX.Y - how does explicitly adding them to the package change the behavior in apko?

xnox:
> This PR fixes a P0 issue where APKO builds were resolving to ancient Python versions (e.g., python-3.12=3.12.2-r6) instead of current versions (e.g., 3.12.10-r1).

what is the full apko config? are there any other shared libraries or packages pinned to a version stream, that consequently pull-down the abi and thus install older versions of python to keep lower abi of shared libraries?
do we perchance have an incomplete abi transition somewhere?

xnox (changes requested):
https://apk.chaindag.dev/https/packages.wolfi.dev/os/x86_64/APKINDEX.tar.gz@etag:7ed053d7944a1af5f43071b48dce2403/APKINDEX?provide=cmd%3Apython3.13&depend=&full=true

shows that cmd:python3.13 is already provided by python-3.13 and it is with a full package version https://apk.chaindag.dev/https/packages.wolfi.dev/os/x86_64/python-3.13-base-3.13.3-r1.apk@sha1:3d9c12285b6aad8628c0c1079a22c25a3b3ca591/.PKGINFO

dakaneye:
Closing this PR as the investigation revealed that the APK packages already contain the correct  provides entries (as shown by @javacruft's analysis of the .PKGINFO). The provides are being auto-generated by melange when binaries are moved to subpackages.

The root cause of the APKO version resolution issue appears to be a Python version mismatch between packages, which is being addressed in PR #54663. That PR aligns py3-cassandra-medusa to use Python 3.11 (matching the cassandra packages) instead of Python 3.12, which should resolve the constraint conflicts that were causing APKO to fall back to ancient package versions.

Thank you to the reviewers for the helpful analysis that led to finding the actual root cause\!

PR #54645, opened 2025-05-29
added package: langfuse-worker

rahulduvedi:
added new package - `langfuse-worker:3.65.3`


langfuse-worker.yaml
```diff
@@ -0,0 +1,238 @@
...
+
+  - runs: |
+      npm install turbo@^1.13.4 --global
+
+      corepack enable
+      corepack prepare pnpm@9.5.0 --activate
+
+      # Configure architecture
+      if [ "${{build.arch}}" = "aarch64" ]; then
+        TARGET_CPU="arm64"
```
mritunjaysharma394: Is this helping us to isolate arch based binaries? I suppose not since you later manually remove them? If they are not helping, can we remove them?

rahulduvedi: Yes, Removed those redundant arch specific configuration, Thanks for pointing out.

PR #54700, opened 2025-05-29
drupal: init package

tulilirockz:
#### For new package PRs only


drupal-11.yaml
```diff
@@ -0,0 +1,298 @@
...
+      export COMPOSER_ALLOW_SUPERUSER=1
+      export COMPOSER_HOME="$(mktemp -d)"
+      install -d ${{targets.destdir}}/opt/drupal
+      mv vendor-binaries/vendor ${{targets.destdir}}/opt/drupal/
+
+      # Move files around to match upstream image structure
+      cp -r vendor-binaries/composer*  ${{targets.destdir}}/opt/drupal/
+      cp -r /usr/src/${{package.name}} ${{targets.destdir}}/opt/drupal/web
+      mv ${{targets.destdir}}/opt/drupal/web/.editorconfig ${{targets.destdir}}/opt/drupal/web/.editorconfig
+      mv ${{targets.destdir}}/opt/drupal/web/.gitattributes ${{targets.destdir}}/opt/drupal/web/.gitattributes
```
mritunjaysharma394: doesn't it look no-op?

drupal-11.yaml
```diff
@@ -0,0 +1,298 @@
...
+      rm -rf ${{targets.destdir}}/opt/drupal/web/.git*
+      chmod 755 ${{targets.destdir}}/opt/drupal/web
```
mritunjaysharma394, line 119: I think while doing install `install -d ${{targets.destdir}}/opt/drupal`, you get these web assets too right? If yes then you can fix the permissions while doing install itself with something like `install -Dm755`

drupal-11.yaml
astrojerms: Yep, thanks, fixed!

drupal-11.yaml
astrojerms: Not exactly, it only creates the drupal folder. I experimented with changing this and the chmod was still required.

mritunjaysharma394 (approved):

PR #54728, opened 2025-05-29
[shard 0] Pin GCC 14 build dependency on packages FTBFSing with GCC 15

sergiodj:
Several of our packages fail to build with GCC 15.  These build failures are being reported to and worked on by upstream, but until everything is OK we need to pin `gcc-14-default` as a build dependency for those packages to guarantee that they will keep building fine.

This is shard 0.

### DIFF
--- a/apache-kvrocks.yaml
+++ b/apache-kvrocks.yaml
@@ -1,7 +1,7 @@
 package:
   name: apache-kvrocks
   version: "2.12.1"
-  epoch: 0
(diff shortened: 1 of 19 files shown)


murraybd (dismissed):

sergiodj:
* Removed ceph (unrelated FTBFS)

murraybd (dismissed):

gaughen (approved):

PR #54743, opened 2025-05-29
[shard 6] Pin GCC 14 build dependency on packages FTBFSing with GCC 15

sergiodj:
Several of our packages fail to build with GCC 15. These build failures are being reported to and worked on by upstream, but until everything is OK we need to pin gcc-14-default as a build dependency for those packages to guarantee that they will keep building fine.

This is shard 6.

### DIFF
--- a/tileserver-gl.yaml
+++ b/tileserver-gl.yaml
@@ -1,7 +1,7 @@
 package:
   name: tileserver-gl
   version: "5.3.1"
-  epoch: 0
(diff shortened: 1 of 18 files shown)


murraybd (dismissed):

murraybd:
cargo running out of memory looks like https://github.com/wolfi-dev/os/issues/54536

gaughen (approved):

PR #54827, opened 2025-05-30
add opentel-java-instrumentation

anushkamittal20:
add opentel-java-instrumentation


opentelemetry-java-instrumentation.yaml
```diff
@@ -0,0 +1,42 @@
...
+  version: "2.16.0"
+  epoch: 0
+  description: OpenTelemetry auto-instrumentation and instrumentation libraries for Java
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - busybox
+  target-architecture:
+    - x86_64
```
Dentrax: Can we drop a comment why its only enabled for single arch?

opentelemetry-java-instrumentation.yaml
```diff
@@ -0,0 +1,42 @@
...
+update:
+  enabled: true
+  github:
+    identifier: open-telemetry/opentelemetry-java-instrumentation
+    strip-prefix: v
+
+test:
+  pipeline:
+    - runs: |
+        test -f /javaagent.jar
```
Dentrax, line 54: Is it possible to increase test coverage a bit, like:

```
- name: Smoke test: Run Java with agent
      runs: |
        echo 'public class Hello { public static void main(String[] args) { System.out.println("Hello!"); } }' > Hello.java
        javac Hello.java
        java -javaagent:/javaagent.jar Hello | grep -q "Hello!"

    - name: Validate agent initialization output
      runs: |
        java -javaagent:/javaagent.jar -version 2>&1 | grep -qi "opentelemetry"
```

(didn't test, gpt-generated)

opentelemetry-java-instrumentation.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/open-telemetry/opentelemetry-java-instrumentation
+      tag: v${{package.version}}
+      expected-commit: d53c0e95b6ba6fdc62d8cda0c370f9189838f932
+
+  - runs: |
+      ./gradlew -PskipTests=true :javaagent:shadowJar
```
Dentrax: why we didn't run `assemble`? https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/CONTRIBUTING.md#building-from-source

Dentrax (changes requested):

opentelemetry-java-instrumentation.yaml
anushkamittal20: Yes, thanks for pointing this out!

opentelemetry-java-instrumentation.yaml
anushkamittal20: gotcha!

opentelemetry-java-instrumentation.yaml
anushkamittal20: you are right, I should use this instead.
let me try this.

Dentrax (dismissed):

opentelemetry-java-instrumentation.yaml
```diff
@@ -0,0 +1,60 @@
...
+      # Cannot find a Java installation on your machine (Linux 6.14.7-r0-linux aarch64) matching: {languageVersion=11, vendor=any vendor, implementation=vendor-specific, nativeImageCapable=false}. Some toolchain resolvers had internal failures: foojay (Service 'SystemInfo' is not available (os=Linux 6.14.7-r0-linux aarch64, enabled=false).).
+      - openjdk-21-default-jdk
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - libudev
+      - openjdk-11
+      - openjdk-17
```
Dentrax, line 23: Are we using those? Should we build this for all supported JDK versions?

opentelemetry-java-instrumentation.yaml
```diff
@@ -0,0 +1,60 @@
...
+      - openjdk-21
+  environment:
+    JAVA_HOME: /usr/lib/jvm/java-21-openjdk
```
Dentrax: Let's have a variable above called `java-version` and replace `21` to `${{vars.java-version}}` everywhere.

opentelemetry-java-instrumentation.yaml
```diff
@@ -0,0 +1,60 @@
...
+  enabled: true
+  github:
+    identifier: open-telemetry/opentelemetry-java-instrumentation
+    strip-prefix: v
+
+test:
+  environment:
+    contents:
+      packages:
+        - openjdk-21-default-jdk
```
Dentrax, line 51: This shouldn't be a runtime dep in the test.

opentelemetry-java-instrumentation.yaml
```diff
@@ -0,0 +1,60 @@
...
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - busybox
+      # upstream simply adds the platform independent jar to the destdir
+      # https://github.com/open-telemetry/opentelemetry-operator/blob/main/autoinstrumentation/java/Dockerfile
+      # when we build the agent on aarch64 it errors out as some dependencies are not available for aarch64
+      # Cannot find a Java installation on your machine (Linux 6.14.7-r0-linux aarch64) matching: {languageVersion=11, vendor=any vendor, implementation=vendor-specific, nativeImageCapable=false}. Some toolchain resolvers had internal failures: foojay (Service 'SystemInfo' is not available (os=Linux 6.14.7-r0-linux aarch64, enabled=false).).
+      - openjdk-21-default-jdk
```
Dentrax: Does it work with `openjdk-21-jre`?

Dentrax (changes requested):

opentelemetry-java-instrumentation.yaml
anushkamittal20: I had added those earlier, removing them

opentelemetry-java-instrumentation.yaml
anushkamittal20: Got it

opentelemetry-java-instrumentation.yaml
anushkamittal20: removed!

opentelemetry-java-instrumentation.yaml
anushkamittal20: it seems to work, and makes sense as well

opentelemetry-java-instrumentation.yaml
```diff
@@ -0,0 +1,62 @@
...
+  github:
+    identifier: open-telemetry/opentelemetry-java-instrumentation
+    strip-prefix: v
+
+test:
+  environment:
+    contents:
+      packages:
+        # complains /bin/sh: javac: not found if not added
+        - openjdk-21-default-jdk
```
Dentrax, line 51: This should be a runtime dep above and we omit it here

anushkamittal20: we dont need it in the package as upstream image doesnt have it as well. Also this increases the size of the image a lot.

here we need it as we are using javac.

Dentrax: ack - thanks!

Dentrax (approved):

PR #54828, opened 2025-05-30
fix(graph-tool): Build for arm64 and for every supported Python version

EyeCantCU:
Builds graph tool for supported versions of Python and enables support for arm64. Additionally, splits away documentation

### DIFF
--- a/py3-graph-tool.yaml
+++ b/py3-graph-tool.yaml
@@ -88,9 +88,19 @@ subpackages:
           # Only link when needed
           sed -i -e 's/ -shared / -Wl,-O1,--as-needed\0/g' libtool
 
-          # Unfortunately cuts the number of threads used in
-          # half, but we need lots of memory per job
-          make -C . -j$(( $(nproc) / 2 ))
+          # Grab available memory
+          MEM_KB=$(awk '/^MemAvailable:/ { print $2; exit }' /proc/meminfo)
+          MEM_MB=$(( MEM_KB / 1024 ))
+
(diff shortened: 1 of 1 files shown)


py3-graph-tool.yaml
```diff
@@ -78,15 +82,17 @@ subpackages:
       - uses: autoconf/configure
         with:
           opts: |
-            --enable-openmp \
             --includedir=/usr/lib/cairomm-1.16 \
-            --disable-debug \
-            --disable-dependency-tracking \
```
EyeCantCU, line 84: Found that behavior here (sans dependency tracking which should be enabled) was default (and that some of these flags aren't valid)

py3-graph-tool.yaml
```diff
@@ -78,15 +82,17 @@ subpackages:
...
-          make -C . -j$(nproc) V=1
+          # Only link when needed
+          sed -i -e 's/ -shared / -Wl,-O1,--as-needed\0/g' libtool
+
+          # Unfortunately cuts the number of threads used in
+          # half, but we need lots of memory per job
+          make -C . -j$(( $(nproc) / 2 ))
       - uses: autoconf/make-install
-      - runs: find ${{targets.contextdir}} -name "*.so" -exec strip -g {} \;
+      - uses: strip
```
jamonation, line 95: This one I think we'll want to double-check. I think there was an issue with scanelf and strip being too conservative and the final package ended up being multiple GB larger than others.

jamonation (approved):

PR #54836, opened 2025-05-30
clickhouse: provide correct version values

astrojerms:
Fixes the clickhouse versioning so clickhouse --version reports the correct "tweak" version


kranurag7 (commented):
a test | grep ${{package.version}} will ensure that this doesn't happen again. 

Also a comment around the logic would be nice in case this runs into issue down the road.

kranurag7 (dismissed):

kranurag7:
thank you; excellent work here. testing it locally, I see an extra `.` at the end. We can consider removing the same as well. 
```bash
/ # clickhouse --version
ClickHouse local version 25.2.2.39.
```

kranurag7 (approved):

PR #54847, opened 2025-05-30
5795 redpanda package complement

utieyin:
"seastar-25.05.0: new package"

Fixes:

Related:

#### For new package PRs only


Dentrax (dismissed):

cryptopp.yaml
```diff
@@ -0,0 +1,95 @@
...
+    replace: _
+    to: mangled-package-version
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: "https://github.com/weidai11/cryptopp.git"
+      tag: "CRYPTOPP_${{vars.mangled-package-version}}"
+      expected-commit: "843d74c7c97f9e19a615b8ff3c0ca06599ca501b"
+
```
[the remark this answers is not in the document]
utieyin: I think `autoconf/configure` expects a configure file which this project doesn't have so using it causes the build to fail

cryptopp.yaml
```diff
@@ -0,0 +1,95 @@
...
+  - uses: autoconf/make
+
+  - uses: autoconf/make-install
+
+  - runs: |
+      mkdir -p ${{targets.destdir}}/usr/bin
+      ln -s /usr/local/bin/cryptest.exe ${{targets.destdir}}/usr/bin/cryptest
```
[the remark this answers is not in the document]
Dentrax: Hmmm should we omit `.exe` file ext in this case?

cryptopp.yaml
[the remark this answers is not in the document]
utieyin:
```
readelf -h /usr/local/bin/cryptest.exe
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 03 00 00 00 00 00 00 00 00 
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - GNU
  ABI Version:                       0
  Type:                              DYN (Position-Independent Executable file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0x1e0fd0
  Start of program headers:          64 (bytes into file)
  Start of section headers:          12005168 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           56 (bytes)
  Number of program headers:         13
  Size of section headers:           64 (bytes)
  Number of section headers:         34
  Section header string table index: 33
  ```

File: seastar.yaml [hunk 3]
```diff
@@ -0,0 +1,244 @@
...
+      tag: seastar-${{package.version}}
+      expected-commit: 8df8212e53577e1d8477a5c901457cd61d88afc7
+
+  - uses: cmake/configure
+    with:
+      opts: |
+        -DSeastar_TESTING=OFF \
+        -DPROTOBUF_SOURCE=SYSTEM \
+        -DBUILD_SHARED_LIBS=ON \
+        -DWITH_SSL=system
```
Dentrax: Should we pass `-DCMAKE_BUILD_TYPE=RelWithDebInfo` which sets `-O3`?

seastar.yaml
```diff
@@ -0,0 +1,244 @@
...
+  - name: ${{package.name}}-libs
+    description: "libraries for ${{package.name}}"
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/usr/lib
+          mv "${{targets.destdir}}"/usr/lib/lib*.so* "${{targets.subpkgdir}}"/usr/lib/
+    test:
+      pipeline:
+        - uses: test/tw/ldd-check
+
```
Dentrax: Upstream does provides docs `-DSeastar_DOCS=ON`, should we have a doc subpkg?

seastar.yaml
Dentrax: Let's explicitly set `-DCMAKE_CXX_STANDARD=20`, as upstream does support 20 and 23: https://github.com/scylladb/seastar/blob/26badcb14c1b8a03547d5b7fb24ab6b5f4c42299/doc/compatibility.md#language-standards

seastar.yaml
```diff
@@ -0,0 +1,244 @@
...
+    runtime:
+      - binutils
+      - boost-dev
+      - c-ares-dev
+      - fmt-dev
+      - gcc
+      - pkgconf
+      - protobuf-dev
+
+environment:
```
Dentrax: There are some build-time deps are missing here: https://github.com/scylladb/seastar/blob/26badcb14c1b8a03547d5b7fb24ab6b5f4c42299/.github/workflows/alpinelinux.yaml#L21C2-L51C23

- bsd-compat-headers
- crypto++-dev
- libpciaccess-dev
- libucontext-dev
- libunwind-dev
- openssl-dev
- py3-yaml
- samurai

(haven't vetted, worth to double-check if those are present already.)

seastar.yaml
```diff
@@ -0,0 +1,244 @@
...
+      repository: "https://github.com/scylladb/seastar.git"
+      tag: seastar-${{package.version}}
+      expected-commit: 8df8212e53577e1d8477a5c901457cd61d88afc7
+
+  - uses: cmake/configure
+    with:
+      opts: |
+        -DSeastar_TESTING=OFF \
+        -DPROTOBUF_SOURCE=SYSTEM \
+        -DBUILD_SHARED_LIBS=ON \
```
Dentrax: Upstream builds static by default whereas we're enabled shared deps, there was a linking issue or smth?

utieyin: I think the default build option for us is dynamic/shared. Also we had issues with `pc:protobuf` been linked to an older version of protobuf than what was in the system. This is why this approach was taken

seastar.yaml
utieyin: Am not sure those are explicitly needed. Lets keep them off for now

seastar.yaml
utieyin: Even though that build flag exists, the build doesn't actually produce any docs

seastar.yaml
utieyin: Same as below, I expect that building a dynamic lib is a better approach here as we don't have to pin versions.

seastar.yaml
```diff
@@ -0,0 +1,246 @@
...
+      expected-commit: 8df8212e53577e1d8477a5c901457cd61d88afc7
+
+  - uses: cmake/configure
+    with:
+      opts: |
+        -DSeastar_TESTING=OFF \
+        -DCMAKE_CXX_STANDARD=20 \
+        -DPROTOBUF_SOURCE=SYSTEM \
+        -DBUILD_SHARED_LIBS=ON \
+        -DWITH_SSL=system
```
jamonation: Where do these options come from? They don't appear to be used:

```
CMake Warning:
  Manually-specified variables were not used by the project:

    PROTOBUF_SOURCE
    WITH_SSL
```

seastar.yaml
```diff
@@ -0,0 +1,246 @@
...
+    with:
+      repository: "https://github.com/scylladb/seastar.git"
+      tag: seastar-${{package.version}}
+      expected-commit: 8df8212e53577e1d8477a5c901457cd61d88afc7
+
+  - uses: cmake/configure
+    with:
+      opts: |
+        -DSeastar_TESTING=OFF \
+        -DCMAKE_CXX_STANDARD=20 \
```
jamonation: I'd unset this, https://github.com/scylladb/seastar/tree/ea8fb869ebe80e155a2b9808c23f9c06db5d1f5c?tab=readme-ov-file#seastars-c-standard-c20-or-c23 notes that:
> The build defaults to the latest standard supported by your compiler

Setting to 20 would pin to an older version and make rolling forward harder.

seastar.yaml
```diff
@@ -56,10 +56,8 @@ pipeline:
     with:
       opts: |
         -DSeastar_TESTING=OFF \
-        -DCMAKE_CXX_STANDARD=20 \
-        -DPROTOBUF_SOURCE=SYSTEM \
         -DBUILD_SHARED_LIBS=ON \
-        -DWITH_SSL=system
+        -DBUILD_STATIC_LIBS=ON
```
jamonation: Why the switch to static build? The other options didn't do anything, but this one ensures using system libraries where available.

utieyin: Upstream builds with static as default, so testing this to see if we can build both and split that static libs out.

utieyin: I tried that and the static package was empty when using `split/static` so removed.

seastar.yaml
```diff
@@ -0,0 +1,243 @@
...
+      - xfsprogs-dev
+      - yaml-cpp-dev
+
+vars:
+  py-version: 3.13
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: "https://github.com/scylladb/seastar.git"
```
EyeCantCU: Seastar for Redpanda has a public repository here: https://github.com/redpanda-data/seastar

Could we package `redpanda-seastar` instead?

EyeCantCU (dismissed):

seastar.yaml
utieyin: @EyeCantCU It doesn't look like they cut new releases but continue to work on branches and merge into master. We ok with building master on this repo? 
Also should this be in enterprise against wolfi?
Finally, do we still package the other open source seastar as just seastar?

redpanda-seastar.yaml
```diff
@@ -0,0 +1,240 @@
...
+        # Try to compile the test program
+        g++ -std=c++20 test_json2code.cc -o test_json2code || echo "Compilation failed, but continuing test"
+
+        # If compilation succeeded, run the program
+        if [ -x test_json2code ]; then
+          ./test_json2code
+        fi
+    - uses: test/pkgconf
+
+update:
```
EyeCantCU, line 238: We need to use the git poller so we continue to get updates

redpanda-seastar.yaml
```diff
@@ -0,0 +1,240 @@
...
+
+  - name: ${{package.name}}-libs
+    description: "libraries for ${{package.name}}"
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/usr/lib
+          mv "${{targets.destdir}}"/usr/lib/lib*.so* "${{targets.subpkgdir}}"/usr/lib/
+    test:
+      pipeline:
+        - uses: test/tw/ldd-check
```
EyeCantCU, line 88: Unsure if splitting these away provides value. What's left in the seastar package after we do this?

redpanda-seastar.yaml
```diff
@@ -0,0 +1,240 @@
...
+
+  - name: ${{package.name}}-libs
+    description: "libraries for ${{package.name}}"
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/usr/lib
+          mv "${{targets.destdir}}"/usr/lib/lib*.so* "${{targets.subpkgdir}}"/usr/lib/
```
EyeCantCU, line 85:
```suggestion
          mv "${{targets.destdir}}"/usr/lib/lib*.so.* "${{targets.subpkgdir}}"/usr/lib/
```

EyeCantCU (changes requested):

redpanda-seastar.yaml
utieyin: There is a binary `seastar-jsoncode.py`in `/us/rbin`

redpanda-seastar.yaml
utieyin: We would be polling every change and potential breaking ones as well if we poll `HEAD` right?

redpanda-seastar.yaml
utieyin: This doesn't work because there are no libraries like this

redpanda-seastar.yaml
wojciechka: @EyeCantCU we talked about it during working session.

The `seastar-jsoncode.py` should work without the libraries, so the main package does not need the dependency on `libs` subpackage.

And the libraries are a subpackage that is meant to be used by `redpanda`. This way `redpanda` could ship without shipping the `seastar-*` into `/usr/bin`.

The seastar `-dev` package is used by `redpanda` package at compile time.

IMHO splitting here seems to make sense.

utieyin:
no longer needed

PR #55039, opened 2025-06-02
fix(protobuf grpc compatibility): Rebuilding all the grps with the new version of protobuf

debasishbsws:
There is a new version of protobuf available(https://github.com/wolfi-dev/os/pull/54579) we need to rebuild all the grpc with this new version this required otherwise we will get a conflict to use both in a single build env

And some lint.


OddBloke (approved):

debasishbsws:
CVE is scan is not running

PR #55047, opened 2025-06-02
glibc: drop most per-locale packages, add -locale-extra

sil2100:
Generating locale subpackages per language causes the glibc package count to go up to 211 for just the locale ones. This is unnecessary and causes issues with the apk copier everytime a new release is performed. We actually only use like 5 main locales + posix. Create an locale-extra package for the rest.

### DIFF
--- a/glibc.yaml
+++ b/glibc.yaml
@@ -2,7 +2,7 @@ package:
   name: glibc
   version: "2.41"
   # Every glibc update causes build disruption; always announce on #eng-psa
-  epoch: 4
+  epoch: 5
   description: "the GNU C library"
   copyright:
     - license: LGPL-2.1-or-later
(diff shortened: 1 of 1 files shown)


glibc.yaml
```diff
@@ -774,6 +572,47 @@ subpackages:
...
+            - posix-libc-utils
+      pipeline:
+        - runs: |
+            set +x
+            pkg=glibc-locale-extra
+            out=$(locale --all-locales) || {
+              echo "FAIL: locale --all-locales failed for $pkg"
+              exit 1
+            }
+            # XXX: the list of locales to check here needs to be kept in sync with
```
javacruft, line 604: I prefer NOTE as this tends to get highlighted more consistently in IDE's

glibc.yaml
```diff
@@ -146,217 +146,15 @@ pipeline:
   - uses: strip
 
 data:
-  - name: locales
```
javacruft, line 149: confirmed list matches those current used in images :+1:

PR #55051, opened 2025-06-02
Allow tshark to run as nonroot

max-allan-cgr:
Allows tshark to run as nonroot

Needs to be a subpackage so that the normal package can still run as root without needing to give capabilities to the container.


tshark-4.4.yaml
```diff
@@ -43,6 +44,26 @@ pipeline:
...
+          # Allow tshark to run as nonroot user
+          setcap cap_net_raw,cap_net_admin=eip "${{targets.contextdir}}/usr/bin/tshark"
+          setcap cap_net_raw,cap_net_admin=eip "${{targets.contextdir}}/usr/bin/dumpcap"
+    test:
+      environment:
+        contents:
+          packages:
+            - libcap-utils
+      pipeline:
+        - uses: test/tw/ldd-check
```
kranurag7, line 66: inside some other packages where we're setting file capabilities we do the following tests.
```bash
        getcap /usr/bin/haproxy | cut -d ' ' -f2 | grep -q -E '^cap_net_bind_service=eip$'
```
would be a great addition here as well to make sure we don't miss file capabilities at package level.

jamonation: cap_net_admin encapsulates binding, and I'd expect tshark to have complete control over an interface

jamonation (dismissed):

jamonation (approved):

jamonation (approved):

PR #55100, opened 2025-06-02
glibc: fix shellcheck errors

dannf:
In glibclc2vpwo_ line 4:
	for locale in "${locales}"; do
                      ^----------^ SC2066 (error): Since you double quoted this, it will not word split, and the loop will only run once.

In glibc1ae0s0em line 1:
[ -e /lib64/ld-linux-* ]
     ^---------------^ SC2144 (error): -e doesn't work with globs. Use a for loop.

For more information:
  https://www.shellcheck.net/wiki/SC2066 -- Since you double quoted this, it ...
  https://www.shellcheck.net/wiki/SC2144 -- -e doesn't work with globs. Use a...

### DIFF
--- a/glibc.yaml
+++ b/glibc.yaml
@@ -2,7 +2,7 @@ package:
   name: glibc
   version: "2.41"
   # Every glibc update causes build disruption; always announce on #eng-psa
-  epoch: 4
+  epoch: 5
   description: "the GNU C library"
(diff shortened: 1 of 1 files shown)


dannf:
Superseded by https://github.com/wolfi-dev/os/pull/54778

PR #55105, opened 2025-06-02
Fix(conflict) current version is depends on a version grpc that has a…

debasishbsws:
… stict dependency on older protobuf rebuildig this so that it can pickup the latest grpc which depnds on the latest protobuf

### DIFF
--- a/ingress-nginx-controller-1.12.yaml
+++ b/ingress-nginx-controller-1.12.yaml
@@ -3,7 +3,7 @@ package:
   name: ingress-nginx-controller-1.12
   version: 1.12.2
   # There are manual changes to review between each package update. See 'vars:' section
-  epoch: 1
+  epoch: 2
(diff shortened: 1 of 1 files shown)


imjasonh (dismissed):

debasishbsws:
Not sure avoiding that error is a good idea or not.

kwmonroe:
> Not sure avoiding that error is a good idea or not.

fwiw, nginx/unit did [something similar](https://github.com/nginx/unit/pull/1548) to disable the gcc 15 warning while working on a compatible [refactor](https://github.com/nginx/unit/pull/1575).

For the current protobuf errors, it looks like we need to bring opentelemetry-cpp-dev's protobuf up from 5.29.4 to jive with protobuf-dev's 5.29.5.

imjasonh (approved):

PR #55133, opened 2025-06-02
feat: add perl-app-cpm, perl-app-cpanminus and all their dependencies

tulilirockz:
These should all have tests, except for the following packages:

`perl-darwin-initobjc` : For full functionality testing we need to have a darwin runner
`perl-menlo{,-legacy}` and `perl-parse-pmfile`: Upstream documentation is incomplete and requires some context from within `perl-app-cpanminus` to work

### DIFF
--- a/perl-app-cpanminus.yaml
+++ b/perl-app-cpanminus.yaml
@@ -31,6 +31,7 @@ update:
   enabled: true
   github:
     identifier: miyagawa/cpanminus
+    use-tag: true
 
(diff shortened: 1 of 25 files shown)


mritunjaysharma394 (dismissed):
Great work, where possible, as discussed async, let's use `git-checkout` pipeline and use `github` for update block, thanks for the great work!

jamonation (approved):

PR #55137, opened 2025-06-02
chore(Makefile): Always generate libraries token

EyeCantCU:
Always generate token for libraries access when chainctl is present

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,7 +107,10 @@ ${CACHEDIR}/.libraries_token.txt: cache
 	mv $${tmpf} ${CACHEDIR}/.libraries_token.txt
 
 .PHONY: lib-token
-lib-token: ${CACHEDIR}/.libraries_token.txt
+lib-token:
+	if command -v chainctl; then \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -107,7 +107,10 @@ ${CACHEDIR}/.libraries_token.txt: cache
 	mv $${tmpf} ${CACHEDIR}/.libraries_token.txt
 
 .PHONY: lib-token
-lib-token: ${CACHEDIR}/.libraries_token.txt
+lib-token:
+	if command -v chainctl; then \
+		$(MAKE) ${CACHEDIR}/.libraries_token.txt; \
```
dannf, line 112: Is there a way to glean from a `melange` YAML file that library access is required? If so, maybe the better solution here would be to evaluate the YAML file and only generate a token in that case (and fail hard if a token is needed and `chainctl` is unavailable)

EyeCantCU: We could parse the package manifest with yq and look for any pipeline preceded by `auth`

dannf: That sounds good to me - both for usability and security reasons.

PR #55166, opened 2025-06-03
Zookeeper-3.9: Official package and versioning fix + Entrypoint subpackage addition

AnmolVirdi:
- Fixed versioning pattern for `zookeeper`
- Fixed packaging for official `zookeeper`
- Added `-oci-entrypoint` package
- Fixed `bitnami compat` to align with new `zookeeper` packaging
- Updated CVE patches


zookeeper-3.9.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: zookeeper-3.9
-  version: "3.9.4.0"
-  epoch: 1
+  version: "3.9.3"
```
kranurag7, line 3: this looks like a downgrade here and even if we build it; the resolver will continue to pull 3.9.4.0 package version. Is this correct that this is latest and we are considering this an upgrade?

AnmolVirdi: @kranurag7 this is an upgrade since `3.9.4-0` is just a release candidate and not official release. The goal is to release `3.9.3` and then while creating zookeeper official image, we'll keep `3.9.4` to ignore version

kranurag7: the tooling doesn't understanding this; it'll continue to treat 3.9.4.0 one as the latest version; we need to withdraw that package first.

and then inside update block please use `ignore-regex-patterns:` to ignore versions like 1.2.3.4

AnmolVirdi: Withdrawing that package will lead to breaking `zookeeper-bitnami:3.9.4` which was accidentally released.

To ignore release candidates, I've added:
 `tag-filter: ^release-3\.9\.\d+$`
 Release candidates follow: 1.2.3-4 pattern

AnmolVirdi: Not being able to release `3.9.3` is an issue, I'll open a thread on how we can sort this out. Dropping `3.9.4-0` package means we will be dropping `3.9.4-0` tagged images as well.

AnmolVirdi:
CVE Advisory update PR:
- https://github.com/wolfi-dev/advisories/pull/19077

zookeeper-3.9.yaml
```diff
@@ -45,35 +37,69 @@ pipeline:
...
+      provides:
+        - zookeeper-oci-entrypoint=${{package.full-version}}
+      runtime:
+        - ca-certificates-bundle
+        - bash
+        - coreutils
+        - glibc-locale-en
+        - merged-usrsbin
+        - netcat-openbsd
+        - wolfi-baselayout
```
kranurag7:
```suggestion
        - netcat-openbsd
```

zookeeper-3.9.yaml
```diff
@@ -91,28 +117,58 @@ subpackages:
...
           cp -r conf/* ${{targets.subpkgdir}}/opt/bitnami/zookeeper/conf.default
 
           # create symlinks for both /entrypoint.sh and /run.sh to make it compatible with bitnami/zookeeper helm chart
           ln -sf /opt/bitnami/scripts/zookeeper/entrypoint.sh "${{targets.subpkgdir}}"/entrypoint.sh
           ln -sf /opt/bitnami/scripts/zookeeper/run.sh "${{targets.subpkgdir}}"/run.sh
     test:
+      environment:
+        contents:
+          packages:
+            - uuidgen
```
kranurag7: is uuidgen required by the script? if yes; we should add it as runtime dep. If no, we can consider removing this.

AnmolVirdi: It's used in the tests to generate uuid while creating a node

AnmolVirdi:
Closing this PR since we're fixing Zookeeper-3.9.4 versions onwards. There will be a follow-up PR.

PR #55171, opened 2025-06-03
curl: fix multi-transfer handling

javacruft:
Cherry pick fix to resolve resizing of multi-transfer handling table,
resolving issues with cargo downloads for rust based packages.


javacruft:
Successful build test with zed that was reliably reproducing the issue this causes.

curl.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: curl
   version: "8.14.0"
-  epoch: 2
+  epoch: 3
```
xnox, line 4: do we need this cherrypick in the rustls curl build too?

javacruft: We do!

curl.yaml
```diff
@@ -15,30 +15,39 @@ package:
 environment:
   contents:
     packages:
```
xnox, line 17: epoch bump

xnox:
i am curious this resolved the build.

i am pondering if we are in a position to merge curl and curl-rustls into a single build.

javacruft:
> i am pondering if we are in a position to merge curl and curl-rustls into a single build.

We can look (later)

javacruft:
I did move both curl packages to use git-checkout rather than one using fetch as well - made it easier to cherry pick the fix.

sil2100 (approved):
I'm always a bit weary of switching from a fetch to a checkout requiring autoreconf runs - but since it builds and is tested to work, let's go.

justinvreeland:
@xnox when you say in a single build do you mean building both libraries separately or build them libcurl as a single multi lib backend?  It's fairly easy to build curl/libcurl to have multiple backends i did it for some experimentation early in the year.

```
f2faf2b22534:/work/packages# curl -V
curl 8.12.1 (aarch64-unknown-linux-gnu) libcurl/8.12.1 OpenSSL/3.4.1 (rustls-ffi/0.15.0/rustls/0.23.25/aws-lc-rs) zlib/1.3.1 brotli/1.1.0 libpsl/0.21.5 nghttp2/1.65.0 OpenLDAP/2.6.9
Release-Date: 2025-02-13
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns ldap ldaps mqtt pop3 pop3s rtsp smtp smtps telnet tftp ws wss
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IPv6 Kerberos Largefile libz MultiSSL PSL SPNEGO SSL threadsafe TLS-SRP UnixSockets
```

xnox:
> @xnox when you say in a single build do you mean building both libraries separately or build them libcurl as a single multi lib backend?  It's fairly easy to build curl/libcurl to have multiple backends i did it for some experimentation early in the year.
> 
> ```
> f2faf2b22534:/work/packages# curl -V
> curl 8.12.1 (aarch64-unknown-linux-gnu) libcurl/8.12.1 OpenSSL/3.4.1 (rustls-ffi/0.15.0/rustls/0.23.25/aws-lc-rs) zlib/1.3.1 brotli/1.1.0 libpsl/0.21.5 nghttp2/1.65.0 OpenLDAP/2.6.9
> Release-Date: 2025-02-13
> Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns ldap ldaps mqtt pop3 pop3s rtsp smtp smtps telnet tftp ws wss
> Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IPv6 Kerberos Largefile libz MultiSSL PSL SPNEGO SSL threadsafe TLS-SRP UnixSockets
> ```

I meant to have sub package that redoes the build. And that's how I have always built curl before.

I haven't seen multi backend curl before - can we still split package it?

Also still not sure if this is a good idea. As sometimes rust, ffi, rustls lag and do not support latest curl.

justinvreeland:
> I haven't seen multi backend curl before - can we still split package it?

I don't believe so. It creates a version of libcurl that response to `CURL_SSL_BACKEND` https://curl.se/libcurl/c/libcurl-env.html so basically the opposite of a split package with two ssl backends.

xnox:
> > I haven't seen multi backend curl before - can we still split package it?
> 
> I don't believe so. It creates a version of libcurl that response to `CURL_SSL_BACKEND` https://curl.se/libcurl/c/libcurl-env.html so basically the opposite of a split package with two ssl backends.

ew, no we don't want that. or possibly we want multi-backend libcurl as an additional build to single-backend libcurls.

justinvreeland:
> ew, no we don't want that. or possibly we want multi-backend libcurl as an additional build to single-backend libcurls.

Honestly i think we want it if only for our curl image. right now our default install of curl is openssl, but our curl image is with curl-rustls. If i had other images and downloaded our curl image but it was rust instead of the openssl version in my other images i'd be a bit surprised I couldn't use it to test if the problem was curl. MultiSSL is a neat solution to that.

xnox:
@justinvreeland many cURL CVEs affect specific backends only; thus having them in separate origins helps with false positive determinations.

PR #55175, opened 2025-06-03
bunch of package split manpages/infodir

AbduAzim424:
Based on this issue (https://github.com/chainguard-dev/internal-dev/issues/11399)

Fixes:
'split/manpages', 'split/infodir' added to the appropriate packages


libpcap.yaml
```diff
@@ -63,6 +63,14 @@ subpackages:
         - uses: test/pkgconf
         - uses: test/tw/ldd-check
 
+  - name: libcap-doc
```
murraybd: This needs to be libpcap not libcap.

libpcap.yaml
```diff
@@ -63,6 +63,14 @@ subpackages:
...
+    description: libcap docs
```
murraybd: This should also be libpcap.

murraybd (changes requested):
libpcap.yaml needs updating

PR #55221, opened 2025-06-03
luajit: bump to latest

Dentrax:
Bump to latest: https://github.com/openresty/luajit2/releases/tag/v2.1-20250117

### DIFF
--- a/luajit.yaml
+++ b/luajit.yaml
@@ -1,7 +1,7 @@
 package:
   name: luajit
-  version: 2.1_p20210510
-  epoch: 12
+  version: 2.1_git20250117
+  epoch: 0
   description: OpenResty's branch of LuaJIT
   copyright:
     - license: MIT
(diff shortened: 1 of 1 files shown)


luajit.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: luajit
-  version: 2.1_p20210510
-  epoch: 12
+  version: 2.1_git20250117
```
OddBloke: I think we'll need to retain the `p` here, as it sorts higher than `git`:

```
$ docker run -it cgr.dev/chainguard/wolfi-base apk version -t 2.1_p20210510 2.1_git20250117
>
```

OddBloke (changes requested):
The version here will need changing: see inline comment. I also wonder if we might now be able to enable updates, as we're now using git tags?

OddBloke (approved):

PR #55268, opened 2025-06-04
quality: extend test kube-metrics-adapter package 

yashpimple:
This PR adds a functional test to validate core behavior of kube-metrics-adapter package

### DIFF
--- a/kube-metrics-adapter.yaml
+++ b/kube-metrics-adapter.yaml
@@ -65,7 +65,6 @@ test:
           # Check if the status code is 200, if so exit 0, otherwise exit 1
           if [ "$STATUS" -eq 200 ]; then
               echo "Health check passed with status code 200"
-              exit 0
           else
               echo "Health check failed with status code $STATUS"
               exit 1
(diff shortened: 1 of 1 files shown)


kube-metrics-adapter.yaml
```diff
@@ -37,7 +37,37 @@ update:
...
+        post: |
+          STATUS=$(curl -k -s -o /dev/null -w "%{http_code}" https://localhost:8443/healthz)
+          # Check if the status code is 200, if so exit 0, otherwise exit 1
+          if [ "$STATUS" -eq 200 ]; then
+              echo "Health check passed with status code 200"
+              exit 0
+          else
+              echo "Health check failed with status code $STATUS"
+              exit 1
+          fi
```
tulilirockz: Doesn't this mean that itll exit always? - if not equal to 200, exit 1, if eq 200 exit 0 - but what about the last curl part?

yashpimple: @tulilirockz i forgot to remove the exit 0 it's fixed now so the `/metrics` check runs properly. Thanks for pointing it out 🙌🏻

mritunjaysharma394 (approved):
thanks, lgtm

PR #55311, opened 2025-06-04
elfutils: New packages `debuginfod` and `libdebuginfod{,-dev}`

sergiodj:
Add new `debuginfod` and `libdebuginfod{,-dev}` packages.

Closes: https://github.com/chainguard-dev/internal-dev/issues/12757

### DIFF
--- a/elfutils.yaml
+++ b/elfutils.yaml
@@ -2,7 +2,7 @@
 package:
   name: elfutils
   version: "0.193"
-  epoch: 0
+  epoch: 1
   description: Utilities and DSOs to handle ELF files and DWARF data
   copyright:
     - license: GPL-3.0-or-later AND ( GPL-2.0-or-later OR LGPL-3.0-or-later )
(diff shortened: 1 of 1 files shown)


wiedenmeier (approved):
Fantastic, thanks for building this. Once these packages are available in the repo I will rebuild the python package discussed here https://github.com/chainguard-dev/internal-dev/issues/12736 which depends on this library and confirm that the build succeeds with `libdebuginfod-dev` specified as a build requirement

sergiodj:
That's great, thanks @wiedenmeier

justinvreeland:
Have you verified this builds our other core packages correctly? I had thought there was a bootstrapping problem with libdebuginfod like: https://sourceware.org/bugzilla/show_bug.cgi?id=25509

That might only be an actually bootstrapping issue maybe not relevant to adding it but worth noting maybe. Either way code looks good to me.

justinvreeland (approved):

PR #55333, opened 2025-06-04
empty package fix 1

AbduAzim424:
The issue with these 3 packages was that the man pages weren't located in standard directories, preventing "split/manpages" from functioning correctly, and the fix involved relocating them to the appropriate standard paths to enable proper division. (not an issue anymore, melange fixes these type of issues from now on)


fatrace.yaml
```diff
@@ -28,11 +28,19 @@ pipeline:
 
   - uses: strip
 
+  - name: install manpages
+    runs: |
+      mkdir -p "${{targets.destdir}}/usr/share/man/man8"
+      cp fatrace.8 "${{targets.destdir}}/usr/share/man/man8/"
```
murraybd: I believe you'll want to us `mv` instead of `cp` otherwise the `fatrace.8` file will exist in both packages. You could test this via local build and inspecting the resultant apks.

AbduAzim424: If I'm not mistaking this step copies man page from upstream folder into the package folder that is later actually moved by "split/manpages"

murraybd (changes requested):
Rather than creating a complicated `install manpages` step in the build pipeline we should modify melange such that it also splits manpages out of `/usr/local/share/man`.

murraybd (changes requested):
bat.yaml will need an epoch bump for the doc package to be generated.

murraybd:
It looks like meson is failing to build on arm64.

AbduAzim424:
it says "invalid memory address or nil pointer dereference" I think this is related to melange rather than meson itself?

murraybd (approved):

PR #55345, opened 2025-06-04
empty package fix 4

AbduAzim424:
Fixed wrong pipeline entry and non-standard manpage placement witin the upstream repository.


murraybd (changes requested):
You'll want to use mv instead of cp so that the man files will no longer be in the non -doc package.

AbduAzim424:
I double checked udns and there are no manpages within the main package. The "cp" in the pipeline definition is used to copy manpages from upstream directory into main package directory and later on udns-doc subpackage stage "split/manpages" script actually 'mv' the copied docs into -doc subpackage. The problem with fatrace was different because I made a mistake in determinig the cause of the problem and applied the wrong solution but that package is fixed as well now.

murraybd (approved):

PR #55357, opened 2025-06-05
"-doc" package fix

AbduAzim424:
Fixes:
'split/manpages', 'split/infodir' added to the appropriate packages


ajayk:
Can you please add proper PR description

AbduAzim424:
Yea sure, I'm just splitting larger PR into smaller ones.

murraybd (dismissed):

murraybd:
Based on the log output it looks like the gdal-doc package is empty.

AbduAzim424:
already looking at it

gdal.yaml
```diff
@@ -229,6 +229,23 @@ subpackages:
             sozip --version | grep ${{package.version}}
             sozip --help
 
+  - name: gdal-doc
+    description: gdal docs
+    pipeline:
+      - runs: |
+          for pyver in 3.10 3.11 3.12 3.13; do
```
murraybd: Hard coding these numbers here makes me nervous as it would require manual updating. Did you look at using the `py-versions` range at all?

AbduAzim424: I wasn't able to pull that of. I tried using range within pipeline but that failed as it is not allowed.

AbduAzim424: Let me try something

AbduAzim424: Ok that didn't work, I do not really know how to solve this without hardcoding

AbduAzim424: I can do separate range enttires but that will spawn separate "-doc" packages for each gdal-py version which I think isn't ideal.

murraybd: Actually, thinking about it more I believe we would want separate documenation packages for each gdal-py version because users would only want to install the documentation for the version they are using and it will match the way the -dev packages are organized.

AbduAzim424: Understood, will do shortly

murraybd (dismissed):

murraybd (approved):

PR #55374, opened 2025-06-05
Fix(update): Use Gitlab repo sourcecode and our toolchain to build

debasishbsws:
- We can use our GitMonitor now which is much accurate that release monitor
- Even the tar and the gitlab source is not an exact match they result in the same artifact

Fix: https://github.com/chainguard-dev/internal-dev/issues/12692

using this will result in less tickets of update bot failures in future.

### DIFF
--- a/cryptsetup.yaml
+++ b/cryptsetup.yaml
@@ -2,7 +2,7 @@
 package:
   name: cryptsetup
   version: 2.7.5
-  epoch: 21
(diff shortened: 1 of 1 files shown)


debasishbsws:
The test passes in Docker runner but failed in QEMU. for this and the for what we already have today.

dnegreira (approved):

PR #55465, opened 2025-06-06
kubernetes-1.33: make kubelet-1.33 subpackage provide kubelet

markusboehme:
Add a `provides` for the `kubelet-1.33` subpackage so that it may be installed as `kubelet`.

### DIFF
--- a/kubernetes-1.33.yaml
+++ b/kubernetes-1.33.yaml
@@ -1,7 +1,7 @@
 package:
   name: kubernetes-1.33
   version: "1.33.1"
-  epoch: 3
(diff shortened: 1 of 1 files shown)


a-crate (approved):

zmarano (dismissed):

a-crate (changes requested):
I think this needs to be rebased, it's not building packages in CI since epoch is 3 at HEAD in main now.

markusboehme:
@a-crate Good catch! Rebased.

a-crate (approved):

PR #55466, opened 2025-06-06
fix(CVE): [jenkins-plugin-manager] Update commons-beanutils transitive dependency to   1.11.0

debasishbsws:
-address GHSA-wxr5-93ph-8wr9

- It was marked pending upstream fix but even it is a transitive dependency it is fixable by forcing the version

- There are total 3 pom.xml file and we need to apply the patch for two of them


debasishbsws:
Cant use the new login module as there are some missing Class in the new version

PR #55511, opened 2025-06-06
nginx-s3-gateway: init package at 0_git20250605

tulilirockz:


nginx-s3-gateway.yaml
```diff
@@ -0,0 +1,92 @@
...
+  epoch: 0
+  description: NGINX S3 Gateway
+  copyright:
+    - license: Apache-2.0
+  # BYO nginx
+  dependencies:
+    provides:
+      - nginx-config=0.${{package.full-version}}
+    runtime:
+      - njs
```
tulilirockz, line 14: They also explicitly add `libedit` on the dockerfile...

nginx-s3-gateway.yaml
```diff
@@ -0,0 +1,128 @@
...
+  name: nginx-s3-gateway
+  version: 0_git20250605
+  epoch: 0
+  description: NGINX S3 Gateway
+  copyright:
+    - license: Apache-2.0
+  # BYO nginx
+  dependencies:
+    provides:
+      - nginx-config=0.${{package.full-version}}
```
kranurag7: what's the reason behind having this provides here?

tulilirockz: Theres most likely no reason to do this, but I couldn't manage to install `nginx-*` without it, it fixes a conflict I was getting where `/etc/nginx/nginx.conf` exists in both this package and `nginx-config`. Unsure what would be the actual right way to do this

kranurag7: I think replaces will be a more suited solution here. 

```yaml
package:
  dependencies:
    replaces:
      - nginx-mainline-config
```

tulilirockz: When doing the `replaces` it seems that `/etc/nginx/nginx.conf` gets overwritten by `nginx-*-config` instead of having the same things from upstream - `provides` correctly omits installing `nginx-config`... @jamonation had mentioned doing something like the following would work but it still brings in `nginx-*-config`
```yaml
  options:
    no-depends: true
    no-provides: true
```

tulilirockz: After testing this for a bit, seems that the `options:` thing works fine for this! - checksum is the same before and after installing `nginx-mainline`:
```shell
11c3ecf01365:/work/packages# apk add --allow-untrusted /work/packages/aarch64/nginx-s3*.apk
# ...
11c3ecf01365:/work/packages# sha256sum /etc/nginx/nginx.conf
881eaf6990c91d119e49032ebba9d0a0351a5edc801c4faaf4a2d640667ee9ee  /etc/nginx/nginx.conf
11c3ecf01365:/work/packages# apk add nginx-config
(1/2) Installing nginx-mainline (1.28.0-r0)
(2/2) Installing nginx-mainline-config (1.28.0-r0)
Executing busybox-1.37.0-r41.trigger
OK: 180 MiB in 90 packages
11c3ecf01365:/work/packages# sha256sum /etc/nginx/nginx.conf
881eaf6990c91d119e49032ebba9d0a0351a5edc801c4faaf4a2d640667ee9ee  /etc/nginx/nginx.conf
```

kranurag7: we use replaces when we want to add two packages together (either at image or via runtime deps) and some files are conflicting. 

for example here: nginx-s3-gateway ships it's own [nginx.conf](https://apk.chaindag.dev/https/apk.cgr.dev/wolfi-presubmit/ce350646b1dea1c24b73e8d2be02bc661e6e940d/x86_64/nginx-s3-gateway-0_git20250605-r0.apk@etag:ab4ba2537437947f91f9e1c80a325cf5/etc/nginx/nginx.conf) file which is different from nginx-mainline-config [nginx.conf](https://apk.dag.dev/https/packages.wolfi.dev/os/x86_64/nginx-mainline-config-1.28.0-r0.apk@sha1:60ce25faaedf92712afc564e8d6d6616d9ef8659/etc/nginx/nginx.conf) 

Given, we want to include the `/etc/nginx/nginx.conf` from the upstream here; we'll use replaces here which will replace the conflicting file `/etc/nginx/nginx.conf` from nginx-mainline-config. 

[SCA](https://github.com/chainguard-dev/melange/blob/main/pkg/sca/sca.go) is not doing anything [important here](https://apk.chaindag.dev/https/apk.cgr.dev/wolfi-presubmit/ce350646b1dea1c24b73e8d2be02bc661e6e940d/x86_64/nginx-s3-gateway-0_git20250605-r0.apk@etag:ab4ba2537437947f91f9e1c80a325cf5/.PKGINFO) so we can avoid setting no-depends and no-provides to true. We use no-depends and no-provides when sca is acting weird (detecting runtime deps for which source is not available `no-depends` use-case; or in case where you don't want to provide `cmd:foo` `no-provides` use-case ).

nginx-s3-gateway.yaml
```diff
@@ -0,0 +1,127 @@
...
+    period: daily
+    reason: Upstream does not maintain tags or releases
+  github:
+    identifier: nginx/nginx-s3-gateway
+
+test:
+  environment:
+    contents:
+      packages:
+        - nginx-stable
```
tulilirockz, line 99: This fails on `nginx-mainline` with:
```
2025/06/07 17:33:58 WARN nginx: [emerg] module "/var/lib/nginx/modules/ngx_http_js_module.so" version 1027005 instead of 1028000 in /etc/nginx/nginx.conf:8
```
Packages:
```
nginx-mainline
nginx-mainline-config
nginx-mainline-mod-http_geoip
nginx-mainline-mod-http_image_filter
nginx-mainline-mod-http_xslt_filter
nginx-s3-gateway
```

kranurag7: nginx-stable looks good here.

kranurag7 (approved):

jamonation (approved):

PR #55580, opened 2025-06-08
chore: Streamline auth flow

EyeCantCU:
* Automate token generation
* Create tokens only when needed
* Remove when finished with them
* Leverage octo-sts

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,8 +111,8 @@ clean-cache:
 	mkdir -p $(pkgname) && \
 	mv $${tmpf} ./$(pkgname)/.guarded-repo.token
 
-guarded_repo_token_targets = $(foreach name,$(pkgs),repo-token/$(name))
-$(guarded_repo_token_targets): repo-token/%: %/.guarded-repo.token
+repo_token_targets = $(foreach name,$(pkgs),repo-token/$(name))
+$(repo_token_targets): repo-token/%: %/.guarded-repo.token
 
 %/.libraries.token:
 	$(eval pkgname := $*)
@@ -124,11 +124,11 @@ $(guarded_repo_token_targets): repo-token/%: %/.guarded-repo.token
 libraries_token_targets = $(foreach name,$(pkgs),lib-token/$(name))
 $(libraries_token_targets): lib-token/%: %/.libraries.token
 
-.PHONY: tokens/%
+.PHONY: tokens-if-needed/%
 tokens/%:
 	$(eval pkgname := $*)
-	@$(call guarded_repo_token,$(pkgname))
-	@$(call guarded_libraries_token,$(pkgname))
+	@$(call repo_token_if_needed,$(pkgname))
+	@$(call libraries_token_if_needed,$(pkgname))
 
 .PHONY: clean-tokens/%
 clean-tokens/%:
@@ -175,7 +175,7 @@ $(pkg_targets): package/%:
 	$(MAKE) yamlfile=$(yamlfile) pkgname=$* packages/$(ARCH)/$(pkgver).apk
 
 packages/$(ARCH)/%.apk: cache $(KEY) $(QEMU_KERNEL_DEP)
-	$(MAKE) tokens/$(pkgname)
+	$(MAKE) tokens-if-needed/$(pkgname)
 	mkdir -p ./$(pkgname)/
 	$(eval SOURCE_DATE_EPOCH ?= $(shell git log -1 --pretty=%ct --follow $(yamlfile)))
 	$(info @SOURCE_DATE_EPOCH=$(SOURCE_DATE_EPOCH) $(MELANGE) build $(yamlfile) $(MELANGE_OPTS) --source-dir ./$(pkgname)/)
@@ -192,7 +192,7 @@ $(dbg_targets): debug/%: cache $(KEY) $(QEMU_KERNEL_DEP)
 	$(eval yamlfile := $*.yaml)
 	$(eval pkgname := $*)
 	$(eval pkgver := $(shell $(MELANGE) package-version $(yamlfile)))
-	$(MAKE) tokens/$(pkgname)
+	$(MAKE) tokens-if-needed/$(pkgname)
 	@printf "Building package $* with version $(pkgver) from file $(yamlfile)\n"
 	mkdir -p ./"$*"/
 	$(eval SOURCE_DATE_EPOCH ?= $(shell git log -1 --pretty=%ct --follow $(yamlfile)))
@@ -206,7 +206,7 @@ $(test_targets): test/%: cache $(KEY) $(QEMU_KERNEL_DEP)
 	$(eval yamlfile := $*.yaml)
 	$(eval pkgname := $*)
 	$(eval pkgver := $(shell $(MELANGE) package-version $(yamlfile)))
-	$(MAKE) tokens/$(pkgname)
+	$(MAKE) tokens-if-needed/$(pkgname)
 	@printf "Testing package $* with version $(pkgver) from file $(yamlfile)\n"
 	bash -xc "trap 'trap - SIGINT SIGTERM ERR; $(MAKE) clean-tokens/$(pkgname); exit 1' SIGINT SIGTERM ERR; $(MELANGE) test $(yamlfile) $(MELANGE_TEST_OPTS) --source-dir ./$(*)/"
 	$(MAKE) clean-tokens/$(pkgname)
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -100,14 +100,34 @@ clean:
 clean-cache:
 	rm -rf ${CACHEDIR}
 
-${CACHEDIR}/.libraries_token.txt: cache
+${CACHEDIR}/.guarded-repo.token: cache
+	tmpf=$(shell mktemp); \
+	chainctl auth octo-sts --identity=guarded-package-repos --scope=chainguard-dev > $${tmpf}; \
```
dannf: I think the `;` would ignore a non-zero exitcode from `chainctl`. That would result in an empty `$${tmpf}` that would then get used as the token file. Maybe use `&&` instead?

Makefile
```diff
@@ -100,14 +100,34 @@ clean:
...
+	mv $${tmpf} ${CACHEDIR}/.guarded-repo.token
+
+.PHONY: repo-token
+repo-token: ${CACHEDIR}/.guarded-repo.token
+
+${CACHEDIR}/.libraries.token: cache
 	tmpf=$(shell mktemp); \
 	chainctl auth login --audience libraries.cgr.dev; \
 	chainctl auth token --audience libraries.cgr.dev > $${tmpf}; \
-	mv $${tmpf} ${CACHEDIR}/.libraries_token.txt
```
dannf, line 107: while pre-existing, same comment about `;` vs `&&` here

Makefile
```diff
@@ -100,14 +100,34 @@ clean:
...
 .PHONY: lib-token
-lib-token: ${CACHEDIR}/.libraries_token.txt
+lib-token: ${CACHEDIR}/.libraries.token
+
+.PHONY: cache-tokens-if-needed/%
+cache-tokens-if-needed/%:
+	auth_needed=$$(yq '.. | select(has("uses")) | select(.uses | test("auth|iamguarded"))' $*); \
+	if [ -n "$$auth_needed" ]; then \
+		$(MAKE) repo-token; \
+		$(MAKE) lib-token; \
```
dannf: its not clear what the mapping between auth/iamguarded and repo-token/lib-token are.  If there's a 1:1 mapping, can we handle them separately so the relationship is clear? Or do both tokens need to be used by both pipelines?

Also, personally I'd probably just make this a couple of make functions, akin to `authget`.

Makefile
```diff
@@ -151,10 +171,12 @@ $(pkg_targets): package/%:
 	$(MAKE) yamlfile=$(yamlfile) pkgname=$* packages/$(ARCH)/$(pkgver).apk
 
 packages/$(ARCH)/%.apk: cache $(KEY) $(QEMU_KERNEL_DEP)
+	$(MAKE) cache-tokens-if-needed/$(yamlfile)
 	mkdir -p ./$(pkgname)/
 	$(eval SOURCE_DATE_EPOCH ?= $(shell git log -1 --pretty=%ct --follow $(yamlfile)))
 	$(info @SOURCE_DATE_EPOCH=$(SOURCE_DATE_EPOCH) $(MELANGE) build $(yamlfile) $(MELANGE_OPTS) --source-dir ./$(pkgname)/)
 	SOURCE_DATE_EPOCH=$(SOURCE_DATE_EPOCH) $(MELANGE) build $(yamlfile) $(MELANGE_OPTS) --source-dir ./$(pkgname)/
+	$(MAKE) clean-tokens
```
dannf: I think this would be racy if someone did a `make -j3 package/foo package/bar package/baz` - that is, the token would be changing or getting deleted out from under the other builds.

This makes me want to go back to the idea of putting the token in the package build directory instead of the cache. Each package would then have its own tokens. They could live for just that package build, and other builds that happen to run at the same time wouldn't get a copy of the tokens from another build. WDYT?

dannf (dismissed):
thanks! this is looking much more secure

Makefile
EyeCantCU: I think the problem we have is with more fine grained filtering, incorporating auth into other pipelines (like the iamguarded pipelines) becomes painful. We have to remember to come back to this and remember to add an exception

I can make this more fine grained and isolated between both use cases. It is something we might want to take another pass at though

If there was a way to "unroll" top level pipelines used in packages, that would make this a little more battle ready

Makefile
EyeCantCU: Didn't catch this, but you're right! We should do the same for guarded repos and libraries. I'll make the change

Makefile
EyeCantCU: I like that approach better, will make the adjustment for this and library tokens

Makefile
EyeCantCU: Consolidated in a couple functions with a comment on the relationship between iamguarded and auth/guarded-repo

Makefile
```diff
@@ -100,14 +104,36 @@ clean:
...
+	$(eval pkgname := $*)
+	tmpf=$(shell mktemp) && \
+	chainctl auth token --audience libraries.cgr.dev > $${tmpf} && \
+	mkdir -p $(pkgname) && \
+	mv $${tmpf} ./$(pkgname)/.libraries.token
+
+libraries_token_targets = $(foreach name,$(pkgs),lib-token/$(name))
+$(libraries_token_targets): lib-token/%: %/.libraries.token
+
+.PHONY: tokens-if-needed/%
```
dannf, line 127: I don't see where this rule is defined, but maybe I'm just tired :/

Makefile
```diff
@@ -280,9 +316,42 @@ check-bootstrap:
...
+    echo "$1 does not use a guarded repo. Skipping token creation."; \
+  fi
+endef
+
+define libraries_token_if_needed
+  $(eval yamlfile := $1.yaml)
+  $(eval pkgname := $1)
+  $(eval pipelines := "auth")
+  $(eval ignore_pipelines := "auth/guarded-repo")
+  if yq e -e '.. | select(has("uses")) \
```
dannf, line 360: btw, is `yq` used before this in our Makefiles? If not, do you think the error mode for not having it installed is clear, or do we need an explicit message? (I haven't tried it)

Makefile
```diff
@@ -280,9 +316,42 @@ check-bootstrap:
...
+  tok=$$(chainctl auth token --audience=$(1)) || \
   { echo "failed token from $(1) for target $@"; exit 1; }; \
   mkdir -p $$(dirname $(2)) && \
   echo "auth-download[$(1)] to $(2) from $(3)" && \
   curl --fail -LS --silent -o $(2).tmp --user "user:$$tok" $(3) && \
-	mv "$(2).tmp" "$(2)"
+  mv "$(2).tmp" "$(2)"
+endef
+
+define repo_token_if_needed
```
dannf, line 338: These look great. So clear, so well formatted. Nice work!

Makefile
```diff
@@ -184,7 +184,7 @@ packages/$(ARCH)/%.apk: cache $(KEY) $(QEMU_KERNEL_DEP)
 	mkdir -p ./$(pkgname)/
 	$(eval SOURCE_DATE_EPOCH ?= $(shell git log -1 --pretty=%ct --follow $(yamlfile)))
 	$(info @SOURCE_DATE_EPOCH=$(SOURCE_DATE_EPOCH) $(MELANGE) build $(yamlfile) $(MELANGE_OPTS) --source-dir ./$(pkgname)/)
-	SOURCE_DATE_EPOCH=$(SOURCE_DATE_EPOCH) $(MELANGE) build $(yamlfile) $(MELANGE_OPTS) --source-dir ./$(pkgname)/
+	bash -xc "trap 'trap - SIGINT SIGTERM ERR; $(MAKE) clean-tokens/$(pkgname); exit 1' SIGINT SIGTERM ERR; SOURCE_DATE_EPOCH=$(SOURCE_DATE_EPOCH) $(MELANGE) build $(yamlfile) $(MELANGE_OPTS) --source-dir ./$(pkgname)/"
```
dannf, line 182: Maybe we should use `.ONESHELL` so you can do this as a normal function:
https://stackoverflow.com/questions/28597794/how-can-i-clean-up-after-an-error-in-a-makefile

I've never used it myself, but seems like a win.

Makefile
EyeCantCU: Fixed this, my bad

Makefile
EyeCantCU: We get a CNF and it fails. I could be more verbose here though, similar to qemu

Makefile
EyeCantCU: I'll give this a try :)

EyeCantCU: I still end up having to wrap the invokation in bash. I'll leave as is for now

dannf (approved):
:heart:

EyeCantCU:
lfg! Thank you Dann!

PR #55603, opened 2025-06-08
Add split/bin pipeline

dannf:
- **pipelines: add split/bin**
- **py3-bokeh: Use split/bin pipeline**

### DIFF
--- a/pipelines/split/bin.yaml
+++ b/pipelines/split/bin.yaml
@@ -1,0 +1,40 @@
+name: Split executable files
+
+needs:
+  packages:
+    - busybox
+
+inputs:
+  package:
+    description: |
+      The package to split executable files from
+    required: false
+
+pipeline:
+  - runs: |
+      pkgname="${{context.name}}"
+      pkgnobin="${pkgname%%-bin}"
+      pkgnobindir="${{targets.outdir}}/$pkgnobin"
+
+      if [ -n "${{inputs.package}}" ]; then
+        PACKAGE_DIR="${{targets.outdir}}/${{inputs.package}}"
+      elif [ "$pkgname" != "$pkgnobin" ] && [ -d "$pkgnobindir" ]; then
+        PACKAGE_DIR="$pkgnobindir"
+      else
+        PACKAGE_DIR="${{targets.outdir}}/${{targets.destdir}}"
+      fi
+
+      if [ "$PACKAGE_DIR" == "${{targets.contextdir}}" ]; then
+        echo "ERROR: Package can not split files from itself!" && exit 1
+      fi
+
+      cd "$PACKAGE_DIR" || exit 0
+
+      for i in usr/bin; do
+        if [ -e "$PACKAGE_DIR/$i" ] || [ -L "$PACKAGE_DIR/$i" ]; then
+          d="${{targets.contextdir}}/${i%/*}" # dirname $i
+          mkdir -p "$d"
+          mv "$PACKAGE_DIR/$i" "$d"
+          rmdir "$PACKAGE_DIR/${i%/*}" 2>/dev/null || :
+        fi
+      done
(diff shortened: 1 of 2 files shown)


pipelines/split/bin.yaml
```diff
@@ -0,0 +1,40 @@
...
+      pkgname="${{context.name}}"
+      pkgnobin="${pkgname%%-bin}"
+      pkgnobindir="${{targets.outdir}}/$pkgnobin"
+
+      if [ -n "${{inputs.package}}" ]; then
+        PACKAGE_DIR="${{targets.outdir}}/${{inputs.package}}"
+      elif [ "$pkgname" != "$pkgnobin" ] && [ -d "$pkgnobindir" ]; then
+        PACKAGE_DIR="$pkgnobindir"
+      else
+        PACKAGE_DIR="${{targets.outdir}}/${{targets.destdir}}"
```
smoser, line 24: this seems wrong, doesn't it? targets.destdir will always be a full path, won't it?
If so, then prefixing with targets.outdir is always wrong. right?

pipelines/split/bin.yaml
```diff
@@ -0,0 +1,40 @@
...
+      fi
+
+      if [ "$PACKAGE_DIR" == "${{targets.contextdir}}" ]; then
+        echo "ERROR: Package can not split files from itself!" && exit 1
```
smoser, line 28: don't && exit 1 here. just exit 1 on a new line. if the echo fails for some reason, then you will actually go _not_ exit here.
shell programming is madness . https://smoser.github.io/2023/04/27/set-e-bad.html

```
$ sh -ec 'false && echo exit 1; echo still here'
still here
```

pipelines/split/bin.yaml
```diff
@@ -0,0 +1,40 @@
...
+      fi
+
+      cd "$PACKAGE_DIR" || exit 0
+
+      for i in usr/bin; do
```
smoser, line 33: shellcheck would tell you that https://www.shellcheck.net/wiki/SC2043 .
I dont care about that too much, but there is value in shellcheck.

pipelines/split/bin.yaml
```diff
@@ -0,0 +1,40 @@
+name: Split executable files
+
+needs:
+  packages:
+    - busybox
+
+inputs:
+  package:
```
smoser, line 8: maybe 'from' is a better name? or redundant 'split-from' ?

pipelines/split/bin.yaml
```diff
@@ -0,0 +1,40 @@
...
+        PACKAGE_DIR="$pkgnobindir"
+      else
+        PACKAGE_DIR="${{targets.outdir}}/${{targets.destdir}}"
+      fi
+
+      if [ "$PACKAGE_DIR" == "${{targets.contextdir}}" ]; then
+        echo "ERROR: Package can not split files from itself!" && exit 1
+      fi
+
+      cd "$PACKAGE_DIR" || exit 0
```
smoser, line 31: this feels strange to me. I'm not sure why we wouldn't fail here.

smoser (changes requested):
I like this a lot, and really simplifies the multi-python workflow. thank you.
I would suggest tw or melange proper for this to avoid the copy / paste maintenance on pipelines/

xnox:
Oh, I do often want to have a split-bin pipeline, however given all split pipelines are in melange, can we please consider testing this here, but actually adding it to melange?

smoser:
oh, i see you largely copied this from melange split/ . so yeah, :+1: to putting this in melange.

pipelines/split/bin.yaml
dannf: i think that's better, but sticking with this for consistency with the others

pipelines/split/bin.yaml
dannf: nice catch, will fix in melange

dannf:
moved to melange: https://github.com/chainguard-dev/melange/pull/2030

PR #55608, opened 2025-06-08
pipelines: Add test/pip-check

dannf:
- **pipelines: add test/pip-check**
- **py3-google-auth: Respect upstream's cachetools pin**

### DIFF
--- a/pipelines/test/tw/pip-check.yaml
+++ b/pipelines/test/tw/pip-check.yaml
@@ -1,0 +1,22 @@
+name: Run 'pip check' to verify runtime deps.
+
+needs:
+  packages:
+    - tw-pip-check
+
+inputs:
+  python:
+    description: path to python interpreter to use
+    default: DEFAULT
+
+pipeline:
+  - name: "pip check"
+    runs: |
+      set +x
+      python="${{inputs.python}}"
+
+      if [ "$python" = "DEFAULT" ]; then
+        tw-pip-check
+      else
+        tw-pip-check --python="$python"
+      fi
--- a/py3-google-auth.yaml
+++ b/py3-google-auth.yaml
@@ -1,7 +1,7 @@
 package:
   name: py3-google-auth
   version: "2.40.3"
-  epoch: 0
+  epoch: 1
   description: Google Authentication Library
   copyright:
     - license: Apache-2.0
@@ -32,6 +32,10 @@ pipeline:
       tag: v${{package.version}}
       expected-commit: ca94ead4035beea4741dc5384449032f8e6f75d8
 
+  - runs: |
+      # Fail to let us know we can unpin
+      grep '"cachetools>=2.0.0,<6.0"' setup.py
+
 subpackages:
   - range: py-versions
     name: py${{range.key}}-${{vars.pypi-package}}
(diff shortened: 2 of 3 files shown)


smoser:
I think this is a win.

I had added pip-check once before, in a #24499 .  It was dropped after discussion with Pris.  Summary can be seen in 2583b5d0c3d95752291596fa50ae6e6e932f3b70 . more details in [slack thread](https://chainguard-dev.slack.com/archives/C06A87T8N66/p1721755310929879).  The summary was "melange sca should do that".

The version there uses pip-zipapp to avoid the pip dependency, and also has a default behavior so it does not need it input.  I think both of those are useful improvements.

I'd also like to suggest moving this to tw where it can be more centrally updated. While I do think that is a more easily maintainable solution, I wouldn't block on that.  (I'd love to see all the pipelines/py moved there to reduce the yaml-shell to one place.

pipelines/test/pip-check.yaml
```diff
@@ -0,0 +1,15 @@
...
+inputs:
+  python:
+    description: python interpreter to use
+
+pipeline:
+  - name: "pip check"
+    runs: |
+      if ! ${{inputs.python}} -c 'import pip' > /dev/null 2>&1; then
+        pyver=$(${{inputs.python}} -c \
+          "import sys; print('.'.join(sys.version.split('.')[:2]))")
```
justinvreeland: Nit: you should use `version_info` or one of the tuples from platform to avoid pulling in weird version info.  Though with the split it's not a huge deal here.


https://docs.python.org/3/library/sys.html#sys.version

dannf: good call, fixed.

dannf:
> I think this is a win.
> 
> I had added pip-check once before, in a #24499 . It was dropped after discussion with Pris. Summary can be seen in [2583b5d](https://github.com/wolfi-dev/os/commit/2583b5d0c3d95752291596fa50ae6e6e932f3b70) . more details in [slack thread](https://chainguard-dev.slack.com/archives/C06A87T8N66/p1721755310929879). The summary was "melange sca should do that".

IIUC, the argument is that `melange` should just magically add the deps instead of failing a test later. I don't disagree. We're asking a dev to manually add deps that it could just figure out. And those deps could become unnecessary overtime.  But `melange` can't do that today, and in the meantime we have broken packages. And there's nothing wrong with testing the result anyway.

> The version there uses pip-zipapp to avoid the pip dependency, and also has a default behavior so it does not need it input. I think both of those are useful improvements.
> 
> I'd also like to suggest moving this to tw where it can be more centrally updated. While I do think that is a more easily maintainable solution, I wouldn't block on that. (I'd love to see all the pipelines/py moved there to reduce the yaml-shell to one place.

OK, your version is better. Let me shove it in `tw`.

nit: one thing that feels disorganized is that we have some test pipelines under `test/`, and others under their ecosystem name (e.g. `py/`). I ended up putting mine under `test/` because it feels very akin to `ldd-check`.

dannf:
> OK, your version is better. Let me shove it in `tw`.

https://github.com/chainguard-dev/tw/pull/95

smoser (approved):

PR #55632, opened 2025-06-09
initial build for electric

developer-guy:
Fixes:

Related:

#### For new package PRs only

**NOTE TO THE REVIEWER**:

_it didn't work until you are in `/app` folder so this is why I start the binary under `bin/electric`_

### DIFF
--- a/electric.yaml
+++ b/electric.yaml
@@ -1,0 +1,113 @@
+package:
+  name: electric
+  version: "1.0.19"
+  epoch: 0
+  description: General-purpose programming language and runtime environment
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - elixir
+      - erlang
+      - erlang-dev
+      - protoc
+  environment:
+    MIX_ENV: "prod"
+    MIX_TARGET: "application"
+
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: 543b20446747ebd9431d7175f4838bd13a7e1779
+      repository: https://github.com/electric-sql/electric
+      tag: "@core/sync-service@${{package.version}}"
+
+  - working-directory: "packages/sync-service"
+    runs: |
+      mix local.hex --force && mix local.rebar --force
+      mix deps.get
+      mix deps.update sentry
+      mix deps.compile
+      mix compile
+      mix release
+      mkdir -p ${{targets.contextdir}}/app
+      cp -r _build/application_prod/rel/electric/* ${{targets.contextdir}}/app/
+      install -Dm0755 _build/application_prod/rel/electric/bin/electric ${{targets.contextdir}}/app/bin/entrypoint
+      install -Dm755 _build/application_prod/rel/electric/bin/electric \
+        ${{targets.contextdir}}/usr/bin/electric
+
+update:
+  enabled: true
+  github:
+    identifier: electric-sql/electric
+    strip-prefix: "@core/sync-service@"
+  ignore-regex-patterns:
+    - "@core/elixir-client"
+    - "@electric-sql/y-electric"
+    - "@electric-sql/react"
+    - "@core/elixir-client"
+    - "@electric-sql/experimental"
+
+test:
+  environment:
+    contents:
+      packages:
+        - postgresql-client
+        - shadow
+        - sudo-rs
+        - postgresql
+        - curl
+        - wait-for-it
+    environment:
+      PGDATA: /tmp/test_db
+      PGUSER: wolfi
+      DATABASE_URL: "postgresql://wolfi:password@localhost:5432/testdb?sslmode=disable"
+      ELECTRIC_INSECURE: "true"
+      ELECTRIC_PORT: "5000"
+  pipeline:
+    - name: "Test database creation"
+      runs: |
+        useradd $PGUSER
+        sudo -u $PGUSER initdb -D $PGDATA
+        sudo -u $PGUSER pg_ctl -D $PGDATA -l /tmp/logfile start
+        createdb testdb
+        psql -lqt | cut -d \| -f 1 | grep -qw testdb
+    - name: "Test basic read/write operations"
+      runs: |
+        psql -d testdb -c "CREATE TABLE test_table (id SERIAL PRIMARY KEY, test_value VARCHAR(50));"
+        psql -d testdb -c "INSERT INTO test_table (test_value) VALUES ('Hello, PostgreSQL!');"
+        psql -d testdb -t -A -c "SELECT test_value FROM test_table WHERE id=1;"
+    - working-directory: /app
+      runs: |
+        set -euo pipefail
+        # avoid having error like "Electric requires wal_level >= logical."
+        psql "$DATABASE_URL" -c "ALTER SYSTEM SET wal_level = 'logical';" > /dev/null 2>&1
+        sudo -u $PGUSER pg_ctl -D /tmp/test_db restart > /dev/null 2>&1
+
+        # Run the electric binary with a simple command to ensure it starts correctly
+        bin/electric start > /tmp/electric.log 2>&1 &
+        ELECTRIC_PID=$!
+        echo "Electric service started with PID $ELECTRIC_PID"
+
+        # Wait for electric to be ready
+        wait-for-it --timeout=30 localhost:$ELECTRIC_PORT
+
+        # Check if it's up
+        status_code=$(curl -s -o /dev/null -w "%{http_code}" "http://localhost:$ELECTRIC_PORT/v1/shape?table=test_table&offset=-1")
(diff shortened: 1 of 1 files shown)


electric.yaml
```diff
@@ -0,0 +1,108 @@
...
+        psql -d testdb -t -A -c "SELECT test_value FROM test_table WHERE id=1;"
+    - working-directory: /app
+      runs: |
+        # avoid having error like "Electric requires wal_level >= logical."
+        psql "$DATABASE_URL" -c "ALTER SYSTEM SET wal_level = 'logical';"
+        sudo -u $PGUSER pg_ctl -D /tmp/test_db restart
+
+        # Run the electric binary with a simple command to ensure it starts correctly
+        bin/electric start &
+        PID=$!
```
AnmolVirdi:
```suggestion
        # avoid having error like "Electric requires wal_level >= logical."
        psql "$DATABASE_URL" -c "ALTER SYSTEM SET wal_level = 'logical';" > /dev/null 2>&1
        sudo -u $PGUSER pg_ctl -D /tmp/test_db restart > /dev/null 2>&1

        # Run the electric binary with a simple command to ensure it starts correctly
        bin/electric start > /tmp/electric.log 2>&1 &
        PID=$!
```

electric.yaml
```diff
@@ -0,0 +1,108 @@
...
+      cp -r _build/application_prod/rel/electric/* ${{targets.contextdir}}/app/
+      cp -r _build/application_prod/rel/electric/bin/electric ${{targets.contextdir}}/app/bin/entrypoint
+      install -Dm755 _build/application_prod/rel/electric/bin/electric \
+        ${{targets.contextdir}}/usr/bin/electric
+
+update:
+  enabled: true
+  github:
+    identifier: electric-sql/electric
+    strip-prefix: "@core/sync-service@"
```
AnmolVirdi, line 48: We might need to add ignore regex for `@core/elixir-client`

developer-guy: I added bunch of regex patterns, thanks!

AnmolVirdi (approved):

electric.yaml
```diff
@@ -0,0 +1,114 @@
...
+    - license: Apache-2.0
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - elixir
+      - erlang
+      - erlang-dev
```
tulilirockz, line 16: We need to be very much careful here, elixir and erlang can have OTP mismatch errors, I have some comments on slack about this

electric.yaml
```diff
@@ -0,0 +1,114 @@
...
+        - wait-for-it
+    environment:
+      PGDATA: /tmp/test_db
+      PGUSER: wolfi
+      DATABASE_URL: "postgresql://wolfi:password@localhost:5432/testdb?sslmode=disable"
+      ELECTRIC_INSECURE: "true"
+  pipeline:
+    - name: "Test database creation"
+      runs: |
+        useradd $PGUSER
```
tulilirockz, line 75: I believe we can do the same thing from https://github.com/wolfi-dev/os/pull/55782#discussion_r2138572054 here too, that way we remove the `shadow` dependency and we hand this off to melange to handle user creation instead

electric.yaml
```diff
@@ -0,0 +1,114 @@
...
+    runs: |
+      mix local.hex --force && mix local.rebar --force
+      mix deps.get
+      mix deps.update sentry
+      mix deps.compile
+      mix compile
+      mix release
+      mkdir -p ${{targets.contextdir}}/app
+      cp -r _build/application_prod/rel/electric/* ${{targets.contextdir}}/app/
+      cp -r _build/application_prod/rel/electric/bin/electric ${{targets.contextdir}}/app/bin/entrypoint
```
tulilirockz:
```suggestion
      install -Dm0755 _build/application_prod/rel/electric/bin/electric ${{targets.contextdir}}/app/bin/entrypoint
```

electric.yaml
```diff
@@ -0,0 +1,114 @@
...
+  - working-directory: "packages/sync-service"
+    runs: |
+      mix local.hex --force && mix local.rebar --force
+      mix deps.get
+      mix deps.update sentry
+      mix deps.compile
+      mix compile
+      mix release
+      mkdir -p ${{targets.contextdir}}/app
+      cp -r _build/application_prod/rel/electric/* ${{targets.contextdir}}/app/
```
tulilirockz, line 39:
```suggestion
      install -Dm0644 _build/application_prod/rel/electric/* ${{targets.contextdir}}/app/
```
I _think_ that works?

electric.yaml
```diff
@@ -0,0 +1,114 @@
...
+    - "@electric-sql/experimental"
+
+test:
+  environment:
+    contents:
+      packages:
+        - postgresql-client
+        - shadow
+        - sudo-rs
+        - glibc-locales
```
tulilirockz: What happens when we dont have `glibc-locales`? This seems like something that should be pulled in as a dependency directly

electric.yaml
```diff
@@ -0,0 +1,114 @@
...
+        psql -d testdb -t -A -c "SELECT test_value FROM test_table WHERE id=1;"
+    - working-directory: /app
+      runs: |
+        set -euo pipefail
+        # avoid having error like "Electric requires wal_level >= logical."
+        psql "$DATABASE_URL" -c "ALTER SYSTEM SET wal_level = 'logical';" > /dev/null 2>&1
+        sudo -u $PGUSER pg_ctl -D /tmp/test_db restart > /dev/null 2>&1
+
+        # Run the electric binary with a simple command to ensure it starts correctly
+        bin/electric start > /tmp/electric.log 2>&1 &
```
tulilirockz, line 93: Is there some way we can specify a port here? Would be handy so we can have more consistent tests (i.e. imagine if upstream changes the default port or something)

electric.yaml
```diff
@@ -0,0 +1,114 @@
...
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - elixir
+      - erlang
+      - erlang-dev
+      - make
```
tulilirockz: `build-base` already pulls `make` in

electric.yaml
developer-guy: yes, we can have `ELECTRIC_PORT`: https://electric-sql.com/docs/api/config#electric-port

electric.yaml
developer-guy: I tried but having other issues like:

```bash
2025/06/11 14:44:13 WARN + su - wolfi -c 'initdb -D '"'"'/tmp/test_db'"'"
2025/06/11 14:44:13 WARN su: can't change directory to '/home/wolfi': Permission denied
2025/06/11 14:44:13 WARN su: can't change directory to '/': Permission denied
```

electric.yaml
developer-guy: let's remove it to see what's gonna happen 🫠

electric.yaml
developer-guy: noted!

electric.yaml
developer-guy: build-base is required otherwise it fails and yes I confirmed that by removing it 🫠

tulilirockz: No no, I mean removing `make`, we can just get rid of `make` because we have `build-base`

electric.yaml
tulilirockz: You can use `run-as: 0` so it wont break like that

electric.yaml
developer-guy: it worked.

electric.yaml
developer-guy: I tried but had another error:

```bash
2025/06/11 18:22:21 WARN [su: authenticate] Password: su: PAM error: Authentication service cannot retrieve authentication info
```

so right now we can merge the PR maybe we can get back to this later on.

electric.yaml
```diff
@@ -0,0 +1,113 @@
...
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: 543b20446747ebd9431d7175f4838bd13a7e1779
+      repository: https://github.com/electric-sql/electric
+      tag: "@core/sync-service@${{package.version}}"
+
+  - working-directory: "packages/sync-service"
+    runs: |
+      mix local.hex --force && mix local.rebar --force
```
tulilirockz: You can specify
```suggestion
      mix local.hex --force && mix local.rebar rebar3 /usr/bin/rebar3
```
We might want to use our rebar3 package from wolfi for this

tulilirockz (dismissed):

electric.yaml
developer-guy: it worked just fine, thanks for the suggestion!

tulilirockz (approved):
Looks good! Maybe we can revisit this eventually to get the `environment.accounts:` thing working

PR #55637, opened 2025-06-09
curl: rebuild without overlinking

xnox:
The 2.28 build of curl linked and references symbols that are not
present in any of the dependencies. This leads to libcurl failures at
runtime. This this with a rebuild against reverted toolchain.

### DIFF
--- a/curl.yaml
+++ b/curl.yaml
@@ -1,7 +1,7 @@
 package:
   name: curl
   version: "8.14.1"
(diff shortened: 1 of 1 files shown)


xnox:
2.28 elastic-build presubmit here is successful and has low abi.

techalchemy (approved):

dannf (approved):

PR #55649, opened 2025-06-09
feat(ffmpeg): enable libass support

igcasarino:
FFMPEG-7.1: Add support for libass

Related: #55648 

- [X] REQUIRED cgr.dev/chainguard/wolfi-base and ghcr.io/wolfi-dev/sdk images successfully build
- [X] REQUIRED cgr.dev/chainguard/wolfi-base and ghcr.io/wolfi-dev/sdk contain no obsolete (no longer built) packages
- [X] Upon launch, does `apk upgrade --latest` successfully upgrades packages or performs no actions


igcasarino:
Hi @AmberArcadia can you help me with this PR? Thanks!

AmberArcadia:
Hey, thanks for your contribution! I've raised this internally and we'll get back to you shortly.

In the mean time, would you be open to adding a test case for this? :)

ffmpeg-7.1.yaml
```diff
@@ -34,8 +34,11 @@ environment:
       - dav1d-dev
       - expat-dev
       - fontconfig-dev
+      - freetype-dev
+      - fribidi-dev
       - harfbuzz-dev
       - lame-dev
+      - libass
```
dannf, line 41: huh, seems like `libass` needs to `split/dev` a subpackage

igcasarino: If we split libass, the base package won't provide anything

dannf: Well noticed. That points to a bigger problem, that we're likely statically linking it. if we are, then there's nothing telling us to rebuild ffmpeg if a CVE is found in libass. Looks like we need to:

1. Add `-Ddefault_library=both` as an `opts` to the `meson/configure` pipeline.
2. Add a -dev package with `split/dev`
3. We should bump any rdeps that build against `libass` to make sure they are linking dynamically... but we don't appear to have any.
4. While we're at it, the build says that we have asm optimizations disabled. Can you add `nasm` as a build-dependency?

igcasarino: PR #57025 Modifies libass

PR #55658, opened 2025-06-09
glibc: stop shipping /etc/ld.so.cache

dannf:
Avoids poisoning our layering scheme, which expects files owned by
a package to remain static after the image build. `apko` will
regenerate this file.


murraybd (approved):

dannf:
I'm doing an image build sanity check here before I merge:
https://github.com/chainguard-images/images-private/pull/10294

PR #55724, opened 2025-06-10
Enable version update for videolan/x264

debasishbsws:
- Moved to use GitLab repo for the source code
- The repo does not release any version, we are using the stable version, the latest commit for update
- Using release monitor as it tracks the latest commit by date as a version

_ Bumping epoch of ffmpeg-7.1 to rebuild with this new version of x264, otherwise there will be a conflict with provides.

### DIFF
--- a/ffmpeg-7.1.yaml
+++ b/ffmpeg-7.1.yaml
@@ -1,7 +1,7 @@
 package:
   name: ffmpeg-7.1
   version: "7.1.1"
-  epoch: 7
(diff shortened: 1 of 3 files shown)


debasishbsws:
All the dependent downstream packages are updated with an epoch bump.

bentasker:
Note on ci-so-check - the deps *are* being bumped.

bentasker (approved):

PR #55739, opened 2025-06-10
airflow-3: use uv

kranurag7:


kranurag7:
Closing given, this is already on main

PR #55782, opened 2025-06-10
quality: extend test it-tools package

yashpimple:
The PR adds functional test to validate that Nginx serves the expected it-tools content over HTTP.

### DIFF
--- a/it-tools.yaml
+++ b/it-tools.yaml
@@ -1,7 +1,7 @@
 package:
   name: it-tools
   version: 2024.10.22
-  epoch: 0
+  epoch: 1
   description: Collection of handy online tools for developers, with great UX
   copyright:
     - license: GPL-3.0
@@ -74,8 +74,30 @@ update:
     tag-filter: v
 
 test:
+  environment:
+    contents:
+      packages:
+        - curl
+        - shadow
   pipeline:
     - name: Verify nginx
       runs: |
         stat /etc/nginx/conf.d/default.conf
         stat /usr/share/nginx/html/index.html
+    # Using manual runs instead of test/daemon-check-output since nginx doesn't emit usable stdout on startup.
+    - name: "Start and verify nginx manually"
+      runs: |
+        useradd -r -s /sbin/nologin nginx 2>/dev/null
+        mkdir -p /var/lib/nginx/tmp/client_body
+        mkdir -p /var/log/nginx
+        mkdir -p /var/run
+
+        nginx -g 'daemon off;' &
+        wait-for-it -t 10 --strict localhost:80
+
+        response=$(curl -sf http://localhost)
+        if ! echo "$response" | grep -iq 'it-tools'; then
+          echo "ERROR: Nginx is running, but expected 'it-tools' content not found in HTTP response"
+          echo "$response"
+          exit 1
+        fi
--- a/it-tools.yaml
+++ b/it-tools.yaml
@@ -92,7 +92,7 @@ test:
         mkdir -p /var/log/nginx
         mkdir -p /var/run
 
-        nginx -g 'daemon off;' &
+        nginx -g 'daemon off;' > /tmp/nginx.log 2>&1 &
         wait-for-it -t 10 --strict localhost:80
 
         response=$(curl -sf http://localhost)
--- a/it-tools.yaml
+++ b/it-tools.yaml
@@ -85,7 +85,6 @@ test:
     contents:
       packages:
         - curl
-        - shadow
         - wait-for-it
   pipeline:
     - name: Verify nginx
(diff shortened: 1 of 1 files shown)


it-tools.yaml
```diff
@@ -74,8 +74,31 @@ update:
     tag-filter: v
 
 test:
+  environment:
+    contents:
```
tulilirockz:
```suggestion
  environment:
    accounts:
      groups:
        - groupname: nginx
          gid: 101
      users:
        - username: nginx
          uid: 101
    contents:
```

it-tools.yaml
```diff
@@ -74,8 +74,31 @@ update:
...
+        - wait-for-it
   pipeline:
     - name: Verify nginx
       runs: |
         stat /etc/nginx/conf.d/default.conf
         stat /usr/share/nginx/html/index.html
+    # Using manual runs instead of test/daemon-check-output since nginx doesn't emit usable stdout on startup.
+    - name: "Start and verify nginx manually"
+      runs: |
+        useradd -r -s /sbin/nologin nginx 2>/dev/null
```
tulilirockz:
```suggestion
```

it-tools.yaml
```diff
@@ -74,8 +74,31 @@ update:
...
+        mkdir -p /var/lib/nginx/tmp/client_body
+        mkdir -p /var/log/nginx
+        mkdir -p /var/run
```
tulilirockz:
```suggestion
        mkdir -p /var/lib/nginx/tmp/client_body /var/log/nginx /var/run
```

it-tools.yaml
```diff
@@ -74,8 +74,31 @@ update:
...
+
+        nginx -g 'daemon off;' > /tmp/nginx.log 2>&1 &
```
tulilirockz:
```suggestion
        nginx -g 'daemon off;' 2>&1 | tee /tmp/nginx.log &
        NGINX_PID=$!
        trap 'kill -9 $NGINX_PID' EXIT
```

it-tools.yaml
```diff
@@ -74,8 +74,31 @@ update:
...
+        wait-for-it -t 10 --strict localhost:80
+
+        response=$(curl -sf http://localhost)
+        if ! echo "$response" | grep -iq 'it-tools'; then
```
tulilirockz: I recommend grepping for something like the description just to make sure the entire thing is working, like: "IT Tools - Handy online tools for developers

tulilirockz (commented):
Looking awesome! I just have a few little things that might be nice to use instead of what we are doing here, mostly nits so this is non-blocking

it-tools.yaml
Dentrax: Let's migrate to `test/daemon-check-output` pipeline.

Dentrax (changes requested):

it-tools.yaml
yashpimple: @Dentrax! initially i made use of `test/daemon-check-output`, but nginx doesn't emit any consistent output to stdout/stderr on successful startup. This caused the `expected_output` check to fail. Additionally to ensure compatibility, especially with QEMU runners, I opted for a manual background start .

it-tools.yaml
```diff
@@ -74,8 +74,36 @@ update:
...
+      groups:
+        - groupname: nginx
+          gid: 101
+      users:
+        - username: nginx
+          uid: 101
+    contents:
+      packages:
+        - curl
+        - shadow
```
tulilirockz:
```suggestion
```
I think you can get rid of this now that we are using `environment.accounts`

it-tools.yaml
tulilirockz: I also don't see any other package with `test/daemon-check-output` using nginx as the daemon

Dentrax (approved):

it-tools.yaml
```diff
@@ -74,8 +74,35 @@ update:
...
+        - wait-for-it
   pipeline:
     - name: Verify nginx
       runs: |
         stat /etc/nginx/conf.d/default.conf
         stat /usr/share/nginx/html/index.html
+    # Using manual runs instead of test/daemon-check-output since nginx doesn't emit usable stdout on startup.
+    - name: "Start and verify nginx manually"
+      runs: |
+        mkdir -p /var/lib/nginx/tmp/client_body /var/log/nginx /var/run
```
Dentrax, line 97: is this needed?

yashpimple: yes @Dentrax we create /var/lib/nginx/tmp/client_body, /var/log/nginx, and /var/run to avoid runtime errors from Nginx when those expected directories are missing or not writable, especially in our package test these paths are commonly required for request buffering and logging.

PR #55812, opened 2025-06-10
create a symlink between /var/run and /run just like google did

developer-guy:
Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/cadvisor.yaml
+++ b/cadvisor.yaml
@@ -35,8 +35,11 @@ subpackages:
     description: Compatibility package for cadvisor
     pipeline:
       - runs: |
-          mkdir -p "${{targets.contextdir}}"/run
-          ln -sf /var/run "${{targets.contextdir}}"/run
+          mkdir -p "${{targets.contextdir}}"/var
+          ln -sf /run "${{targets.contextdir}}"/var/run
+    test:
+      pipeline:
+        - runs: test "$(readlink /var/run)" = "/run"
 
 update:
   enabled: true
--- a/cadvisor.yaml
+++ b/cadvisor.yaml
@@ -1,7 +1,7 @@
 package:
   name: cadvisor
   version: "0.53.0"
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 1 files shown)


cadvisor.yaml
```diff
@@ -30,6 +30,14 @@ pipeline:
 
   - uses: strip
 
+subpackages:
+  - name: ${{package.name}}-compat
+    description: Compatibility package for cadvisor
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.contextdir}}"/run
+          ln -sf /var/run "${{targets.contextdir}}"/run
```
kranurag7: we can consider adding a stat or readlink -f based test here if it's required so that we don't miss this in b/w updates.

developer-guy: this test is not gonna work since that we create that /var/run folder at image level so its giving me an error like /var/run does not exist, so, this is why I left the tests empty

tulilirockz:
Just as a note: This is literally a FHS compliancy issue - [see this](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s13.html) - We should have already been doing this with `wolfi-base` or some other base package like that, maybe a good point to improve on at some point?

cadvisor.yaml
tulilirockz:
```suggestion
          mkdir -p "${{targets.contextdir}}"/var/run
          ln -sf ../run "${{targets.contextdir}}"/var/run
```
Usually the "right way" would be to symlink `/run` to `../run` - does this break anything?

```shell
docker run --rm -it quay.io/fedora/fedora:latest ls -lah /var/run
```
```
lrwxrwxrwx 1 root root 6 Jun 10 06:48 /var/run -> ../run

```

tulilirockz:
```suggestion
          ln -sf /run "${{targets.contextdir}}"/var/run
```

kranurag7 (approved):

tulilirockz (approved):

PR #55957, opened 2025-06-11
jenkins-docker-agent: new pkg script

MickeyPvX:
Adding `jenkins-docker-agent` package for tracking the agent script used in Jenkins inbound agent container images.

#### For new package PRs only


jenkins-2.yaml
```diff
@@ -71,6 +71,13 @@ pipeline:
       git checkout "${remoting_version}"
       mvn -B -V -e --ntp -Dchangelist="${remoting_version}" -Pquick-build clean package
 
+      # The jenkins docker-agent shares similar tags to remoting
+      cd ..
+      git clone https://github.com/jenkinsci/docker-agent
+      cd docker-agent
+      agent_version=$(git tag -l "${remoting_version}*" | sort -r | head -1)
+      git checkout "${agent_version}"
```
kranurag7: we can consider using git-checkout pipeline here with destination?

MickeyPvX: The reason we're doing it this way has to do with the tags that are in the docker-agent repo, which are horribly non-semver - I'm using the same type of logic in the block preceding this, as in, the `jenkins` repo has a `remoting.version` in its pom.xml that this uses to get the right version of `jenkins-remoting`.  The `docker-agent` uses the same tags but with a `-<some number>` suffix on them (to make it worse).

So the `jenkins-remoting` and (new) `jenkins-docker-agent` will share the same version as the main `jenkins` package, but under the hood they are the versions specified by the `jenkins` pom.xml

kranurag7 (approved):

PR #55993, opened 2025-06-11
proj: update cmake flags, expand tests

Dentrax:
Expand tests, update cmake flags to sync with how upstream does.

* https://github.com/OSGeo/PROJ/blob/d87fcea05ed78271ab6210a71ab128bde0f51b61/Dockerfile#L21
* https://github.com/OSGeo/gdal/blob/v3.11.0/docker/ubuntu-full/bh-proj.sh

### DIFF
--- a/proj.yaml
+++ b/proj.yaml
@@ -2,7 +2,7 @@
 package:
   name: proj
   version: "9.6.2"
-  epoch: 0
+  epoch: 1
   description: PROJ coordinate transformation software library
   copyright:
     - license: MIT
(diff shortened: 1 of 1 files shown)


kranurag7 (approved):
this is already the default of cmake/configure pipeline 

```bash
-DCMAKE_INSTALL_PREFIX=/usr
```
but being explicit again is not a blocker here.

other options looks good and obvious. 
thank you very much for expanding the tests here.

PR #56040, opened 2025-06-12
Enable update packages for multiple packages

debasishbsws:
- armadillo.yaml
- libexif.yaml
- librdkafka.yaml
- librdkafka.yaml
- rlwrap.yaml

### DIFF
--- a/armadillo.yaml
+++ b/armadillo.yaml
@@ -1,7 +1,7 @@
 package:
   name: armadillo
-  version: 12.6.7
-  epoch: 2
+  version: 14.4.3
+  epoch: 0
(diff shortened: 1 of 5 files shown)


powersj:
Looks like you correctly updated the gdal but still a ci-so-check failure?

debasishbsws:
@powersj It ment to be fail. [Docs](https://github.com/wolfi-dev/os/blob/main/docs/CI_CHECKS.md#what-to-do-when-the-so-name-check-fails)
> Note: The So-Name CI check will still fail after all the above steps are completed. This is expected behavior, and a Wolfi Maintainer admin will need to override the failed check.

powersj (approved):

PR #56070, opened 2025-06-12
fix: make unshare optional when not needed

89luca89:
Right now the `unshare` command is used as a workaround to a limitation of the bubblewrap runner, this will break on both QEMU and Docker

Make this optional, and skip unshare if not needed

Workaround for: https://github.com/chainguard-dev/prodsec/issues/282


egibs:
CVE fixes are in this PR: https://github.com/wolfi-dev/os/pull/56012

So I think we're good.

egibs (approved):

PR #56089, opened 2025-06-12
kafka-4.0: Add iamguarded-compat

IdlePhysicist:


kafka-4.0.yaml
```diff
@@ -103,6 +103,49 @@ subpackages:
...
 
+  - name: kafka-iamguarded-compat-${{vars.major-minor-version}}
+    description: "compat package with iamguarded/kafka image"
+    dependencies:
+      provides:
+        - kafka-iamguarded-compat=${{package.full-version}}
+      runtime:
+        - merged-usrsbin
+        - wait-for-port
+        - wolfi-baselayout
```
kranurag7: nit: we can remove `merged-usrbin` and `wolfi-baselayout` from runtime deps given we're already usrmerged.

IdlePhysicist: I assume that we can.  I'll push it up

tcnghia (approved):

kafka-4.0.yaml
IdlePhysicist: Done

PR #56095, opened 2025-06-12
Version streamed and filled in versions for numpy

AmberArcadia:
Added version streams for numpy to handle their multiple supported versions (and to get a v2 build for another of our packages)

#### For new version streams


py3-numpy-2.0.yaml
```diff
@@ -0,0 +1,146 @@
...
+vars:
+  pypi-package: numpy
+
+data:
+  # Numpy 2.0 only supports Python 3.10-3.12
+  - name: py-versions
+    items:
+      3.10: '310'
+      3.11: '311'
+      3.12: '312'
```
justinvreeland, line 26: Missing 3.13 or have an extra 3.13 in the supported package.

justinvreeland (changes requested):
Looks good over all. We probably don't need to add the provides for any version but `py3-numpy-1.26` since it's the version that is expected and we don't need to propagate that forward. It'd be ideal if people chose the version they want imo.  Why are we version streaming so many version though, It seems like we'd just want `1.X` and `2.X`?

py3-numpy-2.0.yaml
AmberArcadia: Whoops yep I thought I removed the extra in the supported package, thanks!
Edit: oh actually I did but didn't commit/push it yet haha

AmberArcadia:
@justinvreeland thanks for your review! The reason I did so many versions is they're apparently all still supported: https://endoflife.date/numpy. Also, I've gotten the feeling they'll come in handy during the migration. 

Re: the provides, I know we talked about this and went back and forth. I think having them available but commented out gives us the flexibility of opting in but not requiring the new versions, and makes it easy to change that if we wanted down the line.

AmberArcadia:
I confirmed that this should have 0 impact on other packages and they'll all stay on the old 1.26 version but new version stream. I confirmed this via building all packages locally, then attempting a build of a package with `numpy` and `py3-supported-numpy` and observing the version used was 1.26. Migrations can be performed by dropping `numpy` and swapping `py3-supported-numpy` with `py3-supported-numpy-2.<x>`. We can/will add `numpy` as `provides` in the future when we're ready to bump the default.

AmberArcadia:
Guess I'll be disabling this test too in 1.25... our build system is too good :) 
image

justinvreeland (approved):

kranurag7 (approved):
thank you!! @AmberArcadia 

I checked the index locally, played with it for a while and it looks good. With provides commented out for 2.x; this looks good to land.

AmberArcadia:
Thanks for double checking that this won't accidentally cause any unanticipated migrations!

PR #56148, opened 2025-06-13
distribution: move comment inside entrypoint.sh

kranurag7:
without this we run into exec format error


mritunjaysharma394 (approved):
lgtm, thanks for finding it

PR #56198, opened 2025-06-13
Remove the onetbb.yaml update libtbb.yaml

murraybd:
Thee onetbb source package definition contains the same upstream source code as libtbb's source package definition. The onetbb package, it's variants, including py3-tbb are all unused. Subsequently, let's remove the onetbb package and copy the good bits from it, a -doc subpakage and a better smoke test to libtbb.

```
[ 10:25AM ]  [ brian-murray@log:~/src/wolfi-os(main✗) ]
 $ grep onetbb *.yaml                                                                              (main↑7994|…1)
onetbb.yaml:  name: onetbb
onetbb.yaml:    description: onetbb FIXME
onetbb.yaml:  - name: onetbb-dev
onetbb.yaml:        - onetbb
onetbb.yaml:    description: onetbb dev
onetbb.yaml:  - name: onetbb-doc
onetbb.yaml:    description: onetbb readme
onetbb.yaml:        - onetbb-dev
[ 10:25AM ]  [ brian-murray@log:~/src/wolfi-os(main✗) ]
 $ grep py3-tbb *.yaml                                                                             (main↑7994|…1)
onetbb.yaml:  - name: py3-tbb
```
I also checked enterprise and extra packages.


dannf (changes requested):
My only thought here is that we could keep the onetbb name, since it's the name of the upstream project, and make `libtbb` a subpackage. It does mean we'll now have a useless empty `onetbb` package. But, on the plus side, if we do move to soname-based-library-packages, the `libtbb` package will need to become empty anyway.

Would you mind withdrawing the old unused/packages?

murraybd:
Superseded by https://github.com/wolfi-dev/os/pull/56211.

PR #56211, opened 2025-06-13
Remove libtbb and update opentbb for its removal

murraybd:
The onetbb source package definition contains the same upstream source as libtbb, however there are some slight differences in the contents of the package definition. This PR keeps the onetbb name and adds libtbb as a subpackage to, additionally the onetbb-dev package is renamed libtbb-dev and some more testing is added.


dannf (approved):
nice! thanks for doing that.

shall we withdraw the old packages?

murraybd:
> nice! thanks for doing that.
> 
> shall we withdraw the old packages?

Do you mean the libtbb and libtbb-dev ones?

dannf:
> > nice! thanks for doing that.
> > shall we withdraw the old packages?
> 
> Do you mean the libtbb and libtbb-dev ones?

anything that we are no longer building, so if someone installs it, they won't be getting something unpatched.  I didn't compare, but it might just be the python ones?

PR #56556, opened 2025-06-15
aws-sigv4-proxy package build

developer-guy:
#### For new package PRs only

### DIFF
--- a/aws-sigv4-proxy.yaml
+++ b/aws-sigv4-proxy.yaml
@@ -1,0 +1,69 @@
+package:
+  name: aws-sigv4-proxy
+  version: "1.10"
+  epoch: 0
+  description: This project signs and proxies HTTP requests with Sigv4
+  copyright:
+    - license: Apache-2.0
+
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: fb42fe8db8d4ebed5f6761402d8a2617963d4c72
+      repository: https://github.com/awslabs/aws-sigv4-proxy
+      tag: v${{package.version}}
+
+  - uses: go/build
+    with:
+      output: aws-sigv4-proxy
+      packages: ./cmd/aws-sigv4-proxy
+
+subpackages:
+  - name: ${{package.name}}-compat
+    description: Compatibility package to place binaries in the location expected by upstream Dockerfile
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}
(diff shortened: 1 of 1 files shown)


aws-sigv4-proxy.yaml
```diff
@@ -0,0 +1,38 @@
...
+update:
+  enabled: true
+  github:
+    identifier: awslabs/aws-sigv4-proxy
+    strip-prefix: v
+
+test:
+  pipeline:
+    - runs: |
+        aws-sigv4-proxy --help
```
mamccorm: Given the description of the application, we should be able to do some functional testing at the package level. Ca you add some functional test coverage?

mamccorm: If it expects AWS creds, we may be able to mock it by setting some of the env variables and seeing how far we get. We should be able to at a minimum get to some log validation, but we may be able to get further and do some functional tests. Example: https://github.com/wolfi-dev/os/pull/56476/files

aws-sigv4-proxy.yaml
```diff
@@ -0,0 +1,44 @@
...
+      output: aws-sigv4-proxy
+      packages: ./cmd/aws-sigv4-proxy
+
+subpackages:
+  - name: ${{package.name}}-compat
+    description: Compatibility package to place binaries in the location expected by upstream Dockerfile
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}
+          ln -sf /usr/bin/aws-sigv4-proxy ${{targets.contextdir}}/aws-sigv4-proxy
```
Aditevil, line 27: dependencies:
      runtime:
        - ${{package.name}}
        
     test:
      pipeline:
        - name: Test compatibility symlink
          runs: |
            readlink -f ${{targets.contextdir}}/aws-sigv4-proxy | grep "/usr/bin/aws-sigv4-proxy"

Aditevil (dismissed):

aws-sigv4-proxy.yaml
mamccorm: I'd still like to see more tests than this. The application mentions signing HTTP requests - have we dug into how this application runs and would be tested? It seems like theres functional testing we might be able to do here.

Same comment above re: mocking  AWS creds if we need to

developer-guy: Based on your feedback, I improved the tests to verify that the request is being properly signed by checking for the presence of a signed request entry in the log file.

mamccorm (approved):

PR #56631, opened 2025-06-15
use auth/gradle pipeline during build

kranurag7:
selectively picked from https://github.com/wolfi-dev/os/pull/56225 which were passing


cpanato:
Why is that required? There is a GitHub issue for this?

cpanato (approved):

PR #56743, opened 2025-06-16
Apply /dev/shm fix to melange/init

egibs:
Relates to: #56731

This applies the change from the aforementioned PR to the Melange `init` script.

### DIFF
--- a/melange.yaml
+++ b/melange.yaml
@@ -1,7 +1,7 @@
 package:
   name: melange
   version: "0.26.12"
-  epoch: 0
+  epoch: 1
   description: build APKs from source code
   copyright:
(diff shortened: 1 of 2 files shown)


89luca89:
This package is not used anywhere, we can even drop it, as all production has been moved to newer melange

egibs:
> This package is not used anywhere, we can even drop it, as all production has been moved to newer melange

Ah, yep. Forgot about the recent changes: https://github.com/chainguard-dev/melange/blob/65e385aff51b530faf6c3c001be62e5dc8b91aef/pkg/container/qemu_runner.go#L1469-L1471

We can make a cleanup PR in the future.

PR #56746, opened 2025-06-16
fix: QEMU: correctly propagate mounts in chroot

89luca89:
We need `rbind` to correctly propagate nested mount inside the chroot


aborrero:
LGTM.

egibs (approved):

PR #56747, opened 2025-06-16
kubernetes-reflector: init at 9.1.12

tulilirockz:
New package: `kubernetes-reflector`


MickeyPvX:
Added some basic checks for expected files from the build and that it runs and tries to access the Kubernetes API.  More in-depth tests of reflecting actual resources will be in the image tests.

jamonation (approved):

PR #56806, opened 2025-06-17
add just-newuidmap-newgidmap 

anushkamittal20:
Add a package for just newuidmap and newgidmap binaries

### DIFF
--- a/shadow.yaml
+++ b/shadow.yaml
@@ -2,7 +2,7 @@
 package:
   name: shadow
   version: "4.17.4"
-  epoch: 40
+  epoch: 41
   description: PAM login and passwd utilities
   copyright:
     - license: BSD-3-Clause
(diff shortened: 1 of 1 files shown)


shadow.yaml
```diff
@@ -185,7 +185,7 @@ subpackages:
       - runs: |
           mkdir -p ${{targets.subpkgdir}}/usr/bin
           mv ${{targets.destdir}}/usr/bin/getsubids ${{targets.subpkgdir}}/usr/bin/
-          mv ${{targets.destdir}}/usr/bin/new*idmap ${{targets.subpkgdir}}/usr/bin/
+          cp ${{targets.destdir}}/usr/bin/new*idmap ${{targets.subpkgdir}}/usr/bin/
```
Dentrax: Instead of `cp` and have 2 subpkgs that provides same binaries, was wondering if we should have the `just-newuidmap-newgidmap` as a runtime dep here. And omit this line. This package also has tests for those 2 binaries, it does seem like a duplicate. Wdyt?

```suggestion
```

anushkamittal20: I think I get what you are saying, I have made the changes, wdyt @Dentrax

Dentrax (approved):

PR #56821, opened 2025-06-17
chainguard-2.28: Test build openblas

xnox:
Failing with:

 "gfortran: fatal error: '-fuse-linker-plugin', but liblto_plugin.so not found"

will check if this can be packaged or not


techalchemy (approved):

xnox:
1 neutral check is still broken build on 2.28 =)

PR #56835, opened 2025-06-17
fix: renamed package and added installed container entrypoint scripts

rahulduvedi:
We're creating new package and image named `kubo` upstream project [`ipfs/kubo.`](https://github.com/ipfs/kubo)
The existing images use the `go-ipfs `and package `ipfs` naming for historical reasons, but adopting the correct upstream name improves clarity and consistency.
So renaming the packege `ipfs` to `kubo`
For more context refer to this thread [here](https://chainguard-dev.slack.com/archives/C062D52J3T9/p1750169806574819)

### DIFF
--- a/ipfs.yaml
+++ b/ipfs.yaml
@@ -7,10 +7,10 @@ package:
     - license: MIT
   dependencies:
     runtime:
+      - busybox
+      - gosu
       - "posix-libc-utils" # "getent" required
       - tini
-      - gosu
-      - busybox
     provides:
       - kubo:${{package.full-version}}
 
@@ -61,7 +61,7 @@ pipeline:
       mkdir -p ${{targets.destdir}}/data/ipfs
       mkdir -p ${{targets.destdir}}/ipfs ${{targets.destdir}}/ipns
       mkdir -p ${{targets.destdir}}/container-init.d
-      
+
       # Install container entrypoint scripts
       install -Dm755 bin/container_daemon ${{targets.destdir}}/usr/local/bin/start_ipfs
       install -Dm755 bin/container_init_run ${{targets.destdir}}/usr/local/bin/container_init_run
--- a/ipfs.yaml
+++ b/ipfs.yaml
@@ -1,15 +1,18 @@
 package:
   name: ipfs
   version: "0.35.0"
-  epoch: 2
+  epoch: 3
   description: An IPFS implementation in Go
   copyright:
     - license: MIT
   dependencies:
     runtime:
       - "posix-libc-utils" # "getent" required
+      - tini
+      - gosu
+      - busybox
     provides:
-      - kubo
+      - kubo:${{package.full-version}}
 
 environment:
   contents:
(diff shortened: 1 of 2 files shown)


ipfs.yaml
```diff
@@ -1,15 +1,18 @@
...
   version: "0.35.0"
-  epoch: 2
+  epoch: 3
   description: An IPFS implementation in Go
   copyright:
     - license: MIT
   dependencies:
     runtime:
+      - busybox
+      - gosu
```
mamccorm: Was this intentional? Dropping a shell onto the package as a runtime dependency is a significant change. Did you intend to add this at the test level to facilitate tests?

mamccorm (dismissed):
Blocking query re: addition of runtime deps at the package level

ipfs.yaml
rahulduvedi: Apologies, Yes I did added busybox to facilitate tests, Removed.
Thanks @mamccorm

kubo.yaml
```diff
@@ -1,15 +1,15 @@
 package:
-  name: ipfs
```
Dentrax, line 2: I'm confused a bit here, why we renamed the file and package? Can we add the reasoning in the PR desc?

Dentrax (changes requested):

kubo.yaml
rahulduvedi: Here is the discussion you can refer to :
https://chainguard-dev.slack.com/archives/C062D52J3T9/p1750169806574819

kubo.yaml
```diff
@@ -1,15 +1,15 @@
...
   version: "0.35.0"
-  epoch: 2
+  epoch: 3
   description: An IPFS implementation in Go
   copyright:
     - license: MIT
   dependencies:
     runtime:
+      - gosu
       - "posix-libc-utils" # "getent" required
```
Dentrax, line 11:
```suggestion
      - posix-libc-utils # "getent" required
```

kubo.yaml
```diff
@@ -1,15 +1,15 @@
...
-    provides:
-      - kubo
```
Dentrax, line 12: Should we add `ipfs=${{package.full-version}}` for backward-compatibility?

kubo.yaml
```diff
@@ -46,25 +46,55 @@ pipeline:
         golang.org/x/oauth2@v0.27.0
         github.com/pion/interceptor@v0.1.39
 
-  - runs: |
-      CGO_ENABLED=1 GOOS=$(go env GOOS) GOARCH=$(go env GOARCH) GOFLAGS=-buildvcs=false make build GOTAGS=openssl
-      install -m755 -D ./cmd/ipfs/ipfs "${{targets.destdir}}"/usr/bin/ipfs
+  - uses: go/build
+    with:
+      packages: ./cmd/ipfs
+      output: ipfs
```
Dentrax, line 54: Even though package name is `kubo`, we provide the binary as `ipfs` is that correct?

kubo.yaml
```diff
@@ -46,25 +46,55 @@ pipeline:
...
+  - name: ${{package.name}}-compat
+    description: Compatibility symlinks for IPFS container entrypoint scripts
+    dependencies:
+      runtime:
+        - ${{package.name}}
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/usr/local/bin
+          ln -sf /usr/bin/start_ipfs ${{targets.contextdir}}/usr/local/bin/start_ipfs
+          ln -sf /usr/bin/container_init_run ${{targets.contextdir}}/usr/local/bin/container_init_run
```
Dentrax, line 79: Should we have another symlink for backward-compatibility? `ipfs -> kubo`.

kubo.yaml
```diff
@@ -46,25 +46,55 @@ pipeline:
...
+
+            # Verify the targets are executable
+            test -x /usr/bin/start_ipfs
+            test -x /usr/bin/container_init_run
 
 test:
+  environment:
+    contents:
+      packages:
+        - ${{package.name}}-compat
```
Dentrax: We shouldn't need to do this; if test fails without this change, its likely we did break something.
```suggestion
```

kubo.yaml
rahulduvedi: The package was wrongly named, it is not changed on upstream.
And acc to the discussion we don't need to maintain a backward-compatibility.
Please correct me if I am missing something.

kubo.yaml
rahulduvedi: Safely removed this, I will this in mind for my further work

Dentrax (dismissed):

kubo.yaml
mamccorm: +1, seems like an easy win, and a code comment above with a summary why we're adding

mritunjaysharma394 (dismissed):

mritunjaysharma394 (approved):

PR #56841, opened 2025-06-17
tarball-uploads: no-change rebuilds to generate upstream archive tarballs.

javacruft:
The tarball-upload service in production received fixes today to improve resilience and deal with large tarballs (such as binutils and glibc) which failed in prior rebuilds.

This is triggered from the publication of APK's to the Wolfi archive post merge and build; rebuild selected core packages that make use of this service to publish an upstream tarball to use in SBOM downloadLocation references.

For reference this is a smaller subset of the packages in Wolfi that don't use github, gitlab or upstream release tarballs to build.

### DIFF
--- a/aom.yaml
+++ b/aom.yaml
@@ -5,7 +5,7 @@
 package:
   name: aom
   version: "3.12.1"
-  epoch: 3
+  epoch: 4
   description: Alliance for Open Media (AOM) AV1 codec SDK
   copyright:
     - license: BSD-2-Clause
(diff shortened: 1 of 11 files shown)


smoser:
Could you add some context here as to what this does?

javacruft:
> Could you add some context here as to what this does?

That would be helpful - I'll update the summary

javacruft:
Will need an #eng-psa announce as includes glibc.

murraybd (approved):

PR #56862, opened 2025-06-17
qemu: use autoconf pipeline, explicit build options for aio/io-uring and ensure dummy --version  tests arent broken

tulilirockz:
Older title was too big, I believe this might be ok? 
- Also changed a few things to `install` instead of `cp`
- `qemu-edid` was displaying garbled data on the tests because the `help`/`version` options don't seem to exist
- Hopefully from now on we could specify what exactly we want on the build so we can clear up old dependencies easier in the future

### DIFF
--- a/qemu.yaml
+++ b/qemu.yaml
@@ -1,7 +1,7 @@
 package:
   name: qemu
   version: "10.0.2"
-  epoch: 1
+  epoch: 2
   description: "fast processor emulator"
   copyright:
     - license: GPL-2.0
@@ -28,7 +28,6 @@ environment:
       - liburing-dev
       - ncurses
       - ncurses-dev
-      - ninja-build
       - perl
       - pixman-dev
       - py3-packaging
@@ -47,12 +46,22 @@ pipeline:
       expected-commit: ff3419cbacdc9ad0715c716afeed65bb21a2bbbc
       tag: v${{package.version}}
 
+  - uses: autoconf/configure
+    with:
+      opts: |
+        --enable-system \
+        --enable-slirp \
+        --enable-linux-aio \
+        --enable-linux-io-uring \
+        --enable-vnc \
+        --target-list=x86_64-softmmu,aarch64-softmmu \
+        --prefix=/usr
+
+  - uses: autoconf/make
+
+  - uses: autoconf/make-install
+
   - runs: |
-      mkdir build
-      cd build
-      ../configure --enable-system --enable-slirp --enable-vnc --target-list=x86_64-softmmu,aarch64-softmmu --prefix=${{targets.destdir}}/usr
-      make -j$(nproc)
-      make install
       # Remove this firmware file, which we don't need, but requires so:dld.sl, which we don't have
       rm -f ${{targets.destdir}}/usr/share/qemu/hppa-firmware64.img
       # Can't use strip pipeline; there are some binaries that melange can't strip,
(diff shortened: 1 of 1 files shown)


qemu.yaml
```diff
@@ -82,15 +97,16 @@ subpackages:
...
       pipeline:
         - runs: |
-            qemu-img --version
-            qemu-img --help
+            qemu-img --version | grep "${{package.version}}"
+            qemu-img --help | grep "QEMU disk image utility"
+            qemu-io --version | grep "${{package.version}}"
+            qemu-io --help | grep "QEMU Disk exerciser"
+            qemu-nbd --version | grep "${{package.version}}"
+            qemu-nbd --help | grep "QEMU Disk Network Block Device Utility"
```
smoser: Since you're touching it, please use '-f' (fixed strings) to grep. And Also set 'set -o pipefail' otherwise the left side of the pipe could segfault every time and go unnoticed.

smoser (changes requested):
Was there a specific reason to move off ninja? Have you looked at what other distros do? I'm guessing if they use ninja there is some way to fail on missing options dependencies.

tulilirockz:
> Was there a specific reason to move off ninja? Have you looked at what other distros do? I'm guessing if they use ninja there is some way to fail on missing options dependencies.

It was honestly by accident LOL I completely forgot I removed that for some reason when I was testing this package locally. I'll just re-add it

qemu.yaml
```diff
@@ -178,16 +195,20 @@ test:
     - uses: test/tw/ldd-check
     - uses: test/pkgconf
     - runs: |
+        set -o pipefail
         for i in /usr/bin/qemu-system-*; do
-          $i --help 2>/dev/null && echo "$i --> OKAY" || echo "$i --> FAIL"
-          $i --version 2>/dev/null && echo "$i --> OKAY" || echo "$i --> FAIL"
+          {
+            $i --help | grep -F "QEMU emulator"
+          } 2>/dev/null && echo "$i --> OKAY" || echo "$i --> FAIL"
```
smoser, line 202: 2 things here.
1. my local test here didn't have anything output on stderr, so i dont think you  need to swallow it by redirecting to /dev/null.  if it failed, we'd want to see it.
2. if you FAIL here you'll just print FAIL and go on happily, and no one will ever know unless they're looking at logs.

I guess you didn't actually cause 2, it was like that before. 

something like this maybe:

```bash
set -o pipefail
fails=""
for i in /usr/bin/qemu-system-*; do
  $i --help | grep -F "QEMU emulator" &&
      echo "$i --help: PASS" || { echo "$i --help: FAIL"; fails="$fails $i"; }
done
if [ -n "$fails" ]; then
  exit 1
fi
```

smoser (changes requested):

qemu.yaml
smoser: Alternatively, you can just do:

```
for i in /usr/bin/qemu-syste-*; do
  "$i" --help | grep -F "QEMU emulator"
done
```
  
that works because these scripts run with 'set -e'. if either side of that pipe exits non-zero it will fail.  you just wont get a nice message as to _what_ failed.

smoser:
> It was honestly by accident LOL I completely forgot I removed that for some reason when I was testing this package locally. I'll just re-add it


Maybe I wasn't clear.  You changed the build system from ninja to autoconf.  Why did you do that?

tulilirockz:
> You changed the build system from ninja to autoconf. Why did you do that?

I believe I haven't, though! - The `autoconf` pipeline does the same things we are doing right now:

- [autoconf/configure](https://github.com/chainguard-dev/melange/blob/9e77b79041d636d0712f0eb084e483d8d6e04180/pkg/build/pipelines/autoconf/configure.yaml#L38-L47) - `./configure` + a few other options that we dont need to specify if we use it
- [autoconf/make](https://github.com/chainguard-dev/melange/blob/9e77b79041d636d0712f0eb084e483d8d6e04180/pkg/build/pipelines/autoconf/make.yaml#L20) - `make -J(nproc)` + a target
- [autoconf/make-install](https://github.com/chainguard-dev/melange/blob/9e77b79041d636d0712f0eb084e483d8d6e04180/pkg/build/pipelines/autoconf/make-install.yaml#L20) - `make install` + some options there too

I guess what we gain from it is just not needing to specify a few options like `--prefix` and stuff. If this is an issue I can just remove it and maybe PR this some other time separately, I might not be understanding the ramifications of using these pipelines instead of raw `make blah`.

qemu.yaml
tulilirockz: Sounds perfectly fine to me! Figuring something out here and I'll post a commit soon. Thank you!!

PR #56909, opened 2025-06-18
fix(sqlpad): recreate client yarn env for 7.5.4

kwmonroe:
A recent package update masked failing tests, which happened as a result of broken sqlpad client deps upstream. This was initially noticed on the [package update pr](https://github.com/wolfi-dev/os/pull/55707#pullrequestreview-2931019389) post merge.

This PR fixes the issue by pinning the sqlpad client to a more recent react version and recreating the yarn environment.

Related: https://github.com/chainguard-dev/image-release-stats/issues/5842

### DIFF
--- a/sqlpad.yaml
+++ b/sqlpad.yaml
@@ -1,7 +1,7 @@
 package:
   name: sqlpad
   version: "7.5.4" # when updating check the patch below as it contains dependency version updates which may downgrade if upstream upgrades them
-  epoch: 0
+  epoch: 1
   description: Web-based SQL editor. Legacy project in maintenance mode.
   copyright:
     - license: MIT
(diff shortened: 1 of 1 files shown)


sqlpad.yaml
```diff
@@ -42,6 +42,17 @@ pipeline:
 
       yarn install
 
+  - working-directory: /home/build/client
+    runs: |
+      # TODO(kwmonroe): v7.5.4 client yarn.lock is broken; recreate it with known good deps
```
kbsteere, line 47: Think you can remove the TODO since you regenerate the lock below.

kwmonroe: it was more of a todo for if/when upstream fixes it, we can get rid of the whole block

vishal-chdhry (approved):

sqlpad.yaml
vishal-chdhry: @kbsteere marking this as resolved to fix image build

PR #56960, opened 2025-06-18
Fix/systemd unit tests

89luca89:
Failing tests are because systemd-tests cannot detect us as a container-like solution and assumes we're a full system
So we manually skip what previously was automatically skipped by their checks.

Also we will add to the init the creation of `/.dockerenv` to trick other software to treat us more like a container than a full VM

Fix: https://github.com/chainguard-dev/prodsec/issues/313


zmarano (dismissed):

egibs (dismissed):

systemd.yaml
```diff
@@ -181,7 +182,8 @@ subpackages:
             -s test-tpm2 \
             -s test-fd-util \
             -s test-fstab-generator.sh \
-            -s test-rm-rf
+            -s test-rm-rf \
+            -s test-execute
```
zmarano, line 191: Looks like you need to also skip `test-mount-util` and `test-namespace`

89luca89: They should be auto-skipped now as they detect `/.dockerenv` we're adding now

89luca89: I see they're failing because first they need the new microvm-init, I'll explicitly add them

89luca89: Now PR depends on https://github.com/wolfi-dev/os/pull/56962

89luca89:
Depedns on PR: 

- https://github.com/wolfi-dev/os/pull/56962

zmarano (approved):

PR #57015, opened 2025-06-19
gcc-14: enable gfortran

xnox:
Enable versioned gfortran too. chainguard-2.28 is still on gcc-14 and some projects need gfortran and the 15 gfortran doesn't work with gcc 14 provided crt files and lto plugins. Instead of trying to provide 15 ctr files and lto plugins it is easier to just enable gfortran build in gcc-14 for now for consistent toolchain builds.

### DIFF
--- a/gcc-14.yaml
+++ b/gcc-14.yaml
@@ -1,7 +1,7 @@
 package:
   name: gcc-14
   version: 14.3.0
-  epoch: 1
+  epoch: 2
   description: "the GNU compiler collection - version 14"
   copyright:
     - license: GPL-3.0-or-later WITH GCC-exception-3.1
@@ -98,7 +98,7 @@ pipeline:
             --with-arch=$march \
             --with-tune=$mtune \
             --with-specs=$specs \
-            --enable-languages=c,c++ \
+            --enable-languages=c,c++,fortran \
             --enable-bootstrap \
             --enable-gnu-indirect-function \
             --enable-gnu-unique-object \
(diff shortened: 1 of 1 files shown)


gcc-14.yaml
```diff
@@ -197,22 +197,63 @@ subpackages:
...
   - name: 'gcc-14-default'
     description: 'Use GCC 14 as system gcc'
     dependencies:
       provides:
         - gcc=${{package.full-version}}
+        - gfortran=${{package.full-version}}
       runtime:
         - gcc-14=${{package.full-version}}
+        - gfortran-14=${{package.full-version}}
+        - libgfortran
```
sergiodj, line 247: Shouldn't this be `libgfortran-14=${{package.full-version}}` as well?  Or maybe just omit this runtime dependency since `gfortran-14` already depends on `libgfortran-14`?

sergiodj (changes requested):

gcc-14.yaml
xnox: @sergiodj no it should not! Because libgfortran-14 has link time dev library in /14/ lib directory, yet just compiled binaries typically need runtime library that is backwards compatible and should come from latest build.

This is same as libstdc++-14 not providing a public libstdc++ and that one is currently pulled in indirectly and by accident as toolchain itself needs libstdc++ in global library search path. However gfortran itself is not written in fortran and thus doesn't actually have runtime dependency on libgfortran.

xnox: Without this dep, the just compiled applications cannot execute. For example unit tests written in fortran.

sergiodj (approved):

gcc-14.yaml
xnox: I do agree it is unfortunate.

Maybe all gcc libraries should be shipped in versioned paths, and we should simply add /15/ path to ld.so.conf.d paths like we do for nvidia.

This would look weird and unlike what everyone else does.

Or possibly make gcc-15 provide versioned compileres; and have separate gcc-libs which only provides runtime libs. As in theory we can increase gcc-libs without upgrading compiler. But that is also weird.

PR #57019, opened 2025-06-19
py3-sphinxext-opengraph: add new package

aborrero:
Add new package, required dependency of some other packages, like smartypants, which has some sphinx-based docs that somehow load this opengraph extension.

Fixes: https://github.com/wolfi-dev/os/pull/56819

Related:

#### For new package PRs only

### DIFF
--- a/py3-sphinx.yaml
+++ b/py3-sphinx.yaml
@@ -57,7 +57,6 @@ subpackages:
         - py${{range.key}}-sphinxcontrib-qthelp
         - py${{range.key}}-sphinxcontrib-serializinghtml
         - py${{range.key}}-tomli
-        - py${{range.key}}-sphinxext-opengraph
     pipeline:
       - uses: py/pip-build-install
         with:
--- a/py3-sphinx.yaml
+++ b/py3-sphinx.yaml
@@ -1,7 +1,7 @@
 package:
   name: py3-sphinx
   version: "8.2.3"
-  epoch: 0
+  epoch: 1
   description: Python Documentation Generator
   copyright:
     - license: MIT
@@ -57,6 +57,7 @@ subpackages:
         - py${{range.key}}-sphinxcontrib-qthelp
         - py${{range.key}}-sphinxcontrib-serializinghtml
         - py${{range.key}}-tomli
+        - py${{range.key}}-sphinxext-opengraph
     pipeline:
       - uses: py/pip-build-install
         with:
--- a/py3-sphinx.yaml
+++ b/py3-sphinx.yaml
@@ -1,7 +1,7 @@
 package:
   name: py3-sphinx
   version: "8.2.3"
-  epoch: 1
+  epoch: 0
   description: Python Documentation Generator
   copyright:
     - license: MIT
--- a/py3-sphinxext-opengraph.yaml
+++ b/py3-sphinxext-opengraph.yaml
@@ -5,8 +5,6 @@ package:
   description: sphinxcontrib-devhelp is a sphinx extension to generate unique OpenGraph metadata
   copyright:
     - license: BSD-3-Clause
-  dependencies:
-    provider-priority: "0"
 
 vars:
   pypi-package: sphinxext-opengraph
@@ -15,10 +13,10 @@ vars:
 data:
   - name: py-versions
     items:
-      3.10: '310'
-      3.11: '311'
-      3.12: '312'
-      3.13: '313'
+      3.10: ''
+      3.11: ''
+      3.12: ''
+      3.13: ''
 
 environment:
   contents:
@@ -37,10 +35,6 @@ subpackages:
   - range: py-versions
     name: py${{range.key}}-${{vars.pypi-package}}
     description: python${{range.key}} version of ${{vars.pypi-package}}
-    dependencies:
-      provider-priority: ${{range.value}}
-      provides:
-        - py3-${{vars.pypi-package}}
     pipeline:
       - uses: py/pip-build-install
         with:
@@ -48,6 +42,7 @@ subpackages:
       - uses: strip
     test:
       pipeline:
+        - uses: test/tw/pip-check
         - uses: python/import
           with:
             python: python${{range.key}}
@@ -65,10 +60,7 @@ subpackages:
 
 test:
   pipeline:
-    - uses: python/import
-      with:
-        imports: |
-          # import ${{vars.import}} # AVOID_DEP_CYCLE(sphinx)
+    - uses: test/emptypackage
 
 update:
   enabled: true
--- a/py3-sphinxext-opengraph.yaml
+++ b/py3-sphinxext-opengraph.yaml
@@ -1,0 +1,77 @@
+package:
+  name: py3-sphinxext-opengraph
+  version: 0.10.0
+  epoch: 0
+  description: sphinxcontrib-devhelp is a sphinx extension to generate unique OpenGraph metadata
+  copyright:
+    - license: BSD-3-Clause
+  dependencies:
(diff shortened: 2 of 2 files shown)


py3-sphinxext-opengraph.yaml
```diff
@@ -0,0 +1,77 @@
...
+  pypi-package: sphinxext-opengraph
+  import: sphinxext.opengraph
+
+data:
+  - name: py-versions
+    items:
+      3.10: '310'
+      3.11: '311'
+      3.12: '312'
+      3.13: '313'
```
dannf: no need for the values here, just leave them empty.

py3-sphinxext-opengraph.yaml
```diff
@@ -0,0 +1,77 @@
+package:
+  name: py3-sphinxext-opengraph
+  version: 0.10.0
+  epoch: 0
+  description: sphinxcontrib-devhelp is a sphinx extension to generate unique OpenGraph metadata
+  copyright:
+    - license: BSD-3-Clause
+  dependencies:
+    provider-priority: "0"
```
dannf: no need for a `provider-priority` - we're not transitioning from an earlier `py3-sphinxext-opengraph`

py3-sphinxext-opengraph.yaml
```diff
@@ -0,0 +1,77 @@
...
+      provider-priority: ${{range.value}}
+      provides:
+        - py3-${{vars.pypi-package}}
+    pipeline:
+      - uses: py/pip-build-install
+        with:
+          python: python${{range.key}}
+      - uses: strip
+    test:
+      pipeline:
```
dannf, line 47: please add `test/tw/pip-check`

py3-sphinxext-opengraph.yaml
```diff
@@ -0,0 +1,77 @@
...
+        - py3.11-${{vars.pypi-package}}
+        - py3.12-${{vars.pypi-package}}
+        - py3.13-${{vars.pypi-package}}
+
+test:
+  pipeline:
+    - uses: python/import
+      with:
+        imports: |
+          # import ${{vars.import}} # AVOID_DEP_CYCLE(sphinx)
```
dannf: there should be nothing to import, this package should be empty (after provides are removed). Also, please add `test/emptypackage`

py3-sphinxext-opengraph.yaml
```diff
@@ -0,0 +1,77 @@
...
+        with:
+          python: python${{range.key}}
+      - uses: strip
+    test:
+      pipeline:
+        - uses: python/import
+          with:
+            python: python${{range.key}}
+            imports: |
+              # import ${{vars.import}} # AVOID_DEP_CYCLE(sphinx)
```
dannf: how is this used then? And can add a test that does *that*?

py3-sphinxext-opengraph.yaml
```diff
@@ -0,0 +1,77 @@
...
+      tag: v${{package.version}}
+
+subpackages:
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}
+    description: python${{range.key}} version of ${{vars.pypi-package}}
+    dependencies:
+      provider-priority: ${{range.value}}
+      provides:
+        - py3-${{vars.pypi-package}}
```
dannf: please remove. this pattern is for transitioning from a previously non-multiversioned package. this package is net new.

dannf (changes requested):

py3-sphinxext-opengraph.yaml
aborrero: done.

py3-sphinxext-opengraph.yaml
aborrero: I think importing the library directly is a good to test. I have tested that on my system.

However, the most common way of using this extension is by using the sphinx library directly.

py3-sphinxext-opengraph.yaml
aborrero: done.

py3-sphinxext-opengraph.yaml
aborrero: done.

py3-sphinxext-opengraph.yaml
aborrero: done.

py3-sphinxext-opengraph.yaml
aborrero: done.

py3-sphinx.yaml
```diff
@@ -57,6 +57,7 @@ subpackages:
         - py${{range.key}}-sphinxcontrib-qthelp
         - py${{range.key}}-sphinxcontrib-serializinghtml
         - py${{range.key}}-tomli
+        - py${{range.key}}-sphinxext-opengraph
```
dannf: Does it make sense to have a dependency here? `py3-sphinx` doesn't seem to actually *require* it.  I added pip-check to `py3-sphinx` to see if this was reported as a missing dependency, and it wasn't:

```
2025/06/20 08:33:59 INFO running 'pip check' with /usr/bin/python3.11 at version '3.11'
2025/06/20 08:33:59 INFO execute: /usr/bin/python3.11 /usr/share/pip-zipapp/pip-zipapp.pyz check
2025/06/20 08:34:00 INFO sphinx 8.2.3 has requirement sphinxcontrib-jsmath>=1.0.1, but you have sphinxcontrib-jsmath 1.0.1.dev20241212.
```

Perhaps it should be in `smartypants` or whatever in-between project actually requires it.

aborrero: good point. I will update soon.

py3-sphinxext-opengraph.yaml
dannf: My question was because this appears to be commented-out.
Can you explain the dep cycle that's preventing this? Is it because py3-sphinx is in the same PR?

py3-sphinx.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: py3-sphinx
   version: "8.2.3"
-  epoch: 0
+  epoch: 1
```
dannf: thx. but that begs the question, why bump it at all now?

aborrero: fair.

Updated.

py3-sphinxext-opengraph.yaml
aborrero: I just copy-pasted this code from the other sphinx packages.

I understand now what happens. The plugins and the core package have a sensitive dependency between them.

For example, if have the import here uncommented then I have to declare a runtime dependency on sphinx (the extension wont load without sphinx being installed).

But if I do so, then the pip check will fail because it has a weird error about the version:

2025/06/20 16:02:55 INFO sphinx 8.2.3 has requirement sphinxcontrib-jsmath>=1.0.1, but you have sphinxcontrib-jsmath 1.0.1.dev20241212.

So my proposal is to drop the pip check and leave the import test. Wdyt?

dannf: I think that's a bug with `shinxcontrib-jsmath`'s versioning. The version on pypi is 1.0.1:

https://pypi.org/project/sphinxcontrib-jsmath/

So somehow we're appending a `dev$DATED` extension. I think we should understand why and fix that - is that something you could look at?

dannf: This should do it - https://github.com/wolfi-dev/os/pull/57213

aborrero:
follow up on https://github.com/wolfi-dev/os/pull/57314

PR #57024, opened 2025-06-19
prometheus-operator: add iam subpackage

kranurag7:


prometheus-operator.yaml
```diff
@@ -130,6 +136,34 @@ subpackages:
               exit 1
             }
 
+  - name: ${{package.name}}-iamguarded-compat
+    description: "compat package for iamguarded"
+    pipeline:
+      - runs: |
+          mkdir -p /opt/iamguarded/prometheus-operator/bin/
```
klrmngr, line 143: would it be a good idea to use `${{package.name}}` here (and anywhere else that has `prometheus-operator`) to facilitate easier copy-paste in the future?

kranurag7: keeping it same here will facilitate easier copy-pasting b/w fips and non-fips. If we use package name then that'll change on fips

klrmngr (approved):

maxgio92 (approved):

PR #57122, opened 2025-06-20
openssf-compiler-options: clang 14

javacruft:
Add support for Clang 14.


dannf (approved):
just to confirm - have you verified that these options are all relevant for 14?https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.html

PR #57215, opened 2025-06-21
kubeflow-pipelines: upgrade envoy config

xnox:
Upstream ships
https://github.com/kubeflow/pipelines/commits/master/third_party/metadata_envoy/envoy.yaml
which is used in their dockerfile under the name envoy.yaml

From history, we packaged this file under a different name always, and
our image has always used it under a different name. Not sure why.

Also the contents of our statically vendored config, doesn't seem to
match any versions or copies of the upstream envoy.yaml config. Not
sure if this is intentional (because our envoy is different) or
original enovy.yaml was taken from somewhere else (i.e. some other
repo or docs).

Also I don't know what this does, but imho we should attempt to
upgrade to upstream config of envoy for this image.

This was noticed by @justinvreeland [here](https://github.com/wolfi-dev/os/pull/53546#discussion_r2159565580).

TODO: test image release with this update


cmwilson21:
Thanks for the contribution!

PR #57272, opened 2025-06-23
vale: new package

ajayk:


eslerm:
This package contains many dictionaries. Malcontent found patterns in these dictionaries that match malware. False positive.

srishtih (approved):

PR #57277, opened 2025-06-23
JuiceFS package addition

AnmolVirdi:
**JuiceFS package** `Current version: v1.3.0`
- JuiceFS is a distributed POSIX file system built on top of Redis and S3.

**Type**: Go binary
**REF**: https://github.com/juicedata/juicefs

**Tests:**
- JuiceFS  format volumes
- Tools, Services commands.


juicefs-1.3.yaml
```diff
@@ -0,0 +1,133 @@
...
+        mkdir -p /data/minio
+        minio server /data/minio --address :9000 --console-address :9001 > /tmp/minio.log 2>&1 &
+        wait-for-it localhost:9000 -t 30
+    - name: "Create test bucket using mc"
+      runs: |
+        export MINIO_ACCESS_KEY=minioadmin
+        export MINIO_SECRET_KEY=minioadmin
+        mc alias set local http://localhost:9000 minioadmin minioadmin > /tmp/mc.log 2>&1
+        mc mb local/juicefs-bucket
+        sleep 2 # Wait for juicefs bucket to show up
```
maxgio92: Can we add a retry on failure? This way we'd avoid unnecessary backpressure in CI

juicefs-1.3.yaml
```diff
@@ -0,0 +1,133 @@
...
+    - name: "Format juicefs volume and dump logs"
+      runs: |
+        export ACCESS_KEY=minioadmin
```
maxgio92: We can use `test.environment.environment` so we share the credentials across the steps

juicefs-1.3.yaml
```diff
@@ -0,0 +1,133 @@
...
+        grep -q "FSTree" /tmp/dump.json
+    - name: "Test service commands"
+      runs: |
+        # Creds needed for JuiceFS's built in S3 gateway
+        export MINIO_ROOT_USER=minioadmin
+        export MINIO_ROOT_PASSWORD=minioadmin
+        export WEBDAV_USER=minioadmin
+        export WEBDAV_PASSWORD=minioadmin
+        # Setup S3 Gateway
+        juicefs gateway redis://localhost localhost:9002 > gateway.logs 2>&1 &
```
maxgio92: Can we check these logs?

juicefs-1.3.yaml
```diff
@@ -0,0 +1,133 @@
...
+        wait-for-it localhost:9002
+        # Validate endpoint health
+        curl -v http://localhost:9002/minio/health/ready 2>&1 | grep -q "200 OK"
+        # Validate
+        mc alias set jfs http://localhost:9002 minioadmin minioadmin
+        mc mb jfs/testvolume
+        mc ls jfs 2>&1 | grep "testvolume/"
+        # Start WebDAV server
+        juicefs webdav redis://localhost localhost:9003 > webdav.logs 2>&1 &
```
maxgio92: We can separate each functionality in a separate step (S3 gateway and WebDAV server)

juicefs-1.3.yaml
```diff
@@ -0,0 +1,133 @@
...
+      packages:
+        - redis
+        - minio
+        - mc
+        - wait-for-it
+        - curl
+  pipeline:
+    - name: "Version test"
+      runs: |
+        juicefs --version 2>&1 | grep ${{package.version}}
```
maxgio92:
```suggestion
        juicefs --version 2>&1 | grep -q ${{package.version}}
```

juicefs-1.3.yaml
AnmolVirdi: Updated with `grep -q "Bucket created successfully"` validation

juicefs-1.3.yaml
AnmolVirdi: Updated!

juicefs-1.3.yaml
AnmolVirdi: Thanks, was keeping them in the same block as `service commands` since they needed same env vars. Since they're global now, I've divided them into two.

juicefs-1.3.yaml
AnmolVirdi: Updated, thanks!

juicefs-1.3.yaml
AnmolVirdi: Updated

juicefs-1.3.yaml
```diff
@@ -0,0 +1,139 @@
...
+update:
+  enabled: true
+  ignore-regex-patterns:
+    - '-rc'
+    - '-beta'
+    - '-dev'
+  github:
+    identifier: juicedata/juicefs
+    strip-prefix: v
+    tag-filter: v1.3
```
debasishbsws:
```suggestion
    tag-filter: v1.3.
```

Not an issue but this would be better as the older one will also match with v1.30.x

AnmolVirdi: Ah noted, thanks! Updated.

juicefs-1.2
Dentrax, line 1: This is empty package.

juicefs-1.3.yaml
```diff
@@ -0,0 +1,139 @@
...
+      runs: |
+        juicefs format --bucket http://localhost:9000/juicefs-bucket --access-key $ACCESS_KEY --secret-key $SECRET_KEY redis://localhost testvolume 2>&1 | grep -q "Volume is formatted"
+        juicefs dump redis://localhost /tmp/dump.json
+        grep -q "Sustained" /tmp/dump.json
+        grep -q "DelFiles" /tmp/dump.json
+        grep -q "FSTree" /tmp/dump.json
+    - name: "Test S3 Gateway setup"
+      runs: |
+        # Setup S3 Gateway
+        juicefs gateway redis://localhost localhost:9002 > gateway.logs 2>&1 &
```
Dentrax: Any reasons why we don't use `test/daemon-check-output` pipeline here?

juicefs-1.3.yaml
```diff
@@ -0,0 +1,139 @@
...
+        # Validate endpoint health
+        curl -v http://localhost:9002/minio/health/ready 2>&1 | grep -q "200 OK"
+        # Validate connection
+        mc alias set jfs http://localhost:9002 minioadmin minioadmin
+        mc mb jfs/testvolume
+        mc ls jfs 2>&1 | grep -q "testvolume/"
+    - name: "Test WebDAV Server"
+      runs: |
+        # Start WebDAV server
+        juicefs webdav redis://localhost localhost:9003 > webdav.logs 2>&1 &
```
Dentrax: Same here.

juicefs-1.3.yaml
```diff
@@ -0,0 +1,139 @@
...
+    with:
+      deps: |-
+        golang.org/x/net@v0.38.0
+        github.com/golang/glog@v1.2.4
+        github.com/tidwall/gjson@v1.9.3
+        golang.org/x/oauth2@v0.27.0
+
+  - uses: go/build
+    with:
+      packages: .
```
Dentrax, line 31: Why we don't pass `tags`? https://github.com/juicedata/juicefs/blob/fcef17452db35ca8875ce1475baae37340879d42/Makefile#L50

If there is a reason, can we drop a comment here?

juicefs-1.3.yaml
```diff
@@ -0,0 +1,139 @@
...
+
+  - uses: go/bump
+    with:
+      deps: |-
+        golang.org/x/net@v0.38.0
+        github.com/golang/glog@v1.2.4
+        github.com/tidwall/gjson@v1.9.3
+        golang.org/x/oauth2@v0.27.0
+
+  - uses: go/build
```
Dentrax, line 29: There are other juicefs binaries provided by upstream, should we have a range pipeline here? https://github.com/juicedata/juicefs/blob/fcef17452db35ca8875ce1475baae37340879d42/Makefile#L30-L44

AnmolVirdi: For now, we're not building other juicefs binaries since they require `ceph`, `glusterfs` and `foundationdb` headerfiles.

juicefs-1.3.yaml
AnmolVirdi: After adding the tags, we need `ceph`, `glusterfs` and `foundationdb` dev packages in the build environment. For now, we're just going for building the main `juicefs` binary.
https://github.com/juicedata/juicefs/blob/fcef17452db35ca8875ce1475baae37340879d42/Makefile#L24

juicefs-1.2
AnmolVirdi: @Dentrax could you please let me know more about this? There's `juicefs` binary.

juicefs-1.3.yaml
AnmolVirdi: We had a combined test step for all service commands earlier (s3 gateway and webdav) and recently split it in different test blocks recently.
I've updated tests with using `daemon-check-output`

juicefs-1.3.yaml
AnmolVirdi: Updated.

juicefs-1.2
Dentrax: Sorry - im not following, do we put binary in the repo?

AnmolVirdi: We keep the binary packaged in `/usr/bin` as usual
I'm little confused with understanding this:
https://github.com/wolfi-dev/os/pull/57277#discussion_r2223684079

Dentrax: Please list the staged files in the branch and notice the juicefs-1.2 empty file.

AnmolVirdi: Thanks, updated.

Dentrax (approved):

PR #57284, opened 2025-06-23
openssl: build without jitterentropy on 2.28

xnox:
jitterentropy is compiled to 2.34 glibc ABI level, thus for 2.28
configuration build openssl without it.


sil2100 (approved):
Don't love how it looks, but it's fine!

PR #57288, opened 2025-06-23
Add epoch field to redpanda-25.1 package definition

utieyin:
Adding epoch key to redpanda package to fix https://github.com/chainguard-dev/internal-dev/issues/14038


anushkamittal20 (approved):
lgtm!

PR #57318, opened 2025-06-23
usrmerge /lib - thingsboard version bump

zmarano:


smoser (dismissed):

xnox (dismissed):

xnox (dismissed):

thingsboard.yaml
```diff
@@ -93,16 +99,20 @@ subpackages:
...
       - name: Extract tb-mqtt-transport to target directories
         runs: |
+          mkdir -p ${{targets.subpkgdir}}/usr/lib
           mkdir -p ${{targets.subpkgdir}}/usr/bin
           cd msa/transport/mqtt/target
           dpkg-deb -x tb-mqtt-transport.deb ${{targets.subpkgdir}}
           cp start-tb-mqtt-transport.sh ${{targets.subpkgdir}}/usr/bin
           mkdir -p ${{targets.subpkgdir}}/config
           cp ${{targets.subpkgdir}}/usr/share/tb-mqtt-transport/conf/* ${{targets.subpkgdir}}/config
+          cp -r lib/* ${{targets.subpkgdir}}/usr/lib
```
a-crate: I think you will end up with files in both /lib and /usr/lib this way, since unpacking the deb will stick all the /lib/systemd stuff in targets.subpkdir. I think you need to `mv` or `rm` `${{targets.subpkdir}}/lib`.

We will see when the package builds.

zmarano: Yeah - fixed that- waiting to see when it finishes building.

a-crate (dismissed):

a-crate (approved):

PR #57320, opened 2025-06-23
Add provides jre into the default jdk

debasishbsws:
@xnox

### DIFF
--- a/openjdk-20.yaml
+++ b/openjdk-20.yaml
@@ -141,6 +141,7 @@ subpackages:
         - libxrender
         - libxtst
         - openjdk-20-jre-base
+      provider-priority: 10
     pipeline:
       - runs: |
           _java_home="usr/lib/jvm/java-20-openjdk"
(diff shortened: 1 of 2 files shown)


openjdk-20.yaml
```diff
@@ -252,6 +252,7 @@ subpackages:
         - openjdk-20
       provides:
         - default-jdk=1.20
+        - openjdk-20-jre
```
xnox: this needs to be with `=${{package.full-version}}` and also provider-priority needs to be set here to like value of 5, and on the jre subpackage to like value of 10.

See for example the two libcurl builds.

debasishbsws: @xnox Done.

xnox:
@debasishbsws i think tests are failing because our packaging for java 20 is stale; but also java 20 is eol so we are not  building it anymore. can you please instead do this change to openjdk 24?

or alternatively pick up all the missing deps/patches to openjdk 20 that we otherwise have in other LTS supported streams.

debasishbsws:
@xnox I choose openjdk-20 because it is EOL. Yes I can do this for openjdk-24

xnox (approved):

PR #57337, opened 2025-06-23
Feat/6012: Add `mlflow-iamguarded` Subpackage

Illbjorn:
Add `mlflow-iamguarded-compat` subpackage.


mlflow.yaml
```diff
@@ -140,6 +146,77 @@ subpackages:
...
+          # Use Python in virtual environment
+          sed -i "s|/home/build/venv|/usr/share/mlflow|g" venv/pyvenv.cfg
+          sed -i "s|/home/build/venv|/usr/share/mlflow|g" venv/bin/*
+
+          # Install virtual environment
+          mkdir -p ${{targets.contextdir}}/usr/share/mlflow
+          cp -r venv/* ${{targets.contextdir}}/usr/share/mlflow/
+
+          # Create auth symlink expected in iamguarded Helm chart
+          mkdir -p ${{targets.contextdir}}/iamguarded
```
IdlePhysicist: Based on what I've learnt from @wojciechka we don't need/want the `${{targets.contextdir}}` here because the next pipeline handles it.

```suggestion
          mkdir -p /iamguarded
```

mlflow.yaml
```diff
@@ -140,6 +146,77 @@ subpackages:
...
+          PY_VERSION=$(python -c "import sys; print(f'{sys.version_info.major}.{sys.version_info.minor}')")
+          ln -sf "/usr/share/mlflow/lib/python${PY_VERSION}/site-packages/mlflow/server/auth" ${{targets.contextdir}}/iamguarded/mlflow-basic-auth
```
IdlePhysicist: Same would apply here re: `${{targets.contextdir}}`

```suggestion
          ln -sf "/usr/share/mlflow/lib/python${PY_VERSION}/site-packages/mlflow/server/auth" /iamguarded/mlflow-basic-auth
```

IdlePhysicist (changes requested):
Man, that's a chunky compat package.  Looks good. But I called out two uses of `${{targets.contextdir}}` that I don't think are right.

IdlePhysicist (approved):

PR #57365, opened 2025-06-24
dnm: test clickhouse bump

vishal-chdhry:


kranurag7:
context: after switching to qemu it started failing unexpected but it should be good now after 

```yaml
# NOTE(joshrwolf): This is required because for whatever reason,
        # specifying this with melange "paths" isn't working on linux + docker
        # runners as of 06/20/2025
        setup: |
chown -R nonroot:nonroot /home/build
```
there's a thread about doing a follow-up on why this is required with qemu.

kranurag7 (changes requested):
likely https://github.com/wolfi-dev/os/pull/57365/commits/800fcc6e0d1fccd25713dc41748b59e670094c75 is not required because /opt/i and /i is already moved to contextdir by finalize-compat pipeline.

vishal-chdhry:
@kranurag7 This is from bitnami package, it changed in epoch 44, the commit where iamguarded was added
image

vishal-chdhry:
For some reason I cannot build the package locally so debugging in CI

PR #57378, opened 2025-06-24
feat(superset): add iamguarded compat

wojciechka:


kranurag7 (approved):

kranurag7:
superset v5 support will land via some other patch; this patch is to add iamguarded-compat; force merging this one.

PR #57386, opened 2025-06-24
add timescaledb-compat

anushkamittal20:


timescaledb-compat-pg16.yaml
```diff
@@ -0,0 +1,95 @@
+#nolint:valid-pipeline-git-checkout-tag
+package:
+  name: timescaledb-compat-pg16
```
kranurag7, line 3: Is there a reason we're calling the main package as `-compat` if we're cloning the main repository. Can we call it timescaledb-pg<ver>.yaml

timescaledb-compat-pg16.yaml
```diff
@@ -0,0 +1,95 @@
+#nolint:valid-pipeline-git-checkout-tag
```
kranurag7, line 1: we're using tag and expected-commit, we can omit this.

timescaledb-compat-pg16.yaml
```diff
@@ -0,0 +1,95 @@
...
+    with:
+      repository: https://github.com/timescale/timescaledb-docker
+      branch: main
+      expected-commit: d6339b0f6f8cb0c41c4eb2087034557fda7fb2f1
+
+  - runs: |
+      mkdir -p ${{targets.contextdir}}/docker-entrypoint-initdb.d
+      cp docker-entrypoint-initdb.d/* ${{targets.contextdir}}/docker-entrypoint-initdb.d/
+      rm -f $(pg_config --pkglibdir)/timescaledb-tsl-*.so
+      rm -rf ${{targets.contextdir}}/build
```
kranurag7: we should move this to a separate `-oci-entrypoint` package to keep it separate from the main package.

timescaledb-compat-pg16.yaml
anushkamittal20: this is for the second git-checkout, it causes issue with lint

timescaledb-compat-pg16.yaml
anushkamittal20: makes sense, thanks for the review!

timescaledb-compat-pg16.yaml
```diff
@@ -0,0 +1,103 @@
...
+        - postgresql-${{vars.pg-version}}-client
+        - shadow
+        - sudo-rs
+    environment:
+      PGUSER: wolfi
+  pipeline:
+    - name: Check postgres version
+      runs: |
+        postgres --version
+        psql --version
```
tulilirockz:
```suggestion
        postgres --version | grep -F -e "${{vars.pg-version}}"
        psql --version | grep -F -e "${{vars.pg-version}}"
```
Just to make sure the output isnt broken somehow

timescaledb-compat-pg16.yaml
jamonation: If this is the main timescale package, then it should be `timescaledb` and the file `timescaledb.yaml`?

timescaledb-compat-pg16.yaml
```diff
@@ -0,0 +1,103 @@
...
+    - name: Check extension control file
+      runs: |
+        ls $(pg_config --sharedir)/extension/timescaledb.control
+    - name: Create database and load extension
+      runs: |
+        useradd $PGUSER
```
tulilirockz, line 87: I think you can let melange handle user creation/groups instead of creating it here.

```yaml
environment:
     accounts:
         groups:
             - groupname: wolfi
               gid: 65532
         users:
             - username: wolfi
               gid: 65532
               uid: 65532
         run-as: 0 # so you aren't stuck on the unprivileged user
```

timescaledb-compat-pg16.yaml
```diff
@@ -0,0 +1,103 @@
...
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - cmake
+      - postgresql-${{vars.pg-version}}
+      - postgresql-${{vars.pg-version}}-dev
+
+vars:
+  pg-version: 16
```
jamonation, line 29: It looks like timescale supports multiple versions of postgres, so ranging over the set of 14-17, and building using a subpackage for each with an empty 'timescale' package would cover everything. Or is there a difference between the licensed and OSS versions and postgresql support?

tulilirockz, line 29:
```suggestion
var-transforms:
  - from: ${{package.name}}
    match: ^timescaledb-compat-pg-(\d\.\d+)
    replace: $1
    to: postgresMM
```
You might be able to do something like this instead of specifying the postgres version here

timescaledb-compat-pg16.yaml
```diff
@@ -0,0 +1,103 @@
...
+  version: "2.20.3"
+  epoch: 0
+  description: A time-series database for high-performance real-time analytics packaged as a Postgres extension.
+  # Outside of the "tsl" directory, source code in a given file is licensed under the Apache License Version 2.0
+  # We are only building the opensource bits here.
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - git
```
tulilirockz, line 13: Mind putting a little comment here about why is `git` required at runtime?

tulilirockz (commented):
I have a few things that could be applied for both versions, thank you!

timescaledb-compat-pg16.yaml
anushkamittal20: Hey @jamonation I was taking from this [comment](https://github.com/chainguard-dev/image-requests/issues/5510#issuecomment-2711620934):
> Name of the image/container should clearly identify it as compatible with timescale (as opposed to just naming it "timescale) -- the following options are acceptable: "timescale-compat", "for-timescale", or "compat-timescale"

timescaledb-compat-pg16.yaml
anushkamittal20: no you are right, we could do that, we are building 15 in enterprise and 16,17 here so I thought this would be the way to go.

PR #57436, opened 2025-06-24
Fix(Update): The version transform was resulting in 6.5.20250621 wher…

debasishbsws:
…e it should be 6.5-20250621

this change replace the `_p` to `- `which matches the version used by the upstream.

Also package update to 6.5_p20250621

### DIFF
--- a/ncurses.yaml
+++ b/ncurses.yaml
@@ -157,8 +157,8 @@ subpackages:
 update:
   enabled: true
   version-transform:
-    - match: _p
-      replace: \-
+    - match: \-
+      replace: _p
   release-monitor:
     identifier: 2057
 
--- a/ncurses.yaml
+++ b/ncurses.yaml
@@ -1,7 +1,7 @@
 package:
   name: ncurses
-  version: 6.5_p20241228
-  epoch: 3
+  version: 6.5_p20250621
+  epoch: 0
   description: "console display library"
   copyright:
     - license: MIT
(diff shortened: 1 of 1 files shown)


debasishbsws:
```
2025/06/24 22:14:59 INFO checking package ncurses: ncurses
2025/06/24 22:14:59 INFO [1/1] ncurses: checking release monitor using id 2057

2025/06/24 22:15:00 INFO attempting to bump version to 6.5-20250621
2025/06/24 22:15:00 INFO processing fetch node:
2025/06/24 22:15:00 INFO   uri: https://invisible-mirror.net/archives/ncurses/current/ncurses-${{vars.mangled-package-version}}.tgz
2025/06/24 22:15:00 INFO   evaluated: https://invisible-mirror.net/archives/ncurses/current/ncurses-6.5-20250621.tgz
2025/06/24 22:15:04 INFO   expected-sha256: b16ead99b5e2b19528cae5122b82cd0dbb0a2e1c251d3f53302634de992a18bf
2025/06/24 22:15:04 INFO   expected-sha512: 1b665749b822133e7166c21e44e0cf0d94766a7ed4f96a6138de43c46c8ce85ed0c5d5230c050103de3830e7d2fc35c283262d30990b9de30851d1f999525b85
2025/06/24 22:15:05 INFO fetch was successful package=ncurses
```
It works for the update

ncurses.yaml
```diff
@@ -158,9 +158,7 @@ update:
   enabled: true
   version-transform:
     - match: _p
-      replace: .
-    - match: \-
-      replace: .
+      replace: \-
```
OddBloke: I think the match and replace are the wrong way around here: we want to convert a `-` in the upstream version to `_p` in our version.

sergiodj: Sorry, I'm the person who introduced this extra code.  I failed to take into consideration the fact that we deliberately use `_p` in our version (I don't really understand why, though...), so you're correct that we need to convert `-` to `_p` for the comparison to work.

stevebeattie: > I don't really understand why [we embed _p in our version]

Likely to match the versioning used in alpine: https://pkgs.alpinelinux.org/package/edge/main/x86_64/ncurses

debasishbsws: @OddBloke Yup right, This is reverse on what we should do.

OddBloke (approved):

PR #57447, opened 2025-06-24
longhorn-instance-manager v1.9.0 package addition

AnmolVirdi:
**longhorn-instance-manager-1.9** `Current version: v1.9.0`
- Longhorn Instance Manager manages engine and replica instances on the node.

**Type**: Go binary
**REF**: https://github.com/longhorn/longhorn-instance-manager

**Tests:**
- Version tests using both flag and command for server and client
- Functional tests for `longhorn-instance-manager daemon` using `test/daemon-check-output` test pipeline
- Functional tests for `longhorn-instance-manager process`


longhorn-instance-manager-1.9.yaml
```diff
@@ -0,0 +1,107 @@
...
+    with:
+      packages: .
+      output: longhorn-instance-manager
+      tidy: true
+      vendor: true
+      ldflags: |
+        -X main.Version=v${{package.version}}
+        -X main.GitCommit=$(git rev-parse HEAD)
+        -X main.BuildDate=$(date ${SOURCE_DATE_EPOCH:+ -d@${SOURCE_DATE_EPOCH}} "+%Y-%m-%dT%H:%M:%SZ")
+        -linkmode external -extldflags "-static"
```
kranurag7:
```suggestion
        -linkmode external
```
nit: we should remove this in favor of dynamic linking.

longhorn-instance-manager-1.9.yaml
```diff
@@ -0,0 +1,107 @@
...
+    with:
+      deps: |-
+        golang.org/x/oauth2@v0.27.0
+
+  - uses: go/build
+    with:
+      packages: .
+      output: longhorn-instance-manager
+      tidy: true
+      vendor: true
```
kranurag7: are these required here? are builds not happy with the defaults here.

longhorn-instance-manager-1.9.yaml
AnmolVirdi: Noted and updated

longhorn-instance-manager-1.9.yaml
AnmolVirdi: These were needed in older versions due to issues with longhorn dependencies. Fortunately, those versions are no longer active:
https://github.com/longhorn/longhorn/blob/master/README.md#releases

kranurag7 (approved):

PR #57462, opened 2025-06-24
apicurio-registry: update package

mcaveety:
apicurio-registry: update package

Adds subpackage `apicurio-registry-nginx-config` for compatibility with the upstream image. As `nginx` is used in the image build, we have added that to the build and included all necessary configuration files (see the [Dockerfile here for reference](https://github.com/Apicurio/apicurio-registry/blob/589bee401ec46003e46a9cba186109135f0298bf/ui/Dockerfile#L4)). We have also slightly updated the build pipelines to more closely match that of the upstream repository, [see here for build instructions](https://github.com/Apicurio/apicurio-registry/blob/589bee401ec46003e46a9cba186109135f0298bf/ui/deploy-examples/build-container-image.sh#L10).

Please note that the solitary CVE in `quarkus-vertx` cannot be bumped at this time as it results in a build failure.


apicurio-registry.yaml
```diff
@@ -19,9 +19,11 @@ environment:
       - go
       - icu
       - maven
+      - nginx
       - nodejs-20
       - npm
       - openjdk-17-default-jdk
+      - openssl
```
tulilirockz:
```suggestion
      - nodejs-20
      - npm
      - openjdk-17-default-jdk
```
`nginx` isnt being used anywhere here and `openssl` is unnecessary (see the other review below)

apicurio-registry.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: apicurio-registry
   version: "3.0.9"
-  epoch: 3
+  epoch: 5
```
tulilirockz:
```suggestion
  epoch: 4
```

apicurio-registry.yaml
```diff
@@ -71,12 +73,35 @@ subpackages:
...
+    dependencies:
+      runtime:
+        - docker-nginx
+    pipeline:
+      - working-directory: ui
+        runs: |
+          # Configure certificate and key (per upstream)
+          mkdir -p /etc/pki/nginx/private
+          openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout /etc/pki/nginx/private/server.key -out /etc/pki/nginx/server.crt --batch
+          chown -R 1001:0 /etc/pki/nginx/ && chmod 755 /etc/pki/nginx/private/server.key /etc/pki/nginx/server.crt
```
tulilirockz:
```suggestion
```
If its working as-is right now we probably don't need this. This is just being added to the build rootfs instead of the package itself (`${{targets.subpkgdir}}`)

apicurio-registry.yaml
```diff
@@ -71,12 +73,35 @@ subpackages:
...
+
+          # Copy configuration scripts (per upstream)
+          config_dest="${{targets.contextdir}}/usr/local/bin/"
+          install -Dm644 -o 1001 -g 0 .docker-scripts/create-config.cjs "${config_dest}/create-config.cjs"
+          install -Dm644 -o 1001 -g 0 .docker-scripts/update-base-href.cjs "${config_dest}/update-base-href.cjs"
+          install -Dm644 -o 1001 -g 0 .docker-scripts/entrypoint.sh "${config_dest}/entrypoint.sh"
+          install -Dm644 -o 1001 -g 0 .docker-scripts/nginx.conf "${config_dest}/etc/nginx/nginx.conf"
```
tulilirockz:
```suggestion
          config_dest="${{targets.contextdir}}/usr/local/bin/"
          install -Dm644 .docker-scripts/create-config.cjs "${config_dest}/create-config.cjs"
          install -Dm644 .docker-scripts/update-base-href.cjs "${config_dest}/update-base-href.cjs"
          install -Dm644 .docker-scripts/entrypoint.sh "${config_dest}/entrypoint.sh"
          install -Dm644 .docker-scripts/nginx.conf "${config_dest}/etc/nginx/nginx.conf"
```
I _feel_ like you can omit these, does it break anything?

mcaveety: Not at the moment, but I was doing this as per upstream's configuration

tulilirockz (approved):
We should probably file in an advisory for the CVE issue if we haven't yet!

PR #57496, opened 2025-06-25
Remove sonarqube-10 yaml and withdrawal the packages as it is confusing

debasishbsws:


debasishbsws:
As we build only the latest sonarqube and not the sonarqube-10 we can remove the package entirely

philroche (changes requested):
I don't think we should withdraw. When we do, we are regressing current users of the `10` tag - however confusing.

philroche (approved):

PR #57527, opened 2025-06-25
llvm-16.yaml: Consolidate into unified llvm-16.yaml and remove split …

AbduAzim424:
This PR simplifies the LLVM 16 packaging by consolidating all component YAMLs into a single unified build.

Used the highest epoch from the subpackage as the base package epoch to prevent the build stage from pulling incorrect packages.


llvm-16.yaml
```diff
@@ -0,0 +1,799 @@
...
+            clang-pseudo --version
+            clang-pseudo --help
+
+  - name: lld-${{vars.major-version}}
+    description: LLVM linker
+    dependencies:
+      provides:
+        - llvm-lld-${{vars.major-version}}=${{package.full-version}}
+      runtime:
+        - libLLVM-16=${{package.full-version}}
```
javacruft:
```
  - libLLVM-${{vars.major-version}}=${{package.full-version}}
```

javacruft (approved):

PR #57528, opened 2025-06-25
llvm-15.yaml: Consolidate into unified llvm-15.yaml and remove split …

AbduAzim424:
This PR simplifies the LLVM 15 packaging by consolidating all component YAMLs into a single unified build.

Used the highest epoch from the subpackage as the base package epoch to prevent the build stage from pulling incorrect packages.


AbduAzim424:
@javacruft Hi, can you please take a look at this

llvm-15.yaml
```diff
@@ -0,0 +1,754 @@
+package:
+  name: llvm-15
+  version: 15.0.7
+  epoch: 10
```
javacruft: You'll need to use

```
  epoch: 11
```

in order to ensure this new package version is higher than all of the ones being withdrawn/replaced - clang-15 is currently epoch 10

llvm-15.yaml
```diff
@@ -0,0 +1,754 @@
...
+            clang-linker-wrapper --version
+            clang-linker-wrapper --help
+
+  - name: lld-${{vars.major-version}}
+    description: LLVM linker
+    dependencies:
+      provides:
+        - llvm-lld-${{vars.major-version}}=${{package.full-version}}
+      runtime:
+        - libLLVM-15=${{package.full-version}}
```
javacruft: nit:

```
libLLVM-${{vars.major-version}}=${{package.full-version}}
```

javacruft (changes requested):

javacruft (approved):

PR #57541, opened 2025-06-25
llvm-17.yaml: Consolidate into unified llvm-17.yaml and remove split …

AbduAzim424:
This PR simplifies the LLVM 17 packaging by consolidating all component YAMLs into a single unified build.

Issue description:
- Building LLVM/Clang 17.0.6 with GCC 14 on ARM64/AArch64 systems fails with preprocessor token pasting errors: `error: pasting "kw_" and "[" does not give a valid preprocessing token`.
- GCC 14 automatically defines ARM-specific built-in macros that expand to C++ attribute syntax: `__arm_streaming` -> '[[arm::streaming]]'
- When LLVM's token generation macros attempt to concatenate `kw_` with these identifiers, GCC expands the macro first, resulting in invalid tokens like `kw_[[arm::streaming]]`.
- This was fixed by using the older compiler, `gcc-13`.

The issue has been fixed in more recent LLVM versions, but the patch wasn't backported to LLVM 17.0.6 (https://github.com/llvm/llvm-project/issues/78691).


llvm-17.yaml
```diff
@@ -58,6 +58,8 @@ pipeline:
         -DCMAKE_BUILD_TYPE=Release \
         -DCMAKE_INSTALL_PREFIX=${{vars.llvm-prefix}} \
         -DCMAKE_PREFIX_PATH=${{vars.llvm-prefix}} \
+        -DCMAKE_CXX_FLAGS="-U__arm_streaming" \
+        -DCMAKE_C_FLAGS="-U__arm_streaming" \
```
dannf: Mind adding a comment about this, with a description of the issue and/or a bug link? Something so that a person coming by later will know how to determine if this is still needed.

AbduAzim424: I added the issue reference into package definition and gave explanation in PR description.

dannf: Thanks! Can you elaborate (or provide pointers to) how you determined that undefining `__arm_streaming` is the correct solution? Did you see a discussion somewhere that suggested this approach? I'm also nervous about overriding these CMAKE variables. Do we know that these are just appending flags vs. changing ones that would otherwise be generated?

I do see that a fix was merged upstream in https://github.com/llvm/llvm-project/pull/78704. So I looked to see when that went in :

```
dann-frazier@dann-frazier-laptop:~/git/llvm-project$ git log --grep "Fix build with GCC 14 on ARM" --oneline
d54dfdd1b53f [Clang] Fix build with GCC 14 on ARM (#78704)
dann-frazier@dann-frazier-laptop:~/git/llvm-project$ git tag --contains d54dfdd1b53f | head
llvmorg-18.1.0
llvmorg-18.1.0-rc1
llvmorg-18.1.0-rc2
llvmorg-18.1.0-rc3
llvmorg-18.1.0-rc4
llvmorg-18.1.1
llvmorg-18.1.2
llvmorg-18.1.3
llvmorg-18.1.4
llvmorg-18.1.5
```

Is it maybe already backported to 17, just not yet released?

```
HEAD is now at 6009708b4367 Revert "[runtimes] Add missing test dependencies to check-all (#72955)"
dann-frazier@dann-frazier-laptop:~/git/llvm-project$ git log --grep "Fix build with GCC 14 on ARM" --oneline
dann-frazier@dann-frazier-laptop:~/git/llvm-project$ 
```

No such luck.

And, as you mentioned, backporting the upstream patch ourselves is non-trivial, as it has a lot of dependencies.

Now, alpine appears to have gone with this patch:
https://github.com/alpinelinux/aports/commit/837424684bc037a76e4bb2f3dab943ef8a166652

And eulerOS with this one, which at a quick glance, looks to be the same:
https://github.com/openeuler-mirror/llvm-project/commit/505323d49f4621e5f7210d99fd52dd33a6223fa8

I'd suggest going with the well trod path here.

btw, the fact that we switched from gcc-13 to a later compiler isn't clear from the git logs here. My suggestion would be to make that switch (and the patch that makes it possible) in a separate commit before or after the reorganization. So if someone is looking through the logs to diagnose a regression, they can clearly see that we made a functional change here.

Another option would of course be just going back to gcc-13 - assuming gcc-13's upstream support life is >= llvm-17's.

dannf (changes requested):

AbduAzim424:
Thank you for the review. I came up with this as a hypothetical fix at the moment. From my understanding, this was just a compile time macro collision. I assumed that this change will not eliminate `clang` ARM streaming capabilities but that could lead to errors if LLVM depends on GCC's definition of `__arm_streaming` for something later on. So that's really risky indeed. I ported the patch from `eulerOS` and will see if it works. Regarding the GCC 13 change, `llvm17` wasn't using GCC 13, it had `gcc-14-default` so I didn't touch it, but the `clang-17` actually was, but that change was implicit in the definition. If the patch thing works, I can add a note near the GCC definition mentioning compiler bump for the `clang` and emphasize that in a separate commit.

AbduAzim424:
@javacruft Can you please take a look at this

murraybd:
> @javacruft Can you please take a look at this

If I recall correctly the `count --help` test was causing the build and test phase to take more than 6 hours in the production testing environment. Is that right @AbduAzim424 ?

AbduAzim424:
From my last rebuild on `x86_64` `count --version` printed nothing and `count --help` froze indefinitely causing `test failure` via timeout

llvm-17.yaml
```diff
@@ -0,0 +1,809 @@
+package:
+  name: llvm-17
+  version: 17.0.6
+  epoch: 10
```
javacruft: You are burning quite a few epochs here - 6 is sufficient (clang-17 has the highest one at 5 previously)

javacruft:
Re the command that hangs - do you see this behaviour with no patches applied? I'm concerned we've slightly broken something in clang - some earlier attempts to perform this consolidation took a different approach - https://github.com/wolfi-dev/os/pull/39406/files#diff-693fae19ab1adbbaf9014820ae3944cdcc5a425ba21e2cbcf6ec3c9f77ccc7cb - and switched the gcc compiler for arm64 only - we might consider this as an alternative so we don't have to carry the patch.

javacruft (dismissed):

dannf:
> Re the command that hangs - do you see this behaviour with no patches applied? I'm concerned we've slightly broken something in clang - some earlier attempts to perform this consolidation took a different approach - https://github.com/wolfi-dev/os/pull/39406/files#diff-693fae19ab1adbbaf9014820ae3944cdcc5a425ba21e2cbcf6ec3c9f77ccc7cb - and switched the gcc compiler for arm64 only - we might consider this as an alternative so we don't have to carry the patch.

ftr, I root caused this to being a change in the build environment, where we now use `script` for tests which does not close stdin.

llvm-17.yaml
```diff
@@ -0,0 +1,819 @@
...
+            nvptx-arch --version
+            nvptx-arch --help
+
+  - name: lld-${{vars.major-version}}
+    description: LLVM linker
+    dependencies:
+      provides:
+        - llvm-lld-${{vars.major-version}}=${{package.full-version}}
+      runtime:
+        - libLLVM-17=${{package.full-version}}
```
dannf: *nit* in llvm-16 we use ${{vars.major-version}} - feels like we should do the same here

dannf (dismissed):
I compared the diff w/ llvm-16, and no surprises. I see how you've done the gcc-13 if arm64 thing, and that seems reasonable. One nit inline but, other than that, :+1:

dannf (approved):
```diff
diff --git a/llvm-17.yaml b/llvm-17.yaml
index e1557e1fd5..02cd780cba 100644
--- a/llvm-17.yaml
+++ b/llvm-17.yaml
@@ -435,7 +435,7 @@ subpackages:
       provides:
         - llvm-lld-${{vars.major-version}}=${{package.full-version}}
       runtime:
-        - libLLVM-${{vars.major-version}}=${{package.full-version}}
+        - libLLVM-17=${{package.full-version}}
     pipeline:
       - runs: |
           mkdir -p ${{targets.contextdir}}/usr/bin
```
LGTM :)

PR #57573, opened 2025-06-25
libgdiplus: Bump version, update upstream location

sergiodj:
Mono upstream changed hands to WineHQ (https://www.mono-project.com/).  Update libgdiplus to use the new upstream, bump its version, and fix the build for the upcoming CMake 4.0 package.

### DIFF
--- a/libgdiplus.yaml
+++ b/libgdiplus.yaml
@@ -1,7 +1,7 @@
 package:
   name: libgdiplus
-  version: 6.1
-  epoch: 3
+  version: 6.2
+  epoch: 0
   description: "Open Source Implementation of the GDI+ API"
   copyright:
     - license: MIT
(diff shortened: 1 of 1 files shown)


libgdiplus.yaml
```diff
@@ -67,10 +69,7 @@ subpackages:
 
 update:
   enabled: true
-  github:
-    identifier: mono/libgdiplus
-    use-tag: true
-    tag-filter: 6.
+  git: {}
```
dannf, line 72: is there nothing to track here?

sergiodj: There's no need to adjust any of the default settings.  Just using `git: {}` works fine.

sergiodj: `git:` already checks for tags (which is what we want), and upstream doesn't prefix their tags with anything: it's just the regular package version.

dannf: oh, ok. and the `update:` section doesn't need a repo to monitor? that seems odd, because even in the github case we need to put the repo in both the git pipeline and the update section.

justinvreeland: It pulls it from the checkout pipeline. I thought it was strange the first time I saw it too.

https://github.com/chainguard-dev/melange/blob/main/docs/UPDATE.md#git

dannf: cool, thx @justinvreeland !

dannf (approved):
cmake it so.

PR #57574, opened 2025-06-25
libical: Upgrade to latest version; enable auto-update

sergiodj:
For some reason auto-update was disabled for libical.  Enable it, and update the package to the latest upstream version.  This also fixes a build failure when using CMake 4.0.


dannf (approved):
crazy man, another killer extra!

PR #57689, opened 2025-06-26
kubeflow-jupyter-web-app/cve remediation

kbsteere:
GHSA-pq67-6m6q-mj2v fix


powersj (changes requested):
Need an epoch bump, also if you are commenting it out, can't we just remove? ;)

powersj (approved):

PR #57695, opened 2025-06-26
build grafana-image-renderer for arm64

sanjayk0508:
Given that chromium now supports arm64 ([ref](https://github.com/wolfi-dev/os/pull/44107)), we can now build `grafana-image-renderer` for arm64

also updates version


sanjayk0508:
updating the version manually fails in CI

jamonation (dismissed):

jamonation:
Looks like the `GHSA-rhx6-c78j-4q9w.patch` patch causes the bumped version to fail. Wonder if removing that will get a newer version going?

grafana-image-renderer.yaml
```diff
@@ -1,12 +1,10 @@
 package:
   name: grafana-image-renderer
-  version: "3.12.7"
-  epoch: 0
+  version: "3.12.8"
+  epoch: 1
```
kranurag7:
```suggestion
  epoch: 0
```
nit and non-blocker: for new package updates; this is zero.

kranurag7 (dismissed):

grafana-image-renderer.yaml
sanjayk0508: we have also made a patch by removing the `target-architecture`, so i think we might need to bump the epoch, LMK me if i'm missing something

jamonation: The new version will take precedence over an older epoch, e.g.:
3.12.8-r0 wins over 3.12.7-rN

kranurag7 (approved):

grafana-image-renderer.yaml
sanjayk0508: ah, got it, thank you for the clarification @jamonation

PR #57722, opened 2025-06-27
dataplaneapi: migrate to version streams

kranurag7:
this is used together with haproxy and haproxy <major>.<minor> imports dataplaneapi <major.minor> so this commit migrates dataplaneapi to version streams.


kranurag7:
closing given it doesn't matter mostly, we're good with using latest all the time basically for now.

PR #57723, opened 2025-06-27
nginx-stable: use stable versions correctly

kranurag7:
nginx-stable: use stable versions correctly

    nginx project stable versions are always evenly tagged, it should never
    be odd unless upstream changes something.

    this commit adds a check at build time to ensure that we always use even
    tags for stable nginx builds.

### DIFF
--- a/nginx-stable.yaml
+++ b/nginx-stable.yaml
@@ -40,7 +40,7 @@ environment:
 
 var-transforms:
   - from: ${{package.version}}
-    match: '^(\d+)\.(\d+)\.'
+    match: '^(\d+)\.(\d+)\.(\d+)'
     replace: '$2'
     to: minor-version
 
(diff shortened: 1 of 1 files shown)


nginx-stable.yaml
```diff
@@ -51,16 +58,23 @@ data:
 # TODO look at adding extra modules like alpine does https://git.alpinelinux.org/aports/tree/main/nginx/APKBUILD#n149
 # Also note the nginx inc config https://hg.nginx.org/pkg-oss/file/tip/alpine
 pipeline:
+  - name: validate-stable-version
```
kwmonroe, line 61: nice!

kwmonroe (approved):
lgtm, thanks for the test to make sure we don't end up here in the near future!

PR #57732, opened 2025-06-27
icu 77.1 update

javacruft:
Introduce icu 77.1 using the proposed new approach to shared library package naming.

This ensures that both icu-libs and libicu77 can be installed at the same time, allowing for a smooth transition of all reverse dependencies to the new version.

Add python3 to build environment; used to generate data files during the build.

Switch back to using git-checkout for upstream source.

Reverse build dependencies will be covered under separate PR's; install-ability should be maintained throughout


icu.yaml
```diff
@@ -124,18 +125,13 @@ subpackages:
         - merged-usrsbin
         - wolfi-baselayout
 
-# switch back to this once the GitHub release stops saying pre-release
-# github:
-#   identifier: unicode-org/icu
-#   strip-prefix: release-
 update:
   enabled: true
   manual: true # ICU updates contain ABI breaking changes which require manual intervention
```
murraybd: Do we want this package to continue to have `manual: true` in the update section?

javacruft: I think longer term no; but we need automation to perform the actual transitions when we get a new version.

murraybd (approved):

PR #57741, opened 2025-06-27
py3-pip/cve-remediation

kbsteere:
Upgrade requests to 2.32.4


py3-pip.yaml
```diff
@@ -35,7 +35,7 @@ pipeline:
 
   - uses: patch
     with:
-      patches: fix-CVE-2024-47081.patch
```
aborrero, line 38: maybe we can drop the file `py3-pip/fix-CVE-2024-47081.patch` in this same PR, otherwise it seems it be dead code.

py3-pip/upgrade-requests-to-2.32.4.patch
```diff
@@ -0,0 +1,215 @@
+From 036b1e0f4271ebb6068a75fd0869d7eecd48162c Mon Sep 17 00:00:00 2001
```
aborrero, line 1: Would it be possible to include the URL for this patch in the header of this file?

aborrero (commented):
some comments inline.

py3-pip.yaml
kbsteere: Yeah your correct. Just removed it.

py3-pip/upgrade-requests-to-2.32.4.patch
kbsteere: I could add the mainline upstream that I based this patch on but it is still different since that was mainline and this is patch fixes specifically 25.1.1 tag: https://github.com/pypa/pip/commit/205977a791b1e5da7a20fa01effe5a4f035f6a9d

kbsteere:
advisory pr: https://github.com/wolfi-dev/advisories/pull/20565

kwmonroe (approved):

PR #57765, opened 2025-06-27
keycloak: bump epoch

aborrero:
To remediate CVE-2025-49574 GHSA-9623-mj7j-p9v4.

See also: https://github.com/chainguard-dev/CVE-Dashboard/issues/25410


aborrero:
the epoch bump alone seems to not be enough to remediate the vuln :-( Trying something else.

PR #57767, opened 2025-06-27
python-3.12: backport patch to fix CVE-2025-6069

OddBloke:


OddBloke:
The CVE scan is failing because it can't detect that we've backported the patch: a manual advisory update will be required.

powersj (approved):

PR #57933, opened 2025-06-28
nginx-(mainline|stable): use git backend for update

kranurag7:
nginx-(mainline|stable): use git backend for update

    this will ensure that our nginx are upto date.

    as of now tags are stale and because of filter when series of nginx
    comes out they don't get updated unless we udpate them manully.

    this commit changes the update backend to use git.
    and then we're ignoring odd versions during update for nginx-stable and
    vice-versa for mainline.

### DIFF
--- a/nginx-mainline.yaml
+++ b/nginx-mainline.yaml
@@ -338,9 +338,9 @@ test:
 
 update:
   enabled: true
-  github:
-    identifier: nginx/nginx
-    use-tag: true
-    # MAINLINE VERSIONS MUST USE ODD MINOR VERSIONS (e.g., 1.27.x, 1.29.x)
-    tag-filter: release-1.29
+  # MAINLINE VERSIONS MUST USE ODD MINOR VERSIONS (e.g., 1.27.x, 1.29.x)
+  ignore-regex-patterns:
+    - ^(\d+)\.(\d+)[02468]\.(\d+)$
+  git:
+    tag-filter-prefix: release-
     strip-prefix: release-
--- a/nginx-stable.yaml
+++ b/nginx-stable.yaml
@@ -283,9 +283,9 @@ test:
 
 update:
   enabled: true
-  github:
-    identifier: nginx/nginx
-    use-tag: true
-    # STABLE VERSIONS MUST USE EVEN MINOR VERSIONS (e.g., 1.26.x, 1.28.x)
-    tag-filter: release-1.28
+  # STABLE VERSIONS MUST USE EVEN MINOR VERSIONS (e.g., 1.26.x, 1.28.x)
+  ignore-regex-patterns:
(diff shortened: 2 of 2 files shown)


debasishbsws (approved):
I was confused for ones, why it is matching the reverse.

nginx-mainline.yaml
```diff
@@ -338,9 +338,9 @@ test:
 
 update:
   enabled: true
-  github:
-    identifier: nginx/nginx
-    use-tag: true
-    # MAINLINE VERSIONS MUST USE ODD MINOR VERSIONS (e.g., 1.27.x, 1.29.x)
-    tag-filter: release-1.29
+  # MAINLINE VERSIONS MUST USE ODD MINOR VERSIONS (e.g., 1.27.x, 1.29.x)
```
bentasker, line 341:
```suggestion
  # MAINLINE VERSIONS MUST USE ODD MINOR VERSIONS (e.g., 1.27.x, 1.29.x)
  # SO IGNORE ANY TAG THAT IS NOT ODD
```

nginx-stable.yaml
```diff
@@ -283,9 +283,9 @@ test:
 
 update:
   enabled: true
-  github:
-    identifier: nginx/nginx
-    use-tag: true
-    # STABLE VERSIONS MUST USE EVEN MINOR VERSIONS (e.g., 1.26.x, 1.28.x)
-    tag-filter: release-1.28
+  # STABLE VERSIONS MUST USE EVEN MINOR VERSIONS (e.g., 1.26.x, 1.28.x)
```
bentasker, line 286:
```suggestion
  # STABLE VERSIONS MUST USE EVEN MINOR VERSIONS (e.g., 1.26.x, 1.28.x)
    # SO IGNORE ANY TAG THAT IS NOT EVEN
```

nginx-mainline.yaml
bentasker: I know the next line uses `ignore-regex-patterns` but my eye/brain skipped over that and started wondering why the regex contains even numbers when the comment mentions odd ones.

nginx-mainline.yaml
```diff
@@ -338,9 +338,9 @@ test:
...
+  ignore-regex-patterns:
+    - ^(\d+)\.(\d+)[02468]\.(\d+)$
```
bentasker, line 343: Do we want to make the first digit of the minor version optional?

It looks like what you're doing here is ensuring that we'll match versions where the tens unit isn't even (for example 1.32) which is great.

But, if they ever release (say) a 2.2 this will fail - I think I'm less concerned about the likelihood of that than as someone taking a copy of the regex for use against a different package

I think we _should_ be able to use `^(\d+)\.((\d+)?)[02468]\.(\d+)$`

bentasker (commented):
Have noted a couple of possible changes

PR #57973, opened 2025-06-30
feat: add kubescape-operator package

yashpimple:
The PR added new package name kubescape-operator with functional tests

### DIFF
--- a/kubescape-operator.yaml
+++ b/kubescape-operator.yaml
@@ -35,73 +35,54 @@ test:
     - runs: |
         operator --help
     - uses: test/kwok/cluster
-    - runs: |
-        kubectl config view --minify --raw > /tmp/kubeconfig.yaml
+    - uses: test/daemon-check-output
+      with:
+        setup: |
+          kubectl config view --minify --raw > /tmp/kubeconfig.yaml
 
-        # Create required configuration directories
-        mkdir -p /etc/config
-        mkdir -p /etc/credentials
+          # Create required configuration directories
+          mkdir -p /etc/config /etc/credentials
+          echo "test-access-key" > /etc/credentials/access-key
+          echo "test-account-id" > /etc/credentials/account
 
-        # Create credentials files (required by operator)
-        cat > /etc/credentials/access-key << 'EOF'
-        test-access-key-dummy
-        EOF
-
-        cat > /etc/credentials/account << 'EOF'
-        test-account-id-dummy
-        EOF
-
-        cat > /etc/config/clusterData.json << 'EOF'
-        {
-          "gatewayWebsocketURL": "wss://ggs.kubescape.io",
-          "gatewayRestURL": "https://ggs.kubescape.io",
-          "kubescapeURL": "https://ggs.kubescape.io",
-          "accountID": "test-account",
-          "clusterName": "test-cluster"
-        }
-        EOF
-
-        cat > /etc/config/capabilities << 'EOF'
-        {
-          "capabilities": {
-            "scanImages": true,
-            "scanWorkloads": true,
-            "scanCronJobs": true,
-            "scanNetworkPolicies": true,
-            "scanRegistries": true,
-            "relevantImageScan": true,
-            "vulnerabilityScanning": true,
-            "configurationScanning": true,
-            "runtimeDetection": true,
-            "runtimeMitigation": false,
-            "continuousScanning": true
+          # Creating STANDALONE configuration with empty service URLs
+          cat > /etc/config/clusterData.json << 'EOF'
+          {
+            "kubevulnURL": "",
+            "kubescapeURL": "",
+            "accountID": "test-account-id",
+            "clusterName": "test-cluster"
           }
-        }
-        EOF
+          EOF
 
-        cat > /etc/config/config << 'EOF'
-        {
-          "server": {
-            "host": "0.0.0.0",
-            "port": 8080
-          },
-        "logger": {
-            "level": "info",
-            "format": "json"
-          },
-          "scanning": {
-            "enabled": true,
-            "interval": "24h"
-          },
-          "metrics": {
-            "enabled": true,
-            "port": 8081
+          # Create minimal capabilities for standalone mode
+          cat > /etc/config/capabilities << 'EOF'
+          {
+            "capabilities": {
+              "configurationScan": "enable",
+              "vulnerabilityScan": "disable",
+              "nodeScan": "disable",
+              "continuousScan": "disable",
+              "runtimeObservability": "disable"
+            }
           }
-        }
-        EOF
+          EOF
 
-        kubectl create namespace kubescape
-        operator -kubeconfig /tmp/kubeconfig.yaml 2>&1 | head -20 > /tmp/operator-output.log
-        grep -q "credentials loaded" /tmp/operator-output.log
-        grep -q "Image version" /tmp/operator-output.log
-        echo "Operator started successfully"
+          # Create minimal config for standalone
+          cat > /etc/config/config << 'EOF'
+          {
+            "namespace": "kubescape",
+            "port": "4002",
+            "triggersecurityframework": false,
+            "workerconcurrency": 1
+          }
+          EOF
+
+          kubectl create namespace kubescape
+        start: operator -kubeconfig /tmp/kubeconfig.yaml
+        timeout: 30
+        expected_output: |
+          Image version
+          credentials loaded
+        post: |
+          echo "Operator started successfully"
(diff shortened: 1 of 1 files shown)


kubescape-operator.yaml
```diff
@@ -0,0 +1,113 @@
...
+    with:
+      deps: |-
+        github.com/go-viper/mapstructure/v2@v2.3.0
+        github.com/inspektor-gadget/inspektor-gadget@v0.40.0
+
+  - uses: go/build
+    with:
+      packages: .
+      output: operator
+      ldflags: -s -w
```
kranurag7:
```suggestion

```
this is not required; the defaults are good.

kubescape-operator.yaml
```diff
@@ -0,0 +1,113 @@
...
+
+        kubectl create namespace kubescape
+        operator -kubeconfig /tmp/kubeconfig.yaml 2>&1 | head -20 > /tmp/operator-output.log
+        if grep -q "credentials loaded" /tmp/operator-output.log; then
+          echo "Operator successfully loaded credentials"
+        fi
+
+        if grep -q "Image version" /tmp/operator-output.log; then
+          echo "Operator started successfully"
+        fi
```
kranurag7: let's directly use `grep "foo" /tmp/operator-output.log` and avoid wrapping this inside a if statement. If `grep` is not able to find the logs then it'll give us `exit 1`

kubescape-operator.yaml
```diff
@@ -0,0 +1,108 @@
...
+  enabled: true
+  github:
+    identifier: kubescape/operator
+    strip-prefix: v
+
+test:
+  pipeline:
+    - runs: |
+        operator --help
+    - uses: test/kwok/cluster
```
kranurag7, line 44:
```suggestion

```
nit: duplicate

kubescape-operator.yaml
```diff
@@ -0,0 +1,108 @@
...
+            "level": "info",
+            "format": "json"
+          },
+          "scanning": {
+            "enabled": true,
+            "interval": "24h"
+          },
+          "metrics": {
+            "enabled": true,
+            "port": 8081
```
kranurag7: are we able to get some metrics out from 8081 and some activity with 8080 maybe healthchecks if yes then that would be awesome tests to add because that'll validate that operator is responding.

yashpimple: @kranurag7 unfortunately, i am not able implement curl command at package level because kubescape-operator requires actual Kubernetes nodes with specific metadata and depends on other Kubescape services (kubescape:8080, kubevuln:8081) that need to be running and port-forwarded. Our KWOK test environment lacks these dependencies, causing the operator to crash when trying to access node metadata. But I can add  metrics and health endpoint testing at the image level where we can deploy the complete Kubescape ecosystem with proper service dependencies.

kubescape-operator.yaml
```diff
@@ -0,0 +1,108 @@
...
+    identifier: kubescape/operator
+    strip-prefix: v
+
+test:
+  pipeline:
+    - runs: |
+        operator --help
+    - uses: test/kwok/cluster
+    - uses: test/kwok/cluster
+    - runs: |
```
Dentrax: Can we use the `test/daemon-check-output` pipeline with minimal config provided? Tests does seem a bit noisy and not sure what we are testing here.

Are those configs needed at package-level? Why we are creating those manually?

Dentrax (changes requested):

kubescape-operator.yaml
yashpimple: @Dentrax to test the operator, we need to provide some required configuration files (e.g. clusterData.json, capabilities, and credential files), as the operator expects them at startup under `/etc/config` and `/etc/credentials` without these, the operator fails early with fatal errors like:
```
load clusterData error: Config File "clusterData.json" Not Found in [/etc/config]
```
I made few changes to run the operator as a standalone component using `test/daemon-check-output`

kubescape-operator.yaml
yashpimple: @kranurag7 made few changes to run the operator as a standalone components limited the crash caused by missing node metadata in the KWOK test environment.

kranurag7 (dismissed):

kubescape-operator.yaml
```diff
@@ -0,0 +1,88 @@
...
+update:
+  enabled: true
+  github:
+    identifier: kubescape/operator
+    strip-prefix: v
+
+test:
+  pipeline:
+    - runs: |
+        operator --help
```
tulilirockz:
```suggestion
        set -o pipefail
        operator --help | grep -F -e "(something from the output)"
```
Whenever you are doing --help checks, please do this so you can make sure the output is not being garbled somehow (this happened on the QEMU package and we didnt catch it)

kubescape-operator.yaml
```diff
@@ -0,0 +1,88 @@
...
+            "triggersecurityframework": false,
+            "workerconcurrency": 1
+          }
+          EOF
+
+          kubectl create namespace kubescape
+        start: operator -kubeconfig /tmp/kubeconfig.yaml
+        timeout: 30
+        expected_output: |
+          Image version
```
tulilirockz: `Image version`? Does it print out the package version? If so, would be pretty cool to use `${{package.version}}` here. (if not its ok too)

yashpimple: @tulilirockz he operator only prints "Image version" without the actual package version, so using ${{package.version}} wouldn't reflect the real output.

kubescape-operator.yaml
yashpimple: Hi @tulilirockz, thanks for the suggestion! I tried implementing it, but the test fails because the kubescape-operator has very limited help output.

Here's what it currently prints:
```
Usage of operator:
  -kubeconfig string
        Paths to a kubeconfig. Only required if out-of-cluster.
```
since --help only lists this one flag, and the output is minimal, the grep works correctly but the test still exits with status 1. So for now, I’m keeping the original check to avoid unnecessary test failures.

kubescape-operator.yaml
```diff
@@ -0,0 +1,88 @@
...
+          credentials loaded
+        post: |
+          echo "Operator started successfully"
```
Dentrax: this should be placed under `expected_output`. and lets have `wait-for-it` here to test port, and simple curl request if possible

kubescape-operator.yaml
tulilirockz: I managed to make it work locally by doing this:
```shell
set -o pipefail
operator --help 2>&1 | grep -F -e "Usage of operator"
```
It was failing because `grep` couldn't find any output on `/dev/stdout`, instead, `operator` was outputting to `/dev/stderr` :)

kubescape-operator.yaml
yashpimple: @Dentrax the upstream operator has an issue where it fails to detect the kubeconfig path during startup, even when correctly set in tests. This caused earlier failures, so we moved the validation logic to the post section instead.

Dentrax: If there is a testing issue, we should comment here why we can't have post-step validation here. (a comment in the `post:`) The app panics in its current state, and we wouldn’t have known exactly when it happened—whether it occurred during the build process or afterward.

```
2025/07/02 13:44:22 INFO > [info] Image version: 
2025/07/02 13:44:22 INFO > [info] credentials loaded. accessKeyLength: 0; accountLength: 15
2025/07/02 13:44:22 INFO > [error] error getting cloud metadata. error: failed to get node : resource name may not be empty
2025/07/02 13:44:22 INFO > panic: runtime error: invalid memory address or nil pointer dereference
2025/07/02 13:44:22 INFO > [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x296a458]
2025/07/02 13:44:22 INFO > 
2025/07/02 13:44:22 INFO > goroutine 1 [running]:
2025/07/02 13:44:22 INFO > main.main()
2025/07/02 13:44:22 INFO >     github.com/kubescape/operator/main.go:132 +0x1468
```

Dentrax (dismissed):

kubescape-operator.yaml
yashpimple: added

Dentrax: https://github.com/wolfi-dev/os/pull/57973/commits/fb810b1955ed2f87628f9da61a12cf521e4c0327

developer-guy (approved):
Looks correct to me! I’ve checked the image layers and didn’t see anything other than the binary, so this package meets all the container image requirements. I'm approving it.

PR #57976, opened 2025-06-30
Mariadb operator

rahulduvedi:
mariadb-0.38.1: new package


mariadb-operator.yaml
```diff
@@ -0,0 +1,165 @@
...
+
+  - uses: go/build
+    with:
+      modroot: .
+      packages: ./cmd/backup
+      output: mariadb-operator-backup
+
+  - uses: go/build
+    with:
+      modroot: .
```
kranurag7:
```suggestion

```
this is the default; we can omit this.

mariadb-operator.yaml
```diff
@@ -0,0 +1,165 @@
...
+      packages: ./cmd/init
+      output: mariadb-operator-init
+
+  - uses: strip
```
kranurag7:
```suggestion

```
not required given the go/build pipeline does the necessary stripping for us.

mariadb-operator.yaml
```diff
@@ -0,0 +1,165 @@
...
+
+subpackages:
```
kranurag7: If you want these subpackages then let's create a `data:` block; range over it and do a `go/build` there.

mariadb-operator.yaml
```diff
@@ -0,0 +1,165 @@
...
+  epoch: 0
+  description: A Go operator that automates the deployment and management of MariaDB databases on Kubernetes
+  copyright:
+    - license: MIT
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/mariadb-operator/mariadb-operator.git
+      tag: v${{package.version}}
```
kranurag7, line 13: the tags there are mixed; there's also one `mariadb-operator-vx.y.z` please validate if this is the actual one.

mariadb-operator.yaml
rahulduvedi: I realised while testing that these are subcommands and not sub-packages, Removed

mariadb-operator.yaml
rahulduvedi: Verified we can safely one with prefix`v`

mariadb-operator.yaml
```diff
@@ -0,0 +1,107 @@
...
+        kubectl apply -f https://raw.githubusercontent.com/mariadb-operator/mariadb-operator/main/config/crd/bases/k8s.mariadb.com_connections.yaml
+        kubectl apply -f https://raw.githubusercontent.com/mariadb-operator/mariadb-operator/main/config/crd/bases/k8s.mariadb.com_backups.yaml
+        kubectl apply -f https://raw.githubusercontent.com/mariadb-operator/mariadb-operator/main/config/crd/bases/k8s.mariadb.com_restores.yaml
+        kubectl apply -f https://raw.githubusercontent.com/mariadb-operator/mariadb-operator/main/config/crd/bases/k8s.mariadb.com_sqljobs.yaml
+        kubectl apply -f https://raw.githubusercontent.com/mariadb-operator/mariadb-operator/main/config/crd/bases/k8s.mariadb.com_databases.yaml
+        kubectl apply -f https://raw.githubusercontent.com/mariadb-operator/mariadb-operator/main/config/crd/bases/k8s.mariadb.com_users.yaml
+        kubectl apply -f https://raw.githubusercontent.com/mariadb-operator/mariadb-operator/main/config/crd/bases/k8s.mariadb.com_grants.yaml
+        kubectl apply -f https://raw.githubusercontent.com/mariadb-operator/mariadb-operator/main/config/crd/bases/k8s.mariadb.com_maxscales.yaml
```
Dentrax: nit: instead of getting those from `main`, using `{{package.version}}` preferable and compatible.

mariadb-operator.yaml
```diff
@@ -0,0 +1,107 @@
...
+          Resources
+          Starting manager
+          Starting metrics server
+          Starting EventSource
+          Starting Controller
+          Starting workers
+        post: |
+          echo "Verifying metrics endpoint"
+          curl -sf http://127.0.0.1:8080/metrics | head -20
+          echo "Metrics endpoint is serving as expected"
```
Dentrax: Can we use `grep` instead of `head`? it might be leading false positive.

```suggestion
          curl -sf http://127.0.0.1:8080/metrics | head -20
```

rahulduvedi: added better error handling for this one

mariadb-operator.yaml
rahulduvedi: Used specific version for CRDs

Dentrax (approved):

PR #57980, opened 2025-06-30
initial build for aws-node-termination-handler

developer-guy:
testing is not an option here since the project constantly queries AWS Metadata Service and it keeps failing like this:

```bash
2025/06/30 12:02:52 WARN panic: There was a problem checking for spot ITNs: Unable to parse metadata response: Unable to get a response from IMDS: Get "http://169.254.169.254/latest/meta-data/spot/instance-action": dial tcp 169.254.169.254:80: connect: network is unreachable
```

### DIFF
--- a/aws-node-termination-handler.yaml
+++ b/aws-node-termination-handler.yaml
@@ -1,0 +1,54 @@
+package:
+  name: aws-node-termination-handler
+  version: "1.25.1"
+  epoch: 0
+  description: Gracefully handle EC2 instance shutdown within Kubernetes
+  copyright:
+    - license: Apache-2.0
+
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: 5641359c91288d0589c3c188309ce2771eb93838
+      repository: https://github.com/aws/aws-node-termination-handler
+      tag: v${{package.version}}
+
+  - uses: go/bump
+    with:
+      deps: |-
+        golang.org/x/crypto@v0.35.0
+        golang.org/x/oauth2@v0.27.0
+        golang.org/x/net@v0.38.0
+
+  - uses: go/build
+    with:
+      ldflags: -X github.com/google/ko/pkg/commands.Version=${{package.version}}
+      output: node-termination-handler
+      packages: ./cmd/node-termination-handler.go
+      tags: nthlinux
+
+subpackages:
+  - name: ${{package.name}}-compat
+    description: Compatibility package for the ${{package.name}}
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/
+          ln -sf /usr/bin/node-termination-handler ${{targets.contextdir}}/node-termination-handler
+    test:
+      pipeline:
+        - runs: test "$(readlink /node-termination-handler)" = "/usr/bin/node-termination-handler"
+
+update:
+  enabled: true
+  github:
+    identifier: aws/aws-node-termination-handler
+    strip-prefix: v
+
+test:
+  pipeline:
+    - runs: |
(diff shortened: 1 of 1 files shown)


aws-node-termination-handler.yaml
```diff
@@ -0,0 +1,54 @@
...
+        golang.org/x/crypto@v0.35.0
+        golang.org/x/oauth2@v0.27.0
+        golang.org/x/net@v0.38.0
+
+  - uses: go/build
+    with:
+      ldflags: -X github.com/google/ko/pkg/commands.Version=${{package.version}}
+      output: node-termination-handler
+      packages: ./cmd/node-termination-handler.go
+      tags: nthlinux
```
tulilirockz, line 28: Oooo whats this? Never seen this tag being used

aws-node-termination-handler.yaml
```diff
@@ -0,0 +1,54 @@
...
+    description: Compatibility package for the ${{package.name}}
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/
+          ln -sf /usr/bin/node-termination-handler ${{targets.contextdir}}/node-termination-handler
+    test:
+      environment:
+        contents:
+          packages:
+            - ${{package.name}}
```
tulilirockz:
```suggestion
```
I think you dont need the base package to check if the symlink is valid, usually even if it is broken it _should_ resolve properly

aws-node-termination-handler.yaml
```diff
@@ -0,0 +1,54 @@
...
+update:
+  enabled: true
+  github:
+    identifier: aws/aws-node-termination-handler
+    strip-prefix: v
+
+test:
+  pipeline:
+    - runs: |
+        node-termination-handler -h
```
tulilirockz, line 54:
```suggestion
        set -o pipefail
        node-termination-handler -h | grep -F -e "some output from -h"
```
Lets just make sure this smoke test doesnt blow up somehow. Also, any chance we could do `--version` or some equivalent?

aws-node-termination-handler.yaml
```diff
@@ -0,0 +1,54 @@
...
+        - runs: test "$(readlink /node-termination-handler)" = "/usr/bin/node-termination-handler"
+
+update:
+  enabled: true
+  github:
+    identifier: aws/aws-node-termination-handler
+    strip-prefix: v
+
+test:
+  pipeline:
```
tulilirockz, line 48: Lets write a comment here just so we dont forget why we can't test this as you mentioned on the PR description (git blame should be enough but its good to have a note here if possible)

aws-node-termination-handler.yaml
developer-guy: I saw it here: https://github.com/aws/aws-node-termination-handler/blob/5641359c91288d0589c3c188309ce2771eb93838/Makefile#L43

aws-node-termination-handler.yaml
developer-guy: nope there is no version subcmd or flag.

aws-node-termination-handler.yaml
developer-guy: makes sense, thank you!

aws-node-termination-handler.yaml
developer-guy: but its failing in grep somehow, so I leave it like this.

tulilirockz: Its probably because you need `2>&1`

aws-node-termination-handler.yaml
tulilirockz: Awesome, thank you!

tulilirockz (approved):

Dentrax (approved):

PR #57987, opened 2025-06-30
feat(package): kubelet-csr-approver

rakshitgondwal:


kubelet-csr-approver.yaml
```diff
@@ -0,0 +1,179 @@
...
+          signerName: kubernetes.io/kubelet-serving
+          usages:
+          - digital signature
+          - key encipherment
+          - server auth
+          groups:
+          - system:nodes
+        EOF
+        sleep 10
+        if kubectl get csr test-node-csr-valid -o jsonpath='{.status.conditions[?(@.type=="Approved")].status}' | grep -q "True"; then
```
klrmngr: could we use wait instead of sleep

`kubectl wait --for=condition=Approved csr/test-node-csr-valid --timeout=10s`

kubelet-csr-approver.yaml
```diff
@@ -0,0 +1,179 @@
...
+          request: $(cat /tmp/bad-node.csr | base64 | tr -d '\n')
+          signerName: kubernetes.io/kubelet-serving
+          usages:
+          - digital signature
+          - key encipherment
+          - server auth
+          groups:
+          - system:nodes
+        EOF
+        sleep 10
```
klrmngr: similar comment about wait vs sleep

klrmngr (dismissed):

kubelet-csr-approver.yaml
rakshitgondwal: added

kubelet-csr-approver.yaml
rakshitgondwal: cant add wait here since the kubectl apply is expected to fail

kubelet-csr-approver.yaml
```diff
@@ -0,0 +1,62 @@
...
+        -X github.com/postfinance/kubelet-csr-approver/internal/cmd.ref=$(git rev-parse --abbrev-ref HEAD)
+      output: kubelet-csr-approver
+
+subpackages:
+  - name: ${{package.name}}-compat
+    description: "Compatibility package to place binary in the location expected by kubelet-csr-approver"
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/ko-app
+          ln -sf /usr/bin/kubelet-csr-approver ${{targets.contextdir}}/ko-app/kubelet-csr-approver
```
yashpimple, line 35: lets add an small test `stat /ko-app/kubelet-csr-approver`

rakshitgondwal: added

kubelet-csr-approver.yaml
```diff
@@ -0,0 +1,66 @@
...
+    test:
+      pipeline:
+        - runs: |
+            stat /ko-app/kubelet-csr-approver
```
Dentrax:
```suggestion
	    test "$(readlink /ko-app/kubelet-csr-approver)" = "/usr/bin/kubelet-csr-approver"
```

kubelet-csr-approver.yaml
```diff
@@ -0,0 +1,66 @@
...
+  - uses: git-checkout
+    with:
+      repository: https://github.com/postfinance/kubelet-csr-approver
+      tag: v${{package.version}}
+      expected-commit: fabca038b8165aebc7ee62d3019fa67b37b972a3
+
+  - uses: go/build
+    with:
+      packages: ./cmd/kubelet-csr-approver/
+      tags: debug
```
Dentrax, line 23: Why we set `debug` tag? I couldn't see it in the upstream:

* https://github.com/postfinance/kubelet-csr-approver/blob/main/.ko.yaml
* https://github.com/postfinance/kubelet-csr-approver/blob/59f678e3a30dd40a59bb11d6906014f18cce2861/.github/workflows/publish.yaml#L80-L84


```suggestion
```

kubelet-csr-approver.yaml
```diff
@@ -0,0 +1,66 @@
...
+          --level=5
+        timeout: 30
+        expected_output: |
+          Kubelet-CSR-Approver controller starting.
+          starting server
+          Starting metrics server
+          Serving metrics server
+          Starting EventSource
+          Starting Controller
+          Starting workers
```
Dentrax, line 74: Can we have `post:` step with `wait-for-it` and simple `curl | grep` if possible?

kubelet-csr-approver.yaml
```diff
@@ -0,0 +1,66 @@
...
+  copyright:
+    - license: MIT
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/postfinance/kubelet-csr-approver
+      tag: v${{package.version}}
+      expected-commit: fabca038b8165aebc7ee62d3019fa67b37b972a3
+
```
Dentrax, line 15: Not 100% sure but we might need to do `mkdir -p ${{targets.contextdir}}/var/run/ko` as well.

Dentrax (changes requested):

kubelet-csr-approver.yaml
rakshitgondwal: It is mentioned in the readme: https://github.com/postfinance/kubelet-csr-approver?tab=readme-ov-file#build-and-development

kubelet-csr-approver.yaml
rakshitgondwal: added

Dentrax (dismissed):

Dentrax (approved):

PR #57990, opened 2025-06-30
keycloack: bump quarkus-vertx dependency

aborrero:
Bump quarkus-vertx dependency to remediate CVE-2025-49574 GHSA-9623-mj7j-p9v4.

See also: https://github.com/chainguard-dev/CVE-Dashboard/issues/25410


aborrero:
we are pending an upstream fix.

PR #57991, opened 2025-06-30
feat(pkg): cis-operator

rakshitgondwal:


cis-operator.yaml
```diff
@@ -0,0 +1,72 @@
...
+  - uses: go/bump
+    with:
+      deps: |-
+        github.com/go-viper/mapstructure/v2@v2.3.0
+        golang.org/x/net@v0.38.0
+
+  - uses: go/build
+    with:
+      packages: .
+      tags: netgo,osusergo
```
developer-guy: you don't need to specify them here as they are already given as defaults in go/build pipeline: https://github.com/chainguard-dev/melange/blob/5425a36a8e8b21197e45a4f5e019a4963d585970/pkg/build/pipelines/go/build.yaml#L29

cis-operator.yaml
```diff
@@ -0,0 +1,72 @@
...
+      ldflags: |
+        -X github.com/rancher/cis-operator.Version=${{package.version}}
+        -X github.com/rancher/cis-operator.GitCommit=$(git rev-parse HEAD)
```
developer-guy: lets use --short here as well as they specified here: https://github.com/rancher/cis-operator/blob/969cb5627b97d94e51a69d0e83ef52554cb1fa43/hack/make/build.mk#L19

cis-operator.yaml
```diff
@@ -0,0 +1,72 @@
...
+        timeout: 10
+        expected_output: |
+          Starting CIS-Operator
+          Starting cis.cattle.io/v1
+          Starting /v1, Kind=Service controller
+          Starting /v1, Kind=Pod controller
+          Starting /v1, Kind=ConfigMap controller
+          Starting batch/v1, Kind=Job controller
+        post: |
+          curl -sfSL http://localhost:8080/metrics
```
developer-guy: maybe we could check the response with the expected values to make sure that we are not missing anything? \o/

rakshitgondwal: updated

developer-guy (approved):
I saw all the comments addressed so I'm approving it!

PR #58005, opened 2025-06-30
k8s_gateway: init package at 1.4.0

tulilirockz:


k8s_gateway.yaml
```diff
@@ -0,0 +1,66 @@
+package:
+  name: k8s_gateway
```
kranurag7, line 2:
```suggestion
  name: k8s-gateway
```
And at other places

tulilirockz: This is the project name, though 🤔 Shouldn't we use that instead?

kranurag7: I was thinking that using `-` as standard for package naming is nice and less error prone. This also avoids `-` and `_` confusion as we do more and more.

EyeCantCU: We've used `_` in other package names, this is fine

EyeCantCU (approved):

PR #58014, opened 2025-06-30
cmake-3: Version stream cmake

sergiodj:
With CMake 4.0 almost ready to be uploaded to Wolfi, there are a few packages that still FTBFS with it.  We need to version stream CMake 3.x to make sure these packages keep building fine.

### DIFF
--- a/cmake-3.yaml
+++ b/cmake-3.yaml
@@ -1,0 +1,95 @@
+package:
+  name: cmake-3
+  version: 3.31.7
+  epoch: 0
+  description: "CMake 3.x is an open-source, cross-platform family of tools designed to build, test and package software"
+  dependencies:
+    provides:
+      - cmake=${{package.full-version}}
+  copyright:
+    - license: BSD-3-Clause
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - bzip2-dev
+      - ca-certificates-bundle
+      - curl-dev
+      - expat-dev
+      - jsoncpp-dev
+      - libarchive-dev
+      - libuv-dev
+      - ncurses-dev
+      - nghttp2-dev
+      - openssl-dev
+      - rhash-dev
+      - samurai
+      - wolfi-base
+      - xz-dev
+      - zlib-dev
+
+pipeline:
+  - uses: fetch
+    with:
+      uri: https://www.cmake.org/files/v3.31/cmake-${{package.version}}.tar.gz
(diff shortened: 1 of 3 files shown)


sergiodj:
This is ready for review.

sergiodj:
Hm, this will need a bit more work.

sergiodj:
OK, finally ready now :-).  It was missing a `provides`.

cmake-3.yaml
```diff
@@ -0,0 +1,96 @@
+package:
+  name: cmake-3
+  version: 3.31.7
+  epoch: 0
+  description: "CMake 3.x is an open-source, cross-platform family of tools designed to build, test and package software"
+  dependencies:
+    provider-priority: 10
```
dannf: Is this cargo-culted from `cmake.yaml`? I think this is there so `cmake` is preferred over `cmake-bootstrap`.

Keeping it here will mean both `cmake-3.31.7-r0` and `cmake-3-3.31.7-r0` will be providing `cmake=3.31.7-r0` at priority 10. An install of `cmake=3.31.7` will presumably swap back and forth between `cmake` and `cmake-3` depending on which one has the greater `-rX` number. It would be nice to have something more deterministic.

cmake-3/wolfi-tests/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+project(HelloWorld)
+cmake_minimum_required(VERSION 3.0)
```
dannf: According to https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html , we should get a deprecation warning if this isn't at least 3.10.  I wonder if w shouldn't just put `3.31.7` here, since we'll only move forward. We can also specify a max here. Maybe `(VERSION 3.31.7 3.999)` would be appropriate, to make sure we also don't accidentally get bumped to 4.x?

cmake-3.yaml
sergiodj: Ah, fair.  Yeah, I had removed this bit in the first version of this PR, but then decided to put it back.  Good catch, let me fix it.

cmake-3/wolfi-tests/CMakeLists.txt
sergiodj: I like it!  I think using the max version here makes a lot of sense.  Let me adjust.

dannf (approved):

PR #58023, opened 2025-06-30
update pg-failover-slots to v17

sschubertchainguard:
update pg-failover-slots to v17

### DIFF
--- a/pg-failover-slots-17.yaml
+++ b/pg-failover-slots-17.yaml
@@ -1,7 +1,7 @@
 package:
   name: pg-failover-slots-17
   version: 1.1.0
-  epoch: 3
+  epoch: 0
   description: PG Failover Slots extension for PostgreSQL 17
   copyright:
     - license: BSD-3-Clause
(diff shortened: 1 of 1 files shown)


pg-failover-slots-17.yaml
```diff
@@ -1,8 +1,8 @@
 package:
-  name: pg-failover-slots-16
+  name: pg-failover-slots-17
   version: 1.1.0
   epoch: 3
```
kranurag7: epoch should be zero for a new package.

pg-failover-slots-17.yaml
```diff
@@ -33,7 +33,7 @@ pipeline:
 test:
   pipeline:
     - runs: |
-        if [[ -f /usr/lib/postgresql16/pg_failover_slots.so ]]; then
+        if [[ -f /usr/lib/postgresql7/pg_failover_slots.so ]]; then
```
kranurag7:
```suggestion
        if [[ -f /usr/lib/postgresql17/pg_failover_slots.so ]]; then
```

sschubertchainguard:
Woops! Thanks for catching those @kranurag7 !

PR #58028, opened 2025-06-30
libwebsockets: Backport patch to fix build with CMake 4.0

sergiodj:


dannf (approved):

murraybd:
I don't think this is necessary any longer given pr 59613

PR #58029, opened 2025-06-30
add editcap subpackage to tshark

sanghanan:
Fixes:

Related: https://github.com/chainguard-dev/image-requests/issues/6276
Note: We're building `editcap` to be used as a standalone package if the user does not want to install all of tshark suite, as required in the package request. 

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches


tshark-4.4.yaml
```diff
@@ -45,6 +47,68 @@ pipeline:
...
+        - uses: test/tw/ldd-check
+
+  - name: editcap
+    description: Edit and/or translate the format of capture files
+    dependencies:
+      runtime:
+        - ${{package.name}}-libs
+    pipeline:
+      - runs: |
+          install -Dm755 "${{targets.destdir}}/usr/bin/editcap" "${{targets.contextdir}}/usr/bin/editcap"
```
tulilirockz, line 71:
```suggestion
          mkdir -p "${{targets.contextdir}}/usr/bin"
          mv "${{targets.destdir}}/usr/bin/editcap" "${{targets.contextdir}}/usr/bin/editcap"
```
`install` is just copying the file from one place to another, the tshark package should still have that binary in it if you dont `mv` instead

tshark-4.4.yaml
```diff
@@ -45,6 +47,68 @@ pipeline:
...
+        contents:
+          packages:
+            - tcpdump
+            - curl
+      pipeline:
+        - uses: test/tw/ldd-check
+        - name: Verify installation
+          runs: |
+            editcap --help
+            editcap --version
```
tulilirockz:
```suggestion
            set -o pipefail
            editcap --help | grep -F -e "(some output from editcap)"
            editcap --version | grep -F -e "${{package.version}}"
```
Lets just make sure we are not getting garbled output from it

tshark-4.4.yaml
```diff
@@ -45,6 +47,68 @@ pipeline:
...
+        - name: Download a sample .pcap file
+          runs: |
+            curl https://raw.githubusercontent.com/wireshark/wireshark/master/test/captures/http-ooo2.pcap -o /tmp/test.pcap
```
tulilirockz:
```suggestion
            curl https://gitlab.com/wireshark/wireshark/-/raw/1121a14269bbd9df5327225ac4379d0c961e3d6f/test/captures/http-ooo2.pcap?inline=false -o /tmp/test.pcap
```
We should use the official repository here and maybe pin to a commit just so there is no way these tests will blow up in the future due to file restructuring in the repos

tshark-4.4.yaml
```diff
@@ -45,6 +47,68 @@ pipeline:
...
+        - name: Verify packet split
+          runs: |
+            cd /tmp
+            no_of_packets=$(tcpdump -r test.pcap | wc -l)
+            # splitting each packet into its own file
+            editcap -c 1 test.pcap split.pcap
+            no_of_files=$(ls -1 split* | wc -l)
```
tulilirockz:
```suggestion
            no_of_files=$(find -type f -iname 'split*' | wc -l)
```
`ls`'s output is inconsistent depending on the implementation that we are dealing with, to make sure this wont blow up we can use `gnufind` instead (see https://unix.stackexchange.com/questions/128985/why-not-parse-ls-and-what-to-do-instead)

tshark-4.4.yaml
```diff
@@ -1,13 +1,15 @@
...
-  epoch: 1
+  epoch: 2
   description: "TShark is a network protocol analyzer to dump and analyze network traffic"
   copyright:
     - license: GPL-2.0-only
   dependencies:
     provides:
       - tshark=${{package.full-version}}
+    runtime:
+      - ${{package.name}}-libs
```
AnmolVirdi, line 12: We can remove `subpackage` from runtime dependencies of main package. Ideally, we add all subpackages separately image level.
```suggestion
```

We can copy lib files at subpackage level to keep them distinct.

tshark-4.4.yaml
```diff
@@ -45,6 +47,68 @@ pipeline:
...
+            editcap -c 1 test.pcap split.pcap
+            no_of_files=$(ls -1 split* | wc -l)
+            if [[ "$no_of_packets" -ne "$no_of_files" ]]; then
+              echo "Error: Number of packets ($no_of_packets) does not match number of files ($no_of_files)."
+              exit 1
+            fi
+            echo "Successfully verified packet split."
+        - name: Verify discarding packets
+          runs: |
+            cd /tmp
```
tulilirockz:
```suggestion
```

tshark-4.4.yaml
```diff
@@ -45,6 +47,68 @@ pipeline:
...
+        - name: Verify installation
+          runs: |
+            editcap --help
+            editcap --version
+        - name: Download a sample .pcap file
+          runs: |
+            curl https://raw.githubusercontent.com/wireshark/wireshark/master/test/captures/http-ooo2.pcap -o /tmp/test.pcap
+        - name: Verify packet split
+          runs: |
+            cd /tmp
```
tulilirockz, line 91:
```suggestion
```

You are `cd`ing into the same directory on both tests, its effectively the same as not cding at all since the testing workspace is already ephemeral by default

tulilirockz (commented):
I have a few nits to pick here but this is looking awesome, thank you!!!

tshark-4.4.yaml
```diff
@@ -45,6 +47,68 @@ pipeline:
...
 
 subpackages:
+  - name: ${{package.name}}-libs
+    dependencies:
+      provides:
+        - tshark-libs=${{package.full-version}}
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/usr/lib
+          mv "${{targets.destdir}}"/usr/lib/* "${{targets.subpkgdir}}"/usr/lib
```
AnmolVirdi, line 57: W.r.t the above comment
```suggestion
          mkdir -p "${{targets.subpkgdir}}"/usr/lib
          cp -r "${{targets.destdir}}"/usr/lib/* "${{targets.subpkgdir}}"/usr/lib
```

AnmolVirdi (dismissed):

tshark-4.4.yaml
tulilirockz: The libs are pretty big though. Copying does not seem the way to go here

tshark-4.4.yaml
tulilirockz: [See this](https://github.com/wolfi-dev/os/pull/58029#discussion_r2177772033)

tshark-4.4.yaml
kranurag7: We want to have editcap in the main package here.

sanghanan: @tulilirockz this is a bit of a weird one. The customer wants to install `editcap` without installing the whole wireshark suite.

tshark-4.4.yaml
sanghanan: That makes sense! Thanks for the suggestion

tshark-4.4.yaml
sanghanan: I was a little concerned about how to preserve this, and I didn't think of using a commit. This is perfect, thanks!

tshark-4.4.yaml
sanghanan: I did not know this! Thanks for pointing that out, will change this

tshark-4.4.yaml
sanghanan: So, at each step of the pipeline, we go back to `/home/build`. Sharing the logs here:
```
2025/07/01 11:12:55 INFO running step "Verify discarding packets"
2025/07/01 11:12:55 WARN + '[' -d /home/build ]
2025/07/01 11:12:55 WARN + cd /home/build
2025/07/01 11:12:55 WARN + tcpdump -r test.pcap
2025/07/01 11:12:55 WARN + wc -l
2025/07/01 11:12:55 WARN tcpdump: test.pcap: No such file or directory
```
Would it be better to combine them into a single step instead?

tshark-4.4.yaml
tulilirockz: > We want to have editcap in the main package here.

I was thinking we could just add a runtime dependency for `editcap` on `tshark` instead (@kranurag7 )

tulilirockz: > The customer wants to install editcap without installing the whole wireshark suite.

Yupyup! Splitting it out with `mv` lets us _not_ have `editcap` on `tshark`, then we can just depend on it on runtime then BAM we have the split out subpackage _and_ full tshark when we `apk add tshark` (@sanghanan )

tshark-4.4.yaml
tulilirockz: Huh yeah maybe! Probably best to do that, yeah!

tshark-4.4.yaml
sanghanan: Makes sense, I fixed it!

tshark-4.4.yaml
sanghanan: I see what you mean
But, I think if we were to receive more requests like this in the future (for example: just dumpcap or capinfos), it would be more maintainable to just have a library subpackage like ${{package.name}}-libs.

tshark-4.4.yaml
sanghanan: I went back and used the -libs subpackage instead. This should make it easier to maintain if we need more standalone subpackages in the future.

tshark-4.4.yaml
mcaveety: Yeah, we don't want to move editcap out of tshark. Tshark is the composite of all binaries like mergecap, editcap, dumpcap, etc. and this is an edge case where we want editcap by itself. It makes more sense to make a tshark derivative (editcap) with less, rather than fragmenting the main package and having to re-add as a runtime dependency.

mcaveety (dismissed):
This looks good to me. I think we're taking the right approach by keeping `editcap` inside `tshark`, and creating a separate subpackage for using `editcap` standalone. Sadly we won't be able to get the binary slimmed down much more than this, but that's going to be hard to avoid with the shared libraries needed by each binary. I think we're good to go. 🦈

tshark-4.4.yaml
```diff
@@ -77,9 +140,16 @@ test:
...
+        tshark --help | grep -F -e "Usage: tshark"
+        tshark --version | grep -F -e "${{package.version}}"
+    - name: Test packet capture
       runs: |
-        tshark -w /tmp/out.pcap -c 10 host example.com&
+        tshark -w /tmp/out.pcap -c 20 host example.com &
+        PID=$!
         sleep 1; curl -sI http://example.com; sleep 1
         tshark -r /tmp/out.pcap -Y "http" |grep '200 OK'
+        kill $PID
```
tulilirockz, line 154: Here is a neat trick:
```suggestion
       PID=$!
       trap 'kill $PID' EXIT
       sleep 1; curl -sI http://example.com; sleep 1
       tshark -r /tmp/out.pcap -Y "http" |grep '200 OK'
```

sanghanan: The problem here is that tshark waits until it receives 20 packets, so we're not gonna get an error for a real long time. I can set a timeout, but it didn't feel that necessary, because if grep fails, we should get an `exit 1` and if it passes, we don't have to wait for the timeout.

tulilirockz (approved):

PR #58046, opened 2025-07-01
Knative Serving:  components addition

AnmolVirdi:
Packages added:
- `knative-serving-activator`
- `knative-serving-autoscaler`
- `knative-serving-controller`
- `knative-serving-webhook`

### DIFF
--- a/knative-serving-1.18.yaml
+++ b/knative-serving-1.18.yaml
@@ -16,6 +16,11 @@ pipeline:
       repository: https://github.com/knative/serving
       tag: knative-v${{package.version}}
 
+  - uses: go/bump
+    with:
+      deps: |-
+        github.com/golang-jwt/jwt/v4@v4.5.2
+
 subpackages:
   - name: ${{package.name}}-queue
     description: "Queue component for Knative serving"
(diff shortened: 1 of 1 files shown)


knative-serving-1.18.yaml
```diff
@@ -87,6 +92,560 @@ subpackages:
       provides:
         - knative-serving-queue-compat=${{package.full-version}}
 
+  - name: ${{package.name}}-activator
```
Dentrax, line 95: Is it possible to use `- range` for the components that listed in `items:`? I feel like build steps likely same so we can avoid duplicate and it would significantly easier for the review.

Also, if possible we should use `test/daemon-check-output` for test confidence.

And also, is it possible to avoid those long manifest and apply those directly from upstream? I'm not so sure if we really need those to run (its also ok to write negative-tests as well)

Thanks!

AnmolVirdi: Thanks @Dentrax, I'll add ranges and reduce test sizes by reverting to negative tests.

AnmolVirdi: @Dentrax the tests for all the components will be separate since some require env variables (`-queue`) and some require configmaps which i can minimize.

Please let me know if we can keep these packages without the ranges since tests will be different for all components.

Also `test/daemon-check-output` will always fail in this case since there are errors in the logs. The only way to tests these components is at image level, after deploying knative operator.

Dentrax: ack - thanks for looking at this, resolving the discussion. 🤞

Dentrax:
Thanks for the work here!

Its now superseded by https://github.com/wolfi-dev/os/pull/58145

PR #58047, opened 2025-07-01
openssl: Update to 3.5.1

dannf:
Leaving in draft form until testing is complete.


sil2100 (approved):
+1. The fix-jitter patch is still required, since the changes didn't quite go to upstream yet.

PR #58120, opened 2025-07-01
timescaledb-tune: add compat subpackage

tulilirockz:
Necessary for the timescaledb entrypoint scripts

### DIFF
--- a/timescaledb-tune.yaml
+++ b/timescaledb-tune.yaml
@@ -1,7 +1,7 @@
 package:
   name: timescaledb-tune
   version: 0.18.0
-  epoch: 0
+  epoch: 1
   description: A tool for tuning TimescaleDB for better performance by adjusting settings to match your system's CPU and memory resources.
   copyright:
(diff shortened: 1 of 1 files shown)


timescaledb-tune.yaml
```diff
@@ -25,15 +25,28 @@ update:
     use-tag: true
     strip-prefix: v
 
+subpackages:
+  - name: timescaledb-tune-compat
+    description: "Compatibility subpackage for upstream docker entrypoint scripts"
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.contextdir}}/usr/local/bin"
+          ln -s /usr/bin/timescaledb-tune "${{targets.contextdir}}/usr/local/bin/timescaledb-tune"
```
EyeCantCU, line 34: We should create this symlink in the package that contains the entrypoint script since it depends on it being present anyway (unless this is useful somewhere else)

EyeCantCU (changes requested):

PR #58133, opened 2025-07-01
Implement no-hardening.spec

xnox:
- **gcc: add optional spec file integration to versions 11, 12 and 13**
  

- **gcc: add a no-hardening.spec file**
  Unlike common pattern to use /dev/null instead add an explicit
  no-hardening.spec file, which still has melange and toolchain
  integrations.
  

- **openssf-compiler-options epoch bump**
  

- **gcc: add test cases for no-hardening.spec**

### DIFF
--- a/openssf-compiler-options/usr/lib/gcc/aarch64-unknown-linux-gnu/11/no-hardening.spec
+++ b/openssf-compiler-options/usr/lib/gcc/aarch64-unknown-linux-gnu/11/no-hardening.spec
@@ -1,0 +1,2 @@
+%include_noerr </usr/lib/oldglibc/gcc.spec>
+%include_noerr </home/build/.melange.gcc.spec>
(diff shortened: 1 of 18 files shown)


openssf-compiler-options/usr/lib/gcc/aarch64-unknown-linux-gnu/11/no-hardening.spec
```diff
@@ -0,0 +1,2 @@
+%include_noerr </usr/lib/oldglibc/gcc.spec>
```
justinvreeland, line 1: Doesn't ct-manylinux replace `openssf-compiler-options`. I suppose it makes sense that `no-hardening` would help with the fallback for manylinux and since there's low chances people would have both toolchains installed and also have problems when the older spec file additions. But can these be installed at the same time to provide this fallback?

xnox: It only replaces the oldglibc clang config file which must exist as clang doesn't have "_noerr" include.
all of these spec files are used as is.

justinvreeland (approved):

PR #58141, opened 2025-07-01
python builds fixes for 2.28

xnox:
- **With or without you, it is building with lto**
  

- **Disable optimizations for oldglibc builds due to gcc missmatch**
  

- **Epoch bump**

Due to missmatch of gcc 14.2 and 14.3 between wolfi and 2.28 builds profiling build fails, disable profile guided optimisations in 2.28 builds for now.


xnox:
@OddBloke you will love this green tick on eco-2-28 builds =)

dannf (approved):

PR #58186, opened 2025-07-02
glibc: update ld.so.conf for completed usrmerge

xnox:
Now that usrmerge has been fully completed, the output from ldd often
looks odd. As if all libraries moved to /lib, when in fact they are
all actually in /usr/lib. Update the libc.conf to only use
non-symlinked real library locations. This will make ldd output to be
clean and reference /usr/lib for all libraries.

### DIFF
--- a/glibc.yaml
+++ b/glibc.yaml
@@ -2,7 +2,7 @@ package:
   name: glibc
   version: "2.41"
   # Every glibc update causes build disruption; always announce on #eng-psa
-  epoch: 52
+  epoch: 53
   description: "the GNU C library"
   copyright:
     - license: LGPL-2.1-or-later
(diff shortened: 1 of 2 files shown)


dannf (commented):
Thanks! I've been wanting to do this for a while now.

I think @sergiodj may have a glibc update planned for today. Let me check that first, if not I'll go ahead and merge this as a standalone change.

sergiodj:
Thanks for the ping.  Yes, I can incorporate this change to my PR, for sure.

dannf:
Cool - will go ahead and close this one then as to be superseded by https://github.com/wolfi-dev/os/pull/58020. Thanks again @xnox !

PR #58192, opened 2025-07-02
Builds compat package to run redis w/harbor

srishtih:
This sub-package is being added based on this customer request: https://github.com/chainguard-dev/image-requests/issues/6260

When users try to use 'redis' with harbor, but it appears as if harbor require some customisations to redis in order to run the image alongside harbor. To that end, harbor publishes their own 'redis' image:

```
goharbor/redis-photon
```
The source code for building this image, appears to reside here:
https://github.com/goharbor/harbor/tree/main/make/photon/redis

We don't currently build that image variant today, and given this request it'd be useful to do so. i.e we are adding this package to provide the expected endpoint and conf files.

### DIFF
--- a/harbor-2.13.yaml
+++ b/harbor-2.13.yaml
@@ -1,7 +1,7 @@
 package:
   name: harbor-2.13
   version: "2.13.1"
-  epoch: 2
+  epoch: 3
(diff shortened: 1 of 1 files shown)


anushkamittal20 (approved):
Thanks for working on this!
lgtm!

PR #58254, opened 2025-07-03
Kubescape-storage package addition

AnmolVirdi:
**Kubescape-storage package** `Current version: v0.0.185`
- An Aggregated APIServer for the Kubescape internal storage services.

**Type**: Go binary
**REF**: https://github.com/kubescape/storage

**Tests:**
- Minimal tests to check whether kubescape-storage binary is running APIServer. It needs complete setup with operator


kubescape-storage.yaml
```diff
@@ -0,0 +1,68 @@
...
+        {
+          "clusterName": "test-cluster",
+          "region": "us-test1"
+        }
+        EOF
+    - name: "Functional test"
+      runs: |
+        export KUBERNETES_SERVICE_HOST=127.0.0.1
+        export KUBERNETES_SERVICE_PORT=443
+        storage serve 2>&1 | tee /logs.txt
```
Dentrax: Can we migrate this to `test/daemon-check-output`? See `kubescape-operator` for example.

Dentrax (changes requested):

kubescape-storage.yaml
AnmolVirdi: Updated tests to using `daemon-check`.

kubescape-storage.yaml
```diff
@@ -0,0 +1,59 @@
...
+        {
+          "clusterName": "test-cluster",
+          "region": "us-test1"
+        }
+        EOF
+    - uses: test/daemon-check-output
+      with:
+        start: storage serve
+        expected_output: |
+          APIServer started
```
Dentrax, line 59: Can we check if port is up w wait-for-it in `post:`? without post steps, this might resulting false-positive CIs in the future. i.e., it might exit just after print out the log above.

Dentrax (changes requested):

PR #58258, opened 2025-07-03
kwok: create sa token and ca.cert

Dentrax:
`kwok` pipeline now creates `token`, `ca.crt`, and `namepspace` files for default SA located in `/var/run/secrets/kubernetes.io/serviceaccount`.

Previous work: https://github.com/wolfi-dev/os/pull/50425

### DIFF
--- a/pipelines/test/kwok/cluster.yaml
+++ b/pipelines/test/kwok/cluster.yaml
@@ -58,7 +58,6 @@ pipeline:
 
   - name: Create ServiceAccount and Token
     runs: |
-      #!/usr/bin/env bash
       set -euo pipefail
       # https://kubernetes.io/docs/concepts/security/service-accounts/#default-service-accounts
       if [ "${{inputs.serviceaccount}}" != "true" ]; then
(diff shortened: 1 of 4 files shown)


maxgio92 (dismissed):

pipelines/test/kwok/cluster.yaml
```diff
@@ -45,3 +55,24 @@ pipeline:
 
       kubectl wait --for=condition=Ready nodes --all
       kubectl cluster-info
+
+  - name: Create ServiceAccount and Token
+    runs: |
+      #!/usr/bin/env bash
```
dannf: does the shebang do anything here, or is this just treated as a comment?

If it does work and/or you do need `bash`, I think it needs to be added to `needs.packages`. I'm not seeing any need for `bash` below though.

dannf (changes requested):
thx for improving our tests! pipeline comment inline (i didn't review the rest)

pipelines/test/kwok/cluster.yaml
Dentrax: ah yeah... it might treated as a comment. i've removed it!

Dentrax: done! @dannf

dannf (approved):
thanks!

PR #58273, opened 2025-07-03
sudo/misc

dannf:
- **sudo: update license per melange scanning warnings**
- **sudo: Switch from fetch to git-checkout**
- **sudo: enable updates, now that we're using git**
- **sudo: bump epoch to release 1.9.17-r1**

### DIFF
--- a/sudo.yaml
+++ b/sudo.yaml
@@ -1,7 +1,7 @@
 package:
   name: sudo
   version: 1.9.17
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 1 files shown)


sergiodj (approved):
LGTM.  Thanks for doing this.

I can't say I'm happy with manually having to update the package because of the extra patch level thingy (I started working on melange to try and get this sorted out, but it's hard to do without breaking compatibility with Alpine), but I believe this is the best we can do for now.

PR #58296, opened 2025-07-03
iptables: Add tests to prepare for package refactoring

dannf:
We want to fold ip6tables commands back into iptables, and we want to
split the nftables multi-call binaries into their own package. Before
we do that, let's add tests to make sure the transition is smooth.

This runs some smoke tests on all of the commands we expect to be
in each package, and that there are no other commands in the package
other than those being tested.

After the migration, we'll use the smoke tests to verify that
installing a package brings in the same commands, even if the command
has now moved into a dependency.

### DIFF
--- a/iptables.yaml
+++ b/iptables.yaml
@@ -1,7 +1,7 @@
 package:
   name: iptables
   version: 1.8.11
-  epoch: 22
+  epoch: 23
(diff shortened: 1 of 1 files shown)


sergiodj (approved):
Thanks a lot for this.  Built and tested locally on amd64, everything seems OK.  +1

PR #58387, opened 2025-07-04
add aznfs-mount files/dirs

developer-guy:
Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/aznfs-mount.yaml
+++ b/aznfs-mount.yaml
@@ -1,0 +1,42 @@
+package:
+  name: aznfs-mount
+  version: "2.0.12"
+  epoch: 0
+  description: AZNFS Mount Helper
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - bash
+      - conntrack-tools
+      - coreutils
+      - flock
+      - iptables
+      - procps
+      - util-linux
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/Azure/AZNFS-mount.git
+      tag: ${{package.version}}
+      expected-commit: 9511dd9b91eed27ea68ae83d8ba14fd3d3c67dd1
+
+  - runs: |
+      install -Dm755 src/aznfswatchdog "${{targets.contextdir}}"/usr/bin/aznfswatchdog
+      gcc -O3 src/mount.aznfs.c -o "${{targets.contextdir}}"/usr/bin/mount.aznfs
+      install -Dm755 src/mountscript.sh "${{targets.contextdir}}"/opt/microsoft/aznfs/mountscript.sh
+      install -Dm755 scripts/aznfs_install.sh "${{targets.contextdir}}"/opt/microsoft/aznfs/aznfs_install.sh
+      install -Dm755 lib/common.sh "${{targets.contextdir}}"/opt/microsoft/aznfs/common.sh
+
+update:
+  enabled: true
+  github:
+    identifier: Azure/AZNFS-mount
+    use-tag: true
--- a/blob-csi-1.26.yaml
+++ b/blob-csi-1.26.yaml
@@ -1,7 +1,7 @@
 package:
   name: blob-csi-1.26
   version: "1.26.6"
-  epoch: 0
+  epoch: 1
   description: Azure Blob Storage CSI driver
   copyright:
     - license: Apache-2.0
@@ -102,6 +102,7 @@ subpackages:
         - nfs-utils
         - procps
         - udev
+        - aznfs-mount
 
 update:
   enabled: true


tulilirockz:
Are these compilable from source? I dont see why we shouldn't package them instead of fetching a pre-compiled tarball

blob-csi-1.26.yaml
```diff
@@ -80,6 +90,27 @@ subpackages:
...
+            with:
+              uri: "https://github.com/Azure/AZNFS-mount/releases/download/${{vars.aznfsVer}}/aznfs-${{vars.aznfsVer}}-1.x86_64.tar.gz"
+              expected-sha256: "5b4f0936f3997f94a903fce8fe4e3c3ae37801764e52227e64b7085b4e305d72"
+              directory: /tmp
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/opt/microsoft/aznfs
+
+          install -Dm755 /tmp/mount.aznfs ${{targets.contextdir}}/usr/sbin/mount.aznfs
+          install -Dm755 /tmp/sbin/aznfswatchdog ${{targets.contextdir}}/usr/sbin/aznfswatchdog
+          cp -r /tmp/microsoft/aznfs/* ${{targets.contextdir}}/opt/microsoft/aznfs
```
tulilirockz: I would just like some confirmation before this getting merged: this does not seem right? we'd usually package these up... unless those are scripts and not compiled?

tulilirockz (changes requested):

blob-csi-1.26.yaml
developer-guy: these are scripts yes, you will see it when you extract the tar file https://github.com/Azure/AZNFS-mount/releases/download/2.0.12/aznfs-2.0.12-1.arm64.tar.gz \o/

jamonation: Hmm there's a binary in there:
```
[13:50:12|25-07-04]jamon@derp /tmp/aznfs
 0 % file sbin/mount.aznfs
sbin/mount.aznfs: ELF 64-bit LSB executable, ARM aarch64, version 1 (GNU/Linux), statically linked, BuildID[sha1]=df8404fb496ea6f8477d93fc9152a736ff04fa5b, for GNU/Linux 3.7.0, not stripped
```

developer-guy: found how to build it: https://github.com/Azure/AZNFS-mount/blob/81293c1b46c750ca6d7b13a454fbfae73dfb65fa/package.sh#L49

blob-csi-1.26.yaml
```diff
@@ -1,27 +1,39 @@
...
+  aznfsVer: "2.0.12"
+
 environment:
   contents:
     packages:
       - bash
       - blobfuse2
       - blobfuse2-health-monitor
       - ca-certificates
+      - cmake
```
jamonation: cmake isn't called anywhere is it?

developer-guy: yes, removed!

blob-csi-1.26.yaml
```diff
@@ -1,17 +1,27 @@
...
       - wolfi-baselayout
 
+vars:
+  # --- Manual update required ---
+  # This application depends on aznfs-mount project, which are contained
+  # in a different repository and each blob-csi release is using a specific
+  # version of aznfs-mount project so you need to manually update the
+  # aznfsVer variable below to match the version of aznfs-mount that
+  # blob-csi is using: https://github.com/kubernetes-sigs/blob-csi-driver/blob/v1.26.5/pkg/blobplugin/Dockerfile#L27
+  aznfsVer: "2.0.12"
```
kranurag7: what do you think of getting the version here like this?

```bash
awk -F'=' '/aznfsVer/ {print $2}' ./pkg/blobplugin/Dockerfile
```

we get this version; export it and then do manual git clone. 

I think this way we will also reduce the chances of relying in manual updates. The machine will do the job and if it fails then we'll encounter build failures.

developer-guy: This is definitely doable, but don’t you think it’s a bit outside our usual standards? I know we don’t have strict guidelines around this yet, but at the very least, we should try to maintain consistency with similar packages, if that does not make sense, I'll do that change for sure, thanks for the tip brother <3

Dentrax: FWIW https://github.com/Azure/AZNFS-mount.git should be a separate package, then have it as a runtime dep here.

developer-guy: We did this intentionally, as we wanted to avoid maintaining multiple versions of aznfs-mount, especially since it's not a version-streamed package. The main reason is that blob-csi uses different versions of aznfs-mount, so keeping the aznfs-mount version within the main package makes it much simpler to track and manage the relationship between the two, please tell me what do you think?

jamonation: Hmm no we can't just have manual updates here. New versions will get missed and we'll fall behind.

Is there a way in this package, or the aznfs-mount package to tie the two together based on one or the other's version? e.g. fail a build if they diverge from known commit hashes, but otherwise increment either package?

blob-csi-1.26.yaml
```diff
@@ -22,6 +23,7 @@ environment:
       - conntrack-tools
       - curl
       - fuse3
+      - gcc
```
jamonation: `gcc` isn't used here is it?

developer-guy: ah yes, removed!

aznfs-mount.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - iptables
+      - procps
+      - util-linux
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - gcc
```
tulilirockz:
```suggestion
```
`build-base` already gives you a compiler

aznfs-mount.yaml
```diff
@@ -0,0 +1,45 @@
...
+    with:
+      repository: https://github.com/Azure/AZNFS-mount.git
+      tag: ${{package.version}}
+      expected-commit: 9511dd9b91eed27ea68ae83d8ba14fd3d3c67dd1
+
+  - runs: |
+      mkdir -p "${{targets.contextdir}}"/usr/bin
+      mkdir -p "${{targets.contextdir}}"/opt/microsoft/aznfs/
+      gcc src/mount.aznfs.c -o "${{targets.contextdir}}"/usr/bin/mount.aznfs
+      install -Dm755 src/aznfswatchdog "${{targets.contextdir}}"/usr/bin/aznfswatchdog
```
tulilirockz:
```suggestion
      install -Dm755 src/aznfswatchdog "${{targets.contextdir}}"/usr/bin/aznfswatchdog
      gcc src/mount.aznfs.c -o "${{targets.contextdir}}"/usr/bin/mount.aznfs
```
If you change the order like this, `install -D` will do the job of making the folders for you

tulilirockz (commented):
Looking awesome! Just small nits

tulilirockz (dismissed):
Lookin nice!

aznfs-mount.yaml
```diff
@@ -0,0 +1,42 @@
...
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/Azure/AZNFS-mount.git
+      tag: ${{package.version}}
+      expected-commit: 9511dd9b91eed27ea68ae83d8ba14fd3d3c67dd1
+
+  - runs: |
+      install -Dm755 src/aznfswatchdog "${{targets.contextdir}}"/usr/bin/aznfswatchdog
+      gcc src/mount.aznfs.c -o "${{targets.contextdir}}"/usr/bin/mount.aznfs
```
tulilirockz:
```suggestion
      gcc -O3 src/mount.aznfs.c -o "${{targets.contextdir}}"/usr/bin/mount.aznfs
      strip "${{targets.contextdir}}"/usr/bin/mount.aznfs
```
Actually like dont we usually want this + strip? Kinda bad that there is no pipeline for this here but I think you want these

developer-guy: thanks for the tip!

aznfs-mount.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - bash
+      - conntrack-tools
+      - coreutils
+      - flock
+      - iptables
+      - procps
+      - util-linux
```
jamonation, line 16: How many of these are required for this tool, versus the blob-csi package that it is meant to support?

jamonation (approved):

PR #58419, opened 2025-07-06
7zip: update to 2500

imjasonh:
https://github.com/wolfi-dev/os/pull/58418 updated its release-monitoring ID, which we expected to fail the update config check, but it failed for another reason -- because the 7zip source download was not found with the string `25.00`.

Instead, we want to use the new updated RM ID, but transform it so the version can be `2500` (to be > `2300`), and transform the RM metadata so that its `25.00` becomes `2500`, and potentially follows `2501` etc going forward.

### DIFF
--- a/7zip.yaml
+++ b/7zip.yaml
@@ -1,6 +1,6 @@
 package:
   name: 7zip
-  version: 2301
+  version: 25.00
   epoch: 3
   description: "File archiver with a high compression ratio"
   copyright:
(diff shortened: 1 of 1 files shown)


7zip.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: 7zip
-  version: 2301
+  version: 25.00
   epoch: 3
```
ajayk:
```suggestion
  epoch: 0
```

ajayk (approved):

PR #58464, opened 2025-07-07
cerbos: package update

mcaveety:
Related: https://github.com/chainguard-images/images-private/pull/11088

Updates the package to include correct directories and adds `stat` utility (coreutils) as it is needed for the image level tests.


cerbos.yaml
```diff
@@ -1,11 +1,16 @@
...
-  epoch: 0
+  epoch: 1
   description: Cerbos is the open core, language-agnostic, scalable authorization solution that makes user permissions and authorization simple to implement and manage by writing context-aware access control policies for your application resources.
   copyright:
     - license: Apache-2.0
 
+environment:
+  contents:
+    packages:
+      - coreutils # Needed for image tests
```
AnmolVirdi: This will just add the package to the melange build environment and not as a runtime dependency package for cerbos. Adding coreutils are runtime dep in follow up comment
```suggestion
environment:
  contents:
    packages:
      - coreutils # Needed for image tests
```

cerbos.yaml
```diff
@@ -1,11 +1,16 @@
 package:
   name: cerbos
   version: "0.45.1"
-  epoch: 0
+  epoch: 1
   description: Cerbos is the open core, language-agnostic, scalable authorization solution that makes user permissions and authorization simple to implement and manage by writing context-aware access control policies for your application resources.
   copyright:
     - license: Apache-2.0
```
AnmolVirdi: Before adding this please make sure if this is a package runtime dependency or a dependency for entrypoint script or at image-level. In the second case, we will add this as runtime dependency in compat package level.
```suggestion
  dependencies:
    runtime:
      - coreutils

```

AnmolVirdi (changes requested):

cerbos.yaml
mcaveety: 🤦 thank you Anmol. I didn't realize the distinction there

cerbos.yaml
mcaveety: The cerbos utility requires stat  for its initialization, so I believe it would fall under the first category.

AnmolVirdi: I believe we need to dive deeper and analyse what's going on here:
https://github.com/chainguard-images/images-private/actions/runs/16059862985/job/45323624287?pr=11088#step:24:1061

If the `stat` binary or syscall were missing, you’d see an error more like `exec format error`, `command not found`, or `not implemented`. This log is probably due to a file.

mcaveety:
Closing for now since I have some more direction to move forward and will revisit later if necessary

PR #58478, opened 2025-07-07
add font pkgs

anushkamittal20:


font-stix-otf.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/stipub/stixfonts
+      tag: v${{package.version}}
+      expected-commit: de537a95240e0105546bed58e15b4a61be28cd89
+
+  - runs: |
+      cd ./archive/STIXv1.1.0
```
Dentrax: Why we're building `1.1.0` but not the latest `2.0.2`?

Can we have a comment for it?

anushkamittal20: You are right, let me change that

anushkamittal20: I have removed it from here, pending more discussion

Dentrax (approved):

PR #58519, opened 2025-07-07
wolfi-baselayout: prefer /usr/bin to /usr/sbin in PATH

dannf:
/usr/bin is the canonical path now that we're usrmerged.

### DIFF
--- a/wolfi-baselayout.yaml
+++ b/wolfi-baselayout.yaml
@@ -13,6 +13,10 @@ package:
       - merged-bin
       - merged-usrsbin
       - merged-lib
+  checks:
+    disabled:
+      # fails "usrlocal" linter because it creates /usr/local/lib64
+      - usrmerge
 
 environment:
   contents:
--- a/wolfi-baselayout.yaml
+++ b/wolfi-baselayout.yaml
@@ -1,7 +1,7 @@
 package:
   name: wolfi-baselayout
   version: 20230201
-  epoch: 21
+  epoch: 22
   description: "baselayout data for Wolfi"
   copyright:
     - license: MIT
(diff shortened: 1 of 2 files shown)


sergiodj (dismissed):

dannf:
@zmarano @xnox It does seem odd to me that we're linking an `/usr/local` directory (`/usr/local/lib64`) to `/lib`.  Can you confirm that was intentional? I don't want to add this linter override if it is actually correct :)

xnox:
> @zmarano @xnox It does seem odd to me that we're linking an `/usr/local` directory (`/usr/local/lib64`) to `/lib`.  Can you confirm that was intentional? I don't want to add this linter override if it is actually correct :)

Before we started /usr/local/lib64 was a symlink to a real /usr/local/lib directory.
Both of those facts should still be true.

dannf:
> > @zmarano @xnox It does seem odd to me that we're linking an `/usr/local` directory (`/usr/local/lib64`) to `/lib`.  Can you confirm that was intentional? I don't want to add this linter override if it is actually correct :)
> 
> Before we started /usr/local/lib64 was a symlink to a real /usr/local/lib directory. Both of those facts should still be true.

Ah, so it is - I just parsed it wrong:
```
4d805745b6d0:/work/packages# readlink /usr/local/lib64
lib
```
thx!

zmarano:
> @zmarano @xnox It does seem odd to me that we're linking an `/usr/local` directory (`/usr/local/lib64`) to `/lib`. Can you confirm that was intentional? I don't want to add this linter override if it is actually correct :)

This is as @xnox suggested, where `/usr/local/lib64` is a symlink to `/usr/local/lib`  

`ln -sf lib ${{targets.destdir}}/usr/local/lib64` is creating the right target and resolves to the right place.

zmarano (approved):

PR #58530, opened 2025-07-07
Fix melange microvm dependencies

a-crate:
This is used in test runs, and was split in https://github.com/wolfi-dev/os/commit/ad809da75e3257eaaf502f3e10a0e26213021c4d to its own subpackage.
I looked through qemu_runner.go and didn't see any usage of other binaries in util-linux-misc.

You can reproduce failures by doing `make test/package` and seeing ` ash: script: not found` failures, but not in this repo. The failure happens while mounting the melange cache inside of the VM, which doesn't happen in wolfi because the makefile doesn't use --cache= during tests. You can see it with `make test/linux-qemu-generic` in enterprise-packages.

### DIFF
--- a/melange.yaml
+++ b/melange.yaml
@@ -1,7 +1,7 @@
 package:
   name: melange
   version: "0.29.0"
-  epoch: 0
+  epoch: 1
   description: build APKs from source code
   copyright:
     - license: Apache-2.0
--- a/melange.yaml
+++ b/melange.yaml
@@ -37,7 +37,7 @@ subpackages:
         - mount
         - iproute2
         - openssh-server
-        - util-linux-misc
+        - script
         - e2fsprogs
     pipeline:
       - runs: |
--- a/microvm-init.yaml
+++ b/microvm-init.yaml
@@ -1,7 +1,7 @@
 package:
   name: microvm-init
   version: 0.0.1
-  epoch: 11
+  epoch: 12
   description: Minimal busybox init for microvm workloads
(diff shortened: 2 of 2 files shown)


a-crate:
microvm-init is what melange uses, but the package isn't melange specific and I'm not sure if it's used elsewhere so I opted to add the dependency there instead of replacing util-linux-misc in case it is needed for some other case.

smoser (changes requested):
I don't think this is right.  Its odd, but neither change is fully correct.

melange doesn't actually invoke script on the same system that it runs on, so the dependency isn't useful there.  It builds an initramfs that contains microvm-init and then it ssh's in and executes script.

similarly, microvm-init itself does not execute `script` any more obviously than it invokes `gcc` or `make`.  

melange does know that it needs `cmd:script` in the initramfs that it builds (https://github.com/chainguard-dev/melange/pull/2049.

@89luca89 , I'm pretty sure the correct fix is for melange to have realized that the cache it was using is invalid and to ignore it.  I'm not sure its a huge deal, since we're not changing it all that frequently at all.

89luca89:
@smoser the initramfs cache is invalidated every 24h and a `make clean` will delete it
@a-crate does this happen even after a `make clean`? if so maybe you're encountering this: https://github.com/chainguard-dev/melange/pull/2070 ?

a-crate:
Ah, yes I'm encountering that bug. Thanks Luca.

PR #58543, opened 2025-07-08
add apache-exporter package

sanjayk0508:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/apache-exporter.yaml
+++ b/apache-exporter.yaml
@@ -18,6 +18,7 @@ pipeline:
       deps: |-
         golang.org/x/crypto@v0.35.0
         golang.org/x/net@v0.38.0
+        golang.org/x/oauth2@v0.27.0
 
   - uses: go/build
     with:
--- a/apache-exporter.yaml
+++ b/apache-exporter.yaml
@@ -25,11 +25,32 @@ pipeline:
       packages: .
       output: apache_exporter
       ldflags: |
-        -X github.com/prometheus/common/version.Version=$(cat ./VERSION) \
+        -X github.com/prometheus/common/version.Version=${{package.version}} \
         -X github.com/prometheus/common/version.Revision=$(git rev-parse --short HEAD) \
         -X github.com/prometheus/common/version.Branch=$(git rev-parse --abbrev-ref HEAD) \
         -X github.com/prometheus/common/version.BuildUser=$(whoami)@$(hostname) \
         -X github.com/prometheus/common/version.BuildDate=$(date +"%Y%m%d-%H:%M:%S")
+
+subpackages:
+  - name: ${{package.name}}-compat
+    description: Compatibility package for apache-exporter container environments
+    dependencies:
+      runtime:
+        - ${{package.name}}
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/bin
+          ln -sf /usr/bin/apache_exporter ${{targets.contextdir}}/bin/apache_exporter
+    test:
+      environment:
+        contents:
+          packages:
+            - ${{package.name}}
+      pipeline:
+        - runs: |
+            stat /bin/apache_exporter
+            test "$(readlink /bin/apache_exporter)" = "/usr/bin/apache_exporter"
+            test -x /usr/bin/apache_exporter
 
 update:
   enabled: true
@@ -44,6 +65,7 @@ test:
       packages:
         - curl
         - apache2
+        - wait-for-it
   pipeline:
     - name: "Basic functionality tests"
       runs: |
@@ -60,9 +82,9 @@ test:
         post: |
           #!/bin/sh -e
           # Test that metrics endpoint is accessible even when Apache is down
-          curl -s http://localhost:9117/metrics | grep -q "apache_up"
+          curl -s --retry 3 http://localhost:9117/metrics | grep -q "apache_up"
           # Should show apache_up as 0 since Apache is not accessible
-          curl -s http://localhost:9117/metrics | grep -q "apache_up 0"
+          curl -s --retry 3 http://localhost:9117/metrics | grep -q "apache_up 0"
     - name: "Test with mock Apache server"
       uses: test/daemon-check-output
       with:
@@ -82,13 +104,7 @@ test:
           APACHE_PID=$!
           sleep 3
 
-          # Wait for Apache to be ready
-          for i in $(seq 1 10); do
-            if curl -s http://localhost/server-status?auto > /dev/null 2>&1; then
-              break
-            fi
-            sleep 1
-          done
+          wait-for-it localhost:80 -t 30 --strict
         start: apache_exporter --scrape_uri="http://localhost/server-status?auto"
         timeout: 30
         expected_output: |
@@ -97,7 +113,7 @@ test:
         post: |
           #!/bin/sh -e
           # Test that metrics endpoint is accessible and shows Apache is up
-          curl -s http://localhost:9117/metrics | grep -q "apache_up 1"
+          curl -s --retry 3 http://localhost:9117/metrics | grep -q "apache_up 1"
 
           # Cleanup Apache process
           kill $APACHE_PID 2>/dev/null || true
--- a/apache-exporter.yaml
+++ b/apache-exporter.yaml
@@ -49,9 +49,10 @@ test:
     - name: "Basic functionality tests"
       runs: |
         apache_exporter --version 2>&1 | grep -q "${{package.version}}"
-        apache_exporter --help 2>&1
+        apache_exporter --help
     - name: "Test entrypoint /bin/apache_exporter"
       runs: |
+        stat /bin/apache_exporter
         test -x /bin/apache_exporter
     - name: "Test exporter startup without Apache"
       uses: test/daemon-check-output
@@ -84,7 +85,6 @@ test:
           # Start Apache server
           httpd -D FOREGROUND > /dev/null 2>&1 &
           APACHE_PID=$!
-          sleep 3
 
           wait-for-it localhost:80 -t 30 --strict
         start: apache_exporter --scrape_uri="http://localhost/server-status?auto"
@@ -96,6 +96,3 @@ test:
           #!/bin/sh -e
           # Test that metrics endpoint is accessible and shows Apache is up
           curl -s --retry 3 http://localhost:9117/metrics | grep -q "apache_up 1"
-
-          # Cleanup Apache process
-          kill $APACHE_PID 2>/dev/null || true
--- a/apache-exporter.yaml
+++ b/apache-exporter.yaml
@@ -50,6 +50,9 @@ test:
       runs: |
         apache_exporter --version 2>&1 | grep -q "${{package.version}}"
         apache_exporter --help 2>&1
+    - name: "Test entrypoint /bin/apache_exporter"
+      runs: |
+        test -x /bin/apache_exporter
     - name: "Test exporter startup without Apache"
       uses: test/daemon-check-output
       with:
--- a/apache-exporter.yaml
+++ b/apache-exporter.yaml
@@ -84,9 +84,8 @@ test:
 
           # Start Apache server
           httpd -D FOREGROUND > /dev/null 2>&1 &
-          APACHE_PID=$!
 
-          wait-for-it localhost:80 -t 30 --strict
+          wait-for-it localhost:80 -t 10 --strict
         start: apache_exporter --scrape_uri="http://localhost/server-status?auto"
         timeout: 30
         expected_output: |
--- a/apache-exporter.yaml
+++ b/apache-exporter.yaml
@@ -1,0 +1,102 @@
+package:
+  name: apache-exporter
+  version: "1.0.10"
+  epoch: 0
+  description: apache-exporter exposes Apache server metrics for Prometheus monitoring
+  copyright:
+    - license: MIT
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/Lusitaniae/apache_exporter
+      tag: v${{package.version}}
+      expected-commit: c4206b529038d79f20f2eeba98a9780c8cb4a2c2
+
+  - uses: go/bump
+    with:
+      deps: |-
+        golang.org/x/crypto@v0.35.0
+        golang.org/x/net@v0.38.0
+
+  - uses: go/build
+    with:
+      packages: .
+      output: apache-exporter
+      ldflags: |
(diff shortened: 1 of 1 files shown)


apache-exporter.yaml
```diff
@@ -0,0 +1,103 @@
...
+        golang.org/x/crypto@v0.35.0
+        golang.org/x/net@v0.38.0
+        golang.org/x/oauth2@v0.27.0
+
+  - uses: go/build
+    with:
+      packages: .
+      output: apache_exporter
+      ldflags: |
+        -X github.com/prometheus/common/version.Version=$(cat ./VERSION) \
```
Aditevil: -X github.com/prometheus/common/version.Version=${{package.version}}

apache-exporter.yaml
```diff
@@ -0,0 +1,103 @@
...
+          APACHE_PID=$!
+          sleep 3
+
+          # Wait for Apache to be ready
+          for i in $(seq 1 10); do
+            if curl -s http://localhost/server-status?auto > /dev/null 2>&1; then
+              break
+            fi
+            sleep 1
+          done
```
Dentrax: I'd use `wait-for-it` and pass `--retry` to `curl`.

apache-exporter.yaml
```diff
@@ -0,0 +1,103 @@
...
+    with:
+      packages: .
+      output: apache_exporter
+      ldflags: |
+        -X github.com/prometheus/common/version.Version=$(cat ./VERSION) \
+        -X github.com/prometheus/common/version.Revision=$(git rev-parse --short HEAD) \
+        -X github.com/prometheus/common/version.Branch=$(git rev-parse --abbrev-ref HEAD) \
+        -X github.com/prometheus/common/version.BuildUser=$(whoami)@$(hostname) \
+        -X github.com/prometheus/common/version.BuildDate=$(date +"%Y%m%d-%H:%M:%S")
+
```
Dentrax, line 33: Needs `-compat` subpkg: https://github.com/Lusitaniae/apache_exporter/blob/634fead49017138d65f802b96c15b30c9af73aca/Dockerfile#L12

Dentrax (changes requested):

apache-exporter.yaml
sanjayk0508: if i'm not wrong we need to put our entrypoint of main package to `/bin/apache_exporter` and symlink the compat to `/bin/apache_exporter` as well?

Aditevil: By default the binary should be in` /usr/bin` I think.  But the entrypoint is `/bin/apache_exporter`, so we need a `-compat `subpkg to create a symlink

sanjayk0508: I think, if I'm not wrong, the `/bin` symlink to `/usr/bin` already handles the compatibility, so the `-compat` package isn't needed here. I tried adding `-compat`, but it caused a conflict because `/bin` was already pointing to `/usr/bin`

```
e0053c2ef053:/work/packages# ls -l /bin
lrwxrwxrwx    1 root     root             7 Jul  8 00:25 /bin -> usr/bin
```

Aditevil: A simple test to run that entrypoint(`/bin/apache_exporter`) would do then.

sanjayk0508: here's what i found, `/bin` is a symlink to `/usr/bin` but `/bin/apache_exporter` is not a symlink to `/usr/bin/apache_exporter`, and if i try to create a symlink, it conflicts because `/bin` is a symlink to `/usr/bin`, not sure what to do here

Aditevil: I'm able to run this just fine. 

```
496234fc0bb9:/work/packages# /bin/apache_exporter
time=2025-07-11T13:14:48.651Z level=INFO source=apache_exporter.go:66 msg="Starting apache_exporter" version="(version=1.0.10, branch=HEAD, revision=c4206b5)"
time=2025-07-11T13:14:48.651Z level=INFO source=apache_exporter.go:67 msg="Build context" build="(go=go1.24.5, platform=linux/amd64, user=build@aditya-tirmanwar, date=20250711-13:12:48, tags=netgo,osusergo)"
time=2025-07-11T13:14:48.651Z level=INFO source=apache_exporter.go:68 msg="Collect metrics from" scrape_uri=http://localhost/server-status/?auto
time=2025-07-11T13:14:48.655Z level=INFO source=tls_config.go:347 msg="Listening on" address=[::]:9117
time=2025-07-11T13:14:48.655Z level=INFO source=tls_config.go:350 msg="TLS is disabled." http2=false address=[::]:9117

^Ctime=2025-07-11T13:15:04.019Z level=INF
```O source=apache_exporter.go:74 msg="Caught signal. Wait 2 seconds..." sig=interrupt

Comment (Aditevil) [reply apache-exporter.yaml:33]:
```
496234fc0bb9:/work/packages# readlink -f /bin/apache_exporter
/usr/bin/apache_exporter
```

Comment (sanjayk0508) [reply apache-exporter.yaml:33]: my bad, probably used older package to debug, have added the entrypoint test now, PTAL

Comment (Aditevil) [reply apache-exporter.yaml:33]: Looks good. Nice work!

Comment (sanjayk0508) [reply apache-exporter.yaml:33]: thank you for looking into this!

**Review by Aditevil — DISMISSED** [review]

File: apache-exporter.yaml [hunk 4]
```diff
@@ -0,0 +1,101 @@
...
+    contents:
+      packages:
+        - curl
+        - apache2
+        - wait-for-it
+  pipeline:
+    - name: "Basic functionality tests"
+      runs: |
+        apache_exporter --version 2>&1 | grep -q "${{package.version}}"
+        apache_exporter --help 2>&1
```
EyeCantCU: There isn't a need to redirect output here

```suggestion
        apache_exporter --version | grep -q "${{package.version}}"
        apache_exporter --help
```

apache-exporter.yaml
```diff
@@ -0,0 +1,101 @@
...
+    - name: "Test entrypoint /bin/apache_exporter"
+      runs: |
+        test -x /bin/apache_exporter
```
EyeCantCU, line 57: This will fail silently. Please either use `stat` or set debug

apache-exporter.yaml
```diff
@@ -0,0 +1,101 @@
...
+              SetHandler server-status
+              Require all granted
+          </Location>
+          ExtendedStatus On
+          EOF
+
+          # Start Apache server
+          httpd -D FOREGROUND > /dev/null 2>&1 &
+          APACHE_PID=$!
+          sleep 3
```
EyeCantCU: This sleep is unnecessary

apache-exporter.yaml
```diff
@@ -0,0 +1,101 @@
...
+        expected_output: |
+          Starting apache_exporter
+          Listening on
+        post: |
+          #!/bin/sh -e
+          # Test that metrics endpoint is accessible and shows Apache is up
+          curl -s --retry 3 http://localhost:9117/metrics | grep -q "apache_up 1"
+
+          # Cleanup Apache process
+          kill $APACHE_PID 2>/dev/null || true
```
EyeCantCU: This is a separate shell, APACHE_PID will not be set. You should daemonize apache2 instead (in the start block) and use the exporter in post. Also, please avoid changing the return code and redirecting output. It hides valuable output and will change the outcome of our tests

EyeCantCU (dismissed):
Packaging looks good, flagged a few issues in tests

apache-exporter.yaml
sanjayk0508: weird, although the logs show the version in `stdout`, it still fails. I also checked how the version string is handled, and it seems something is being printed to `stderr`. Using `2>&1` resolves it by capturing both streams

apache-exporter.yaml
sanjayk0508: I tried daemonizing apache2 in the start (similar to production env), but given `daemon-check-output` monitors a single foreground process. When apache is daemonized, it runs in the background, which prevents the pipeline from tracking its output properly

Keeping apache in foreground will allow us to control the process and verify the exporter’s functionality, checking the `apache_up` metric, without losing visibility into apache's status. Please let me know if i'm overlooking something

apache-exporter.yaml
```diff
@@ -0,0 +1,98 @@
...
+          <Location "/server-status">
+              SetHandler server-status
+              Require all granted
+          </Location>
+          ExtendedStatus On
+          EOF
+
+          # Start Apache server
+          httpd -D FOREGROUND > /dev/null 2>&1 &
+          APACHE_PID=$!
```
Dentrax: nit: It seems `APACHE_PID` not used below, so we can remove it.

apache-exporter.yaml
```diff
@@ -0,0 +1,98 @@
...
+
+          wait-for-it localhost:80 -t 30 --strict
```
Dentrax: nit: `-t 30` would exceed the main timeout of `test/daemon-check-output` so it should be smaller than that value.

Dentrax (dismissed):
2 nits but current shape does seem good as well, so approved!

Dentrax (dismissed):

anushkamittal20 (dismissed):
lgtm, all comments are resolved

apache-exporter.yaml
```diff
@@ -0,0 +1,97 @@
...
+
+  - uses: go/build
+    with:
+      packages: .
+      output: apache_exporter
+      ldflags: |
+        -X github.com/prometheus/common/version.Version=${{package.version}} \
+        -X github.com/prometheus/common/version.Revision=$(git rev-parse --short HEAD) \
+        -X github.com/prometheus/common/version.Branch=$(git rev-parse --abbrev-ref HEAD) \
+        -X github.com/prometheus/common/version.BuildUser=$(whoami)@$(hostname) \
```
tulilirockz:
```suggestion
        -X github.com/prometheus/common/version.BuildUser=linky@chainguard.dev \
```
Lets use something static here as these can change between runners

apache-exporter.yaml
```diff
@@ -0,0 +1,97 @@
...
+    contents:
+      packages:
+        - curl
+        - apache2
+        - wait-for-it
+  pipeline:
+    - name: "Basic functionality tests"
+      runs: |
+        apache_exporter --version 2>&1 | grep -q "${{package.version}}"
+        apache_exporter --help
```
tulilirockz:
```suggestion
        set -o pipefail
        apache_exporter --version 2>&1 | grep -F -e "${{package.version}}"
        apache_exporter --help 2>&1 | grep -F -e "usage: apache_exporter"
```
Lets just make sure we dont get broken output

apache-exporter.yaml
```diff
@@ -0,0 +1,97 @@
...
+      with:
+        start: apache_exporter --scrape_uri="http://localhost:9999/server-status?auto"
+        timeout: 15
+        expected_output: |
+          Starting apache_exporter
+          Listening on
+        post: |
+          #!/bin/sh -e
+          # Test that metrics endpoint is accessible even when Apache is down
+          curl -s --retry 3 http://localhost:9117/metrics | grep -q "apache_up"
```
tulilirockz:
```suggestion
          curl -s --retry 3 http://localhost:9117/metrics | grep -F -e "apache_up"
```

apache-exporter.yaml
```diff
@@ -0,0 +1,97 @@
...
+          # Should show apache_up as 0 since Apache is not accessible
+          curl -s --retry 3 http://localhost:9117/metrics | grep -q "apache_up 0"
```
tulilirockz:
```suggestion
          curl -s --retry 3 http://localhost:9117/metrics | grep -F -e "apache_up 0"
```

apache-exporter.yaml
```diff
@@ -0,0 +1,97 @@
...
+    - name: "Test exporter startup without Apache"
+      uses: test/daemon-check-output
+      with:
+        start: apache_exporter --scrape_uri="http://localhost:9999/server-status?auto"
+        timeout: 15
+        expected_output: |
+          Starting apache_exporter
+          Listening on
+        post: |
+          #!/bin/sh -e
```
tulilirockz:
```suggestion
          set -o pipefail
```

apache-exporter.yaml
```diff
@@ -0,0 +1,97 @@
...
+  - uses: go/build
+    with:
+      packages: .
+      output: apache_exporter
+      ldflags: |
+        -X github.com/prometheus/common/version.Version=${{package.version}} \
+        -X github.com/prometheus/common/version.Revision=$(git rev-parse --short HEAD) \
+        -X github.com/prometheus/common/version.Branch=$(git rev-parse --abbrev-ref HEAD) \
+        -X github.com/prometheus/common/version.BuildUser=$(whoami)@$(hostname) \
+        -X github.com/prometheus/common/version.BuildDate=$(date +"%Y%m%d-%H:%M:%S")
```
tulilirockz:
```suggestion
        -X github.com/prometheus/common/version.BuildDate=$(date -u -d "@$SOURCE_DATE_EPOCH" +"%Y-%m-%dT%H:%M:%SZ")
```
Best to use SOURCE_DATE_EPOCH here as binaries in different dates will result in slightly differently-sized binaries if we dont do this

apache-exporter.yaml
```diff
@@ -0,0 +1,97 @@
...
+          wait-for-it localhost:80 -t 10 --strict
+        start: apache_exporter --scrape_uri="http://localhost/server-status?auto"
+        timeout: 30
+        expected_output: |
+          Starting apache_exporter
+          Listening on
+        post: |
+          #!/bin/sh -e
+          # Test that metrics endpoint is accessible and shows Apache is up
+          curl -s --retry 3 http://localhost:9117/metrics | grep -q "apache_up 1"
```
tulilirockz:
```suggestion
          # Test that metrics endpoint is accessible and shows Apache is up
          curl -s --retry 3 http://localhost:9117/metrics | grep -F -e "apache_up 1"
```

tulilirockz (dismissed):
I just have tiny nits here, mostly to just fix some reliability issues.

- There are tiny things that kind of break reproducibility
- Grepping for fixed strings is always very good so we dont accidentally use an expression instead of grepping for a string

apache-exporter.yaml
sanjayk0508: thanks for the review, have updated them now

Dentrax (approved):

jamonation (approved):

Aditevil (approved):

PR #58550, opened 2025-07-08
add font-tibetan-machine

anushkamittal20:


Dentrax (commented):
Upstream font name is `TibetanMachineUni.ttf`, whereas we put it as `TibMachUni-1.901b.ttf`, is that correct?

https://packages.debian.org/sid/all/fonts-tibetan-machine/filelist

Dentrax (approved):

PR #58580, opened 2025-07-08
ceph: fix FTBFS with cmake 4

aborrero:
Build ceph with cmake-3 as it doesn't fully support cmake 4.

See also:
 https://github.com/wolfi-dev/os/issues/58496


sergiodj (changes requested):
You forgot to bump the epoch :-).

aborrero:
> You forgot to bump the epoch :-).

thanks!

sergiodj (approved):

PR #58583, opened 2025-07-08
go-1.20: Don't depend on binutils-gold

sergiodj:
We don't need to depend on `binutils-gold` anymore.


dannf (approved):
assuming it doesn't need the patch update

sergiodj:
@dannf Thanks.  It actually already has the patch; I'd just forgotten to remove the dependency.

PR #58597, opened 2025-07-08
azure-ipam: make package reproducible

imjasonh:
Disclaimer: I have no earthly clue whether this makes the package non-functional. Dropping timestamps from the gzip output makes it reproducible, but I don't know if those are load-bearing for the actual app to work.

If this isn't okay, we can drop this change and note that we expect this package not to be reproducible.


MickeyPvX (approved):
I was able to load this into a test image and it seems to work fine 👍

PR #58604, opened 2025-07-08
postfix: Add `ld.so.conf.d` snippet

sergiodj:
While doing some melange work I noticed that `postfix-stone`'s binaries depend on libraries that are installed under `/usr/lib/postfix`, but because we don't ship a `ld.so.conf.d` for that, it's not possible to run the binaries.

Let's rectify that.  I don't feel comfortable enabling `ldd-check` just yet because melange isn't generating proper dependencies for the libraries (see https://github.com/chainguard-dev/melange/pull/2072 for a current WIP).

### DIFF
--- a/postfix.yaml
+++ b/postfix.yaml
@@ -1,7 +1,7 @@
 package:
   name: postfix
   version: "3.10.2"
-  epoch: 3
+  epoch: 4
   description: Secure and fast drop-in replacement for Sendmail (MTA)
   copyright:
     - license: IPL-1.0 OR EPL-2.0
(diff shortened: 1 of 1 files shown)


murraybd (approved):

xnox:
> While doing some melange work I noticed that `postfix-stone`'s binaries depend on libraries that are installed under `/usr/lib/postfix`, but because we don't ship a `ld.so.conf.d` for that, it's not possible to run the binaries.
> 
> Let's rectify that. I don't feel comfortable enabling `ldd-check` just yet because melange isn't generating proper dependencies for the libraries (see [chainguard-dev/melange#2072](https://github.com/chainguard-dev/melange/pull/2072) for a current WIP).

i wonder if runpath on the libraries in question and the binaries in question. Those libraries are not supposed to become public, do they?

sergiodj:
> > While doing some melange work I noticed that `postfix-stone`'s binaries depend on libraries that are installed under `/usr/lib/postfix`, but because we don't ship a `ld.so.conf.d` for that, it's not possible to run the binaries.
> > Let's rectify that. I don't feel comfortable enabling `ldd-check` just yet because melange isn't generating proper dependencies for the libraries (see [chainguard-dev/melange#2072](https://github.com/chainguard-dev/melange/pull/2072) for a current WIP).
> 
> i wonder if runpath on the libraries in question and the binaries in question. Those libraries are not supposed to become public, do they?

Yeah, I thought about runpath, too.  I'm not entirely sure if these libraries are supposed to be public or not; one of them is call `postfix-global` or some such...  Let me look at what Debian does...

OK, it seems like Debian uses `rpath`:

https://sources.debian.org/src/postfix/3.10.2-1/debian/rules/#L51

So maybe this is the way to go here as well.  I'll be leaving on PTO tomorrow, and I'm EODing now.  Let me see if I can prepare a quick PR before I leave.

PR #58605, opened 2025-07-08
re-package crossplane-aws-providers

Aditevil:
The monolithic AWS provider is deprecated as of `2024-06-12` and is no longer maintained. We need to refactor the packages to the provider family architecture. https://docs.upbound.io/providers/provider-families/

Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches


Aditevil:
NOTE: Runtime dependency on `terraform-provider-aws` was removed because it was only needed for the original monolithic build

crossplane-provider-aws.yaml
```diff
@@ -77,22 +50,50 @@ data:
...
+        with:
+          packages: ./cmd/provider/${{range.key}}
+          output: ${{range.value}}
+          ldflags: |
+            -X github.com/crossplane-contrib/provider-upjet-aws/internal/version.Version=v${{package.version}}
       - runs: |
-          up xpkg xp-extract xpkg.upbound.io/upbound/${{range.value}}:v${{package.version}}
           mkdir -p "${{targets.subpkgdir}}"
-          gunzip out.gz -c > "${{targets.subpkgdir}}"/package.yaml
+          crane export xpkg.upbound.io/upbound/${{range.value}}:v1.23.1 - | tar -xOf - package.yaml > ${{targets.subpkgdir}}/package.yaml
```
maxgio92: Shouldn't be `${{package.version}}`?
I see we have 1.23.1 here but the tag on the monorepo is 1.23.0, so I'm wondering what is their release lifecycle 🤔

Aditevil: I left a comment at the top of the manifest, to explain this. 
```
 # NOTE: When updating this package, verify the latest upstream package versions from:
# https://marketplace.upbound.io/providers
#
# Upbound only allows free tier to pull 'latest' versions, so ensure all provider
# versions in the 'data' section below are updated to match the latest available
# versions on the marketplace. "crane export xpkg.upbound.io/upbound/${{range.value}}:<update-to-latest>"
```

We can only pull latest as part of free tier and they don't cut a release for every minor version on github.

image


https://github.com/crossplane-contrib/provider-upjet-aws/releases

crossplane-provider-aws.yaml
```diff
@@ -77,22 +50,50 @@ data:
...
+      pipeline:
+        - uses: test/kwok/cluster
+        - runs: |
+            mkdir -p /tls/server
+            openssl req -x509 -newkey rsa:2048 -keyout /tls/server/tls.key \
+              -out /tls/server/tls.crt -days 1 -nodes -subj "/CN=test"
+            chmod 600 /tls/server/tls.key
+            chmod 644 /tls/server/tls.crt
+            # Run the provider with timeout and expect connection error
+            timeout 10 ${{range.value}} 2>&1 | grep -E "(logger.*provider-aws)" || echo "Provider started with expected behavior"
```
maxgio92:
```suggestion
        - name: Running the operator
          uses: test/daemon-check-output
          with:
            setup: |
              mkdir -p /tls/server
              openssl req -x509 -newkey rsa:2048 -keyout /tls/server/tls.key \
                -out /tls/server/tls.crt -days 1 -nodes -subj "/CN=test"
              chmod 600 /tls/server/tls.key
              chmod 644 /tls/server/tls.crt
            start: |
              ${{range.value}}
            timeout: 20
            expected_output: |
              "logger":"provider-aws"
              EnableBetaManagementPolicies
            post: |
              echo "Provider started with expected behavior"
```

maxgio92: We may also want to add [the CRDs](https://github.com/crossplane-contrib/provider-upjet-aws/tree/v1.23.0/package/crds) but apparently the controller manager is not logging anything after the first info log to say that the Beta feature ManagementPolicies is enabled 🫠 
So we cannot test really that more code path is executed.

crossplane-provider-aws.yaml
maxgio92: Thanks for clarifying it @Aditvil-Dev. I'd put that version into a variable to make it more visible, something like `latest_freetier_version` or `latest_version`.

crossplane-provider-aws.yaml
Aditevil: Yeah. That is as far as I could get with package tests.
Thankyou for looking into this

crossplane-provider-aws.yaml
Aditevil: Nice suggestion.

maxgio92: Sorry for the confusion. I'd like at least to pickup the same version we're building, so using `${{package.version}}`.

I'd also strive to produce the `package.yaml` which contains the CRDs, the provider configuration and some other metadata out of the sources we have, instead of getting it built.

I've tried with:
```shell
make generate
make build.all
```

but I didn't get further than image building. In order to do it we need to initialize also the git submodule that the _build_ directory references, because it contains building harness like many Make targets, among which the `build.all` for instance.

crossplane-provider-aws.yaml
```diff
@@ -1,31 +1,23 @@
...
   version: "1.23.0"
-  epoch: 0
-  description: Official AWS Provider for Crossplane by Upbound
+  epoch: 1
+  description: Official AWS Providers for Crossplane by Upbound
   copyright:
     - license: Apache-2.0
-  resources:
-    cpu: 16
-    memory: 16Gi
```
EyeCantCU: Let's restore the resources section

crossplane-provider-aws.yaml
```diff
@@ -77,22 +50,50 @@ data:
...
   enabled: true
-  github:
-    identifier: crossplane-contrib/provider-upjet-aws
+  git:
     strip-prefix: v
-
-test:
-  pipeline:
-    # AUTOGENERATED
-    - runs: |
```
EyeCantCU, line 100: We can use `test/emptypackage` here

crossplane-provider-aws.yaml
EyeCantCU: If `latest` isn't an option, we should provide a variable towards the top of the package manifest

EyeCantCU (changes requested):

crossplane-provider-aws.yaml
```diff
@@ -80,22 +56,63 @@ data:
...
+      runtime:
+        - ${{package.name}}-${{range.key}}
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}"/usr/local/bin
+          ln -sf /usr/bin/${{range.value}} "${{targets.subpkgdir}}"/usr/local/bin/provider
+    test:
+      pipeline:
+        - runs: |
+            readlink -f /usr/local/bin/provider | grep "${{range.value}}"
```
maxgio92: We can use `test/tw/symlink-check` here

PR #58651, opened 2025-07-09
add font stix-otf

anushkamittal20:
Note: Buiding only for amd as currently there are issues building for arm.

### DIFF
--- a/font-stix-otf.yaml
+++ b/font-stix-otf.yaml
@@ -24,7 +24,7 @@ pipeline:
       expected-commit: de537a95240e0105546bed58e15b4a61be28cd89
 
   - runs: |
-      pip install cmake==3.22.6 
+      pip install cmake==3.22.6
       pip install -r requirements.txt
       python tools/tirobuild.py STIXbuild.yml
       mkdir -p "${{targets.contextdir}}/usr/share/fonts/stix-fonts/"
--- a/font-stix-otf.yaml
+++ b/font-stix-otf.yaml
@@ -1,6 +1,6 @@
 package:
   name: font-stix-otf
-  version: 2.14
+  version: 1.1.0
   epoch: 0
   description: telugu fonts
   copyright:
@@ -12,7 +12,7 @@ environment:
       - build-base
       - busybox
 
-# not the latest 2.0.2; but is the requested version
+# not the latest 2.0.2; we will bump to latest once this is in
 pipeline:
   - uses: git-checkout
     with:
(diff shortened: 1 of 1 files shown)


font-stix-otf.yaml
```diff
@@ -0,0 +1,34 @@
+package:
+  name: font-stix-otf
+  version: 1.1.0
```
Dentrax:
```suggestion
  version: 2.14
```

font-stix-otf.yaml
```diff
@@ -0,0 +1,34 @@
...
+    packages:
+      - build-base
+      - busybox
+
+# not the latest 2.14; we will bump to latest once this is in
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/stipub/stixfonts
+      tag: v2.14
```
Dentrax:
```suggestion
      tag: v${{package.version}}
```

Dentrax (changes requested):

font-stix-otf.yaml
```diff
@@ -0,0 +1,34 @@
...
+      expected-commit: de537a95240e0105546bed58e15b4a61be28cd89
+  - runs: |
```
Dentrax, line 30: As upstream supports multiple fonts: https://github.com/stipub/stixfonts/tree/master/fonts

FWIW we should use `- range` here build all of it as subpackages.

Dentrax (changes requested):

font-stix-otf.yaml
Dentrax: master no longer provides fonts dir (Thanks to Anushka for vis)

font-stix-otf.yaml
```diff
@@ -0,0 +1,42 @@
+package:
+  name: font-stix-otf
+  version: 2.14
+  epoch: 0
+  description: telugu fonts
+  copyright:
+    - license: OFL-1.1
+ target-architecture:
+    - x86_64
```
Dentrax, line 10: Can we have a comment here why this is only enabled for x86_64?

font-stix-otf.yaml
```diff
@@ -0,0 +1,42 @@
...
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - cargo-c
+      - maturin
+      - py3-pip
+      - python3
+      - python3-dev
+      - rust
```
Dentrax: rust and cargo is needed here?

font-stix-otf.yaml
anushkamittal20: Hey Furkan, I was just trying things out. the build started failing for me after the most recent changes

font-stix-otf.yaml
anushkamittal20: not really, I am refactoring the PR atm

font-stix-otf.yaml
Dentrax: is this still the case? if so, it'd better to drop a comment here for the reasoning.

anushkamittal20: I have now added a comment, there is an ongoing discussion but as of now: cffsubr doesnt build on aarch64 (I think it is a cmake issue)

Dentrax (approved):

PR #58653, opened 2025-07-09
llvm 15 renames

javacruft:
- **dotnet-6: update LLVM 15 package references**
- **dotnet-7: update LLVM 15 package references**
- **dotnet-8: update LLVM 15 package references**
- **dotnet-9: update LLVM 15 package references**
- **datawire-envoy-1.31: update LLVM 15 package references**


dannf (dismissed):

dotnet-7.yaml
```diff
@@ -23,7 +23,6 @@ environment:
       - busybox
       - ca-certificates-bundle
       - clang-15
-      - clang-15-default
```
sil2100: I think this one might have been missed. We'd need an epoch bump here + the renames, since the llvm packages were left untouched.

sil2100 (changes requested):
All good, just missed dotnet-7 changes

dotnet-7.yaml
javacruft: Urgh - this was hangover from me splitting out dotnet-7 as it FTBFS

sil2100 (approved):

PR #58708, opened 2025-07-09
dotnet-7: update LLVM 15 package references

javacruft:
Replace llvm15, llvm15-cmake-default, llvm15-dev, llvm15-tools with llvm-15-dev

### DIFF
--- a/dotnet-7.yaml
+++ b/dotnet-7.yaml
@@ -2,7 +2,7 @@ package:
   name: dotnet-7
   # Remove `-sb` from the git-checkout tag at the next release
   version: 7.0.120
-  epoch: 1
+  epoch: 2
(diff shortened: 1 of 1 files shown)


dotnet-7.yaml
```diff
@@ -23,17 +23,13 @@ environment:
       - busybox
       - ca-certificates-bundle
       - clang-15
-      - clang-15-default
```
dannf, line 26: I wondered why this was safe, but I see that clang-15 includes `/usr/bin/clang`.  Should we instead follow the `gcc` pattern and allow clang versions to be co-installable, with `-default` packages that symlink things in?

dannf (approved):

PR #58725, opened 2025-07-09
Shellcheck fix 1

AbduAzim424:
Fix some `shellcheck` warnings before enabling it globally.


luajit.yaml
```diff
@@ -37,7 +37,7 @@ pipeline:
       export LUA_LIB_DIR="$LUAJIT_LIB/lua"
       export LUAJIT_INC=/usr/include/luajit-${{vars.major-minor-version}}
 
-      make CCDEBUG=-g PREFIX=/usr -j $(nproc)
+      make CCDEBUG=-g PREFIX=/usr -j "$(nproc)"
```
smoser: shell is such a trainwreck.
these two lines are not actually identical in the case that 'nproc' fails.
and because 'nproc' is used inline as it is, nothing will catch it's failure.

so if nproc were to fail, before you got 'make -j' which happens to mean  'no limit'.
now you'll get 'make -j ""' which will fail with `make: the '-j' option requires a positive integer argument`.

Other places in this pr, you have 'make "-j$(nproc)"' which then ends with the existing behavior (which is _still_ probably not what people wanted, but does not change behavior).

So....
tldr; I think just make this 'make "-j$(nproc)"' but the right fix is probably for melange to set  and guarantee an environment variable or something.

logstash-integration-jdbc.yaml
```diff
@@ -57,7 +57,7 @@ test:
         - jruby-9.4
   pipeline:
     - runs: |
-        logstash-plugin install $(find / -type f -name "${{package.name}}-*.gem")
+        logstash-plugin install "$(find / -type f -name "${{package.name}}-*.gem")"
```
smoser: this could be wrong now.  you _were_ going to call `logstash-plugin install` with multiple arguments , but now you're going to call it with only one argument containing multiple paths.

jellyfin.yaml
```diff
@@ -105,7 +105,7 @@ test:
           echo "Test failed: Jellyfin is not listening on port 8096."
           echo "----- Jellyfin Debug Logs -----"
           cat /tmp/jellyfin.log
-          kill $JELLYFIN_PID || true
+          kill "$JELLYFIN_PID" || true
```
smoser, line 108: fwiw, i'm 90% sure that JELLYFIN_PID here will never have a value, because shell variables are not going to be carried across 'runs' sections.

gengetopt.yaml
```diff
@@ -90,7 +90,7 @@ test:
             option "dependant" D "Depends on str-opt" int optional dependon="str-opt"
         ' > test.ggo
         gengetopt --input=test.ggo
-        ls -l | grep cmdline.c
+        ls -l cmdline.c 2>/dev/null || echo "cmdline.c not found"
```
smoser: this _was_ going to fail if there was no 'cmdline.c' file.  (because grep would have not found 'cmdline.c' in 'ls -l' output and woudl have exited non-zero).

Now you're guaranteed to swallow that error with a 'cmdline.c not found' message that no one will ever read. I don't think that is what was intended.

if we just want to see if 'cmdline.c' exists, 
`[ -f cmdline.c ]` or more helpfully:
`[ -f "cmdline.c" ] || { echo "FATAL: cmdline.c did not exist"; exit 1; }'`

jellyfin.yaml
AbduAzim424: Yes you're 100% right and therefore I merged these 2 test blocks.

datadog-agent.yaml
```diff
@@ -546,33 +546,33 @@ subpackages:
...
-            s6-fillurandompool --help 2>&1 >/tmp/datadog-agent.log &
-            s6-format-filter --version 2>&1 >/tmp/datadog-agent.log &
-            s6-format-filter --help 2>&1 >/tmp/datadog-agent.log &
-            s6-head --version 2>&1 >/tmp/datadog-agent.log &
-            s6-head --help 2>&1 >/tmp/datadog-agent.log &
-            s6-ls version 2>&1 >/tmp/datadog-agent.log &
-            s6-ls help 2>&1 >/tmp/datadog-agent.log &
-            s6-mkdir --version 2>&1 >/tmp/datadog-agent.log &
+            s6-cat --version >/tmp/datadog-agent.log 2>&1
+            s6-cat --help >/tmp/datadog-agent.log 2>&1 &
```
smoser, line 550: the trailng `&` is almost certainly a typo. If you background these things, you 
1. background them
2. swallow any error they would give

(you didn't add the `&`, but you carried it forward)

AbduAzim424: Thank you for looking into this

PR #58742, opened 2025-07-10
go-1.2*: Improve tests

sergiodj:
We need to expand the testsuite used for the Golang toolchain in order to:

- Make sure the compiler can build a real project (let's use "fzf" because it's relatively simple and fast to compile).

- Test the usage of an external linker, particularly because of our OpenSSF compiler flags.

Relates: https://github.com/chainguard-dev/internal-dev/issues/14796
Relates: #58493

### DIFF
--- a/go-1.20.yaml
+++ b/go-1.20.yaml
@@ -1,7 +1,7 @@
 package:
   name: go-1.20
   version: 1.20.14
-  epoch: 6
+  epoch: 7
   description: "the Go programming language"
   copyright:
     - license: BSD-3-Clause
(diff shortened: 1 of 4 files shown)


sil2100 (approved):
This is good. It's always tricky to rely on an external resource for testing, but I think we do that either way.

sergiodj:
Thanks, @sil2100.  Indeed, I'm not overly happy with using external resources for testing either, but since there's precedence I decided to go along with it.

I'll prepare the same tests for our enterprise Golang toolchain.

PR #58877, opened 2025-07-10
initial build for ratify package

developer-guy:
Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/ratify.yaml
+++ b/ratify.yaml
@@ -1,0 +1,156 @@
+package:
+  name: ratify
+  version: "1.4.0"
+  epoch: 0
+  description: Artifact Ratification Framework (CNCF Sandbox)
+  copyright:
+    - license: Apache-2.0
+
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: f3f6229dee1051e4beb580ec17ad68b7a45f23af
+      repository: https://github.com/notaryproject/ratify
+      tag: v${{package.version}}
+
+  - uses: go/bump
+    with:
+      deps: |-
+        golang.org/x/crypto@v0.35.0
+        golang.org/x/oauth2@v0.27.0
+        golang.org/x/net@v0.38.0
+        github.com/go-jose/go-jose/v3@v3.0.4
+        github.com/go-jose/go-jose/v4@v4.0.5
+        github.com/golang-jwt/jwt/v4@v4.5.2
+        github.com/golang-jwt/jwt/v5@v5.2.2
+        github.com/cloudflare/circl@v1.6.1
+        github.com/open-policy-agent/opa@v1.4.0
+        github.com/go-viper/mapstructure/v2@v2.3.0
+
+  - uses: go/build
+    with:
+      ldflags: |
+        -X github.com/ratify-project/ratify/internal/version.GitCommitHash=$(git rev-parse HEAD)
+        -X github.com/ratify-project/ratify/internal/version.GitTag=${{package.version}}
+      output: ratify
+      packages: ./cmd/ratify
+
+  - runs: |
+      mkdir -p ${{targets.contextdir}}/opt/ratify
+      install -Dpm644 -t "${{targets.contextdir}}/opt/ratify/" config/config.json
+
+data:
+  - name: plugins
+    items:
+      sbom:
+      licensechecker:
+      schemavalidator:
+      vulnerabilityreport:
+
+subpackages:
+  - name: ${{package.name}}-compat
+    description: Compatibility package for the ${{package.name}}
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/app
+          ln -sf /usr/bin/ratify ${{targets.contextdir}}/app/ratify
+    test:
+      pipeline:
+        - runs: test "$(readlink /app/ratify)" = "/usr/bin/ratify"
+
+  - range: plugins
+    name: ${{package.name}}-${{range.key}}
+    description: Ratify ${{range.key}} plugin
+    pipeline:
+      - uses: go/build
+        with:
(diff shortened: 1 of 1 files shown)


ratify.yaml
```diff
@@ -0,0 +1,157 @@
...
+        github.com/golang-jwt/jwt/v5@v5.2.2
+        github.com/cloudflare/circl@v1.6.1
+        github.com/open-policy-agent/opa@v1.4.0
+        github.com/go-viper/mapstructure/v2@v2.3.0
+
+  - uses: go/build
+    with:
+      ldflags: |
+        -X github.com/ratify-project/ratify/internal/version.GitCommitHash=$(git rev-parse HEAD)
+        -X github.com/ratify-project/ratify/internal/version.GitTag=${{package.version}}
```
anushkamittal20, line 34: should we add: -X github.com/ratify-project/ratify/internal/version.GitTreeState=$(git diff --quiet && echo clean || echo dirty) as done here: https://github.com/notaryproject/ratify/blob/main/Makefile#L25C24-L25C54

developer-guy: thanks for spotting this out actually but if you look at the goreleaser file https://github.com/notaryproject/ratify/blob/f3f6229dee1051e4beb580ec17ad68b7a45f23af/.goreleaser.yml#L23 you will see that they actually do not use GitTreeState variable and they use goreleaser to release the project https://github.com/notaryproject/ratify/blob/f3f6229dee1051e4beb580ec17ad68b7a45f23af/.github/workflows/release.yml#L36-L43

ratify.yaml
```diff
@@ -0,0 +1,157 @@
...
+      output: ratify
+      packages: ./cmd/ratify
+
+  - runs: |
+      mkdir -p ${{targets.contextdir}}/opt/ratify
+      install -Dm644 config/config.json ${{targets.contextdir}}/opt/ratify/config.json
```
tulilirockz:
```suggestion
      install -Dpm644 -t "${{targets.contextdir}}/opt/ratify/" config/config.json
```
Mostly just a best-practice thing :P Doesnt matter much though

ratify.yaml
```diff
@@ -0,0 +1,157 @@
...
+  - range: plugins
+    name: ${{package.name}}-${{range.key}}-compat
+    description: Ratify ${{range.key}} compatibility package
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/opt/ratify/plugins
+          ln -sf /usr/bin/${{range.key}} ${{targets.contextdir}}/opt/ratify/plugins/${{range.key}}
+    test:
+      pipeline:
+        - runs: test "$(readlink /opt/ratify/plugins/${{range.key}})" = "/usr/bin/${{range.key}}"
```
tulilirockz, line 79: Amazing job on this! Chef's kiss

ratify.yaml
```diff
@@ -0,0 +1,157 @@
...
+    strip-prefix: v
+
+test:
+  environment:
+    contents:
+      packages:
+        - curl
+  pipeline:
+    - runs: |
+        curl -sSLO https://raw.githubusercontent.com/deislabs/ratify/main/test/testdata/notation.crt
```
tulilirockz: I think we can use our own self-signed certificate here instead, `mkcert -cert-file cert.pem -key-file key.pem localhost 127.0.0.1 ::1` should do it for you

tulilirockz:
```suggestion
        curl -sSLO https://raw.githubusercontent.com/notaryproject/ratify/cb7ffd11a99446731468de435b43f44d4afeaa49/test/testdata/notation.crt
```
If we absolutely cant use a self-signed certificate we should at least pin this to a specific commit just to make sure this wont blow up due to repository restructuring

ratify.yaml
```diff
@@ -0,0 +1,157 @@
...
+                        }
+                    }
+                ]
+            }
+        }
+        EOF
+
+        ratify verify -c config.json -s ghcr.io/deislabs/ratify/notary-image:signed > verificationResult.json
+
+        if ! grep -q '"isSuccess": true' verificationResult.json; then
```
tulilirockz:
```suggestion
        if ! ratify verify -c config.json -s ghcr.io/deislabs/ratify/notary-image:signed | grep -F -e '"isSuccess": true'; then
```
Might be a bit cleaner + we should use `-F -e` when possible because we are writing some stuff we dont want to be misinterpreted by bash

ratify.yaml
mamccorm: Does goreleaser add that header already? As we're not using goreleaser we might need to add it. Can you check the upstream binary vs ours and see if we're missing any ldflags

tulilirockz (changes requested):
I have a tiny nit + something I'd like some comments on, everything looks great, though! Amazing job with this package, just learned something new today :)

ratify.yaml
developer-guy: yes I just checked that the version command's output is matching with the one we built.

ratify.yaml
developer-guy: I'm always open for best-practices, so thank you!

ratify.yaml
developer-guy: I was following the documentation here: https://ratify.dev/docs/next/quickstarts/ratify-cli so I have just added it as note to the tests.

developer-guy: yes I fixed it with the one mentioned in the documentation, thank you!

ratify.yaml
developer-guy: thank you, yes it looks more cleaner for sure!

tulilirockz (approved):

PR #58879, opened 2025-07-10
add additonal ldflags for cis-operator

rakshitgondwal:
This PR adds additional `ldflags` for `cis-operator` package.


kranurag7 (commented):
Inside tests; we can invoke `version` or `--version` if it's available to and `| grep ${{package.version}}` to ensure ldflags are being preserved and are correct in b/w package bumps.

kranurag7 (approved):

PR #58933, opened 2025-07-10
Include SCAP content for all supported RHEL products in SSG package

egibs:
The RHEL products were originally excluded from our SSG package due to compilation errors but this is no longer the case (and `RHEl7` is no longer supported).

RHEL8-10 will now be included as opposed to just RHEL10.


stevebeattie (approved):
Thanks!

PR #58938, opened 2025-07-10
postfix: Use `rpath` instead of `ld.so.conf.d` for internal libraries

sergiodj:
@xnox pointed out that postfix libraries might be internal rather than public, and after checking the Debian package I noticed that he's right.

Postfix even provides a configuration option to specify what the `rpath` should be when linking binaries, so let's use that instead of shipping an `ld.so.conf.d` snippet.

I've also added a runtime dependency on `postfix` to `postfix-stone`, otherwise its binaries won't run.

Relates: https://github.com/chainguard-dev/melange/pull/2072
Relates: #58604

### DIFF
--- a/postfix.yaml
+++ b/postfix.yaml
@@ -1,7 +1,7 @@
 package:
   name: postfix
   version: "3.10.3"
-  epoch: 0
+  epoch: 1
   description: Secure and fast drop-in replacement for Sendmail (MTA)
   copyright:
     - license: IPL-1.0 OR EPL-2.0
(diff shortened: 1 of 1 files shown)


postfix.yaml
```diff
@@ -145,6 +145,7 @@ pipeline:
       	meta_directory=/etc/postfix \
       	daemon_directory=/usr/libexec/postfix \
       	shlib_directory=/usr/lib/postfix \
+        SHLIB_RPATH="-Wl,-rpath,/usr/lib/postfix" \
```
xnox: Odd indentation visible in the diff.

xnox (dismissed):

postfix.yaml
sergiodj: Huh, the odd indentation is actually in the file.  I've adjusted the PR.

dannf (approved):

PR #58954, opened 2025-07-10
feat(dotnet): Support all SDK feature bands

EyeCantCU:
Previously, we were not building all feature bands for the .NET SDK. This adds support for all supported feature bands and cleans up our packaging for .NET

Please see Microsoft's documentation covering the supported versions of the .NET SDK here:

https://learn.microsoft.com/en-us/dotnet/core/porting/versioning-sdk-msbuild-vs#targeting-and-support-rules

Changes for existing packages -

dotnet-bootstrap:
- Do not generate dependencies with SCA. Previously, we removed a library to work around an issue where installing the bootstrap'd SDK would fail because it depended on an ABI version unavailable in Wolfi. Instead of going through this hassle every time resolution fails, just don't generate dependencies. We include the ones we need at buildtime anyway and pray things "just work"
- If things don't "just work", the build will in a clear way
- Remove buildtime dependencies that were never used at all

dotnet:
- Don't include the SDK as we need to provide feature bands. The whole point of this PR
- Stop depending on the exact version of .NET used for bootstrap as an older SDK is retrieved anyway (9.0.106). If bootstrap needs to be bumped, the build will fail in an obvious way
- Drop unused buildtime dependencies
- Bump to latest Clang
- Remove default packages that appear to serve no purpose and migrate the virtual provides they had to the actual packages that provide them


dotnet-8.yaml
```diff
@@ -14,176 +14,134 @@ package:
...
     pipeline:
       - runs: |
-          mkdir -p "${{targets.subpkgdir}}"/usr/share/dotnet/shared
-          mv "${{targets.destdir}}"/usr/share/dotnet/shared/Microsoft.NETCore.App "${{targets.subpkgdir}}"/usr/share/dotnet/shared/
-    dependencies:
-      runtime:
-        - dotnet-8
-    test:
-      pipeline:
-        - uses: test/tw/ldd-check
```
javacruft, line 97: I think the test pipeline got dropped - was that intentional?

EyeCantCU: Not intentional at all. Restored

dotnet-9.yaml
```diff
@@ -12,7 +12,13 @@ package:
     runtime:
       - icu
     provides:
-      - dotnet=${{package.version}}
+      - dotnet=${{package.full-version}}
```
sergiodj, line 15: I'm assuming this change is intentional because I see a similar one down below.  Out of curiosity, what's the benefit of tightening the `provides` to `package.full-version` instead of using the regular `package.version`?

EyeCantCU: It's only slightly more granular and aligns with versioning for "real" provides. It would be useful if I wanted to install a specific version of `dotnet` using the virtual provides without caring about the version stream. For example, if I wanted to test changes between 9.0.7-r1 and 9.0.7-r2, I can resolve to `dotnet~9.0.7-r1/r2` whereas right now that would have to be done by providing the full package name + version. Are people actually doing that though? Maybe, probably not, but behaviorally it is more consistent

sergiodj (approved):
LGTM to the extent that I understand the dotnet ecosystem :-).  Thanks for doing this!

EyeCantCU:
Thank you for reviewing!

dotnet-9.yaml
```diff
@@ -50,122 +48,102 @@ pipeline:
...
-
-          for i in sdk sdk-manifests templates; do
-            mv "${{targets.destdir}}"/usr/share/dotnet/$i "${{targets.contextdir}}"/usr/share/dotnet/
-          done
-    dependencies:
-      runtime:
-        - dotnet-9-runtime
-        - aspnet-9-runtime
-        - dotnet-9-targeting-pack
-        - aspnet-9-targeting-pack
```
vishal-chdhry, line 149: @EyeCantCU I see that we have removed `aspnet-9-runtime` as a runtime dependency of dotnet-9-sdk.

That causes a change in existing behaviour: https://github.com/chainguard-dev/prodsec/issues/386

dotnet-sdk-9.0.3.yaml
```diff
@@ -0,0 +1,127 @@
...
+  dependencies:
+    # Only retain for latest feature band as one may want to install
+    # multiple bands at once
+    provides:
+      - dotnet-sdk=${{package.full-version}}
+      - dotnet-${{vars.major-version}}-sdk=${{package.full-version}}
+    runtime:
+      - dotnet-${{vars.major-version}}-runtime
+      - dotnet-${{vars.major-version}}-targeting-pack
+      - icu
```
vishal-chdhry, line 20: `aspnet` runtime dependencies are not present here

PR #59006, opened 2025-07-11
Updated crossplane-provider-keycloak to 1.9.2

AmberArcadia:
Update check issue is known and will be fixed after the intermediary versions are merged as per the escalation.


powersj (approved):
+1 Assuming CI passes :)

AmberArcadia:
Worked last time so it better not flake haha

PR #59146, opened 2025-07-14
add py3-msgspec package

kranurag7:

### DIFF
--- a/py3-msgspec.yaml
+++ b/py3-msgspec.yaml
@@ -35,10 +35,6 @@ subpackages:
   - range: py-versions
     name: py${{range.key}}-${{vars.pypi-package}}
     description: ${{vars.pypi-package}} installed for python${{range.key}}
-    dependencies:
-      provides:
-        - py3-${{vars.pypi-package}}
-      provider-priority: ${{range.value}}
     pipeline:
       - uses: py/pip-build-install
         with:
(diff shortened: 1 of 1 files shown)


py3-msgspec.yaml
```diff
@@ -0,0 +1,61 @@
...
+        with:
+          python: python${{range.key}}
+      - uses: strip
+    test:
+      pipeline:
+        - uses: python/import
+          with:
+            python: python${{range.key}}
+            import: ${{vars.pypi-package}}
+
```
AnmolVirdi, line 66: We can add a minimal functional test for the library (Not a blocker):
```suggestion

        - name: Test some simple date/time functions
          runs: |
            cat >> test.py <<EOF
            import ${{vars.pypi-package}}

            class User(msgspec.Struct):
                id: int
                name: str
                email: str
                active: bool = True
                
            user = User(id=1, name="test", email="test@example.com")
            encoded = msgspec.json.encode(user)
            print("Encoded JSON:", encoded)
            EOF

            python${{range.key}} test.py | grep -q '"id":1,"name":"test","email":"test@example.com","active":true'
```

kranurag7: ack, thanks for the suggestion. Most of these py3-* packages are temporary and will be replaced soon in few weeks once we've libraries available to us via pipelines.

AnmolVirdi (approved):
LGTM.

PR #54531, opened 2025-05-28
feat(virtualpackage.yaml): Add new virtualpackage test pipeline with example usage

philroche:
Based on provider-priority and the same name provided by other packages, it was determined we have many virtual packages that should never be installed.

Instead one of the providing packages should be installed.

This package test ensures this remains true and does not regress.

Virtual package is defined as a package that has another package providing the exact same name with a higher provider priority.

This is tracked @ https://github.com/wolfi-dev/os/issues/44820 and internal to chainguard @ https://github.com/chainguard-dev/internal-dev/issues/12573

Signed-off-by: philroche <phil.roche@chainguard.dev>

PR #59164, opened 2025-07-14
erl{25-28}-rebar3: set provides rebar3 + provider priority based on erlang version

javacruft:
Ensure that the most recent Erlang rebar3 package is used when a package just declares a dependency on rebar3.

### DIFF
--- a/electric.yaml
+++ b/electric.yaml
@@ -1,7 +1,7 @@
 package:
   name: electric
   version: "1.0.24"
-  epoch: 0
+  epoch: 1
   description: General-purpose programming language and runtime environment
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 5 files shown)


aborrero:
might be a good candidate to introduce the new virtualpackage tests, see https://github.com/wolfi-dev/os/pull/54531

javacruft:
> might be a good candidate to introduce the new virtualpackage tests, see #54531

@aborrero added as requested - should I be checking for the real package name as well?

aborrero:
> > might be a good candidate to introduce the new virtualpackage tests, see #54531
> 
> @aborrero added as requested - should I be checking for the real package name as well?

Yes, if you add this:

```
  real-pkg-name: "${{package.name}}"
```

It would be even a more robust check.

aborrero (dismissed):
LGTM.

aborrero (approved):
LGTM.

PR #59166, opened 2025-07-14
pipelines: add new metapackage test

aborrero:
Add new metapackage test pipeline, to verify that a meta package complies with three things:
 * it is empty
 * it has at least 1 runtime dependency
 * it contains the 'meta' keyword in the description

While at it, add the new test to the py3-fsspec package to demonstrate usage.

### DIFF
--- a/pipelines/test/metapackage.yaml
+++ b/pipelines/test/metapackage.yaml
@@ -1,0 +1,39 @@
+name: metapackage
+
+# A metapackage is an empty package that only declares dependencies on other packages.
+# This package test ensures a metapackage remains correct and does not regress.
+#
+# Please note the difference with virtual packages. They both act as kind of "shortcuts"
+# to get other packages installed, but with important nuances:
+# Virtual packages don't get installed. Real packages relate to virtual packages via 'provides' directive.
+# Metapackages do get installed, but they are empty. They relate to the actual packages via normal runtime dependency
+pipeline:
+  - uses: test/emptypackage
+  - name: metapackage keyword validation
+    description: Checks that a metapackage description contains the 'meta' keyword
+    runs: |
+      pkg="${{context.name}}"
+
+      # the tail trick here skips the title line and leaves only the description
+      description=$(apk info --installed --description $pkg | tail -n +2)
+
+      if ! echo "$description" | grep -qi "meta"; then
+        echo "FAIL: metapackage [$pkg] description does not contain 'meta' keyword"
+        echo "Description: $description"
+        exit 1
+      fi
+      echo "PASS: metapackage [$pkg] description contains 'meta' keyword"
+  - name: metapackage dependency validation
+    description: Checks that a metapackage contains at least one runtime dependency
+    runs: |
+      pkg="${{context.name}}"
+
+      # the tail trick here skips the title line and leaves only the lines for dependencies
+      # then we remove empty lines, and count remaining lines
+      depends_count=$(apk info --installed --depends $pkg | tail -n +2 | grep -v ^$ | wc -l)
+
+      if [ "${depends_count}" == "0" ]; then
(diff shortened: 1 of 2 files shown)


pipelines/test/metapackage.yaml
```diff
@@ -0,0 +1,29 @@
...
+        echo "FAIL: metapackage [$pkg] description does not contain 'meta' keyword"
+        echo "Description: $description"
+        exit 1
+      fi
+      echo "PASS: metapackage [$pkg] description contains 'meta' keyword"
+  - name: metapackage dependency validation
+    description: Checks that a metapackage contains at least one runtime dependency
+    runs: |
+      pkg=$(basename ${{targets.contextdir}})
+      depends_count=$(apk info -e -R $pkg | tail -n +2 | grep -v ^$ | wc -l)
```
philroche: This is tricky to understand. What is being checked here? Can we use long form options if available to make it easier to understand?

pipelines/test/metapackage.yaml
```diff
@@ -0,0 +1,29 @@
+name: metapackage
+
+# A metapackage is an empty package that only declares dependencies on other packages.
+# This package test ensures a metapackage remains correct and does not regress.
```
philroche, line 4: Meta and vitual package distinction often confuses me. As we now have virtual package test @ https://github.com/wolfi-dev/os/pull/54531/files . Can you ad a sentence here on how meta differs from virtual?

pipelines/test/metapackage.yaml
```diff
@@ -0,0 +1,29 @@
...
+pipeline:
+  - uses: test/emptypackage
+  - name: metapackage keyword validation
+    description: Checks that a metapackage description contains the 'meta' keyword
+    runs: |
+      pkg=$(basename ${{targets.contextdir}})
+      description=$(apk info -e -d $pkg | tail -n +2)
```
philroche: Can we test this in an easier way by testing the melange definition or is there value in checking that it is present in the APK info too?

pipelines/test/metapackage.yaml
philroche: Can we test this in an easier way by testing the melange definition or is there value in checking that it is present in the APK info too?

pipelines/test/metapackage.yaml
aborrero: metapackages are meant to be installed, so testing the package "as-installed" provides a tiny bit of added value compared to only checking the melange definition.

pipelines/test/metapackage.yaml
aborrero: Melange doesn't expose this as template variables.

Also, see the other comments about the added value of testing packages "as-installed".

pipelines/test/metapackage.yaml
```diff
@@ -0,0 +1,39 @@
+name: metapackage
+
+# A metapackage is an empty package that only declares dependencies on other packages.
+# This package test ensures a metapackage remains correct and does not regress.
+#
+# Please note the difference with virtual packages. They both act as kind of "shortcuts"
+# to get other packages installed, but with important nuances:
+# Virtual packages don't get installed. Real packages related to virtual packages via 'prodives' directive.
```
javacruft: Typo

```suggestion
# Virtual packages don't get installed. Real packages related to virtual packages via 'provides' directive.
```

javacruft (changes requested):
One minor typo but other than that looks good to me.

javacruft (approved):

pipelines/test/metapackage.yaml
philroche: +1

pipelines/test/metapackage.yaml
philroche: +1

PR #59174, opened 2025-07-14
Update ruby3.2-aws-eventstream to 1.4.0

debasishbsws:
This PR updates ruby3.2-aws-eventstream to the latest version 1.4.0.

Changes made:
- Updated package version to 1.4.0
- Updated expected commit SHA if applicable

Generated automatically by package update script.


AmberArcadia (changes requested):
I would probably make a few adjustments to the script. As much as I like a "hello", that's probably superfluous haha. I also ran it and my terminal exited, so maybe we have a different failure behavior as I didn't even get an opportunity to see what went wrong. Also if it went wrong at all then that also seems problematic.

PR #59211, opened 2025-07-14
Create symlink for metrics-server binary at root of filesystem

mamccorm:
Creates root level symlinks to the binaries, for closer alignment with other deployment mechanisms, which may expect to find the binaries here

### DIFF
--- a/coredns.yaml
+++ b/coredns.yaml
@@ -47,7 +47,7 @@ subpackages:
             set -o pipefail
             readlink -f /coredns | grep /usr/bin/coredns
             /coredns --version
-            coredns --help
+            /coredns --help
 
   - name: kuma-coredns
     description: CoreDNS with plugins used by Kuma
--- a/coredns.yaml
+++ b/coredns.yaml
@@ -1,7 +1,7 @@
 package:
   name: coredns
   version: "1.12.2"
-  epoch: 0
+  epoch: 1
   description: CoreDNS is a DNS server that chains plugins
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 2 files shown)


metrics-server.yaml
```diff
@@ -36,10 +36,26 @@ pipeline:
...
+    test:
+      environment:
+        contents:
+          packages:
+            - ${{package.name}}
+      pipeline:
+        - runs: |
+            readlink -f /metrics-server | grep /usr/bin/metrics-server
+            /metrics-server --help
+
```
tulilirockz:
```suggestion
        - runs: |
            set -o pipefail
            readlink -f /metrics-server | grep -F -e "/usr/bin/metrics-server"
            /metrics-server --help | grep -F -e (output from the thing)
            # maybe also a --version here too!

```

coredns.yaml
```diff
@@ -32,6 +32,22 @@ pipeline:
...
+    test:
+      environment:
+        contents:
+          packages:
+            - ${{package.name}}
+      pipeline:
+        - runs: |
+            readlink -f /coredns | grep /usr/bin/coredns
+            /coredns -version
+
```
tulilirockz:
```suggestion
        - runs: |
            set -o pipefail
            readlink -f /coredns | grep -F -e "/usr/bin/coredns"
            /coredns -version | grep -F -e "${{package.version}}"

```

metrics-server.yaml
```diff
@@ -36,10 +36,26 @@ pipeline:
...
   - uses: strip
 
-# Note, we don't need to fetch any upstream files, as there aren't any custom
-# entrypoints or scripts required for compatibility with the metrics-server
-# Bitnami helm chart.
 subpackages:
+  - name: ${{package.name}}-compat
+    pipeline:
+      - working-directory: ${{targets.subpkgdir}}
+        runs: |
```
kranurag7:
```suggestion
      - runs: |
```
Nit: working-directory can be omitted here.

tulilirockz (commented):
Tiny nit, would be nice doing this just to make sure we dont have garbled output

tulilirockz (approved):
Seems fine? Maybe I missed something? Would be good to have another review

PR #59252, opened 2025-07-15
k3s-1.32: add version-streamed package for rancher-2.11 compatibility

jamie-albert:
## Summary

This PR adds k3s-1.32 as a version-streamed package in the os repository to support rancher-2.11.

## Context

From discussion with Jon Johnson:

- rancher-2.11 (in os) requires k3s 1.32.x for compatibility
- rancher-2.10 (in enterprise) requires k3s 1.31.x for compatibility  
- The main k3s package has moved to 1.33.x, breaking rancher-2.11 builds
- k3s produces a subpackage (k3s-multicall) that rancher depends on

## Problem Being Solved

Since packages in os cannot depend on packages in enterprise-packages, we need k3s-1.32 in os to support rancher-2.11. This follows the pattern where version-streamed packages remain in os when public packages depend on them.

## Implementation

- Added k3s-1.32.yaml with version streaming configured to track v1.32 tags
- Based on the k3s-1.32 package from https://github.com/chainguard-dev/enterprise-packages/pull/27327
- Provides k3s-multicall-1.32 subpackage for rancher-2.11 to depend on

## Testing

The package includes comprehensive tests for:
- Basic functionality of all k3s components
- Multicall symlink verification
- Static build verification for k3s-static-1.32

## Notes

As Jon mentioned: "k3s-1.31 belongs in wolfi because rancher-2.11 depends on it, even though it's not the latest k3s version." The same logic applies to k3s-1.32.

### DIFF
--- a/k3s-1.32.yaml
+++ b/k3s-1.32.yaml
@@ -1,0 +1,307 @@
+package:
+  name: k3s-1.32
+  version: "1.32.6.1"
+  epoch: 0
+  description:
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - busybox
+      - conntrack-tools
+      - containerd-shim-runc-v2
+      - ip6tables # this pulls in iptables as well
+      - ipset # required for network policy controller
+      - kmod
+      - libseccomp
+      - merged-bin
+      - mount
+      - nftables
+      - runc
+      - umount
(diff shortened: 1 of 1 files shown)


jamie-albert:
vulnerabilities are expected, will need advisories but multicall dependency relevant to rancher has no vulnerabilities.

PR #59385, opened 2025-07-16
fix(perl-*): rebuild all perl packages with perl 5.42.0

vishal-chdhry:
## Explanation

Perl v5.42.0 just came out. Rebuilding all perl package with latest perl to fix version mismatch errors.

Example error: 
```shell
 Perl API version v5.40.0 of Net::SSLeay does not match v5.42.0 at /usr/lib/perl5/vendor_perl/Net/SSLeay.pm line 1024.
```


aborrero (approved):
LGTM.

PR #59396, opened 2025-07-16
cilium envoy 1.17/llvm 17 renames

javacruft:
Build dependency renames and drops for new single source LLVM 17 package


sil2100 (approved):
Was curious about the build flag fix that's added there, but that's okay.

PR #59400, opened 2025-07-16
rust-1.76: update LLVM 17 package references

javacruft:
Build dependency renames and drops for new single source LLVM 17 package

### DIFF
--- a/rust-1.76.yaml
+++ b/rust-1.76.yaml
@@ -1,7 +1,7 @@
 package:
   name: rust-1.76
   version: 1.76.0
-  epoch: 3
+  epoch: 4
   description: "Empowering everyone to build reliable and efficient software."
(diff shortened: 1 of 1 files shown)


rust-1.76.yaml
```diff
@@ -94,8 +92,8 @@ pipeline:
       unset CARGO_PROFILE_RELEASE_OPT_LEVEL
       unset CARGO_PROFILE_RELEASE_PANIC
       unset CARGO_PROFILE_RELEASE_CODEGEN_UNITS
-      export CFLAGS="$CFLAGS -O2 -Iusr/include/llvm17"
```
sil2100, line 97: Now that was a bug..!

sil2100 (approved):

PR #59449, opened 2025-07-16
add /data/flux for flux-operator

sschubertchainguard:
add data compat package for flux-operator to allow deployment of a FluxInstance without specifying `spec.distribution.artifact`

### DIFF
--- a/flux-operator.yaml
+++ b/flux-operator.yaml
@@ -3,6 +3,9 @@ package:
   version: "0.24.1"
   epoch: 2
   description: Flux Operator creates and manages Flux instances running in Kubernetes
+  dependencies:
+    runtime:
+      - tzdata
   copyright:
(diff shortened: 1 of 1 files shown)


flux-operator.yaml
```diff
@@ -19,6 +22,10 @@ pipeline:
       output: flux-operator
       ldflags: -X main.VERSION=${{package.version}}
 
+  - runs: |
+      mkdir -p "${{targets.contextdir}}"/data
+      cp -R config/data/* ${{targets.contextdir}}/data
```
tulilirockz, line 27: Shouldn't this be under a compat subpackage?

kranurag7: this is required by the main application itself so I think landing the config inside main package looks good here similar to arangodb and others.

jamonation (approved):

PR #59457, opened 2025-07-16
Rebuild other perl-dev rdeps for the perl 5.42.0 release

dannf:
Generated by:
`wolfictl bump $(grep -- '- perl-dev$' *.yaml | grep -v ^perl- | cut -d: -f1)`


dannf:
Sharded out into smaller commits

PR #59497, opened 2025-07-16
Move k3s-1.32 from enterprise-packages to os and rancher-2.11 fixes

jamie-albert:
## Summary

  This PR moves k3s-1.32 from enterprise-packages to os to fix dependency resolution issues for rancher-2.11.

##   Problem

  K3s version streaming was initially implemented by moving k3s-1.31 and k3s-1.32 to enterprise-packages. However,
  this created a critical dependency issue:
  - rancher-2.11 (in os) requires k3s-multicall-1.32
  - Packages in os cannot see dependencies from enterprise-packages
  - This causes rancher-2.11 builds to fail due to missing k3s-multicall-1.32 dependency

##   Background

  K3s is currently not version-streamed in os, causing it to build the latest version (1.33) which is incompatible
  with rancher packages that have strict minor version requirements:
  - rancher-2.11 requires k3s 1.32.x
  - rancher-2.10 requires k3s 1.31.x

  Without version streaming, this incompatibility causes builds to use stale k3s-multicall dependencies that
  accumulate CVEs over time.

##   Solution

  Move k3s-1.32 from enterprise-packages back to os to ensure rancher-2.11 can access its required dependency.

  Per discussion with @jonjohnson, version-streamed k3s packages should remain in os when they're dependencies of
  packages in os, even if they're not the latest version. This ensures transparency and keeps dependencies co-located
   with the packages that require them.

##   Changes

  - Move k3s-1.32.yaml from enterprise-packages to os
  - Update rancher-2.11.yaml to use the correct k3s-multicall-1.32 dependency instead of the unversioned
  k3s-multicall
  - Ensures k3s-multicall-1.32 subpackage is available for rancher-2.11
  - Maintains proper dependency resolution while allowing ongoing CVE remediation

##   Note on K3s 1.32 Containerd Issue

  Be aware that k3s versions 1.32.5 and later include containerd v2.0.5-k3s1.32 which has known issues with Docker
  environments due to aggressive overlayfs snapshotter detection. Currently using k3s 1.32.6 which works in QEMU but
  may have issues in Docker runners.

### DIFF
--- a/k3s-1.32.yaml
+++ b/k3s-1.32.yaml
@@ -1,0 +1,307 @@
+package:
+  name: k3s-1.32
+  version: "1.32.2.1"
+  epoch: 0
+  description:
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - busybox
+      - conntrack-tools
+      - containerd-shim-runc-v2
+      - ip6tables # this pulls in iptables as well
+      - ipset # required for network policy controller
+      - kmod
+      - libseccomp
+      - merged-bin
+      - mount
+      - nftables
+      - runc
+      - umount
+      # Do not include runtime dependencies already included in the multicall k3s
+      # - ctr
+      # - crictl
+      # - containerd
+      # - kubectl
+      - wolfi-baselayout
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - crane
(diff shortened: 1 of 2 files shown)


jamie-albert:
remaining cve will require an advisory, can not bump multicall version higher than this until upstream supports containerd v2.0.5

kbsteere (approved):

PR #59565, opened 2025-07-17
add helm-mapkubeapis plugin package

yashpimple:
add helm plugin packages for `helm-mapkubeapis`
Related: https://github.com/helm/helm-mapkubeapis

#### For new package PRs only

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches


helm-mapkubeapis.yaml
```diff
@@ -0,0 +1,96 @@
...
+    #         spec:
+    #           containers:
+    #           - name: nginx
+    #             image: nginx:1.25.2
+    #     EOF
+    #     helm install deprecated-release /tmp/deprecated-chart --namespace test-ns --create-namespace --debug
+    #     helm mapkubeapis deprecated-release --namespace test-ns 2>&1
+    #     echo "Successfully test passed"
+    - runs: |
+        helm plugin list | grep -i 'mapkubeapis'
```
mritunjaysharma394: Try to add ` > out.log 2>&1 ` after all the grep, this should unblock you probably

mritunjaysharma394:
```suggestion
        helm plugin list | grep -i 'mapkubeapis'  > out.log 2>&1 
```

helm-mapkubeapis.yaml
```diff
@@ -0,0 +1,98 @@
...
+      deps: |-
+        github.com/docker/docker@v26.0.0-rc3
+        golang.org/x/oauth2@v0.27.0
+        helm.sh/helm/v3@v3.17.4
+
+  - runs: |
+      cat Makefile
+      make build
+      mkdir -p "${{targets.contextdir}}/root/.local/share/helm/plugins/helm-mapkubeapis/bin"
+      install -Dm755 ./bin/mapkubeapis "${{targets.contextdir}}/root/.local/share/helm/plugins/helm-mapkubeapis/bin/mapkubeapis"
```
mritunjaysharma394: The hardcoding of root here causes failure and this should not be done anyways

helm-mapkubeapis.yaml
```diff
@@ -0,0 +1,104 @@
...
+        golang.org/x/oauth2@v0.27.0
+        helm.sh/helm/v3@v3.17.4
+
+  # Dropping helm plugins in /usr/libexec since they are binaries and
+  # configuration intended to be run by the helm command.
+  # See https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s07.html
+  # Installers of this package will have to set this environment var
+  # HELM_PLUGINS=/usr/libexec/helm-plugins
+  - runs: |
+      cat Makefile
```
mritunjaysharma394: I think cat was for debugging, do we need it here?

helm-mapkubeapis.yaml
```diff
@@ -0,0 +1,104 @@
...
+              app: test
+          replicas: 1
+          template:
+            metadata:
+              labels:
+                app: test
+            spec:
+              containers:
+              - name: nginx
+                image: nginx:1.25.2
```
mritunjaysharma394: avoid hardcoding versions. use latest tag if possible?

helm-mapkubeapis.yaml
```diff
@@ -0,0 +1,104 @@
...
+    - uses: test/kwok/cluster
+    - name: "Tests for helm-mapkubeapis plugin"
+      runs: |
+        set -x
+        mkdir -p /tmp/deprecated-chart/templates
+        export KUBECONFIG=$HOME/.kwok/clusters/kwok/kubeconfig.yaml
+        cat > /tmp/deprecated-chart/Chart.yaml <<EOF
+        apiVersion: v2
+        name: deprecated-chart
+        version: 0.1.0
```
mritunjaysharma394: Where are we getting this version from?

yashpimple: the version is just a placeholder for test chart we dynamically create during the test run. It's not sourced from anywhere just a simple valid version field required by helm chart spec for it to install correctly.

helm-mapkubeapis.yaml
yashpimple: i will remove it

helm-mapkubeapis.yaml
```diff
@@ -0,0 +1,104 @@
...
+        helm.sh/helm/v3@v3.17.4
+
+  # Dropping helm plugins in /usr/libexec since they are binaries and
+  # configuration intended to be run by the helm command.
+  # See https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s07.html
+  # Installers of this package will have to set this environment var
+  # HELM_PLUGINS=/usr/libexec/helm-plugins
+  - runs: |
+      cat Makefile
+      make build
```
mritunjaysharma394: can we use go/build pipeline rather? Also, we can configure INSTALL_DIR there too by either changing our `output:` or using symlink and you can add another pipeline to copy the config and yamls

yashpimple: @mritunjaysharma394 i tried implementing this approach, but keep on see some issue with `go/build` Pipeline
The `go/build` step completes without errors but doesn't create the expected binary:
```
uses: go/build
  with:
    packages: ./cmd/mapkubeapis
    output: usr/libexec/helm-plugins/helm-mapkubeapis/bin/mapkubeapis
    ldflags: -X main.version=v${{package.version}}
```
which results in below error 
`install: can't stat 'usr/libexec/helm-plugins/helm-mapkubeapis/bin/mapkubeapis': No such file or directory`

the pipeline completes without errors but doesn't create the binary file which results the test and other pipeline to fail  due to which i am switch back to the `make build` approach

Aditevil: are you sure about the binary not being created? It could be in `/usr/bin` by default

yashpimple: Thanks @Aditvil-Dev i totally overlooked that part! 🙈 i've now updated the pipeline to correctly symlink it to the expected plugin path under `usr/libexec/helm-plugins/...` and also added a step to install the plugin metadata files (plugin.yaml, Map.yaml). everything seems to be working smoothly now

helm-mapkubeapis.yaml
```diff
@@ -0,0 +1,99 @@
...
+  name: helm-mapkubeapis
+  version: 0.6.1
+  epoch: 0
+  description: Helm plugin to map and update deprecated Kubernetes APIs in Helm release manifests
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  environment:
+    CGO_ENABLED: "0"
```
Aditevil: isn't this set by default?(Now that we use go/build)

helm-mapkubeapis.yaml
```diff
@@ -0,0 +1,99 @@
...
+              labels:
+                app: test
+            spec:
+              containers:
+              - name: nginx
+                image: nginx:latest
+        EOF
+        helm install deprecated-release /tmp/deprecated-chart --namespace test-ns --create-namespace --debug --kubeconfig $HOME/.kwok/clusters/kwok/kubeconfig.yaml
+        helm mapkubeapis deprecated-release --namespace test-ns 2>&1
+        echo "Successfully test passed"
```
Aditevil: Can we use `uses: test/daemon-check-output` and check for output. From upstream README example I see useful logs that can be validated.

https://github.com/helm/helm-mapkubeapis#:~:text=of%20the%20release-,Example%20output%3A,-%24%20helm%20mapkubeapis

Last nit & bit and it should be good to go.

yashpimple: updated the PR to make use of `test/daemon-check-output`

Aditevil (approved):
nice work.

PR #59750, opened 2025-07-17
refactor cluster-api packages

Aditevil:
- Changing binary name to `kubeadm-control-plane-controller`, to match [upstream image name](https://github.com/kubernetes-sigs/cluster-api/blob/7b5d4b0ea46328a9af3ae8ef2bf72d28ec29d87a/docs/release/role-handbooks/release-lead/README.md?plain=1#L222).

- Entrypoint for controller image is `/manager`.


jamonation (dismissed):

cluster-api-1.10.yaml
```diff
@@ -64,44 +61,43 @@ test:
...
-        - ${{package.name}}-clusterctl-compat
+        - git
   pipeline:
-    - uses: git-checkout
-      with:
-        expected-commit: 737edc88faed28b0bd81f6768ebe2ea276fdbe3a
-        repository: https://github.com/kubernetes-sigs/cluster-api
-        tag: v${{package.version}}
+    - runs: |
+        git clone --branch v${{package.version}} --depth 1 https://github.com/kubernetes-sigs/cluster-api.git
```
EyeCantCU, line 69: Why did `git-checkout` have to be dropped?

cluster-api-1.10.yaml
```diff
@@ -64,44 +61,43 @@ test:
...
         expected_output: |
           Starting workers
           Starting webhook server
           Starting EventSource
           All workers finished
+        timeout: 30
     - uses: test/daemon-check-output
       with:
-        start: kubeadm-controlplane-controller
+        start: /usr/bin/kubeadm-control-plane-controller
```
EyeCantCU, line 83: Why are we using the absolute path instead of implicit now?

EyeCantCU (changes requested):

cluster-api-1.10.yaml
Aditevil: to avoid having to manually update each time there is a new version.

cluster-api-1.10.yaml
Aditevil: Just for nit(ness), since I removed all `-compat` packages from tests.

EyeCantCU (approved):

PR #59917, opened 2025-07-18
update license from Apache-2.0 to LGPL-2.0-or-later in jruby-9.4.yaml

utieyin:
jruby - existing package

Adding correct license for the application as the previous value of Apache-2.0 is incorrect according to https://github.com/jruby/jruby/blob/master/COPYING


mritunjaysharma394 (commented):
I think we should bump the epoch

mritunjaysharma394 (approved):

PR #59925, opened 2025-07-18
feat: new package rancher-system-upgrade-controller

rahulduvedi:
new package: `rancher-system-upgrade-controller:0.16.0`

### DIFF
--- a/rancher-system-upgrade-controller.yaml
+++ b/rancher-system-upgrade-controller.yaml
@@ -1,0 +1,120 @@
+package:
+  name: rancher-system-upgrade-controller
+  version: "0.16.0"
+  epoch: 0
+  description: A Kubernetes controller that safely automates node upgrades in your cluster, coordinating OS and Kubernetes version updates with minimal disruption.
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - tzdata
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/rancher/system-upgrade-controller
+      tag: v${{package.version}}
+      expected-commit: 185ff9f785cd68d4fd09b0856b12b1ee4196cce9
+
+  - uses: go/bump
+    with:
+      deps: |-
+        golang.org/x/net@v0.38.0
+        golang.org/x/oauth2@v0.27.0
+
+  - uses: go/build
+    with:
+      packages: .
+      output: system-upgrade-controller
+      ldflags: |
+        -X github.com/rancher/system-upgrade-controller/pkg/version.Version=${{package.version}}
+        -X github.com/rancher/system-upgrade-controller/pkg/version.Revision=$(git rev-parse HEAD)
+        -X github.com/rancher/system-upgrade-controller/pkg/version.Branch=HEAD
+        -X github.com/rancher/system-upgrade-controller/pkg/version.BuildUser=reproducible@reproducible
+        -X github.com/rancher/system-upgrade-controller/pkg/version.BuildDate=$(date -u -d "@$SOURCE_DATE_EPOCH" +%Y-%m-%d-%H:%M)
+
+update:
+  enabled: true
+  ignore-regex-patterns:
+    - -rc
+  github:
+    identifier: rancher/system-upgrade-controller
+    strip-prefix: v
+
+test:
+  pipeline:
+    - name: "Basic binary verification"
+      runs: |
+        system-upgrade-controller --version | grep -q "${{package.version}}"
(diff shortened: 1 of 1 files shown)


rancher-system-upgrade-controller.yaml
```diff
@@ -0,0 +1,120 @@
...
+                node-role.kubernetes.io/agent: ""
+            serviceAccountName: system-upgrade
+            concurrency: 1
+            version: "test-version"
+            upgrade:
+              image: alpine:latest
+              command: ["sh", "-c"]
+              args: ["echo test"]
+          EOF
+          kubectl get plan test-plan -n system-upgrade -o jsonpath='{.metadata.name}' | grep -q "test-plan"
```
Dentrax: No need `grep`, as `kubectl` would return error if its no exist anyway.
```suggestion
          kubectl get plan test-plan -n system-upgrade -o jsonpath='{.metadata.name}'
```

rancher-system-upgrade-controller.yaml
```diff
@@ -0,0 +1,120 @@
...
+
+          # Wait for the plan to be processed and check node labels to verify if it ran successfully
+          sleep 5
+          kubectl get nodes -o yaml | grep -q "plan.upgrade.cattle.io/test-plan"
```
Dentrax: Let's omit `grep` here as well, and use label selectors, `-l, --selector=''`.

rancher-system-upgrade-controller.yaml
```diff
@@ -0,0 +1,120 @@
...
+            namespace: system-upgrade
+          spec:
+            nodeSelector:
+              matchLabels:
+                node-role.kubernetes.io/agent: ""
+            serviceAccountName: system-upgrade
+            concurrency: 1
+            version: "test-version"
+            upgrade:
+              image: alpine:latest
```
Dentrax:
```suggestion
              image: cgr.dev/chainguard/wolfi-base:latest
```

rancher-system-upgrade-controller.yaml
```diff
@@ -0,0 +1,120 @@
+package:
+  name: rancher-system-upgrade-controller
+  version: "0.16.0"
+  epoch: 0
+  description: A Kubernetes controller that safely automates node upgrades in your cluster, coordinating OS and Kubernetes version updates with minimal disruption.
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - tzdata
```
Dentrax, line 10: Do we need additional runtime deps here? See: https://github.com/rancher/system-upgrade-controller/blob/185ff9f785cd68d4fd09b0856b12b1ee4196cce9/Dockerfile.dapper#L12-L21

rahulduvedi: I'm not sure but I think this is actual production Dockerfile https://github.com/rancher/system-upgrade-controller/blob/master/package/Dockerfile
The above Dockerfile is for development, not production runtime.

rancher-system-upgrade-controller.yaml
rahulduvedi: Changed the image to cgr.dev/chainguard/wolfi-base

rancher-system-upgrade-controller.yaml
rahulduvedi: Understood, Omitted grep here.

rancher-system-upgrade-controller.yaml
rahulduvedi: Understood, made the requested change.

rancher-system-upgrade-controller.yaml
Dentrax: Thanks for checking!

rancher-system-upgrade-controller.yaml
```diff
@@ -0,0 +1,120 @@
...
+            upgrade:
+              image: cgr.dev/chainguard/wolfi-base:latest
+              command: ["sh", "-c"]
+              args: ["echo test"]
+          EOF
+          kubectl get plan test-plan -n system-upgrade -o jsonpath='{.metadata.name}'
+
+          # Wait for the plan to be processed and check node labels to verify if it ran successfully
+          sleep 5
+          kubectl get nodes -l plan.upgrade.cattle.io/test-plan
```
Dentrax, line 120: IIRC selector takes key=value.

Dentrax (approved):

rancher-system-upgrade-controller.yaml
Dentrax: ignore me, it seems this also works.

PR #59926, opened 2025-07-18
llvm-18: fix a dangling symlink for libclang-cpp.so.18

javacruft:
This is specifically causing a problem in bpftrace.


mritunjaysharma394 (approved):

dannf (approved):
could you just run ldconfig on it?

PR #59972, opened 2025-07-18
xorgproto: enable legacy headers (a.k.a xf86miscproto)

Dentrax:
`xf86miscproto` is deprecated and replaced with `xorgproto`: https://gitlab.freedesktop.org/xorg/proto/xf86miscproto/-/blob/master/autogen.sh?ref_type=heads

To build this: https://github.com/wolfi-dev/os/pull/59993 we do need some headers. It builds if we do enable xorgproto with legacy support. libx11-dev doesn't provide `xf86mscstr.h`.

But some headers like `XKBgeom.h XKBsrv.h XKBstr.h` is still provided by `libx11`, which resulting conflict during install:

```
installing apk packages: installing packages: installing libx11-dev (ver:1.8.12-r1 arch:aarch64): unable to install files for pkg libx11-dev: writing header for "usr/include/X11/extensions/XKBgeom.h": packages map[libx11-dev:libx11 xorgproto:xorgproto] has conflicting file: "usr/include/X11/extensions/XKBgeom.h"
```

So remove those from libx11 and re-build xorgproto with having those headers. So we can build `libXxf86misc`.

### DIFF
--- a/libx11.yaml
+++ b/libx11.yaml
@@ -1,7 +1,7 @@
 package:
   name: libx11
   version: "1.8.12"
-  epoch: 1
+  epoch: 2
   description: X11 client-side library
   copyright:
     - license: XFree86-1.1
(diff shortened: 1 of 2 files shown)


Dentrax:
~Superseded by https://github.com/wolfi-dev/os/pull/59993~ (I was wrong, we still need this)

Dentrax:
Superseded by:
* https://github.com/wolfi-dev/os/pull/60251
* https://github.com/wolfi-dev/os/pull/60252

PR #60020, opened 2025-07-18
crossplane-provider-aws: attempt to build without closed source component

tulilirockz:


tulilirockz:
Other crossplane-provider-aws PRs do this already and got merged I believe

PR #60114, opened 2025-07-19
jaeger: No-change rebuild to fix file permissions

javacruft:


aborrero:
now it builds fine locally, should be ready for review and merge.

AbduAzim424:
FYI, If we want to keep type checking, using `npm` only instead of `yarn` is the solution. P.S upstream confirmed that they do not use `yarn` (https://github.com/jaegertracing/jaeger-ui/issues/2982)

aborrero:
> FYI, If we want to keep type checking, using `npm` only instead of `yarn` is the solution. P.S upstream confirmed that they do not use `yarn` ([jaegertracing/jaeger-ui#2982](https://github.com/jaegertracing/jaeger-ui/issues/2982))

thanks! updated.

egibs (approved):

PR #60213, opened 2025-07-21
JuiceFS CSI Driver Package addition

AnmolVirdi:
**juicefs-csi-driver package** `Current version: v0.29.1`
- JuiceFS CSI Driver implements the CSI specification, allowing JuiceFS to be integrated with Kubernetes.

**Type**: Go binary
**REF**: https://github.com/juicedata/juicefs-csi-driver

**Tests:**
- Basic version and completion script tests.

**Note:** 
- Functional tests for JuiceFS CSI Interface require complete Pod-Level Setup.
- `gluster`, `foundationdb` and `ceph` are not dependencies for `juicefs-csi-driver` but for `juicefs` binary which is packaged separately. They're plugins mostly meant for enterprise use, therefore excluded.


juicefs-csi-driver.yaml
```diff
@@ -0,0 +1,74 @@
...
+      runtime:
+        - juicefs # Needed for JuiceFS CSI Driver functions, attached symlinks too.
+        - tini # Need in upstream entrypoint, acts as the init process.
+        - lsof # Check which process is using a file, mount point, or socket.
+        - procps # Check running JuiceFS processes
+        - strace # Analyze what a JuiceFS mount process is doing at the syscall level
+        - iproute2 # Check interface and routing info if JuiceFS
+        - net-tools # Check open ports or connections
+        - tcpdump # Used to trace network packets
+        - coreutils # Needed for working with volumes
```
kranurag7: let's move this to top-level package similar to other csi drivers. do we need strace only in ours or it's in theirs only. If it's not a necessary requirement for image and only for debugging then we can ship inside `-dev` same for tcpdump.

juicefs-csi-driver.yaml
```diff
@@ -0,0 +1,74 @@
...
+update:
+  enabled: true
+  ignore-regex-patterns:
+    - '-rc'
+    - '-chart'
+  github:
+    identifier: juicedata/juicefs-csi-driver
+    strip-prefix: v
+
+# CSI Interface Testing Requires Pod-Level Setup
```
kranurag7, line 65: please check `secrets-store-csi-driver-provider-azure.yaml` for simple testing of csi driver and `cert-manager-csi-driver.yaml` for a complex setup of same at package level. 

would be great if we can include the same at package level as well.

AnmolVirdi: I've gone through all other csi-driver examples, `juicefs-csi-driver` does not return any logs (even with proper setup) when testing in a non-k8s runtime. There's a strict requirement to test it at image level.
```
bash-5.3# juicefs-csi-driver
bash-5.3# juicefs-csi-driver --help
juicefs csi driver

Usage:
  juicefs-csi [flags]
  juicefs-csi [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  help        Help about any command
  upgrade     upgrade mount pod smoothly

Flags:
      --add_dir_header                            If true, adds the file directory to the header of the log messages
      --alsologtostderr                           log to standard error as well as files (no effect when -logtostderr=true)
      --by-process                                CSI Driver run juicefs in process or not. default false.
      --cache-client-conf                         Cache client config file. default false.
      --config string                             Paths to a csi config file. default empty
      --enable-manager                            Enable pod manager in csi node. default false.
      --endpoint string                           CSI endpoint (default "unix://tmp/csi.sock")
      --format-in-pod                             Put format/auth in pod
  -h, --help                                      help for juicefs-csi
      --kubeconfig string                         Paths to a kubeconfig. Only required if out-of-cluster.
      --leader-election                           Enables leader election. If leader election is enabled, additional RBAC rules are required. 
      --leader-election-lease-duration duration   Duration, in seconds, that non-leader candidates will wait to force acquire leadership. Defaults to 15 seconds. (default 15s)
      --leader-election-namespace string          Namespace where the leader election resource lives. Defaults to the pod namespace if not set.
      --log_backtrace_at traceLocation            when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                            If non-empty, write log files in this directory (no effect when -logtostderr=true)
      --log_file string                           If non-empty, use this log file (no effect when -logtostderr=true)
      --log_file_max_size uint                    Defines the maximum size a log file can grow to (no effect when -logtostderr=true). Unit is megabytes. If the value is 0, the maximum file size is unlimited. (default 1800)
      --logtostderr                               log to standard error instead of files (default true)
      --nodeid string                             Node ID
      --one_output                                If true, only write logs to their native severity level (vs also writing to each lower severity level; no effect when -logtostderr=true)
      --provisioner                               Enable provisioner in controller. default false.
      --reconciler-interval int                   interval (default 5s) for reconciler (default 5)
      --skip_headers                              If true, avoid header prefixes in the log messages
      --skip_log_headers                          If true, avoid headers when opening log files (no effect when -logtostderr=true)
      --stderrthreshold severity                  logs at or above this threshold go to stderr when writing to files and stderr (no effect when -logtostderr=true or -alsologtostderr=true) (default 2)
  -v, --v Level                                   number for the log level verbosity
      --validating-webhook                        Enable validation webhook in controller. default false.
      --version                                   Print the version and exit.
      --vmodule moduleSpec                        comma-separated list of pattern=N settings for file-filtered logging
      --webhook                                   Enable mutating webhook in controller for sidecar mode. default false.
      --webhook-cert-dir string                   Admission webhook cert/key dir. (default "/etc/webhook/certs")
      --webhook-port int                          Admission webhook port. (default 9444)

Use "juicefs-csi [command] --help" for more information about a command.

juicefs-csi-driver.yaml
AnmolVirdi: Updated runtime dependencies with only essential deps. We can keep the rest in `-dev`.

kranurag7 (changes requested):
notes:
- have we tested the following https://juicefs.com/docs/csi/troubleshooting/#csi-dashboard 
- why are not including gcluster, foundationdb and ceph (some of it are not there in index but can we include the ones that are there) if there's some context about not having those then please include it in the description. 
- juicefs cli needs to be packages as a subpackage as well and then `ln -s /usr/local/bin/juicefs /usr/bin/mount.juicefs` ref: https://github.com/juicedata/juicefs-csi-driver/blob/master/docker/csi.Dockerfile

AnmolVirdi:
> notes:
> 
> * have we tested the following https://juicefs.com/docs/csi/troubleshooting/#csi-dashboard
> * why are not including gcluster, foundationdb and ceph (some of it are not there in index but can we include the ones that are there) if there's some context about not having those then please include it in the description.
> * juicefs cli needs to be packages as a subpackage as well and then `ln -s /usr/local/bin/juicefs /usr/bin/mount.juicefs` ref: https://github.com/juicedata/juicefs-csi-driver/blob/master/docker/csi.Dockerfile

1. Dashboard is associated with separate image: https://hub.docker.com/r/juicedata/csi-dashboard/tags

2. `gluster`, `foundationdb` and `ceph` are not dependencies for `juicefs-csi-driver` but for `juicefs` binary which is packaged separately. They're plugins mostly meant for enterprise use.

3. `juicefs cli` in this context works same as our `juicefs` mount binary but has some optional downloading instructions (if juicefs enterprise binary is missing), the intent is to only use its CLI functions. We're not including the enterprise binaries. [`juicefs.py`:`/usr/bin/juicefs`, `/usr/local/juicefs/mount/jfsmount`] (Both are enterprise contents)


`juicefs-csi-driver`
Enterprise components: `/usr/bin/juicefs` (juicefs.py) and `/usr/local/juicefs/mount/jfsmount` (Enterprise JuiceFS)
Community components: `/usr/local/bin/juicefs` (CE juicefs) and `/bin/mount.juicefs`(Symlink)

`csi-driver` will only work with community `juicefs` if we exclude enterprise `jfsmount` binary in the bundle.

cc @kranurag7

kranurag7 (approved):

PR #59993, opened 2025-07-18
add libXxf86misc

Dentrax:
~BLOCKED by: https://github.com/wolfi-dev/os/pull/59972~

```
configure: error: Package requirements (xproto x11 xextproto xext xf86miscproto) were not met:

Package 'xf86miscproto' not found
```

Fixes:

Related:

#### For new package PRs only

- [ ] This PR is marked as fixing a pre-existing package request bug
  - [ ] Alternatively, the PR is marked as related to a pre-existing package request bug, such as a dependency
- [X] REQUIRED - The package is available under an OSI-approved or FSF-approved license
- [X] REQUIRED - The version of the package is still receiving security updates
- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`)

#### For new version streams

- [ ] The upstream project actually supports multiple concurrent versions.
- [ ] Any subpackages include the version string in their package name (e.g. `name: ${{package.name}}-compat`)
- [ ] The package (and subpackages) `provides:` logical unversioned forms of the package (e.g. `nodejs`, `nodejs-lts`)
- [ ] If non-streamed package names no longer built, open PR to withdraw them (see [WITHDRAWING PACKAGES](https://github.com/wolfi-dev/os/blob/main/WITHDRAWING_PACKAGES.md))

#### For package updates (renames) in the base images

When updating packages part of base images (i.e. cgr.dev/chainguard/wolfi-base or ghcr.io/wolfi-dev/sdk)
- [ ] REQUIRED cgr.dev/chainguard/wolfi-base and ghcr.io/wolfi-dev/sdk images successfully build
- [ ] REQUIRED cgr.dev/chainguard/wolfi-base and ghcr.io/wolfi-dev/sdk contain no obsolete (no longer built) packages
- [ ] Upon launch, does `apk upgrade --latest` successfully upgrades packages or performs no actions

#### For security-related PRs

- [ ] The security fix is recorded in the [advisories](https://github.com/wolfi-dev/advisories) repo

#### For version bump PRs

- [ ] The `epoch` field is reset to 0

#### For PRs that add patches

- [ ] Patch source is documented

PR #60252, opened 2025-07-21
xorgproto: enable legacy headers

Dentrax:
Part 1/2: https://github.com/wolfi-dev/os/pull/60251

Part 2/2:

`xf86miscproto` is deprecated and replaced with `xorgproto`: https://gitlab.freedesktop.org/xorg/proto/xf86miscproto/-/blob/master/autogen.sh?ref_type=heads

To build this: https://github.com/wolfi-dev/os/pull/59993 we do need some headers. It builds if we do enable xorgproto with legacy support. libx11-dev doesn't provide `xf86mscstr.h`.

But some headers like `XKBgeom.h XKBsrv.h XKBstr.h` is still provided by `libx11`, which resulting conflict during install:

```
installing apk packages: installing packages: installing libx11-dev (ver:1.8.12-r1 arch:aarch64): unable to install files for pkg libx11-dev: writing header for "usr/include/X11/extensions/XKBgeom.h": packages map[libx11-dev:libx11 xorgproto:xorgproto] has conflicting file: "usr/include/X11/extensions/XKBgeom.h"
```

So remove those from libx11 and re-build xorgproto with having those headers. So we can build `libXxf86misc`.

### DIFF
--- a/xorgproto.yaml
+++ b/xorgproto.yaml
@@ -1,7 +1,7 @@
 package:
   name: xorgproto
   version: "2024.1"
-  epoch: 2
+  epoch: 3
   description: Combined X.Org X11 protocol headers
   copyright:
     - license: BSD-2-Clause AND MIT AND X11
@@ -24,6 +24,10 @@ pipeline:
       uri: https://xorg.freedesktop.org/archive/individual/proto/xorgproto-${{package.version}}.tar.gz
 
   - uses: meson/configure
+    with:
+      # Building the xf86miscproto requires legacy support
+      opts: |
+        -Dlegacy=true
 
   - uses: meson/compile
 


xorgproto.yaml
```diff
@@ -1,10 +1,13 @@
...
   name: xorgproto
   version: "2024.1"
-  epoch: 2
+  epoch: 3
   description: Combined X.Org X11 protocol headers
   copyright:
     - license: BSD-2-Clause AND MIT AND X11
+  dependencies:
+    provides:
+      - xf86miscproto=${{package.full-version}} # With legacy support enabled, it can be drop-in replacement for xf86miscproto
```
dannf: I don't see `xf86miscproto` anywhere - if we don't need to transition things, I'd prefer to not add a provides.

Dentrax: That makes sense, removed!

dannf (approved):

PR #60254, opened 2025-07-21
cluster-api: move clusterctl to it's own subpackage

kranurag7:
we wanted to have a provides of clusterctl so migrate this to it's
own subpackage and add another clusterctl provides in there.


kranurag7:
closing given it's not required anymore. we can revisit in future if required.

PR #60264, opened 2025-07-21
feat: llvm-20: enable mlir

techalchemy:
mlir is needed for some AIML builds; enable it for our llvm builds.

Start with llvm-20. Once merged there, we can backport as far back as needed.

### DIFF
--- a/llvm-20.yaml
+++ b/llvm-20.yaml
@@ -1,7 +1,7 @@
 package:
   name: llvm-20
   version: "20.1.8"
-  epoch: 0
+  epoch: 1
   description: Low-level virtual machine ${{vars.major-version}} - core frameworks
   copyright:
     - license: Apache-2.0
@@ -74,7 +74,7 @@ pipeline:
         -DLLVM_ENABLE_FFI=ON \
         -DLLVM_ENABLE_LIBCXX=ON \
         -DLLVM_ENABLE_PIC=ON \
-        -DLLVM_ENABLE_PROJECTS="clang;lld;clang-tools-extra" \
+        -DLLVM_ENABLE_PROJECTS="clang;lld;mlir;clang-tools-extra" \
         -DLLVM_ENABLE_RUNTIMES="compiler-rt;libcxx;libcxxabi;libunwind" \
         -DLLVM_ENABLE_RTTI=ON \
         -DLLVM_ENABLE_SPHINX=OFF \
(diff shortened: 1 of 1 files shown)


xnox:
All but 19 worked; maybe drop that one for now, and land enabling this for all others first?

techalchemy:
dropped 19 and 20 now due to failures, the rest should work ok. may need to loop back around to the failures in the future

dannf:
I am a bit nervous about adding support in older versions that we can't enable in the latest. Maybe we should get this working in 20 first, then backport?

I wonder if we want to split out any subpackages here. Debian uses `libmlir-VER-dev`, `libmlir-VER` and `mlir-VER-tools`. @javacruft thoughts?

dannf:
I think `llvm-20` may have failed test due to a race with an upstream version bump. Pushed that change again just to confirm...

javacruft:
> I am a bit nervous about adding support in older versions that we can't enable in the latest. Maybe we should get this working in 20 first, then backport?
> 
> I wonder if we want to split out any subpackages here. Debian uses `libmlir-VER-dev`, `libmlir-VER` and `mlir-VER-tools`. @javacruft thoughts?

:+1: on this suggestion - I was just reviewing 14 and came to the same conclusion myself.

javacruft (changes requested):
Inline with @dannf comment please can we split the resulting shared libraries, static objects headers and tools into separate mlir packages:

libmlir-${{vars.major-version}}
libmlir-${{vars.major-version}}-dev
mlir-${{vars.major-version}}-tools

thanks!

dannf:
The failures w/ llvm-20 are actually `ldd-check` failures like:
```
/usr/bin/mlir-runner: missing libMLIR.so.20.1
```
My guess is we're not actually copying that library into an apk at the moment.

sergiodj (dismissed):

llvm-20.yaml
```diff
@@ -591,6 +591,74 @@ subpackages:
           mv ${{targets.destdir}}/${{vars.llvm-prefix}}/include/*unwind* ${{targets.contextdir}}/usr/include/
           mv ${{targets.destdir}}/${{vars.llvm-prefix}}/include/mach-o/*unwind* ${{targets.contextdir}}/usr/include/mach-o/
 
+  - name: libmlir-${{vars.major-version}}
+    description: LLVM libmlir
+    dependencies:
+      runtime:
+        - libmlir-${{vars.major-version}}=${{package.full-version}}
```
javacruft, line 598: Should this be a provides?

javacruft:
```suggestion
      provides:
        - libmlir=${{package.full-version}}
```

dannf: oh, nice catch!

javacruft (approved):

PR #60278, opened 2025-07-21
chore: Streamline auth take two

EyeCantCU:
* Automate token generation
* Use Octo STS policy for guarded repos

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,27 +124,11 @@ $(repo_token_targets): repo-token/%: %/.guarded-repo.token
 libraries_token_targets = $(foreach name,$(pkgs),lib-token/$(name))
 $(libraries_token_targets): lib-token/%: %/.libraries.token
 
-.PHONY: is-google-service-account
-is-google-service-account:
-	@if ! curl -f -H "Metadata-Flavor: Google" \
-			  http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/identity?audience=issuer.enforce.dev \
-			  -o /dev/null; then \
-		echo "Google SA token not found."; \
-		exit 1; \
-	fi
-	@echo "Google SA token found."
-
-.PHONY: _tokens-if-needed
-_tokens-if-needed:
-	$(call repo_token_if_needed,$(PKGNAME))
-	$(call libraries_token_if_needed,$(PKGNAME))
-
 .PHONY: tokens-if-needed/%
 tokens-if-needed/%:
 	$(eval pkgname := $*)
-	@if ! $(MAKE) is-google-service-account 2>/dev/null; then \
-		$(MAKE) _tokens-if-needed PKGNAME=$(pkgname); \
-	fi
+	$(call repo_token_if_needed,$(pkgname))
+	$(call libraries_token_if_needed,$(pkgname))
 
 .PHONY: clean-tokens/%
 clean-tokens/%:
--- a/Makefile
+++ b/Makefile
@@ -352,10 +352,9 @@ mv "$(2).tmp" "$(2)"
 endef
 
 define repo_token_if_needed
-$(eval yamlfile := $1.yaml)
 $(eval pkgname := $1)
 
-@if grep -qE 'uses:.*(auth/guarded-repo|iamguarded/)' $(yamlfile); then \
+@if [ -n "$$($(MAKE) -s compile/$(pkgname) | jq -r '.. | .uses? | strings | select(test("auth/guarded-repo"))')" ]; then \
 	echo "Creating guarded repo token for $(pkgname)…"; \
 	$(MAKE) repo-token/$(pkgname); \
 else \
@@ -364,13 +363,9 @@ fi
 endef
 
 define libraries_token_if_needed
-$(eval yamlfile := $1.yaml)
 $(eval pkgname := $1)
 
-$(eval num_auth_pipelines := $(shell grep -cE 'uses:.*auth/' $(yamlfile)))
-$(eval num_ignored_auth_pipelines := $(shell grep -cE 'uses:.*(auth/guarded-repo|iamguarded/)' $(yamlfile)))
-
-@if [ $(num_auth_pipelines) -gt $(num_ignored_auth_pipelines) ]; then \
+@if [ -n "$$($(MAKE) -s compile/$(pkgname) | jq -r '.. | .uses? | strings | select(test("auth/(?!guarded-repo)"))')" ]; then \
 	echo "Creating guarded libraries token for $(pkgname)…"; \
 	$(MAKE) lib-token/$(pkgname); \
 else \
--- a/Makefile
+++ b/Makefile
@@ -84,6 +84,10 @@ else
 	MELANGE_OPTS += -r ${WOLFI_REPO}
 endif
 
+# Evaluate all package manifests and packages
+yamls := $(wildcard *.yaml)
+pkgs := $(subst .yaml,,$(yamls))
+
 ${KEY}:
 	${MELANGE} keygen ${KEY}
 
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -280,9 +332,38 @@ check-bootstrap:
...
+echo "auth-download[$(1)] to $(2) from $(3)" && \
+curl --fail -LS --silent -o $(2).tmp --user "user:$$tok" $(3) && \
+mv "$(2).tmp" "$(2)"
+endef
+
+define repo_token_if_needed
+$(eval yamlfile := $1.yaml)
+$(eval pkgname := $1)
+
+@if grep -qE 'uses:.*(auth/guarded-repo|iamguarded/)' $(yamlfile); then \
```
stevebeattie: Probably we should avoid generating tokens for commented out `uses:` statements, with something like the following:

```suggestion
@if grep -qE '^[^#]*uses:.*(auth/guarded-repo|iamguarded/)' $(yamlfile); then \
```

Makefile
```diff
@@ -280,9 +332,38 @@ check-bootstrap:
...
+	echo "$(pkgname) does not use a guarded repo. Skipping token creation."; \
+fi
+endef
+
+define libraries_token_if_needed
+$(eval yamlfile := $1.yaml)
+$(eval pkgname := $1)
+
+$(eval num_auth_pipelines := $(shell grep -cE 'uses:.*auth/' $(yamlfile)))
+$(eval num_ignored_auth_pipelines := $(shell grep -cE 'uses:.*(auth/guarded-repo|iamguarded/)' $(yamlfile)))
```
stevebeattie: Same change here to match:

```suggestion
$(eval num_auth_pipelines := $(shell grep -cE '^[^#]*uses:.*auth/' $(yamlfile)))
$(eval num_ignored_auth_pipelines := $(shell grep -cE '^[^#]*uses:.*(auth/guarded-repo|iamguarded/)' $(yamlfile)))
```

Makefile
```diff
@@ -100,14 +104,52 @@ clean:
...
-	mv $${tmpf} ${CACHEDIR}/.libraries_token.txt
-
-.PHONY: lib-token
-lib-token: ${CACHEDIR}/.libraries_token.txt
+%/.guarded-repo.token:
+	$(eval pkgname := $*)
+	tmpf=$(shell mktemp) && \
+	chainctl auth octo-sts --identity=guarded-package-repos --scope=chainguard-dev > $${tmpf} && \
+	mkdir -p $(pkgname) && \
+	mv $${tmpf} ./$(pkgname)/.guarded-repo.token
```
stevebeattie, line 112: If the chainctl call or (less likely) mkdir fails, then the temporary file created will be left around. We should probably trap and cleanup if this happens here and in the package/.libraries.token generator targets

stevebeattie (commented):
This is looking pretty nice, a couple of nits.

Makefile
EyeCantCU: Moved to using melange compile, no more grep-fu :)

Makefile
EyeCantCU: No longer an issue :)

Makefile
EyeCantCU: I think this is good as is, but feel free to disagree and I can revisit. Every recipe that calls this has a trap that performs cleanup of the file after this target is called, regardless of whether or not chainctl succeeds. There isn't a great way to handle this in the target itself because it has been called by another consumer. If chainctl fails, that failure will be printed to stderr and will not be written to the file. If mkdir fails... we probably have worse things to worry about. We are still left with a file that contains whatever was written to stdout, but that's only possible if manually invoked, which I hope is a very very rare edge case after this

stevebeattie:
Nice, I really like the change to using melange compile to get json output to parse. However, it wasn't working for me locally, with my gnu make on ubuntu 24.04, `make tokens-if-needed/PKG` would succeed, but attempting to build the package, jq would fail parsing the output like so:

```
jq: parse error: Invalid numeric literal at line 1, column 5
```

I ended up putting a `tee` command inline to see what `make compile` was generating, and found that despite the `-s` argument to make, it was emitting the directory information, like so:

```
make[3]: Entering directory '/home/USER/git/wolfi-dev/os'
```

Using the following patch to add `--no-print-directory` makes it work for me, and also reduces the number of MELANGE_RUNNER warnings I see in the output (I don't export a setting for MELANGE_RUNNER in my shell):

```diff
diff --git a/Makefile b/Makefile
index 22a7778acf..dd9b8fe559 100644
--- a/Makefile
+++ b/Makefile
@@ -16,8 +16,8 @@ KEY ?= local-melange.rsa
 REPO ?= $(shell pwd)/packages
 QEMU_KERNEL_REPO := https://apk.cgr.dev/chainguard-private/
 
-ifeq (${MELANGE_RUNNER},)
-MELANGE_RUNNER = qemu
+ifeq ($(MELANGE_RUNNER),)
+export MELANGE_RUNNER = qemu
 $(warning ****************************** WARNING ******************************)
 $(warning *** MELANGE_RUNNER is unset. The default runner is now qemu, which)
 $(warning *** requires chainctl authentication to access the Chainguard kernel.)
@@ -338,7 +338,7 @@ endef
 define repo_token_if_needed
 $(eval pkgname := $1)
 
-@if [ -n "$$($(MAKE) -s compile/$(pkgname) | jq -r '.. | .uses? | strings | select(test("auth/guarded-repo"))')" ]; then \
+@if [ -n "$$($(MAKE) --no-print-directory -s compile/$(pkgname) | jq -r '.. | .uses? | strings | select(test("auth/guarded-repo"))')" ]; then \
 	echo "Creating guarded repo token for $(pkgname)…"; \
 	$(MAKE) repo-token/$(pkgname); \
 else \
@@ -349,7 +349,7 @@ endef
 define libraries_token_if_needed
 $(eval pkgname := $1)
 
-@if [ -n "$$($(MAKE) -s compile/$(pkgname) | jq -r '.. | .uses? | strings | select(test("auth/(?!guarded-repo)"))')" ]; then \
+@if [ -n "$$($(MAKE) --no-print-directory -s compile/$(pkgname) | jq -r '.. | .uses? | strings | select(test("auth/(?!guarded-repo)"))')" ]; then \
 	echo "Creating guarded libraries token for $(pkgname)…"; \
 	$(MAKE) lib-token/$(pkgname); \
 else \
```

but I worry `--no-print-directory` might be a GNU makeism.

smoser:
>Using the following patch to add --no-print-directory makes it work for me,

drive by comments.

2 other options:
1. make an actual target pkgname.yaml.json and make other targets depend on that.  fine to put that into a compiled/ directory (compiled/pkgname.json or something) then parse that with the jq above.
2. just redirect make's output to /dev/null

EyeCantCU:
I really like the idea of writing the compiled manifest to `compiled/packagename.json`... Let's do that. Less error prone, and we don't have to silence output for the compile target to get what we need

EyeCantCU:
Oh my gosh, this is so much better

pipelines/auth/guarded-repo.yaml
```diff
@@ -0,0 +1,43 @@
...
+    description: which GitHub repo to clone (e.g., "chainguard-dev/ecosystems-cassandra")
+    required: true
+  identity:
+    description: the identity to use for OctoSTS
+    required: false
+    default: elastic-build
+  cleanup-token:
+    description: whether or not to remove token file after use
+    default: true
+
```
stevebeattie, line 21: I like this but unfortunately, there are a few packages that have build pipelines that end up with multiple `uses: iamguarded/build-compat` steps, so either the `cleanup-token` option needs to be surfaced upwards into that pipeline or we drop the cleanup operation entirely.

Current offenders are in:
```
cert-manager-1.19.yaml
valkey-8.1.yaml
clickhouse-25.9.yaml
```

stevebeattie (changes requested):

pipelines/auth/guarded-repo.yaml
EyeCantCU: I've disabled specifically for iamguarded for now and left behavior in tact. Validated there's nowhere else we do this

stevebeattie (approved):
Thanks, this is working well for me.

We should probably have a followup to have `make clean` clear out the `compiled/` directory, and I would like to figure out a decent way to clean out the token before any build steps start, without breaking multi-pipeline builds.

justinvreeland (approved):

PR #60319, opened 2025-07-22
add helm-set-status plugin package

yashpimple:
Fixes:

Related: https://github.com/k3s-io/helm-set-status

#### For new package PRs only

- [] This PR is marked as fixing a pre-existing package request bug

#### For new version streams

#### For security-related PRs

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/helm-set-status.yaml
+++ b/helm-set-status.yaml
@@ -1,0 +1,72 @@
+package:
+  name: helm-set-status
+  version: 0.3.0
+  epoch: 0
+  description: Helm plugin to set release status
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - ca-certificates-bundle
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - git
+      - go
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/k3s-io/helm-set-status.git
+      tag: v${{package.version}}
+      expected-commit: 260bbd88a52aba1a76d44ed2ae66dccf5615dbf3
+
+  - uses: go/bump
+    with:
+      deps: |-
+        github.com/containerd/containerd@v1.7.27
+        github.com/docker/docker@v26.1.5
+        golang.org/x/net@v0.38.0
+        golang.org/x/oauth2@v0.27.0
+        helm.sh/helm/v3@v3.17.4
+
+  - runs: |
+      cat Makefile
+      make build
+      mkdir -p "${{targets.contextdir}}/root/.local/share/helm/plugins/helm-set-status"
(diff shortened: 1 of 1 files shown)


helm-set-status.yaml
```diff
@@ -0,0 +1,78 @@
...
+        golang.org/x/oauth2@v0.27.0
+        helm.sh/helm/v3@v3.17.4
+
+  # Dropping helm plugins in /usr/libexec since they are binaries and
+  # configuration intended to be run by the helm command.
+  # See https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s07.html
+  # Installers of this package will have to set this environment var
+  # HELM_PLUGINS=/usr/libexec/helm-plugins
+  - runs: |
+      cat Makefile
```
mritunjaysharma394: no need for cat of Makefile

helm-set-status.yaml
```diff
@@ -0,0 +1,78 @@
...
+      make build
```
mritunjaysharma394: can we use go/build pipeline rather? Also, we can configure INSTALL_DIR there too by either changing our output: or using symlink and you can add another pipeline to copy the config and yamls

yashpimple: updated the package to make use of `go/build` pipeline 👍

helm-set-status.yaml
```diff
@@ -0,0 +1,70 @@
...
+  - uses: go/build
+    with:
+      packages: ./cmd/helm-set-status
+      output: helm-set-status
+      tags: static_build,osusergo
+      ldflags: |
+        -X main.version=v${{package.version}}
+
+  - runs: |
+      INSTALL_DIR="${{targets.destdir}}/usr/libexec/helm-plugins/${{package.name}}"
```
anushkamittal20, line 39: is the binary also expected at `/bin/helm-set-status`: https://github.com/k3s-io/helm-set-status/blob/main/Makefile#L27

yashpimple: @anushkamittal20 the upstream Makefile builds binary at `bin/helm-set-status` (i.e., relative `./bin/ `directory), but in context of our `go/build` pipeline step, the binary is built directly into `/usr/bin/helm-set-status` further is then symlink to `/usr/libexec/helm-plugins/helm-set-status/` which is the expected path for Helm plugins (by default).

also, since Helm discovers plugins using $HELM_PLUGINS env variable (not /bin or /usr/bin), we explicitly set `HELM_PLUGINS=/usr/libexec/helm-plugins` in our tests env and it works as intended. This path is dynamic and configurable by users, so helm doesn’t rely on a fixed location like `/bin`. Here is similar plugin for reference [helm-mapkubeapis](https://github.com/wolfi-dev/os/pull/59565/files#diff-ccb110e401937a9a4d0f9d6f691a178c17c27f5f3f18d869493be256421634ffR37).

anushkamittal20 (approved):

Aditevil (approved):

PR #60324, opened 2025-07-22
auditwheel allow libanl linking

xnox:
libanl is a glibc component between versions 2.2.5 and 2.33, which
should be allowed by default in those older policies. Newer builds do
not gain such a dependency, as libanl got folded into libc itself, but
it is harmless.

Cherry-picked from:
- https://github.com/pypa/auditwheel/pull/607

### DIFF
--- a/py3-auditwheel.yaml
+++ b/py3-auditwheel.yaml
@@ -1,7 +1,7 @@
 package:
   name: py3-auditwheel
   version: "6.4.1"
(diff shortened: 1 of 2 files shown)


sergiodj (approved):

smoser (approved):

techalchemy (approved):
lgtm

PR #60457, opened 2025-07-23
kube-fluentd-operator: bump dependency to fix segfault errors

MickeyPvX:
Updating the `fluentd-plugin-systemd` dependency to a version that resolves [this](https://github.com/ledbettj/systemd-journal/pull/96) issue


kube-fluentd-operator.yaml
```diff
@@ -16,6 +16,7 @@ package:
       - merged-bin
       - ruby-3.2
       - ruby3.2-bundler
+      - systemd
```
xnox, line 19: this does not seem right.

systemd should not be pulled into the container or running in the container.

does the new dependency execute some binaries from that package? or does it require services to be running from it?

xnox (changes requested):

kube-fluentd-operator.yaml
MickeyPvX: nope, you're right, had that in based on the discussion on the issue, but shouldn't be in the final package build...fixing...

xnox (approved):

xnox (changes requested):
> Completed in 2m — No packages were built!

is this change out of date with main line? and needs rebase + epoch rebump?

MickeyPvX:
> > Completed in 2m — No packages were built!
> 
> is this change out of date with main line? and needs rebase + epoch rebump?

Odd - rebuilt with the latest package and now I'm getting the segfault error...let me see what's going on

xnox:
> > > Completed in 2m — No packages were built!
> > 
> > 
> > is this change out of date with main line? and needs rebase + epoch rebump?
> 
> Odd - rebuilt with the latest package and now I'm getting the segfault error...let me see what's going on

fun..... and does adding systemd package in the container / test environment makes things better?

MickeyPvX:
> > > > Completed in 2m — No packages were built!
> > > 
> > > 
> > > is this change out of date with main line? and needs rebase + epoch rebump?
> > 
> > 
> > Odd - rebuilt with the latest package and now I'm getting the segfault error...let me see what's going on
> 
> fun..... and does adding systemd package in the container / test environment makes things better?

you read my mind - just did it, and yes, it seems to get the container running properly

xnox (approved):

PR #60467, opened 2025-07-23
add symlink-check subpackage to tw

kranurag7:
adds symlink-check subpackage to tw and bumps to v0.0.22


pipelines/test/tw/ldd-check.yaml
```diff
@@ -14,7 +14,6 @@ inputs:
       The files to exclude from running `ldd` on and check for missing deps.
       Multiple file names must be separated by space.
     required: false
-    default: ""
```
smoser: i think we need this back on tw/

smoser (approved):

PR #60495, opened 2025-07-23
JuiceFS Compat: Update runtime dependency

AnmolVirdi:
- Add `dash-binsh` package which is needed in entrypoint when deployed in `k8s` runtime by `juicefs-csi-driver` helm chart.


Aditevil (commented):
epoch bump and its GTG(good to go :))

Aditevil (approved):

PR #60680, opened 2025-07-25
spilo pgqd

utieyin:
pgqd-3.5: new package

This is a package required by spilo-17

#### For new package PRs only


pgqd.yaml
```diff
@@ -0,0 +1,163 @@
...
+      ./autogen.sh
+
+  - uses: autoconf/configure
+
+  - uses: autoconf/make
+
+  - uses: autoconf/make-install
+
+  - runs: |
+      mkdir -p ${{targets.destdir}}/etc/pgqd
```
anushkamittal20: should this be in a compat pkg?

pgqd.yaml
```diff
@@ -0,0 +1,163 @@
...
+    packages:
+      - build-base
+      - libevent-dev
+      - libtool
+      - pkgconf-dev
+      - postgresql-${{vars.pg-version}}-dev
+      - py3-docutils
+
+vars:
+  pg-version: "17"
```
anushkamittal20, line 24: I might be lacking context here, is this version streamed? or maybe we can have a comment on why pg17?

utieyin: Oh no, this is just a variable to pin the version of postgresql that we want to use to package pgqd

pgqd.yaml
utieyin: I would be inclined to agree but a single file in a compat package that upstream ships is perhaps an overkill?

pgqd.yaml
Dentrax: Upstream didn't version stream this so we can use the latest pg here.

Dentrax:
LGTM as APK content matches with upstream's, and no additional build-time deps are needed.
```
-rwxr-xr-x 0/0          148368 2025-07-30 13:04 usr/bin/pgqd
-rw-r--r-- 0/0             579 2025-07-30 13:04 etc/pgqd.ini
```

Dentrax (approved):

PR #60693, opened 2025-07-25
pipelines/test/virtualpackage: extend to verify APK declares virtual packages as expected

aborrero:
Extend the virtualpackage test to verify the APK declares the virtual package as expected.

### DIFF
--- a/pipelines/test/virtualpackage.yaml
+++ b/pipelines/test/virtualpackage.yaml
@@ -13,7 +13,7 @@ inputs:
     default: ''
 
 pipeline:
-  - name: virtual package checks
+  - name: virtual package installation checks
     description: Check that a virtual package does not install itself
     runs: |
       get_installed_version() {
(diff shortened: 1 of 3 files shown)


debasishbsws (dismissed):
Love it

aborrero:
image

pipelines/test/virtualpackage.yaml
```diff
@@ -41,3 +41,24 @@ pipeline:
           exit 1
         fi
       fi
+
+  - name: virtual package configuration checks
+    description: Check that a a real package declares it provides the expected virtual package
```
smoser: typo.

pipelines/test/virtualpackage.yaml
```diff
@@ -41,3 +41,24 @@ pipeline:
...
+    runs: |
+      virtual_pkg=${{inputs.virtual-pkg-name}}
+      real_pkg=${{inputs.real-pkg-name}}
+
+      if [ "${real_pkg}" == "" ] ; then
+        echo "SKIP: no 'real-pkg-name' was set, so skipping evaluation."
+        exit 0
+      fi
+
+      declared_provides=$(apk info --installed --provides $real_pkg | tail -n +2 | grep -v ^$)
```
smoser: does `--quiet` help you here? then you can drop the tail -n+2 i think.

you should also quote "$real_pkg" there.

pipelines/test/virtualpackage.yaml
```diff
@@ -41,3 +41,24 @@ pipeline:
...
+
+      if echo "$declared_provides" | grep -q $virtual_pkg ; then
```
smoser: quote virtual_pkg and do you want to limit the grep to whole line?  `grep -x -F "${virtual_pkg}"` will require matching a whole line.

smoser (commented):
some comments inline there.
2 more general comments
1. virtual packages are non-ideal and we are kind of moving away from them.  in the python examples you have here, only a human should ever type `apk add py3-ansible-core` and when they do, they need to be aware that it might not actually have the same python as a subsequent or prior `apk add py3-other-package`.
2. I'm happy to have stuff like this in tw, where it can reasonably be written in non-shell.

sergiodj (changes requested):
I've also left comments on https://github.com/chainguard-dev/extra-packages/pull/3158 which apply to the change you're doing here (I've just noticed that my comments overlap with @smoser's ones, but there are other minor things I've asked too).

aborrero:
thanks for the review @smoser and @sergiodj I just sent an updated version.

sergiodj (approved):

PR #60710, opened 2025-07-25
gnupg: add patch to not use 3DES in FIPS mode

sil2100:
Might be useful when working with more restrictive libgcrypt libraries.


xnox (approved):

xnox:
3DES is museum cryptography and dead to us.

we have switched things like this in all other implementations we ship too.

PR #60714, opened 2025-07-25
shellcheck: shard 6

AbduAzim424:
Fix some shell pre-commit errors before enabling globally

### DIFF
--- a/terraform-provider-kubernetes.yaml
+++ b/terraform-provider-kubernetes.yaml
@@ -29,9 +29,9 @@ pipeline:
 
   - runs: |
       GOARCH="$(go env GOARCH)"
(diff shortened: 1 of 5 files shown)


terraform-provider-kubernetes.yaml
```diff
@@ -18,7 +18,7 @@ pipeline:
 
   - uses: go/bump
     with:
-      deps: |-
+      deps: |
```
dannf: why this change?

dannf (changes requested):
One question inline

terraform-provider-kubernetes.yaml
AbduAzim424: I reverted these in other repos. I wasn't sure initially, Will fix instantly

dannf (approved):

PR #60766, opened 2025-07-25
spilo python-etcd

utieyin:
python-etcd: new package

Fixes:

Related:

#### For new package PRs only


py3-etcd.yaml
```diff
@@ -0,0 +1,89 @@
...
+
+  - name: py3-supported-${{vars.pypi-package}}
+    description: meta package providing ${{vars.pypi-package}} for supported python versions.
+    dependencies:
+      runtime:
+        - py3.10-${{vars.pypi-package}}
+        - py3.11-${{vars.pypi-package}}
+        - py3.12-${{vars.pypi-package}}
+        - py3.13-${{vars.pypi-package}}
+        # - uses: strip
```
Dentrax: Commented out

py3-etcd.yaml
```diff
@@ -0,0 +1,89 @@
...
+      packages:
+        - etcd~3.5
+    environment:
+      ETCD_ENABLE_V2: true
+  pipeline:
+    - uses: py/one-python
+      with:
+        content: |
+          etcd &
+          sleep 2
```
Dentrax: Can we `wait-for-it` the port to avoid CI flakes?

py3-etcd.yaml
```diff
@@ -0,0 +1,92 @@
+package:
+  name: py3-etcd
+  version: "0.4.5"
+  epoch: 0
+  description: A python client for etcd.
+  copyright:
+    - license: MIT
+
+vars:
+  py-version: "3.13"
```
Dentrax: I think this is not needed as we use `py-versions`?

Dentrax (dismissed):

Dentrax (approved):

PR #60869, opened 2025-07-28
fix syntax issues in configs after latest yam release

kranurag7:
fix syntax issues in configs after latest yam release

    we release new version of yam which sorts out dependencies. this requires
    changes in configs which are part of this diff.
    epoch is not bumped because this patch contains the same dependencies
    but in a sorted order.


aborrero (dismissed):
LGTM.

AmberArcadia:
Apologies for the chaos, I wasn't expecting CI to hoover up my fix :grimacing: I appreciate you beating me to it with the cleanup fix though! Happy to help out with the other repos too.

dannf (approved):

PR #60881, opened 2025-07-28
ICU 77 - pre-transition linker fixes

javacruft:
pkg-config from ICU >= 76 is more selective in terms of which libraries are returned for ldflags.

Update packages which need both icuuc and icui18n but only provide one when calling pkg-config.

These updates have also been tested with https://github.com/wolfi-dev/os/pull/57732


sil2100 (approved):
Thanks!

PR #60897, opened 2025-07-28
regclient: build other binaries

Dentrax:
Build regbot and regsync as well.


kranurag7 (approved):
very clean; with provides in place; this looks good to land.

PR #60898, opened 2025-07-28
nova initial package build

developer-guy:
Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/nova.yaml
+++ b/nova.yaml
@@ -1,0 +1,77 @@
+package:
+  name: nova
+  version: "3.11.6"
+  epoch: 0
+  description: Find outdated or deprecated Helm charts running in your cluster.
+  copyright:
+    - license: Apache-2.0
+
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: 49362fe62667d6dda248ba58c119cdbe15eee5d7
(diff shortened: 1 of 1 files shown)


nova.yaml
```diff
@@ -0,0 +1,77 @@
...
+        - helm
+        - kubectl
+        - jq
+  pipeline:
+    - uses: test/kwok/cluster
+    # let's deploy something old on top of the cluster
+    - runs: |
+        kubectl config view --minify --raw > /tmp/kubeconfig.yaml
+        # the latest version of components was 0.8.0 at the time of writing this, but we want to test against an older version
+        kubectl apply --kubeconfig /tmp/kubeconfig.yaml -f https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.7.2/components.yaml
```
Dentrax, line 56: Can we use the latest version instead of 0.7.2?

developer-guy: This is on purpose so this is how nova works actually so we need to deploy an older version and run nova against the older version to make sure that this is an older version and it needs to be updated

flexiondotorg (approved):
Looks good. Test confirms outdated resources are detected.

Dentrax (approved):

PR #60924, opened 2025-07-28
ICU 77: Tranche 0 no change rebuilds

javacruft:
No change rebuilds for ICU 77 transition.


tensorflow-core.yaml
```diff
@@ -2,7 +2,7 @@ package:
   name: tensorflow-core
   description: Framework for data-graph oriented computing (core libraries, oneDNN build)
   version: "2.19.0"
-  epoch: 3
+  epoch: 5
```
dannf: Skipped a number

dannf (changes requested):

tensorflow-core.yaml
javacruft: Dropped as seeing a test failure (unrelated to transition).

dannf (dismissed):

murraybd:
I removed php-8.4 because it already rebuild with icu-77.

https://apk.chaindag.dev/https/packages.wolfi.dev/os/x86_64/php-8.4-8.4.11-r0.apk@sha1:004cb46fef69cf04fc510eb42d0e067353a0d790/.melange.yaml

dannf (approved):

PR #60988, opened 2025-07-29
add missing runtime dependency for klipper-helm

yashpimple:
Related: https://github.com/k3s-io/klipper-helm

The `klipper-helm` entry script requires few utilities that were missing from runtime dependencies, updated the PR and added missing `busybox` since `grep` and `tr` used by the entry script and `jq`: required for JSON processing in helm operations

These dependencies resolve test failures during k8s testing.


developer-guy (approved):
looks ok to me!

PR #60998, opened 2025-07-29
postgresql plpython

utieyin:
Compiling postgresql-17 with python libraries support


utieyin:
There are a few extensions (plpgsql, plperl, plpython) that can be compiled into postgres by passing flags.
I have enable the python extension in this change. Since this is an existing image, not sure if all the python header files need to be completely moved into the specific package.

postgresql-17.yaml
```diff
@@ -43,6 +43,7 @@ environment:
       - net-tools
       - openldap-dev
       - openssl-dev
+      - py3-supported-python-dev
```
dannf: please use `py3-supported-build-base-dev`

postgresql-17.yaml
```diff
@@ -195,6 +197,23 @@ subpackages:
       provides:
         - libpq=${{package.full-version}}
 
+  - name: plpython3
```
dannf: could you python multi-version this? Right now it is only being built for python3.13. Ideally we have interchangeable `py3.X-plpython` packages that can be selected depending on which python version an image requires.

postgresql-17.yaml
```diff
@@ -241,7 +260,6 @@ subpackages:
           packages:
             - pkgconf
       pipeline:
-        - uses: test/pkgconf
```
dannf, line 244: why is this being dropped?

postgresql-17.yaml
```diff
@@ -195,6 +197,23 @@ subpackages:
...
+      provider-priority: 17
+      provides:
+        - plpython=${{package.full-version}}
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/usr/lib/postgresql${{vars.major-version}}
+          mkdir -p ${{targets.subpkgdir}}/usr/share/postgresql${{vars.major-version}}/extension
+      - runs: |
+          mv ${{targets.destdir}}/usr/share/postgresql${{vars.major-version}}/extension/plpython* ${{targets.subpkgdir}}/usr/share/postgresql${{vars.major-version}}/extension/
+          mv ${{targets.destdir}}/usr/lib/postgresql${{vars.major-version}}/plpython* ${{targets.subpkgdir}}/usr/lib/postgresql${{vars.major-version}}/
```
dannf: note:  we're working on adding `shellcheck` support as a linter, and I think this will cause a shellcheck error and thus a failure when that lands. It's recommended alternative is to mv `$(echo ${{targets.destdir}}/usr/lib/postgresql${{vars.major-version}}/plpython*)` so that the shell will expand the glob.

dannf (dismissed):

postgresql-17.yaml
dannf: (alternatively, please pin everything to use `python3.13` and not `python3`, so things don't magically break when `python3.14` becomes the default)

utieyin: we would have to recompile postgres-17 and all other versions of postgres against python3.x in a different package which is probably an overkill and so have pinned this extension against 3.13 python.

Like we have done in other packages such as spark where the main package is already version streamed, we drop a fixed version of the dependent package

postgresql-17.yaml
utieyin: Formatting error, fixed

Dentrax (dismissed):

postgresql-17.yaml
```diff
@@ -43,11 +43,16 @@ environment:
       - net-tools
       - openldap-dev
       - openssl-dev
+      - python-${{vars.py-version}}-dev
```
dannf: please use `py${{vars.py-version}}-build-base-dev`. `python-X-dev` will land a `/usr/bin/python3` that points at `python3.13`. But things that build with it will inherent a dependency on `python3`, not `python3.13`. This means that when `python3` becomes `python3.14`, things will break.

This of course means that you need to figure out how to make the build use `/usr/bin/python3.13` instead of `/usr/bin/python3`.

postgresql-17.yaml
```diff
@@ -92,6 +97,7 @@ pipeline:
         --with-gssapi \
         --with-ldap \
         --with-llvm \
+        --with-python \
         --enable-thread-safety
```
dannf: I'm concerned that nothing in this configuration is telling it to use `/usr/bin/python${{vars.py-version}}` to do the build. Maybe the build is smart enough to figure out that only `python3.13` is there, and that's what it should use to build it. But it could also be that some other build dependency is pulling in `python3`, and it is being used even though you happen to have `python3.13` in the build environment. For some reason elastic hasn't built this, so I can't verify.

Can you somehow demonstrate that `/usr/bin/python3.13` is what is being used in the build?

postgresql-17.yaml
```diff
@@ -195,6 +198,23 @@ subpackages:
...
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/usr/lib/postgresql${{vars.major-version}}
+          mkdir -p ${{targets.subpkgdir}}/usr/share/postgresql${{vars.major-version}}/extension
+      - runs: |
+          mv $(echo ${{targets.destdir}}/usr/share/postgresql${{vars.major-version}}/extension/plpython*) ${{targets.subpkgdir}}/usr/share/postgresql${{vars.major-version}}/extension/
+          mv $(echo ${{targets.destdir}}/usr/lib/postgresql${{vars.major-version}}/plpython*) ${{targets.subpkgdir}}/usr/lib/postgresql${{vars.major-version}}/
+    test:
+      pipeline:
+        - uses: test/tw/ldd-check
```
dannf: Is there a simple test case for the python extension you could add - even just something that prints e.g. a version? It would be nice if we could run that test w/ the python3.13 interpreter to make sure we're building for the version we think we are.

utieyin: Have added tests to verify python version in plpython. Change the py-test-version to something other than 3.13 and tests fail

postgresql-17.yaml
utieyin: Have added a PYTHON env variable to specify the version of python to use in compiling postgres

postgresql-17.yaml
```diff
@@ -195,6 +200,68 @@ subpackages:
...
+          mkdir -p ${{targets.subpkgdir}}/usr/share/postgresql${{vars.major-version}}/extension
+      - runs: |
+          mv $(echo ${{targets.destdir}}/usr/share/postgresql${{vars.major-version}}/extension/plpython*) ${{targets.subpkgdir}}/usr/share/postgresql${{vars.major-version}}/extension/
+          mv $(echo ${{targets.destdir}}/usr/lib/postgresql${{vars.major-version}}/plpython*) ${{targets.subpkgdir}}/usr/lib/postgresql${{vars.major-version}}/
+    test:
+      environment:
+        contents:
+          packages:
+            - postgresql-17-client
+            - postgresql-17
```
Dentrax: Let's get the versions from transform matrix

postgresql-17.yaml
```diff
@@ -17,6 +17,7 @@ environment:
   environment:
     # Used to fetch docker scripts
     ALPINE_VERSION: "3.21"
+    PYTHON: "python3.13"
```
dannf: Since we can't ${{vars.py-version}} here, could you set it in the `configure` step? That does mean not no longer using the configure pipeline. See tevent.yaml as an example. (We really should add environment variables as options to that pipeline...)

Otherwise, when we do automation to try and bump python versions, we may miss this one.

postgresql-17.yaml
```diff
@@ -195,6 +200,68 @@ subpackages:
       provides:
         - libpq=${{package.full-version}}
 
+  - name: plpython3
```
dannf: Please include the postgres major version in the name like with the other packages.

postgresql-17.yaml
```diff
@@ -195,6 +200,68 @@ subpackages:
...
+    description: PostgreSQL Python libraries
+    dependencies:
+      provider-priority: 17
+      provides:
+        - plpython=${{package.full-version}}
```
dannf: I don't see a reason to introduce a `plpython` name. virtualpackages can cause mysterious issues with our resolver, so I prefer we only add them if we have a real use case. This just seems like an alias.

postgresql-17.yaml
```diff
@@ -195,6 +200,68 @@ subpackages:
       provides:
         - libpq=${{package.full-version}}
 
+  - name: plpython3
+    description: PostgreSQL Python libraries
+    dependencies:
```
dannf: is melange already adding the appropriate `python3.13`-specific dependencies? (I imagine `so:libpython3.13` or similar?). See below comment in testing

dannf (dismissed):
Looking good :)

dannf (dismissed):
The `3.10` version looks good:
```
~ $ apk info -L py3.10-plpython
WARNING: opening /home/dann_frazier_chainguard_dev/os/packages: No such file or directory
WARNING: opening from cache https://apk.cgr.dev/wolfi-presubmit/e51dce432247a198c3af56bd61956b0b0e0be658: No such file or directory
WARNING: opening from cache https://packages.wolfi.dev/os: No such file or directory
py3.10-plpython-17.5-r10 contains:
usr/lib/postgresql17/plpython3.so
usr/share/postgresql17/extension/plpython3u--1.0.sql
usr/share/postgresql17/extension/plpython3u.control
var/lib/db/sbom/py3.10-plpython-17.5-r10.spdx.json

~ $ ldd /usr/lib/postgresql17/plpython3.so
	linux-vdso.so.1 (0x00007fa699337000)
	libpython3.10.so.1.0 => /usr/lib/libpython3.10.so.1.0 (0x00007fa698f61000)
	libc.so.6 => /usr/lib/libc.so.6 (0x00007fa698d61000)
	libm.so.6 => /lib/libm.so.6 (0x00007fa698c74000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fa699339000)
```

But the `3.11`-`3.13` versions seem to be shipping the same `3.10` binary:
```
~ $ apk info -L py3.11-plpython
WARNING: opening /home/dann_frazier_chainguard_dev/os/packages: No such file or directory
WARNING: opening from cache https://apk.cgr.dev/wolfi-presubmit/e51dce432247a198c3af56bd61956b0b0e0be658: No such file or directory
WARNING: opening from cache https://packages.wolfi.dev/os: No such file or directory
py3.11-plpython-17.5-r10 contains:
usr/lib/postgresql17/plpython3.so
usr/share/postgresql17/extension/plpython3u--1.0.sql
usr/share/postgresql17/extension/plpython3u.control
var/lib/db/sbom/py3.11-plpython-17.5-r10.spdx.json

~ $ ldd /usr/lib/postgresql17/plpython3.so
	linux-vdso.so.1 (0x00007f6003391000)
	libpython3.10.so.1.0 => /usr/lib/libpython3.10.so.1.0 (0x00007f6002fbb000)
	libc.so.6 => /usr/lib/libc.so.6 (0x00007f6002dbb000)
	libm.so.6 => /lib/libm.so.6 (0x00007f6002cce000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f6003393000)
```

The checksums are the same in each package, so I suspect some kind of install script bug.

postgresql-17.yaml
```diff
@@ -407,6 +417,96 @@ subpackages:
...
+            --with-libxml \
+            --with-libxslt \
+            --with-icu \
+            --with-lz4 \
+            --with-zstd \
+            --with-gssapi \
+            --with-ldap \
+            --with-llvm \
+            --with-python \
+            --enable-thread-safety
```
dannf: These configure options are now duplicated, and at risk of becoming out-of-sync. I'd suggest using a variable. See https://github.com/wolfi-dev/os/blob/f79bd298614f506707b99320ab727f2d0dd24a85/talloc.yaml#L10 as an example.

I'd also suggest using `--with-python` for the main package as well - but just delete the python plugin afterwards. It's possible that there is - or will be - code that needs to be enabled in the main package to make use of the python plugin.

postgresql-17.yaml
```diff
@@ -407,6 +417,96 @@ subpackages:
...
+            CREATE OR REPLACE FUNCTION py_version() RETURNS text AS \$\$
+            import sys
+            return f'Python {sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}'
+            \$\$ LANGUAGE plpython3u;" 2>&1
+
+            # Get the Python version from plpython
+            PLPYTHON_VERSION=$(psql -t -A -d plpython_testdb -c "SELECT py_version();")
+
+            # Get the system Python version
+            SYSTEM_VERSION=$(python --version)
```
dannf: I was given the feedback that we should always use `sys.version_info` for comparisons, because we can't guarantee that the `--version` output will be stable. You could just run the same code you're doing above here, like in this example:
https://github.com/wolfi-dev/os/blob/f79bd298614f506707b99320ab727f2d0dd24a85/mlflow.yaml#L198

Of course, you could also just compare [ "$PLPYTHON_VERSION" = "${{vars.py-version}}" ] here and avoid the whole system version detection. We're relying on that macro already to install the system python, so installing it then querying it kind of "the long way around".

postgresql-17.yaml
```diff
@@ -407,6 +417,96 @@ subpackages:
...
+            --with-libxslt \
+            --with-icu \
+            --with-lz4 \
+            --with-zstd \
+            --with-gssapi \
+            --with-ldap \
+            --with-llvm \
+            --with-python \
+            --enable-thread-safety
+      - uses: autoconf/make
```
dannf, line 422: I wonder if there's a more specific target we could call to avoid building *everything* 5 times.

dannf (dismissed):
Great! Let's get this merged!

I have some non-blocking comments inline that you might consider for a follow-up update.

postgresql-17.yaml
utieyin: Not that I know :-)

postgresql-17.yaml
utieyin: Can't use vars since the config options also contain vars, so went with env vars

postgresql-17.yaml
```diff
@@ -13,10 +13,21 @@ package:
...
+      3.11: "311"
+      3.12: "312"
+      3.13: "313"
+
 environment:
   environment:
     # Used to fetch docker scripts
     ALPINE_VERSION: "3.21"
+    # Common configure flags for both main package and plpython builds
+    CONFIGURE_FLAGS: >-
```
dannf, line 29: nice!

dannf (approved):

Dentrax (approved):

PR #61024, opened 2025-07-29
feat: Add Conjur CLI

EyeCantCU:


conjur-cli.yaml
```diff
@@ -0,0 +1,35 @@
...
+  name: conjur-cli
+  description: Conjur command line interface
+  version: 8.1.2
+  epoch: 0
+
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: 556d4a21686e782dfa27177615a0882de9b9f7de
+      repository: https://github.com/cyberark/conjur-cli-go
```
mamccorm, line 11: > This repository was archived by the owner on Dec 7, 2023. It is now read-only.

mamccorm (dismissed):
Looks like the project is archived

mamccorm (approved):

conjur-cli.yaml
```diff
@@ -0,0 +1,35 @@
...
+        -X "github.com/cyberark/conjur-cli-go/pkg/version.Tag=$(git rev-parse --short HEAD)"
+        -X "github.com/cyberark/conjur-cli-go/pkg/version.Version=${{package.version}}"
+
+update:
+  enabled: true
+  github:
+    identifier: cyberark/conjur-cli-go
+    strip-prefix: v
+    tag-filter: v
+    use-tag: true
```
mamccorm, line 29: Nit: do we need this? looks like they are publishing releases

EyeCantCU: We don't, but we're already using the tag to reference the release when we check the repository out with git so we might as well use the same behavior when checking for updates

PR #61030, opened 2025-07-29
Adjust file permissions across several packages

egibs:


logstash-9.yaml
```diff
@@ -305,7 +305,7 @@ subpackages:
             - username: logstash
               gid: 1001
               uid: 1001
-          run-as: 1001
+          run-as: 0
```
justinvreeland, line 302: This test sets up users to run the the test as that user. If this is supposed to work as that user then this shouldn't be changed and if the test doesn't need to use the user you should get rid of the creation of the user and group.

repmgr-17.yaml
```diff
@@ -207,8 +210,8 @@ subpackages:
                       /opt/bitnami/scripts/postgresql-repmgr/entrypoint.sh /opt/bitnami/scripts/postgresql-repmgr/run.sh
                 timeout: 60
                 expected_output: |
-                  Initializing Repmgr
-                  Initializing PostgreSQL
+                  Querying all partner nodes for common upstream node
+                  There are no nodes with primary role
                   Starting PostgreSQL
```
justinvreeland: Do you know what's going on here? I don't see anything that would suggest we'd expect the output to change.  Why is the permissions update causing this?

justinvreeland (changes requested):

repmgr-17.yaml
egibs: I don't. The output between the Bitnami test and the Iamguarded tests is entirely different for some reason.

logstash-9.yaml
egibs: Right. I was confused by this because the QEMU runner doesn't work with a non-root user because it can't configure any of the mounts on startup and the Iamguarded test also runs as root.

repmgr-17.yaml
egibs: Fixed in [`cbf01e7` (#61030)](https://github.com/wolfi-dev/os/pull/61030/commits/cbf01e7bd73c1a54401bc3441febd06159650c83).

logstash-9.yaml
egibs: Fixed in [`cbf01e7` (#61030)](https://github.com/wolfi-dev/os/pull/61030/commits/cbf01e7bd73c1a54401bc3441febd06159650c83).

We run as `0` but use sudo to run the actual commands as the `logstash` user.

justinvreeland (approved):

PR #61032, opened 2025-07-29
Fix CVE: CVE-2022-1471 cassandra-reaper

debasishbsws:


debasishbsws:
not possible. pending upstream fix is still valid.

PR #61082, opened 2025-07-29
feat: adds package kube-arangodb

rahulduvedi:
`kube-arangodb:1.2.50`: new package

### DIFF
--- a/kube-arangodb.yaml
+++ b/kube-arangodb.yaml
@@ -1,0 +1,114 @@
+package:
+  name: kube-arangodb
+  version: "1.2.50"
+  epoch: 0
+  description: ArangoDB Kubernetes Operator - manages deployments of the ArangoDB database in Kubernetes
+  copyright:
+    - license: Apache-2.0
+
+environment:
+  environment:
+    # Disable CGO to create a statically linked binary that works in musl-based ArangoDB pods
+    CGO_ENABLED: "0"
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/arangodb/kube-arangodb/
+      tag: ${{package.version}}
+      expected-commit: 7ff4e5f5c79b2822cbd7d50afd7eaaf801e9451d
+
+  - uses: go/bump
+    with:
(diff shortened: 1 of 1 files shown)


kube-arangodb.yaml
```diff
@@ -0,0 +1,114 @@
...
+        -X github.com/arangodb/kube-arangodb/pkg/version.version=${{package.version}}
+        -X github.com/arangodb/kube-arangodb/pkg/version.buildDate=$(date -d@${SOURCE_DATE_EPOCH} -u +'%Y-%m-%dT%H:%M:%SZ')
+        -X github.com/arangodb/kube-arangodb/pkg/version.build=$(git rev-parse --short HEAD)
+
+  - runs: |
+      mkdir -p ${{targets.destdir}}/licenses
+      cp LICENSE ${{targets.destdir}}/licenses/LICENSE
+
+subpackages:
+  - name: ${{package.name}}-compat
```
klrmngr, line 37: instead of adding envoy as a runtime dep to the subpackage can we add envoy @ the image level?

kube-arangodb.yaml
```diff
@@ -0,0 +1,114 @@
+package:
+  name: kube-arangodb
+  version: "1.2.50"
+  epoch: 2
```
klrmngr:
```suggestion
  epoch: 0
```

klrmngr (dismissed):

kube-arangodb.yaml
kranurag7: fips equivalent of same is keeping it at package level; I don't have a strong preference here both achieve the same goal. If still; we can do a follow-up to move it at image level.

rahulduvedi: The upstream Makefile explicitly requires envoy via `--build-arg "ENVOY_IMAGE=$(ENVOY_IMAGE)"`, proving it's fundamental to kube-arangodb's operation , Making this a package-level dependency ensures it's always available and prevents runtime failures.

kranurag7 (approved):

PR #61088, opened 2025-07-29
neuvector-scanner: add neuvector-db as sub-package; remove separate package

kbsteere:
fix: add neuvector-db to neuvector-scanner as subpackage. This address issues of the
 database not being built every day and adds versioning to the database, based on the scanner version.
cleanup: remove separate nevector-db melange file. Subpackage exists with neuvector-scanner


neuvector-scanner.yaml
```diff
@@ -59,6 +61,33 @@ subpackages:
           mkdir -p ${{targets.contextdir}}/usr/local/bin
           install -Dm755 monitor/monitor ${{targets.contextdir}}/usr/local/bin/monitor
 
+  - name: neuvector-db
```
mamccorm, line 63: I see @EyeCantCU is on the reviews, but I recall an intentional decision to remove this in the past

neuvector-scanner.yaml
```diff
@@ -59,6 +61,33 @@ subpackages:
...
+          # Fetch NeuVector scanner from docker hub
+          # https://hub.docker.com/r/neuvector/scanner/tags
+          # we don't use the dbgen tool for this since it's very slow and
+          # super flaky. Extracting is faster and is more reliable.
+          # This may change in the future if neuvector isn't published to DH
+          crane export neuvector/scanner:${{package.version}} scanner.tar
+
+          # Extract image
+          mkdir -p scanner
+          tar xvf scanner.tar -C scanner
```
OddBloke: A minor nit: we could pass (I think) `etc/neuvector/db/cvedb` here to _only_ extract the part of the tarball we actually need, saving some build time and disk space.

neuvector-scanner.yaml
```diff
@@ -68,6 +97,9 @@ test:
 
 update:
   enabled: true
+  schedule:
+    period: daily
+    reason: NeuVector scanner is updated daily for database updates
```
OddBloke: I don't _think_ we need to build these `daily`: it looks like a new release gets tagged for every new database version, so I think we should almost always get a new version when the `daily` version would contain a change. (This also avoids having to build the trunk version of the `scanner` code, which I think would require a bit more analysis to choose to do.)

neuvector-scanner.yaml
```diff
@@ -1,22 +1,24 @@
...
   version: "3.875"
-  epoch: 0
+  epoch: 1
   description: NeuVector vulnerability scanner for the SUSE NeuVector Container Security Platform
   copyright:
     - license: Apache-2.0
   dependencies:
     runtime:
       - ca-certificates
-      - neuvector-db
```
OddBloke: This shouldn't be removed: we still need to pull in the subpackage when the main `-scanner` is installed.

neuvector-scanner.yaml
```diff
@@ -1,22 +1,24 @@
...
-      - neuvector-scanner-monitor
       - neuvector-scanner-task
       - neuvector-sigstore-interface
+    provides:
+      - neuvector-db
```
OddBloke: This should be removed: it currently means the main package shadows the subpackage.

neuvector-scanner.yaml
```diff
@@ -59,6 +61,33 @@ subpackages:
...
+          mkdir -p ${{targets.contextdir}}/etc/neuvector/db
+          cp scanner/etc/neuvector/db/cvedb ${{targets.contextdir}}/etc/neuvector/db/cvedb
+    test:
+      pipeline:
+        - runs: |
+            if [[ -f "/etc/neuvector/db/cvedb" ]]; then
+              echo "CVE DB found!" && exit 0
+            else
+              echo "CVE DB not found!" && exit 1
+            fi
```
OddBloke, line 85: Another nit, but in addition to this test, extending the top-level `test:` with something like `scanner -v -d /etc/neuvector/db/` might give us a legit test of the database; I wasn't able to confirm this works (because I can't currently install the DB :sweat_smile:), but it does emit:

```
Error: Read db file fail: open /etc/neuvector/db/cvedb: no such file or directory
```
and `--help` says:
```
  -v	show cve database version
```
so it might work as a standalone test.

neuvector-scanner.yaml
OddBloke: > I see @EyeCantCU is on the reviews, but I recall an intentional decision to remove this in the past

@mamccorm  To _remove_ this, or to move it to a separate package? We currently do have a separate `neuvector-db` package: this change will ensure that it remains current.

kbsteere: Only thing I can think of is that there is a neuvector-scanner-fips version. We could create a fips only db but it's kind of pointless since we are not doing anything to it in any way. Which means that package relies on this one to build first since this contains the db.

kbsteere: @EyeCantCU do you have the previous context as to why this was split out from neuvector scanner?

EyeCantCU: Hey Kyle. I would pull the database personally. It's not consumed by any of our images today, and it's likely to lag behind whatever database is provided in the latest image as it's updated constantly

cmwilson21:
Thanks for the contribution!

PR #61095, opened 2025-07-29
No change rebuild for the icu-77 transition

murraybd:


murraybd:
If https://github.com/wolfi-dev/os/pull/61101 is merged then qt5-qtbase should be able to pass its tests.

javacruft (approved):

PR #61137, opened 2025-07-30
code-server: fix issue with broken depends on musl.

javacruft:


sil2100 (dismissed):
Removing unneeded stuff is music to my ears (especially if it can help with broken depends generation).

sil2100 (approved):

PR #61152, opened 2025-07-30
chor: Using git monitor as a update monitor

debasishbsws:
The upstream has released alpha and beta version from last 20 releases, using git monitor to use tag and ignore unwanted regex pattern

**This do not required an `epoch` bump.**


aborrero (approved):
LGTM-

PR #61164, opened 2025-07-30
chore: Withdraw versions of Firefox that provide vendored libraries

EyeCantCU:
I'm seeing this from apk right now:

```
solving "libnss" constraint:   libnss-3.94-r0.apk disqualified because firefox-128.0-r0.apk already provides so:libfreeblpriv3.so
```

`no-depends` was enabled in all later versions of Firefox, but offending versions were never pulled


aborrero (approved):
LGTM.

PR #61215, opened 2025-07-30
Ingress-nginx-controller: Adjustments to better match upstream configs and changes

AmberArcadia:
Better aligned our setup with upstream build script. Enabled missing features/components like njs, some patches, fixed iamguarded chart version, etc.


kwmonroe (approved):
@AmberArcadia really nice job, thanks!!

PR #61218, opened 2025-07-30
Bump tensorflow-core with Axion fixes

egibs:
This PR bumps `tensorflow-core` and ensures it builds and imports correctly given the new Axion instance types and cleans up old config that also contributes to test failures related to the new instances.

While these instances support BTI and PAC, Tensorflow only supports PAC and will cause SIGILL crashes as soon as `import tensorflow` is executed in Python if `-mbranch-protection=standard` is used for the build.


tensorflow-core.yaml
```diff
@@ -52,7 +54,6 @@ environment:
       - perl
       - py3-astunparse
       - py3-pip
-      - py3-pybind11-dev
```
egibs, line 55: This was pointing to an old/unmaintained version prior to the `-dev` merge of the package.

tensorflow-core.yaml
```diff
@@ -89,7 +90,7 @@ environment:
     TF_DOWNLOAD_CLANG: 0
     TF_NEED_CUDA: 0
     TF_CUDA_CLANG: 0
-    CLANG_CUDA_COMPILER_PATH: /usr/bin/clang
+    CLANG_COMPILER_PATH: /usr/bin/clang
```
egibs, line 93: I don't think this changes much but since we don't do anything with CUDA it's technically more correct.

tensorflow-core.yaml
```diff
@@ -156,6 +161,7 @@ subpackages:
         - py${{range.key}}-opt-einsum
         - py${{range.key}}-pip
         - py${{range.key}}-protobuf
+        - py${{range.key}}-pybind11
```
egibs, line 164: TBD if this is required, but the SIGILL errors pointed to `pybind11` so it seemed good to include at any rate.

tensorflow-core.yaml
```diff
@@ -226,28 +236,5 @@ update:
     tag-filter: v
 
 test:
-  environment:
-    contents:
-      packages:
-        - py3-tensorflow-core
```
egibs, line 232: This will crash on the new instance types and isn't built any longer (plus the commands below are handled in the subpackages now).

murraybd (approved):

PR #61283, opened 2025-07-31
openssf-compiler-options: Consolidate x86 GCC flags

sergiodj:
We're using different flags between different versions of GCC on x86. For example, GCC 14 and 15 are missing `-fcf-protection=full`.  Let's consolidate things.

### DIFF
--- a/openssf-compiler-options.yaml
+++ b/openssf-compiler-options.yaml
@@ -1,7 +1,7 @@
 package:
   name: openssf-compiler-options
   version: 20240627
-  epoch: 26
+  epoch: 27
   description: "Compiler Options Hardening Guide for C and C++"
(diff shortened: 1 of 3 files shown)


sergiodj:
@xnox I'm wondering if there's a reason for this discrepancy.  I couldn't find anything on `git log`, but maybe I'm missing something.

xnox (changes requested):
They are implied by fhardened and do automatic detection of glibc as to high they can go, and allow easy opt outs by specifying an alternative.

Doing them expanded, removes ability to partially opt out, and may break 2.28 builds.

that's why fhardened is preferred over the expanded versions.

And older gcc did not have that shortcut.

sergiodj:
Ah, gotcha.  I knew I was missing something!  Thanks for the quick reply.  Let's close this one, then.

PR #61296, opened 2025-07-31
pre-commit: use shellcheck

dannf:
Add a pre-commit check that extracts the code from each `runs:` snippet and runs it through `shellcheck`, configured to only report errors. There are still some false positives, but not a terrible percentage.

There are PRs open to fix all existing failures.


tulilirockz:
> [IS THIS WHAT YOU WANT?](https://github.com/wolfi-dev/os/pull/61296/commits/fef6577b7dfacd0b56835c2daee238ec0cce020a)

It wasnt it 💔

sergiodj (approved):

PR #61313, opened 2025-07-31
feat: add logging-operator-node-exporter and logging-operator-custom-runner

wojciechka:
#### For new package PRs only

### DIFF
--- a/kube-logging-operator-node-exporter.yaml
+++ b/kube-logging-operator-node-exporter.yaml
@@ -1,0 +1,65 @@
+package:
+  name: kube-logging-operator-node-exporter
+  version: "6.0.1"
+  epoch: 0
+  description: Logging operator for Kubernetes - node exporter helper files
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - busybox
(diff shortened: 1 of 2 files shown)


kube-logging-operator-node-exporter.yaml
```diff
@@ -0,0 +1,65 @@
...
+update:
+  enabled: true
+  github:
+    identifier: kube-logging/logging-operator
+
+test:
+  pipeline:
+    - runs: |
+        mkdir -p /buffers /prometheus/node_exporter/textfile_collector
+        /usr/bin/logging-operator-node-exporter-buffer-size.sh </dev/null >/dev/null 2>/dev/null &
```
Dentrax, line 57: Should we have a test daemon pipeline here instead?

wojciechka: This is a simple script, it does not print any output, only processes files in a loop.

I do not think it makes sense. It may not even work because it does not output anything and the test has expected output that is required.

I'd rather leave it as is.

Dentrax (approved):

PR #61571, opened 2025-08-04
add sasl-xoauth2 package

sanjayk0508:
related to: https://github.com/chainguard-dev/image-requests/issues/5962

### DIFF
--- a/sasl-xoauth2.yaml
+++ b/sasl-xoauth2.yaml
@@ -1,0 +1,147 @@
+package:
+  name: sasl-xoauth2
+  version: "0.25"
+  epoch: 0
+  description: "SASL plugin for XOAUTH2"
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - ca-certificates-bundle
+      - cyrus-sasl
+      - py3-msal
+      - python-3
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - cmake-3
+      - curl-dev
+      - cyrus-sasl-dev
+      - gcc-14-default
+      - jsoncpp-dev
+      - openssl-dev
+      - pkgconf
+      - py3-msal
+      - python-3
+      - python-3-dev
+      - wolfi-base
+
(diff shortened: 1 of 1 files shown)


sasl-xoauth2.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://github.com/tarickb/sasl-xoauth2
+      tag: release-${{package.version}}
+      expected-commit: 637356f5effaeb7336ab6ebaf2fad1fe0e9a6a50
+
+  - runs: |
+      # Disable docs build by commenting out the add_subdirectory(docs) line
```
Dentrax: Why we don't build the docs and provide `-doc` subpkg?

sasl-xoauth2.yaml
```diff
@@ -0,0 +1,117 @@
...
+      opts: |
+        -DCMAKE_INSTALL_PREFIX=/usr \
+        -DCMAKE_INSTALL_LIBDIR=lib \
+        -DCMAKE_INSTALL_SYSCONFDIR=/etc \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
+        -DSASL_PLUGIN_DIR=/usr/lib/sasl2 \
+        -DBUILD_SHARED_LIBS=ON \
+        -DBUILD_DOCS=OFF \
+        -DBUILD_MAN_PAGES=OFF \
```
Dentrax, line 55: Same here

sasl-xoauth2.yaml
sanjayk0508: we actually need `pandoc` dep for doc, which we don't have packaged, and considering `sasl-xoauth2` is dep itself for postfix and no standalone image for it, so skipped it. But let me know if you think we should include it anyways

Dentrax: Ah thanks for clarification, in this case can we update the comment here, it should represent the _why_.

Dentrax (approved):

PR #61579, opened 2025-08-04
correct release/ build for arm too

anushkamittal20:


font-stix-otf.yaml
```diff
@@ -1,44 +1,31 @@
...
-      source venv/bin/activate
-      pip3 install -r requirements.txt
-      python tools/tirobuild.py STIXbuild.yml
       mkdir -p "${{targets.contextdir}}/usr/share/fonts/stix-fonts/"
-      mv ./output/STIXbuild/STIXTwoMath/OTF/* "${{targets.contextdir}}/usr/share/fonts/stix-fonts/"
-      mv ./output/STIXbuild/STIXTwoText/OTF/* "${{targets.contextdir}}/usr/share/fonts/stix-fonts/"
+      mv ./fonts/static_otf/ "${{targets.contextdir}}/usr/share/fonts/stix-fonts/"
 
 update:
   enabled: false
```
Dentrax: Can we set the `enabled: true` and pass `schedule:`?

anushkamittal20: what kind of schedule do youthink we should add here 🤔

anushkamittal20: tried a different approach here!

font-stix-otf-arm.yaml
```diff
@@ -0,0 +1,37 @@
...
+  target-architecture:
+    - aarch64
+  dependencies:
+    provides:
+      - font-stix-otf=${{package.full-version}}
+
+environment:
+  contents:
+    packages:
+      - build-base
```
kranurag7:
```suggestion

```
nit: `build-base` is required only when we're compiling something, in this case only busybox will be good.

font-stix-otf-arm.yaml
```diff
@@ -0,0 +1,36 @@
...
+      uri: https://github.com/stipub/stixfonts/archive/refs/tags/v2.13b171.tar.gz
+      expected-sha256: 1e76b9ab0bb08372ff73ad5b58d9116260e9058d1fce4b83fe1e213c3b9c947f
+
+  - runs: |
+      mkdir -p "${{targets.contextdir}}/usr/share/fonts/stix-fonts/"
+      mv ./fonts/static_otf/* "${{targets.contextdir}}/usr/share/fonts/stix-fonts/"
+
+update:
+  enabled: false
+  exclude-reason: 2.14 tag was never officially released, we will need to look out for the next official build to fetch binary from.
```
kranurag7: non-blocker: can we also update the comment here? thinking...

jamonation (approved):

PR #61671, opened 2025-08-05
adds powerdns (pdns), its backends and tools, and pdns-recursor

Dentrax:
Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

```
<!--ci-cve-scan:fail-any-->
```

#### For version bump PRs

#### For PRs that add patches


pdns-5.2.yaml
```diff
@@ -0,0 +1,392 @@
...
+        --enable-remotebackend-zeromq \
+        --with-libcrypto=/usr \
+        --enable-ixfrdist \
+        --enable-reproducible
+
+  - uses: autoconf/make
+
+  - uses: autoconf/make-install
+
+  - runs: PDNS_TEST_NO_IPV6=1 make check
```
jamonation: Why no IPv6 check?

pdns-5.2.yaml
```diff
@@ -0,0 +1,392 @@
...
+
+  - name: Install files
+    runs: |
+      mkdir -p ${{targets.destdir}}/etc/init.d/
+      install -Dm755 pdns.initd ${{targets.destdir}}/etc/init.d/pdns
+      install -Dm755 pdns.conf ${{targets.destdir}}/etc/pdns/pdns.conf
```
jamonation: Where do these come from? What are they intended to support?

pdns-5.2/pdns.initd
jamonation, line 1: Why is this file included? It looks like it was removed 2+ years ago? https://github.com/PowerDNS/pdns/pull/13345

pdns-5.2/pdns.conf
jamonation, line 1: Where is this file coming from and where is it used? It appears that the .yml-template file is generated from here: https://github.com/PowerDNS/pdns/blob/master/pdns/recursordist/rec-rust-lib/table.py

So our package should do that in the build pipeline, and then use the generated config file as opposed to shipping a static version that may drift over time.

jamonation (changes requested):
I don't think the pdns-recursor.initd, pdns.conf, pdns.initd, and recursor.conf files match what I'd expect from adding the pdns package elsewhere. These files should be generated or copied in place from the pdns source tree directly.

pdns-5.2.yaml
Dentrax: I was following the APKBUILD, let me test without that.

pdns-5.2/pdns.conf
Dentrax: I copied those from Alpine's repo, we should remove those in this case right? https://gitlab.alpinelinux.org/alpine/aports/-/tree/master/community/pdns

pdns-5.2.yaml
Dentrax: Dropped a comment, disabled for aarch64

pdns-5.2/pdns.initd
Dentrax: Removed!

pdns-5.2/pdns.conf
Dentrax: Removed!

pdns-5.2.yaml
Dentrax: Removed!

jamonation (dismissed):

pdns-5.2.yaml
```diff
@@ -0,0 +1,420 @@
...
+      provides:
+        - pdns-recursor-compat=${{package.full-version}}
+      runtime:
+        - python-${{vars.py-version}}
+        - py${{vars.py-version}}-jinja2
+        - tini
+        - libcap-utils
+        - coreutils # For env -S support
+        - merged-usrsbin
+        - wolfi-baselayout
```
EyeCantCU: Not needed

pdns-5.2.yaml
```diff
@@ -0,0 +1,420 @@
...
+              mkdir -p ${{targets.contextdir}}/etc/powerdns/templates.d
+              install -Dm755 startup.py ${{targets.contextdir}}/usr/local/sbin/pdns_recursor-startup
+              install -Dm755 recursor.conf ${{targets.contextdir}}/etc/powerdns/recursor.conf
+              ln -sf /usr/bin/pdns_recursor ${{targets.contextdir}}/usr/local/sbin/pdns_recursor
+              ln -sf /usr/bin/rec_control ${{targets.contextdir}}/usr/local/sbin/rec_control
+    test:
+      pipeline:
+        - runs: |
+            test "$(readlink /usr/local/sbin/pdns_recursor)" = "/usr/bin/pdns_recursor"
+            test "$(readlink /usr/local/sbin/rec_control)" = "/usr/bin/rec_control"
```
EyeCantCU: `test` will fail silently without enabling debug, making it unclear where tests fail for anyone that may be debugging them. I'd prefer we used a pattern like this though:

```sh
found="$(readlink -fv /path/to/link)"
expected="/expected/path"
if [ $found != $expected ]; then
  echo "Invalid symlink!"
  echo "Expected: $expected"
  echo "Found: $found"
  exit 1
fi
echo "Valid symlink. Found: $found"
```

pdns-5.2.yaml
```diff
@@ -0,0 +1,420 @@
...
+      provides:
+        - pdns-backend-${{range.key}}=${{package.full-version}}
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.contextdir}}/usr/lib/pdns
+          mv ${{targets.destdir}}/usr/lib/pdns/lib${{range.key}}backend.so ${{targets.contextdir}}/usr/lib/pdns/lib${{range.key}}backend.so
+    test:
+      pipeline:
+        # gmysql backend needs libmariadb.so.3 on the runtime, which we can't provide runtime deps conditionally when using ranges.
+        # Apko would resolve it when building anyway, so we skip the ldd check only for the gmysql backend.
```
EyeCantCU, line 269: Looking into this... If it isn't being resolved by apk, it's not going to be resolved by apko

pdns-5.2.yaml
```diff
@@ -0,0 +1,420 @@
...
+          mv ${{targets.destdir}}/usr/bin/${{range.key}} ${{targets.contextdir}}/usr/bin/${{range.key}}
+    test:
+      pipeline:
+        - uses: test/tw/ldd-check
+        - runs: |
+            # saxfr is a special case, it requires IP-address, port, zone to run and does not have a --help option.
+            # So we handle it separately.
+            if [ "${{range.key}}" != "saxfr" ]; then
+              ${{range.key}} --help
+            else
```
EyeCantCU: Let's use an elif here explicitly. I know we're already promised `saxfr` exclusively, but it wouldn't hurt

pdns-5.2.yaml
```diff
@@ -0,0 +1,420 @@
...
+        - ${{package.name}}-backend-gpgsql
+        - ${{package.name}}-backend-gsqlite3
+        - ${{package.name}}-backend-ldap
+        - ${{package.name}}-backend-lmdb
+        - ${{package.name}}-backend-lua2
+        - ${{package.name}}-backend-pipe
+        - ${{package.name}}-backend-remote
+    pipeline:
+      - runs: mkdir -p ${{targets.contextdir}}
+    description: Virtual package that installs all PowerDNS backends
```
EyeCantCU, line 313: We have tests for meta packages and virtual packages (in the pipelines folder)

pdns-5.2.yaml
```diff
@@ -0,0 +1,420 @@
...
+        - ${{package.name}}-tool-pdns_notify
+        - ${{package.name}}-tool-saxfr
+        - ${{package.name}}-tool-sdig
+        - ${{package.name}}-tool-stubquery
+        - ${{package.name}}-tool-zone2json
+        - ${{package.name}}-tool-zone2ldap
+        - ${{package.name}}-tool-zone2sql
+    pipeline:
+      - runs: mkdir -p ${{targets.contextdir}}
+    description: Virtual package that installs all PowerDNS tools
```
EyeCantCU, line 346: Same here, let's add a test

pdns-5.2.yaml
```diff
@@ -0,0 +1,420 @@
...
+        timeout: 30
+        expected_output: |
+          Listening on controlsocket
+          UDP server bound
+          TCP server bound
+          Done launching threads, ready to distribute questions
+        post: |
+          wait-for-it -t 5 --strict localhost:53 -- echo ":53 is up"
+          dig @127.0.0.1 -p 53 example.com SOA +norecurse | grep -qi "NO ERROR"
+          sdig 127.0.0.1 53 example.com SOA | grep -qi "No Error"
```
EyeCantCU, line 433: `dig` already returns nonzero on error. Adding `grep` here isn't needed

pdns-5.2.yaml
```diff
@@ -0,0 +1,420 @@
...
+            expected_output: |
+              YAML config found and processed
+              Enabled TCP data-ready
+              Listening for queries
+              Launching worker threads
+              Launching tcpworker threads
+              Enabled multiplexer
+            post: |
+              wait-for-it -t 5 --strict localhost:8082 -- echo ":8082 is up"
+              wait-for-it -t 5 --strict localhost:53 -- echo ":53 is up"
```
EyeCantCU: `wait-for-it` is already verbose on failure, the `echo` isn't necessary

pdns-5.2.yaml
```diff
@@ -0,0 +1,420 @@
...
+          pdnsutil add-record example.com ns1 A 127.0.0.1
+        start: "pdns_server"
+        timeout: 30
+        expected_output: |
+          Listening on controlsocket
+          UDP server bound
+          TCP server bound
+          Done launching threads, ready to distribute questions
+        post: |
+          wait-for-it -t 5 --strict localhost:53 -- echo ":53 is up"
```
EyeCantCU: Same here

EyeCantCU (changes requested):

pdns-5.2.yaml
Dentrax: IIRC i hit some cases where dig / sdig didn't return non-zero for other non-(NO ERROR) cases, that grep check provides much confidence for me so I'd rather keep it here, if does it make sense to you as well.

pdns-5.2.yaml
Dentrax: +1 to idea but there is no other test cases in the `compat` so we wouldn't expect it to fail.

But there is an ongoing migration to `test/tw/symlink-check` test pipeline, but it fails in this case due to a known issue, we do follow-up here. wdyt?

pdns-5.2.yaml
Dentrax: Not sure if understand this, there is no other cases actually, its just "saxdr" or "others". 🤔 Could you clarify a bit?

pdns-5.2.yaml
EyeCantCU: I'm sorry, I'm a little lost. If we wouldn't expect the case to fail then there wouldn't be a need to have the test in the first place. I would just like whatever we use in the interim to be verbose. If you run the command above, you'll see output is totally silent regardless of whether it passes or fails so it wouldn't be clear to anyone during the debug process that this is the cause of the failure other than this being the only two tests for this package. Let's disregard whatever the future has in store, we can fix this now

pdns-5.2.yaml
EyeCantCU: The logic works right now but it's backwards. The "saxfr" "or others" isn't true. The test is exclusively for saxfr. The behavior is the opposite of that right now. The gate should be flipped so that either if the plugin is saxfr, run the test for saxfr, otherwise generic --help or if it isn't saxfr, run the generic help test, if it is saxfr, run the test for saxfr (not all other cases that may or may not be saxfr)

pdns-5.2.yaml
EyeCantCU: If we're sure there are cases where it's not returning non-zero, let's keep it, but if there aren't, we should remove it

pdns-5.2.yaml
Dentrax: That makes sense tbh, thanks for the idea. See https://github.com/wolfi-dev/os/pull/61671/commits/15e53b860df5b114fc78a4b66a9424c02e193e09

pdns-5.2.yaml
Dentrax: Ah that makes sense, flipped the logic here: https://github.com/wolfi-dev/os/pull/61671/commits/15e53b860df5b114fc78a4b66a9424c02e193e09

pdns-5.2.yaml
Dentrax: Do you think its a bug with melange or smth?

EyeCantCU: The resolver correctly picks up the package that provides library... I'd imagine we don't maintain the package anymore

EyeCantCU (approved):

PR #61677, opened 2025-08-05
Bump the version of docker-library

max-allan-cgr:
Get the new version of docker-library.
The dockerd-entrypoint,sh includes `-addext  keyUsage=critical,digitalSignature,keyCertSign`
Which Python now requires to be set on CA certificates.


xnox (approved):

xnox:
We should have this cronned.

PR #61709, opened 2025-08-05
feat(test pipeline): Adding a static package test pipeline

debasishbsws:
- static packages that contain only static libraries, which are linked into applications at compile time rather than at runtime.
- this pipeline checks that the package contains only static libraries. and fails if it finds other files or empty package


pipelines/test/staticpackage.yaml
```diff
@@ -0,0 +1,42 @@
...
+        echo "See:"
+        apk info -qL "$static_pkg"
+        echo "This package [$static_pkg] do not contain usable static libraries"
+        echo "Please check the package build for proper static library installation, and either:"
+        echo "  (a) fix the static subpackage build to actually include .a files (check the split/static pipeline), or"
+        echo "  (b) remove the test/staticpackage pipeline"
+        exit 1
+      fi
+
+      echo "Package [$static_pkg] contains only valid static libraries, total static files: $static_files"
```
aborrero: we usually have here a string like `PASS: <msg>`.

Same for errors.

pipelines/test/staticpackage.yaml
```diff
@@ -0,0 +1,42 @@
...
+              exit 1
+              ;;
+          esac
+        fi
+      done
+
+      if [ $static_files -eq 0 ]; then
+        echo "See:"
+        apk info -qL "$static_pkg"
+        echo "This package [$static_pkg] do not contain usable static libraries"
```
aborrero: I've recently learn that we try to have `FAIL` or `PASS` heading strings in this kind of messages produced by test pipelines.

pipelines/test/staticpackage.yaml
```diff
@@ -0,0 +1,40 @@
...
+# this packages only contains `.a` files, no other files are expected. this pipeline checks that the package contains only static libraries.
+pipeline:
+  - name: static package check
+    runs: |
+      # Get our static package name, which is usually a sub package
+      static_pkg=$(basename ${{targets.contextdir}})
+      static_files=0
+
+      # Test static library files
+      for static_file in $(apk info -qL "$static_pkg" | grep -v "^var/lib/db/sbom"); do
```
aborrero: Here is a proposal for a maybe more compact logic:

```
file_list="$(apk info --quiet --contents $static_pkg"| grep -v ^$ | grep -v var/lib/db/sbom)"

if [ "$(echo $file_list | grep .a$ | wc -l)" = 0 ] ; then
    echo "FAIL: package [$static_pkg] does not contain any static library '.a' files."
    exit 1
fi

if [  "$(echo $file_list | grep -v .a$ | wc -l)" != 0 ] ; then
   echo "FAIL: package [$static_pkg] contains unexpected files beyond the static library '.a' files."
   echo $file_list
   exit 1
fi

echo "PASS: ..."
```
This way we avoid the for loop. What do you think?

debasishbsws: I want to have more verbose out put but love the logic, I will try to reduce the line count, to be more compact.

pipelines/test/staticpackage.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+      # Check if package contains any non-.a files
+      if [ $(echo "$file_list" | grep -v "\.a$" | wc -l) -ne 0 ]; then
+        echo "FAIL: Package [$static_pkg] contains unexpected files beyond static library '.a' files:"
+        echo "$file_list" | grep -v "\.a$"
+        echo "      Expected only .a files and .spdx.json metadata"
+        exit 1
+      fi
+
+      echo "PASS: Package [$static_pkg] contains only valid static libraries, total static files: $static_files"
```
aborrero: this `$static_files` var is unset.

debasishbsws: Ahh right it should be `file_list`

smoser: Are you open to putting this into tw? 
You can see a very similar PR to what it would look like https://github.com/chainguard-dev/tw/pull/130

debasishbsws: @smoser I am open to do that.

debasishbsws: @smoser After reviewing tw and thinking it through, I believe this test fits better alongside the existing tests for metapackages, virtual packages, empty packages, and documentation packages.

So I think i would like to have this test pipeline here instead of tw.

smoser: ultimately, i'd like to get those other things to tw also.

tw allows to more easily test and improve pipelines and tools that are used in pipelines.

We can be more consistent on output and pipeline behavior.

We can write tests for the pipelines, and use things like shellcheck, or replace bash with a more suitable programming language with ease.

aborrero: > ultimately, i'd like to get those other things to tw also.
> 
> tw allows to more easily test and improve pipelines and tools that are used in pipelines.
> 
> We can be more consistent on output and pipeline behavior.
> 
> We can write tests for the pipelines, and use things like shellcheck, or replace bash with a more suitable programming language with ease.

I would be happy to prepare a PR for that repo.

I will send you a message via slack to learn a bit more how the tw repo is made available to the other 3 melange pacakge repos.

debasishbsws: if thats the final plan I can start doing that sure.

aborrero (approved):
LGTM.

debasishbsws:
@smoser I am @aborrero are working on some test improvements and as the plan is to move every test is tw. we talked and we plan to merge this now and we can start moving all the package type test from here to tw.

PR #61724, opened 2025-08-05
glu, glew: split glu package into `-dev` variant + clean deps

tulilirockz:
Makes sense as we are shipping pkgconf configs on the main package as well


kranurag7 (approved):
looks good to land, verified that glew is the only thing that's using glu.

Dentrax (approved):

PR #61776, opened 2025-08-06
add: headlamp package

yashpimple:
Related: https://github.com/kubernetes-sigs/headlamp/tree/main

#### For new package PRs only

#### For new version streams

- [X] Any subpackages include the version string in their package name (e.g. `name: ${{package.name}}-compat`)

#### For security-related PRs

```
<!--ci-cve-scan:fail-any-->
```

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/headlamp.yaml
+++ b/headlamp.yaml
@@ -9,7 +9,7 @@ package:
     runtime:
       - busybox
       - gnutar
-      - nodejs-${{vars.node-version}}s
+      - nodejs-${{vars.node-version}}
 
 vars:
   node-version: 18 # Headlamp requires Node.js 18 for frontend builds as of v0.34.0
--- a/headlamp.yaml
+++ b/headlamp.yaml
@@ -46,9 +46,8 @@ pipeline:
       mkdir -p "${{targets.destdir}}/headlamp/plugins"
       mkdir -p "${{targets.destdir}}/headlamp/static-plugins"
       mkdir -p "${{targets.destdir}}/tools"
-      cp container/build-manifest.json "${{targets.destdir}}/tools/"
-      cp container/fetch-plugins.sh "${{targets.destdir}}/tools/"
-      chmod +x "${{targets.destdir}}/tools/fetch-plugins.sh"
+      install -m 0644 container/build-manifest.json "${{targets.destdir}}/tools/"
+      install -m 0755 container/fetch-plugins.sh "${{targets.destdir}}/tools/"
       cd "${{targets.destdir}}/tools"
       ./fetch-plugins.sh "${{targets.destdir}}/headlamp/static-plugins"
 
--- a/headlamp.yaml
+++ b/headlamp.yaml
@@ -7,7 +7,11 @@ package:
     - license: Apache-2.0
   dependencies:
     runtime:
-      - nodejs-18
+      - bash-binsh
+      - nodejs-${{vars.node-version}}
+
+vars:
+  node-version: 18 # Headlamp requires Node.js 18 for frontend builds as of v0.34.0
 
 environment:
   contents:
@@ -16,7 +20,7 @@ environment:
       - curl
       - gnutar
       - jq
-      - nodejs-18
+      - nodejs-${{vars.node-version}}
       - npm
 
 pipeline:
--- a/headlamp.yaml
+++ b/headlamp.yaml
@@ -1,0 +1,92 @@
+package:
+  name: headlamp
+  version: "0.34.0"
+  epoch: 0
+  description: A Kubernetes web UI that is fully-featured, user-friendly and extensible.
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - nodejs
+
+environment:
+  contents:
+    packages:
+      - busybox
+      - curl
+      - gnutar
+      - jq
+      - nodejs
+      - npm
+
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: 4b2d7b4bbbe82371220707ef464af3c7b450e6dc
+      repository: https://github.com/kubernetes-sigs/headlamp
+      tag: v${{package.version}}
+
+  - working-directory: backend
+    runs: go mod download
+
+  - working-directory: backend
+    uses: go/build
+    with:
+      packages: ./cmd
+      output: headlamp-server
+
+  - working-directory: frontend
+    runs: |
+      npm ci --only=prod
+      npm run build
+      mkdir -p "${{targets.destdir}}/headlamp/frontend"
+      cp -r build/* "${{targets.destdir}}/headlamp/frontend/"
+
+  - runs: |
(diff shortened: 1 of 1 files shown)


headlamp.yaml
```diff
@@ -0,0 +1,91 @@
+package:
+  name: headlamp
+  version: "0.34.0"
+  epoch: 0
+  description: A Kubernetes web UI that is fully-featured, user-friendly and extensible.
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - nodejs
```
Dentrax: Should we pin nodejs-18 here? https://github.com/kubernetes-sigs/headlamp/blob/c0a379dbbb78a3d54707bdbba83ea7d57f53f1e7/Dockerfile#L30

headlamp.yaml
```diff
@@ -0,0 +1,91 @@
...
+pipeline:
+  - uses: git-checkout
+    with:
+      expected-commit: 4b2d7b4bbbe82371220707ef464af3c7b450e6dc
+      repository: https://github.com/kubernetes-sigs/headlamp
+      tag: v${{package.version}}
+
+  - working-directory: backend
+    runs: go mod download
+
```
Dentrax: Why we are doing this? Can't we pass a `tidy: true` to go/build pipeline below?

headlamp.yaml
```diff
@@ -0,0 +1,91 @@
...
+          kubectl config view --minify --raw > /tmp/kubeconfig.yaml
+        start: |
+          /headlamp/headlamp-server -html-static-dir=/headlamp/frontend -plugins-dir=/headlamp/plugins -listen-addr=127.0.0.1 -port=8080 -kubeconfig /tmp/kubeconfig.yaml
+        timeout: 30
+        expected_output: |
+          Creating Headlamp handler
+          Listen address: 127.0.0.1:8080
+          Plugins dir: /headlamp/plugins
+        post: |
+          curl -sSf http://127.0.0.1:8080/ | grep -q "<title>Headlamp</title>"
```
Dentrax: Can we wait for port `:8080`?

headlamp.yaml
```diff
@@ -0,0 +1,91 @@
...
+        setup: |
+          kubectl config view --minify --raw > /tmp/kubeconfig.yaml
+        start: |
+          /headlamp/headlamp-server -html-static-dir=/headlamp/frontend -plugins-dir=/headlamp/plugins -listen-addr=127.0.0.1 -port=8080 -kubeconfig /tmp/kubeconfig.yaml
+        timeout: 30
+        expected_output: |
+          Creating Headlamp handler
+          Listen address: 127.0.0.1:8080
+          Plugins dir: /headlamp/plugins
+        post: |
```
Dentrax, line 94: Can we also test `:4466` port? https://github.com/kubernetes-sigs/headlamp/blob/c0a379dbbb78a3d54707bdbba83ea7d57f53f1e7/Dockerfile#L95

yashpimple: You're right, the default image exposes port 4466. I used 8080 in test setup to avoid any local conflicts, but I’ll update it to use 4466 instead so it matches the actual image behavior.

headlamp.yaml
yashpimple: fixed 👍

headlamp.yaml
```diff
@@ -0,0 +1,92 @@
...
+  - runs: |
+      mkdir -p "${{targets.destdir}}/headlamp/plugins"
+      mkdir -p "${{targets.destdir}}/headlamp/static-plugins"
+      mkdir -p "${{targets.destdir}}/tools"
+      install -m 0644 container/build-manifest.json "${{targets.destdir}}/tools/"
+      install -m 0755 container/fetch-plugins.sh "${{targets.destdir}}/tools/"
+      cd "${{targets.destdir}}/tools"
+      ./fetch-plugins.sh "${{targets.destdir}}/headlamp/static-plugins"
+
+subpackages:
```
Dentrax, line 57: Do we need to have plugins subpkg as well? Does upstream image contain this? https://github.com/kubernetes-sigs/headlamp/blob/main/Dockerfile.plugins

headlamp.yaml
```diff
@@ -0,0 +1,92 @@
...
+  name: headlamp
+  version: "0.34.0"
+  epoch: 0
+  description: A Kubernetes web UI that is fully-featured, user-friendly and extensible.
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - nodejs-18
+
```
Dentrax, line 16: Let's have a new variable here:

```
vars:
  node-version: 18 # And explain why we set 18 and when to bump this
```

Dentrax, line 16: https://github.com/kubernetes-sigs/headlamp/blob/main/container/fetch-plugins.sh

We need to include `bash-binsh` to runtime deps. And that script calls `curl`, `tar`, etc.

Dentrax (changes requested):
I compared the [filesystem](https://apk.chaindag.dev/https/apk.cgr.dev/wolfi-presubmit/a2d66148dc94fa3545c8b7f501b10e75282dc340/aarch64/headlamp-0.34.0-r0.apk@etag:34343038376637323061633636633864333764363834333733303437316566372d37/) with upstream image `ghcr.io/headlamp-k8s/headlamp:latest`, and its matching! Dropped some small comments.

headlamp.yaml
yashpimple: Hi @Dentrax in this case, a separate plugin subpackage isn't necessary since we're only bundling the default plugins directly via `[fetch-plugins.sh](https://github.com/kubernetes-sigs/headlamp/blob/main/container/fetch-plugins.sh)` (similar to how upstream does it in [Dockerfile](https://github.com/kubernetes-sigs/headlamp/blob/main/Dockerfile)) where they only include prometheus plugin by default.

Also upstream maintains and distributes each plugin as a [separate image](https://github.com/headlamp-k8s/plugins), so in this [image request](https://github.com/chainguard-dev/image-requests/issues/6303) we decide to ship and build plugins independently, since it make more sense to split them into separate packages for future offerings.

headlamp.yaml
```diff
@@ -0,0 +1,96 @@
...
+  name: headlamp
+  version: "0.34.0"
+  epoch: 0
+  description: A Kubernetes web UI that is fully-featured, user-friendly and extensible.
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - bash-binsh
+      - nodejs-${{vars.node-version}}
```
Dentrax: Do we need curl and tar as well? re: fetch script content

yashpimple: After inspecting the upstream image, we can confirmed that curl and bash are not included at runtime. since we run `fetch-plugins.sh` only during build time to copy static plugin assets, this script are not required in final build. tar is included in upstream, so i have added it in our runtime dependencies.

Dentrax (approved):

PR #61813, opened 2025-08-06
tests for grafana-alloy

Aditevil:
Follow-up to this comment - https://github.com/chainguard-images/images-private/pull/13389#pullrequestreview-3090426648

Without libsystemd the tests will fail -
`Error: /etc/alloy/config.alloy:39:1: Failed to build component: building component: creating journal reader: unable to open a handle to the library`

### DIFF
--- a/grafana-alloy.yaml
+++ b/grafana-alloy.yaml
@@ -1,7 +1,7 @@
 package:
   name: grafana-alloy
   version: "1.10.0"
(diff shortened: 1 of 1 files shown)


kranurag7 (commented):
nice, should we move the dependencies at package level only, thinking...

kranurag7 (approved):

PR #61829, opened 2025-08-06
kayenta 2.46 package build

developer-guy:
Fixes:

Related:

#### For new package PRs only


stevebeattie:
Local review of malcontent scan looks okay.

kayenta-2.46.yaml
```diff
@@ -0,0 +1,78 @@
...
+  epoch: 0
+  description: Automated Canary Service
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - openjdk-17-jre
+    provides:
+      - kayenta=${{package.full-version}}
+
```
Dentrax, line 13: I think we should have a Java version variable here:

```
vars:
  java-version: 17 # And the reason why we pinned 17 and when to update this variable in the future
```

kayenta-2.46.yaml
```diff
@@ -0,0 +1,78 @@
...
+        - redis
+        - curl
+    environment:
+      JAVA_HOME: /usr/lib/jvm/java-17-openjdk
+  pipeline:
+    - uses: test/daemon-check-output
+      with:
+        setup: |
+          redis-server > /tmp/redis.logs 2>&1 &
+          sleep 2
```
Dentrax: Shouldn't this be instead:

```suggestion
        setup: redis-server
```

kayenta-2.46.yaml
```diff
@@ -0,0 +1,78 @@
...
+        start: /opt/kayenta/bin/kayenta
+        timeout: 120
+        expected_output: |
+          Started Main
+          Tomcat started
+        post: |
+          sleep 5 # Allow some time for the service to start
+          # Test health endpoint
```
Dentrax, line 70: Daemon test already awaits the process to spin up.
```suggestion
          wait-for-it -t 5 --strict localhost:8090
```

kayenta-2.46.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+  - uses: auth/gradle
+
+  - uses: patch
+    with:
+      patches: upgrade-dependencies.patch
+
+  - runs: |
+      export GRADLE_OPTS="-Dorg.gradle.daemon=false -Xmx2g -Xms2g"
+      export JAVA_HOME=/usr/lib/jvm/java-17-openjdk
```
Dentrax: nit: I'd move these to `.environment.environment` for vis.

kayenta-2.46.yaml
developer-guy: nope it should be running in the background

kayenta-2.46.yaml
developer-guy: nope, there is some time needed for api to ready, wait-for-it is not working as I expected.

kayenta-2.46.yaml
```diff
@@ -0,0 +1,83 @@
...
+  - uses: patch
+    with:
+      patches: upgrade-dependencies.patch
+
+  - runs: |
+      ./gradlew --info --stacktrace build kayenta-web:installDist \
+          -x :kayenta-core:test \
+          -x :kayenta-integration-tests:test
+
+      mkdir -p ${{targets.contextdir}}/opt/kayenta
```
anushkamittal20, line 45: I am not super sure, but should this be a compat package?

kayenta-2.46.yaml
```diff
@@ -0,0 +1,83 @@
...
+      tag: v${{package.version}}
+
+  - uses: auth/gradle
+
+  - uses: patch
+    with:
+      patches: upgrade-dependencies.patch
+
+  - runs: |
+      ./gradlew --info --stacktrace build kayenta-web:installDist \
```
anushkamittal20, line 41: correct me if I am wrong, maybe we dont need info and stacktrace here? they weould just print stuff for the build right?

developer-guy: I left this option in place because, in the event of a failure, we might need to reopen this, what do you think?

kayenta-2.46.yaml
developer-guy: it could be I guess, lemme check real quick, thanks for the suggestion!

kayenta-2.46.yaml
anushkamittal20: definitely, no harm!

kayenta-2.46.yaml
developer-guy: I noticed that we were copying a lot of other files into /opt, not just the binary. I think it would be better to keep it that way, as it makes managing everything in /opt much simpler just like upstream did, what do you think?

kayenta-2.46.yaml
```diff
@@ -0,0 +1,83 @@
...
+      cp -r kayenta-web/build/install/kayenta/* ${{targets.contextdir}}/opt/kayenta
+
+test:
+  environment:
+    contents:
+      packages:
+        - jq
+        - redis
+        - curl
+        - wait-for-it
```
anushkamittal20: should we remove wait-for-it

developer-guy: yes we don't need that anymore, thanks!

anushkamittal20 (dismissed):
lgtm!

kayenta-2.46.yaml
```diff
@@ -0,0 +1,82 @@
...
+
+test:
+  environment:
+    contents:
+      packages:
+        - jq
+        - redis
+        - curl
+    environment:
+      JAVA_HOME: /usr/lib/jvm/java-17-openjdk
```
Dentrax:
```suggestion
      JAVA_HOME: /usr/lib/jvm/java-${{vars.java-version}}-openjdk
```

kayenta-2.46.yaml
Dentrax: the test daemon pipeline supposed to run it in the background already, what error did you see?

kayenta-2.46.yaml
Dentrax: make sense, lets keep the sleep as-is and add wait-for-it to avoid CI flake.

Dentrax (approved):

PR #61920, opened 2025-08-07
openssl: avoid potentially broken ARMV8_UNROLL12_EOR3

xnox:
Fix potential unaligned access in providers on ARMv9 instances.

Actually not too sure if it is unaligned access, or just in general ARMV8_UNROLL12_EOR3 not working, or not working in some cases but not others.

### DIFF
--- a/openssl.yaml
+++ b/openssl.yaml
@@ -1,8 +1,7 @@
-#nolint:forbidden-repository-used,forbidden-keyring-used
 package:
   name: openssl
   version: "3.5.2"
-  epoch: 0
+  epoch: 1
   description: "the OpenSSL cryptography suite"
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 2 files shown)


openssl.yaml
```diff
@@ -1,8 +1,7 @@
-#nolint:forbidden-repository-used,forbidden-keyring-used
```
justinvreeland, line 1: Is this just cruft that wasn't removed from when it was added?

justinvreeland (commented):
Have you tested it with wget to verify it works yet? I'm happy to try it if you haven't.

openssl.yaml
xnox: cruft that wasn't removed, when we stopped using repositories and keyring in the melange yaml back when we had to urgently ship some stuff.

xnox:
> Have you tested it with wget to verify it works yet? I'm happy to try it if you haven't.

tested and it does work. But I don't know why.

Also not sure if it is safer to disable that always, because it seems like it can "get unlucky" as in rebuilds / contents of the app that is running openssl seems to matter so far.

xnox:
I don't like that i don't understand the issue. Also this is not critical as it for the 3.4 module.

PR #61940, opened 2025-08-07
cluster-api-provider-vsphere package build

developer-guy:
Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/cluster-api-provider-vsphere-1.13.yaml
+++ b/cluster-api-provider-vsphere-1.13.yaml
@@ -1,0 +1,73 @@
+package:
+  name: cluster-api-provider-vsphere-1.13
+  version: "1.13.1"
+  epoch: 0
+  description: Kubernetes-native declarative infrastructure for vSphere.
+  copyright:
+    - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


cluster-api-provider-vsphere-1.13.yaml
```diff
@@ -0,0 +1,60 @@
...
+update:
+  enabled: true
+  github:
+    identifier: kubernetes-sigs/cluster-api-provider-vsphere
+    strip-prefix: v
+
+test:
+  pipeline:
+    - runs: |
+        manager --help 2>&1 | grep "Usage of manager"
```
Dentrax, line 60: Is it possible to do some negative-testing with `test/daemon-check-output` even though it would fail? We can alse check the version as well: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/d125fdaf10da6b1f1219d042887d640a735a8e89/main.go#L243 and `Starting manager` log.

It might need `CAPV` CRDs but we can ignore those here.

wojciechka (approved):

PR #61975, opened 2025-08-08
disable openssf flags for squid to work around invalid evex instruction

jamonation:


xnox:
let's try this:
- https://github.com/wolfi-dev/os/pull/62004

once it builds, does that one look fine / better?

jamonation:
Does look better, tested and `squid` no longer exits with sigkill illegal instruction. So opting for the more conservative approach in #62004

PR #61994, opened 2025-08-08
mesa: withdraw APK's pre-migration to libglvd

javacruft:
24.2.6-r2 removed a number of shared libraries and associated
scaffolding from the mesa package, making use of libglvnd instead.

Due to the version of libglvnd (1.7.0) this creates some odd
resolution behaviour where 24.2.6-r1 get resolved instead of libglvnd
plus associated subpackages.

Withdraw all mesa generated APK's prior to this split.


javacruft:
I've validated this withdrawal using the tools appearing in stereo - build dependencies across all packaging repos still resolve correctly, as do images and vms.

justinvreeland (approved):

PR #62004, opened 2025-08-08
squid: disable -march=native

xnox:
The minimum required abi is encoded in our toolchains, and we must
stick to it for now.

Reported by: @jamonation


jamonation (approved):
LGTM, thanks for looking into it

PR #62081, opened 2025-08-08
Fix `test/compiler/asan` on GCC 14

sergiodj:
I don't know how this test ever passed.  libasan is installed in a non-standard path and can't be found without setting `LD_LIBRARY_PATH`, which leads to an error when executing the test programs.

I came up with this hack to automatically set `LD_LIBRARY_PATH` if a new parameter `version` is provided for the compiler.

### DIFF
--- a/gcc-14.yaml
+++ b/gcc-14.yaml
@@ -1,7 +1,7 @@
 package:
   name: gcc-14
   version: 14.3.0
-  epoch: 5
+  epoch: 6
   description: "the GNU compiler collection - version 14"
   copyright:
     - license: GPL-3.0-or-later WITH GCC-exception-3.1
--- a/pipelines/test/compiler/asan.yaml
+++ b/pipelines/test/compiler/asan.yaml
@@ -27,7 +27,14 @@ pipeline:
         return 0;
       }
       EOF
-      ${{inputs.cc}} noleak.c -o noleak -fsanitize=address -lasan
+      ver=$(echo "${{inputs.cc}}" | sed -n 's@[a-z0-9]\+-\([0-9]\+\)@\1@p')
+      cc="${{inputs.cc}}"
+      rpath=""
+      if [ -n "${ver}" ]; then
+        cc=$(echo "${{inputs.cc}}" | cut -d'-' -f1)
+        rpath="-Wl,-rpath,/usr/lib/${cc}/${{host.triplet.gnu}}/${ver}/"
+      fi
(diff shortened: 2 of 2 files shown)


xnox:
it might have silently used the gcc 15's asan instead which is global?

can we use rpath on the binary?

sergiodj:
I thought about that, but then why is it not using/installing GCC 15's ASAN anymore in the test environment?

sergiodj:
> can we use rpath on the binary?

I was torn between using `rpath` or `LD_LIBRARY_PATH`.  I can use the former if you'd prefer.

sergiodj:
Changed to `rpath`.

gcc-14.yaml
```diff
@@ -355,7 +355,8 @@ test:
         cc: gcc-14
     - uses: test/compiler/asan
       with:
-        cc: gcc-14
+        cc: gcc
```
xnox: version is not used together with cc, so should this not instead still say gcc-14?

```suggestion
        cc: gcc-14
```

sergiodj: The problem is that if `cc` is `gcc-14`, then the `rpath` will be invalid.  Argh.

Let me adjust `${{inputs.cc}}` accordingly.

pipelines/test/compiler/asan.yaml
```diff
@@ -5,6 +5,10 @@ inputs:
     description: |
       CC compiler to use
     required: true
+  version:
```
dannf: I think long-term we need a solution that just works regardless of compiler version, or there's really no good way to consume this feature in a generic way. In which case, this version parameter is only needed temporarily, and we'll need to claw it back at some point. To avoid that - perhaps we should just autodetect the compiler version in the test code for now and rpath as needed (or always?), and revisit when we've a better general solution for libasan.

xnox: as far as I understand, matching libasan is always required. Thus yes, parsing $cc --version; and doing rpath off that is likely the best way forward forever.

sergiodj: Thanks for the review.

What I can do is detect whether the compiler name has an explicit version in it.  We only need to `rpath` in this case; if we're dealing with the default compiler, its libraries are installed in `/usr/lib` instead of `/usr/lib/gcc/ARCH/VERSION`.

Let me adjust the PR to do that.

sergiodj:
Updated.

@dannf is this what you had in mind?

dannf (approved):

PR #62518, opened 2025-08-10
add missing runtime dependency for headlamp-plugin-flux

yashpimple:
Related: https://github.com/headlamp-k8s/plugins

Added `busybox` to runtime dependencies to provide `/bin/sh` and core utilities (mkdir, cp, chown) needed by the `initContainer` script that copies plugin files into Headlamp’s pluginsDir. Removed nodejs from runtime since the plugin ships as static assets and does not require node.js at execution time. This aligns the image behavior with upstream and prevents init step failures.

### DIFF
--- a/headlamp-plugin-flux.yaml
+++ b/headlamp-plugin-flux.yaml
@@ -1,7 +1,7 @@
 package:
   name: headlamp-plugin-flux
   version: "0.4.0"
-  epoch: 0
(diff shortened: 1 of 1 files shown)


kranurag7 (approved):
thank you, validated from upstream image and diff looks good; upstream explicitly is using busybox utilities over coreutils.

PR #62562, opened 2025-08-11
5583 spilo wal g hotfix

utieyin:
Wal-g is needed by spilo and adding a runtime dependency on python here was causing issues as it was fixing the python version at 3.11. Have now moved the python version into the test environment so it is no longer a hard dependency. Have tested with spilo and all works fine.

Added new dir to plpython postgres as this was missing headers


AnmolVirdi (approved):
LGTM. These changes do not break any other image/package requirements. Only meant for compatibility against Spilo package.

PR #62566, opened 2025-08-11
Tetragon package addition, bundled with Tetra CLI and Tetragon-operator

AnmolVirdi:
**Tetragon package** `Current version: v1.5.0`
- eBPF-based Security Observability and Runtime Enforcement

**Type**: Go binary
**REF**: https://github.com/cilium/tetragon

**Tests:**
- `Tetragon`:  Versioning, completion script, Daemon test, metrics validations
- `TetraCLI`: Versioning, completion script, core commands, loglevel
- `Tetragon-operator`: Daemon Test, CRDs check, Healthchecks

### DIFF
--- a/tetragon.yaml
+++ b/tetragon.yaml
@@ -78,6 +78,13 @@ subpackages:
           cp -r /etc/profile.d/bash_completion.sh "${{targets.subpkgdir}}/etc/bash/bash_completion.sh"
           cp -r /etc/bash_completion.d/000_bash_completion_compat.bash "${{targets.subpkgdir}}/etc/bash_completion.d/000_bash_completion_compat.bash"
           cp -r /usr/share/bash-completion/bash_completion "${{targets.subpkgdir}}/usr/share/bash-completion/bash_completion"
+    test:
+      pipeline:
+        - name: "Check for scripts"
+          runs: |
+            test -s /etc/bash/bash_completion.sh
+            test -s /etc/bash_completion.d/000_bash_completion_compat.bash
+            test -s /usr/share/bash-completion/bash_completion
 
   - name: tetra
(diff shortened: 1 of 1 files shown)


tetragon.yaml
```diff
@@ -0,0 +1,221 @@
...
+  name: tetragon
+  version: 1.5.0
+  epoch: 0
+  description: "eBPF-based Security Observability and Runtime Enforcement"
+  copyright:
+    - license: Apache-2.0
+  dependencies:
+    runtime:
+      - bpftool
+      - ca-certificates-bundle
```
rahulduvedi, line 11: we can safely omit this as this is added by the pipeline by default:
https://github.com/chainguard-dev/melange/blob/main/pkg/build/pipelines/go/build.yaml#L7

AnmolVirdi: It's runtime dependency, not build time @rahulduvedi

rahulduvedi: can you remove that from buildtime deps?

tetragon.yaml
```diff
@@ -0,0 +1,221 @@
...
+      - gops
+      - iproute2
+
+environment:
+  environment:
+    LOCAL_CLANG: "1"
+  contents:
+    packages:
+      - bash-completion
+      - ca-certificates-bundle
```
rahulduvedi: can we remove it from here, These are buildtime deps.

AnmolVirdi: Updated, thanks.

rahulduvedi:
can we do same changes request here in the other package
just some cosmetic changes, Thank you
Great tests!

rahulduvedi (approved):

PR #63050, opened 2025-08-13
Speed up pre-commit checks

egibs:
The current pre-commit Action doesn't reliably cache its configuration, so we spend cycles initializing the environment each time. Also, we can avoid running against all files unless the pre-commit configuration changes which is about 4-5 times faster per commit.


xnox (approved):

xnox:
And now i guess we need to run this twice to see that it all works, right?

egibs:
> And now i guess we need to run this twice to see that it all works, right?

~I think rebasing any PR will work since the cache key will already exist.~

Edit: yea, we'll need to recreate the cache in `main` and then future runs should be able to retrieve the key

PR #63156, opened 2025-08-14
geos-3.13: No-change rebuild for 2.28 toolchain

cjwatson:
Needed for some Python libraries (e.g. rasterio).


xnox (approved):

xnox (changes requested):
hm. checking stuff.

Built on x86_64, failed on arm64.

xnox (approved):

PR #63197, opened 2025-08-14
ruby*-net-http-persistent: convert to git update backend

OddBloke:
Upstream stopped using GitHub Releases (before the version we currently have, confusingly).


OddBloke:
I'll let automation do the bumping, so disregard the PUCC failure.

AmberArcadia (approved):

PR #63198, opened 2025-08-14
py3-decorator: convert to git update backend

OddBloke:
Upstream haven't used GitHub Releases since before the version we currently have.


OddBloke:
I'll let automation handle the bump, please disregard the PUCC failure.

AmberArcadia (approved):

PR #63208, opened 2025-08-14
Fix more OpenJDK FTBFSes with glibc 2.42

sergiodj:


xnox (approved):

xnox:
sort of want to remove all of these jdks - we don't care about them actually and we don't need them.

sergiodj:
> sort of want to remove all of these jdks - we don't care about them actually and we don't need them.

That'd be great.  I've been meaning to ask whether it makes sense to keep them around...

PR #63223, opened 2025-08-14
RFC: python-3.13: Add support for dist-packages

dannf:
Add support for a dist-packages directory that sits next to site-packages. Libraries installed as apks can live here, keeping them separate from the pip-installed libraries that will continue to install into site-packages by default.

This is an experiment based on [PEP-668](https://peps.python.org/pep-0668). The hope was that we could somehow "protect" the deps used by the system from stacks installed by tools like pip. But it turns out we don't really get much benefit from it, for reasons described here: https://peps.python.org/pep-0668/#create-separate-distro-and-local-directories.

TLDR; the "EXTERNALLY-MANAGED" setting is per python-install, not per directory. So while we can separate where thing get installed, this is not a mechanism to prevent them from interfering with one another. They are in the same namepace, an pip will just as happily remove an apk-installed package from under dist-packages as it would remove a pip-installed package from under site-packages. We *could* set the EXTERNALLY-MANAGED flag, but that would just result in users having to set a flag to override this.

The only benefit I see is with potential conflicts between filepaths between apk's containing individual libraries and apk's containing bundled pip-installations. But even then, we'd just be allowing an install of libraries that may actually conflict at runtime.

### DIFF
--- a/pipelines/py/pip-build-install-bootstrap.yaml
+++ b/pipelines/py/pip-build-install-bootstrap.yaml
@@ -20,6 +20,7 @@ pipeline:
   - name: "pip-bootstrap build"
     runs: |
       export SOURCE_DATE_EPOCH=315532800
+      export CHAINGUARD_PYTHON_INSTALL_LAYOUT="chainguard_system"
       py=${{inputs.python}}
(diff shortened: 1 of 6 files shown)


cmwilson21:
Thanks for the contribution!

PR #63245, opened 2025-08-15
chore(crossplane-provider-aws): Build individual resource packages instead of monolith provider

EyeCantCU:
Credit to @Aditvil-Dev for the initial implementation. Reverting a prior revert to restore this

Changes:

Providers are published in the crossplane-contrib org on GHCR so we can avoid xpkg.io altogether. This enables us to avoid regenerating the same CRDs manually and avoid extraction of the package CR by referencing the package by its version on the marketplace, which is inconsistent with how the providers are tagged on GitHub

Additionally, retain the package names for the Crossplane AWS family provider so that it's just drop in

Also, move from crane to crank. Previously, we were extracting a tarball of the entire image to get ahold of the package CR. Using crank, we only extract the package CR from the layer that contains it, saving a significant amount of time (at least locally)


mritunjaysharma394 (dismissed):
looks good, thanks for the detailed note!

mritunjaysharma394 (approved):

PR #58274, opened 2025-07-03
Add new package: nftables-slim/1.1.3

yxun:
Ad new package: nftables-slim/1.1.3

This PR is related to a discussion in https://kubernetes.slack.com/archives/C03MASRQP4M/p1747869530558059
I updated the existing nftables configure options in a new yaml file and removed some unused built-time dependencies.
The goal of this new package is to provide a slim package so that users can use nftables via a control process and no interactive user usage.

Thanks,

#### For new package PRs only

- [ ] This PR is marked as fixing a pre-existing package request bug
  - [ ] Alternatively, the PR is marked as related to a pre-existing package request bug, such as a dependency
- [ ] REQUIRED - The package is available under an OSI-approved or FSF-approved license
- [ ] REQUIRED - The version of the package is still receiving security updates
- [ ] This PR links to the upstream project's support policy (e.g. `endoflife.date`)

#### For version bump PRs

- [ ] The `epoch` field is reset to 0

PR #63313, opened 2025-08-15
nftables: add a nftables-slim package

kranurag7:
nftables-slim package was added to wolfi via
https://github.com/wolfi-dev/os/pull/58274

this commit moves the same to a subpackage and this
way we'll make one single package and updates will
also be more aligned.


kranurag7:
\\ cc @yxun fyi; everything is same but we're moving this as a sub-package to existing nftables.yaml for better maintenance.

stevebeattie (approved):
Thanks for doing this @kranurag7 ! Looks great and confirmed doesn't change the nftables-slim package as compared to the results of the version bump in https://github.com/wolfi-dev/os/pull/63308/

yxun:
@kranurag7 I see the build environment is different between the nftables.yaml and nftables-slim.yaml file

I need to test this new sub package installation and let you know if the unused dependencies are still a problem in our base image building later.

The main concern is :

> The initial concern was the dependencies when installing the existing nftables.yaml package. It pulls many other dependencies and additional unused binary files.
That causes an Istio side wolfi based image from 10+MB to 30+MB when installing the existing nftables.yaml package.
We only need that nft package which is expected around 10+MB.

with the nftables-slim.yaml environment , I can verify the concern was be addressed. I will try the new nftables.yaml sub package installation and reply later.

nftables-slim.yaml
```
package:
  name: nftables-slim
  version: "1.1.4"
  epoch: 0
  description: Netfilter tables userspace tools without interactive CLI
  copyright:
    - license: GPL-2.0-or-later

environment:
  contents:
    packages:
      - autoconf
      - automake
      - bison
      - build-base
      - busybox
      - flex
      - libmnl-dev
      - libnftnl-dev
```

nftables.yaml
```
  dependencies:
    runtime:
      - merged-usrsbin
      - wolfi-baselayout
environment:
  contents:
    packages:
      - asciidoc
      - autoconf
      - automake
      - bison
      - build-base
      - busybox
      - ca-certificates-bundle
      - flex
      - gmp-dev
      - jansson-dev
      - libedit-dev
      - libmnl-dev
      - libnftnl-dev
      - ncurses-dev
      - py3-gpep517
      - py3-setuptools
      - py3-wheel
      - readline-dev
```

yxun:
I tested this . All good, thanks again.

PR #63410, opened 2025-08-16
openssl: add MD5 and SHA1 to the base provider

xnox:
This is for compatibility with upstream cpython and dotnet fetching of
digests with -fips property, all of which alternatively document that MD5 is deprecated for security purposes. See examples of this at:

- https://github.com/dotnet/runtime/blob/58d1c2e3e9bc76a4a9e02af75eeba210800f54eb/src/native/libs/System.Security.Cryptography.Native/pal_evp.c#L320

- https://github.com/python/cpython/blob/3663b2ad54c9e15775a605facf69da8f5ee8d335/Modules/_hashopenssl.c#L676

- https://github.com/aws/s2n-tls/blob/83dcc99c93a17fcb8f4aeffaf543d962fa672cf7/crypto/s2n_hash.c#L60

Documentation on the property query strings is here:
- https://docs.openssl.org/3.5/man7/property/#queries

Other distributions instead allow to load the default provider which has high memory cost, and exposes the whole lot of unapproved algorithms to the applications, none of which are desired for pure backwards compatible CRC functionality.
To achieve similar behaviour we currently further patch python to load the default provider, with this change we will be able to drop those patches.

### DIFF
--- a/openssl.yaml
+++ b/openssl.yaml
@@ -2,7 +2,7 @@
 package:
   name: openssl
   version: "3.5.2"
-  epoch: 0
+  epoch: 1
   description: "the OpenSSL cryptography suite"
   copyright:
     - license: Apache-2.0
@@ -45,7 +45,7 @@ pipeline:
 
   - uses: patch
     with:
-      patches: fix-jitter.patch
+      patches: fix-jitter.patch 0001-baseprovider-add-MD5-and-SHA1.patch
 
   - name: Create dbg sourcecode
     runs: |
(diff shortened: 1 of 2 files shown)


sil2100 (approved):
Feels weird to enable deprecated algorithms in the base provider. Normally I'd be worried that this would enable use of those algorithms for our FIPS provider, but from what I see when `fips=yes` is set, it simply won't allow it, even if it's coming from the base provider, right? So this feels safe.

xnox:
> Feels weird to enable deprecated algorithms in the base provider. Normally I'd be worried that this would enable use of those algorithms for our FIPS provider, but from what I see when `fips=yes` is set, it simply won't allow it, even if it's coming from the base provider, right? So this feels safe.

Correct!

In the openssl.cnf for FIPS we have
```
[algorithm_sect]
default_properties = fips=yes
```
Which means anything that doesn't have that by default is not used. And we do have many algorithms without fips=yes in fips provider (for example Ed25519 in fips provider 3.1.2). And we also have many things in base provider with fips=yes - all the encoding/decoding routines which are not cryptography but just data marshaling/unmarshaling (der to pem, pem to der). So indeed it is safe, and we already have things without fips=yes in our stock configuration, as well as things in the base provider that are sometimes used. Thus we already know that everything is using these look ups correctly.

PR #63457, opened 2025-08-18
6135 ceph package

utieyin:
Updating the ceph package to allow `cephadm` recognise wolfiOS as an operating system


ceph.yaml
```diff
@@ -134,7 +135,7 @@ pipeline:
 
   - uses: patch
     with:
-      patches: fix-osd-class-dir.patch
+      patches: fix-osd-class-dir.patch wolfi-update.patch update-podman.patch
```
mritunjaysharma394: is it possible to add test in package level to ensure the intended functionality of this patch works?

ceph.yaml
```diff
@@ -11,6 +11,7 @@ package:
   dependencies:
```
mritunjaysharma394, line 11: We also need to bump epoch

ceph.yaml
utieyin: Test added for wolfi patch

ceph.yaml
utieyin: Epoch bumped

flexiondotorg (changes requested):
EDIT: This recommendation is incorrect, as discussed in Slack. Sorry for the confusion :bow: 

~~The patch is looking good, but there is a package name variation between Alpine and Wolfi for the packages `cephadm` tries to install via the `install_packages()` function.~~

~~## Package Name Mappings~~

Package Purpose | Alpine Name | Wolfi Name |
-- | -- | -- |
CA Certificates | ca-certificates | ca-certificates-bundle |
Python 3 | python3 | python-3.12 |
Podman | podman | podman |
Time Sync | chrony | chrony |
LVM Tools | lvm2 | lvm2 |~~

~~Here's an example of how the Apk class could be update to support Alpine and Wolfi:~~

```python
class Apk(Packager):
    DISTRO_NAMES = ['wolfi', 'alpine']
    
    # Package name mappings for Wolfi
    WOLFI_PACKAGE_MAP = {
        'ca-certificates': 'ca-certificates-bundle',
        'python3': 'python-3.12',
        # Add other mappings as needed
    }
    
    def __init__(
        self,
        ctx: CephadmContext,
        stable: Optional[str],
        version: Optional[str],
        branch: Optional[str],
        commit: Optional[str],
        distro: Optional[str],
        distro_version: Optional[str],
    ) -> None:
        super(Apk, self).__init__(
            ctx, stable=stable, version=version, branch=branch, commit=commit
        )
        assert distro is not None
        self.ctx = ctx
        self.tool = 'apk'
        self.distro = distro
        self.distro_version = distro_version
    
    def _map_package_name(self, pkg: str) -> str:
        """Map package names for Wolfi if needed."""
        if self.distro == 'wolfi':
            return self.WOLFI_PACKAGE_MAP.get(pkg, pkg)
        return pkg
    
    def add_repo(self) -> None:
        # For Wolfi/Alpine, we don't add custom Ceph repos as packages
        # should be provided through the standard repositories
        logger.info('Skipping repo addition for %s - using system packages' % self.distro)
    
    def rm_repo(self) -> None:
        # No custom repos to remove
        pass
    
    def install(self, ls: List[str]) -> None:
        # Map package names for Wolfi
        mapped_packages = [self._map_package_name(pkg) for pkg in ls]
        logger.info('Installing packages %s...' % mapped_packages)
        call_throws(self.ctx, [self.tool, 'add'] + mapped_packages)
    
    def install_podman(self) -> None:
        logger.info('Attempting podman install...')
        try:
            self.install(['podman'])
        except Error:
            logger.info('Podman did not work. Falling back to docker...')
            self.install(['docker'])
```

~~As Wolfi already includes these dependencies in the ceph package's runtime dependencies, these packages should be installed. However, adding the package name mapping ensures `cephadm` will work correctly if it needs to install any missing packages.~~

flexiondotorg (dismissed):

flexiondotorg (approved):

PR #63550, opened 2025-08-18
protobuf.yaml: update to 31.1 without v3.x update filter

AbduAzim424:
Update `protobuf` to upstream version `31.1` and remove `v3.x` update filtering tag. Re-build all compatible dependent packages.

References: 
   - Related issue: https://github.com/chainguard-dev/internal-dev/issues/12473
   - `UPB` upstream dependency issue: https://github.com/protocolbuffers/protobuf/issues/22954


protobuf.yaml
```diff
@@ -37,14 +29,15 @@ pipeline:
...
-      expected-commit: f5de0a0495faa63b4186fc767324f8b9a7bf4fc4
+      expected-commit: 74211c0dfc2777318ab53c2cd2c317a2ef9012de
       cherry-picks: |
         main/ced605d0e6a7ad20985375b596b2ca6720e07737: utf8_range: add version marker to library
 
   - runs: |
       cd third_party
       git submodule update --init --recursive
 
+  # LIBUPB=ON dep -> https://github.com/protocolbuffers/protobuf/issues/22954
```
justinvreeland: Don't need the comment since It's clear from the PR. though I appreciate it for the review.

protobuf.yaml
```diff
@@ -37,14 +29,15 @@ pipeline:
     with:
       repository: https://github.com/protocolbuffers/protobuf
       tag: v${{package.version}}
-      expected-commit: f5de0a0495faa63b4186fc767324f8b9a7bf4fc4
+      expected-commit: 74211c0dfc2777318ab53c2cd2c317a2ef9012de
       cherry-picks: |
         main/ced605d0e6a7ad20985375b596b2ca6720e07737: utf8_range: add version marker to library
 
   - runs: |
```
justinvreeland, line 36: We should get rid of this too there's no reason we'd vendor dependencies for this package. Might not fit with this set of work but I'm reasonably certain we'll have issues with this as part of the rebuild anyway.

justinvreeland:
Also be sure to add `-Dprotobuf_LOCAL_DEPENDENCIES_ONLY=ON` so that we don't pull in bad deps.

justinvreeland:
This looks good, not sure what's going on with the onnx either the logs aren't super helpful from what I'm seeing. Looks like there's some additional packages in the so bump warning as well. Hopefully this doesn't get tooo large : (

justinvreeland (approved):
This looks good to me as far as I can tell the missing `so:` check failures are packages that have issue or sufficiently deep in the dependency graph that it seems to mess with the elastic build solver. We can pick those up in new PRs once this lands.

PR #63612, opened 2025-08-19
fix(llvm-20): correct cmake packaging for dev subpackages

daemoninstitute:
The `libmlir-20-dev` subpackage was incorrectly moving the entire `/usr/lib/llvm-20/lib/cmake` directory. This resulted in essential files like `LLVMConfig.cmake` being removed from the main `llvm-20-dev` package, breaking its usability.

This commit refines the packaging logic to only move the mlir-specific subdirectory. The `epoch` has also been bumped to 2.

```
<!--ci-cve-scan:fail-any-->
```

### DIFF
--- a/llvm-20.yaml
+++ b/llvm-20.yaml
@@ -1,7 +1,7 @@
 package:
   name: llvm-20
   version: "20.1.8"
-  epoch: 1
+  epoch: 2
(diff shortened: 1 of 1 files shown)


dannf (approved):
thanks!

PR #63683, opened 2025-08-19
add py3 versions to xfsprogs

sanghanan:
Adding multiple py3 versions to xfs_protofile to fix dependency issues with cephcsi.

Context:
`cephcsi` has two runtime dependencies:
* `ceph` → requires Python 3.11 specifically
* `xfsprogs-extra` → built against `python3` which currently resolves to Python 3.13.

This leads to a conflict when building the cephcsi image:
```
solving "ceph" constraint: resolving "ceph-19.2.3-r3.apk" deps:
solving "python-3.11" constraint:
  python-3.11-3.11.1-r5.apk disqualified because python-3.13-3.13.6-r0.apk already provides python3
```

To fix this, I extracted `xfs-protofile` from `xfsprogs-extra`, because it is the only component in that package that actually requires Python. I bundled `xfs-protofile` separately for multiple Python versions. This means that:
- `xfsprogs-extra` itself no longer depends on a specific Python version.
- I can choose the required  `xfs-protofile` dependency for `cephcsi` (e.g., `xfs-protofile-py-3.11`) without causing conflicts with `ceph`’s Python 3.11 requirement.
-  By default, `xfs-protofile` still resolves to Python 3.13, so other packages that depend on it are unaffected.

### DIFF
--- a/xfsprogs.yaml
+++ b/xfsprogs.yaml
@@ -11,6 +11,14 @@ package:
       - wolfi-baselayout
       - xfsprogs-core
 
+data:
+  - name: py-versions
+    items:
+      3.10: "310"
+      3.11: "311"
+      3.12: "312"
+      3.13: "313"
+
 environment:
   contents:
     packages:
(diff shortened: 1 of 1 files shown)


jamonation (commented):
For consistency, can the new package be called `xfsprogs-protofile-...`?

sanghanan:
> For consistency, can the new package be called `xfsprogs-protofile-...`?

@jamonation done

jamonation (approved):

PR #63700, opened 2025-08-20
harbor-2.13: unpin swagger

dannf:
We've had `swagger` pinned for over a year now. Unfortunately, I couldn't find a record of what the exact problem was so we just seem to have been cargo-culting it.

The package builds/tests fine today with the pin removed, and our `harbor` images test/build fine against it.

### DIFF
--- a/harbor-2.13.yaml
+++ b/harbor-2.13.yaml
@@ -1,7 +1,7 @@
 package:
   name: harbor-2.13
   version: "2.13.2"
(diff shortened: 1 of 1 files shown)


dannf:
Image verification https://github.com/chainguard-images/images-private/pull/15077 (apologies for the internal link)

EyeCantCU (approved):

PR #63749, opened 2025-08-20
tw - sync pipelines and melange yaml with tw at v0.0.25

smoser:
octo-sts did a merge at #63636, but it did not / does not sync pipelines or melange yaml.

This syncs those back in.


smoser:
Closing this for now.
The failure is due to https://github.com/chainguard-dev/tw/issues/169.

PR #63756, opened 2025-08-20
fix: add coreutils for init container compatibility

rahulduvedi:
Flannel Helm chart's init containers require POSIX utilities (cp, mv, etc.) to copy CNI binaries and configuration files to the host filesystem. Without these utilities, pods fail with `exec: cp: executable file not found in $PATH`
check here https://github.com/flannel-io/flannel/blob/v0.27.2/chart/kube-flannel/templates/daemonset.yaml

### DIFF
--- a/flannel.yaml
+++ b/flannel.yaml
@@ -7,8 +7,8 @@ package:
     - license: Apache-2.0
   dependencies:
     runtime:
-      - coreutil
       - ca-certificates
+      - coreutil
       - ip6tables
       - iproute2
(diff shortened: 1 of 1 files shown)


kranurag7 (approved):
looks good (ref: https://github.com/flannel-io/flannel/blob/1bf0b6de86034efb0ad313e091716a829abdfe01/chart/kube-flannel/templates/daemonset.yaml#L39-L63)

PR #63757, opened 2025-08-20
sqlite: Enable TCL extension

IdlePhysicist:
In this commit I enable the compilation of the TCL extension for SQLite.

I had to change to a different source archive as the TCL extension
library was not included in the autoconf tarball.


jamonation (approved):
LGTM, thanks for getting the TCL extension working

PR #63817, opened 2025-08-21
apk-tools: future proof CRC algorithms access

xnox:
In some cases apk-tools needs access to MD5 and SHA1 for CRC
purposes. Future proof this access by preffering fips algorithms, but
not requiring them. This results in FIPS usage for RSA256 signature
validation, and yet allows to parse all existing SHA1 manifest CRC
digests as well even when SHA1 is removed.

### DIFF
--- a/apk-tools.yaml
+++ b/apk-tools.yaml
@@ -1,7 +1,7 @@
 package:
   name: apk-tools
   version: "2.14.10"
-  epoch: 7
+  epoch: 8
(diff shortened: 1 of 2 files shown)


sil2100 (approved):
Ok, Dimitri explained to me what the `?` in `?fips=yes` means, and I was able to find OpenSSL docs using `?provider=fips` in the same way. If you tested that this indeed works by preferring the FIPS provider as expected, let's merge it!

PR #63875, opened 2025-08-21
busybox/cve remediation

kbsteere:
fix: Created patch for CVE-2024-58251


kbsteere:
Per Valery Ushakov. My fix does specifically address the CVE but he is also looking at fixing some other related issues. I have submitted another patch which address both the CVE and the other issues. Waiting for response back. 

```
Valery Ushakov response:
This might, pedantically speaking, tick off the CVE as reported, but
the very same process will trigger the exact same problem with ps(1)
&c that sanitizes C0 controls, but doesn't sanitize C1 controls (and
doesn't sanitize the {comm} part at all, btw, even for C0).  I
recently posted about that, but got no feedback:

https://lists.busybox.net/pipermail/busybox/2025-August/091682.html
https://lists.busybox.net/pipermail/busybox/2025-August/091683.html
```

eslerm (approved):
Thanks @kbsteere! Appreciate that you raised this upstream. Two of these patches are not being applied, which is intentional and ack'ing the applied patch. Cheers!

kbsteere:
https://lists.busybox.net/pipermail/busybox/2025-August/091706.html

PR #64039, opened 2025-08-22
hardening-check: ignore .git versions

OddBloke:
These appear to be due to an additional source artifact upstream, rather than indicating a different release.


OddBloke:
Please ignore the PUCC failure, I'll let automation perform the bump.

AmberArcadia (approved):

PR #64314, opened 2025-08-26
Liquibase jar fixes

AmberArcadia:
Fixes: Liquibase was missing a variety of expected components, especially commons-collections4.jar, causing runtime issues. These components are now validated to be present using lists from upstream, and additional checks have been added to ensure we're building and using the source code correctly. 

### Pre-review Checklist

```
<!--ci-cve-scan:fail-any-->
```

### DIFF
--- a/liquibase.yaml
+++ b/liquibase.yaml
@@ -1,7 +1,7 @@
 package:
   name: liquibase
   version: "4.33.0"
-  epoch: 2
+  epoch: 3
   description: "Liquibase is a database schema change management solution that enables you to revise and release database changes faster and safer from development to production."
   copyright:
     - license: Apache-2.0
--- a/liquibase.yaml
+++ b/liquibase.yaml
@@ -120,41 +120,40 @@ pipeline:
       tar -xzf liquibase-dist/target/liquibase-${{package.version}}.tar.gz -C /tmp/liquibase-check
       cd /tmp/liquibase-check
 
-      # Check 1: WOLFI_MARKER in liquibase-core.jar matches exactly
-      echo "==> Checking liquibase-core.jar for markers"
-      jar -xf internal/lib/liquibase-core.jar WOLFI_MARKER.txt 2>/dev/null || true
-      if [ -f WOLFI_MARKER.txt ]; then
-        ACTUAL_CORE_MARKER=$(cat WOLFI_MARKER.txt)
-        echo "Expected core marker: $EXPECTED_CORE_MARKER"
-        echo "Actual core marker:   $ACTUAL_CORE_MARKER"
-        if [ "$ACTUAL_CORE_MARKER" = "$EXPECTED_CORE_MARKER" ]; then
-          echo "✓ WOLFI_MARKER matches exactly - using our build"
-        else
-          echo "✗ WOLFI_MARKER doesn't match - not using our build!"
-          exit 1
-        fi
-      else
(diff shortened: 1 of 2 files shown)


liquibase.yaml
```diff
@@ -60,6 +85,94 @@ pipeline:
       tar -zxvf liquibase-dist/target/${{package.name}}-${{package.version}}.tar.gz -C ${{targets.destdir}}/usr/share/java/liquibase
       ln -sf /usr/lib/jvm/default-jvm ${{targets.destdir}}/opt/java/openjdk
 
+  - name: Check verification markers
```
OddBloke, line 88: I think this is a good test for the problem: I wonder if it should be an actual `test:` instead of being part of the build, so we know that we're validating the files that end up in the package?

liquibase.yaml
```diff
@@ -60,6 +85,94 @@ pipeline:
...
+      echo "==> Validating distribution contents against expected files"
+      # List tarball contents, excluding directories (which end with /)
+      tar -tzf liquibase-dist/target/${{package.name}}-${{package.version}}.tar.gz | grep -v '/$' | sort > actual-contents.txt
+
+      # Check against the standard expected contents
+      echo "Comparing with expected-distribution-contents.txt"
+      # Filter out commercial components from expected files since we build with !liquibase-commercial
+      grep -v 'liquibase-commercial.jar' liquibase-dist/expected-distribution-contents.txt | sort > expected-filtered.txt
+
+      if diff -u expected-filtered.txt actual-contents.txt; then
```
OddBloke, line 103: This `diff`'s output will be written to the log, so I don't know if we need the `comm` calls below.

liquibase.yaml
```diff
@@ -60,6 +85,94 @@ pipeline:
...
+        if [ "$ACTUAL_STANDARD_MARKER" = "$EXPECTED_STANDARD_MARKER" ]; then
+          echo "✓ WOLFI_STANDARD matches exactly - using our build"
+        else
+          echo "✗ WOLFI_STANDARD doesn't match - not using our build!"
+          exit 1
+        fi
+      else
+        echo "✗✗✗ FAILURE: WOLFI_STANDARD.txt not found in liquibase-core.jar"
+        exit 1
+      fi
```
OddBloke: These two checks could potentially be condensed into either a for loop or a helper function, to reduce repetition.

AmberArcadia: Ah yeah good point, that makes sense.

liquibase.yaml
AmberArcadia: I thought about that too, but it uses info pulled from the source code so it'd be a bit more difficult to do. Plus it's more of an assembly check vs a runtime functionality check.

liquibase.yaml
AmberArcadia: I liked the cleaned up / formatted output, but it is technically redundant.

liquibase.yaml
OddBloke: I guess my concern is that we could be putting a different built artifact into the final package, meaning these tests aren't exercising the actual contents of the package.

Would asserting against the contents of `melange-out/...` be viable?

OddBloke (approved):
LGTM, thanks!

liquibase.yaml
AmberArcadia: I thought about that too, but since it was extracting files from jars and creating a few other files, it felt cleaner to just keep it all separate since both copies are created from the same tarball anyway.

PR #64407, opened 2025-08-27
fix: add symlink to cilium image to match upstream

astrojerms:
Fix https://github.com/chainguard-dev/customer-issues/issues/2417 with symlink compat package.


tulilirockz (commented):
Doesn't this need a bump?

MickeyPvX (approved):

PR #64420, opened 2025-08-27
cmake,cmake-bootstrap: bump to version 4.1.1

OddBloke:


OddBloke:
I've checked with the OS team and as this is only a patch release we can handle it ourselves.

powersj (approved):

PR #64502, opened 2025-08-28
Make cpio and gnutar co-installable.

smoser:
cpio and tar both ship /usr/lib/rmt (remote magnetic tape).

It probably would have been sufficient to just remove /usr/lib/rmt from cpio and no one would notice - I do not think too many people are using remote tape drives at this point, but might as well keep it working.

### DIFF
--- a/cpio.yaml
+++ b/cpio.yaml
@@ -1,10 +1,16 @@
 package:
   name: cpio
   version: "2.15"
-  epoch: 4
+  epoch: 5
   description: tool to copy files into or out of a cpio or tar archive
   copyright:
     - license: GPL-3.0-or-later
+  dependencies:
+    runtime:
+      - gnutar-rmt
+
+vars:
+  rmt-path: /usr/libexec/rmt
 
 environment:
   contents:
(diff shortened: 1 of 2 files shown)


cpio.yaml
```diff
@@ -54,4 +63,8 @@ test:
       runs: cpio --version | grep ${{package.version}}
     - runs: |
         find /usr/local -print -depth | cpio -ov > foo.cpio
+    - runs: |
+        p=${{vars.rmt-path}}
+        [ -x "$p" ] && [ -f "$p" ] && echo "PASS: $p is executable file" ||
+          echo "FAIL: $p is not executable file"
```
sergiodj: This will print a failure message but won't actually return failure.  Is it intentional?  Should it be something like:

```suggestion
        p=${{vars.rmt-path}}
        [ -x "$p" ] && [ -f "$p" ] && echo "PASS: $p is executable file" ||
          (echo "FAIL: $p is not executable file"; exit 1)
```

?

smoser: very good catch, @sergiodj thank you.

sergiodj (approved):

PR #64596, opened 2025-08-29
fix(kubernetes-dns-node-cache): track tags instead of releases

bentasker:
Upstream use an unusual approach to tracking releases: Github releases are only created for new minor versions, with patch versions only being tracked by tags.

Switch to tracking tags to ensure that we pick up on new versions


bentasker:
The PUCC failure is expected: before this change, automation would only have seen the `.0` release

aborrero (approved):
LGTM.

PR #64667, opened 2025-08-29
sudo: encode upstream patch level into our versioning

OddBloke:
This makes it clear which upstream artifact we've built from, and should enable automation to correctly identify new versions (allowing us to enable automated updates).


sergiodj (approved):
Been meaning to tackle this for a while.  I like the `p0` trick.  Thanks!

PR #64796, opened 2025-08-31
Revert "wget: Unpin gcc-14-default"

xnox:
This reverts commit 0d64b87ea7f9d68314e49525e325a51493a4fba4.

Still broken for fips. Please see:
- https://github.com/chainguard-dev/internal-dev/issues/17777


sil2100 (approved):
Let's let @sergiodj know to look into this later!

PR #64817, opened 2025-09-01
update postfix to use rspamd

sanjayk0508:
We're replacing `opendkim` with `rspamd`, as [opendkim](https://github.com/trusteddomainproject/OpenDKIM) is no longer actively maintained, `rspamd` supports both DKIM signing and verification, more context can be found here - https://github.com/chainguard-dev/image-requests/issues/5962

the scripts files used for replacing rspamd: https://github.com/bokysan/docker-postfix/tree/master/scripts

### DIFF
--- a/postfix.yaml
+++ b/postfix.yaml
@@ -291,41 +291,6 @@ pipeline:
       EOF
 
 subpackages:
-  - name: ${{package.name}}-compat
-    description: "Compatibility package for postfix"
-    checks:
-      disabled:
-        - usrmerge
-    pipeline:
-      - runs: |
-          mkdir -p "${{targets.contextdir}}"/usr/sbin
-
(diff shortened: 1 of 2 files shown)


postfix.yaml
```diff
@@ -1,31 +1,35 @@
...
+      - busybox
+      - cronie
       - merged-usrsbin
       - rspamd
+      - rsyslog
       - sasl-xoauth2
       - supervisor
       - wolfi-baselayout
 
+# Dual-source package: scripts from github (docker-postfix), source code from official Postfix tarball
```
Dentrax, line 24: I'm not confident on this tbh, it feels like these should be separate packages instead. Its a bit unusual approach and would make hard to maintain. But happy to discuss more about on this.

sanjayk0508: Dm'd to discuss this

PR #64824, opened 2025-09-01
py3-supported-numpy: transition to version stream

aborrero:
The package 'py3-supported-numpy' is not provided by any numpy version.

Use a version-streamed variant instead.


aborrero:
outdated PR, closing.

PR #64835, opened 2025-09-01
Newer pcre version for njs

vivian-rook:
getting `undefined symbol: pcre_free` in nginx-s3-gateway with older version

https://github.com/chainguard-dev/image-release-stats/issues/6720


aborrero (approved):
LGTM.

PR #64851, opened 2025-09-01
pixi: clean up package + add compat subpackage

tulilirockz:


pixi.yaml
```diff
@@ -42,8 +31,44 @@ update:
     identifier: prefix-dev/pixi
     strip-prefix: v
 
+subpackages:
+  - name: ${{package.name}}-compat
+    description: Compatibility subpackage for Pixi upstream image
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.subpkgdir}}/usr/local/bin"
+          ln -s ../../bin/pixi "${{targets.subpkgdir}}/usr/local/bin/pixi"
```
Dentrax, line 40: Just checking... Are we sure on this?

```
usr/local/bin/pixi -> ../../bin/pixi
```

tulilirockz: It should be pointing to `/usr/bin/pixi`, yup!

tulilirockz: If you want to I can set it to `/usr/bin` directly instead of using a relative symlink here

Dentrax: It should work but thinking... what if i'm in a dir like `/foo/bar/baz/qux/quux` and run the `/usr/local/bin/pixi` here?

tulilirockz: Oh it works fine, relative symlinks are interpreted from where the symlink is at instead of where your $PWD is. Even if you are on `/foo/bar/baz/quix/quux` its going to interpret it as "/usr/bin/pixi" instead of `../../bin/pixi`

EyeCantCU (approved):

PR #64937, opened 2025-09-02
glibc: introduce additional virtualpackage tests

aborrero:

### DIFF
--- a/glibc.yaml
+++ b/glibc.yaml
@@ -2,7 +2,7 @@ package:
   name: glibc
   version: "2.42"
   # Every glibc update causes build disruption; always announce on #eng-psa
-  epoch: 0
+  epoch: 1
   description: "the GNU C library"
   copyright:
     - license: LGPL-2.1-or-later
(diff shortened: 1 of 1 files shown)


AmberArcadia (approved):

sergiodj (changes requested):
Thanks @aborrero .

I'm wondering if it makes sense to have this test in this particular scenario.  The reason is that `apk list installed` won't list any `so:` provides, even those that are actually installed:

```
# apk list installed so:libc.so.6
#
```

So in a way, this test is not really checking anything that could really fail.

WDYT?

aborrero:
> Thanks @aborrero .
> 
> I'm wondering if it makes sense to have this test in this particular scenario. The reason is that `apk list installed` won't list any `so:` provides, even those that are actually installed:
> 
> ```
> # apk list installed so:libc.so.6
> #
> ```
> 
> So in a way, this test is not really checking anything that could really fail.
> 
> WDYT?

I guess this is up to you. The truth is this PR was submitted by an automated script, I only noticed the 'peculiar' of this package later.

sergiodj:
ACK, thanks @aborrero (and thanks a lot for working on this!).  For the reasons I outlined, I am going to close this PR without merging it.  We can revisit this later if needed.

Cheers.

PR #64963, opened 2025-09-02
remove packages

joshrwolf:
removes all the subpackages. unless someone tells me otherwise, I'm intentionally not bumping epoch's because rebuilding doesn't change the desired behavior here.

a withdraw will come in a follow up PR


kranurag7 (approved):

kranurag7:
Looks good thinking all corner cases; if something is using these compats by mistake then we'll very likely find it in world builds of packages and they'll surface via image honks too.

imjasonh (approved):

joshrwolf:
> Looks good thinking all corner cases; if something is using these compats by mistake then we'll very likely find it in world builds of packages and they'll surface via image honks too.

yep! going to be monitoring things after all these pieces land, I'm sure we've missed some implicit deps

james-mchugh:
Out of curiosity, why were these packages removed? Our team took advantage of these. Was it a maintenance burden? Or was it related to the Bitnami license changes? Or something else?

PR #64987, opened 2025-09-02
restore: add APK restore

javacruft:
Use the new restore endpoint to provide the capability to restore
APK's that where previously withdrawn from APK repositories.

This is triggered by any updates to restored-packages.txt, which is
simply a list of apk names to restore; restoration is applied to both
architectures that Wolfi supports.


sil2100 (approved):
This looks good. Checked that the magic `jq` call properly converts the format to a json file. All else seems in par with withdrawal.

PR #65010, opened 2025-09-02
py3-protobuf: disable aarch64 ASM optimization

aborrero:
It results in illegal instruction in our aarch64 CPUs.

### DIFF
--- a/py3-protobuf.yaml
+++ b/py3-protobuf.yaml
@@ -1,7 +1,7 @@
 package:
   name: py3-protobuf
   version: "6.32.0"
-  epoch: 0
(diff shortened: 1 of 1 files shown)


sergiodj (approved):
LGTM.  This also seems like a good upstream bug to file.  Our ARM64 architectures are pretty conventional and should be able to run this code.

aborrero:
> LGTM. This also seems like a good upstream bug to file. Our ARM64 architectures are pretty conventional and should be able to run this code.

Done: https://github.com/protocolbuffers/protobuf/issues/23306

PR #65254, opened 2025-09-04
git - add git-subtree to installation.

smoser:


git.yaml
```diff
@@ -194,6 +202,9 @@ test:
   environment:
     environment:
       HOME: /tmp
+    contents:
```
justinvreeland: man is needed for subtree to work?

justinvreeland (dismissed):
Just a question but this is fine.

git.yaml
smoser: I pushed a commit that does build the docs and `--help` will work.

justinvreeland (approved):

PR #65327, opened 2025-09-04
2509 docker selenium

utieyin:
Update docker selenium to fix permissions and clean up files not used


docker-selenium.yaml
```diff
@@ -134,16 +132,22 @@ subpackages:
...
+              mkdir -p /var/run/supervisor
+              chown seluser:seluser /var/run/supervisor
+            start: |
+              su seluser -c /opt/bin/entry_point.sh
             timeout: 60
             expected_output: |
               supervisord started
 
-  - name: ${{package.name}}-config
+  - name: ${{package.name}}-compat
```
mritunjaysharma394, line 146: We seem to be changing a subpackage name? Have you verified it is not being used anywhere else?

utieyin: It is not used anywhere else

mritunjaysharma394 (approved):

PR #65386, opened 2025-09-05
feat(dotnet-8/9): Make co-installable

EyeCantCU:
.NET SDKs and runtimes work with the latest version of the .NET host so always use the latest version. This previously prevented us from co-installing different versions of .NET as they strictly depended on the version of the host built alongside the current .NET SDK and runtime. This does not prevent installing older .NET host versions if desired (though maybe we should figure out how to not allow that without adding a replaces for every host package and prior .NET version...)

This also splits the .NET Standard targeting pack into separate subpackages. Both 8 and 9 install 2.1.0 causing conflicts when attempting to install both versions. Because the targeting pack is the same between both versions, we can safely split it away and install the targeting pack provided by the latest .NET Standard package. If the major.minor version ever changes, this will require the version be bumped, but I've added a check when building the subpackage that will pick up on that

### DIFF
--- a/dotnet-8.yaml
+++ b/dotnet-8.yaml
@@ -14,6 +14,13 @@ package:
     provides:
       - dotnet=${{package.full-version}}
 
+vars:
+  # This is the major.minor version of the .NET Standard
+  # targeting pack. We validate this version is correct
+  # below, but as the version progresses, this will need
+  # to be manually bumped.
+  netstandard-version: '2.1'
+
 var-transforms:
   - from: ${{package.version}}
     match: ^(\d+)\.\d+\.\d+$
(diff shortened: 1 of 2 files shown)


Dentrax:
Thanks for the work @EyeCantCU, I've tried this locally but for some reason can't able to co-install, am i missing something?

```
d3b341c52c64:/work/packages# apk add aarch64/dotnet-8-8.0.119-r2.apk
ERROR: aarch64/dotnet-8-8.0.119-r2.apk: UNTRUSTED signature
d3b341c52c64:/work/packages# apk add aarch64/dotnet-8-8.0.119-r2.apk --allow-untrusted
(1/5) Installing icu77-data-full (77.1-r1)
(2/5) Installing libstdc++ (15.2.0-r2)
(3/5) Installing libicu77 (77.1-r1)
(4/5) Installing icu (77.1-r1)
(5/5) Installing dotnet-8 (8.0.119-r2)
Executing busybox-1.37.0-r50.trigger
OK: 57 MiB in 20 packages
d3b341c52c64:/work/packages# apk add aarch64/dotnet-9-9.0.109-r2.apk --allow-untrusted
  dotnet-8-8.0.119-r2:
    conflicts: dotnet-9-9.0.109-r2[dotnet=8.0.119-r2]
    satisfies: world[dotnet-8><Q1P+PPkQivsypqDpknf8B6e57zsFQ=]
  dotnet-9-9.0.109-r2:
    conflicts: dotnet-8-8.0.119-r2[dotnet=9.0.109-r2]
    satisfies: world[dotnet-9><Q1LXEtKXXUexvHfzentbQJRrmDE+I=]
```

---

Another point would be, our confidence would much higher if we could add a test to `dotnet-8` to install `dotnet-9` and a test to `dotnet-9` to install `dotnet-8` to ensure that those are co-installable along with all components. WDYT?

EyeCantCU:
@Dentrax you're attempting to install multiple versions of the .NET host and not the runtime or SDK. You can't install both hosts concurrently and that's intentional as the latest host will recognize prior versions of the runtime and SDK

justinvreeland (approved):
👍🏻

PR #65422, opened 2025-09-05
Kubernetes latest is now 1.34

zmarano:
Should fix
```
panic: error creating image.tar: failed to build layer image: installing apk packages:
installing packages: installing kubectl-latest (ver:0-r10 arch:aarch64):
unable to install files for pkg kubectl-latest: writing header for "usr/bin/kubectl":
packages map[kubectl-1.34-default:kubernetes-1.34 kubectl-latest:kubernetes-latest] has conflicting file: "usr/bin/kubectl"
```

Due to https://github.com/wolfi-dev/os/pull/64708

### DIFF
--- a/kubernetes-latest.yaml
+++ b/kubernetes-latest.yaml
@@ -1,7 +1,7 @@
 package:
   name: kubernetes-latest
   version: 0
-  epoch: 10
+  epoch: 11
   description: "Compatibility infrastructure for Kubernetes components"
   copyright:
     - license: GPL-2.0-or-later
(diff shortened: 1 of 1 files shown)


kranurag7 (approved):
Ohh yeah!! thanks Zach, sorry missed this.

xnox:
I am pondering if we can redo kubernetes packaging to match how we handle all other versioned stream things that allow to be co installed and to be default.

I.e. python, GCC.

They all have one melange yaml per version, allow to coinstall them, and allow to pick which one is default, and "latest" is done via depending on a provides.

PR #65554, opened 2025-09-08
openjdk-{8,11}: improve backwards compatibility

javacruft:
Sometime in the history of these packages we published -base or
-headless versions of the jre subpackage; we've seen a few customers
using these package names from our extensive APK history, and as such
they would not be getting updates.

Provide provides for these old package names to ensure that updates to
the latest version of the package occur for these customers when they
rebuild their images.

### DIFF
--- a/openjdk-11.yaml
+++ b/openjdk-11.yaml
@@ -143,7 +143,7 @@ subpackages:
         - libxtst
         - ttf-dejavu
       provides:
-        - ${{package.name}}-jre-headless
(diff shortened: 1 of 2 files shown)


javacruft:
FWIW I think we should do this at the point of subpackage removal, and then drop the provides sometime later but its hard for consumers to know they are doing the wrong thing!

justinvreeland (approved):

PR #65568, opened 2025-09-08
gcc: configure glibc 2.42

sschepens:
We're now using GLIBC 2.42, we should configure gcc accordingly

### DIFF
--- a/gcc.yaml
+++ b/gcc.yaml
@@ -1,7 +1,7 @@
 package:
   name: gcc
   version: "15.2.0"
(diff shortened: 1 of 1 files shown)


sergiodj (approved):

sschepens:
@sergiodj is this supposed to automatically merge?

sergiodj:
> @sergiodj is this supposed to automatically merge?

Yes.  Somehow it hasn't started building yet, let me try adding another label.

PR #65686, opened 2025-09-09
docker: include /usr/lib/docker plugins directory with dockerd

a-crate:
Some packages expect this to exist. I am not sure what the best subpackage for this is, but dockerd seems appropriate, I assume that's what is using the plugins.

### DIFF
--- a/docker.yaml
+++ b/docker.yaml
@@ -1,7 +1,7 @@
 package:
   name: docker
   version: "28.4.0"
-  epoch: 0 # CVE-2025-47907
+  epoch: 1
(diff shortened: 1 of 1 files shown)


roaksoax (approved):
lgtm!

xnox (approved):

a-crate:
Note that there's a typo (I think, anway) in `dockerd` so actually everything in that subpackage is installed to `docker` subpackage. `dockerd` is empty. I'm keeping the status quo and ignoring this for now, and keeping the plugin dir with the binary.

PR #65754, opened 2025-09-10
erlang-26: improve backwards compatibility

javacruft:
We've seen some usage of subpackages that where removed at 26.1-r1 in
the wild; provide backwards compatible provides to cover those seen and
ensure users actually get updates - they are currently pinned to 26.1
which very out of date.


sil2100 (approved):
That's a lot of provides. But I think we should start being more aware of upgrade scenarios, like a normal distro. Especially that even if we keep containers in mind, there's still 'upgrades' like in someone using apk add with a constant package name.

PR #65757, opened 2025-09-10
feat: add flag for rrsync binary

rakshitgondwal:
This PR adds the `--with-rrsync` to the `rsync` package. This flag ships the `rrsync` binary alongside the `rsync` package.


rsync.yaml
```diff
@@ -37,7 +37,8 @@ pipeline:
         --without-included-popt \
         --without-included-zlib \
         --disable-md2man \
-        --disable-openssl
+        --disable-openssl \
+        --with-rrsync
```
mritunjaysharma394, line 41: Thanks for the work!
Is this installing rrysnc? Which I think is a perl script as per related discussion on internet?
https://serverfault.com/a/1161919

How can we verify that we have rrysnc with this change

rakshitgondwal: https://chainguard-dev.slack.com/archives/C0627BW5GLF/p1757503975701889?thread_ts=1757501255.325919&cid=C0627BW5GLF

rakshitgondwal: added the test, ptal, thanks

mritunjaysharma394 (approved):

PR #65806, opened 2025-09-10
gc: withdraw APK's older than 12 months.

javacruft:
As part of ongoing housekeeping for Wolfi, we will be removing APK's
older than 12 months.  Older versions of APK's are assessed based on
usage within the archive, use in images and vms, use as build
dependencies and based on user telemetry reviews.

This is the September 10th Garbage Collection point.


sil2100 (approved):
Checked a few of the removed packages to see if they're sane, then did a diff against the expected withdrawn-packages.txt list and got no difference. That's as much as I can review this. Looks good.

PR #65819, opened 2025-09-10
Enable virtual provides for numpy version streams

AmberArcadia:
In an attempt to improve and solve the numpy versioning situation, I've now enabled virtual packages on the newer numpys, and then we can migrate everything forwards slowly and safely. 

This should not be merged before the other numpy PRs from me today. 

```
<!--ci-cve-scan:fail-any-->
```


aborrero:
ouch, I think I messed up the rebase.

AmberArcadia:
This is now obsolete.

PR #65820, opened 2025-09-10
Numpy version locks 2

AmberArcadia:
In an attempt to improve and solve the numpy versioning situation, I've locked usages of numpy to <2.0 so that we can enable virtual packages on the newer numpys, and then migrate everything forwards slowly and safely. 

```
<!--ci-cve-scan:fail-any-->
```


aborrero (approved):
LGTM.

PR #65850, opened 2025-09-11
rust: cleanup obsolete versions

xnox:
Cleanup obsolete versions of rust, that are over 6 months past EOL.
Eventually binaries will be archied after a year:
- https://github.com/orgs/wolfi-dev/discussions/63023

Also imho it is best to keep just the one rust build in Wolfi,
bootstrapped from a rustup fetched compiler.


sergiodj (approved):
That's great, thank you!

PR #65855, opened 2025-09-11
wolfi-baselayout: correct shadow permissions

xnox:
Note shadow file can be 0000 which many security hardened operating
systems use. In practice it is modified by setuid / priviledged
processes or an external hypervisor, and thus its permissions are
largely irrelevant.

Tested VMs and docker, but possibly this change needs to be tested in
other container implementations too (podman, k8s).

Note fedora containers ship shadow with 0000 permissions by default.

### DIFF
--- a/wolfi-baselayout.yaml
+++ b/wolfi-baselayout.yaml
@@ -1,7 +1,7 @@
 package:
   name: wolfi-baselayout
   version: 20230201
-  epoch: 23
+  epoch: 24
   description: "baselayout data for Wolfi"
   copyright:
     - license: MIT
@@ -68,7 +68,7 @@ pipeline:
         install -m644 vendor/${i} "${{targets.destdir}}"/${i}
       done
 
-      install -m600 vendor/etc/shadow "${{targets.destdir}}"/etc/shadow
+      install -m000 vendor/etc/shadow "${{targets.destdir}}"/etc/shadow
 
       ln -s /etc/crontabs "${{targets.destdir}}"/var/spool/cron/crontabs
       ln -s /proc/mounts "${{targets.destdir}}"/etc/mtab
@@ -87,6 +87,7 @@ test:
       runs: |
         test -f /etc/passwd
         test -f /etc/shadow
+        [ "$(stat -c %a /etc/shadow)" = "0" ]
         test -f /etc/os-release
         test -f /etc/host.conf
         test -d /var/empty


wolfi-baselayout.yaml
```diff
@@ -64,7 +64,7 @@ pipeline:
         install -m644 vendor/${i} "${{targets.destdir}}"/${i}
       done
 
-      install -m600 vendor/etc/shadow "${{targets.destdir}}"/etc/shadow
+      install -m000 vendor/etc/shadow "${{targets.destdir}}"/etc/shadow
```
89luca89, line 71: should we consider initializing also /etc/gshadow and keep it 0000? Seems like it's in use in fedora too, we don't use it at the moment tho

xnox: We could, along with subuid / subgroups too. However, not having the file at all is also compliant. So would rather keep this scoped to this bug fix alone.

I think introducing gshadow is a feature =) and warrants a separate pull request.

89luca89: Perfect, we can do one for gshadow and one for sub{uid,gid} :+1:

89luca89 (dismissed):

Taffer:
Is there some sort of "sane environment" test that could check to make sure it's `000`?

xnox:
> Is there some sort of "sane environment" test that could check to make sure it's `000`?

it is checked elsewhere.

smoser:
If you're going to land this, can I get https://github.com/wolfi-dev/os/pull/62947 added also?

smoser (changes requested):
consider https://github.com/wolfi-dev/os/pull/62947 ?

smoser:
> > Is there some sort of "sane environment" test that could check to make sure it's `000`?
> 
> it is checked elsewhere.

It surely would not hurt to add 

```yaml
   - name: "/etc/shadow perms as expected"
     runs: |
       perms=$(busybox stat -c "%a" /etc/shadow)
       [ "$perms" = "0" ]
```

And would actually excercise that apko picked up the permissions correctly and the test environment extracted them correctly.

smoser:
> consider #62947 ?

this is now landed, and did no bump epoch.  Can you add the test I requested, and then we can land this one.

xnox:
@Taffer @smoser test added

Also rebased onto latest main.

smoser (approved):

PR #65862, opened 2025-09-11
boost: switch to shared library naming policy

javacruft:
Rename libboost-* packages to include soversion so that we can deal with
boost version updates as transitions, improving the general stability of the archive.

Move new shared libraries out of boost and into libboost subpackages.

Move binaries from boost to boost-tools.

These changes ensure co-install-ability between library versions.

### DIFF
--- a/boost.yaml
+++ b/boost.yaml
@@ -226,10 +226,17 @@ subpackages:
     description: "Development files for Boost"
     dependencies:
       runtime:
-        - boost
+        - boost-tools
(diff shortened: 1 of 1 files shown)


justinvreeland (dismissed):

javacruft:
I'll merge tomorrow AM and then deal with the few packages that still need to transition

justinvreeland (approved):

PR #65888, opened 2025-09-11
fix(tests): switch tests from using redis to valkey

dnegreira:
Since we have withdrawn the redis package from our archives, some
packages were still relying on it for some tests.
Valkey is an in-place replacement for redis, as we are currently
building and maintaining it, we will make the switch now.

Signed-off-by: David Negreira <david.negreira@chainguard.dev>


chisel.yaml
```diff
@@ -61,8 +61,8 @@ test:
     contents:
       packages:
         - curl
-        - redis
-        - redis-cli
+        - valkey-8.1
```
sergiodj: I'd use `valkey` here instead of `valkey-8.1`, otherwise we'll need to bump these versions later.

sergiodj (changes requested):

chisel.yaml
dnegreira: Good catch, fixed now, thanks!

xnox (approved):

sergiodj (approved):

murraybd:
I'm not convinced this build is going to finish...

dnegreira:
> I'm not convinced this build is going to finish...

re-kicking it ... if it continues stuck I will separate each package into their own PR

PR #66324, opened 2025-09-16
Provide python version aligned tkinter

justinvreeland:


smoser (commented):
Is there any reason not to just drop the un-versioned provides?

justinvreeland:
> Is there any reason not to just drop the un-versioned provides?

Still used by some packages as a builddep. But we should remove it once I update that package.

smoser (commented):
it is just one package in enterprise-packages... can we just remove it now and fix that one?

smoser (approved):

justinvreeland:
> it is just one package in enterprise-packages... can we just remove it now and fix that one?

I guess we could've but i don't like having the gap where it's not buildable not a real issue though.

PR #66328, opened 2025-09-16
qt6-qtbase: use GCC instead of clang

renner0e:
GCC has less dependencies and is using the OpenSSF security hardening
flags

### DIFF
--- a/qt6-qtbase.yaml
+++ b/qt6-qtbase.yaml
@@ -3,7 +3,7 @@
 package:
   name: qt6-qtbase
   version: "6.9.2"
-  epoch: 0
+  epoch: 1
   description: qt
   copyright:
     - license: LGPL-3.0-or-later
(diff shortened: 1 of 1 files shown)


qt6-qtbase.yaml
```diff
@@ -63,17 +56,9 @@ pipeline:
...
-      export CXXFLAGS="$CXXFLAGS -I/usr/lib/llvm-${{vars.llvm-vers}}/include"
-      export CC=clang
-      export CXX=clang++
-
-      cmake -B build -G Ninja -Wno-dev \
+  - uses: cmake/configure
+    with:
+      opts: |
         -DCMAKE_BUILD_TYPE=Release \
         -DCMAKE_INSTALL_PREFIX=/usr \
```
tulilirockz:
```suggestion
```
I think we can get rid of this. https://github.com/chainguard-dev/melange/blob/8772644a1be7357c644b3003d2a25b7d2fd2988f/pkg/build/pipelines/cmake/configure.yaml#L21-L23

sergiodj:
I like it!  Thanks @renner0e.

sergiodj (approved):

PR #66329, opened 2025-09-16
fix(selenium): adjust paths to align with upstream; confs should be 0644

kwmonroe:
Match upstream dockerfiles wrt installing files in our subpkgs.

- *.conf files were being installed with +x permissions but don't need to be, so this PR also `s/755/644` where appropriate
- make the firefox test accurately test firefox

### DIFF
--- a/docker-selenium.yaml
+++ b/docker-selenium.yaml
@@ -5,7 +5,7 @@ package:
   # 'package format error' when trying to install the package. The workaround is
   # to replace '-' with '.', then mangling the version to replace back.
   version: "4.35.0.20250828"
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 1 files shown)


docker-selenium.yaml
```diff
@@ -54,7 +54,7 @@ environment:
       - x11vnc
       - yq
   environment:
-    TC: UTC
+    TZ: UTC
```
justinvreeland, line 57: What's with the TC -> TZ change.

justinvreeland (approved):

docker-selenium.yaml
kwmonroe: i think it was just a flub.. i can't find mention of a `TC` envar upstream, but `TZ` is pretty prevalent::

https://github.com/SeleniumHQ/docker-selenium/blob/d9f2777614a95bca9b9660911315d5cd96677070/README.md?plain=1#L1103-L1110

PR #66384, opened 2025-09-17
restore: add note re withdrawals

javacruft:
Add a note to ensure that restored packages are not listed in
withdrawn-packages.txt otherwise they might get withdrawn again if files
are added to this file.


murraybd:
Could we add a pre-commit hook for this?

sil2100:
Added a pre-commit PR here: https://github.com/chainguard-dev/pre-commit-hooks/pull/21

sil2100 (approved):
This looks good.

PR #66443, opened 2025-09-17
Switch packages to using ver-check instead of --version

murraybd:
Switch packages to using tw's ver-check.


yace.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: yace
   version: 0.62.1
-  epoch: 7 # CVE-2025-47907
+  epoch: 8 # CVE-2025-47907
```
sergiodj: Small nit, but I prefer to remove the `# CVE...` comment when I bump epochs, otherwise it might give the impression that this bump is fixing the CVE.

sergiodj (dismissed):

yarn-berry.yaml
sergiodj, line 1: Missing epoch bump here.

sergiodj (approved):

PR #66450, opened 2025-09-17
test(varnish-modules): ensure compat with available varnish

kwmonroe:
`varnish-modules` needs to be rebuilt for compat with the recent build of `varnish`.  This isn't the [first time](https://github.com/wolfi-dev/os/pull/35759) we've had to resync the varnish pkg builds.

We just need an epoch bump / rebuild to fix this. Let's also add a test to ensure whatever we build here is compatible with the currently available varnish[-dev] packages; when this test fails, it will be a fast signal in our pkg update queue that we're out of sync.

### DIFF
--- a/varnish-modules.yaml
+++ b/varnish-modules.yaml
@@ -1,7 +1,7 @@
 package:
   name: varnish-modules
   version: "0.26.0"
-  epoch: 1
+  epoch: 2
   description: "Collection of Varnish Cache modules (vmods) by Varnish Software"
   copyright:
     - license: BSD-2-Clause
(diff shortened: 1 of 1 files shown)


kwmonroe:
ack that there's a new pkg update check failure, but we don't want to upgrade to those until the current incompatibility with varnish-7 and varnish-mod-0.26 is resolved.

kbsteere (dismissed):

aborrero (dismissed):
LGTM

aborrero (approved):
LGTM.

PR #66486, opened 2025-09-18
chore(lifecycle): Grant read access to detection job

jml:
Needs this to read the package definitions to make decisions about detection.

See https://github.com/chainguard-dev/internal-dev/issues/20128


joedborg (approved):

jml:
I think this might need to be force merged, but I don't have the permissions needed.

cpanato (approved):
need force merge

PR #66534, opened 2025-09-18
feat: add systemctl as a standalone binary package

89luca89:
It is needed sometime to have systemctl but not the whole systemd package

### DIFF
--- a/systemd.yaml
+++ b/systemd.yaml
@@ -18,6 +18,7 @@ package:
       - merged-usrsbin
       - quota-tools
       - systemd-boot
+      - systemd-systemctl
       - wolfi-baselayout
 
 vars:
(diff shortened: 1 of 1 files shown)


systemd.yaml
```diff
@@ -359,6 +360,26 @@ subpackages:
             kernel-install --version
             kernel-install --help
 
+  - name: "systemd-systemctl"
+    description: "systemd systemctl utility"
+    dependencies:
+      runtime:
+        - libsystemd-shared
```
a-crate, line 367: Are these not picked up by melange SCA? 

also, `merged-*` and wolfi-baselayout is not necessary for new packages.

89luca89: will remove `merged-*`
while for libsystemd-shared, it is not picked up by SCA

xnox (commented):
Note some OSes even ship statistically linked systemctl for greater flexibility. So this is inline with many other ways of dicing systemd.

xnox (approved):

PR #66563, opened 2025-09-19
fix(py3-numpy-1.26.yaml): Bump py3-numpy-1.26.yaml epoch so it is the latest version - and NOT enterprise

philroche:
PR https://github.com/wolfi-dev/os/pull/66533 restored py3-numpy-1.26.yaml but did not bump the epoch so we had conflicting versions in enterprise and os
os should have the latest version


aborrero (approved):
LGTM

PR #66762, opened 2025-09-20
Test building a lot of python.

justinvreeland:
I think this is what's happening:

When we introduced the 2.X version stream for numpy we provided `py3-numpy` the yaml files for these packages were removed but the packages remain in the repo. Because of how we multi version python3 it's possible to resolved deps by a different python version as a dependcy.  For exapmle if you do:

- py3-numpy
- py3-h5py

You might install python3.13-numpy-1.26 and python-3.13-numpy-2.2 to resolve both deps. Nothing ties them together. This is intentionall behavior and is how we can build multiple versions of python.

As we've been removing increasingly older packages some of those older but used packages have not been included anymore. And we've been adding constraints that we didn't before.  This has surface some places like above that the solver doesn't cover. For example we have a py3-numpy=2.2 package but we also we don't do that for 2.3.

While the unversioned python packages are useful for end users who want to install any version of a particular package they're problematic for consistency so we try to specify the version stream to ensure depencies are fixed.

To make this more complicated we can't really fix this forward. Since we don't withdraw packages when we remove the files if we have a package that provides.

The structure of the deps for python should be:

Since we already provide the unversioned python for some packages we can't get rid of it untill we withdraw the packages and remove them. Fortunatly we can mostly fix this. We need to ensure we conflict when the version streamms provide the same thing and we need to provide the unversioned version.

So:

py3.XX-module needs to provide `py3-module` without a version so it can be found with `py3-module` as well as be installed without other python versions of the same module like `py3.XX+1-module`.  When we version stream a python module itself we need to now provide `py3-module` unversioned to be found and potentially resolve as expected.  It also needs to conflict on the other version stream, whilepython version streams are co-installable the modules themselves are not.  So `py3.12-module-new-version-stream` needs to also provide `py3.12-module=new-version-stream` so that we can conflict during package resolution time. This also means anything that has a strict requirement for a particular version stream then it also needs to be version streamed and conflicted. If you build software with a specific version of numpy it can require that minimum version at runtime.

We also need to *explicitly* exclude old versions of the packages that popup to resolve for somethign like `py3-module-verison-stream` but not providing the correct conflicts.

Anywho these take forever to build but i think they're close so i want to see if CI can handle it.

Fixes:

Related:

#### For new package PRs only

#### For new version streams

#### For security-related PRs

```
<!--ci-cve-scan:fail-any-->
```

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/modelmesh-runtime-adapter.yaml
+++ b/modelmesh-runtime-adapter.yaml
@@ -1,22 +1,26 @@
 package:
   name: modelmesh-runtime-adapter
   version: 0.12.0
-  epoch: 16
+  epoch: 17
   description: Unified runtime-adapter package of the sidecar containers which run in the modelmesh pods
   copyright:
     - license: Apache-2.0
   dependencies:
     runtime:
-      - py3-cachetools
-      - py3-google-auth
-      - py3-google-auth-oauthlib
-      - py3-importlib-metadata
-      - py3-numpy
-      - py3-oauthlib
-      - py3-requests-oauthlib
-      - py3-rsa
-      - py3-tensorflow-core
-      - py3-zipp
+      - py3.12-cachetools
+      - py3.12-google-auth
+      - py3.12-google-auth-oauthlib
+      - py3.12-importlib-metadata
+      - py3.12-numpy~1
+      - py3.12-oauthlib
+      - py3.12-requests-oauthlib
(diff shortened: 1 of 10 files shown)


cmwilson21:
Thanks for the contribution!

PR #66809, opened 2025-09-22
py3-keras: update numpy dependency

aborrero:
Update dependency on numpy to be the 2.2 version stream. This is because:
 * h5py requires numpy 2.2
 * both keras packages need to be in sync wrt. what numpy version to use


aborrero:
this is wrong, or incomplete:
 * there is also py3-keras
 * numpy-2.2 is an enterprise package

aborrero:
I think I will go this route instead: https://github.com/chainguard-dev/enterprise-packages/pull/34137

PR #66812, opened 2025-09-22
2609 jaeger stream

utieyin:
Fixes:

Related:

#### For new version streams

#### For version bump PRs


jaeger-2/fix-build-react-icons-typescript-errors.patch
```diff
@@ -0,0 +1,40 @@
...
+diff --git a/packages/plexus/package.json b/packages/plexus/package.json
+index c49bd0f3..840ab218 100644
+--- a/packages/plexus/package.json
++++ b/packages/plexus/package.json
+@@ -53,7 +53,7 @@
+     "d3-selection": "^3.0.0",
+     "d3-zoom": "^3.0.0",
+     "memoize-one": "6.0.0",
+-    "react-icons": "^5.0.1"
++    "react-icons": "5.0.1"
```
tulilirockz: I think this could be a sed expression instead? `sed "/react-icons/s/: ".*"/: \"5.0.1\"/"`

jaeger-2.yaml
```diff
@@ -147,3 +149,10 @@ test:
         timeout: 60
         expected_output: |
           Admin server started
+
+update:
+  enabled: true
+  github:
+    use-tag: true
+    identifier: jaegertracing/jaeger
+    strip-prefix: v
```
Dentrax, line 158: Can we also pin the version prefix to `2.` to pick always given version suffix in the package name.

jaeger-2.yaml
```diff
@@ -1,23 +1,25 @@
...
-      expected-commit: c17f97d4add9f96e52852f8f238689568e8a48d9
+      expected-commit: 076ea6f40af6b3f9f75a86caa23972fc0b99ca71
       recurse-submodules: true
 
   - working-directory: jaeger-ui
-    uses: patch
-    with:
-      patches: ../fix-build-react-icons-typescript-errors.patch
+    runs: |
+      sed -i '/react-icons/s/: "\^5\.0\.1"/: "5.0.1"/' packages/plexus/package.json
```
Dentrax: This would result build failure due to conflict in the future; so we could use the npm "resolutions" to bump the deps (there are some example)

Dentrax (dismissed):

Dentrax (approved):

PR #66861, opened 2025-09-22
llvm-21: Add new package for release 21.x

AbduAzim424:
- Add `llvm-21.yaml` package based on `LLVM 21.1.1` release
  - Update `openssf-compiler-options` to support `clang-21` hardening


openssf-compiler-options.yaml
```diff
@@ -1,14 +1,14 @@
...
-  epoch: 27
+  epoch: 28
   description: "Compiler Options Hardening Guide for C and C++"
   url: https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.html
   copyright:
     - license: CC-BY-4.0
 
 vars:
-  clang-major-versions: 14 15 16 17 18 19 20
+  clang-major-versions: 14 15 16 17 18 19 20 21
```
sergiodj, line 11: Hey Azim,

I know you're still working on this PR, but just leaving a comment here to remind to also extend the current tests as well.

See https://github.com/wolfi-dev/os/pull/66889 for a related PR.

Thanks!

sergiodj: Actually, let me correct myself.

It's not possible to co-install more than one clang version, which means that we will need to test the latest one only.  So you only need to adjust the test to make sure it's using `clang-21`.

AbduAzim424: Hi, thank you. I removed it after testing locally because I though we are keeping 1 test per package as a sanity check. Will add soon. Thank you again)

openssf-compiler-options.yaml
```diff
@@ -1,14 +1,14 @@
 package:
   name: openssf-compiler-options
   version: 20240627
-  epoch: 29
+  epoch: 31
```
javacruft: Double bump :)

llvm-21.yaml
```diff
@@ -0,0 +1,917 @@
...
+  resources:
+    cpu: 32
+    memory: 64Gi
+  dependencies:
+    provides:
+      - llvm=${{package.full-version}}
+
+vars:
+  llvm-prefix: "/usr/lib/llvm-21"
+  clang-prefix: "/usr/lib/llvm-21/lib/clang/21"
```
javacruft: I know they are like this in -20 but both of these could be generated in var-transforms, avoiding a hardcode of 21 - will help when we do 22 :)

justinvreeland:
This has been in draft so I haven't been reviewing it.  @AbduAzim424 do you want reviews on this or is this just here so you can test with the pre-submit repo?

AbduAzim424:
Yes this PR can be merged, I wanted to make sure what to do with failing packages before making it approvable

justinvreeland (approved):
👍🏻

PR #66911, opened 2025-09-22
py3-numpy-1.26: Make it conflict with all curret 2.X versions

justinvreeland:
This should make debugging and fixing things a bit easier and leave the sovler some room to find better matches itself.

Before:

```
2025/09/22 16:04:34 INFO installing py3.12-numpy-1.26 (1.26.5-r6) arch=x86_64
2025/09/22 16:04:34 INFO installing py3.12-numpy-2.2 (2.2.6-r0) arch=x86_64
Error: building "amd64" layer: installing apk packages: installing packages: installing py3.12-numpy-2.2 (ver:2.2.6-r0 arch:x86_64): unable to install files for pkg py3.12-numpy-2.2: writing header for "usr/lib/python3.12/site-packages/numpy/__config__.py": packages map[py3.12-numpy-1.26:py3-numpy-1.26 py3.12-numpy-2.2:py3-numpy-2.2] has conflicting file: "usr/lib/python3.12/site-packages/numpy/__config__.py"
2025/09/22 16:04:34 INFO error during command execution: building "amd64" layer: installing apk packages: installing packages: installing py3.12-numpy-2.2 (ver:2.2.6-r0 arch:x86_64): unable to install files for pkg py3.12-numpy-2.2: writing header for "usr/lib/python3.12/site-packages/numpy/__config__.py": packages map[py3.12-numpy-1.26:py3-numpy-1.26 py3.12-numpy-2.2:py3-numpy-2.2] has conflicting file: "usr/lib/python3.12/site-packages/numpy/__config__.py"
```

after:

```
2025/09/22 16:05:19 WARN getting local index /Users/justin/src/wolfi-dev/wolfi/packages/x86_64/APKINDEX.tar.gz: stat: stat /Users/justin/src/wolfi-dev/wolfi/packages/x86_64/APKINDEX.tar.gz: no such file or directory
Error: locking config: resolving apk packages: for arch "amd64": solving "py3.12-numpy-1.26=1.26.5-r7" constraint:   py3.12-numpy-1.26-1.26.5-r0.apk disqualified because "1.26.5-r0" does not satisfy "py3.12-numpy-1.26=1.26.5-r7"
  py3.12-numpy-1.26-1.26.5-r4.apk disqualified because "1.26.5-r4" does not satisfy "py3.12-numpy-1.26=1.26.5-r7"
[...]

PR #66912, opened 2025-09-22
withdraw remaining py3.X-numpy-1.26 package

justinvreeland:
withdraw remaining py3.X-numpy-1.26 package. This will mean that there will only be a py3.X-numpy-1.26 package that conflicts with the py3-numpy-2.X packages in the repo which should make it easier to find the packages that are using the wrong requires as well as help the solver find real combinations that work.

so we only have the confliting one.

This was re-added recently so it's surprising there's so few.

Took the list from apk.chaindag.dev

See also:

https://github.com/wolfi-dev/os/pull/66911


justinvreeland:
Oh i think i'm missing a few here.

AmberArcadia (approved):

PR #66924, opened 2025-09-23
tshark-4.4: test privileged version of tshark + add timelimit

stevebeattie:
Test that the variant of tshark with filesystemp capabilities set can capture packets as an unprivileged user ("build") and also limit the duration of the packet capture tests to 10 seconds, to ensure that tshark eventually quits in the face of network problems.

### DIFF
--- a/tshark-4.4.yaml
+++ b/tshark-4.4.yaml
@@ -1,7 +1,7 @@
 package:
   name: tshark-4.4
   version: "4.4.9"
-  epoch: 0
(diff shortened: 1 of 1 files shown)


stevebeattie:
Converting to draft because the dropped-privilege test fails in docker, used on arm64; this needs either
- a fix to land in melange's docker runner to fix up fs permissions to support non-root builds
- converting the arm runners to use the qemu runner.

PR #66995, opened 2025-09-23
py3-numpy-2.3: Allow it to be used as py3-numpy and conflict with others

justinvreeland:
This makes py3-numpy-2.3 provide `py3-numpy` so that `py3-numpy`_can_ install it.
It also provides `py${{range.key}}-numpy=${{package.version}}` so that other python version modules providing numpy conflict. Additionally it provides `py3-numpy-supported=${{package.version}}` to conflict with other numpy supported packages. This is not strictly needed since the `py3.XX-numpy...` package will conflict but this should cause the solver to point out that supported is conflicting instead of the module.

### DIFF
--- a/py3-numpy-2.3.yaml
+++ b/py3-numpy-2.3.yaml
@@ -1,7 +1,7 @@
 package:
   name: py3-numpy-2.3
   version: "2.3.3"
-  epoch: 0
+  epoch: 1
   description: The fundamental package for scientific computing with Python.
   copyright:
     - license: BSD-3-Clause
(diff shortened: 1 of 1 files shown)


philroche:
@justinvreeland Should we rebuild all packages that currently depend on py3-numpy to see if they need pinning before we make this change?

justinvreeland:
> @justinvreeland Should we rebuild all packages that currently depend on py3-numpy to see if they need pinning before we make this change?

I can't see anything depending on `py3-numpy` anymore after we pinned things. I built a fair number of packages with this and they seem fine.

But since nothing is using this right now maybe this is a good test case for rolling out the numpy changes across packages without relying on the unversioned python picking the most recent version.

My thinking right now for what makes sense is to keep the `py3-numpy-bin` provides. Remove the `py3-numpy` provides. And bumping packages from `py3.XX-numpy-2.2` to `py3.XX-numpy-2.3. This seems like the most likely path forward for these packages anyway and should make the version splitting issues we've seen go away. I still want to double check and build some images to make sure this won't popup as an issue there instead of in the packages.

cmwilson21:
Thanks for the contribution!

PR #67008, opened 2025-09-23
systemd: cherry-pick upstream patch exposing system UIDs/GIDs in pkgconf

markusboehme:
Cherry-pick upstream patch 346b7b6b ("pkgconf: expose variables for system-alloc-{uid,gid}-min") that will likely only land in systemd v259. It will be useful to sanity check UID/GID values added to login.defs owned by the shadow package.

Related: https://github.com/wolfi-dev/os/pull/64005 (which does this for maximum UIDs already)

#### For PRs that add patches

- [X] Patch source is documented (cherry-pick of upstream commit, commit ID in patch and commit message)

### DIFF
--- a/systemd.yaml
+++ b/systemd.yaml
@@ -1,6 +1,5 @@
 package:
   name: systemd
-  # 35712.patch can be dropped when 258 releases
   version: "258"
   epoch: 0
   description: The systemd System and Service Manager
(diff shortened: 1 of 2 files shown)


markusboehme:
Fixed the systemd update to v258 in https://github.com/wolfi-dev/os/pull/67009 and rebased onto it.

89luca89 (approved):

PR #67020, opened 2025-09-23
fix(rancher-charts): Retain latest git commit

EyeCantCU:
The latest commit is used to determine the version of the repository containing charts used by Rancher

Also drop Helm 3 charts as a runtime dependency of Rancher 2.12


kranurag7 (approved):
read the changes and went through the notes on the issue, looks good and safe to land. as a follow-up, we can consider removal of `rancher-helm3-charts` from wolfi as well to avoid confusion and maintain less.

Issue #58294, opened 2025-07-03
Building with GCC 15 on ARM64 sometimes fail due to GCS


sergiodj:
A number of build failures are being reported which relate to using GCC 15 on ARM64.  The compiler recently gained more capabilities w.r.t. [GCS](https://docs.kernel.org/arch/arm64/gcs.html), and that's interacting badly with `ld`.

For example, when trying to use `cgo`  on ARM64 with GCC 15, we currently see the following problem:

```
$ cat > 1.go << _EOF_
        package main
        /*
        #include <stdlib.h>
        #include <stdio.h>
        void hello() {
            printf("%s\\n", "Hello from cgo!");
            fflush(stdout);
        }
        */
        import "C"
        func main() {
            C.hello()
        }
_EOF_
$ go run 1.go
# command-line-arguments
/usr/lib/go/pkg/tool/linux_arm64/link: running gcc failed: exit status 1
/usr/local/bin/gcc -s -o $WORK/b001/exe/1 -rdynamic /tmp/go-link-624876020/go.o /tmp/go-link-624876020/000000.o /tmp/go-link-624876020/000001.o /tmp/go-link-624876020/000002.o /tmp/go-link-624876020/000003.o /tmp/go-link-624876020/000004.o /tmp/go-link-624876020/000005.o /tmp/go-link-624876020/000006.o /tmp/go-link-624876020/000007.o /tmp/go-link-624876020/000008.o /tmp/go-link-624876020/000009.o /tmp/go-link-624876020/000010.o /tmp/go-link-624876020/000011.o /tmp/go-link-624876020/000012.o /tmp/go-link-624876020/000013.o /tmp/go-link-624876020/000014.o /tmp/go-link-624876020/000015.o -O2 -g -O2 -g -lpthread
/tmp/go-link-624876020/go.o: warning: GCS is required by -z gcs, but this input object file lacks the necessary property note.
/tmp/go-link-624876020/000015.o: warning: GCS is required by -z gcs, but this input object file lacks the necessary property note.
/usr/lib/gcc/aarch64-unknown-linux-gnu/15/../../../../lib64/libgcc_s.so.1: warning: GCS is required by -z gcs, but this shared library lacks the necessary property note. The dynamic loader might not enable GCS or refuse to load the program unless all the shared library dependencies have the GCS marking.
[...]

PR #67026, opened 2025-09-24
DNM: OpenSSF update

sergiodj:
Update our compiler flags according to version 20250904.

This pretty much entails adding `-fzero-init-padding-bits=all` to the
list of compilation flags.  There have not been many updates to the
OpenSSF list since our last update.

I'm taking this opportunity to re-enable ARM64's GCS support.  This is
an important security feature which had to be disabled due to
https://github.com/wolfi-dev/os/issues/58294.  That issue is no longer
applicable because our ARM64 golang defaults to using GNU `ld` now,
which has proper support for GCS.

This has been tested by rebuilding the archive.  I also did some
manual tests to make sure the GCS feature works as expected.

Relates: https://github.com/chainguard-dev/internal-dev/issues/14606

### DIFF
--- a/openssf-compiler-options.yaml
+++ b/openssf-compiler-options.yaml
@@ -1,7 +1,7 @@
 package:
   name: openssf-compiler-options
-  version: 20240627
-  epoch: 33
+  version: 20250904
+  epoch: 0
   description: "Compiler Options Hardening Guide for C and C++"
   url: https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.html
   copyright:
--- a/openssf-compiler-options/usr/lib/gcc/aarch64-unknown-linux-gnu/15/openssf.spec
+++ b/openssf-compiler-options/usr/lib/gcc/aarch64-unknown-linux-gnu/15/openssf.spec
@@ -1,6 +1,17 @@
+# For GCC 15, -fhardened enables:
+#
+# -D_FORTIFY_SOURCE=3
+# -D_GLIBCXX_ASSERTIONS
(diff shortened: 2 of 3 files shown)


xnox:
> That issue is no longer
> applicable because our ARM64 golang defaults to using GNU `ld` now,
> which has proper support for GCS

well, some of our images do not use GOTOOLCHAIN=local, and instead use GOTOOLCHAIN=local+auto which can auto-update to non-chainguard go toolchain.

and our go images ship openssf-compiler options, thus I think we might need to push the breaking change to update GOTOOLCHAIN=local.

Fun upstream uses GOTOOLCHAIN=local and we don't

sergiodj:
> > That issue is no longer
> > applicable because our ARM64 golang defaults to using GNU `ld` now,
> > which has proper support for GCS
> 
> well, some of our images do not use GOTOOLCHAIN=local, and instead use GOTOOLCHAIN=local+auto which can auto-update to non-chainguard go toolchain.

Ouch!  I didn't know that.

> and our go images ship openssf-compiler options, thus I think we might need to push the breaking change to update GOTOOLCHAIN=local.
> 
> Fun upstream uses GOTOOLCHAIN=local and we don't

Hm, that sounds like a *real* breaking change, indeed.  I'll ping you in private to discuss it better.  Thanks for the comment!

xnox:
> Hm, that sounds like a _real_ breaking change

actually it might be easy to land, yeah let's chat elsewhere.

xnox:
Both changes can happen independent of each other.

sergiodj:
@xnox I believe you can approve this, right?  Otherwise I can ask someone else from the OS team.

xnox (approved):

PR #67082, opened 2025-09-24
py3-nltk: numpy-2.2: update dependant packages to use it

aborrero:


aborrero:
re-triggering CI

efbar (approved):

PR #67168, opened 2025-09-24
fix(superset-5.0): correct conditional usage

bentasker:
The previous change introduced square brackets `[[` resulting in the conditional being evaluated with `test`. 

`test` does not look at exit statuses and so `am_i_root` always evaluated true.

This PR also switches to using a patch rather than `sed` to help make the logic clearer when reviewing YAML.

### DIFF
--- a/superset-5.0.yaml
+++ b/superset-5.0.yaml
@@ -1,7 +1,7 @@
 package:
   name: superset-5.0
   version: "5.0.0"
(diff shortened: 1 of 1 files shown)


IdlePhysicist (approved):
This works for me, but you can add a patch file to the repo in the superset-5.0 directory, and then reference the file in the build.

PR #67416, opened 2025-09-25
Remove remaining eudev deps in wolfi and use systemd

justinvreeland:
These should've been cleaned up a while ago.

### DIFF
--- a/thingsboard.yaml
+++ b/thingsboard.yaml
@@ -1,7 +1,7 @@
 package:
   name: thingsboard
   version: "4.2"
-  epoch: 7 # GHSA-2hmj-97jw-28jh
+  epoch: 8 # GHSA-2hmj-97jw-28jh
   description: "Open-source IoT Platform - Device management, data collection, processing and visualization."
   copyright:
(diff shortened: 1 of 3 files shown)


thingsboard.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: thingsboard
   version: "4.2"
-  epoch: 7 # GHSA-2hmj-97jw-28jh
+  epoch: 8 # GHSA-2hmj-97jw-28jh
```
murraybd, line 4: I thought @jamie-albert advocated for removing these comments?

justinvreeland: Happy to change it to whatever is right. I just wolfictl bumped these and it didn't remove the comment.

xnox: It is only safe to remove these if all PRs for this package have been merged and there are no outstanding drafts.

xnox: Also harmless.

murraybd (approved):

PR #67500, opened 2025-09-26
libtool: switch to building from tarball

OddBloke:
The `bootstrap` script performs a `git clone` from an unreliable server, currently experiencing a days-long outage.

This also stops ignoring odd-minor-versioned releases: those are no longer reserved for development snapshots.

Closes: #66525

### DIFF
--- a/libtool.yaml
+++ b/libtool.yaml
@@ -1,7 +1,7 @@
 package:
   name: libtool
   version: 2.4.7
-  epoch: 6
+  epoch: 7
   description: "GNU libtool"
   copyright:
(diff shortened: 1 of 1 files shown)


OddBloke:
I'll let automation perform the bump to the new version, as I'd like to get the infra-related FTBFS fixed independent of any issues the newer version might have.

(We do have https://github.com/wolfi-dev/os/pull/66525 open already, but libtool's website doesn't yet list 2.6.0 as stable so this PR will close that out. I think release-monitoring is a more reliable source than git.)

IdlePhysicist (approved):

PR #67507, opened 2025-09-26
Cilium 1.18 more testing

murraybd:
This copies the testing for other cilium packages over to 1.18.

### DIFF
--- a/cilium-1.18.yaml
+++ b/cilium-1.18.yaml
@@ -1,7 +1,7 @@
 package:
   name: cilium-1.18
   version: "1.18.2"
-  epoch: 0 # GHSA-2464-8j7c-4cjm
+  epoch: 1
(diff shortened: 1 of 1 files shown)


cilium-1.18.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: cilium-1.18
   version: "1.18.2"
-  epoch: 0 # GHSA-2464-8j7c-4cjm
+  epoch: 1
```
dannf, line 4: do we want to drop the # GHSA thing?

murraybd: AIUI it's added by the bots to indicate that a specific issue e.g. rebuild has been addressed to prevent them from continually updating a package. Given the epoch has been bumped the comment no longer serves any purpose.

justinvreeland (approved):

PR #67617, opened 2025-09-29
6135 ceph package

utieyin:
Adding changes to fix pyo3 issues with ceph mgr


ceph.yaml
```diff
@@ -126,15 +132,27 @@ environment:
       - xfs-scrub
       - xfsprogs
   environment:
-    CXXFLAGS: "-include cstdint"
+    NODE_OPTIONS: "--openssl-legacy-provider"
+    CEPH_BUILD_VIRTUALENV: "build"
+    RUSTFLAGS: "build_rustflags --cfg pyo3_unsafe_allow_subinterpreters"
```
javacruft: I don't believe this is actually needed now we have the cherry pick in place to avoid this completely.

javacruft (approved):

PR #67639, opened 2025-09-29
Revert ossl 3.5.3

sil2100:
Only use this in case of emergencies.


xnox (dismissed):

xnox:
Note as we can't build packages, openssl should be reverted back to last version & last revision, without rebuilding it.

Ditto openssh should be withdrawn and downgraded to last version & revision.

Such that this PR elastic build should say "no packages to build" and withdrawn-packages has -r6 openssh and 3.5.3 openssl apks.

xnox (approved):

PR #67675, opened 2025-09-29
feat: Add Python pipelines to consume Chainguard Libraries

EyeCantCU:
This adds a few pipelines for both authentication with our indexes, and for building packages that consume Chainguard Libraries:

- **py/bundle**: Leverages uv to install Python packages from Guarded Libraries. Used exclusively for bundles of Python packages
- **auth/uv**: Configures uv to pull Chainguard Libraries
- **auth/pip**: Configures pip to pull Chainguard Libraries

A few packages have been moved over to Chainguard Libraries in the process:

- **jupyter-base-notebook**: Demonstrates packaging of a Python bundle, uses some libraries. Some, such as notebook itself, were not present at the time of testing
- **superset-5.0**: Demonstrates working authentication with pip in a virtual environment, consumes some libraries. Will need to be fully migrated in the future

This change also globally disallows downloading external builds of Python with uv. External Python builds should never be used to produce packages.


jupyter-base-notebook.yaml
```diff
@@ -1,59 +1,59 @@
...
   version: "7.4.6"
-  epoch: 0
+  epoch: 40
   description: Jupyter Notebook
+  options:
+    no-provides: true
   dependencies:
     runtime:
-      - python-3.12
+      - python-${{vars.py-version}}
```
justinvreeland: @smoser just moved these to `python-version` so i'd like to do that.

jupyter-base-notebook.yaml
```diff
@@ -1,59 +1,59 @@
 package:
   name: jupyter-base-notebook
   version: "7.4.6"
-  epoch: 0
+  epoch: 40
```
justinvreeland: 40?

py3-poetry.yaml
```diff
@@ -25,19 +25,12 @@ data:
 environment:
   contents:
     packages:
-      - build-base
       - busybox
-      - ca-certificates-bundle
       - py3-supported-pip
       - py3-supported-poetry-core
-      - wolfi-base
```
justinvreeland, line 33: Why remove wolfi base?

jupyterhub-k8s-hub.yaml
```diff
@@ -1,46 +1,45 @@
...
-      - py3-oauthenticator
-      - py3-psycopg2
-      - py3-pycurl
-      - py3-pyjwt
-      - py3-pymysql
-      - py3-python-dateutil
-      - py3-sqlalchemy-cockroachdb
-      - py3-statsd
-      - py3-tornado
-      - python3
```
justinvreeland, line 33: Oh wow that's a huge improvement versioning them.

pipelines/auth/pip.yaml
```diff
@@ -0,0 +1,53 @@
...
+      echo "Removing pip config, if exists..."
+      rm -rf $HOME/.config/pip && mkdir -p $HOME/.config/pip
+
+      python_index="python"
+      if [ "${{inputs.guarded}}" = "false" ]; then
+        python_index="python-all"
+      fi
+
+      echo "Creating pip config..."
+      cat > $HOME/.config/pip/pip.conf <<EOF
```
justinvreeland: Double quote this and 35 please. I don't think we run any shellcheck on pipelines unfortunatly.

justinvreeland (changes requested):
Looks good, looking forward to trying it out.

tulilirockz:
+1 here, I _really_ want this

py3-poetry.yaml
```diff
@@ -25,19 +25,12 @@ data:
...
 
 pipeline:
-  - uses: git-checkout
-    with:
-      repository: https://github.com/python-poetry/poetry
-      tag: ${{package.version}}
-      expected-commit: b9e5d79fc57de2f2e60973019d56662b7398440b
```
javacruft, line 40: I was concerned that removing the git-checkout block might have an unintended side effect on automatic updates but AFAICT (have code-surfed for 10 minutes) I don't think that it will.

javacruft: That said it does have a side-effect on the build-time SBOM we publish for the package - it no longer has an upstream source package and associated downloadLocation populated so we're going to loose some information.

The libraries published python module does contain an SBOM which includes this information but we're not digesting that into the APK SBOM right now - that will need changes into melange.

py3-poetry.yaml
EyeCantCU: It seems unnecessary to include it if it isn't used directly at build time

jupyter-base-notebook.yaml
EyeCantCU: I wanted to make it obvious this thing is built with libraries now at this point. Will increment by 1 though

py3-poetry.yaml
EyeCantCU: Thanks for the heads up, I'll restore the checkout

pipelines/auth/pip.yaml
Taffer: Hooking `shellcheck` up in your editor is always a good idea.

justinvreeland: Even the venerable emacs gets super annoyed trying to parse shell inside a multi string yaml file.  I would like to make a melange lsp or something that does it better if i ever find time. It'd be nice to get completion on the pipelines too. Anywho this is why we have [tw](https://github.com/chainguard-dev/tw) because checking shell inline is annoying, for stuff like this i cheat and use `yq` though. `shellcheck (curl https://raw.githubusercontent.com/wolfi-dev/os/c3a06703fa45f81fca3e1cf57c24eda73f7ecd52/pipelines/auth/pip.yaml | yq '.pipeline[0].runs' | psub)`

Taffer: We have patch files as stand-alone files in `{package}/`; what about shell scripts? Or is there a strong preference for keeping them inline in the `{package}.yaml`?

jamonation:
Looks like `/usr/bin/jupyterhub` is missing. Today it comes from our `py3.13-jupyterhub-bin` package.

justinvreeland:
It looks like Jupyter lost all of it's cmd provides:

```
- 	Provides: []string{
- 		"cmd:alembic=7.4.6-r0", "cmd:certipy=7.4.6-r0", "cmd:debugpy-adapter=7.4.6-r0",
- 		"cmd:debugpy=7.4.6-r0", "cmd:httpx=7.4.6-r0", "cmd:ipython3=7.4.6-r0",
- 		"cmd:ipython=7.4.6-r0", "cmd:jlpm=7.4.6-r0", "cmd:jsonpointer=7.4.6-r0",
- 		"cmd:jsonschema=7.4.6-r0", "cmd:jupyter-dejavu=7.4.6-r0",
- 		"cmd:jupyter-events=7.4.6-r0", "cmd:jupyter-execute=7.4.6-r0",
- 		"cmd:jupyter-kernel=7.4.6-r0", "cmd:jupyter-kernelspec=7.4.6-r0",
- 		"cmd:jupyter-lab=7.4.6-r0", ...,
- 	},
```

justinvreeland:
The so dep changes seem slightly suspicious too:
```
  	Dependencies: []string{
- 		"cmd:python3.12",
+ 		"cmd:python3.13",
- 		"python-3.12",
+ 		"python-3.13",
  		"so:ld-linux-aarch64.so.1",
  		"so:libc.so.6",
- 		"so:libdl.so.2",
  		"so:libgcc_s.so.1",
- 		"so:libm.so.6",
  		"so:libpthread.so.0",
- 		"so:librt.so.1",
  		"so:libstdc++.so.6",
```

EyeCantCU:
@justinvreeland I intentionally enabled no-provides as the package is a bundle of multiple Python packages. We have other packages that provide those commands as well

A caller of `cmd:jupyterhub` likely wants `py3.13-jupyterhub-bin` and not `jupyter-base-notebook`

I also incremented Python to 3.13 explaining the library changes

EyeCantCU:
`libm`, `librt`, and `libdl` are curious... looking into those

jamonation:
Is it worth splitting out the package changes so that the auth changes can be used here separately?

tulilirockz:
+1 for splitting

EyeCantCU:
It doesn't. Let's wait until we've dug into this more and understand what's going on

EyeCantCU:
All changes are related to us building against newer glibc. Let's proceed

justinvreeland:
> All changes are related to us building against newer glibc. Let's proceed
Awesome thanks for looking into that!

justinvreeland (dismissed):

pipelines/auth/uv.yaml
```diff
@@ -0,0 +1,45 @@
...
+      fi
+
+      python_index="python"
+      if [ "${{inputs.guarded}}" = "false" ]; then
+        python_index="python-all"
+      fi
+
+      echo "Creating uv config..."
+      mkdir -p /etc/uv
+      cat > /etc/uv/uv.toml <<EOF
```
dannf, line 44: The pip config is local to the user configuration - should we do the same for `uv` for consistency? (e.g. `~/.config/uv/uv.toml`)?

build-x86_64.env
```diff
@@ -2,4 +2,7 @@ export GOFLAGS=""
 export GOTOOLCHAIN=local
 
 # Use a reproducible hash seed when compiling Python bytecode
-export PYTHONHASHSEED=0
\ No newline at end of file
+export PYTHONHASHSEED=0
+
+# Do not download external Python builds with UV
+UV_PYTHON_DOWNLOADS=never
```
dannf, line 8: Clearly the right thing to do - but have you verified that nothing relies on this today?

pipelines/auth/pip.yaml
```diff
@@ -0,0 +1,53 @@
+name: Configure pip to fetch Chainguard Libraries dependencies.
```
dannf, line 1: Do we need separate pipelines for pip and uv, or could we just configure them both in the same pipeline?

pipelines/auth/pip.yaml
```diff
@@ -0,0 +1,53 @@
...
+      echo "Removing pip config, if exists..."
+      rm -rf "$HOME/.config/pip" && mkdir -p "$HOME/.config/pip"
+
+      python_index="python"
+      if [ "${{inputs.guarded}}" = "false" ]; then
+        python_index="python-all"
+      fi
+
+      echo "Creating pip config..."
+      cat > "$HOME/.config/pip/pip.conf" <<EOF
```
dannf, line 46: Should we error out if this file already exists (same q for uv)?

pipelines/auth/pip.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - jq
+
+inputs:
+  identity:
+    description: the identity to use for Chainguard Libraries
+    default: "720909c9f5279097d847ad02a2f24ba8f59de36a/a49c7fedc33adf69"
+  group:
+    description: the group to use for Chainguard Libraries
+    default: "720909c9f5279097d847ad02a2f24ba8f59de36a"
+  guarded:
```
dannf: how about "no-index-fallback" or "guarded-only"?

pipelines/auth/pip.yaml
```diff
@@ -0,0 +1,53 @@
...
+      # Fetch the token with: make lib-token
+      if [ -f /var/cache/melange/.libraries_token.txt ]; then
+        cgtoken=$(cat /var/cache/melange/.libraries_token.txt)
+      else
+        idtoken=$(curl --fail-with-body -sSL -H "Metadata-Flavor: Google" http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/identity?audience=issuer.enforce.dev)
+        cgtoken=$(curl --fail-with-body -sSL -H "Authorization: Bearer ${idtoken}" 'https://issuer.enforce.dev/sts/exchange?aud=libraries.cgr.dev&scope=${{inputs.group}}&identity=${{inputs.identity}}' | jq -r '.token')
+      fi
+
+      echo "Removing pip config, if exists..."
+      rm -rf "$HOME/.config/pip" && mkdir -p "$HOME/.config/pip"
```
dannf, line 38: Should we error here instead of continuing? That would make debugging easier, in case some package was relying on an existing configuration here.

pipelines/py/bundle.yaml
```diff
@@ -0,0 +1,36 @@
+name: Install Python packages with UV
+
+needs:
+  packages:
+    - python-${{inputs.python-version}}-base
+    - uv
+
+inputs:
+  dependencies:
+    description: "Indirect dependencies to install."
```
dannf: It's not clear to me what this means, or what syntax to use to specify them.

pipelines/py/bundle.yaml
```diff
@@ -0,0 +1,36 @@
...
+    default: ""
+  package:
+    description: "Name of packages to install."
```
dannf: It's called "package" but says "packages". Is this just for 1 package or multiple? I expect we'd normally want the full pin syntax. What if we allowed passing in a requirements file and/or lockfile instead?

pipelines/py/bundle.yaml
```diff
@@ -0,0 +1,36 @@
...
+    required: true
+  python-version:
+    description: "Python version to use."
```
dannf: How about a python path instead? Looks like `UV_PYTHON` supports that. It would be more consistent with our other python pipelines, and would lend itself to venvs in cases where that's necessary.

dannf (changes requested):
Thanks @EyeCantCU! some comments inline

pipelines/auth/uv.yaml
EyeCantCU: Will do

build-x86_64.env
EyeCantCU: I checked. Everywhere we use uv today we also provide a system installation of Python. This more a safe measure to prevent this from ever happening versus forcing a stop to anything we're currently doing

pipelines/auth/pip.yaml
EyeCantCU: We cam throw them in the same pipeline

pipelines/auth/pip.yaml
EyeCantCU: An error if the file exists is a good idea, let's do that

pipelines/auth/pip.yaml
EyeCantCU: As in just changing the name or default behavior as well? I don't think we'd want fallback to be default instead of opt in (because then we know what packages we need to revisit, then again you could do the same in reverse but it would be harder to filter)

pipelines/auth/pip.yaml
EyeCantCU: We force override uv to prioritize using our config as otherwise it will default to any local uv config. We could error out if this happens to exist though

pipelines/py/bundle.yaml
EyeCantCU: We have some packages like jupyter base notebook that install indirect dependencies. This allows you to pass a list of indirect deps

pipelines/py/bundle.yaml
EyeCantCU: I figured we would just install the package by name if it existed. We could fallback to supporting a lockfile or requirements file though

EyeCantCU: Typo, one package that all direct deps would be retrieved via

pipelines/py/bundle.yaml
EyeCantCU: We can do that

pipelines/py/bundle.yaml
dannf: Yeah, I suppose ultimately we want to support both. It would be annoying to have to generate a requirements file if you just need a single package. But there are a number of packages that will benefit from `requirements.txt` support.

pipelines/auth/pip.yaml
dannf: just the name. The default should be to *only* install guarded things. I like names that i can read and I immediately understand broadly what it does. `guarded` true/false doesn't meet that bar for me. We're installing guarded libs either way. "only-guarded" or "no-index-fallback" or something like that tells me what it does on the tin.

TBH, I'm actually not even sure I like having a way to enable a fallback. I'm sure we'll need an escape hatch at times, but I'd prefer something more granular and declarative than "whatever the resolver happens to choose at that moment". I'd like for developers to explicitly list the things that they are pulling from any other index and why.

WDYT about putting this on hold? Or making it a flag that is clearly only meant for debug?

pipelines/auth/uv.yaml
dannf: hm.. given we've a pipeline to do the installs, do we need to even write anything persistent, or just set environment variables in the install pipeline?

pipelines/auth/pip.yaml
EyeCantCU: Definitely agree with changing the name, let's do that. But I think we're going to need fallback as an option because we have existing things that already retrieve packages from other indexes (like jupyter base notebook). I'd consider any amount of libraries that came from us in those packages to be a net improvement. But for net new things we are adding, we need to stick with things we've built. I will make it clearer that it isn't some escape hatch for building net new things, and needs to be used exclusively for debug or older things

We can gate individual packages in our uv pipeline to accommodate that. I like the idea... It probably makes the most sense for us to just do that so that we instantly see what packages we need to build and drop auth setup with pip altogether

EyeCantCU: Hmm.. Can't specify sources in the uv config directly. You only get that level of granularity if you create and initialize a project which isn't what we want for packaging. I'll leave the toggle but rename and add clarity

pipelines/py/bundle.yaml
EyeCantCU: Made this clearer in `py/compile-requirements`

pipelines/auth/pip.yaml
EyeCantCU: Added clarity and renamed

pipelines/py/bundle.yaml
```diff
@@ -0,0 +1,71 @@
...
+    - uv
+
+inputs:
+  package:
+    description: |
+      "Main package. Used to validate package version and requirements are in sync."
+    default: ""
+  python:
+    description: "Path to Python executable."
+    default: "/usr/bin/python3"
```
smoser, line 15: I'm not exactly sure in this scenario.
but /usr/bin/python3 is painful in many ways as it is "sticky" and gets written into shbangs and other places.  Its best to just declare the version.

smoser (commented):
general comment. 
this is a lot of yaml bash.

tw provides one way of getting out of yaml-bash (even if only replacing it with bash).  It also provides a solution to copying pipelines around.

pipelines/py/bundle.yaml
EyeCantCU: I can walk that back

EyeCantCU:
I agree. I started incorporating requirements generation over in tw the other day and have a draft up over there for compiling requirements

cmwilson21:
Thanks for the contribution!

PR #67688, opened 2025-09-30
need-root: package builds that do setcap need to run as root

stevebeattie:
Set the build phase to run as root for packages that set filesystem capabilities on some of the binaries.


kranurag7 (approved):
looks good to land, excited for what's coming next.

PR #67808, opened 2025-10-01
gnupg: Update to version 2.4.8

sergiodj:
- Refresh and update existing patches.

- Remove 0320-gpg-default-to-aes256.patch; upstream now defaults to
  AES-256.

- Remove 0410-make-gpg-zip-use-tar-from-path.patch; upstream doesn't
  ship gpg-zip anymore.

- Update release-monitor identifier to 1215, which tracks the latest
  non-dev package version.


gnupg.yaml
```diff
@@ -333,8 +304,6 @@ subpackages:
     test:
       pipeline:
         - runs: |
-            gpg-zip --version
-            gpg-zip --help
             gpgparsemail --help
             gpgsplit --version
```
murraybd: While you are here switching to tw's ver-check would be nice.

sergiodj: Ah, shoot.  I was going to do it and then forgot.  Let me do it now.

sergiodj: Done.

murraybd (approved):

PR #67809, opened 2025-10-01
Add debianutils package

jamonation:
#### For new package PRs only
- [X] Functional tests are present.

### DIFF
--- a/debianutils.yaml
+++ b/debianutils.yaml
@@ -1,0 +1,199 @@
+package:
+  name: debianutils
+  version: 5.23.2
+  epoch: 0
+  description: Miscellaneous utilities specific to Debian
+  copyright:
+    - license: GPL-2.0-or-later
+    - license: SMAIL-GPL
+
+environment:
+  contents:
+    packages:
+      - autoconf
+      - automake
+      - build-base
+      - busybox
+      - coreutils
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://salsa.debian.org/debian/debianutils.git
+      tag: debian/${{package.version}}
+      expected-commit: b999bb0969471d302f1db85e58f17dcb64f274c2
+
+  - uses: autoconf/configure
+
+  - runs: |
+      # Build only the C programs, skip po4a translations
+      make run-parts tempfile ischroot
+
+  - runs: |
+      # Install manually to avoid make install tripping on .po generation
+      install -d ${{targets.destdir}}/usr/bin
+      install -m755 run-parts ${{targets.destdir}}/usr/bin/
+      install -m755 tempfile ${{targets.destdir}}/usr/bin/
+      install -m755 ischroot ${{targets.destdir}}/usr/bin/
+      install -m755 which.debianutils ${{targets.destdir}}/usr/bin/
+      install -m755 savelog ${{targets.destdir}}/usr/bin/
+      install -m755 installkernel ${{targets.destdir}}/usr/bin/
+      install -m755 add-shell ${{targets.destdir}}/usr/bin/
+      install -m755 remove-shell ${{targets.destdir}}/usr/bin/
+      install -m755 update-shells ${{targets.destdir}}/usr/bin/
+      # Copy license info in to cover which.debianutils and savelog
+      install -d ${{targets.destdir}}/usr/share/doc/debianutils
+      install -m644 debian/copying ${{targets.destdir}}/usr/share/doc/debianutils/
+
+  - uses: strip
+
+test:
+  environment:
+    contents:
+      packages:
+        - coreutils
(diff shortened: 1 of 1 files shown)


debianutils.yaml
```diff
@@ -0,0 +1,199 @@
...
+      # Build only the C programs, skip po4a translations
+      make run-parts tempfile ischroot
+
+  - runs: |
+      # Install manually to avoid make install tripping on .po generation
+      install -d ${{targets.destdir}}/usr/bin
+      install -m755 run-parts ${{targets.destdir}}/usr/bin/
+      install -m755 tempfile ${{targets.destdir}}/usr/bin/
+      install -m755 ischroot ${{targets.destdir}}/usr/bin/
+      install -m755 which.debianutils ${{targets.destdir}}/usr/bin/
```
EyeCantCU: This should be `/usr/bin/which`. Then, above, under `dependencies` we can add a:

```yaml
replaces:
  - busybox
```

Allowing the `debianutils` version of `which` to replace the `busybox` provided version

jamonation: Hmm, smart!

jamonation: Added. I suppose `/usr/bin/which.debianutils` symlinking to `/usr/bin/which` is overkill here?

debianutils.yaml
```diff
@@ -0,0 +1,202 @@
...
+
+        # Test which.debianutils (no --version flag)
+        which.debianutils echo
+
+        # Test shell management scripts exist
+        test -f /usr/bin/add-shell
+        test -f /usr/bin/remove-shell
+        test -f /usr/bin/update-shells
+        test -f /usr/bin/installkernel
+        test -f /usr/bin/savelog
```
EyeCantCU: If we want to check for existence of a file, I'd prefer we used `stat`. We get more verbose information about the file we're looking for versus only whether it exists or not. If debug isn't enabled, we also lose visibility when `test` fails so the problem because less obvious when we're debugging failing tests

debianutils.yaml
```diff
@@ -0,0 +1,202 @@
...
+        echo "run-parts test passed"
+    - runs: |
+        set -euo pipefail
+
+        # Test tempfile creates unique files
+        TEMP1=$(tempfile)
+        TEMP2=$(tempfile)
+        test "$TEMP1" != "$TEMP2"
+        test -f "$TEMP1"
+        test -f "$TEMP2"
```
EyeCantCU: Same here. Though, if we want to be sure the files created aren't the same, let's make it an if statement with verbosity

debianutils.yaml
```diff
@@ -0,0 +1,202 @@
...
+          exit 1
+        fi
+
+        # Test update-shells functionality
+        update-shells
+
+        # Restore backup
+        if [ -f /tmp/shells.backup ]; then
+          mv /tmp/shells.backup /etc/shells
+        fi
```
EyeCantCU: If `/tmp/shell.backup` doesn't exist, tests carry on normally when they should fail

debianutils.yaml
```diff
@@ -0,0 +1,202 @@
...
+        echo "which.debianutils test passed"
+    - runs: |
+        set -euo pipefail
+
+        # Create test log file
+        echo "test log content" > /tmp/test.log
+
+        # Test savelog rotation
+        savelog /tmp/test.log
+        test -f /tmp/test.log.0
```
EyeCantCU: Same as above

debianutils.yaml
jamonation: Personally don't think these are useful tests. Came from an earlier iteration before making functional tests. Shall I remove them?

debianutils.yaml
jamonation: True, and not needed in any case so will remove.

debianutils.yaml
jamonation: Added [here](https://github.com/wolfi-dev/os/pull/67809/files#diff-5d76bf184b75607e74a2df6836cd2f18f1a8c4dfadfdb2d7bb5a50a3ce826cf9R75-R80)

debianutils.yaml
jamonation: Updated [here](https://github.com/wolfi-dev/os/pull/67809/files#diff-5d76bf184b75607e74a2df6836cd2f18f1a8c4dfadfdb2d7bb5a50a3ce826cf9R129-R132)

debianutils.yaml
jamonation: Removed

debianutils.yaml
jamonation: Updated

EyeCantCU (approved):

PR #67842, opened 2025-10-01
nginx-mod-otel build

MickeyPvX:
Builds the `nginx-mod-otel` subpackage.

Related: https://github.com/chainguard-dev/image-requests/issues/7001

### DIFF
--- a/nginx-mainline.yaml
+++ b/nginx-mainline.yaml
@@ -3,7 +3,7 @@ package:
   # MAINLINE VERSIONS MUST USE ODD MINOR VERSIONS (e.g., 1.27.x, 1.29.x)
   # Must also bump njs at the same time
   version: "1.29.1"
-  epoch: 3
+  epoch: 4
   description: HTTP and reverse proxy server (mainline version)
   copyright:
     - license: BSD-2-Clause
@@ -13,9 +13,7 @@ package:
     # Let's ensure that we have config. This gets the default, but can also
     # be override by users.
     runtime:
-      - merged-usrsbin
       - nginx-config
-      - wolfi-baselayout
 
 environment:
   contents:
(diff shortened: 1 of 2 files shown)


nginx-mainline.yaml
```diff
@@ -45,6 +50,11 @@ var-transforms:
     replace: '$2'
     to: minor-version
 
+vars:
+  # OpenTelemetry commit SHA compatible with nginx mainline
+  # Based on latest nginx tag from opentelemetry-cpp-contrib
+  otel-sha: "5be4b76ebf67fee4d09f897649fd8d9348ba5d73"
```
maxgio92, line 56: Would it be possible to check that the supported version is aligned with nginx upstream, like we did for ingress-nginx-controller/otel-plugin here https://github.com/wolfi-dev/os/commit/78949bc0ca7e925b0c25ff3162c9617641d05a7e?

If so, it would prevent us from building unsupported or old versions of the plugin.

nginx-stable.yaml
```diff
@@ -25,16 +23,23 @@ environment:
       - build-base
       - busybox
       - ca-certificates-bundle
+      - cmake
+      - curl-dev
       - gd-dev
       - geoip-dev
+      - git
```
maxgio92, line 30: Optionally we can omit it since `git-checkout` brings it.

nginx-mainline.yaml
```diff
@@ -26,16 +24,23 @@ environment:
       - build-base
       - busybox
       - ca-certificates-bundle
+      - cmake
+      - curl-dev
       - gd-dev
       - geoip-dev
+      - git
```
maxgio92, line 31: Optionally we can omit it since `git-checkout` brings it.

nginx-mainline.yaml
MickeyPvX: So I'm just coming across what I think is the proper nginx-otel package...which is not the opentelemetry one...this may need a redo, so I'm going to put this into draft while I sort this out 🤦

PR #67860, opened 2025-10-01
Migration openjdk-24 -> openjdk-25

javacruft:
OpenJDK 25 is an LTS release; bump packages using 24.


sergiodj (approved):
An idea for the future is to also use variables here to define the Java version.

PR #67862, opened 2025-10-01
Updates to 'argo-workflows' naming scheme

mamccorm:
A number of updates to the naming scheme used for argo-workflows, to align with some internal processes re: version streams. The 'provides' allows for the latest version of the package to be continue to referenced using the old format.


argo-workflows-3.7.yaml
```diff
@@ -1,10 +1,13 @@
 package:
-  name: argo-workflows
+  name: argo-workflows-3.7
   version: "3.7.2"
-  epoch: 0 # GHSA-2464-8j7c-4cjm
+  epoch: 0
```
mamccorm: epoch left at 0 is intentional, as this is a new package name

kranurag7: we can bump the epoch here because images will pull this package using provides and that'll be the older package. (ref: https://apk.dag.dev/https/packages.wolfi.dev/os/x86_64/argo-workflows-3.7.2-r0.apk@sha1:00b5abc5b69e0582cdf2d4ec2c49e4353efdc3b5/.PKGINFO) 

`argo-workflows` of version `3.7.2-r0` exists in the index. I'm not sure if doing it this way will overwrite that or not.

mamccorm: ack, yea i'll bump it to be safe

EyeCantCU (approved):

PR #67869, opened 2025-10-01
fix conda25.9.0 bump

catmsred:
fix: add default channels to conda tests
    
Starting in 25.9, conda requires the user to explicitly add channels.

Relates: #67692

See [conda release notes](https://github.com/conda/conda/releases/tag/25.9.0) for details on the deprecation.

### DIFF
--- a/conda-base.yaml
+++ b/conda-base.yaml
@@ -26,8 +26,7 @@ pipeline:
       conda create --yes --name init
 
       mkdir -p "${{targets.contextdir}}"/root/conda
-      cp -r ${HOME}/conda/* ${HOME}/.conda/envs/init/* \
-        "${{targets.contextdir}}"/root/conda
+      cp -r ${HOME}/.conda/envs/init/* "${{targets.contextdir}}"/root/conda
       echo "root_prefix: /root/conda" > "${{targets.destdir}}"/root/.condarc
 
 update:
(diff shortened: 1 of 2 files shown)


IdlePhysicist:
@catmsred is this PR nearly ready to go?

catmsred:
@IdlePhysicist The upstream change to conda is breaking (previously, any script that didn't explicitly add channels would be able to install packages; now it won't) so I wouldn't want to merge this change until that's been announced.

Additional work (that could be done as separate PR) is that conda-base should be a subpackage of conda as they need to have tightly couple versioning.

IdlePhysicist:
So is it correct to say that we're blocked by the upstream at the moment?

catmsred:
> So is it correct to say that we're blocked by the upstream at the moment?

I'm not sure that's a correct framing.  Upstream has had deprecation warnings around this for some time it seems.  We are blocked on our own breaking change announcement.

catmsred:
Manual testing of version linking:
`make package/conda-base` fails on 25.9 when only 25.7 is available
`make package/conda` creates conda 25.9
`make test/conda` passes
`make package/conda-base` now works for 25.9
`make test/conda-base` passes

jamie-albert (approved):

PR #68000, opened 2025-10-03
trino: report available plugins

aborrero:
Report available plugins for easier maintenance.

### DIFF
--- a/trino.yaml
+++ b/trino.yaml
@@ -1,7 +1,7 @@
 package:
   name: trino
   version: "477"
-  epoch: 0 # GHSA-mmxm-8w33-wc4h, GHSA-3p8m-j85q-pgmj
+  epoch: 1 # GHSA-mmxm-8w33-wc4h, GHSA-3p8m-j85q-pgmj
   description: The distributed SQL query engine for big data, formerly known as PrestoSQL
   copyright:
     - license: Apache-2.0
@@ -58,7 +58,8 @@ pipeline:
       # Re-sort the pom.xml
       ./mvnw sortpom:sort -q
 
-  - runs: |
+  - name: "main package build"
+    runs: |
       set -x
 
       export JAVA_HOME=/usr/lib/jvm/java-25-openjdk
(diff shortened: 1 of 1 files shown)


trino.yaml
```diff
@@ -83,6 +84,14 @@ pipeline:
       cp ./client/trino-cli/target/trino-cli-*-executable.jar ${{targets.destdir}}/usr/bin/trino
       chmod 0755 ${{targets.destdir}}/usr/bin/trino
 
+  # NOTE: just information in case you need to update the plugin list below
+  - name: "print available plugins in the upstream source"
+    runs: |
+      items=$(ls ./core/trino-server/target/trino-server-${{package.version}}/plugin/ | sort)
```
mosabua, line 90: this is wrong .. this just reports the plugins that are by default packaged in the tarball.

mosabua (changes requested):
I think I need to explain how this works in Trino.

* All plugins use the packaging trino-plugin defined in the pom.xml
* They are all located in https://github.com/trinodb/trino/tree/master/plugin
* They all produce a binary zip archive that can be used to install them. as of 477 these can be found on github release (used to be maven central before) - links to all of them are in the docs as well
* there are two tarballs and respective docker container - trino-server-core and trino-server
* core contains a minimal set
* trino-server contains nearly all plugins
* we are driving an effort to remove more plugins from trino-server
* the trino-packages project provides means to create a custom tarball and a custom docker container based on core

So for wolfi I think we want to create apk's for each plugin that is found in the source folder

powersj:
@aborrero and @mosabua what's the plan for this PR?

mosabua:
No plan from my end. The current approach is wrong so I think its safe to close this PR and start a new one whenever we think we need to do something here ... I will probably send another PR when we release Trino 478 and we can revisit then or just keep doing the work manually .. new plugins are not added that often.

trino.yaml
mosabua: Correct would be 

`ls ./plugin/ | sort`

btw

powersj:
Thanks - I'm going to close then.

PR #68037, opened 2025-10-03
qemu - split edk2 files out to dependent subpackage

smoser:
The edk2 files are useful in some scenarios by themselves, so let the user just pull them without all of qemu and deps.

### DIFF
--- a/qemu.yaml
+++ b/qemu.yaml
@@ -1,7 +1,7 @@
 package:
   name: qemu
   version: "10.1.0"
-  epoch: 0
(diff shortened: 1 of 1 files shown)


xnox (approved):

xnox:
Yes malcontent qemu-system-x86_64 pretends to be a computer.

xnox:
This split is also useful, because if/when our edk2 package is fully formed, we can switch to only using firmware builds from edk2 package, without these prebuilt files from qemu.

PR #68099, opened 2025-10-05
Adds a number of core runtime dependencies required by jenkins-inbound-agent image

mamccorm:
Adds a number of additional dependencies required for core Jenkins operations. These are installed on the jenkins-inbound-agent image.


jamonation (dismissed):
Confirmed these are added on top of a base OS image so are functional requirements

jamonation (approved):

PR #68112, opened 2025-10-06
llvm-{20,21}: make libLLVM co-installable

javacruft:
libLLVM is always installed to version specific paths so ensure that
different major versions are always co-installable otherwise we end up
with package resolver issues in the archive as different binaries
depend on different versions of libLLVM.

### DIFF
--- a/llvm-20.yaml
+++ b/llvm-20.yaml
@@ -1,7 +1,7 @@
 package:
   name: llvm-20
   version: "20.1.8"
-  epoch: 4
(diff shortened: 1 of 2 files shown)


javacruft:
Nothing directly depends on libLLVM (and TBH it should never be used in this way).

sil2100 (approved):
Indeed libLLVM doesn't seem to be used. If it breaks anything, we'll see it rather quick!

PR #68117, opened 2025-10-06
orthanc-{gdcm,python}: Pin boost-dev~1.88

javacruft:
Align with core orthanc package and pin boost version to 1.88 until
upstream supports >= 1.89.


sil2100 (approved):
+1, I indeed see orthanc (and all the other orthanc-* packages) pinning to 1.88.

PR #68138, opened 2025-10-06
ensure older grpc versions remain guarded

javacruft:
grpc <= 1.70 requires protobuf-dev=3.29.5; however this version of the
package is old/unguarded and not installable any longer.

Introduce a new protobuf-25.9 (using versioning aligned to current
protobuf package) along versioned subpackages, using the SOVERSION for
all of the shared library packages.

protobuf-25.9 and grpc <= 1.70 also need an older version of abseil-cpp.

Introduce a new abseil-cpp-20250127 package along with versioned subpackages
using the SOVERSION for the shared library package ensuring co-installabilty
with newer versions.

Update grpc <= 1.70, falco and libpulsar to use new versioned packages.


sil2100 (commented):
This looks good, just trying to understand the rationale for using <major.minor.0> as the library naming scheme. In general we're rather only keying on the major revision, right? At most major.minor should be enough? Can you tell me more?

javacruft:
> This looks good, just trying to understand the rationale for using <major.minor.0> as the library naming scheme. In general we're rather only keying on the major revision, right? At most major.minor should be enough? Can you tell me more?

It all rather depends on the project being packaged and the soversion used in the projects build:

https://apk.chaindag.dev/https/apk.cgr.dev/wolfi-presubmit/e551a581757af39b50d92ff609f68e57271b5bce/aarch64/libprotobuf-lite29.5.0-29.5-r0.apk@etag:4d12f38e89dd71110406cf2f1c715b28/.PKGINFO

I *think* I have this right for protobuf - but ultimately so long as we generate a unique package name we're OK

grpc-1.68.yaml
```diff
@@ -156,11 +152,10 @@ subpackages:
...
     dependencies:
       runtime:
-        - abseil-cpp-dev=20250127.1-r4
-        - abseil-cpp=20250127.1-r4
+        - abseil-cpp-20250127-dev
         - ${{package.name}}
-        - re2=2025.07.17-r1
-        - re2-dev=2025.07.17-r1
+        - re2-dev
+        - re2-dev
```
justinvreeland: nit: duplicate re2-dec

justinvreeland (dismissed):
Just a small change to remove the duplicate re2-dev lines. We don't need to include re2 since it should(?) resolve correctly without the conflicts on abseil.

grpc-1.66.yaml
```diff
@@ -140,12 +136,11 @@ subpackages:
...
       runtime:
-        - abseil-cpp-dev=20250127.1-r4
-        - abseil-cpp=20250127.1-r4
+        - abseil-cpp-20250127-dev
         - ${{package.name}}
         - opentelemetry-cpp-dev
-        - re2=2025.07.17-r1
-        - re2-dev=2025.07.17-r1
+        - re2-dev
+        - re2-dev
```
sil2100: Duplicate, remove please!

grpc-1.67.yaml
```diff
@@ -156,12 +152,11 @@ subpackages:
...
       runtime:
-        - abseil-cpp-dev=20250127.1-r4
-        - abseil-cpp=20250127.1-r4
+        - abseil-cpp-20250127-dev
         - ${{package.name}}
         - opentelemetry-cpp-dev
-        - re2=2025.07.17-r1
-        - re2-dev=2025.07.17-r1
+        - re2-dev
+        - re2-dev
```
sil2100: Same here ^

grpc-1.69.yaml
```diff
@@ -156,12 +152,11 @@ subpackages:
...
       runtime:
-        - abseil-cpp-dev=20250127.1-r4
-        - abseil-cpp=20250127.1-r4
+        - abseil-cpp-20250127-dev
         - ${{package.name}}
         - opentelemetry-cpp-dev
-        - re2=2025.07.17-r1
-        - re2-dev=2025.07.17-r1
+        - re2-dev
+        - re2-dev
```
sil2100: Same here ^

grpc-1.70.yaml
```diff
@@ -156,11 +152,10 @@ subpackages:
...
     dependencies:
       runtime:
-        - abseil-cpp-dev=20250127.1-r4
-        - abseil-cpp=20250127.1-r4
+        - abseil-cpp-20250127-dev
         - ${{package.name}}
-        - re2=2025.07.17-r1
-        - re2-dev=2025.07.17-r1
+        - re2-dev
+        - re2-dev
```
sil2100: Same here ^

sil2100 (changes requested):
This looks good but there's still some `re2-dev` duplicated lines. Once those are cleaned up, we're good to merge.

sil2100 (approved):
LGTM now!

PR #68177, opened 2025-10-06
added some tests for the aznfs-mount package

dustinkirkland:
This package was added recently without any tests.

Testing it as best as we can -- check that the critical files are installed, executable, and pass a syntax check.  Try running the command (which previously complained about a missing directory and log file, which we fix with an epoch bump).

### DIFF
--- a/aznfs-mount.yaml
+++ b/aznfs-mount.yaml
@@ -1,7 +1,7 @@
 package:
   name: aznfs-mount
   version: "2.0.12"
-  epoch: 3
+  epoch: 4
   description: AZNFS Mount Helper
   copyright:
     - license: Apache-2.0
@@ -67,10 +67,3 @@ update:
   github:
     identifier: Azure/AZNFS-mount
     use-tag: true
-
-# test added by a robot (binary)
-test:
-  pipeline:
-    - runs: |
-        stat /usr/bin/aznfswatchdog
-        stat /usr/bin/mount.aznfs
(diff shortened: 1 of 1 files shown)


aznfs-mount.yaml
```diff
@@ -35,6 +35,32 @@ pipeline:
       install -Dm755 src/mountscript.sh "${{targets.contextdir}}"/opt/microsoft/aznfs/mountscript.sh
       install -Dm755 scripts/aznfs_install.sh "${{targets.contextdir}}"/opt/microsoft/aznfs/aznfs_install.sh
       install -Dm755 lib/common.sh "${{targets.contextdir}}"/opt/microsoft/aznfs/common.sh
+      mkdir -p "${{targets.contextdir}}"/opt/microsoft/aznfs/data
+      touch "${{targets.contextdir}}"/opt/microsoft/aznfs/data/aznfs.log
```
dannf, line 39: I know nothing about this package - but I'd guess based on the directory structure that `data` might be intended to be mounted into the container from the host. In which case, there's a chance that having this in the package itself might confuse something. @developer-guy would you mind validating this in the context of the image? (And if it happens to be a problem to create this inside the package, maybe we should add a test to confirm that we did *not* do so).

dannf (dismissed):
LGTM - just one point of caution inline

aznfs-mount.yaml
dustinkirkland: Good point....  Let me know which way we should go....

developer-guy (dismissed):

aborrero:
will make the linter happy and submit myself.

AmberArcadia (approved):

PR #68190, opened 2025-10-07
Add v18 for postgres dependent packages

IdlePhysicist:
- **postgis-18: Add pkg, remove 17**
  Here I add postgis-18 and remove 17.
  

- **repmgr-18: Add pkg, remove 17**
  Here I add repmgr-18 and remove 17


aborrero (approved):
LGTM.

PR #68226, opened 2025-10-07
abseil-cpp: make libraries co-installable

javacruft:
Move shared libraries to a new soversion'ed subpackage to ensure that
this package is co-installable with the previous version, ensuring
smooth transitions.


justinvreeland (approved):
it looks like the .so files are only using the first 4 numbers `2508` which is annoying but shouldn't be an issue.

PR #68291, opened 2025-10-07
DNM - Add Python 3.14 without unversioned "provides"

Taffer:
DNM due to potential need to clean up patches, and we want to provide a GIL-less version as well.


Taffer:
Actually, I can test this locally, sorry about that.

PR #68297, opened 2025-10-07
Add Python 3.14

Taffer:
Add Python 3.14 *without* an unversioned "provides" so we can more carefully integrate the changes.

This version does *not* have the experimental free-threaded JIT enabled, that will come separately in a new package.

### DIFF
--- a/python-3.14.yaml
+++ b/python-3.14.yaml
@@ -1,0 +1,329 @@
+package:
+  name: python-3.14
+  version: "3.14.0"
+  epoch: 0
+  description: "the Python programming language"
+  copyright:
+    - license: PSF-2.0
+  resources
+  # We're leaving this out of providing python3; we'll test build critical
+  # packages separately to ensure they're ready for Python 3.14.
+  #
+  # dependencies:
+  #   provides:
+  #     - python3=${{package.full-version}}
+  #     - python-3=${{package.full-version}}
+  #   runtime:
+  #     - ${{package.name}}-base=${{package.full-version}}:
+    cpu: 8
+    memory: 8Gi
+
+environment:
+  contents:
+    packages:
+      - build-base
+      - busybox
+      - bzip2-dev
+      - ca-certificates-bundle
+      - expat-dev
+      - gdbm-dev
+      - libcap-utils
+      - libffi-dev
+      - libx11-dev
+      - linux-headers
+      - mpdecimal-dev
+      - ncurses-dev
+      - openssl-dev
+      - readline-dev
+      - sqlite-dev
+      - tcl-dev
+      - tk-dev
+      - xz-dev
+      - zlib-dev
+
+# creates helpful python3.M and 3.M variables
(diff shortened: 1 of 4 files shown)


python-3.14.yaml
```diff
@@ -0,0 +1,324 @@
...
+          cp "${{targets.outdir}}/${{package.name}}-base/usr/bin/${{vars.python}}" "${{targets.contextdir}}/usr/bin/"
+          # Note this sets the capability the test using getcap fails but the
+          # capability is set in images with this package.
+          setcap cap_net_bind_service=+eip "${{targets.contextdir}}/usr/bin/${{vars.python}}"
+
+  - name: "${{package.name}}-tk"
+    dependencies:
+      provides:
+        - py3-tkinter=${{package.full-version}}
+        - py3.13-tkinter=${{package.full-version}}
```
sergiodj:
```suggestion
        - py${{vars.pyversion}}-tkinter=${{package.full-version}}
```

sergiodj (changes requested):
Not really an in-depth review, but there's a small nit on the `python-tk` subpackage.

python-3.14.yaml
```diff
@@ -0,0 +1,324 @@
...
+          mkdir -p "${{targets.contextdir}}/usr/bin"
+          cp "${{targets.outdir}}/${{package.name}}-base/usr/bin/${{vars.python}}" "${{targets.contextdir}}/usr/bin/"
+          # Note this sets the capability the test using getcap fails but the
+          # capability is set in images with this package.
+          setcap cap_net_bind_service=+eip "${{targets.contextdir}}/usr/bin/${{vars.python}}"
+
+  - name: "${{package.name}}-tk"
+    dependencies:
+      provides:
+        - py3-tkinter=${{package.full-version}}
```
kranurag7: lately we've not been using `py3-<foo>` provides so probably we can omit this. Ideally we're pinning/using `py<x.y>-foo`

python-3.14.yaml
Taffer: Ooh, good catch, these `provides` should be  removed for 3.14.

Taffer:
> @dannf looking into our internal messaging around adding `py3-` builds with 3.14.

We are "GO" to merge.

justinvreeland:
> This version does not have the experimental free-threaded JIT enabled, that will come separately in a new package.

Can you expand on this? Is it a second interpreter or some kind of plugin?

Taffer:
> Can you expand on this? Is it a second interpreter or some kind of plugin?

It's a different build of the interpreter; I was intending on adding a new rev (err, `epoch`) for this to enable it *when requested* via the `PYTHON_JIT=1` environment variable.

I would have included it in this one but was having difficulty getting it to build, and thought getting this in the repo was a higher prio.

Taffer:
Sniped by https://github.com/wolfi-dev/os/pull/68454

PR #68328, opened 2025-10-08
Garbage Collection - 8th October 2025

javacruft:
Withdraw packages older that 12 months that are no longer in use in any
images, vms and builds and have no other reverse dependencies.

Validated using dependency analysis and withdrawal testing in stereo.


sil2100 (approved):
List seems sane.

PR #68343, opened 2025-10-08
py3-gobject3: enable auto updates

aborrero:
There were disabled for unknown reasons.

Checking manually, they seem to work once enabled.

See also: https://github.com/chainguard-dev/internal-dev/issues/13693


aborrero:
PUCC is precisely the thing we are fixing here. My hope is that with this PR automation will send next a PR updating the version of the package.

AmberArcadia (approved):

PR #68344, opened 2025-10-08
py3-mwoauth: update to version 0.3.0

aborrero:
And enable auto-updates.


OddBloke:
Looks like we're missing a dependency on `six`.

dnegreira (approved):

PR #68345, opened 2025-10-08
py3-onetimepass: retire package

aborrero:
Package is abandoned upstream, and is not a real dependency of anything,
including chainguard images.


aborrero:
This is wrong, the package is still required, see https://github.com/jupyterhub/nativeauthenticator/blob/bdc603f964ec54a87d3b687069fc410df3faf614/setup.py#L32

PR #68348, opened 2025-10-08
clang-18-dev -> clang-18

javacruft:
clang-18-dev was a provides in older clang-18 package versions; this is
no longer required and is keeping an older clang-18 package version
pinned into the archive.

### DIFF
--- a/calico-3.30.yaml
+++ b/calico-3.30.yaml
@@ -1,7 +1,7 @@
 package:
   name: calico-3.30
   version: "3.30.3"
-  epoch: 3 # GHSA-2464-8j7c-4cjm
+  epoch: 3
   description: "Cloud native networking and network security"
   copyright:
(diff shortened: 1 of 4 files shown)


calico-3.30.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: calico-3.30
   version: "3.30.3"
-  epoch: 2 # GHSA-2464-8j7c-4cjm
+  epoch: 3 # GHSA-2464-8j7c-4cjm
```
sergiodj: I make a point of removing the comment after the epoch when I bump it, otherwise it can give the impression that this new epoch fixes the security issue referenced in the comment.

sergiodj (dismissed):

sergiodj (approved):

PR #68400, opened 2025-10-08
Switch multiple packages to using tw's help-check

murraybd:
`fsck.zfs --help` didn't actually do anything and caused the help-check test to fail, so I dropped it.

### DIFF
--- a/zizmor.yaml
+++ b/zizmor.yaml
@@ -1,7 +1,7 @@
 package:
   name: zizmor
   version: "1.14.2"
-  epoch: 0 # GHSA-xwfj-jgwm-7wp5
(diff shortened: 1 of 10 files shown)


zizmor.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: zizmor
   version: "1.14.2"
-  epoch: 0 # GHSA-xwfj-jgwm-7wp5
+  epoch: 1 # GHSA-xwfj-jgwm-7wp5
```
sergiodj: Would it be possible to remove this and other instances of comments after epoch bumps?

sergiodj (changes requested):

sergiodj (approved):

PR #68466, opened 2025-10-09
fix(flannel): bypass br-netfilter check for aarch64 CI tests

debasishbsws:
This patch adds a fix for the aarch64 test failures by bypassing the upstream br_netfilter check in Flannel's main.go, which causes issues in our CI Docker runner.

Flannel performs a runtime check to verify if the br_netfilter kernel module is loaded: https://github.com/flannel-io/flannel/blob/c2afcd3b6892fe0e08ae0c7631bb752a335472da/main.go#L274-L288

While this check passes on x86 (QEMU-based) runners, it fails on aarch64 where we use Docker-based runners. The log message shows:
  'main.go:278] Failed to check br_netfilter: stat /proc/sys/net/bridge/bridge-nf-call-iptables: no such file or directory'

This seems to be related to kernel config availability, as the /proc/sys/net/bridge directory is missing in aarch64 CI runners. Interestingly, running the same Docker-based build locally on macOS (aarch64) works fine.

Since this issue also exists in the last release (see: https://github.com/wolfi-dev/os/pull/67940), this patch bypasses the check *only* for aarch64 to allow CI to continue and test the rest of the functionality across architectures.

Note: Local testing on aarch64 Mac systems passes, suggesting this is specific to the CI environment with some recent changes.

### DIFF
--- a/flannel.yaml
+++ b/flannel.yaml
@@ -1,6 +1,6 @@
 package:
   name: flannel
-  version: "0.27.3"
+  version: "0.27.4"
   epoch: 0 # CVE-2025-47907
   description: flannel is a network fabric for containers, designed for Kubernetes
   copyright:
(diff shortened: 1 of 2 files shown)


aborrero (approved):
LGTM.

aborrero:
I think the patch is worth submitting upstream for evaluation.

PR #68516, opened 2025-10-09
Remove Python 3.14 "provides" and update CODEOWNERS

Taffer:
Python 3.14 should not be the provider for `python` or `python3` at this time.


a-crate:
I think you also need to withdraw existing packages (and maybe bump epoch for new builds) to stop solver conflicts.

a-crate (approved):

PR #68548, opened 2025-10-10
fix(falco-no-driver): add missing container plugin

maxgio92:
The container plugin is a required plugin, so it's required also to test the falco-no-driver package.
This PR adds the plugin as a subpackage, and make it a runtime dependency of falco-no-driver.
Furthremore, it adds tests since we have the ability to run Falco now that we have the plugin.
With the occasion, it refactors the build pipeline and aligns it to upstream.


falco-no-driver.yaml
```diff
@@ -60,45 +64,148 @@ pipeline:
...
   pipeline:
-    # AUTOGENERATED
-    - runs: |
-        falco --version
+    - name: Smoke test
+      runs: |
         falco --help
+    - name: "Verify falco version"
+      runs: |
+        falco --version 2>/dev/null | grep -E "^Falco\ version" | cut -d ':' -f2 | tr -d ' ' | grep -E "^${{package.version}}$"
```
murraybd: It'd probably be a lot easier to use tw's ver-check test.

maxgio92: Good catch! Thanks. I've Just replaced, PTAL

falco-no-driver.yaml
```diff
@@ -15,20 +16,23 @@ package:
...
       - bpftool
       - build-base
       - busybox
       - c-ares-dev
+      - ca-certificates-bundle
       - clang-${{vars.llvm-vers}}
+      - clang-${{vars.llvm-vers}}-extras
       - cmake
       - cpp-httplib
+      - curl
```
mamccorm: Is curl actually needed to produce the main package, or just part of the tests?

maxgio92: Nice catch @mamccorm. We actually just need curl's header files at build time, so I've removed curl from build and test time's deps.

falco-no-driver.yaml
```diff
@@ -60,45 +63,148 @@ pipeline:
       expected-commit: d8e430e352396abde7b4309c76039f08042e2512
       recurse-submodules: true
 
-  - runs: |
-      mkdir skeleton-build && cd skeleton-build
-      cmake -DUSE_BUNDLED_DEPS=ON -DBUILD_FALCO_MODERN_BPF=ON -DCREATE_TEST_TARGETS=Off -DFALCO_VERSION=${{package.version}} ..
-      make ProbeSkeleton -j$(nproc)
+  - name: Configure modern BPF probe skeleton
```
utieyin, line 68: We could use the cmake pipeline here

```suggestion
  - name: Configure modern BPF probe skeleton
     uses: cmake/configure
       with:
         opts: ----
     
  
```

utieyin: We could also do same for the build with cmake/build and set the outdir to `skeleton-build`

maxgio92: Thanks @utieyin, done, PTAL

maxgio92: Keep in mind that we need https://github.com/chainguard-dev/melange/pull/2211 to use `cmake/build` for building the BPF probe and Falco

utieyin (approved):
LGTM

PR #68550, opened 2025-10-10
add 3.14 support for py3-pip

kranurag7:

### DIFF
--- a/py3-pip.yaml
+++ b/py3-pip.yaml
@@ -5,8 +5,6 @@ package:
   description: The PyPA recommended tool for installing Python packages.
   copyright:
     - license: MIT
-  dependencies:
-    provider-priority: 0
 
 vars:
   pypi-package: pip
@@ -68,7 +66,6 @@ subpackages:
     dependencies:
       runtime:
         - py${{range.key}}-${{vars.pypi-package}}-base
-      provider-priority: ${{range.value}}
     pipeline:
       - runs: |
           mkdir -p ${{targets.contextdir}}/usr/bin
--- a/py3-pip.yaml
+++ b/py3-pip.yaml
@@ -68,8 +68,6 @@ subpackages:
     dependencies:
       runtime:
         - py${{range.key}}-${{vars.pypi-package}}-base
-      provides:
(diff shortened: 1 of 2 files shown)


Taffer (dismissed):
Assuming this just makes the 3.14 versions of `pip` and `setuptools` available, I think this is right. How's that for an endorsement? 🤪

py3-pip.yaml
```diff
@@ -18,13 +16,15 @@ data:
       3.11: "311"
       3.12: "312"
       3.13: "313"
+      3.14: "314"
```
dannf: the values here can all be deleted now i think

py3-pip.yaml
```diff
@@ -18,13 +16,15 @@ data:
...
 
 environment:
   contents:
     packages:
       - build-base
       - busybox
       - py3-supported-python
+      - python-3.14
```
dannf: I think this should just be python-3.14-base (we don't need `/usr/bin/python3` to be 3.14)

py3-pip.yaml
```diff
@@ -66,9 +66,6 @@ subpackages:
     dependencies:
       runtime:
         - py${{range.key}}-${{vars.pypi-package}}-base
-      provides:
-        - py3-${{vars.pypi-package}}
-      provider-priority: ${{range.value}}
```
dannf, line 71: Have you verified that there are no remaining consumers of `py3-pip` ?

py3-setuptools.yaml
dannf, line 54: while you're in here - would you mind adding the pip-check pipeline?

py3-pip.yaml
kranurag7: some of them are consuming it implicitly at images (meaning they adding py3-foo and py3-foo consuming py3-pip-wheel at runtime) but that number is less and we will be pinning those cases.

bentasker (approved):

javacruft (approved):

py3-pip.yaml
smoser: related https://github.com/wolfi-dev/os/pull/71841

PR #68554, opened 2025-10-10
Adding entrypoint subpackage for official curl image

rahulduvedi:
Minor Fix: Entrypoint instructions for our curl image


AnmolVirdi (approved):
LGTM. Upstream uses entrypoint script to run curl.

PR #68592, opened 2025-10-10
abseil-cpp: export cmake targets

aborrero:
Use this option to export cmake targets.

These new cmake targets can be used by other dependent packages when building their code.

We have detected the need for this while working on the onnxruntime suite of packages.


aborrero:
I don't think we need this after all, comparing the files usr/lib/cmake/absl)abslTargets.cmake pre/post this patch, they are the same.

PR #68594, opened 2025-10-10
clamav-1.4: version-stream subpackages

OddBloke:
So we can introduce `clamav-1.5`.

### DIFF
--- a/clamav-1.4.yaml
+++ b/clamav-1.4.yaml
@@ -1,7 +1,7 @@
 package:
   name: clamav-1.4
   version: "1.4.3"
-  epoch: 2
+  epoch: 3
   description: An anti-virus toolkit for UNIX eis-ng backport
   copyright:
     - license: GPL-2.0-only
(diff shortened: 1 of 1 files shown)


Taffer (approved):
Looks correct to me, but I don't know what implications arise from having version-streamed sub-packages...

dannf (commented):
I rarely create version streams myself. But when I do, I do it like this. The only odd bit to me is `freshclam` becoming `clamav-1.4-freshclam` instead of `freshclam-1.4`. The latter feels nicer because image definitions that have both mainline and version-streamed builds could just have naive "add suffix if version streamed" logic instead of special casing this one. Of course, that likely requires adding a var-transform to extract the version.

javacruft (approved):

javacruft:
> I rarely create version streams myself. But when I do, I do it like this. The only odd bit to me is `freshclam` becoming `clamav-1.4-freshclam` instead of `freshclam-1.4`. The latter feels nicer because image definitions that have both mainline and version-streamed builds could just have naive "add suffix if version streamed" logic instead of special casing this one. Of course, that likely requires adding a var-transform to extract the version.

I think having the base package name included as the prefix is OK as it still provides "freshclam"

PR #68614, opened 2025-10-10
Build Python 3.14 with free-threaded JIT

Taffer:
This adds the experimental free-threaded JIT to Python 3.14. It's off by default; people wanting to use it must export the `PYTHON_JIT=1` environment variable to enable it.

### DIFF
--- a/python-3.14.yaml
+++ b/python-3.14.yaml
@@ -19,6 +19,7 @@ environment:
       - busybox
       - bzip2-dev
       - ca-certificates-bundle
+      # This is a year old. Hopefully 3.14.x updates will move this to current.
       - clang-19
       - expat-dev
       - gdbm-dev
@@ -26,6 +27,7 @@ environment:
       - libffi-dev
       - libx11-dev
       - linux-headers
+      # This is a year old. Hopefully 3.14.x updates will move this to current.
       - llvm-19
       - mpdecimal-dev
       - ncurses-dev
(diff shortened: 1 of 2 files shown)


python-3.14.yaml
```diff
@@ -281,7 +328,20 @@ subpackages:
...
+      runs: |
+        # main package should provide 'python' and 'python3'.
+        python version-check.py ${{package.version}}
+        python3 version-check.py ${{package.version}}
+        pydoc3 --version
+        pydoc3 --help
+        python --version
+        python --help
+        python3 --version
+        python3 --help
```
murraybd: These (pydoc, python and python3) could use tw's ver-check and help-check pipelines.

Taffer: Right you are, I'll update those!

python-3.14.yaml
```diff
@@ -19,15 +19,18 @@ environment:
...
       - bzip2-dev
       - ca-certificates-bundle
+      - clang-19
       - expat-dev
       - gdbm-dev
       - libcap-utils
       - libffi-dev
       - libx11-dev
       - linux-headers
+      - llvm-19
```
justinvreeland, line 31: these seems abit old for a feature that came out this year. Any idea why it uses 19?

python-3.14.yaml
```diff
@@ -19,15 +19,18 @@ environment:
...
       - mpdecimal-dev
       - ncurses-dev
       - openssl-dev
+      - python3
```
justinvreeland: Don't use unversioned python but i also don't like adding an explict python dep to our python. Does this need to be the same 3.x version to build the jit?

justinvreeland (changes requested):
I don't think we need to replicate all the tests with the jit on we certainly don't need to do it for the python package name install test.   It'd be nice to have a test that verifies that the jit is on though.

python-3.14.yaml
Taffer: I'll test it, but it was specifically yelling for `clang-19` and I assumed.

python-3.14.yaml
Taffer: It requires a Python 3.x to exist; `python3` is supposed to be the latest stable Python 3.x, isn't it?

The JIT version isn't a separate build artifact, so we can't (shouldn't?) build/use a non-JIT version from this code bundle.

python-3.14.yaml
Taffer: Yup, 19 *specifically*. Hopefully they can make it less version-specific in 3.14.x updates.

Removed a bunch of redundant tests and added a specific JIT test.

python-3.14.yaml
Taffer: I'll switch this to 3.13 and add a comment about it being needed for enabling the JIT.

python-3.14.yaml
justinvreeland: Not uncommon with llvm tbh just curious
a pep about this for future delvers.
https://peps.python.org/pep-0774/

python-3.14.yaml
justinvreeland: > It requires a Python 3.x to exist; python3 is supposed to be the latest stable Python 3.x, isn't it?

it is but I don't like using this in builds because it's a moving target.  It also sets us up to withdraw python3.X-1 and then have a bootstrapping issue.

justinvreeland (approved):
Really dislike that we just drop those files into the build directory but we can't do anything about that.

PR #68884, opened 2025-10-14
fix(kaniko): patch usrmerge symlinks

sanghanan:
Fixes: https://github.com/chainguard-dev/customer-issues/issues/2579
This patch removes the symlinks between `/bin`, `/sbin` and `/usr/bin`,  `lib`, `lib64` and `/usr/lib`. It creates directories in their place. This doesn't fix any of the underlying issues with how kaniko resolves symlinks, and is just a wolfi-specific short term patch.


tulilirockz (dismissed):
This looks good to me: Context, this pull request is meant to make it so that kaniko can have a clean rootfs so that it can extract `FROM (image)`to `/` without breaking due to usr-merge symlinks. This is a patchwork fix that should be eventually cleaned up by allowing the terraform providers to _not_ include `wolfi-baselayout` or any folder outside of the specified ones.

tulilirockz (approved):

PR #68885, opened 2025-10-14
Use make check and switch to tw's help-check and ver-check

murraybd:


attr.yaml
```diff
@@ -96,9 +99,10 @@ update:
 test:
   pipeline:
     # AUTOGENERATED
```
justinvreeland: Get rid of this comment.

justinvreeland (changes requested):

justinvreeland (approved):

PR #68887, opened 2025-10-15
Build xz via git-checkout rather than fetch

egibs:
This PR builds `xz` via `git-checkout` instead of `fetch` since the former is our accepted practice when possible. The circular dependency comment is interesting and likely outdated since I was able to build and test locally without issue.

The following languages are excluded when using `--no-po4a`:
```
[po4a_langs] de fr it ko pt_BR ro sr sv uk
```
> code for [reference](https://github.com/tukaani-project/xz/blob/8d26b72915e0d373f898b55935505857c30dbdb3/po4a/po4a.conf#L7)

Which explains the `ci-diff-report` removals:
image

### DIFF
--- a/xz.yaml
+++ b/xz.yaml
@@ -1,7 +1,7 @@
 package:
   name: xz
   version: "5.8.1"
-  epoch: 5
+  epoch: 6
   description: "Library and CLI tools for XZ and LZMA compressed files"
   copyright:
     - license: GPL-3.0-or-later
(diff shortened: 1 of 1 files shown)


justinvreeland (commented):
Seems fine. I believe we have a configure pipeline. we should make a autoconf/autogen/autoreconf pipeline if we don’t have one.

justinvreeland (approved):

PR #68895, opened 2025-10-15
vector: apply patch for FIPS-compatibility

sil2100:
Internal use of PKCS12 is causing issues when trying to run in fips-only mode. Upstream has a PR up that changes that, which we should pull in here to still be able to use vector in our FIPS specific images. Hopefully the PR will be reviewed and merged soon, with us being able to either switch to a git cherry-pick or simply drop it altogether with the next vector release.

### DIFF
--- a/vector.yaml
+++ b/vector.yaml
@@ -1,7 +1,7 @@
 package:
   name: vector
   version: "0.50.0"
-  epoch: 0
+  epoch: 1
   description: End-to-end observability data pipeline
   copyright:
     - license: MPL-2.0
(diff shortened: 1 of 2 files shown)


jamonation:
Tested with a pem key/cert and looks good here. I can hit the https interface on 127.0.0.1:9999
```
openssl s_client -connect 127.0.0.1:9999
. . .
No client certificate CA names sent
Peer signing digest: SHA256
Peer signature type: rsa_pss_rsae_sha256
Peer Temp Key: ECDH, prime256v1, 256 bits
---
SSL handshake has read 2435 bytes and written 377 bytes
Verification error: self-signed certificate
---
New, TLSv1.2, Cipher is ECDHE-RSA-AES128-GCM-SHA256
Protocol: TLSv1.2
Server public key is 4096 bit
Secure Renegotiation IS supported
No ALPN negotiated
SSL-Session:
    Protocol  : TLSv1.2
    Cipher    : ECDHE-RSA-AES128-GCM-SHA256
```

jamonation (approved):
Would be good to set up a key/cert and test in the package at some point now that we've got this patch in place

sil2100:
+1 on the test! However, I think we can do that in a separate PR (flooded with interrupts right now!).

PR #68974, opened 2025-10-15
go-1.25: introduce additional tests

aborrero:


aborrero:
bogus test. The package is not empty.

PR #69234, opened 2025-10-16
py3-python-pypi-mirror: switch to git update backend

OddBloke:
We need to look at tags, not releases.


OddBloke:
This is an `update`-only change, no epoch bump required.

AmberArcadia (approved):

PR #69277, opened 2025-10-16
fix(airflow-3): Convert to venv

EyeCantCU:
There's more to fix here, but this is a start. All of the `pip upgrade`s are bumping other dependencies in the environment (and potentially even downgrading dependencies in some instances)

This resolves a few problems:
- Creates a virtual environment so that providers are easier to install
- Avoids completely reinstalling airflow (potentially multiple times) when providers that list it as a dep are installed
- Stop removing pycache (airflow will start faster at the expense of size)

In the future we'll want to lock dependencies in the project before bumping dependencies that contain vulnerabilities, but this will do for now

### DIFF
--- a/airflow-3.yaml
+++ b/airflow-3.yaml
@@ -1,7 +1,7 @@
 package:
   name: airflow-3
   version: "3.1.0"
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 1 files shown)


mritunjaysharma394 (approved):
brilliant and neat, thanks for the improvements here!

PR #69278, opened 2025-10-16
ssh-import-id - split into multi-version.

smoser:
This was depending on "python3" any old python, but would not work as it used a library outside of stdlib (py3-distro). So if py3-distro came from a different version of python than 'python3' got, the py3-distro would not be available to /usr/bin/ssh-import-id

Also drop wget, which just isnt' used.

### DIFF
--- a/ssh-import-id.yaml
+++ b/ssh-import-id.yaml
@@ -1,18 +1,22 @@
 package:
   name: ssh-import-id
   version: 5.11
-  epoch: 3
+  epoch: 4
   description: securely retrieve an SSH public key and install it locally
   copyright:
     - license: GPL-3.0-or-later
-  dependencies:
-    runtime:
-      - ca-certificates-bundle
-      - openssh-client
-      - openssh-server
-      - py3-distro
-      - python3
-      - wget
+
+vars:
+  pypi-package: ssh-import-id
+  import: ssh_import_id
+
+data:
+  - name: py-versions
+    items:
+      3.10: '310'
+      3.11: '311'
+      3.12: '312'
+      3.13: '313'
 
 environment:
   contents:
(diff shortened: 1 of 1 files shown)


smoser:
I came here because cloud-init has a specific python version (3.13 currently) and depends on ssh-import-id, but ssh-import-id might pull in a different python which would get 2 pythons in an image.

ssh-import-id.yaml
```diff
@@ -32,24 +34,61 @@ pipeline:
...
-    runs: python setup.py install --prefix=/usr --root="${{targets.destdir}}"
+  - range: py-versions
+    name: py${{range.key}}-${{vars.pypi-package}}-bin
+    description: Executable binaries for ${{vars.pypi-package}} installed for python${{range.key}}
+    dependencies:
+      provider-priority: ${{range.value}}
+      runtime:
+        - py${{range.key}}-${{vars.pypi-package}}
+      provides:
+        - ssh-import-id=${{package.full-version}}
```
smoser, line 67: This line is the one I have most question about.

justinvreeland: This all should be fine. But providing the unversioned py3 package is what gets us in trouble. We'd be better off not providing it and requiring people to specify a specific bin version. This could potentially leave you with a split versions of python still if cloud-init is updated to 3.14 before this or different packages pull in different versions of python.

justinvreeland (approved):
Any particular questions?

ssh-import-id.yaml
smoser: Right. so cloud-init should not use 'ssh-import-id' as a dep. It should declare the dep on 'py3.13-ssh-import-id'.

Do you want to drpo the ssh-iport-id entirely?  we can do that i guess we can, but we break anything that was previously using `ssh-import-id`.  This makes the old pattern still work while things can explicitly do the right thing.

smoser: thanks.

PR #69311, opened 2025-10-17
libcmis: drop doc subpackage

aborrero:
It is empty, we don't generate docs on purpose, because we are missing
a dependency.


powersj:
@aborrero build failed, un-assigning sustaining team

powersj (approved):

PR #69337, opened 2025-10-17
Seaweedfs: Install OCI entrypoint at root location and add packages

rahulduvedi:
Fixes: 
1. dash-binsh is required on the image level, helm installation is failing without it 
check here : https://github.com/seaweedfs/seaweedfs/blob/master/k8s/charts/seaweedfs/templates/master/master-statefulset.yaml#L124
2. entrypoint is expected at root by the docker file 
check here: https://github.com/seaweedfs/seaweedfs/blob/master/docker/Dockerfile.e2e#L41

### DIFF
--- a/seaweedfs.yaml
+++ b/seaweedfs.yaml
@@ -1,10 +1,13 @@
 package:
   name: seaweedfs
   version: "3.97"
-  epoch: 1 # GHSA-2464-8j7c-4cjm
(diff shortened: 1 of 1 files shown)


developer-guy (approved):
lgtm

developer-guy (dismissed):
lgtm

kranurag7 (approved):

PR #69339, opened 2025-10-17
librdkafka: ignore -DEV1 releases

OddBloke:
Closes: #69303


OddBloke:
This is an `update`-only change, so no epoch bump is necessary.

AmberArcadia (approved):

PR #69357, opened 2025-10-17
Builds PEAR as a subpackage of php

srishtih:
Builds PEAR package as a sub-package of php-8.x. 
References: https://github.com/wolfi-dev/os/pull/63748/files


srishtih:
Considered a separate build for pear, but there's nothing being built here. The php project does a much cleaner job of bundling all the dependencies, so going with this approach.

kranurag7 (commented):
looks good, thanks for the work and adding notes, I will consider removing provides in here considering the fact that we are not multi-versioning across php versions meaning we'll never be able to guarantee that pulling in php-pear will give you latest. Down the road it'll lead to confusion, ack on other php packages doing it but that had lead to confusions in past.

php-8.4.yaml
```diff
@@ -434,6 +439,44 @@ subpackages:
...
+          mkdir -p ${{targets.subpkgdir}}/usr/share/php
+
+          # Move PEAR executables
+          mv ${{targets.destdir}}/usr/bin/pear ${{targets.subpkgdir}}/usr/bin/
+          mv ${{targets.destdir}}/usr/bin/peardev ${{targets.subpkgdir}}/usr/bin/
+          mv ${{targets.destdir}}/usr/bin/pecl ${{targets.subpkgdir}}/usr/bin/
+
+          # Move PEAR libraries
+          if [ -d "${{targets.destdir}}/usr/share/php" ]; then
+            mv ${{targets.destdir}}/usr/share/php/* ${{targets.subpkgdir}}/usr/share/php/
```
kranurag7:
```suggestion
          if [ -d "${{targets.destdir}}/usr/share/php/pear" ]; then
            mv ${{targets.destdir}}/usr/share/php/pear ${{targets.subpkgdir}}/usr/share/php/
```

kranurag7: see a [html file](https://apk.chaindag.dev/https/apk.cgr.dev/chainguard/x86_64/php-8.4-8.4.14-r0.apk@sha1:d6a8b56ad10432788e0246ec3dce73ef60aa654c/usr/share/php/fpm/status.html) moving from main to this sub-package (maybe it's good but this we'll be backwards compatible and will avoid any whoops)

srishtih: nice catch.

kranurag7 (approved):

PR #69358, opened 2025-10-17
chore(daemon-check-output): remove leading/trailing quotes from error…

Aditevil:
… strings

Some of our package tests have quoted error strings and they need to be removed for proper error detection.

### DIFF
--- a/pipelines/test/daemon-check-output.yaml
+++ b/pipelines/test/daemon-check-output.yaml
@@ -203,6 +203,8 @@ pipeline:
           set --
           while read line ; do
               [ "$line" = "" ] && continue
+              # Strip leading and trailing quotes from error patterns
+              line=$(echo "$line" | sed 's/^"\(.*\)"$/\1/' | sed "s/^'\(.*\)'$/\1/")
               set -- "$@" "$line"
           done < "$ERRORS_F"
           num_error_strings=$#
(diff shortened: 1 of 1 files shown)


pipelines/test/daemon-check-output.yaml
```diff
@@ -203,6 +203,8 @@ pipeline:
           set --
           while read line ; do
               [ "$line" = "" ] && continue
+              # Strip leading and trailing quotes from error patterns
+              line=$(echo "$line" | sed 's/^"\(.*\)"$/\1/' | sed "s/^'\(.*\)'$/\1/")
```
tulilirockz:
```suggestion
              line=$(echo "$line" | sed -e 's/^"\(.*\)"$/\1/' -e "s/^'\(.*\)'$/\1/")
```

Could we do this instead? I _need_ to save 3 milisseconds! 😆 
```
bash-3.2$ echo "hello" | sed 's/^"\(.*\)"$/\1/' | sed "s/^'\(.*\)'$/\1/"
hello
bash-3.2$ echo "'hello'" | sed 's/^"\(.*\)"$/\1/' | sed "s/^'\(.*\)'$/\1/"
hello
bash-3.2$ echo "'hello'" | sed -e 's/^"\(.*\)"$/\1/' -e "s/^'\(.*\)'$/\1/"
hello
bash-3.2$ echo "'hello'" | sed -e 's/^"\(.*\)"$/\1/' -e "s/^'\(.*\)'$/\1/"
hello
```

sergiodj (approved):

PR #69508, opened 2025-10-20
fix(airflow-3): Respect constraints

klrmngr:
This addresses Airflow vendoring an incompatible version of Werkzeug

Several Airflow dependencies were previously bumped every time we would go to bump a dependency to remediate a CVE. This happened because `pip install --upgrade` does not respect constraints of other packages on its own so any dependencies bumped would also lead to the dependencies of those dependencies being bumped as well

Airflow published constraints for each prod release we can use to respect the versions of dependencies bundled in every Airflow release. This allows us to also use our `bump-constraints` pipeline that respects the constraints of all packages in an Airflow installation, making it much harder for us to accidentally bump a dependency to an unsupported version

This also moves Airflow over to PyPI (we were previously including several dependencies from PyPI as part of the remediation process), speeding up builds of Airflow as now we do not need to build the Airflow wheel or build the frontend

### DIFF
--- a/airflow-3.yaml
+++ b/airflow-3.yaml
@@ -38,8 +38,9 @@ vars:
   # This list is based on providers that were part of the 2.x version of this package, as well as manually added dependencies
   # to ensure we're building dependent providers too (not pulling from PyPi). This is an ordered list.
   # Use / as a separator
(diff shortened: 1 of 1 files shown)


astrojerms (approved):
looks good!

PR #69543, opened 2025-10-20
Test using ldd-check switch to tw's ver-check and help-check

murraybd:


postgresql-16.yaml
```diff
@@ -434,40 +434,17 @@ test:
       PGDATA: /tmp/test_db
       PGUSER: wolfi
   pipeline:
+    - uses: test/tw/help-check
+      with:
+        bins: ecpg initdb pg_archivecleanup pg_checksums pg_config pg_controldata pg_ctl pg_resetwal pg_rewind pg_test_fsync pg_upgrade pg_waldump pgbench postgres
+    - uses: test/tw/ver-check
+      with:
+        bins: initdb pg_resetwal pg_waldump pg_upgrade pg_archivecleanup postgres pg_config pg_controldata pg_ctl pg_checksums ecpg pg_test_timing pgbench pg_test_fsync pg_rewind
```
sergiodj: The fact that this list and the one above are not in the same order makes me want to sort them by hand :-(.  Would it be possible to do that please?

sergiodj (changes requested):

postgresql-16.yaml
murraybd: I've resolved this and fixed my script for switching to ver-check.

sergiodj (approved):

PR #69597, opened 2025-10-21
Spilo no dep wal g

utieyin:
There is no need to have a hard dependency on a specific version of postgres for wal-g. Removing that dependency


wal-g.yaml
```diff
@@ -23,6 +16,7 @@ environment:
 
 vars:
   python-version: "3.13"
+  pg-version: "18"
```
jamonation: If there's no version dependency, then this shouldn't be used? e.g anywhere using it ought to be a plain `postgresql` or `postgresql-client`

utieyin: I thought it would be better to pin the version used in testing, will remove it

Dentrax (approved):
LGTM, checked no other packages depend on this; and I'd be great if you could update the PR description why it's not needed.

PR #69631, opened 2025-10-21
ssh-import-id - drop ssh-keygen dep

smoser:
This is copied from upstream PR https://code.launchpad.net/~smoser/ssh-import-id/+git/ssh-import-id/+merge/494576


smoser:
Will wait for this to land upstream.

PR #69650, opened 2025-10-21
Ldd check tesseract

murraybd:
Switch to using tw's help-check and ver-check, add ldd-check to the main and subpackage.


sergiodj:
It looks like it needs to B-D on `giflib-dev`.

sergiodj (approved):

PR #69686, opened 2025-10-22
apache-arrow: tidy protobuf build dependencies

javacruft:
Just depend on protobuf-dev and drop libprotobuf and libprotoc; with the
new shared library naming approach the names of these packages will
change over time and the current version of protobuf-dev will always
point to the right versions.


sil2100 (approved):
It builds so it still works!

PR #69712, opened 2025-10-22
Normalize gnupg, gpgme with FIPS editions

Taffer:
This normalizes the parameterization and tests in the `gnupg.yaml` and `gpgme.yaml` with their FIPS counterparts in `enterprise-packages`.

Also adds some source info to the `gnupg` patches that didn't already have it.

### DIFF
--- a/gnupg.yaml
+++ b/gnupg.yaml
@@ -1,7 +1,7 @@
 package:
   name: gnupg
   version: 2.4.8
-  epoch: 2
+  epoch: 3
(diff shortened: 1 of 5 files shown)


sergiodj (approved):
I like the idea of using variable substitution in general, but I'm not a big fan of over-using it like we're doing here.  Having said that, I'd rather have sync'ed packages across the repos, so +1 :-).

Taffer:
yeah, I'm not sure how much is too much, or where it's better not to use it

PR #69769, opened 2025-10-22
Drop `gccgo`

sergiodj:


go-1.20.yaml
```diff
@@ -19,7 +19,6 @@ environment:
       - build-base
       - busybox
       - ca-certificates-bundle
-      - gcc-go
```
xnox, line 22: go needs to build-depend on itself here instead.

sergiodj: Yep, just noticed (and fixed) it.  Thanks!

sergiodj:
ftpmirror.gnu.org is acting out again.  Will retry later.

sergiodj:
Ready for review now.

xnox (approved):

PR #69858, opened 2025-10-23
openssf-compiler-options: Add hardening check for gfortran

cjwatson:
This is fairly basic since it's not entirely clear whether things like source fortification even theoretically apply to Fortran, but it does at least check that immediate binding and control flow integrity work.


xnox (approved):

stevebeattie (approved):
This looks good to me, my attempts to get gfortran to emit stask-clash-protection instructions failed.

justinvreeland (approved):

PR #69868, opened 2025-10-23
transition: protobuf 33

javacruft:
Bump package epochs for no change rebuild with new protobuf.


dannf (approved):
```
$ pr-only-does.py -t bumps-epochs https://github.com/wolfi-dev/os/pull/69868
Unexpected content [-      - libprotobuf]
No, https://github.com/wolfi-dev/os/pull/69868 does *not* just bump epochs
```

But I see that change is intentional/expected, so +1

PR #69886, opened 2025-10-23
apache-nifi: FTBFS issue

jamie-albert:
attempting to reproduce fail to build from source issue


aborrero:
I tested a few things suggested by claude, and they don't work right away. Will need deeper research.

kwmonroe (approved):

PR #69947, opened 2025-10-24
protobuf/fix build envs

javacruft:
- **py3-tensorflow-model-analysis: tidy protobuf-dev build dependencies.**
- **rethinkdb: drop surplus build dependencies.**


justinvreeland (approved):
Are they vendored in or pulled in by sca or something?

PR #69959, opened 2025-10-24
Better Python 3.14 JIT build options

Taffer:
Some tweaks intended to improve experimental JIT performance, although the "real" deal requires some GPO support that we don't have yet.

### DIFF
--- a/python-3.14.yaml
+++ b/python-3.14.yaml
@@ -1,7 +1,7 @@
 package:
   name: python-3.14
   version: "3.14.0"
-  epoch: 3
+  epoch: 4
   description: "the Python programming language"
   copyright:
     - license: PSF-2.0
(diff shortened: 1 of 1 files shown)


python-3.14.yaml
```diff
@@ -73,17 +73,23 @@ pipeline:
...
 
+  # Experimental JIT support
+  #
+  # --enable-optimizations with clang requires PGO support (in particular,
+  # we're missing libclang_rt.profile). This build configuration won't be
+  # optimal and might not produce any performance gains.
   - name: Configure
     runs: |
       ./configure \
+         CC=clang \
```
sergiodj, line 84: OOC, does this mean that upstream claims that using `clang` instead of `gcc` results in better performance?

Taffer: `clang`'s required for the JIT currently, but they're intending to remove that; AFAIK it's not a performance thing, they just haven't gotten it solid with `gcc` yet.

sergiodj (approved):

PR #70012, opened 2025-10-25
improve the tests and use go/build pipeline

developer-guy:
Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/newrelic-fluent-bit-output.yaml
+++ b/newrelic-fluent-bit-output.yaml
@@ -1,67 +1,53 @@
 package:
   name: newrelic-fluent-bit-output
   version: "2.4.0"
-  epoch: 3 # CVE-2025-4674
+  epoch: 4 # CVE-2025-4674
   description: A Fluent Bit output plugin that sends logs to New Relic
   copyright:
     - license: Apache-2.0
 
-environment:
-  contents:
-    packages:
-      - build-base
-      - busybox
-      - ca-certificates-bundle
-      - go-1.20
-      - wolfi-baselayout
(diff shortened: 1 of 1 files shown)


newrelic-fluent-bit-output.yaml
```diff
@@ -11,54 +11,44 @@ environment:
...
       - runs: |
-          mkdir -p ${{targets.subpkgdir}}/fluent-bit/bin/
+          mkdir -p ${{targets.contextdir}}/fluent-bit/bin/
           # The upstream chart expects the fluent-bit binary to be in /fluent-bit/bin/fluent-bit
-          ln -sf /usr/bin/fluent-bit ${{targets.subpkgdir}}/fluent-bit/bin/fluent-bit
-    dependencies:
-      runtime:
-        - newrelic-fluent-bit-output
-    # test added by a robot (compat)
+          ln -sf /usr/bin/out_newrelic.so ${{targets.contextdir}}/fluent-bit/bin/out_newrelic.so
```
murraybd, line 33: Given that there are .so files here you should also add `test/tw/ldd-check` to the test pipeline.

developer-guy: we were doing that at the end of the manifest in the test section but I have added it here too, thanks!

newrelic-fluent-bit-output.yaml
```diff
@@ -11,54 +11,45 @@ environment:
     packages:
```
rakshitgondwal: i think we should not be needing these build time deps anymore?

developer-guy: let me check

developer-guy: you are absolutely right!

rahulduvedi (approved):

PR #70064, opened 2025-10-27
py3-gevent: allow building with recent gcc version

toabctl:
https://github.com/gevent/gevent/issues/2028 got fixed and py3-gevent builds fine with a recent gcc (tested with 15.2) version.

### DIFF
--- a/py3-gevent.yaml
+++ b/py3-gevent.yaml
@@ -1,7 +1,7 @@
 package:
   name: py3-gevent
   version: "25.9.1"
-  epoch: 0
+  epoch: 1
   description: Coroutine-based network library
   copyright:
     - license: MIT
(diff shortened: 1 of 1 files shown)


aborrero:
please bump the epoch of the file :-) Otherwise LGTM.

philroche:
```
epoch: 1
```

To force a rebuild and publication of this new version

toabctl:
@aborrero @philroche done.

kranurag7 (approved):

kranurag7 (changes requested):
looking closely I think we can omit `gcc` explicitly here (ref: https://github.com/wolfi-dev/os/blob/dfbc13d976c2eec279ebcfe9f0c47b089024b115/py3-supported-python.yaml?plain=1#L87-L91) 

py3-supported-build-base-dev pulls in `build-base` which in turns pulls in latest gcc at build-time.

kranurag7 (approved):
thanks!!

PR #70072, opened 2025-10-27
Add no-GIL free threading Python 3.14

Taffer:
Adds the no-GIL "free threading" version of Python 3.14.

This is a *compile time* option, and you are *not* expected to have both `python-3.14` and `python-3.14-freethreading` installed at the same time.


python-3.14t.yaml
```diff
@@ -0,0 +1,325 @@
...
+  epoch: 0
+  description: "the Python programming language"
+  copyright:
+    - license: PSF-2.0
+  resources:
+    cpu: 8
+    memory: 8Gi
+  dependencies:
+    runtime:
+      - ${{package.name}}-base=${{package.full-version}}
```
justinvreeland: If this is a drop in replacement it probably needs a `provides:` as well.

python-3.14t.yaml
```diff
@@ -0,0 +1,325 @@
...
+      export LD_LIBRARY_PATH="${{targets.destdir}}/usr/lib${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}"
+      ${{targets.destdir}}/usr/bin/${{vars.python}} -m compileall --invalidation-mode=unchecked-hash \
+          -r100 ${{targets.destdir}}/usr/lib/
+
+  - uses: strip
+
+subpackages:
+  - name: "${{package.name}}-base"
+    description: "${{package.name}} without /usr/bin/python3"
+    dependencies:
```
justinvreeland: also will need a provide here.

python-3.14t.yaml
```diff
@@ -0,0 +1,325 @@
...
+        ${{vars.python}} version-check.py ${{package.version}}
+        ${{vars.python-t}} version-check.py ${{package.version}}
+    - name: Verify working python3 -m venv
+      runs: |
+        d=$(mktemp -d)
+        ${{vars.python-t}} -m venv "$d"
+        $d/bin/pip list
+        $d/bin/pip check
+        rm -Rf "$d"
+    - name: Verify no-GIL build
```
justinvreeland: 👍🏻

python-3.14t.yaml
dannf: I don't believe this is ABI compatible.

python-3.14-freethreading.yaml
```diff
@@ -0,0 +1,325 @@
+package:
+  name: python-3.14-freethreading
```
dannf: Can we call it `python-3.14t` ? It's shorter, and any apk's we build for its ABI could be `py3.14t-foo`, so our `data: py-versions` pattern could continue.

dannf (commented):
Did you consider making this a subpackage of `python-3.14.yaml` to avoid skew?

python-3.14t.yaml
justinvreeland: That's kind of crazy. But we still need to be able to use this as a drop in replacement for existing python modules.

xnox: I believe we will not be able to. This build has a different ABI suffix, and thus different build of compiled extensions and pyc cache files. Thus ideally it should be made co-installable, and yes iterate the world to build for this type.

I believe this is similar to debug builds which too are different ABI.

python-3.14-freethreading.yaml
xnox: Yes please. And try to make it co-installable

python-3.14t.yaml
Taffer: AFAIK we're not doing `provides:` for 3.14 yet; if that's changed I should update the regular package as well.

Taffer:
> Did you consider making this a subpackage of `python-3.14.yaml` to avoid skew?

I did, but it's a configuration/compile time change, which seemed like too much for a subpackage; is that something subpackages can do?

Update: Jason said it sounded reasonable, so I'll give it a go. Please stand by!

python-3.14t.yaml
Taffer: I suppose this is the first source of no-GIL, so maybe it's fine? I'll add it in.

python-3.14t.yaml
Taffer: The 3.13 package doesn't have one for `-base`, so we need to add these throughout?

Taffer: Also, I want to make sure nobody gets this by accident in case there are issues caused by removing the GIL. Presumably in the future it'll be the default.

dannf:
> > Did you consider making this a subpackage of `python-3.14.yaml` to avoid skew?
> 
> I did, but it's a configuration/compile time change, which seemed like too much for a subpackage; is that something subpackages can do?
> 
> Update: Jason said it sounded reasonable, so I'll give it a go. Please stand by!

One thing I'd be worried about if we were to provide `python-3.14` (which I currently think we should not do) is that - with separate YAML files - the epochs would be staggered, so sometimes free-threading would be "latest", sometimes non-free-threading would be.

Taffer:
No-GIL Python 3.14 is now a sub-package that contains `python`, `python3`, `python3.14`, `python3.14t` binaries; I'm assuming you'd install this over? the regular Python 3.14 if you wanted to experiment with it, or instead of the `python-3.14-base`?

I was unable to get the "can we make a venv" test for the subpackage to work due to things getting confused about where a Python 3.14 would come from for `py3-pip-wheel`. AFAIK that only uses pure Python modules, so there's no reason for it not to work with the no-GIL version, if we could tell it "we have a Python 3.14 base equivalent, don't worry about it".

python-3.14.yaml
```diff
@@ -140,6 +145,118 @@ pipeline:
...
+        - python3t=${{package.full-version}}
+        - python-3t=${{package.full-version}}
+      runtime:
+        # 'import _tkinter' will fail with ImportError on 'libtcl9tk9.0.so'
+        # which is provided by tk.  melange SCA does not find this dependency.
+        - tk
+    pipeline:
+      # Hopefully this is cached. If not, we could copy it after checking out
+      # in the main package build.
+      - uses: git-checkout
```
sil2100: I am unsure about the caching part. I don't think melange ever caches git clones - there is a melange cache but it's more for other build artifacts like Go modules etc. Unsure about any caching above melange from the build infra too.
Also, I would prefer a copy step right after the `git-checkout` of the original source because a separate checkout requires us to not only copy-paste a lot of boiler plate, but also we need to ensure that the same cherry-picks/patches are applied on top of both `git-checkout` calls. This is problematic and we can get out-of-skew really quickly by accident.

python-3.14.yaml
```diff
@@ -78,27 +86,27 @@ pipeline:
...
+  - uses: autoconf/configure
+    with:
+      opts: |
+        CC=clang \
+        --host=${{host.triplet.gnu}} \
+        --build=${{host.triplet.gnu}} \
+        --target=${{host.triplet.gnu}} \
+        --prefix=/usr \
+        --enable-experimental-jit=yes-off \
          --with-tail-call-interp \
```
sil2100: I think whitespace is weird here.

python-3.14.yaml
```diff
@@ -140,6 +145,118 @@ pipeline:
...
+      - uses: git-checkout
+        working-directory: no-gil
+        with:
+          expected-commit: ebf955df7a89ed0c7968f79faec1de49f61ed7cb
+          repository: https://github.com/python/cpython.git
+          tag: v${{package.version}}
+          cherry-picks: |
+            3.14/d11e69d6203080e3ec450446bfed0516727b85c3: Fix CVE-2025-8291
+      - name: Force use of system libraries
+        working-directory: no-gil
```
sil2100: Since we need to repeat the working-directory stanza with every pipeline step here, I think you can remove some of this boilerplate by wrapping the subpackage in a top level pipeline with a `working-directory`. Example as seen in melange `examples/working-directory.yaml`:

```
pipeline:
(...)
  - working-directory: /home/build/bar
    pipeline:
      - runs: |
          echo "current working directory: $(pwd)"
```

This would only mean one setting of `working-directory` then, yay! And I *think* that should work.

sil2100 (changes requested):
Thank you, this is looking good. Some inline proposals and changes I'd like to see before this goes in, but I really like the subpackage approach much more. I was concentrating more on the melange side than the Python side - I'll leave that to Justin.

Checked the python-t package contents and they seemed sane. The ABI incompatibility between the two versions in certain cases is problematic, and makes me wonder about how we want to consume this package later. But for now I feel it is more of an experiment.

One thing I didn't point out in inline comments is that I was also thinking in how we can deduplicate the configure flags between the two versions. Maybe we could use `vars` for this? But then again I don't feel that would look good. So maybe something for the future.

Taffer:
> Checked the python-t package contents and they seemed sane. The ABI incompatibility between the two versions in certain cases is problematic, and makes me wonder about how we want to consume this package later. But for now I feel it is more of an experiment.

Yeah, this is more of a thing for people who want to test their stuff with no GIL; it was specifically requested by someone (IIRC Cisco). In my mind, you install your thing, then this over top of the existing Python and hopefully everything keeps working. :crossed_fingers: 

> One thing I didn't point out in inline comments is that I was also thinking in how we can deduplicate the configure flags between the two versions. Maybe we could use `vars` for this? But then again I don't feel that would look good. So maybe something for the future.

I did have a go at this, but the `${{host.triplet.gnu}}` doesn't seem to be available to `vars:`; I'll see if I can come up with something while I'm resolving other comments here.

python-3.14.yaml
```diff
@@ -43,16 +43,46 @@ environment:
...
+    --enable-loadable-sqlite-extensions \
+    --enable-shared \
+    --with-lto \
+    --with-computed-gotos \
+    --with-dbmliborder=gdbm:ndbm \
+    --with-system-expat \
+    --with-system-libmpdec \
+    --without-ensurepip \
+    --with-builtin-hashlib-hashes=blake2 \
+    --with-wheel-pkg-dir=/usr/share/python-wheels
```
sil2100, line 62: Previously this was different for python3.14t than for python3.14 (prefixed with the subpackage path variable?). Not sure if it's important?

sil2100 (commented):
Looks good from my end! One question inline as I just want to make sure it's all good as expected. Maybe now it's good but previously it was not correct?

python-3.14.yaml
Taffer: I'm not sure it matters; wheels with pre-built binaries inside (is that a thing?) might not be compatible with the no-GIL python, but that's also true for other architectures/platforms. We might need to add a `python-3.14t-dev` package for people to rebuild things (I *think* it's OK to leave that until we find out it's required).

sil2100 (approved):
Fair enough! I guess this is good. Not a Python toolchain expert, but it looks correct.

PR #70090, opened 2025-10-27
DNM: openjdk-25: Test

sergiodj:
DNM.


sergiodj:
This was done to verify that #70087 is not reproducible, which is confirmed.

PR #70091, opened 2025-10-27
chore: move posix-cc-wrappers inline with gcc-<n>

kranurag7:
In current state, when we install any `gcc-<n>` package then the latest
gcc comes up as well.

this comes up because we `posix-cc-wrappers` is a runtime dependency
to `gcc-<n>`

and `posix-cc-wrappers` itself depend on latest gcc at runtime.

```yaml
  dependencies:
    runtime:
      - gcc
```
this commit moves posix-cc-wrappers inline with gcc and builds a
subpackage with name `posix-cc-wrappers-<n>` with a provides of
`posix-cc-wrappers` for backwards compatibility.

after building the package inline, we pin the runtime dependency of `gcc-<n>`
like

`- posix-cc-wrappers-${{vars.major-version}}=${{package.full-version}}`

this will always pull in the inline `posix-cc-wrappers-<n>` that we're
building with pinned epoch and revision.

### DIFF
--- a/gcc-14.yaml
+++ b/gcc-14.yaml
@@ -157,37 +157,38 @@ subpackages:
       runtime:
         - gcc=${{package.full-version}}
     pipeline:
-      - working-directory: '${{targets.destdir}}/usr/bin'
-      - runs: |
-          ln -s gcc cc
-      - runs: |
-          cat >c89 <<'EOF'
-          #!/bin/sh
-          _flavor="-std=c89"
-          for opt; do
-            case "$opt" in
-            -ansi|-std=c89|-std=iso9899:1990) _flavor="";;
-            -std=*) echo "$(basename $0) called with non ANSI/ISO C option $opt" >&2
-                    exit 1;;
-            esac
-          done
-          exec cc $_flavor ${1+"$@"}
-          EOF
-      - runs: |
-          cat >c99 <<'EOF'
-          #!/bin/sh
-          _flavor="-std=c99"
-          for opt; do
-            case "$opt" in
-            -std=c99|-std=iso9899:1999) _flavor="";;
-            -std=*) echo "$(basename $0) called with non ISO C99 option $opt" >&2
-                    exit 1;;
-            esac
-          done
-          exec cc $_flavor ${1+"$@"}
-          EOF
-      - runs: |
-          chmod 755 c?9
+      - working-directory: '${{targets.contextdir}}/usr/bin'
+        pipeline:
+          - runs: |
+              ln -s gcc cc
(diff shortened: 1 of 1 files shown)


sergiodj (changes requested):
Thanks, @kranurag7!

It's late here so this is not a full review, but rather a brain dump of what I'm thinking.

While I do agree that the current situation is not ideal, I'm also a bit hesitant to add multiple subpackage definitions for each version of gcc/clang we ship, especially because these subpackages will all be exactly the same.

Something I've been thinking is that we could actually *relax* the current dependencies of `posix-cc-wrappers`, and drop the runtime requirement for `gcc`.  You see, the only job this package has to perform is install symlinks for `cc`, `c89` and `c99`.  That's all.  It doesn't need to depend on a compiler, and in fact one could say that it *shouldn't* depend on one.  It's actually the other way around: the compilers should depend on this package.

As I said, it's late here so there may be flaws in my proposal, but I believe it is a cleaner approach to what we have today.  We will need to make some small adjustments to the `test` pipeline, but otherwise it should work fine.

WDYT?

sergiodj:
Note to self:

```
gcc-14-14.3.0-r8 depends on:
binutils
glibc-dev
libquadmath
libstdc++-14-dev
openssf-compiler-options
posix-cc-wrappers
so:ld-linux-x86-64.so.2
so:libc.so.6
so:libgmp.so.10
so:libisl.so.23
so:libm.so.6
so:libmpc.so.3
so:libmpfr.so.6
so:libstdc++.so.6
so:libz.so.1
so:libzstd.so.1
```

xnox:
Merge it onto openssf compiler options?
Or even drop it completely?
Default compiler standards have increased above 89 and 99, and compiling code to those standards is no longer relevant.
I don't know how popular it is still to use cc, given that gcc Vs clang matters.

sergiodj:
> Merge it onto openssf compiler options? Or even drop it completely? Default compiler standards have increased above 89 and 99, and compiling code to those standards is no longer relevant. I don't know how popular it is still to use cc, given that gcc Vs clang matters.

We still need `cc` at least; there are some packages that still rely on it.  I agree that it should be possible to just drop `c89` and `c99` (but let's double-check first).

Also, I like the idea of moving the symlink to `openssf-compiler-options`.  Let me work on that.

@kranurag7 I think this PR can be closed now that we've decided on a different approach?  Thanks for doing that, btw!

xnox:
let's mark it draft; and only close if we think we have the symlinks in the right places.

cmwilson21:
Thanks for the contribution!

PR #70095, opened 2025-10-28
refactor: move ocb removal

anushkamittal20:
move removing ocb to compat package so ocb can be used when needed

### DIFF
--- a/opentelemetry-collector.yaml
+++ b/opentelemetry-collector.yaml
@@ -65,12 +65,11 @@ subpackages:
           packages: .
           modroot: ./cmd/builder
           output: ocb
-
     test:
       pipeline:
-      - uses: test/tw/help-check
-        with:
-          bins: ocb
+        - uses: test/tw/help-check
+          with:
+            bins: ocb
 
 update:
   enabled: true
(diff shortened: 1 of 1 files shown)


rahulduvedi:
can you provide more context from this one? 
and can you please add symlink test (if we decide to keep it)

rahulduvedi (approved):

rahulduvedi:
LGTM, Please wait for a one more approval here please.
Thanks @anushkamittal20

opentelemetry-collector.yaml
```diff
@@ -58,6 +57,42 @@ subpackages:
...
+              - gomod: go.opentelemetry.io/collector/exporter/debugexporter v${{package.version}}
+
+            receivers:
+              - gomod: go.opentelemetry.io/collector/receiver/otlpreceiver v${{package.version}}
+
+            processors:
+              - gomod: go.opentelemetry.io/collector/processor/batchprocessor v${{package.version}}
+            EOF
+
+            ocb --config test-manifest.yaml --skip-compilation --skip-get-modules
```
rahulduvedi, line 94: Although this looks good to me, But I am not sure what are we verifying here?

anushkamittal20: ocb takes the manifest and generates the skeleton here - directory structure, I could add a check for that but the command would simply fail if it didnt work 🤔

PR #70101, opened 2025-10-28
chore: remove runtime deps of posix-cc-wrappers

kranurag7:
chore: remove runtime deps of posix-cc-wrappers

    we have gcc as runtime dependency to posix-cc-wrappers and each gcc
    includes posix-cc-wrappers as runtime depedency meaning whichever
    gcc-<ver> we pull in, we always bring in latest gcc.

    this commit removes gcc as runtime dependency from posix-cc-wrappers
    and we should be good with this because this package mostly carries a
    symlink and shell scripts.

### DIFF
--- a/posix-cc-wrappers.yaml
+++ b/posix-cc-wrappers.yaml
@@ -1,13 +1,10 @@
 package:
   name: posix-cc-wrappers
-  version: 1
+  version: 2
   epoch: 7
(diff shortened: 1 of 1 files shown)


xnox:
yeah I never understood this circular dependency.

sergiodj (approved):
LGTM, thanks!

PR #70102, opened 2025-10-28
openssl: add patch to block deprecated HMAC API in FIPS mode

xnox:
When using deprecated HMAC API, ensure that digest is from the FIPS
provider. This should block creating FIPS HMAC with unapproved digests.

### DIFF
--- a/openssl.yaml
+++ b/openssl.yaml
@@ -1,7 +1,7 @@
 package:
   name: openssl
   version: "3.6.0"
-  epoch: 1
+  epoch: 2
   description: "the OpenSSL cryptography suite"
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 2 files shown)


Taffer (approved):

xnox:
This regressed because of

> HMAC_Init_ex() initializes or reuses a HMAC_CTX structure to use the hash function evp_md and key key. If both are NULL, or if key is NULL and evp_md is the same as the previous call, then the existing key is reused. ctx must have been created with HMAC_CTX_new() before the first use of an HMAC_CTX in this function.
> 
> If HMAC_Init_ex() is called with key NULL and evp_md is not the same as the previous digest used by ctx then an error is returned because reuse of an existing key with a different digest is not supported.
> 
> HMAC_Init() initializes a HMAC_CTX structure to use the hash function evp_md and the key key which is key_len bytes long.

And the code didn't account for the case when MD is null; or if MD is without a provider yet (no implementation bound); and thus 

```
 OSSL_PROVIDER_get0_name(EVP_MD_get0_provider(md))
```

Would segfault when EVP_MD_get0_provider returns NULL, and get0_name then tries to deference it.

PR #70152, opened 2025-10-28
fix(tpm2-tss): split dev man subpackages

89luca89:
Split package in -dev and -doc

### DIFF
--- a/tpm2-tss.yaml
+++ b/tpm2-tss.yaml
@@ -1,7 +1,7 @@
 package:
   name: tpm2-tss
   version: "4.1.3"
-  epoch: 2
+  epoch: 3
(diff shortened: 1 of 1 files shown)


a-crate (commented):
This introduces new empty dirs to `var/run`. Previously they were in `usr/local/var/run`. Can we switch to `run/...` by using `--with-runstatedir`?

Not including the empty dirs would be best but we are already shipping them in the other prefix, so maybe we don't want to remove them now.

89luca89:
> --with-runstatedir

Done, also added a step to remove empty dirs just to be sure

a-crate (approved):

PR #70167, opened 2025-10-28
qemu - add qemu-user, small cleanups

smoser:
- specify python version, and use python3-as-3.13
 - add libjpeg-turbo-dev for jpeg support in vnc
 - drop python sphinx as it was not actually building docs anyway due to not specific version of rtd


smoser:
After installing `qemu-user-binfmt` on a x86_64 workstation, and `sudo systemctl start systemd-binfmt` you can:

```
$ make ARCH=aarch64 MELANGE_RUNNER=bubblewrap package/wolfi-base
```

You wont need to run the service if the package is installed at boot time.

qemu.yaml
```diff
@@ -209,6 +221,66 @@ subpackages:
...
+
+  - name: ${{package.name}}-user-binfmt
+    description: "binfmt files for qemu-user"
+    dependencies:
+      runtime:
+        - ${{package.name}}-user
+    pipeline:
+      - runs: |
+          # package the symlinks to qemu-user that systemd-binfmt.service(8) will see
+          # but remove the targets of those symlinks whic are provided by qemu-user
```
dannf: nit: typo'd "which"

qemu.yaml
```diff
@@ -209,6 +221,66 @@ subpackages:
...
+                  --target-list=aarch64-linux-user,i386-linux-user,x86_64-linux-user \
+                  --prefix=/usr
+          make -j$(nproc)
+          make install DESTDIR=${{targets.contextdir}}
+          rm -rf ${{targets.contextdir}}/usr/share/qemu
+          cd ../..
+
+          chmod ugo+x binfmt-install
+          ./binfmt-install ${{targets.contextdir}} ${{build.arch}} aarch64 i386 x86_64
+          # keep /usr/share/qemu/binfmt.d which are just inert files
```
dannf: a "keep" comment before a "rm -Rf" seems confusing. I see they are different directories, but the relation isn't clear for me.

89luca89 (dismissed):

qemu.yaml
```diff
@@ -209,6 +221,66 @@ subpackages:
...
+    test:
+      pipeline:
+        - uses: test/tw/ver-check
+          with:
+            bins: qemu-aarch64 qemu-i386 qemu-x86_64
+            version-flag: "-version"
+        - uses: test/tw/help-check
+          with:
+            bins: qemu-aarch64 qemu-i386 qemu-x86_64
+            help-flag: -h
```
dannf, line 263: Should we check that the binaries are actually static?

dannf (dismissed):
nice! just some nits inline

qemu.yaml
smoser: how? ldd?

dannf (approved):

PR #70180, opened 2025-10-28
21912 ceph

utieyin:
Using `uv pip` to build ceph to remove the bcrypt py03 issues and other dependencies missing from the current package vs upstream


ceph-19.yaml
```diff
@@ -221,9 +217,36 @@ pipeline:
...
+        packaging \
+        pecan \
+        pyOpenSSL \
+        python-dateutil \
+        requests \
+        Routes \
+        scikit-learn \
+        scipy \
+        urllib3 \
+        Werkzeug
```
javacruft, line 240: The upstream source tree has many requirements.txt files that define exactly what's needed in terms of dependencies and versions - please lets use those instead so that we track with the upstream project.

javacruft (changes requested):
If possible I'd like to see the installed python not installed to site-packages; a location such as /opt/chainguard/${{package.name}} is unique and won't create co-install issues with any other packages.  We need to teach python (specifically in the context of the ceph-mgr modules) how to actually find it! that might be via one of the compile flags.

ceph-19.yaml
utieyin: There is a handy script that tries to crawl `src/pybind/mgr/requirements.txt` for all the required packages but falls about 10% short.

PR #70268, opened 2025-10-29
Apache-nifi-registry remove jetty-http2-common from pombump to fix TLS startup

rahulduvedi:
Fixes missing `jetty-http2-common-12.0.27.jar` causing TLS startup failure with NoClassDefFoundError: org/eclipse/jetty/http2/SessionContainer.

Root cause: pombump adds missing dependencies to DependencyManagement section with default scope:import, which interfered with Maven's transitive dependency resolution. Since `jetty-http2-common` is already a transitive dependency of `jetty-http2-server`, the pombump entry prevented natural inclusion.

Verified: r2 APK now contains `jetty-http2-common-12.0.27.jar` and matches upstream `apache/nifi-registry:latest` image 1:1.


AnmolVirdi (approved):
LGTM.
Using updated `pombump-deps.yaml` for `apache-nifi-registry`

PR #70284, opened 2025-10-29
chore: remove posix-cc-wrappers from runtime of gcc

kranurag7:
we previously discussed about c89 and c99 being obselete here https://github.com/wolfi-dev/os/pull/70091#issuecomment-3454419489

openssf-compiler-options which is runtime dependency to each gcc also carries the symlink that links cc to gcc-wrapper.
`ln -s gcc-wrapper cc`

https://github.com/wolfi-dev/os/blob/09e8401c821f0e5dab24b90e2a327160431e2c3d/openssf-compiler-options.yaml?plain=1#L32

PATH wise this wins (`/usr/local/bin/cc`) over what comes from posix-cc-wrappers (`/usr/bin/cc`) so out of the box posix-cc-wrappers is not used anyways.

this commit drops posix-cc-wrappers from runtime dependency of gcc.


sergiodj (changes requested):
Can you also remove it from the other GCC versions we carry?  `gcc-14`, `gcc-13`, etc.

sergiodj:
And with that, I believe we should just remove the `posix-cc-wrappers` package altogether.

xnox (changes requested):
i want to check this interactively.

cause i thought that gcc-wrapper, needs real link in /usr/bin/ to exec, unless it does translation from cc to gcc or somesuch.

PR #70293, opened 2025-10-29
keycloak-config-cli: revert logback dependency and bump epoch

jamie-albert:
## Summary
Reverts the logback-core dependency addition from https://github.com/wolfi-dev/os/pull/69697 and bumps epoch to 51 instead.

## Changes
- Removed logback-core dependency from pombump-deps.yaml
- Bumped epoch from 50 to 51

## Rationale
This dependency bump breaks the image, reverted and added to [autobump ignore list. ](https://github.com/chainguard-dev/infra-images/pull/2928)

## Verification
- Build: `make package/keycloak-config-cli`
- Scan: `wolfictl scan packages/*/*/keycloak-config-cli-*.apk`

### DIFF
--- a/keycloak-config-cli.yaml
+++ b/keycloak-config-cli.yaml
@@ -1,7 +1,7 @@
 package:
   name: keycloak-config-cli
   version: 6.4.0
-  epoch: 50 # GHSA-25qh-j22f-pwp8
+  epoch: 51
   description: Import YAML/JSON-formatted configuration files into Keycloak - Configuration as Code for Keycloak.
   copyright:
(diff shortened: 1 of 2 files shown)


powersj (approved):

kwmonroe (approved):
are we sure reversion is necessary?  we had a similar problem in zookeeper that was resolved by moving the `logback-core` pin to a more generic `logback-version`:

https://github.com/chainguard-dev/enterprise-packages/pull/36998/files

would a similar change here allow us to remediate differently?

PR #70295, opened 2025-10-29
Makefile - do not always make apks

smoser:
The case right now is that 'make package/py3-pytest-timeout' calls 'make packages/x86_64/py3-pytest-timeout-2.4.0-r3.apk'. That wil always be re-made because it depends on the phony target cache.

Moving the phony dependency to the 'package/py3-pytest' call makes _it_ always get remade, but thats ok. It was phony itself.

Before we had:

    $ make package/py3-pytest-timeout
    ... calls packages/x86_64/py3-pytest-timeout-2.4.0-r3.apk, does work...
    $ make pkgname=py3-pytest yamlfile=py3-pytest-timeout \
        packages/x86_64/py3-pytest-timeout-2.4.0-r3.apk
    ... does work every time ...

Now we have

    $ make pkgname=py3-pytest yamlfile=py3-pytest-timeout \
        packages/x86_64/py3-pytest-timeout-2.4.0-r3.apk
    make: 'packages/x86_64/py3-pytest-timeout-2.4.0-r3.apk' is up to date.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,13 +144,15 @@ kernel/%/vmlinuz: kernel/%/linux.apk
 yamls := $(wildcard *.yaml)
 pkgs := $(subst .yaml,,$(yamls))
 pkg_targets = $(foreach name,$(pkgs),package/$(name))
-$(pkg_targets): package/%:
+$(pkg_targets): package/%: cache
 	$(eval yamlfile := $*.yaml)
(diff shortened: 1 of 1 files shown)


EyeCantCU (approved):

gaughen (approved):
approving based on RJ's approval

PR #70351, opened 2025-10-30
openssl: revert recent FIPS HMAC block

sil2100:
This causes issues with node.


xnox (approved):

8none1 (approved):
xnox +1, so I approve on that basis

javacruft (approved):

PR #70396, opened 2025-10-30
Remove redundant OpenSSL provider patch

Taffer:
Investigating whether we can remove one of the FIPS-related patches.


python-3.13.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: python-3.13
   version: "3.13.9"
-  epoch: 0
+  epoch: 1
```
Taffer, line 4: ha, I didn't actually change 3.13 :facepalm: 

if I pick "Delete File" from the "..." menu on the right, will it delete the *change* or the entire file? I don't trust it to do the right thing

Taffer:
Actually, back to DNM, I'm going to see if I can figure out why 3.13 thinks it needs this.

EyeCantCU (approved):

PR #70476, opened 2025-10-31
feat(gstreamer): Init subpackage gst-plugins-good @ 1.27.2

SamD2021:
Fixes: https://github.com/chainguard-dev/image-requests/issues/6361

Related: https://github.com/wolfi-dev/os/pull/58498

#### For new package PRs only


gst-plugins-good.yaml
```diff
@@ -0,0 +1,115 @@
...
+        - ${{package.name}}
+    pipeline:
+      - uses: split/dev
+    description: ${{package.name}} dev
+    test:
+      pipeline:
+        - uses: test/pkgconf
+        - uses: test/tw/ldd-check
+          with:
+            packages: ${{subpkg.name}}
```
murraybd: IIRC this is the default with ldd-check so this is unnecessary.

gst-plugins-good.yaml
```diff
@@ -0,0 +1,115 @@
...
+
+  - name: ${{package.name}}-doc
+    pipeline:
+      - uses: split/manpages
+    description: ${{package.name}} manpages
```
murraybd: There is a test pipeline for docs which would be good to incorporate.

gst-plugins-good.yaml
```diff
@@ -0,0 +1,115 @@
...
+update:
+  enabled: true
+  git:
+    tag-filter-prefix: "1."
+
+test:
+  pipeline:
+    - uses: test/tw/ver-check
+      with:
+        bins: gst-inspect-1.0
```
murraybd: Given that gstreamer provides this binary I think checking the version of it is an unnecessary test.

opencore-amr.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+  - uses: autoconf/make-install
+
+  - uses: strip
+
+subpackages:
+  - name: opencore-amr-dev
+    description: Development files for opencore-amr
+    pipeline:
+      - uses: split/dev
```
murraybd, line 41: This provides some .pc files so adding pkgconf test is appropriate.

opencore-amr.yaml
```diff
@@ -0,0 +1,46 @@
...
+
```
murraybd, line 46: The primary package ships .so files so adding a test with ldd-check is appropriate too.

gst-plugins-good.yaml
SamD2021: This helps since it caught that there was no manpages offered by gst-plugins-good so we can remove the docs subpackage all together :)

gstreamer.yaml
```diff
@@ -2,30 +2,59 @@
...
+      - libtheora-dev
+      - libvorbis-dev
       - libxml2-dev
+      - libxv-dev
+      - mesa-dev
+      - mesa-gbm
+      - mesa-glx
+      - mesa-libgallium
       - meson
+      - opencore-amr
```
murraybd: Why is this needed when we have opencore-amr-dev?

gstreamer.yaml
```diff
@@ -2,30 +2,59 @@
...
+      - opus-dev
+      - orc-compiler
+      - orc-dev
       - perl
       - python3
+      - qt5-qtbase-dev
+      - qt5-qtbase-x11
+      - qt6-qtbase-dev
+      - qt6-qtbase-x11
+      - samba-dev
```
astrojerms, line 54: All of these added dependencies, are they required for the gstreamer build, or just for gst-plugins-good? Would they make sense to just be under the gst-plugins-good subpackage dependencies?

gstreamer.yaml
```diff
@@ -74,6 +103,51 @@ subpackages:
...
+          - uses: meson/compile
+          - uses: meson/install
+          - uses: strip
+    description: GStreamer streaming media framework good plug-ins
+    test:
+      pipeline:
+        # AUTOGENERATED
+        - runs: |
+            echo "Testing gst-plugins-good..."
+            for plugin in jpegdec pngdec autodetect udp audioconvert; do gst-inspect-1.0 $plugin >/dev/null && echo "✓ $plugin found"; done
```
astrojerms: Why are we redirecting to /dev/null here? Typically we want to see the output of even step, even if it may be verbose.

opencore-amr.yaml
```diff
@@ -0,0 +1,50 @@
+package:
+  name: opencore-amr
```
astrojerms, line 2: Any tests we can add to this package?

gstreamer.yaml
SamD2021: IIRC we can only add subpackage specific dependencies if they are runtime ones, so these build dependencies for the subpackages would have to go on the main shared build environment.

gstreamer.yaml
SamD2021: I agree, I'll change it to keep the output!

opencore-amr.yaml
SamD2021: I can't see additional test for this at the moment. This is a codec library where the existing test/pkgconf and test/ldd-check tests make sense since they verify the package builds correctly and provides proper dev interfaces. As far as I'm aware, functional codec testing happens in consuming apps like GStreamer, since it's not really end user facing and would require specific AMR audio samples and infra for audio testing

murraybd (approved):

PR #70482, opened 2025-10-31
fix update block of fluxcd-kustomize-mutating-webhook

kranurag7:
we're more consistent and correct if we drop the `v` in there

ref: https://github.com/xunholy/fluxcd-kustomize-mutating-webhook/tags


kranurag7:
skipping epoch bump given the changes are only related to update block.

powersj (approved):

Issue #64231, opened 2025-08-26
QEMU doesn't support new `termios` interface on glibc 2.42


robbiemcmichael:
I noticed that glibc-2.42 has broken TTY detection when running under QEMU user space emulation. For example, running the following on an amd64 machine:

```
$ docker run --rm -it --platform linux/arm64 cgr.dev/chainguard/wolfi-base:latest
test -t 0 && echo is a tty || echo not a tty
not a tty
```

Comparing this to the expected output which has a prompt and reports a TTY:

```
$ docker run --rm -it --platform linux/amd64 cgr.dev/chainguard/wolfi-base:latest
36e20e975d92:/# test -t 0 && echo is a tty || echo not a tty
is a tty
```

I ran the image on a full arm64 VM and it appears to be fine, which suggests the problem is limited to QEMU user space emulation.

Reverting back to glibc-2.41-r56 with the following Dockerfile is sufficient to fix the problem:

```Dockerfile
FROM cgr.dev/chainguard/wolfi-base:latest
RUN apk add --no-cache glibc=2.41-r56 ld-linux=2.41-r56 libcrypt1=2.41-r56
```

I'm not sure if this is a bug with glibc or QEMU, but I thought it was worth raising an issue here since others might run into the same bug.

robbiemcmichael:
Using `QEMU_STRACE=1` on the `test -t 0` command, I can see that my custom built glibc-2.41-r56 image uses the [`TCGETS` ioctl](https://www.man7.org/linux/man-pages/man2/TCSETS.2const.html):

```
docker run --rm -it --platform linux/arm64 -e QEMU_STRACE=1 wolfi:custom test -t 0
...
1 ioctl(0,TCGETS,0x00007f67de32f9a0) = 0 ({c_iflag = ICRNL|IXON,c_oflag = OPOST|ONLCR,c_cflag = B38400,CS8,CREAD,c_lflag = ISIG|ICANON|ECHO|ECHOE|ECHOK|ECHOCTL|ECHOKE|IEXTEN,c_cc = "",c_line = ''})
1 exit_group(0)
```

Trying the same thing with glibc-2.42-r0, I get this:


```
$ docker run --rm -it --platform linux/arm64 -e QEMU_STRACE=1 cgr.dev/chainguard/wolfi-base:latest test -t 0
...
1 ioctl(0,0x802c542a,0x4b8cf9f8) = -1 errno=25 (Not a tty)
1 exit_group(1)
```

[...]

PR #70485, opened 2025-10-31
Add "cross" support for local-wolfi and dev-container

smoser:
This allows you to run "cross" arch for 'local-wolfi', 'dev-container', and 'dev-container-wolfi'.

So if you're running x86_64:

    make ARCH=aarch64 local-wolfi
    % uname -m
    aarc64

If you're on aarch64:

    $ make ARCH=x86_64 local-wolfi
    % uname -m
    x86_64

These require you to have qemu-user and binfmt-misc set up.

The weirdness with --entrypoint and '-il' is somehow related the "cross-ness".  Both bash and sh seem to need it, or they do not go into interacive mode (and you wont see a prompt).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,6 +68,15 @@ WOLFI_REPO ?= https://packages.wolfi.dev/os
 WOLFI_KEY ?= https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
 BOOTSTRAP ?= no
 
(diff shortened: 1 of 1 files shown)


sergiodj (approved):

smoser:
for reference, https://github.com/wolfi-dev/os/issues/64231 has more context on the `-il` garbage.

PR #70538, opened 2025-11-02
openssl: add patch to block deprecated HMAC API in FIPS mode

xnox:
When using deprecated HMAC API, ensure that digest is from the FIPS
provider. This should block creating FIPS HMAC with unapproved
digests.

Improved implementation over previous attempt, see:
- https://github.com/wolfi-dev/os/pull/70102
- https://github.com/wolfi-dev/os/pull/70351

### DIFF
--- a/openssl.yaml
+++ b/openssl.yaml
@@ -1,7 +1,7 @@
 package:
   name: openssl
   version: "3.6.0"
-  epoch: 3
+  epoch: 4
   description: "the OpenSSL cryptography suite"
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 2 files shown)


Taffer (changes requested):
Is there a test we can add somewhere demonstrating the problem (node-fips segfaults) and that this patch fixes it?

We want to start "requiring" `apkregress` tests (see [this playbook](https://eng.inky.wtf/docs/teams/guarded-os/playbooks/testing-package/#testing-everything)) for packages that have lots of impact, like OpenSSL.

Taffer (approved):

xnox:
re-verified against enterprise-package main tests; and this presubmit catches things there.

PR #70700, opened 2025-11-02
New package openssl-hardened-dev

xnox:
Add a new package openssl-hardened-dev. It configures an OpenSSL
header that removes access to all deprecated APIs. Allow installing it
as a build-time dependency. This allows to simulate builds against
OpenSSL 3 that is configured with no-deprecated API/ABI, as well as
OpenSSL 4.

Use this new openssl-hardended-dev to compile apk-tools. This shows
all the build-time issues where deprecated APIs are used, and allows
to safely port apk-tools away from them. It remains API/ABI compatible
with the current OpenSSL 3 builds, and will enable to move to OpenSSL
4 ABI with a no-change rebuild.

### DIFF
--- a/apk-tools.yaml
+++ b/apk-tools.yaml
@@ -1,7 +1,7 @@
 package:
   name: apk-tools
   version: "2.14.10"
-  epoch: 8
+  epoch: 9
   description: "apk-tools (Wolfi package manager)"
   copyright:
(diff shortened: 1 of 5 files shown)


Taffer (approved):
This is a great idea, love it!

PR #70883, opened 2025-11-03
feat: new -jdk11 package variants.

Aditevil:
Creates JDK11 variant for newrelic-infrastructure-bundle.


nri-kafka.yaml
```diff
@@ -48,7 +66,19 @@ update:
...
+        version-flag: "-show_version"
+    - name: Test binary execution
+      uses: test/daemon-check-output
+      with:
+        start: nri-kafka
+        expected_output: |
+          com.newrelic.kafka
+          protocol_version
+          integration_version
+        error_strings: ""
```
rahulduvedi: Can probably omit this one? I mean `error_strings: ""`
Or we added this to skip some known error?
As giving expect error_strings it does check for the default ones.

nri-cassandra.yaml
```diff
@@ -47,7 +65,27 @@ update:
...
+        bins: nri-cassandra
+        version-flag: "-show_version"
+    - name: "Test binary execution"
+      uses: test/daemon-check-output
+      with:
+        start: |
+          nri-cassandra
+        expected_output: |
+          Connection refused to host
+        error_strings: |
```
rahulduvedi, line 91: Also just checking if we can omit this as then the pipeline will check for a range of default errors

nri-jmx.yaml
```diff
@@ -52,7 +70,38 @@ update:
...
+              beans:
+                - query: type=Memory
+                  attributes:
+                    - HeapMemoryUsage.Used
+          EOF
+        start: |
+          sh -c "COLLECTION_FILES=/tmp/jmx-test.yml nri-jmx"
+        expected_output: |
+          Connection refused to host
+        error_strings: |
```
rahulduvedi, line 107: sorry for the same comment, but want to kindly check if we can also remove it from here as well.

nri-cassandra.yaml
Aditevil: The tests will fail as it is, so wanted to ignore those error strings.

```
COLLECTION_FILES=/jmx-test.yml nri-jmx
[ERR] Failed to open JMX connection, error: failed to open JMX connection, error: connection error: can't connect to JMX server: 'service:jmx:rmi:///jndi/rmi://localhost:9999/jmxrmi', error: 'Failed to retrieve RMIServer stub: javax.naming.ServiceUnavailableException [Root exception is java.rmi.ConnectException: Connection refused to host: localhost; nested exception is: \n    java.net.ConnectException: Connection refused (Connection refused)]', Config: (Hostname: 'localhost', Port: '9999', IsJBossStandaloneMode: 'false', IsRemote: 'false', UseSSL: 'false', Username: '<EMPTY>', Password: '<EMPTY>', Truststore: '', TruststorePassword: '<EMPTY>', Keystore: '', KeystorePassword: '<EMPTY>', RequestTimeoutMs: '10000', Verbose: 'false'), Config: (Hostname: 'localhost', Port: '9999', IsJBossStandaloneMode: 'false', IsRemote: 'false', UseSSL: 'false', Username: '<EMPTY>', Password: '<EMPTY>', Truststore: '', TruststorePassword: '<EMPTY>', Keystore: '', KeystorePassword: '<EMPTY>', RequestTimeoutMs: '10000', Verbose: 'false')
```

rahulduvedi: Do you think we should document this?
So someday we don't remove this accidentally during a cleanup.

nri-kafka.yaml
```diff
@@ -48,7 +66,18 @@ update:
...
+        bins: nri-kafka
+        version-flag: "-show_version"
+    - name: Test binary execution
+      uses: test/daemon-check-output
+      with:
+        start: nri-kafka
+        expected_output: |
+          com.newrelic.kafka
+          protocol_version
+          integration_version
```
rahulduvedi, line 84: is it possible to add some post test for all packages? for more confidence.
(although seems a bit difficult here which is understandable if we cannot)

Aditevil: These are simple go binaries, basic tests should do fine here. I can claude it out and write complex tests but doesn't make sense tbh

nri-cassandra.yaml
Aditevil: added a comment.

nri-kafka.yaml
rahulduvedi: I can understand.
Thanks for adding comments.
This looks good to me
Thanks @Aditevil

rahulduvedi (approved):

PR #70926, opened 2025-11-03
Implement debian aggregate utility

ericsmalling:
- [x] This PR is marked as fixing a pre-existing package request bug: https://github.com/chainguard-dev/image-requests/issues/6939
- [x] REQUIRED - The version of the package is still receiving security updates (via Debian)


aggregate.yaml
```diff
@@ -0,0 +1,120 @@
...
+    to: mangled-package-version
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: https://salsa.debian.org/salvage-team/aggregate.git
+      tag: debian/${{vars.mangled-package-version}}
+      expected-commit: 74b302efd6b21d6d729e14aea5b76e25d89f1372
+
+  - runs: "for patchfile in $(cat debian/patches/series); do\n  patch -p1 < debian/patches/$patchfile    \ndone    \n"
```
ericsmalling: Is there a better way to do this?  the patches are all numerically prefixed so if our standard "patch" tooling has a way to run them all in order, that could work too.

jamonation: There is indeed!

```
  - uses: patch
    with:
      series: series
```

I don't know how it works with `debian/patches/series` and the subdirectory structure though. You could mv the whole thing to the root of the build tree any apply that way. See [openldap](https://github.com/wolfi-dev/os/tree/main/openldap-2.6) and its yaml manifest for an example.

ericsmalling: ok, will update this

jamonation (approved):
LGTM

PR #70952, opened 2025-11-03
glibc: Cherry-pick upstream fixes for 2.42 branch

sergiodj:


glibc.yaml
```diff
@@ -54,7 +54,37 @@ pipeline:
...
+        release/2.42/master/e34453cd6a8c592c325756ff3c7ac0afd3975cb4: AArch64: Fix SVE powf routine [BZ #33299]
+        release/2.42/master/1166170d95863e5a6f8121a5ca9d97713f524f49: libio: Define AT_RENAME_* with the same tokens as Linux
+        release/2.42/master/46b4e37c9e0619d0cf065ba207c29996b326a06f: nss: Group merge does not react to ERANGE during merge (bug 33361)
+        release/2.42/master/18fd689cdced8348e42991964557cddea0ba2dc5: nptl: Fix MADV_GUARD_INSTALL logic for thread without guard page (BZ 33356)
+        release/2.42/master/bf48b17a28066a54f172e0d63da9fc5dc60c6355: x86: Detect Intel Wildcat Lake Processor
+        release/2.42/master/ab8c1b5d62d7be2c3c23f535bea3d7fff19c53ae: x86: Detect Intel Nova Lake Processor
+        release/2.42/master/6de12fc9ad56bc19fa6fcbd8ee502f29b5170d47: aarch64: define macro for calling __libc_arm_za_disable
+        release/2.42/master/256030b9842a10b1f22851b1de0c119761417544: aarch64: clear ZA state of SME before clone and clone3 syscalls
+        release/2.42/master/71874f167aa5bb1538ff7e394beaacee28ebe65f: aarch64: tests for SME
```
xnox, line 87: We can also switch to build from the release/2.42 branch; pin by commit; and change update stanza to be periodic, such that we land smaller batches of fixes going forward. And automatic lifecycle would open the PR requests.

Also it will make the SBOM cleaner of just pointing at the 71874f commit.

sergiodj: The problem with that is that each glibc update needs to be thoroughly tested: I need to rebuild it, and then run reverse dependency tests on all arches, and then do full archive rebuilds (we *could* skip these, but I feel more comfortable making sure everything rebuilds OK), and then do world rebuilds for images...

Having the bot open automatic PRs for glibc would introduce the possibility of someone inadvertently approving it and then introducing breakages everywhere.

A compromise here would be to switch to build from the `release/2.42` branch, pin by commit, but **don't** enable auto updates.

xnox: well, you can have code owner set to yourself; and then set the update bot to poll at longer intervals (i cannot remember if monthly is supported).

If and when CVE happens, it might be easier to accept the auto-update bot PR as well.

sergiodj: Yeah, CODEOWNERS is already set to the OS team.  Let me think more about the implications of doing it like how you're suggesting.

sergiodj:
Test results:

* World rebuild for `wolfi`: OK
* World rebuild for `extras`: OK
* World rebuild for `enterprise`: OK
* `apkregress` on amd64: OK
* `apkregress` on arm64: OK

glibc.yaml
sergiodj: After discussing this with the team (and with @sil2100 in a call), here's what we decided:

1) We prefer the verbosity of the `cherry-pick` section, so we will keep things as is for now.
2) We agree with the SBOM should be improved in order to record cherry-picks; this is something that would benefit many other packages as well.
3) We're not comfortable enabling auto-updates for glibc, for the reasons I explained earlier.  So we will refrain from doing that.

We can revisit (1) in the coming cycles, but I prefer to not change the way we do things this cycle.

Taffer (approved):

PR #71002, opened 2025-11-04
Correcting file structure for apache-nifi-registry-toolkit

rahulduvedi:
Fixes for Toolkit subpackage:
- Fixed nested extraction using `bsdtar --strip-components=1` instead of `unzip`
- Before: `nifi-toolkit-registry-2.6.0/nifi-registry-toolkit-2.6.0/` (nested)
- After: `nifi-toolkit-registry-2.6.0/{bin,classpath,conf,lib}` (flat, matches upstream)

Also removed `curl` package which as added a convenience, [check here there is not runtime usage of curl ](https://github.com/search?q=repo%3Aapache%2Fnifi+path%3A%2F%5Enifi-registry%5C%2F%2F+curl&type=code)

Removed creation of dir's on to the image level for consistency.

### DIFF
--- a/apache-nifi-registry.yaml
+++ b/apache-nifi-registry.yaml
@@ -9,7 +9,6 @@ package:
     runtime:
       - bash # Required by scripts with #!/bin/bash shebang
       - coreutils # Provides readlink, dirname, basename, pwd, uname, expr used by nifi-registry.sh and cli.sh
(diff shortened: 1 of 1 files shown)


AnmolVirdi (approved):
LGTM.

PR #71031, opened 2025-11-04
libmspack: update upstream source, enable auto updates

aborrero:
And while at it, add some more tests.

### DIFF
--- a/libmspack.yaml
+++ b/libmspack.yaml
@@ -2,7 +2,7 @@
 package:
   name: libmspack
   version: "1.11"
-  epoch: 0
+  epoch: 1
   description: Library for Microsoft CAB compression formats
   copyright:
     - license: LGPL-2.1-only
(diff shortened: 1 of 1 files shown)


aborrero:
lint seems to be failing for unrelated reasons:

```
2025/11/04 16:57:42 ERRO Package: redka: [background-process-without-redirect]: background process missing output redirect: redka& (WARNING)
2025/11/04 16:57:42 ERRO Package: ruby3.2-aws-eventstream: [update-disabled-reason]: auto-update is disabled but no reason is provided (WARNING)
2025/11/04 16:57:42 ERRO Package: ruby3.2-aws-partitions: [update-disabled-reason]: auto-update is disabled but no reason is provided (WARNING)
```

aborrero:
wait, this might be wrong as well.

vaikas:
@aborrero Yeah, this should not be using the 1.11 version train. See: https://github.com/wolfi-dev/os/pull/70412 which I believe is the right way to fix this file instead. It was failing before the package update check because the tooling that checks it needs to be updated.

PR #68945, opened 2025-10-15
docker: introduce additional tests and drop dockerd subpackage

aborrero:
Introduce additional tests.

While at it, drop the dockerd subpackage, which is empty. No other
package or image depends on it. The files that were supposed to be
included in the dockerd subpackage are actually in the main docker
package.

Signed-off-by: Arturo Borrero Gonzalez <arturo.borrero@chainguard.dev>

PR #71048, opened 2025-11-04
docker - return of dockerd

smoser:
This adds back the dockerd package that was dropped under https://github.com/wolfi-dev/os/pull/68945.

Other changes:
 * add docker-init package
 * build with VERSION="${{package.version}}" so that --version has reasonable output. It was showing: Docker version dev, build f8215cc2..-unsupported Now we see: Docker version 28.5.1, build f8215cc2...-unsupported
 * add busybox as a dependency for shell entrypoints
 * update the smoke check test. It doesn't really make sense to '--version' and '--help' here, since we're just checking that these programs are available. In fact,
     - docker-proxy --help is undocumented or wrong
     - docker-init --version would show tini-static's version, which would be confusing. - docker is from docker-cli, so --version is not useful output.

### DIFF
--- a/docker.yaml
+++ b/docker.yaml
@@ -1,38 +1,17 @@
 package:
   name: docker
   version: "28.5.1"
-  epoch: 2
+  epoch: 3
   description: A meta package for Docker Engine and Docker CLI
   copyright:
     - license: Apache-2.0
   dependencies:
     runtime:
-      - btrfs-progs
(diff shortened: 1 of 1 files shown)


smoser:
@EyeCantCU i specifically would like input on the -init package... i'm not sure what the point  is and if that should be a dependency of `docker` or not... please review.

89luca89 (approved):

PR #71079, opened 2025-11-05
resize Keycloak 26.4 package

yashpimple:
The keycloak-26.4 package was shipping two copies of the Keycloak distribution. This PR removes the versioned duplicate and keeps the canonical path used by upstream.


AnmolVirdi (approved):
LGTM. Removing duplication of keycloak files.
The CVEs reported(newly published) are already on Sustaining board so we can skip this check for now.

PR #71096, opened 2025-11-05
dimitriledkov/vms 112 configure and build util linux with selinux

xnox:
- **Fix FTBFS due to incorect libs in ncursesw6-config**
  

- **util-linux: enable selinux support**
  Add libselinux-dev build dependency.
  
Fixes VMS-112

### DIFF
--- a/ncurses.yaml
+++ b/ncurses.yaml
@@ -1,7 +1,7 @@
 package:
   name: ncurses
   version: 6.5_p20251025
-  epoch: 0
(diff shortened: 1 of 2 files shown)


markusboehme (approved):
The ncurses change looks fine, but unrelated. Please make sure to use a merge commit to preserve the individual change in its context.

xnox:
well, util-linux wouldn't build without that change =) so it was necessary

PR #71129, opened 2025-11-05
Builds opencv with opencv-contrib modules

srishtih:
Builds opencv with modules from opencv-contrib and adds improved tests.


opencv.yaml
```diff
@@ -104,6 +123,14 @@ pipeline:
   - uses: strip
 
 subpackages:
+  # Addresses the dependency sharing issue between opencv and libopencv sub-packages
+  - name: opencv-libtbb
+    description: "Vendored Intel TBB library for OpenCV"
+    pipeline:
+      - runs: |
+          mkdir -p ${{targets.subpkgdir}}/usr/lib
+          mv ${{targets.destdir}}/usr/lib/libtbb*.so* ${{targets.subpkgdir}}/usr/lib/
```
murraybd: There is a new split/lib pipeline which would reduce the number of steps here. Additionally, you should add the ldd-check test.

murraybd (dismissed):
Please add the test/tw/ldd-check and try using split/lib.

opencv.yaml
srishtih: I'm thinking it might be cleaner if we did not split the libopencv files into seperate sub-packages. Then, we wouldn't need a separate `opencv-libtbb` package. The package specification in the image build will be readable and there's a smaller chance of missing something.

klrmngr (approved):
lgtm

kranurag7 (approved):

PR #71209, opened 2025-11-06
kube-logging-operator: add fluentd-outputs subpackage with all plugins

rahulduvedi:
Adds kube-logging-operator-fluentd-outputs subpackage matching upstream `ghcr.io/kube-logging/logging-operator/fluentd:6.1.0-full`

  Implementation

  - Installs fluentd 1.19.0 + 40 output/filter plugins to /usr/local/bundle via gem install
  - Includes all plugins from upstream outputs/Gemfile and filters/Gemfile
  - Sets GEM_HOME=/usr/local/bundle to avoid conflicts with system Ruby gems
  - Runtime dependencies: ruby-3.2, geoip, libmaxminddb

### DIFF
--- a/kube-logging-operator.yaml
+++ b/kube-logging-operator.yaml
@@ -13,7 +13,6 @@ environment:
       - automake
       - build-base
       - busybox
-      - ca-certificates-bundle
       - geoip-dev
       - git
       - libmaxminddb-dev
@@ -79,8 +78,6 @@ subpackages:
       no-provides: true
     dependencies:
       runtime:
-        - geoip
-        - libmaxminddb
         - ruby-3.2
     pipeline:
       - runs: |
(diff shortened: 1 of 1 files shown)


klrmngr:
how are we planning on building the image? are we adding this new package as an extra package? if so, is there any way we can avoid duplicating transitive deps between the packages we build from source and the packages we gem install?

rahulduvedi:
> how are we planning on building the image? are we adding this new package as an extra package? if so, is there any way we can avoid duplicating transitive deps between the packages we build from source and the packages we gem install?

Hello @klrmngr I am planning to add this as a extra package in the image.
I am not so sure about the latter part, there were a lot of conflicts while building this package and thats why I am installing `fluentd` first in the sub package rather than adding it as a runtime dep.
Do you have any other suggestions regarding this that can help?

klrmngr:
have you tried installing this at the image level yet? when i was doing something similar i ran into conflicts then, unsure if this method avoids those

klrmngr (dismissed):
package looks to be good, but may need to change this again when building the image
tentative lgtm

rahulduvedi:
Hello @klrmngr I was able to make image with this subpackage and I have verified all the 40 plugins are now present.
Requesting a re-review 🙏 
@klrmngr

klrmngr (approved):

PR #71218, opened 2025-11-06
fix(jistsucom-bulker): Update to upstream beta release 2.11.913-beta.20251103114200.fb5fbbf

Ankush-Pathak:
jistsucom-bulker is currently FTBFS. See failing PR: https://github.com/wolfi-dev/os/pull/70816
Upgrading to latest upstream release fixes the build failure. However, the only concern is that the latest release tag indicates it is meant to be a "beta" release. Though, there doesn't seem to be any documented distinction between non-beta and beta releases upstream.

### DIFF
--- a/jitsucom-bulker.yaml
+++ b/jitsucom-bulker.yaml
@@ -1,7 +1,7 @@
 package:
   name: jitsucom-bulker
-  version: "2.11.0"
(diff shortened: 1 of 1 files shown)


aborrero (approved):
LGTM.

PR #71265, opened 2025-11-06
chore(gcc): Add versioned provides

EyeCantCU:
There may be situations (such as those where applications are built with the NVCC which supports a limited range of GCC versions) that we may want to pin an application to what is currently the latest version of GCC

This also adds an additional `-default` provider for GCC so that packages that are intentionally pinned to what was the latest version of GCC continue to use that version by default instead of the latest version

### DIFF
--- a/gcc.yaml
+++ b/gcc.yaml
@@ -1,7 +1,7 @@
 package:
   name: gcc
   version: "15.2.0"
-  epoch: 3
+  epoch: 4
   description: "the GNU compiler collection"
   copyright:
     - license: GPL-3.0-or-later WITH GCC-exception-3.1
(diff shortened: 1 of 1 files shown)


sergiodj (approved):
That's great, thanks @EyeCantCU.  I've had this in my TODO list for the next GCC release, but we might as well do it right now.

I'd like to have stricter runtime dependency versioning for `gfortran` as well, but we can do that in another PR.  Thanks!

gcc.yaml
```diff
@@ -1,14 +1,17 @@
...
+  epoch: 4
   description: "the GNU compiler collection"
   copyright:
     - license: GPL-3.0-or-later WITH GCC-exception-3.1
   resources:
     cpu: 16
     memory: 16Gi
   dependencies:
+    provides:
+      - gcc-${{vars.major-version}}=${{package.full-version}}
```
dannf, line 13: Thanks! I feel like we should also include gcc-15/g++-15 executables if we provide this.

EyeCantCU: After running into another problem today, I couldn't agree more. Will get a PR up in a moment

PR #71476, opened 2025-11-08
add nocaps haproxy

jamonation:
Creates an `haproxy-3.2-nocaps` package without net_bind capability on the binary. Doing it this way means not breaking anyone who relies on or has workarounds for the existing binary ***with*** the capability set on it.

### DIFF
--- a/haproxy-3.2.yaml
+++ b/haproxy-3.2.yaml
@@ -1,7 +1,7 @@
 package:
   name: haproxy-3.2
   version: "3.2.8"
-  epoch: 2
+  epoch: 1
   description: "A TCP/HTTP reverse proxy for high availability environments"
   copyright:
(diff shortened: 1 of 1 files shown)


haproxy-3.2.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: haproxy-3.2
   version: "3.2.8"
-  epoch: 0
+  epoch: 2
```
murraybd: It looks like the epoch got double bumped somehow.

jamonation: Ah yeah leftover from main version bump in my first commit, nice spotting

jamonation:
```suggestion
  epoch: 1
```

murraybd (approved):

PR #71488, opened 2025-11-08
Fix apk-tools database path

### LINKED ISSUE #66951: apk-tools: package should create `/lib/apk/db` instead of `/var/lib/apk`
This doesn't break anything but I thought that I should point out something about the packaging of "apk-tools". 

I was investigating why Wolfi has the `/var/lib/apk` directory but empty with no files inside, while Alpine does not have the directory at all. I traced it to the `install -d` at https://github.com/wolfi-dev/os/blob/85c498ab0a477dcc8d071420561391b957471475/apk-tools.yaml#L48

Checking against Alpine's package of apk-tools, it seems that as of [around package version 2.14.1-r1 in aports](https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/62836) they switched their package to create `/lib/apk/db` instead of `/var/lib/apk`. Apparently apk shifted away from `/var/lib/apk` quite a long time ago, based on the commit referenced in that merge request.

I think this means that Wolfi's apk-tools package could follow the same change, since I don't see any Wolfi-specific patch to apk that contradicts this.

sDaman830:
Change apk database location from /var/lib/apk to /lib/apk/db to align with modern apk standards and Alpine Linux.

The old /var/lib/apk path is deprecated. apk has used /lib/apk/db since around version 2.14.1. This change removes the creation of an empty, unused /var/lib/apk directory.

Fixes #66951

### Changes
- Changed directory creation from `/var/lib/apk` to `/lib/apk/db` 
- Bumped epoch from 9 to 10


cmwilson21:
Thanks for the contribution!

PR #71489, opened 2025-11-08
az: no change rebuild for compatibility with changes in py3-cryptography dependencies.

javacruft:
py3-cryptography recently added typing-extensions as a dependency - rebuild to ensure that its not re-installed as part of the pip install.

We need to revisit this package - pip installing into /usr is going to be brittle.

### DIFF
--- a/az.yaml
+++ b/az.yaml
@@ -1,7 +1,7 @@
 package:
   name: az
   version: "2.79.0"
-  epoch: 0
+  epoch: 1
   description: Azure CLI
   copyright:
     - license: MIT


kranurag7 (approved):
tested with `LOCAL_WOLFI_EXTRA_REPO=https://apk.cgr.dev/wolfi-presubmit/fb27d9b09850e97bdcffaf9a7b4be65016fd6dda make local-wolfi` and this looks good to land. 

if more are impacted then we might want to revisit https://github.com/wolfi-dev/os/pull/71490 down the road.

PR #71584, opened 2025-11-10
py3-transformer: drop package

aborrero:
It has been relocated to the enterprise repository for dependency reasons (it depends
on a package on the enterprise repository).


aborrero:
going on a different route, see https://github.com/chainguard-dev/enterprise-packages/pull/38201#pullrequestreview-3444573645

PR #71595, opened 2025-11-10
fix(flannel-cni-plugin): add busybox runtime dep and add /flannel symlink

rakshitgondwal:
add busybox as a runtime dependency and add /flannel symlink as both are expected by the upstream helm chart
https://github.com/flannel-io/flannel/blob/97b6771a642511a676f51df17184f6852b425df9/chart/kube-flannel/templates/daemonset.yaml#L40

### DIFF
--- a/flannel-cni-plugin.yaml
+++ b/flannel-cni-plugin.yaml
@@ -1,10 +1,14 @@
 package:
   name: flannel-cni-plugin
   version: 1.8.0.2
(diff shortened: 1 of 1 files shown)


rahulduvedi (approved):

rahulduvedi:
Changes look good, Thanks for the work on this, 
Busybox is indeed needed for this.

PR #71699, opened 2025-11-11
fix(libjxl): add docs package to split off man pages

astrojerms:
Currently libjxl is failing to install due to some conflicting man pages with brotli. This splits off the man pages into a -docs package.


kranurag7 (commented):
we can also add a test/docs check in there

kranurag7 (approved):

PR #71727, opened 2025-11-11
libpipeline: switch to git-checkout

cjwatson:


justinvreeland (approved):
Savana does become inaccessible a bit to frequently : (

justinvreeland (approved):

PR #71742, opened 2025-11-11
py3-{pip,setuptools}: Restore provides + priorities

javacruft:
Restore the provider priorites (but not the py3-<name> provides) to
ensure that apko makes the right choices when picking py3.XX-<name>
pacakges.

### DIFF
--- a/py3-pip.yaml
+++ b/py3-pip.yaml
@@ -1,10 +1,12 @@
 package:
   name: py3-pip
   version: "25.3"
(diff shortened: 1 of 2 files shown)


Taffer (dismissed):

Taffer (dismissed):

Taffer (approved):

philroche:
post merge +1

PR #71759, opened 2025-11-11
fix(brotli): include versioned `.so` files on main brotli package (via runtime dependencies)

tulilirockz:
Not having this makes it so packages that depend on `brotli` at runtime can't load brotli libraries properly


Taffer (approved):
LGTM

PR #71795, opened 2025-11-12
dotnet-10: New package

sergiodj:
Add .NET 10 to the archive.

Some changes from version 9 that are worth mentioning:

- `CFLAGS` is set to `-Wno-error=used-but-marked-unused` due to compilation errors. (thanks @EyeCantCU)
- The `_FORTIFY_SOURCE=3` hardening flag had to be disabled because of compilation errors generated due to the fact that the runtime build is not optimized. (thanks @EyeCantCU)
- `file` had to be added as a B-D.
- The `netstandard-10-targeting-pack` subpackage had to be dropped because upstream is apparently not providing a built-in NETStandard anymore.  I couldn't find much information about this other than https://github.com/dotnet/dotnet/pull/2402 and its related changes, but I confirmed that Fedora's dotnet 10 build also removes their netstandard targeting pack subpackage.
- An upstream fix had to be backported to make the build pass. (thanks @pethin)
- We're now passing `--branding rtm` to `build.sh` when building. (thanks @pethin)

### DIFF
--- a/dotnet-10.yaml
+++ b/dotnet-10.yaml
@@ -1,0 +1,294 @@
+package:
+  name: dotnet-10
+  version: "10.0.100"
+  epoch: 0
+  description: ".NET ${{vars.major-version}} host"
+  copyright:
+    - license: MIT
+  resources:
+    cpu: 32
+    memory: 64Gi
+  dependencies:
+    runtime:
+      - icu
+    provider-priority: ${{vars.major-version}}
+    provides:
+      - dotnet=${{package.full-version}}
+
+vars:
+  # This is the major.minor version of the .NET Standard
+  # targeting pack. We validate this version is correct
+  # below, but as the version progresses, this will need
+  # to be manually bumped.
+  netstandard-version: '2.1'
+  clang-version: 20
+
+var-transforms:
+  - from: ${{package.version}}
+    match: ^(\d+)\.\d+\.\d+$
+    replace: "$1"
+    to: major-version
+
+environment:
+  environment:
+    DOTNET_CLI_TELEMETRY_OPTOUT: 1
+    CFLAGS: "-Wno-error=used-but-marked-unused"
+  contents:
+    packages:
+      - bash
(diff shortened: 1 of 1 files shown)


pethin:
Build is erroring with
```
/home/build/eng/common/tools.sh: line 1: tempDir: unbound variable
failed to build package: unable to run package dotnet-10 pipeline: unable to run pipeline: exit status 1
failed to build: command melange failed: exit status 1
build failed: command melange failed: exit status 1
```

Root cause is this issue: https://github.com/dotnet/source-build/issues/5403

Temporary solution is to modify [eng/source-build-toolset-init.sh‎](https://github.com/dotnet/dotnet/pull/3248/files#diff-6be68ffe91d15b95cf60db13fdc3e07324c4b4cb7665617de775a66ebe1c70a5)

pethin:
`--branding rtm` must be passed to `./build.sh` to produce artifacts with the .NET 10 GA version strings. See https://github.com/dotnet/dotnet/releases/tag/v10.0.100

sergiodj:
> `--branding rtm` must be passed to `./build.sh` to produce artifacts with the .NET 10 GA version strings. See https://github.com/dotnet/dotnet/releases/tag/v10.0.100

Thanks.  I'm back on working on this now.  I'm doing some local tests and will update the PR accodingly.

EyeCantCU (approved):
Great work!

pethin:
I'm not sure why the .NET Standard targeting pack was removed either. The only documentation I found was https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-1-0#when-to-target-netx0-vs-netstandard

> App components
> If you're using libraries to break down an application into several components, we recommend you target net10.0.

> Reusable libraries
> If you don't need to support .NET Framework, you could target .NET Standard 2.1 or .NET 10. We recommend you skip .NET Standard 2.1 and go straight to .NET 10.

PR #71841, opened 2025-11-12
py3-supported-python - Add python 3.14

smoser:
Sparked by the addition of 3.14 to py3-pip in
https://github.com/wolfi-dev/os/pull/68550

This seems like the right thing to do.

### DIFF
--- a/py3-build.yaml
+++ b/py3-build.yaml
@@ -1,7 +1,7 @@
 package:
   name: py3-build
   version: "1.3.0"
-  epoch: 0
+  epoch: 1
   description: A simple, correct Python build frontend
   copyright:
     - license: MIT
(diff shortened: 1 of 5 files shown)


kranurag7 (dismissed):
looks good in theory but is this a hint that we'll multiversion again ;) just kidding..

kranurag7:
@smoser some previous work here https://github.com/wolfi-dev/os/pull/68535 

Probably the same, last I remember is I wasn't able to debug what's failing because elastic build page became too long. (via phone, can share more details in a bit)

smoser:
I'm going to close this for now, as not interested in resolving the circular dependencies. :-(

but I do that we want 'py3-supported-python' with 3.14

PR #71858, opened 2025-11-12
remove /opt/cni/bin symlink from flannel-cni-plugin

rakshitgondwal:
remove `/opt/cni/bin` symlink from `flannel-cni-plugin` for helm upstream helm chart compatibility.
upstream's helm chart is trying to cp the flannel binary from `/flannel` to `/opt/cni/bin` 
https://github.com/flannel-io/flannel/blob/97b6771a642511a676f51df17184f6852b425df9/chart/kube-flannel/templates/daemonset.yaml#L47

This cp is failing since the binary already exists over there.

### DIFF
--- a/flannel-cni-plugin.yaml
+++ b/flannel-cni-plugin.yaml
@@ -1,7 +1,7 @@
 package:
   name: flannel-cni-plugin
   version: 1.8.0.2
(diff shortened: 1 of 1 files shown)


klrmngr (approved):
lgtm

PR #71921, opened 2025-11-13
Fix installation path for NETStandard.Library.Ref

sergiodj:
Currently `dotnet-{8,9}` are installing the `NETStandard.Library.Ref` pack under `/usr/share/dotnet/packs/2.1.0/`, i.e., the `NETStandard.Library.Ref/` component of the path is being inadvertently stripped because of globbing rules.  We should fix this by explicitly referring to the NETStandard directory that will be moved to the subpackage.  Let's also take this opportunity to fully create the `/usr/share/dotnet/packs/` directories.

### DIFF
--- a/dotnet-8.yaml
+++ b/dotnet-8.yaml
@@ -1,7 +1,7 @@
 package:
   name: dotnet-8
   version: "8.0.122"
-  epoch: 0
+  epoch: 1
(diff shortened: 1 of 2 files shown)


EyeCantCU (approved):
Thanks for fixing this!

PR #71929, opened 2025-11-13
fix(clamav): enable TCP port + add clamd-milter defaults

sanjayk0508:
have updated `TCPSocket` to enable port 3310 and added default configs for `clamav-milter.conf` matching upstream

ref - https://github.com/Cisco-Talos/clamav-docker/blob/main/clamav/1.5/alpine/Dockerfile#L61-L81

### DIFF
--- a/clamav-1.5.yaml
+++ b/clamav-1.5.yaml
@@ -171,8 +171,8 @@ subpackages:
 
           mv "${{targets.destdir}}"/etc/clamav/clamd.conf.sample "${{targets.subpkgdir}}"/etc/clamav/clamd.conf
 
-          # Enable TCPSocket 3310 (required for remote TCP access)
-          sed -i 's/^#TCPSocket.*/TCPSocket 3310/' "${{targets.subpkgdir}}"/etc/clamav/clamd.conf
+          # Enable TCPSocket 3310 and set TCPAddr
+          sed -i 's/^#TCPSocket.*/TCPSocket 3310/; s/^#TCPAddr localhost/#TCPAddr 0.0.0.0/' "${{targets.subpkgdir}}"/etc/clamav/clamd.conf
     dependencies:
       provides:
         - clamav-daemon=${{package.full-version}}
@@ -292,6 +292,15 @@ subpackages:
 
           mv "${{targets.destdir}}"/etc/clamav/clamav-milter.conf.sample \
             "${{targets.subpkgdir}}"/etc/clamav/clamav-milter.conf
+
+          # Configure milter defaults
+          sed -i -e "s:^\(Example\):\# \1:" \
+            -e "s|.*\(MilterSocket\) .*|\1 inet:7357|" \
+            -e "s:.*\(User\) .*:\1 clamav:" \
+            -e "s:^\#\(LogFile\) .*:\1 /var/log/clamav/milter.log:" \
+            -e "s:^\#\(LogTime\).*:\1 yes:" \
+            -e "s|.*\(ClamdSocket\) .*|\1 unix:/run/clamav/clamd.sock|" \
+            "${{targets.subpkgdir}}"/etc/clamav/clamav-milter.conf
     description: ClamAV milter
     test:
       pipeline:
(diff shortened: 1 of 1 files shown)


clamav-1.5.yaml
```diff
@@ -170,6 +170,9 @@ subpackages:
           mv "${{targets.destdir}}"/usr/bin/clamconf "${{targets.subpkgdir}}"/usr/bin/
 
           mv "${{targets.destdir}}"/etc/clamav/clamd.conf.sample "${{targets.subpkgdir}}"/etc/clamav/clamd.conf
+
+          # Enable TCPSocket 3310 and set TCPAddr
+          sed -i 's/^#TCPSocket.*/TCPSocket 3310/; s/^#TCPAddr localhost/#TCPAddr 0.0.0.0/' "${{targets.subpkgdir}}"/etc/clamav/clamd.conf
```
tulilirockz: I think we should do this to the `/etc/clamd/clamd.conf.sample` file instead as that one is being moved here anyways

clamav-1.5.yaml
```diff
@@ -98,7 +98,7 @@ pipeline:
         -e "s:.*\(User\) .*:\1 clamav:" \
         -e "s:^\#\(LogFile\) .*:\1 /var/log/clamav/clamd.log:" \
         -e "s:^\#\(LogTime\).*:\1 yes:" \
-            -e "s:^\#\(AllowSupplementaryGroups\).*:\1 yes:" \
+        -e "s:^\#\(AllowSupplementaryGroups\).*:\1 yes:" \
         "${{targets.destdir}}"/etc/clamav/clamd.conf.sample
```
tulilirockz, line 102: Here
```suggestion
        "${{targets.destdir}}"/etc/clamav/clamd.conf.sample
        
        sed -i 's/^#TCPSocket.*/TCPSocket 3310/; s/^#TCPAddr localhost/#TCPAddr 0.0.0.0/' "${{targets.destdir}}"/etc/clamav/clamd.conf.sample
```

clamav-1.5.yaml
tulilirockz: https://github.com/wolfi-dev/os/pull/71929/files#r2535214083

sanjayk0508: tried changing it in `/etc/clamd/clamd.conf.sample` itself, but not sure why the config doesn't change, im assuming when we move from `destdir` to `subpkgdir` it may copy instead of move (preserving the original), nginx does the same as well, ref - https://github.com/wolfi-dev/os/blob/main/nginx-mainline.yaml#L166-L171

tulilirockz:
The modifications I made seem to be valid:
```
feb695bf356b:/tmp/etc/clamav# cat clamd.conf  | grep 3310
TCPSocket 3310
feb695bf356b:/tmp/etc/clamav#
```

clamav-1.5.yaml
```diff
@@ -362,6 +365,16 @@ subpackages:
...
+          # init scripts process CLAMD_CONF_* env vars which disables TCP socket
+          # append TCP at end of script (after all config processing) to ensure defaults remain enabled
+          for script in "${{targets.contextdir}}"/init "${{targets.contextdir}}"/init-unprivileged; do
+            cat >> "$script" << 'EOFPATCH'
+          # Ensure TCP socket enabled (override any CLAMD_CONF_* env var changes)
+          sed -i "s|.*TCPSocket.*|TCPSocket 3310|" /etc/clamav/clamd.conf 2>/dev/null || true
+          sed -i "s|.*TCPAddr.*|#TCPAddr 0.0.0.0|" /etc/clamav/clamd.conf 2>/dev/null || true
+          sed -i "s|#TCPAddr localhost|#TCPAddr 0.0.0.0|g" /etc/clamav/clamd.conf 2>/dev/null || true
+          EOFPATCH
+          done
```
tulilirockz: Why do we need to do this though?

tulilirockz:
Seems to be working great!


```bash
# Without modifications
adb779a39e8c:/work/packages# echo "PING" | nc localhost 3310
adb779a39e8c:/work/packages#
```

```bash
adb779a39e8c:/work/packages# echo "PING" | nc localhost 3310
PONG
adb779a39e8c:/work/packages#
```

clamav-1.5.yaml
tulilirockz: Turns out theres no need for this.

tulilirockz (approved):

Aditevil (approved):

PR #71942, opened 2025-11-13
btrfs-progs: Use OpenSSL, ver-check and help-check

Taffer:
OpenSSL is our preferred route to FIPS, nice of `btrfs-progs` to support various crypto implementations!

Also changes the tests to use ver-check and help-check.


btrfs-progs.yaml
```diff
@@ -90,18 +90,12 @@ update:
...
-        btrfsck --help
-        btrfstune --help
-        fsck.btrfs --version
-        mkfs.btrfs --version
-        btrfs --help
-        fsck.btrfs --help
-        mkfs.btrfs --help
+    - uses: test/tw/ver-check
+      with:
+        bins: btrfs mkfs.btrfs
```
murraybd, line 95: What happened to fsck.btrfs? (I realize some of these don't actually return a version but it'd be good to know for sure!)

murraybd (changes requested):

btrfs-progs.yaml
Taffer: It doesn't output anything useful with `-v`, `-V`, `--version`, just a usage message.

Taffer: it was intentional, `ver-check` failed on it :grin:

murraybd (approved):

PR #71222, opened 2025-11-06
chore: add dotnet-sdk provides to dotnet-x

kranurag7:
dotnet-sdk is used by gdal config and we had that provides till
dotnet-7, but the same is not there for dotnet-{8,9} and new builds
of gdal will use old deps.

this adds dotnet-sdk to new versions of dotnet

PR #71951, opened 2025-11-13
Add `provides` for `dotnet-sdk` versions

sergiodj:
Currently, doing `apk add dotnet-sdk` will incorrectly install `dotnet-7-sdk`.  This is incorrect and we don't even publish `dotnet-7` anymore.  Fix by adding a virtual `provides` to each `dotnet-X-sdk` version.

### DIFF
--- a/dotnet-10.yaml
+++ b/dotnet-10.yaml
@@ -1,7 +1,7 @@
 package:
   name: dotnet-10
   version: "10.0.100"
-  epoch: 0
+  epoch: 1
   description: ".NET ${{vars.major-version}} host"
   copyright:
     - license: MIT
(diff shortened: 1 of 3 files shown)


sergiodj:
There's been another attempt at doing this here: #71222 

Apparently there's a co-installability problem that we need to solve first (see #65386).

sergiodj:
FWIW `dotnet-8-sdk` isn't currently co-installable with `dotnet-9-sdk`.  So I don't know if there's really an issue here.

sergiodj:
#71978 should fix the existing co-installability problems with `dotnet-*-sdk`, which will hopefully allow us to move forward here.

sergiodj:
We won't be adding virtual provides for dotnet subpackages anymore.  Closing this one.

Cc @kranurag7

PR #72064, opened 2025-11-14
docker - restore busybox

Aditevil:
We moved `busybox` to `dockerd-oci-entrypoint` since it wasn’t directly required by `docker` itself and was only needed to satisfy the `dockerd-oci-entrypoint` subpackage. However, our existing packages and images still assume that `busybox` is provided by `docker` package.


smoser (approved):
thank you.

PR #72193, opened 2025-11-17
Remove libretls as an unused depedency

xnox:
libretls is a fork of openssl. And yet, tofu-controller and py3-xmlsec
seem to use both libretls and openssl at the same time.

Around alpine 3.15 time, it did choose to often use libretls, but has
since switched to default to openssl. (See all the commits in aports
that remove usage of libretls).

For tofu-controller, upstream also agrees that the libretls dependency is potentially redundant:
- https://github.com/flux-iac/tofu-controller/pull/1634

Try to build these packages without libretls, which is duplicate of
openssl, likely unused and does not support FIPS.

For py3-xmlsec the dep is build-time only, which didn't generate any runtime dependencies, and thus is redundant.

`make image/tofu-controller image/tofu-controller-fips` passed with this change


sergiodj (approved):

kranurag7 (approved):
> For py3-xmlsec the dep is build-time only, which didn't generate any runtime dependencies, and thus is redundant.

affirmative. thank you.

PR #72340, opened 2025-11-19
tomcat-11.0: remove redundant tests and refactor building to use more variables

tulilirockz:
This is mostly so that this is easier to maintain in the future, goes with other tomcat versions.

### DIFF
--- a/tomcat-11.0.yaml
+++ b/tomcat-11.0.yaml
@@ -1,7 +1,7 @@
 package:
   name: tomcat-11.0
   version: "11.0.14"
-  epoch: 0
+  epoch: 1
   description: Apache Tomcat Web Server
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


tomcat-11.0.yaml
```diff
@@ -40,42 +36,33 @@ pipeline:
...
+      tee build.properties <<EOF
       skip.installer=true
       base.path=$PWD
       compile.debug=false
       EOF
 
 subpackages:
   - range: openjdk-versions
-    name: ${{package.name}}-${{range.value}}
+    name: ${{package.name}}-openjdk-${{range.key}}
```
Aditevil, line 47: whats the in-degree and out-degree of this node?

tulilirockz: Doesn't change anything, we are basically jsut changing openjdk-(version) to openjdk-(version). I'm just making it a bit more explicit since this is a package name

Aditevil (approved):

PR #72374, opened 2025-11-19
php-8.5: New package

sergiodj:
The main change here is that upstream made opcache built-in, which means that we no longer have to build it as a shared library.

### DIFF
--- a/php-8.5.yaml
+++ b/php-8.5.yaml
@@ -1,0 +1,496 @@
+package:
+  name: php-8.5
+  version: "8.5.0"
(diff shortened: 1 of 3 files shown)


sergiodj:
Upstream hasn't made the official announcement yet, but the `php-8.5.0` tag already exists.  I'll remove the draft state once the release is official.

sergiodj:
Upstream has just officially announced PHP 8.5.

justinvreeland (approved):

PR #72426, opened 2025-11-20
kaniko: Improve baselayout patch for better compatibility

markusthoemmes:
This adds a couple of more files to remove to the existing patch fixing usrmerge and baselayout for improves kaniko compatibility. In particular, it removes /etc/protocols, /etc/services and /etc/shells to avoid potential conflicts on those files.

It also slightly reworks the patch to surface any failure during those operations as warnings properly and deduplicates some of the code.

### DIFF
--- a/kaniko.yaml
+++ b/kaniko.yaml
@@ -1,7 +1,7 @@
 package:
   name: kaniko
   version: "1.25.5"
-  epoch: 0 # GHSA-pwhc-rpq9-4c8w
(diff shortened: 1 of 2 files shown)


markusthoemmes:
The actual diff can be better seen in https://github.com/markusthoemmes/kaniko/commit/678e708e598770f5a2aed8f3c1ddc7ad5920e977.

priyawadhwa (dismissed):

kranurag7 (approved):

PR #73067, opened 2025-11-24
opentelemetry: re-add shared libraries

justinvreeland:
These were removed inadvertantly in https://github.com/wolfi-dev/os/pull/35855/

No one is actually using them (I can see nothing depending on their .so files) but it is super weird they're not there.

### DIFF
--- a/opentelemetry-cpp.yaml
+++ b/opentelemetry-cpp.yaml
@@ -1,7 +1,7 @@
 package:
   name: opentelemetry-cpp
   version: "1.23.0"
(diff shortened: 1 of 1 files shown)


murraybd (approved):

justinvreeland:
I think I probably need to update this to use the library versioning stuff as well.

justinvreeland:
> I think I probably need to update this to use the library versioning stuff as well.

Nope, they're not versioned. Static libraries appear unchanged as well.

PR #73072, opened 2025-11-24
test: increase coverage of kubernetes operators

maxgio92:
This commit adds tests of the operators to ensure that the manager and possibly also the worker reconcilers start as expected. The tests run against kwok, so since there is no kubelet running, tests go as deep as possible.

See also:
* #73096
* #73097
* #73098
* #73099


kranurag7 (dismissed):

maxgio92:
Better to split into different PRs.

kranurag7 (approved):

PR #73084, opened 2025-11-24
openjdk-8: build directly from OpenJDK git repo

javacruft:
Refactor the OpenJDK 8 package to align build approach to newer OpenJDK package releases.

- Use git-checkout for upstream source
- Trace updates via upstream git repo
- Use distinct image build targets for JDK and JRE packages
- Add ldd-checks for main and -jre packages
- Add -doc package
- Add -dbg package
- Pick patches for linking and glibc compatibility from newer OpenJDK releases

Why do this? The IcedTea distribution we've used to date appears to have some performance penalties and we don't have full control over the flags passed to the compiler - specifically some Hotspot/JIT related operations are ~33% slower.

A few points to note:

 - we're not tracking specific build tags - just the ga tag - so we can't actually determine the build number to pass to the configure step - the resulting build has the right internal update version but not build number (always b00).
 - The JRE and JDK packages are built from specific images produced by the build process, rather than the JRE just being a subset of the files from the JDK (and the main JDK package depending on the JRE package); this is aligned to newer OpenJDK package versions but not the original IcedTea based openjdk-8.


EyeCantCU (approved):
Performance significantly improves, test coverage improves, nothing we build actually leverages IcedTea, and we're packaging the JRE using the JRE image now... What's not to like?

PR #73178, opened 2025-11-25
mysql-9.5, percona-server-9.1: Add tmpfiles.d snippet

cjwatson:
This ensures that the appropriate subdirectory of `/run` exists when running in a VM, where `/run` is a tmpfs.

Related: https://docs.google.com/document/d/1pydotE6GRdgWP9xmO9YFFX6_f7PdMa7wMWRQSzH8bFw

### DIFF
--- a/mysql-9.5.yaml
+++ b/mysql-9.5.yaml
@@ -94,11 +94,6 @@ pipeline:
         runs: |
           mv "${{targets.destdir}}"/usr/usr/lib/* "${{targets.destdir}}"/usr/lib/
           rmdir -p --ignore-fail-on-non-empty "${{targets.destdir}}"/usr/usr/lib
-      # We don't currently use the systemd services, but we configure with
-      # systemd support in order to install a suitable tmpfiles.d snippet.
-      - name: "Remove systemd services"
-        runs: |
-          rm -r "${{targets.destdir}}"/usr/lib/systemd/
       - name: "Remove extras"
         runs: |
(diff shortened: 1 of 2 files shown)


mysql-9.5.yaml
```diff
@@ -76,15 +78,27 @@ pipeline:
...
       - uses: cmake/build
       - uses: cmake/install
+      # CMAKE_INSTALL_PREFIX=/usr confuses this.
+      - name: "Fix installation path of systemd components"
+        runs: |
+          mv "${{targets.destdir}}"/usr/usr/lib/* "${{targets.destdir}}"/usr/lib/
+          rmdir -p --ignore-fail-on-non-empty "${{targets.destdir}}"/usr/usr/lib
+      # We don't currently use the systemd services, but we configure with
+      # systemd support in order to install a suitable tmpfiles.d snippet.
+      - name: "Remove systemd services"
```
a-crate: Systemd units are useful for us in VMs. It would be nice to split them into a `${{package.name}}-systemd-units` subpackage (which is the pattern we use for nginx, and other unit files) as long as we're now building them.

I'm fine with this as-is as well though.

a-crate (dismissed):

mysql-9.5.yaml
cjwatson: OK, done (at least in theory - I've only looked over the results by eye, not actually tried building a VM where I could test them).  Please re-review and see if this fits the right pattern.

a-crate (approved):

PR #73196, opened 2025-11-25
chore: apply patch to pipenv for FIPS compatibility

Aditevil:
pipenv fails in FIPS mode with `_hashlib.UnsupportedDigestmodError` when generating resolution cache keys. Python 3.12's
[gh-127301](https://github.com/wolfi-dev/os/blob/main/python-3.12/gh-127301.patch) patch ensures that when _hashlib/OpenSSL is available and OpenSSL is in FIPS mode, builtin hash implementations require `usedforsecurity=False` for MD5/SHA1 usage.


Aditevil:
https://github.com/wolfi-dev/advisories/blob/main/py3-pipenv.advisories.yaml

Taffer (approved):
LGTM

PR #73285, opened 2025-11-26
restore a few openssl build-dependencies

xnox:
To rebuild a historical build of openssl, restore a few toolchain
packages.


sil2100 (approved):
+1, we can always withdraw them again once we're done.

PR #73306, opened 2025-11-27
Fix: Handle nginx-stable mail module size corruption (#73275)

BHAVISHYA2005:
Fix: Handle nginx-stable mail module size corruption (#73275)

Fixes: #73275

Related: N/A

## Description

The `nginx-stable-mod-mail` subpackage was generating a corrupted size field in the APKINDEX, 
reporting an impossible size of `18446744073709551615` (UINT64_MAX / 2^64-1) instead of the 
actual package size.

### Root Cause
The mail module subpackage extraction was failing silently:
- Mail is configured as `--with-mail=dynamic` (should generate a .so file)
- The pipeline tries to move `ngx_mail_module.so` but it doesn't exist in the expected location
- The empty/malformed subpackage causes a size calculation overflow in melange
- Result: Size field gets set to UINT64_MAX

### Solution
Added specific error handling for the mail module case in the subpackage pipeline:
- Check if `ngx_mail_module.so` exists before attempting to move it
- Log a warning if the file is not found

### Verification
The issue was reproduced with:
```bash
curl -s https://packages.wolfi.dev/os/x86_64/APKINDEX.tar.gz | tar -xzf - APKINDEX
grep -B5 "^S:18446744073709551615" APKINDEX

### DIFF
--- a/nginx-stable.yaml
+++ b/nginx-stable.yaml
@@ -2,7 +2,7 @@ package:
   name: nginx-stable
   # STABLE VERSIONS MUST USE EVEN MINOR VERSIONS (e.g., 1.26.x, 1.28.x)
   version: "1.28.0"
-  epoch: 10
+  epoch: 11
   description: HTTP and reverse proxy server (stable version)
   copyright:
     - license: BSD-2-Clause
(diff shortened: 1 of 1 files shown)


cmwilson21:
Thanks for the contribution!

PR #73308, opened 2025-11-27
chore: move crossplane to version stream

kranurag7:
crossplane is version streamed and this commit migrates crossplane to
the same


crossplane-2.1.yaml
```diff
@@ -1,15 +1,14 @@
 package:
-  name: crossplane
+  name: crossplane-2.1
   version: "2.1.1"
-  epoch: 1 # GHSA-j5w8-q4qc-rx2x
+  epoch: 2 # GHSA-j5w8-q4qc-rx2x
```
developer-guy, line 4: should we reset the epoch here since this is a new package or is it necessary to bump the epoch to kick off the build?

developer-guy (changes requested):
we should add `tag-filter-prefix: 2.1` to updates section.

crossplane-2.1.yaml
kranurag7: it's a new package so ideally we should reset to 0 but we are also providing the crossplane which was the previous package name hence epoch bump.

developer-guy (dismissed):
lgtm

crossplane-2.1.yaml
```diff
@@ -1,15 +1,11 @@
 package:
-  name: crossplane
+  name: crossplane-2.1
```
AnmolVirdi, line 2: Since this will be version streamed, should we need to add `dependencies.provides` here?

kranurag7: oversight, updated via https://github.com/wolfi-dev/os/pull/73308/commits/53cf1af0f0092fbc39152791e79a0203d7ae474c

crossplane-2.1.yaml
```diff
@@ -44,7 +43,10 @@ pipeline:
   - uses: strip
```
AnmolVirdi:
```suggestion
```

AnmolVirdi (commented):
Minor nit, the rest lgtm.

rahulduvedi (approved):

PR #73340, opened 2025-11-27
nftables: Add JSON output support 

max-allan-cgr:
Now that we are shipping iptables-wrappers in images, they are likely to use nftables, and then will need the json output support. This option, described in the source code's INSTALL, adds support for JSON.

### DIFF
--- a/nftables.yaml
+++ b/nftables.yaml
@@ -41,6 +41,7 @@ pipeline:
   - uses: autoconf/configure
     with:
       opts: |
+        --with-json \
         --sbindir=/usr/bin
 
   - uses: autoconf/make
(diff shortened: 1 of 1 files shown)


kranurag7 (dismissed):

nftables.yaml
```diff
@@ -41,6 +41,7 @@ pipeline:
   - uses: autoconf/configure
     with:
       opts: |
+        --with-json \
```
maxgio92, line 48: I know we don't have much tests, but can we test that it's compiled in?

Something like
```
nft -j list ruleset | jq >/dev/null
```

without the json support the output would otherwise be

```shell
# nft -j list ruleset
Error: JSON support not compiled-in
```

maxgio92 (approved):

PR #73574, opened 2025-12-02
fix(chrony): set makestep correctly

89luca89:
Fixes VMS-228

### DIFF
--- a/chrony.yaml
+++ b/chrony.yaml
@@ -1,7 +1,7 @@
 package:
   name: chrony
   version: "4.8"
-  epoch: 4
+  epoch: 5
   description: NTP client and server programs
   copyright:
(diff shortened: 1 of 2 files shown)


xnox (approved):

markusboehme:
I know this PR is merged already, but neither the commit nor the issue has any information on why this setting is correct and the previous one is not, and what governs correct/incorrect. It would be very helpful to have that kind of information logged in either (or both) places in case the configuration needs to be modified in the future.

89luca89:
For reference:

The change will Step the clock if offset > 1 second unlimited (-1) number of times, instead of only the first 3 times.
This ensures we catch up time quickly even for systems were we get sleep/hibernation or fast resumes

PR #73596, opened 2025-12-02
fix: add tmpfiles for faillock dirs

89luca89:
Ref VMS-231

### DIFF
--- a/linux-pam/pam-faillock.conf
+++ b/linux-pam/pam-faillock.conf
@@ -1,0 +1,3 @@
+ # Create faillock directory with correct SELinux context                        
+ d /var/log/faillock 0700 root root - -                                          
+ Z /var/log/faillock - - - - -     
--- a/linux-pam.yaml
+++ b/linux-pam.yaml
@@ -52,8 +52,8 @@ pipeline:
       done
 
       # make "unix_chkpwd" shadow group and enable g+s
-      chgrp shadow ${{targets.destdir}}/bin/unix_chkpwd \
-        && chmod g+s ${{targets.destdir}}/bin/unix_chkpwd
+      chgrp shadow ${{targets.destdir}}/usr/bin/unix_chkpwd \
+        && chmod g+s ${{targets.destdir}}/usr/bin/unix_chkpwd
 
       # Don't ship /var/run
       rm -r ${{targets.destdir}}/var/run
--- a/linux-pam.yaml
+++ b/linux-pam.yaml
@@ -1,7 +1,7 @@
 package:
   name: linux-pam
   version: "1.7.1"
-  epoch: 2
+  epoch: 3
   description: Linux PAM (Pluggable Authentication Modules for Linux)
   copyright:
     - license: BSD-3-Clause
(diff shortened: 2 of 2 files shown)


linux-pam/pam-faillock.conf
```diff
@@ -0,0 +1,3 @@
+ # Create faillock directory with correct SELinux context                        
+ d /var/log/faillock 0700 root root - -                                          
+ Z /var/log/faillock - - - - -
```
xnox, line 3: is it the right place? i somehow was under impression that they are files and in /run/ these days. Or like it is configurable and they should be there.

89luca89: It's right there, the /run/faillock one is correctly generated with the right context already

egibs:
The `ci-diff-report` changes are expected:
```diff
usr/bin/unix_chkpwd
+ NOTE: permissions changed for usr/bin/unix_chkpwd: 755 -> 2755
+ WARNING: usr/bin/unix_chkpwd now has special permissions: setgid
- -rwxr-xr-x 91792 2025-10-16 12:19:44 unix_chkpwd
-   APK-TOOLS.checksum.SHA1: "a8b870aa360a77e09f1f922c5e510b4774218d95"
+ grwxr-xr-x 91840 2025-12-02 17:21:44 unix_chkpwd
+   APK-TOOLS.checksum.SHA1: "3441e675a0f35770434a2eda1437a3c8c50bac3a"
```
due to:
```sh
chgrp shadow ${{targets.destdir}}/usr/bin/unix_chkpwd \
  && chmod g+s ${{targets.destdir}}/usr/bin/unix_chkpwd
```

89luca89:
> The `ci-diff-report` changes are expected:
> 
> ```diff
> usr/bin/unix_chkpwd
> + NOTE: permissions changed for usr/bin/unix_chkpwd: 755 -> 2755
> + WARNING: usr/bin/unix_chkpwd now has special permissions: setgid
> - -rwxr-xr-x 91792 2025-10-16 12:19:44 unix_chkpwd
> -   APK-TOOLS.checksum.SHA1: "a8b870aa360a77e09f1f922c5e510b4774218d95"
> + grwxr-xr-x 91840 2025-12-02 17:21:44 unix_chkpwd
> +   APK-TOOLS.checksum.SHA1: "3441e675a0f35770434a2eda1437a3c8c50bac3a"
> ```
> 
> due to:
> 
> ```shell
> chgrp shadow ${{targets.destdir}}/usr/bin/unix_chkpwd \
>   && chmod g+s ${{targets.destdir}}/usr/bin/unix_chkpwd
> ```

Yea I think before the chgrp whas silently failing I guess

a-crate:
Out of curiosity, why is this a tmpfiles.d snippet instead of just `mkdir -p ${{targets.destdir}}/var/log/faillock` in the package build script?

sergiodj:
> Out of curiosity, why is this a tmpfiles.d snippet instead of just `mkdir -p ${{targets.destdir}}/var/log/faillock` in the package build script?

I believe it's because of the SELinux context thing, although it should be possible to use `mkdir -pZ ...`.  I'll wait for a reply before approving this PR.

89luca89:
> mkdir -pZ

That won't work, we need it to be created at runtime to get the right context, else it is created with the var_log one:

With mkdir:

```
 # ls -ladZ /var/log/faillock/
drwx------. 2 root root unconfined_u:object_r:var_log_t:s0 18 Dec  3 07:38 /var/log/faillock/
```

With tmpfile:

```
 # ls -ladZ /var/log/faillock/
drwx------. 2 root root system_u:object_r:faillog_t:s0 18 Dec  3 07:42 /var/log/faillock/
```

sergiodj (approved):

PR #73641, opened 2025-12-02
Switch to iptables-wrappers

Taffer:
Everything that can be using `iptables-wrappers` should be using `iptables-wrappers`. These all built/tested locally ("works on my machine"), so :crossed_fingers: ...

### DIFF
--- a/flannel.yaml
+++ b/flannel.yaml
@@ -1,7 +1,7 @@
 package:
   name: flannel
   version: "0.27.4"
-  epoch: 4 # GHSA-j5w8-q4qc-rx2x
+  epoch: 5
   description: flannel is a network fabric for containers, designed for Kubernetes
   copyright:
     - license: Apache-2.0
@@ -9,9 +9,8 @@ package:
     runtime:
       - ca-certificates
       - coreutils
-      - ip6tables
       - iproute2
-      - iptables
+      - iptables-wrappers
       - nftables
       - strongswan
       - wireguard-tools
@@ -36,8 +35,7 @@ pipeline:
       deps: |-
         golang.org/x/crypto@v0.45.0
 
-  - if: ${{build.arch}} == 'aarch64'
-    uses: patch
+  - uses: patch
     with:
       patches: disableBrNetfilterCheck.patch
 
@@ -74,6 +72,7 @@ test:
         - etcd
         - jq
         - iproute2
+        - iptables-wrappers
   pipeline:
     - name: "Check flanneld version"
       runs: |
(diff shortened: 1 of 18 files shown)


flannel.yaml
```diff
@@ -122,13 +121,8 @@ test:
...
         # Run flanneld in background
-        if [ "${{build.arch}}" = "aarch64" ]; then
-          flanneld --etcd-endpoints=$ETCD_ENDPOINTS --iface=$IFACE --iptables-forward-rules=false --disable-br-netfilter-check > /tmp/flannel.log 2>&1 &
-          FLANNEL_PID=$!
-        else
-          flanneld --etcd-endpoints=$ETCD_ENDPOINTS --iface=$IFACE --iptables-forward-rules=false > /tmp/flannel.log 2>&1 &
-          FLANNEL_PID=$!
-        fi
+        flanneld --etcd-endpoints=$ETCD_ENDPOINTS --iface=$IFACE --iptables-forward-rules=false --disable-br-netfilter-check > /tmp/flannel.log 2>&1 &
+        FLANNEL_PID=$!
```
sergiodj, line 125: I don't understand this change.  Is it related to `iptables-wrappers`?

sergiodj (dismissed):
Thanks @Taffer.  Small question about a change.

flannel.yaml
Taffer: The `disableBrNetfilterCheck.patch` adds the `--disable-br-netfilter-check` option :thinking:, which is for

> address compatibility issues in containerized environments, particularly for CI/CD pipelines test using Docker runners on aarch64 architecture

This "compatibility issue" seems to have also migrated to x86-64, at least on my system (Linux 6.17.9-arch1-1). I don't *think* it's related to `iptables-wrappers`, but I did run into it while testing the change. I'm also assuming it's not important, since it's just disabled for ARM.

linkerd2-proxy-init.yaml
```diff
@@ -1,14 +1,15 @@
...
   description: "Init container that sets up the iptables rules to forward traffic into the Linkerd2 sidecar proxy"
   copyright:
     - license: Apache-2.0
   dependencies:
     runtime:
-      - ip6tables
-      - iptables-xtables-privileged
+      # Is something needed to handle the iptables-xtables-privileged dep
+      # removed in the switch to iptables-wrappers?
+      - iptables-wrappers
```
EyeCantCU: This image as ran as non-root so the user won't have permission to manage network interfaces or create sockets using xtables

The solution is probably to add a replaces directive to `iptables-xtables-privileged` that's basically:

```yaml
dependencies:
  replaces:
    - iptables
 ```

Allowing iptables-xtables-privileged to replace the contents of iptables. Then we'd include both iptables-wrappers and iptables-xtables-privileged at runtime

File: linkerd2.yaml [hunk 3]
```diff
@@ -218,7 +218,7 @@ subpackages:
         - grep
         - iproute2
         - iptables
-        - iptables-xtables-privileged
+        - iptables-xtables-privileged # WHAT TO DO about iptables-wrappers change?
```
EyeCantCU: See above, that will do the trick here too

EyeCantCU (changes requested):

linkerd2-proxy-init.yaml
Taffer: `iptables-xtables-privileged` just has the `xtables-*-multi` bins inside, so it doesn't conflict with `iptables-wrappers` (which depends on `xtables`); not sure if we need to change the `iptables-xtables-priviledged` subpackage in `iptables.yaml` to use `replaces: xtables` or not… either it already works, or it's not being used anywhere.

linkerd2.yaml
Taffer: We're able to remove `iptables` and `iptables-wrappers` when `iptables-xtables-privileged` is required; it provides all of the correct things.

Taffer:
`rancher-2.12` needs https://github.com/wolfi-dev/os/pull/73714 to land before it'll succeed, I was tricked locally by an `apk` feature into thinking it could go in the batch already.

EyeCantCU (approved):
Great work on this! Documenting for anyone passing by - linkerd2 will need an iptables-wrappers-xtables-privileged packaged, or alternatively, just an xtables-privileged package will need to be created that only includes xtables and provides xtables (so that it can be used with iptables-wrappers). Removing iptables at runtime for now is the right call

PR #73692, opened 2025-12-03
remove test.key from tornado's test

developer-guy:
this creates noice in vuln scanners so we have to remove the test.key

### DIFF
--- a/airflow-3.yaml
+++ b/airflow-3.yaml
@@ -1,7 +1,7 @@
 package:
   name: airflow-3
   version: "3.1.3"
-  epoch: 0 # GHSA-r397-ff8c-wv2g
+  epoch: 1 # GHSA-r397-ff8c-wv2g
   description: Platform to programmatically author, schedule, and monitor workflows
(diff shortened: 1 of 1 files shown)


kranurag7 (approved):

Dentrax (approved):

kranurag7:
attached cves don't have any fixes.

kranurag7:
verified [key is not present](https://apk.chaindag.dev/size/apk.cgr.dev/wolfi-presubmit/794796e1ea24d1cdd283475eea95bb33986d45a3/aarch64/airflow-3-3.1.3-r1.apk@etag:30376636633237303337636536373038613164633434333866626330396432352d3830), looks good to land

mamccorm:
Please don't remove these files until we're sure they aren't on the upstream image by default and serve a purpose (demos / installations). If they exist upstream then we should not be deviating unless for good reason. I've dropped an internal comment but we might loop back and revert this depending on that discussion!

PR #73714, opened 2025-12-03
Be more forceful about replacing iptables/ip6tables

Taffer:
Hoping this is the right change to resolve an issue I'm having with `iptables-wrappers` and `linkerd2`; posting to get a pre-submit URL for testing.

### DIFF
--- a/iptables.yaml
+++ b/iptables.yaml
@@ -200,6 +200,8 @@ subpackages:
       runtime:
         - merged-sbin
         - wolfi-baselayout
+      replaces:
+        - xtables
 
   # iptables-apply is a bash script, and therefore depends on bash
   # only a subset of iptables users use this script, so we keep
--- a/iptables.yaml
+++ b/iptables.yaml
@@ -1,7 +1,7 @@
 package:
   name: iptables
   version: 1.8.11
-  epoch: 28
+  epoch: 29
   description: Linux kernel firewall, NAT and packet mangling tools
   copyright:
     - license: GPL-2.0-or-later
--- a/iptables-wrappers.yaml
+++ b/iptables-wrappers.yaml
@@ -1,7 +1,7 @@
 package:
   name: iptables-wrappers
   version: 2
-  epoch: 3 # CVE-2025-61725
+  epoch: 4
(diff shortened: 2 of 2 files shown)


iptables.yaml
```diff
@@ -200,6 +200,8 @@ subpackages:
       runtime:
         - merged-sbin
         - wolfi-baselayout
+      replaces:
+        - xtables
```
EyeCantCU, line 204: Sadly, the problem with this package is it isn't just `xtables`, it also includes everything in `iptables` (see L183). So if it replaces xtables, that is unfortunately not sufficient because `iptables-wrappers` can only replace `iptables` and `ip6tables`

I believe this is the only package like this... We only set capabilities for xtables. We could make this package just `xtables-privileged`, provide `xtables`, and set `provider-priority: 1` for `xtables` and `provider-priority: 1` for `xtables-privileged`. Then include `iptables-wrappers` and `xtables-privileged` at runtime

EyeCantCU (changes requested):

iptables.yaml
EyeCantCU: Actually, I'd imagine the reason this would have worked is because APK allows packages from the same origin to replace each other

Taffer: I was under the impression that `xtables` is the right thing, which is what `iptables-wrappers` will end up calling in the end. The `iptables-xtables-priviledged` creates symlinks for *all the things*, but they're going to the `xtables-*-multi` binary.

On the other hand, `xtables` only includes `xtables-*-multi` and none of the other things. So maybe we do need a pared down one for `xtables-priviledged`? I have no idea what I'm doing. ;-)

Taffer: Regardless of what we end up doing, `rancher-2.12` needs this change to work correctly; why don't we merge this and I can carry on with the `iptables-wrappers` work and we'll make a task to clean this up some more later?

EyeCantCU: Sounds like a plan

EyeCantCU (approved):

Taffer:
New task: https://github.com/chainguard-dev/internal-dev/issues/24623

PR #74007, opened 2025-12-04
go-1.25: epoch bump to build with go-1.25.5

dnegreira:
Build with go 1.25.5 to remediate CVE-2025-61727 and CVE-2025-61729

Signed-off-by: David Negreira <david.negreira@chainguard.dev>

### DIFF
--- a/go-1.25.yaml
+++ b/go-1.25.yaml
@@ -1,7 +1,7 @@
 package:
   name: go-1.25
   version: "1.25.5"
-  epoch: 0
+  epoch: 1
   description: "the Go programming language"
   copyright:
     - license: BSD-3-Clause


xnox (changes requested):
i would be surpised if this worked.

in theory, the go toolchain does 3 stage bootstrap the in the end, the code used to build the final toolchain is itself.

Thus even with a rebuild I expect all binaries in this package to be built by 1.25.5 go already.

Or is there a leak in the bootstrap process, with us shipping interim go binaries which are built with the bootstrap compiler?

aborrero (approved):

Ankush-Pathak (approved):

dnegreira:
closing since automation has kicked in

PR #74484, opened 2025-12-04
CODEOWNERS: Use ruby-*.yaml instead of specifying a version

sergiodj:
The ruby module packages we have are named `rubyM.N-*.yaml`, which means that we can actually use globbing to match all Ruby versions on CODEOWNERS.


sil2100 (approved):
+1 on this, considering that ruby-packages start with rubyX.Y not ruby-X.Y, so we should be good.

PR #73391, opened 2025-11-28
openjdk-{17,21}: fix use of java.util.random in jre images.

javacruft:
Add `jdk.random` to the list of boot modules which means it will be included in the JRE image.

The fix in  OpenJDK 23 that resolves this issue moves the classes into `java.base` however that's a breaking change so taking this approach should resolve the issue without breaking any users who are directly using `jdk.random` with either jlink or in their own Java code.

PR #74710, opened 2025-12-05
openjdk-24+: add jdk.crypto.ec by default to jre

xnox:
Add empty jdk.crypto.ec module to JRE jimage by default, for compat
with 3rd party modules that need it.

22+ JREs from various vendors also ship empty jdk.crypto.ec by
default.

This is similar to previous fix for a different jre module:
- https://github.com/wolfi-dev/os/pull/73391

### DIFF
--- a/openjdk-24.yaml
+++ b/openjdk-24.yaml
@@ -1,7 +1,7 @@
 package:
   name: openjdk-24
   version: "24.0.2"
-  epoch: 8
+  epoch: 9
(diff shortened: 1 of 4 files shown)


sergiodj (approved):
Thanks for tackling this; I was starting to work on this yesterday but got sidetracked with personal stuff.

I'm assuming we'll have to replicate this fix to our other JDK variants

PR #74714, opened 2025-12-05
ca-certificates: Move compatibility links into separate package

markusthoemmes:
In order to be able to reuse the compatibility provided by the linked locations in other bundles, this moves their creation into a separate package.

Out of an abundance of caution, this keeps ca-certificates-bundle as-is for now. Eventually, we can make it depend on the new package.

### DIFF
--- a/ca-certificates.yaml
+++ b/ca-certificates.yaml
@@ -1,7 +1,7 @@
 package:
   name: ca-certificates
   version: "20251003"
-  epoch: 0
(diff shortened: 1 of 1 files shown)


xnox (changes requested):
Are you sure about this?

Note if one needs to override cert bundle, it is best to declare replaces and then provide the cert bundle in all the multiple locations it needs to be in.

PR #74719, opened 2025-12-05
pipelines: go: add assert-no-crypto check

markusboehme:
Move the no-cryptography check from `gops.yaml` into the new `go/assert-no-crypto` pipeline. Then use it to ensure the build of `iptables-wrappers` can be used in an environment requiring FIPS compliance.

### DIFF
--- a/gops.yaml
+++ b/gops.yaml
@@ -1,7 +1,7 @@
 package:
   name: gops
   version: 0.3.28
-  epoch: 22 # CVE-2025-61729
+  epoch: 23 # CVE-2025-61729
   description: gops is a command to list and diagnose Go processes currently running on your system.
   copyright:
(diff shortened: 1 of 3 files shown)


markusboehme:
epoch-bot is failing, but I'm intentionally not bumping the epoch since this is a build-time check that is not affecting the output.

xnox:
> epoch-bot is failing, but I'm intentionally not bumping the epoch since this is a build-time check that is not affecting the output.

please bump the epoch - the rebuild is cheap; and we want to see the CI run to see that the build passes with this change.

xnox (changes requested):
please rebuild both packages.

xnox (approved):

EyeCantCU (approved):

PR #74814, opened 2025-12-05
openssf-compiler-options: additional include path for libraries

sergiusens:


dannf:
+1. I suggested this, so I'd like to get someone else from OS to review it. Another option would be for libraries to write to `/home/build/.melange.spec` - but that's kinda weird for us, because our buildenv doesn't have a `/home/build`. I also like that this leaves a reminder in `openssf-compiler-options` that `melange` isn't the only user of this, in case we re-architect things later.

xnox (approved):

PR #74830, opened 2025-12-05
fix(tailscale): Withdraw 1.92.0

catmsred:
tailscale tags more versions that it releases. A 1.92.0 version of
tailscale was released by Chainguard when only 1.90.8 was available
as an upstream release. This change withdraws the 1.92.0 tailscale
package.

Patch crypto version to 0.45.0 to remediate GHSA-j5w8-q4qc-rx2x

Relates: https://github.com/chainguard-dev/CVE-Dashboard/issues/47184

### DIFF
--- a/tailscale.yaml
+++ b/tailscale.yaml
@@ -1,7 +1,7 @@
 package:
   name: tailscale
-  version: "1.92.0"
-  epoch: 1 # CVE-2025-61729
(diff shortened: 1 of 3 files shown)


philroche (approved):

philroche:
Moving to draft so we can stage the merging with an image/tag withdrawal

PR #74898, opened 2025-12-06
test jitter fallback

xnox:
- **openssl: correct comments in tests**
  During jitter testing the grep asserts are correct, but comments were
  left unchanged from the earlier getrandom testing. Update the comments
  to match the implementation and the test-case name / intention.
  

- **openssl-dbg: fixup source code snapshot for generated code**
  fixup source code snapshot for all the generated code
  

- **bump epoch**
  

- **Fixup HMAC patch to be upstreamble**
  

- **Perform throw-away canary tests of jitter and non-validated fips**
  This helps to catch issues with future fips modules.
  

- **Update fix-jitter.patch for 3.2+ providers**
  As highlighted by the new throwawy compile test, update
  fix-jitter.patch for 3.2+ providers.

### DIFF
--- a/openssl.yaml
+++ b/openssl.yaml
@@ -1,7 +1,7 @@
 package:
   name: openssl
   version: "3.6.0"
-  epoch: 4
+  epoch: 5
   description: "the OpenSSL cryptography suite"
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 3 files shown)


Taffer (approved):
LGTM

PR #74918, opened 2025-12-07
Makefile: fix SOURCE_DATE_EPOCH

xnox:
The latest git commit is more important, than the last commit that
touched the .yaml, because files outside of individual .yaml can
change, and affect the build via build-dependencies.

Also SOURCE_DATE_EPOCH 0 is not supported as it breaks zip file
format, the minimum value is 1980.

As a side-effect, this fixes building uncommitted packages which use
zip files - anything with python wheels, java jars, cargo crates etc.

This now also matches the implementation in production in elastic
builds too.


smoser:
thank you.

smoser:
https://github.com/wolfi-dev/os/pull/75616 was a simmilar change, @xnox pointed me at this.
I adjusted and put up https://github.com/wolfi-dev/os/pull/75637

smoser:
replaced by https://github.com/wolfi-dev/os/pull/75637

PR #74946, opened 2025-12-07
pulsar: python multiversion + a number of fixes

dannf:
My intent here was just to python multiversion py3-pulsar-client, one of the few remaining items on our old [missing multiversion checklist](https://github.com/chainguard-dev/internal-dev/issues/5334) that linear keeps bugging me about finishing. But I found that the python module currently fails to import, which required fixes to libpulsar. While there, I noticed that we had a bunch of pinned build-deps that I managed to clean out.

- **py3-pulsar-client: cleanup unneeded build-deps on ca-certs/wolfi-base**
- **py3-pulsar-client: Add test/tw/pip-check and missing dep it found**
- **py3-pulsar-client: Use py3-build-base-dev**
- **py3-pulsar-client: Use py/pip-build-install**
- **py3-pulsar-client: Use cmake pipelines**
- **py3-pulsar-client: Unpin gcc**
- **py3-pulsar-client: python multiversion**
- **libpulsar: apply patch to allow us to unpin abseil/protobuf**
- **Revert "libpulsar: require boost-dev<1.87 due to missing io_service alias"**
- **libpulsar: libpulsar-dev should depend on libpulsar**
- **py3-pulsar-client: Add python/import tests**
- **libpulsar: Add regression test for the python library**
- **libpulsar: bump epoch to 1**
- **py3-pulsar-client: bump epoch to 1**

### DIFF
--- a/libpulsar.yaml
+++ b/libpulsar.yaml
@@ -67,5 +67,20 @@ update:
     strip-prefix: v
 
 test:
+  environment:
+    contents:
+      packages:
+        - py3.10-pulsar-client
+        - py3.13-pulsar-client
   pipeline:
     - uses: test/tw/ldd-check
+    # They python library dynamically links against this package,
+    # check that we haven't obviously broken something.
+    - uses: python/import
+      with:
+        python: python3.10
+        import: pulsar
+    - uses: python/import
+      with:
+        python: python3.13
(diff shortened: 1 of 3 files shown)


Taffer (approved):
LGTM, nice cleanups

PR #75175, opened 2025-12-08
cbindgen: switch to git update backend to query tags

OddBloke:
Upstream haven't created a release for the past couple of tagged releases, and Firefox is now failing to build because it needs one of those versions (so they evidently are legit releases!).

### DIFF
--- a/cbindgen.yaml
+++ b/cbindgen.yaml
@@ -49,6 +49,5 @@ test:
 
 update:
   enabled: true
-  github:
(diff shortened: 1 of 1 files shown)


OddBloke:
No epoch bump required for an `update`-only change, and I want automation to take care of the actual package update.

kwmonroe (approved):
lgtm, thx for explaining the version-check failures

PR #75439, opened 2025-12-09
py3-pytz: introduce additional tests

aborrero:


aborrero:
will send another PR later with a more accurate test

PR #75460, opened 2025-12-09
openssl: assert that entropy is provided in the get_user_entropy

xnox:
There is no way to put the provider into an error state from the core
callback. Also without entropy some providers might not correctly
enter into error state. Hence in such scenarios calling assert,
causing OpenSSL to die is the best course of action to ensure the
provider enters error state and does not produce any output.

This callback is used by 3.2+ providers only, and is not used in
non-fips images. The throw away build does excercise that code-path by
executing the full openssl test suite.

### DIFF
--- a/openssl.yaml
+++ b/openssl.yaml
@@ -1,7 +1,7 @@
 package:
   name: openssl
   version: "3.6.0"
-  epoch: 5
+  epoch: 6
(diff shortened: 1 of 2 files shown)


Taffer (approved):
Oof, but yes definitely.

sil2100 (approved):
+1

PR #75465, opened 2025-12-09
py3: introduce additional tests

aborrero:


aborrero:
sending more files on the same PR.

PR #75486, opened 2025-12-09
microvm-init: FEATURE add init.d scripting system

antitree:
This PR adds a change to the microvm-init script we use for building via QEMU (internally and externally). It adds an optional feature to support an "init.d" style folder of additional scripts to run during OS creation. 

This is required so that we do not make a bunch of improvements to our internal Chainguard infrastructure and allow the public to do something similar. 

Changes:
- Add init.d script runner that sources scripts from /mount/opt/melange/init.d/
- Scripts are executed in alphanumeric order before chroot
- Scripts are sourced (not executed) to share environment with parent process
- Failures halt boot due to 'set -e' at script start
- Epoch bumped from 14 to 25

### DIFF
--- a/microvm-init.yaml
+++ b/microvm-init.yaml
@@ -1,7 +1,7 @@
 package:
   name: microvm-init
   version: 0.0.1
-  epoch: 15
+  epoch: 16
   description: Minimal busybox init for microvm workloads
   copyright:
     - license: Apache-2.0
--- a/microvm-init.yaml
+++ b/microvm-init.yaml
@@ -1,7 +1,7 @@
 package:
   name: microvm-init
   version: 0.0.1
-  epoch: 14
+  epoch: 25
   description: Minimal busybox init for microvm workloads
   copyright:
     - license: Apache-2.0
@@ -16,12 +16,10 @@ package:
       - openssh-server
       - util-linux-misc
       - xfsprogs
-
 environment:
   contents:
     packages:
       - busybox
-
 pipeline:
   - runs: |
       mkdir -p ${{targets.destdir}}/
@@ -72,7 +70,6 @@ pipeline:
       Match LocalPort 2223
           ChrootDirectory none
       EOF
-
 update:
   enabled: false
   exclude-reason: |
--- a/microvm-init.yaml
+++ b/microvm-init.yaml
@@ -1,7 +1,7 @@
 package:
   name: microvm-init
   version: 0.0.1
-  epoch: 25
+  epoch: 15
   description: Minimal busybox init for microvm workloads
   copyright:
     - license: Apache-2.0
--- a/microvm-init.yaml
+++ b/microvm-init.yaml
@@ -1,7 +1,7 @@
 package:
   name: microvm-init
   version: 0.0.1
-  epoch: 16
+  epoch: 15
   description: Minimal busybox init for microvm workloads
   copyright:
     - license: Apache-2.0
--- a/microvm-init.yaml
+++ b/microvm-init.yaml
@@ -16,10 +16,12 @@ package:
       - openssh-server
       - util-linux-misc
       - xfsprogs
+
 environment:
   contents:
     packages:
       - busybox
+
 pipeline:
   - runs: |
       mkdir -p ${{targets.destdir}}/
@@ -70,6 +72,7 @@ pipeline:
       Match LocalPort 2223
           ChrootDirectory none
       EOF
+
 update:
   enabled: false
   exclude-reason: |
--- a/microvm-init/init
+++ b/microvm-init/init
@@ -103,6 +103,7 @@ mount --rbind /proc /mount/proc
 mount --rbind /sys /mount/sys
 mount --rbind /tmp /mount/tmp
 mount -t tmpfs -o mode=1777 tmpfs /mount/run
+
 mount -t cgroup2 -o nsdelegate,memory_recursiveprot,nosuid,nodev,noexec cgroup2 /mount/sys/fs/cgroup
 mkdir -p /mount/dev/pts
 mount -t devpts devpts -o noexec,nosuid,newinstance,ptmxmode=0666,mode=0620,gid=tty /mount/dev/pts/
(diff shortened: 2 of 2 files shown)


microvm-init.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: microvm-init
   version: 0.0.1
-  epoch: 14
+  epoch: 25
```
xnox: that's a jump, is this right?

antitree: Yes. Sorry my local iterations. I'll fix

microvm-init/init
```diff
@@ -140,6 +141,29 @@ echo "127.0.0.1 ${HOSTNAME:-"wolfi-vm"}" >> /mount/etc/hosts
...
 
+#########################
+# INIT.D SCRIPT RUNNER  #
+#########################
+
+echo "[INIT] Checking for init.d scripts..." > /dev/console
+
+# Run scripts from /mount/opt/melange/init.d/ in numeric order
+# Scripts are sourced (run in parent process) and must succeed (set -e)
+if [ -d /mount/opt/melange/init.d ]; then
```
89luca89: All occurences of /mount/opt should be /opt

microvm-init/init
```diff
@@ -140,6 +141,29 @@ echo "127.0.0.1 ${HOSTNAME:-"wolfi-vm"}" >> /mount/etc/hosts
...
+	for script in /mount/opt/melange/init.d/*; do
+		if [ -f "$script" ] && [ -r "$script" ]; then
+			echo "[INIT] Sourcing init.d script: $(basename $script)" > /dev/console
+			# Source the script - runs in parent process with full environment
+			# Will halt on error due to 'set -e' at top of this script
+			. "$script"
```
89luca89: This should execute not source, to act as an init.d (maybe usefult in the future for other uses)

89luca89 (changes requested):

microvm-init/init
89luca89: Why do we want to preserve the environment?

microvm-init/init
antitree: Yes. Just so I remember why putting a comment for myself: all of the scripts should assume to execute outside of the chroot so there's no reason to include them inside.

antitree:
## Updated with Additional Features

I've updated this PR to include two additional improvements:

### 1. Fixed init.d Path (Addresses Reviewer Feedback)

**Resolved:** Changed `/mount/opt/melange/init.d/` → `/opt/melange/init.d/`

As @89luca89 correctly pointed out, these scripts run **before** the chroot, so they should be at `/opt/melange/init.d/` not `/mount/opt/melange/init.d/`.

### 2. Added QEMU_ADDITIONAL_PACKAGES Support

**New Feature:** The init script now reads `melange.additional_packages` from the kernel cmdline and installs them before running init.d scripts.

**How it works:**
```bash
# In melange (host side)
QEMU_ADDITIONAL_PACKAGES=hello-wolfi,strace melange build pkg.yaml

# In microvm-init (guest side)
# Reads: melange.additional_packages=hello-wolfi,strace from /proc/cmdline
# Runs: chroot /mount apk add --no-cache hello-wolfi strace
```

**Benefits:**
- 🔧 Install debugging tools (strace, gdb, tcpdump) for troubleshooting
- 📦 Add runtime dependencies not in build environment  
- ⚡ Quick iteration without modifying build configs
- 🔒 Install security monitoring hooks like `microvm-tetragon-hook`

**Execution Order:**
1. Network setup
2. DNS configuration  
3. **→ Install additional packages** (new)
4. Run init.d scripts (which can now use the additional packages)
5. Start SSH and build process

**Related PR:**
- melange: https://github.com/chainguard-dev/melange/pull/2265

Let me know if you'd like any changes to the approach!

microvm-init.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: microvm-init
   version: 0.0.1
-  epoch: 14
+  epoch: 16
```
egibs: You can leave the epoch bump at `15`.

antitree:
## Updated: Simplified to Path Fix Only

I've updated this PR based on discussion about the best approach for additional packages:

### Changes in This PR

**✅ Fixed init.d Path** (Addresses @89luca89's feedback):
- Changed `/mount/opt/melange/init.d/` → `/opt/melange/init.d/`
- Correct because init.d scripts run **before** the chroot

**❌ Removed Runtime Package Installation**:
- The previous commit included code to install packages at runtime via kernel cmdline
- After discussion, we decided the **initramfs approach is better**:
  - Packages are baked into initramfs by melange
  - Faster (no network required)
  - Simpler (no runtime apk operations)
  - More reliable

### How Additional Packages Work Now

**In melange** (host side):
```bash
QEMU_ADDITIONAL_PACKAGES=hello-wolfi,strace melange build pkg.yaml
```

**What happens:**
1. Melange reads `QEMU_ADDITIONAL_PACKAGES` env var
2. Adds packages to initramfs image: `packages: [microvm-init, hello-wolfi, strace]`
3. Packages are baked into initramfs during build
4. Packages are available immediately when VM boots
5. No runtime installation needed ✨

### Summary

This PR now focuses on **just the init.d path fix** which was the original reviewer feedback. The additional packages feature is handled entirely in melange via the initramfs approach.

**Related PR:**
- melange: https://github.com/chainguard-dev/melange/pull/2265 (implements initramfs approach)

antitree:
## Updated: Changed to Execute Instead of Source

I've addressed @89luca89's second piece of feedback about execution method.

### Changes

**✅ Scripts Now Execute in Separate Process:**
- Changed from `. "$script"` (sourcing) → `"$script"` (executing)
- Changed check from `-r` (readable) → `-x` (executable)
- Updated comments and logs to say "Executing" instead of "Sourcing"

**Why This Change:**
- ✅ Better isolation (separate process)
- ✅ Mirrors traditional init.d behavior
- ✅ Allows future flexibility (scripts can't accidentally modify init state)
- ✅ More predictable behavior

### Summary of All Changes in This PR

1. **Fixed init.d path**: `/mount/opt` → `/opt` (scripts run before chroot)
2. **Changed execution method**: Sourcing → Executing (traditional init.d behavior)

Both requested changes from review are now addressed! 🎉

microvm-init/init
antitree: Ok this aspect is fixed. Only /opt used

microvm-init/init
antitree: I ported this to run as a separate process and tested it. Seems to work fine for the one script we have.

egibs (approved):
I tested these changes locally and they install a test script as expected.

LGTM but deferring to @89luca89 for final approval in case any final changes are needed.

89luca89 (approved):
LGTM

PR #75560, opened 2025-12-10
actions-runner-controller: remove sleep binary

aborrero:
Previous to this patch, this package was providing /usr/bin/sleep, but we already
have that via coreutils.

Drop from here to avoid APK file conflicts on tests (and/or images).


dnegreira:
Adding for future reference, the built in sleep was added by upstream for their docker image:
https://github.com/actions/actions-runner-controller/pull/2305

dnegreira (approved):

PR #75705, opened 2025-12-11
harbor-distribution subpackage addition

AnmolVirdi:
- `harbor` distribution subpackage since upstream harbor registry uses `distribution` version `v2.8.3`
- This uses harbor's fork of distribution package, similar to upstream.
- https://github.com/goharbor/harbor/blob/main/Makefile#L119-L121
- Added a `drift` check to fail the build when upstream updates it.


harbor-2.14.yaml
```diff
@@ -231,6 +237,85 @@ subpackages:
...
+          expected-commit: 4772604ae973031ab32dd9805a4bccf61d94909f
+      - runs: |
+          # Adds source modules to $GOPATH
+          mkdir -p "${DISTRIBUTION_DIR}"
+          cp -rf . "${DISTRIBUTION_DIR}"
+      - runs: |
+          # distribution v2.8.3 does not have go.mod file and therefore, it's not supported by Melange's go/build pipelines
+          set -eux
+          # GOPATH-style build
+          # We can't use melange's go/build pipeline because it primarily requires go.mod file which does not exist in source code for Distribution v2.8.3
```
utieyin, line 272: nit: the comment on 263 already explains why we are not using the go build pipeline

AnmolVirdi: Updated, thanks!

harbor-2.14.yaml
```diff
@@ -53,183 +60,83 @@ pipeline:
   - uses: strip
 
 subpackages:
-  - name: ${{package.name}}-jobservice
```
mritunjaysharma394, line 56: Curious, why are we deleting this subpackage?

harbor-2.14.yaml
```diff
@@ -53,183 +60,83 @@ pipeline:
...
+          # https://github.com/goharbor/harbor/blob/main/Makefile#L113
+          grep -qF "REGISTRYVERSION=v${{vars.harbor-distribution-release-branch}}" Makefile
+      - uses: git-checkout
         with:
-          packages: ./jobservice
-          output: harbor-jobservice
-          modroot: ./src
+          repository: https://github.com/goharbor/distribution
+          branch: release/${{vars.harbor-distribution-release-branch}}
+          expected-commit: 0c62ec3ec8df7ea91c819448236a67f0777f929f # needs to be updated with each package release
```
mritunjaysharma394: Let's use a var in the beginning for the commit as well so that it is easy to change while bumping things

harbor-2.14.yaml
AnmolVirdi: Reverted, apologies for the issue.

harbor-2.14.yaml
```diff
@@ -53,183 +60,83 @@ pipeline:
...
+
+          cd "${DISTRIBUTION_DIR}"
+          # Copy vendored dependencies to GOPATH so Go can find them
+          # The vendor directory uses old-style vendoring that's incompatible with newer Go
+          cp -r vendor/* /usr/lib/go/src/
+          rm -rf vendor
+
+          GO111MODULE="off" go build \
+            -tags "include_oss,include_gcs" \
+            -ldflags "-w -X ${DISTRIBUTION_DIR}/version.Version=${{vars.custom-distribution-version}} -X ${DISTRIBUTION_DIR}/version.Revision=$(git rev-parse --short HEAD) -X ${DISTRIBUTION_DIR}/version.Package=${DISTRIBUTION_PKG}" \
```
mritunjaysharma394: We are using the forked version, have we pulled upstream harbor distribution to make sure that we are aliging with ldflags?

harbor-2.14.yaml
```diff
@@ -1,14 +1,19 @@
...
     - license: Apache-2.0
   dependencies:
     provides:
       - harbor=${{package.full-version}}
 
+# To be updated as upstream harbor distribution requirements, when drift check fails for harbor-distribution subpackage.
+vars:
+  custom-distribution-version: 2.8.3
+  harbor-distribution-release-branch: 2.8
+
```
mritunjaysharma394, line 17: Let's add the fork's commit sha for it here for release/2.8 branch

harbor-2.14.yaml
```diff
@@ -231,6 +238,85 @@ subpackages:
...
+  - name: ${{package.name}}-distribution
+    description: harbor registry custom package built using Distribution v2.8.3
+    dependencies:
+      provides:
+        - harbor-distribution=${{package.full-version}}
+    pipeline:
+      - name: "Drift Check"
+        runs: |
+          # https://github.com/goharbor/harbor/blob/main/Makefile#L113
+          grep -qF "REGISTRYVERSION=v${{vars.harbor-distribution-release-branch}}" Makefile
```
mritunjaysharma394: I believe we should use this rather `REGISTRY_SRC_TAG` as the release branch is our source of truth, that version tag can be brittle
https://github.com/goharbor/harbor/blob/fbb45d8c9421dea821855dd3e5f0130d25af6d41/Makefile#L119

harbor-2.14.yaml
AnmolVirdi: Yes, we're aligning with the upstream: 
- Upstream dockerfile uses makefile from `goharbor/distribution`: https://github.com/goharbor/harbor/blob/main/make/photon/registry/Dockerfile.binary
- Makefile ldflags and config: https://github.com/goharbor/distribution/blob/release/2.8/Makefile#L33

harbor-2.14.yaml
AnmolVirdi: Updated, thanks

harbor-2.14.yaml
AnmolVirdi: That would be better, yes. Thanks for pointing this @mritunjaysharma394!
I've updated this.

harbor-2.14.yaml
AnmolVirdi: Updated, thanks!

harbor-2.14.yaml
```diff
@@ -231,6 +239,85 @@ subpackages:
               artifact annotation version v1alpha1 registered
               Registering database: type-PostgreSQL
 
+  # Sub-package to serve harbor-registry/distribution version from goharbor/distribution repo(Harbor's Fork for distribution)
+  # To be updated when the drift check fails.
+  # If the drift check test fails, this means that the upstream images no longer use older Distribution version and have succesfully migrated to using the next release branch.
+  - name: ${{package.name}}-distribution
```
mritunjaysharma394: Just a thought and open suggestion, should we rename this to upstream like harbor-photon-registry? 🤔 

That will save us from confusion with the provides block: 
        - harbor-distribution=

As there the  =${{package.full-version}} is of harbor and not of distribution (and rightly so)

AnmolVirdi: I believe, ideally we should name it `harbor-distribution` or `harbor-registry`_(breaking change)_ since `photon` comes from `photonOS` which might be mis-interpreted.

AnmolVirdi: But you're right. Adding `-photon` makes more sense here.
I'm bundling `entrypoint` files as well so that this passes our 1:1 match with the upstream image.

mritunjaysharma394 (approved):

PR #75745, opened 2025-12-11
chore: remove leading/trailing quotes from error strings

Aditevil:
followup for https://github.com/wolfi-dev/os/pull/69358

### DIFF
--- a/nri-cassandra.yaml
+++ b/nri-cassandra.yaml
@@ -1,7 +1,7 @@
 package:
   name: nri-cassandra
   version: "2.16.0"
-  epoch: 0 # CVE-2025-61729
+  epoch: 2 # CVE-2025-61729
   description: New Relic Infrastructure Cassandra Integration
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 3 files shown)


caddy.yaml
```diff
@@ -147,10 +147,10 @@ test:
         start: caddy run --config modules.Caddyfile
         expected_output: "serving initial configuration"
         error_strings: |
-          "msg":"error"
-          "level":"error"
-          "panic"
-          "fatal"
+          msg:error
```
smoser: this one is actually wierd... I feel like the user was expecting literal `"level":"error"` (like from a json lines output).  And the change here would _miss_ that error.

caddy.yaml
```diff
@@ -112,10 +112,10 @@ test:
         start: caddy run --config static.Caddyfile
         expected_output: "serving initial configuration"
         error_strings: |
-          "msg":"error"
-          "level":"error"
-          "panic"
-          "fatal"
+          msg:error
```
smoser: same as below.

smoser (changes requested):
Some of the changes here will cause us to _miss_ errors. (I think... hard to know for sure without looking further)

The `sed` cleanups that are done in the pipeline (added https://github.com/wolfi-dev/os/pull/69358) only remove the leading and trailing `"`.

smoser (approved):

PR #75750, opened 2025-12-11
Add clang to -dev packages runtime

Taffer:
Adds `clang` as a runtime dependency to the `-dev` subpackages.

Addresses https://github.com/chainguard-dev/customer-issues/issues/2893

### DIFF
--- a/python-3.14.yaml
+++ b/python-3.14.yaml
@@ -1,7 +1,7 @@
 package:
   name: python-3.14
   version: "3.14.2"
-  epoch: 1 # CVE-2025-12084
+  epoch: 2
   description: "the Python programming language"
   copyright:
     - license: PSF-2.0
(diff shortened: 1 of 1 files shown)


sergiodj:
Thanks, Chris.  How would you feel about taking the opportunity to create a `vars.clang-version: 19` and using it everywhere?  It should make it easier to make sure everything clang-related is updated in tandem.

sergiodj (changes requested):
Marking as "request changes" just in case.

Taffer:
CVE-2024-3220 is Low prio, has no fix yet, and has a work-around for affected platforms (Windows): https://github.com/python/cpython/issues/130514

sergiodj (approved):

kranurag7 (approved):

PR #75823, opened 2025-12-12
Remove Old original flux and only use the version-stream flux-2.7

debasishbsws:


Aditevil (dismissed):

kranurag7 (commented):
we need to add provides for the compat as well.

kranurag7 (approved):

PR #76104, opened 2025-12-16
Revert iptables-wrappers changes

Taffer:
Wholesale switching to `iptables-wrappers` has caused other issues, so we're going to back off and figure out a better approach.


Taffer:
potentially jumping the gun

PR #76183, opened 2025-12-17
add: jdk versioning for jenkins-2 pkg

anushkamittal20:
Now that upstream supports jdk 17 and jdk 21, we are trying to add packages for both jenkins and jenkins-docker-agent


mjramirez-dev (approved):
LGTM

PR #76409, opened 2025-12-18
feat(kube-agent-updater): package build

developer-guy:
Fixes:

Related:

#### For new package PRs only

### DIFF
--- a/teleport-18.yaml
+++ b/teleport-18.yaml
@@ -1,7 +1,7 @@
 package:
   name: teleport-18
   version: "18.6.0"
-  epoch: 0 # GHSA-g754-hx8w-x2g6
+  epoch: 1 # GHSA-g754-hx8w-x2g6
   description: The easiest, and most secure way to access and protect all of your infrastructure.
   copyright:
     - license: AGPL-3.0-only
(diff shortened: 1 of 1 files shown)


teleport-18.yaml
```diff
@@ -90,6 +90,66 @@ pipeline:
 
   - uses: strip
 
+subpackages:
+  - name: ${{package.name}}-kube-agent-updater
```
mritunjaysharma394, line 94: I think even this would be verseion streamed right?

teleport-18.yaml
```diff
@@ -90,6 +90,66 @@ pipeline:
...
+    description: Kubernetes agent updater for Teleport
```
mritunjaysharma394, line 95: We should also use provides if this is version streamed too

teleport-18.yaml
developer-guy: so the package name would be like `teleport-18-kube-agent-updater`, is that ok? or should we add the version info at the end of the sub-package? what do you think?

mritunjaysharma394: I would suggest observe what other such veresion streamed packages do in their compat ;)

teleport-18.yaml
```diff
@@ -90,6 +90,71 @@ pipeline:
...
+        - teleport-kube-agent-updater-compat=${{package.full-version}}
+    pipeline:
+      - runs: |
+          mkdir -p "${{targets.contextdir}}/"
+          ln -sf ./usr/bin/teleport-kube-agent-updater "${{targets.contextdir}}/teleport-kube-agent-updater"
+    test:
+      environment:
+        contents:
+          packages:
+            - ${{package.name}}-kube-agent-updater
```
kranurag7: nit: this can be removed given the package is already at runtime.

kranurag7 (commented):
looks good mostly, clean work!! one small inline comment and a rebase to main.

Aditevil (approved):

PR #76460, opened 2025-12-19
erlang-{25,27,28}: switch to git-checkout

cjwatson:
erlang-26 had already been switched a while ago.


Taffer (approved):
LGTM

PR #76480, opened 2025-12-19
gnutar: switch to git-checkout

cjwatson:
For some reason this has the effect of turning on more compiler strictness, some of which fails.  But it's easy to relax that a bit.


Taffer (approved):
LGTM

PR #76500, opened 2025-12-19
DNM: Use iptables-nft instead of iptables-wrappers

Taffer:
DO NOT MERGE - also, remove skip:package-version-check and skip:staging-package-version-check labels!

This is currently just to provide a pre-submit repo for testing the `iptables-nft` package in images.

This switches things to use `iptables-nft` (the current tech kernel firewall). See INC102 for details.


Taffer:
Splitting this due to large.

PR #76578, opened 2025-12-21
7zip: switch to git-checkout

cjwatson:
It's unfortunate that we're using a mangled variant of the version that we then have to unmangle to get back to the git tag, but since 25.01 < 2501 I guess we're stuck with it.


Taffer (approved):
LGTM

PR #76621, opened 2025-12-22
xorg: switch to git-checkout

cjwatson:

### DIFF
--- a/font-xproto.yaml
+++ b/font-xproto.yaml
@@ -1,7 +1,7 @@
 package:
   name: font-xproto
   version: 7.0.31
-  epoch: 4
+  epoch: 5
   description: X.org font xproto
   copyright:
     - license: MIT
@@ -14,12 +14,16 @@ environment:
       - build-base
       - busybox
       - ca-certificates-bundle
+      - pkgconf-dev
+      - util-macros
 
 pipeline:
-  - uses: fetch
+  - uses: git-checkout
     with:
-      expected-sha256: 6d755eaae27b45c5cc75529a12855fed5de5969b367ed05003944cf901ed43c7
-      uri: https://www.x.org/releases/individual/proto/xproto-${{package.version}}.tar.gz
+      # TODO: This repository is deprecated and archived - can we drop this package?
+      repository: https://gitlab.freedesktop.org/xorg/proto/xproto
+      tag: xproto-${{package.version}}
+      expected-commit: f3b7b856777500113cee524dbd295fcbbce11fab
 
   - uses: autoconf/configure
 
--- a/fontsproto.yaml
+++ b/fontsproto.yaml
@@ -1,7 +1,7 @@
 package:
   name: fontsproto
   version: 2.1.3
-  epoch: 2
+  epoch: 3
   description: X.org fontsproto
   copyright:
     - license: MIT-open-group
@@ -14,12 +14,16 @@ environment:
       - build-base
       - busybox
       - ca-certificates-bundle
+      - pkgconf-dev
+      - util-macros
 
 pipeline:
-  - uses: fetch
+  - uses: git-checkout
     with:
-      expected-sha256: 72c44e63044b2b66f6fa112921621ecc20c71193982de4f198d9a29cda385c5e
-      uri: https://www.x.org/releases/individual/proto/fontsproto-${{package.version}}.tar.gz
+      # TODO: This repository is deprecated and archived - can we drop this package?
+      repository: https://gitlab.freedesktop.org/xorg/proto/fontsproto
+      tag: fontsproto-${{package.version}}
+      expected-commit: df8c05f7c0253a36589d96efa52938215eff9d4d
 
   - uses: autoconf/configure
     with:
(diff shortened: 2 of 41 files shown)


font-xproto.yaml
```diff
@@ -14,12 +14,16 @@ environment:
...
+      - pkgconf-dev
+      - util-macros
 
 pipeline:
-  - uses: fetch
+  - uses: git-checkout
     with:
-      expected-sha256: 6d755eaae27b45c5cc75529a12855fed5de5969b367ed05003944cf901ed43c7
-      uri: https://www.x.org/releases/individual/proto/xproto-${{package.version}}.tar.gz
+      # TODO: This repository is deprecated and archived - can we drop this package?
```
Taffer, line 23: Probably; I couldn't find a reference to it anywhere other than `bdftopcf` using `font-xproto-dev`. If `bdftopcf` can build without it, obviously.

fontsproto.yaml
```diff
@@ -14,12 +14,16 @@ environment:
...
+      - pkgconf-dev
+      - util-macros
 
 pipeline:
-  - uses: fetch
+  - uses: git-checkout
     with:
-      expected-sha256: 72c44e63044b2b66f6fa112921621ecc20c71193982de4f198d9a29cda385c5e
-      uri: https://www.x.org/releases/individual/proto/fontsproto-${{package.version}}.tar.gz
+      # TODO: This repository is deprecated and archived - can we drop this package?
```
Taffer, line 23: Probably: same deal as `font-xproto`.

libxkb.yaml
```diff
@@ -1,7 +1,8 @@
+# TODO: Misnamed duplicate of libxkbfile - should be dropped/merged?
```
Taffer, line 1: Definitely. `freerdp-3` and `freerdp` reference `libxkb-dev` though.

`r1` of this package is already in the withdrawn packages list, but I didn't dig up the *why* of that.

libxss.yaml
```diff
@@ -1,7 +1,8 @@
+# TODO: Duplicate of libxscrnsaver - should be dropped/merged?
```
Taffer, line 1: Definitely: `libsdl3` and `x11vnc` reference it; `1.2.4-r2` already in the withdrawn packages.

libxv.yaml
```diff
@@ -13,28 +13,28 @@ environment:
...
-  - uses: fetch
+  - uses: git-checkout
     with:
-      uri: https://www.x.org/releases/individual/lib/libXv-${{package.version}}.tar.xz
-      expected-sha512: 80d7a11e6415fbe0fc50c3c2a1abf8f0f2ec38446c9c8d88ff48875cd94b8949cb1028f2ab37476c4b25cbd7eac34dde9132dd998c4e04ea576b95ae411ba946
+      repository: https://gitlab.freedesktop.org/xorg/lib/libxv
+      tag: libXv-${{package.version}}
+      expected-commit: e1cde54538060c4fd3a3d02e3d2e2b7e5da7bff9
+
+  - uses: autoconf/configure
```
Taffer, line 31: Nice cleanup!

Taffer (changes requested):
Looks like there are a couple of updates required too (probably just rebasing will pick those up).

Dunno if you want to clean up those duplicates now or in a separate PR.

font-xproto.yaml
cjwatson: I updated `bdftopcf` to use `xorgproto` instead.  I'm not sure how withdrawing packages works, though; my thought had been to do that in a follow-up of some kind.

fontsproto.yaml
cjwatson: I updated `bdftopcf` to use `xorgproto` instead.  I'm not sure how withdrawing packages works, though; my thought had been to do that in a follow-up of some kind.

libxkb.yaml
cjwatson: https://github.com/wolfi-dev/os/pull/76640

Should I drop the changes to this package here, or keep them?

libxss.yaml
cjwatson: This one's a judgement call.  The repository name is `libxscrnsaver`, and the `pkg-config` file name is `xscrnsaver.pc`; but on the other hand the shared library is `libXss.so`.  I lean slightly towards keeping `libxscrnsaver` - what do you think?

libxkb.yaml
cjwatson: Looks like the withdrawal was just garbage-collection of a specific old version whose apk we didn't need to keep around any more.  I think that means it would be best to keep these changes for now, and withdraw later.

libxss.yaml
Taffer: What do other distros call it? I'd like 'libXss', but I don't actually know what it's normally called. I checked Fedora, it's got `libXScrnSaver` so `libxscrnsaver` is probably the right one. :+1:

Taffer:
re: package-version-check and staging-package-version-check

And what's with these cloud platforms having low limits? Like "Sorry, we had to truncate this directory to 1,000 files. 3069 entries were omitted from the list." uh, GitHub, did you guys not learn how to stream results? Or maybe just dump them all, we're adults, we can handle our web page taking another 100ms to load. :roll_eyes:

Taffer (approved):

PR #76476, opened 2025-12-19
tileserver-gl: downgrade libvips dependency

aborrero:
We are observing some libvips errors when using tileserver, try with a lower version
to see if it is a regression.

Signed-off-by: Arturo Borrero Gonzalez <arturo.borrero@chainguard.dev>

PR #76630, opened 2025-12-22
fix(tileserver-gl): also pin libvips version in -compat package

bentasker:
The -compat package is consumed by our image, so the dep also requires pinning there.

This is a follow on from https://github.com/wolfi-dev/os/pull/76476

### DIFF
--- a/tileserver-gl.yaml
+++ b/tileserver-gl.yaml
@@ -153,6 +153,8 @@ subpackages:
       provides:
         - ${{package.name}}-entrypoint=${{package.full-version}}
       runtime:
+        # override auto-generated dep so we can force an older versioin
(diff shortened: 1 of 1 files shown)


IdlePhysicist (dismissed):

tileserver-gl.yaml
```diff
@@ -153,6 +153,8 @@ subpackages:
       provides:
         - ${{package.name}}-entrypoint=${{package.full-version}}
       runtime:
+        # override auto-generated dep so we can force an older versioin
```
egibs:
```suggestion
        # override auto-generated dep so we can force an older version
```

IdlePhysicist (approved):

PR #76663, opened 2025-12-22
git: switch to git-checkout

cjwatson:

### DIFF
--- a/git.yaml
+++ b/git.yaml
@@ -1,7 +1,7 @@
 package:
   name: git
   version: "2.52.0"
-  epoch: 0
+  epoch: 1
   description: "distributed version control system"
   copyright:
     - license: GPL-2.0-or-later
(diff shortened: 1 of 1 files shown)


kranurag7 (commented):
@cjwatson are we good with using code mirror; ack that this is particularly fairly upto date. Is there a source tree hosted by kernel itself.

cjwatson:
> @cjwatson are we good with using code mirror; ack that this is particularly fairly upto date. Is there a source tree hosted by kernel itself.

I suppose it would be better to use the true upstream repository, yes.  Done.

Taffer:
I think the `kernel.org` build issue is a bug in `git-checkout` behaviour vs `fetch` (they're using Anubis to block AI scrapers :+1:). Someone with some elastic-build knowledge will have to look into it in the new year.

See <https://github.com/chainguard-dev/internal-dev/issues/24698>

cjwatson:
I switched to kernel.googlesource.com in line with what seemed to be the consensus in that issue.

sergiodj (approved):

PR #76736, opened 2025-12-23
DNM: make Clickhouse go faster on ARM

Taffer:
Clickhouse needs all the CPU it can get, and `-DNO_ARMV81_OR_HIGHER=1` has been throttling it.

Note that we can only do this if, in fact, our ARM target on AWS is Graviton 3.


Taffer:
This is not the right change, and we can't require Graviton 2; I'll try to align our builds better with what up-stream does on ARM.

PR #76817, opened 2025-12-26
Ruby 4.0 + gems

sergiodj:
This is Ruby 4.0 + its associated gems.

Changes:

- Ruby 4.0 incorporated the following gems: `benchmark`, `logger` and
  `reline`.  This means that we don't need to create separate packages
  for them.

- Several of our Ruby module packages had a hardcoded `rubyMM`
  variable set to 3.4.  They now use `var-transforms` to derive the
  Ruby version from the package name.

- ruby4.0-octokit: Cherry-pick fix to bump octokit requirement

- ruby4.0-jruby-openssl: Remove. Upstream said on
  jruby/jruby-openssl#337 that it doesn't make sense to ship this module
  for Ruby as it is a JRuby specific project.  There are no recent
  reverse deps for jruby-openssl either, so let's remove it.

- ruby4.0-rails-8.1: Add `ruby${{vars.rubyMM}}-psych` and `pkgconf` to
  test dependencies.  Test by installing `sidekiq` instead of
  `bcrypt` (the latter fails to build with Ruby 4.0).


sergiodj:
Bundling everything in a single PR doesn't work very well.  Let's try splitting things into separate PRs.

PR #76834, opened 2025-12-26
[shard-3] Ruby 4.0 gems

sergiodj:
Ruby 4.0 gems -- shard 3


justinvreeland (approved):

ruby4.0-quantile.yaml
```diff
@@ -0,0 +1,61 @@
...
+    packages:
+      - build-base
+      - busybox
+      - ca-certificates-bundle
+      - git
+      - ruby-${{vars.rubyMM}}
+      - ruby-${{vars.rubyMM}}-dev
+
+pipeline:
+  - uses: fetch
```
Taffer, line 21: I think we should switch these to `git-checkout`

PR #76894, opened 2025-12-29
chore: migrate kubernetes-latest to 1.35 stream

kranurag7:


debasishbsws (dismissed):

debasishbsws (approved):
LGTM! 👍
This is a great consolidation that simplifies our Kubernetes package maintenance. The approach of moving the `-latest` subpackages into the version-specific configs (like `kubernetes-1.35.yaml`) makes a lot of sense.

this will produce version streams like `kubectl-latest-1.35.0-r1` and `kubectl-latest-1.36.0-r0` when 1.36 releases. the package manager will always resolve to the highest version number, so we will automatically get the true latest version. This is better than the previous approach where we had to manually update a static pointer.

PR #76954, opened 2025-12-30
feat(ruby-3.2-aws-sdk-3): bump to version 1.209.0

bentasker:
This is being done primarily to bring in the change in 1.208.0 which remediates CVE-2025-14762

Have been through the [upstream changelog](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-s3/CHANGELOG.md#12090-2025-12-23) and there are no breaking changes listed between the two versions

### DIFF
--- a/ruby3.2-aws-sdk-core.yaml
+++ b/ruby3.2-aws-sdk-core.yaml
@@ -1,7 +1,7 @@
 #nolint:valid-pipeline-git-checkout-tag
 package:
   name: ruby3.2-aws-sdk-core
-  version: "3.227.0"
+  version: "3.240.0"
   epoch: 0
   description: Provides API clients for AWS. This gem is part of the official AWS SDK for Ruby.
   copyright:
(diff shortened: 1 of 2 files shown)


bentasker:
We need to update sdk core too

> /usr/lib/ruby/3.2.0/rubygems/specification.rb:2303:in `raise_if_conflicts': Unable to activate aws-sdk-s3-1.209.0, because aws-sdk-core-3.227.0 conflicts with aws-sdk-core (~> 3, >= 3.234.0) (Gem::ConflictError)

xnox (dismissed):
The new versions, match the gems/$project/VERSION file

It would be nice to assert that {{package.version}} equals to the VERSION file such that build fails or is successful upon version upgrades.

bentasker:
> It would be nice to assert that {{package.version}} equals to the VERSION file such that build fails or is successful upon version upgrades.

Oh that's a good shout

AmberArcadia (approved):

PR #76976, opened 2025-12-30
fix(crossplane-azure): Refactoring

srishtih:


crossplane-provider-azure.yaml
Aditevil, line 69: just a nit, looks good.
```suggestion
git:
    strip-prefix: v
```

Aditevil (approved):

PR #77021, opened 2025-12-31
Move pg_cron-16 and -18 to Wolfi to match postgresql versions

murraybd:
postgresql 16 and 18 are in Wolfi so these extensions should be too


tulilirockz (changes requested):
Lets wait until the other PR is merged/packages withdrawn, just added this review to soft-block it

tulilirockz (approved):

PR #77150, opened 2026-01-04
Refractor/terragrunt

rahulduvedi:
`terragrunt`: runtime dep either `opentofu` and `terraform` can be added depending upon license constraint 
`busybox`: is added because it is present in the upstream image


anushkamittal20 (approved):
lgtm!
Thanks for the description!

PR #77192, opened 2026-01-05
Use iptables-nft instead of iptables-wrappers (2 of 2)

Taffer:
To clean up the `iptables-wrappers` mess, we should use the `iptables-nft` package instead; new packages should also use `iptables-nft` and we should deprecate `iptables` and/or make it an alias for `iptables-nft`.


EyeCantCU (approved):
Migration looks good, also validated resource allocation for rancher is correct and that is good as well. Great work!

PR #77219, opened 2026-01-05
pgpool2- add iamguarded build package

astrojerms:


pgpool2-4.6.yaml
```diff
@@ -144,6 +144,48 @@ subpackages:
...
+      - runs: |
+          make -C src/parser V=1
+      - uses: autoconf/make
+        with:
+          opts: "clean" # The binary embeds the sysconfdir path at build time into the binary, so we do a clean rebuild.
+      - uses: autoconf/make-install
+      - uses: strip
+      - runs: |
+          mkdir -p "${{targets.contextdir}}/var/run/pgpool"
+          # Move binaries from  /opt/iamguarded/pgpool/bin to /usr/bin then symlinks
```
srishtih: Sorry, I think I'm missing context about why we are moving bins here

astrojerms: I'm just mirroring our typical setup of "binaries in /usr/bin, with symlinks elsewhere pointing to /usr/bin." It isn't required like that, but at the package level this setup means that the binaries are in the path without me setting PATH and I can do ver-check and health-check easily.

srishtih (dismissed):

srishtih (approved):

PR #77230, opened 2026-01-05
iptables-nft for all your iptables needs

Taffer:
Make `iptables-nft` replace `iptables-wrappers` as well. The idea being that this will make it easier to migrate packages to `iptables-nft` without explosions.

Source of *that* theory is `rancher-2.13` failing in https://github.com/wolfi-dev/os/pull/77192 due to *something* down in the dependency tree pulling in `iptables-wrappers` while `iptables-nft` was already there.

### DIFF
--- a/iptables.yaml
+++ b/iptables.yaml
@@ -1,7 +1,7 @@
 package:
   name: iptables
   version: 1.8.11
-  epoch: 30
+  epoch: 31
   description: Linux kernel firewall, NAT and packet mangling tools
   copyright:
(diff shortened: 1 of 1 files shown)


sergiodj (approved):
I ran some tests locally (mainly rebuilding the packages from https://github.com/wolfi-dev/os/pull/77192 and making sure everything tests correctly), and I'm OK with this change.  I haven't been following this iptables battle too closely so feel free to wait for another review before merging :-).

PR #77249, opened 2026-01-06
Log the SHA256 of the Octo STS token

mattmoor:
We do the same thing in our `octo-sts/action` here: https://github.com/octo-sts/action/blob/f603d3be9d8dd9871a265776e625a27b00effe05/index.js#L113-L115

This will let us correlate the hashes of failed builds with exchanges on the server side.


mattmoor:
From previous run with an epoch bump claude chose at random:
image

EyeCantCU (approved):

PR #77322, opened 2026-01-06
cinc-auditor: new package + ruby 3.4 ed25519 and bcrypt_pbkdf gems

stevebeattie:
Add the cinc-auditor package: https://cinc.sh/start/auditor/

This package is an open source distribution of the Chef Software, Inc.  Inspec utility for testing and auditing infrastructure: https://docs.chef.io/inspec/

The cinc-auditor package depends on ~260 gems, many of which are not already packaged; this is why gem install is used to pull gems remotely into a vendored location (`/usr/lib/cinc-auditor/gems`).

Also include two ruby gems, bcrypt_pbkdf and ed25519, that are not included as dependencies in the gemspec files, but end up getting included as part of the cinc build process.

Because of the way the cinc project works, they do not publish a git repo of the upstream inspec repo with their changes applied but instead only distribute release tarballs.

v2:
- add cinc-auditor update info from release-monitoring.org
- fix ruby3.4-bcrypt-pbkdf test script

Ref: https://github.com/chainguard-dev/prodsec/issues/588

#### For new package PRs only

### DIFF
--- a/cinc-auditor.yaml
+++ b/cinc-auditor.yaml
@@ -1,0 +1,113 @@
+# cinc-auditor is a free distribution of the open source Chef Software
+# Inc. inspec tool. See https://cinc.sh/start/auditor/
+package:
+  name: cinc-auditor
+  version: 7.0.95
+  epoch: 0
+  description: cinc-auditor is an OSS tool for applying inspec profiles
+  copyright:
+    - license: Apache-2.0
+      license-path: LICENSE
+    - license: LicenseRef-Third-Party
+      license-path: NOTICE.TXT
+  dependencies:
+    runtime:
+      # additional gem dependencies added in the upstream cinc-auditor
+      # habitat/plan.sh script that can't be listed as part of the Gemfile
+      # for some reason
+      - ruby${{vars.ruby-version}}-bcrypt_pbkdf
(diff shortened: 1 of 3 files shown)


cinc-auditor.yaml
```diff
@@ -0,0 +1,113 @@
...
+      - busybox
+      - gcc
+      - make
+      - ruby${{vars.ruby-version}}-bcrypt_pbkdf
+      - ruby${{vars.ruby-version}}-ed25519
+      - ruby-${{vars.ruby-version}}
+      - ruby-${{vars.ruby-version}}-dev
+
+pipeline:
+  - uses: fetch
```
egibs, line 37: Can we use `git-checkout` with the v7 feature branch? 

https://gitlab.com/cinc-project/distribution/auditor/-/tree/cinc-auditor-7?ref_type=heads (`88f210e4b1e7fa09eeb03b0d7235357ebfcc8899`)

They don't use tags or releases, unfortunately.

stevebeattie: Unfortunately, that repo is the tooling the cinc project uses to modify the upstream repo and build the release tarball. It's *possible* we could work from it to produce the similar artifacts, but not necessarily straightforward.

egibs (approved):

PR #77382, opened 2026-01-07
Migrate git.kernel.org -> kernel.googlesource.com

javacruft:
Update packages that use git.kernel.org to use kernel.googlesource.com; we see sporadic network issues connecting to git.kernel.org which may be related to our usage patterns.


8none1 (approved):
I tested a few URLs at random, and they worked.

PR #77397, opened 2026-01-07
fluent-bit: de-vendor

javacruft:
Unfortunately fluent-bit is vendoring a lot of it's dependencies even when we include the system version.

This adds some missing builddeps as well as telling CMAKE to prefer the system deps.


8none1 (approved):
+1

PR #77618, opened 2026-01-09
dnssec-root: Switch to git-checkout

cjwatson:
This does mean that the package will fail to build when upstream changes (perhaps `git-checkout` needs a mode to check out a specific commit for this sort of case?).  However, we were previously using a commit from 2017-02-05 without a very clear reason, so this might be better than silently falling behind.

### DIFF
--- a/dnssec-root.yaml
+++ b/dnssec-root.yaml
@@ -1,7 +1,7 @@
 package:
   name: dnssec-root
   version: "20190225"
-  epoch: 3
+  epoch: 4
   description: The DNSSEC root key(s)
   copyright:
     - license: CC-PDDC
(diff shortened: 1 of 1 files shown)


dnssec-root.yaml
```diff
@@ -1,7 +1,7 @@
 package:
   name: dnssec-root
   version: "20190225"
```
javacruft, line 3: I think we should refresh the time based version alongside the commit; this appears to be what Alpine does and I guess we're mostly modeled from there.

javacruft (changes requested):
Just one observation re the version field and bumping to todays date.

dnssec-root.yaml
cjwatson: OK, I took this approach in the replacement PR in stereo (https://github.com/chainguard-dev/stereo/pull/12369).

cjwatson:
Closing in favour of https://github.com/chainguard-dev/stereo/pull/12369.

PR #77637, opened 2026-01-09
python-3.14: Move JIT build to subpackage

javacruft:
Make the default package build GCC based (removing the need for clang as
a dependency in the dev packages) and push the JIT enabled build into a
subpackage so that interested users can still try it.

### DIFF
--- a/python-3.14.yaml
+++ b/python-3.14.yaml
@@ -1,7 +1,7 @@
 package:
   name: python-3.14
   version: "3.14.2"
-  epoch: 2
+  epoch: 3
   description: "the Python programming language"
   copyright:
     - license: PSF-2.0
@@ -19,7 +19,7 @@ environment:
       - busybox
       - bzip2-dev
       - ca-certificates-bundle
-      # This is a year old. Hopefully 3.14.x updates will move this to current.
+      # clang, llvm and python-3.13 are required for the JIT subpackage build
       - clang-${{vars.clang-version}}
       - expat-dev
       - gdbm-dev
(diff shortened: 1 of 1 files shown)


python-3.14.yaml
```diff
@@ -232,6 +222,94 @@ subpackages:
...
+          # --enable-optimizations with clang requires PGO support (in particular,
+          # we're missing libclang_rt.profile). This build configuration won't be
+          # optimal and might not produce any performance gains.
+          - runs: |
+              export PATH="/usr/lib/llvm-${{vars.clang-version}}/bin:$PATH"
+          - uses: autoconf/configure
+            with:
+              opts: |
+                CC=clang-${{vars.clang-version}} \
+                --enable-experimental-jit=yes-off \
```
Taffer: Should this just be `yes` (or whatever) to turn it on by default for this package?

Taffer (dismissed):

python-3.14.yaml
```diff
@@ -105,22 +101,16 @@ pipeline:
...
-      tar -cf - --exclude=no-gil . | (cd no-gil; tar -xf -)
-
-  # Experimental JIT support
-  #
-  # --enable-optimizations with clang requires PGO support (in particular,
-  # we're missing libclang_rt.profile). This build configuration won't be
-  # optimal and might not produce any performance gains.
+      mkdir no-gil build-jit
+      tar -cf - --exclude=no-gil --exclude=build-jit . | (cd no-gil; tar -xf -)
+      tar -cf - --exclude=no-gil --exclude=build-jit . | (cd build-jit; tar -xf -)
```
xnox: `tar -xC no-gil`

"-f -" is redundant, as default is to take stdin
"cd foo;" can be replaced with `-C dir` arg (change dir)
this thus removes the subsell need of () as well.

python-3.14.yaml
javacruft: I like that - done

Taffer (approved):
Nice `tar` change, I didn't know about the `C` option!

PR #77792, opened 2026-01-10
daemon-check-output: fix(pdns-5.3): enable webserver for recursor tests

debasishbsws:
The recursor tests were failing because they checked for port 8082 (webserver) but the webserver was not enabled in the default config.

Added PDNS_RECURSOR_API_KEY environment variable which triggers the startup script to enable the webserver on port 8082.


AmberArcadia:
Build failing :(

AmberArcadia (approved):

PR #77797, opened 2026-01-10
Deamon-check-output: fix(seaweedfs): remove /dir/assign check that requires volume servers

debasishbsws:
The /dir/assign endpoint requires volume servers to allocate storage for file uploads, but this test only runs the master server.

Removed the check since /cluster/status and /dir/status are sufficient to verify the master server is working correctly.


AmberArcadia:
@debasishbsws a conflict has arisen

dnegreira (approved):

PR #77835, opened 2026-01-11
feat(crossplane-provider-aws): Builds additional providers

srishtih:


kranurag7 (changes requested):
with additional provides in; we can close this.

PR #77896, opened 2026-01-12
Tweak ARM build to improve v8 compatibility

Taffer:
Due to our default hardening flags, ClickHouse builds *may* be including CPU instructions that aren't strictly compatible with the `NO_ARMV81_OR_HIGHER` build option.


Taffer:
Moved to `stereo`.

PR #77907, opened 2026-01-12
iptables: cherry-pick fix for interface comparisons in rule checks

markusboehme:
When checking for the existence of a rule, iptables-nft accidentally ignores interfaces in the rule and will hence signal the presence of rules that are in fact missing. Cherry-pick the unreleased fix from the upstream project.

#### For PRs that add patches

- [X] Patch source is documented

### DIFF
--- a/iptables.yaml
+++ b/iptables.yaml
@@ -1,7 +1,7 @@
 package:
   name: iptables
   version: 1.8.11
-  epoch: 31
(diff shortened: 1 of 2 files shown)


a-crate (approved):
lgtm, but I'm a lot less familiar with iptables than Chris or Arturo.

PR #77932, opened 2026-01-12
gtk-2/3: fix co-installability of gtk 2 & 3

a-crate:
It's desirable for GTK 2 & 3 to be installable together for packages that depend on both libraries, but they conflict on the presence of the `gtk-update-icon-cache` utility.

Other distros (debian, gentoo) split `gtk-update-icon-cache` into its own package and just ship version 3. This has the side effect of either making gtk-2 depend on gtk-3, or making gtk-2 not depend on `gtk-update-icon-cache` depending on how things are done.

Since our packages previously already included `gtk-update-icon-cache`, I went with the 'make gtk2 depend on gtk3' route. This is a bit odd, but our gtk2 package already depends on gtk4 anyway so not that unusual.
If users desire a gtk3-less gtk-2 install, they can still install `gtk-2.0-update-icon-cache` to make that work.


a-crate:
Alpine ships `gtk-update-icon-cache` as a separate utility as well, they ship version 4 and have everything depend on that. If we want to go that route, I can make a subpackage of gtk4 that provides `cmd:gtk-update-icon-cache` as well. This would make gtk4's update-icon-cache the default, and others installable explicitly by the user as described above.
This works, but is not necessary as gtk renamed the utility to `gtk4-update-icon-cache` in gtk4, so no more conflicts.

sergiodj:
I like the current solution.  If we have problems later we can revisit and go the Alpine route.

sergiodj (approved):

PR #77983, opened 2026-01-13
feat(grafana): add grafana-iamguarded-compat

yashpimple:


kranurag7:
@yashpimple can we also include the same for other version streams as well.

yashpimple:
> @yashpimple can we also include the same for other version streams as well.

@kranurag7 i have add the test to other version also ref: https://github.com/chainguard-dev/enterprise-packages/pull/46768

kranurag7 (approved):
looks good to me.

kranurag7 (changes requested):
let's analyze the runtime deps again.

PR #77990, opened 2026-01-13
create separate package for ceph go deps

utieyin:
Created a standalone package that contains `go` dependencies to satisfy deploying ceph in fips mode
ceph in fips mode requires certain packages to also be in fips mode.
This package just holds those dependencies and is meant to be used together with the main ceph package


toabctl:
What's the advantage of creating a separate new package here?

anushkamittal20 (approved):

PR #78056, opened 2026-01-14
chore: remove argocd-repo-server subpackage

kranurag7:
follow-up after images, this removes argocd-repo-server
subpackage and then moves the runs steps to main package
including the runtime dependencies.

### DIFF
--- a/argo-cd-3.2.yaml
+++ b/argo-cd-3.2.yaml
@@ -13,10 +13,10 @@ package:
       - gnupg
       - gpg
       - gpg-agent
-      - tzdata
       - helm~3
       - kustomize
       - openssh
+      - tzdata
     provides:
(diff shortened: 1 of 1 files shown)


argo-cd-3.2.yaml
```diff
@@ -1,11 +1,22 @@
...
+      - git-lfs
+      - gnupg
+      - gpg
+      - gpg-agent
+      - helm~3
+      - kustomize
+      - openssh
+      - tzdata
     provides:
       - argo-cd=${{package.full-version}}
```
xnox, line 21: Possibly want to provide repo-server here to keep image builds working.

As otherwise images will not build with latest until we drop repo-server references there too.

xnox (approved):

PR #78221, opened 2026-01-15
Qackage improve/amazon k8s cni.yaml

rawlingsj:
## Summary

Add aws-k8s-agent testing, CNI stdin input parsing tests, and memove manual version/help checks because these binaries do not print help and version info.

Made 1 improvement(s) to test configuration.

---
🤖 Generated with qackage-improve CLI

> **Note:** If you need to make manual changes to this PR, apply the `skip:qackage-improve` label so the reconciler won't overwrite them.

### DIFF
--- a/amazon-k8s-cni.yaml
+++ b/amazon-k8s-cni.yaml
@@ -1,7 +1,7 @@
 package:
   name: amazon-k8s-cni
   version: "1.21.1"
-  epoch: 2
+  epoch: 3
   description: Networking plugin repository for pod networking in Kubernetes using Elastic Network Interfaces on AWS
   copyright:
     - license: Apache-2.0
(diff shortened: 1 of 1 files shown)


amazon-k8s-cni.yaml
```diff
@@ -127,21 +127,119 @@ test:
...
+
+        # Create a mock aws-cni source binary for testing
+        mkdir -p /tmp/cni-source
+        cp /usr/bin/aws-cni /tmp/cni-source/
+
+        # Change to source directory and test the copy mechanism
+        cd /tmp/cni-source
+
+        # aws-vpc-cni requires specific environment - test it can at least start
+        if output=$(HOST_CNI_BIN_PATH=/tmp/host-cni-bin aws-vpc-cni 2>&1 || true); then
```
aborrero: I think because we are using `|| true` here, and there is not a single error condition, this test can hardly report a failure (beyond failing the mkdir / cp / cd operations).

I would suggest we introduce a more explicit error condition here. What do you think?

amazon-k8s-cni.yaml
```diff
@@ -127,21 +127,119 @@ test:
...
+        # Test CNI ADD command parsing (will fail without real environment but should parse input)
+        export CNI_COMMAND=ADD
+        output=$(cat /tmp/cni-config.json | aws-cni 2>&1 || true)
+        echo "CNI ADD output:"
+        echo "$output"
+
+        # Plugin should attempt to process the request even if it fails due to missing resources
+        if echo "$output" | grep -iE 'error|failed|missing|cannot|unable'; then
+          echo "Plugin correctly attempts to process ADD command"
+        fi
```
aborrero, line 218: same here. What about having a more explicit error condition?

aborrero (changes requested):

amazon-k8s-cni.yaml
rawlingsj: Hmm yeah - that's odd actually as there's already an explicit skill rule for that. Lemme dig into it - thanks!

kwmonroe: i think it's ok.  in this case, we know the bin will fail and therefore need the `|| true`.  we're just making sure it fails like we expect when we grep the output below.

rawlingsj: ok thanks - I've improved this and the agent skills so hopefully future PRs will always do this.

aborrero:
LGTM.

aborrero (approved):

aborrero:
LGTM.

aborrero (approved):

PR #78229, opened 2026-01-15
ruby-4.0: Fix tag-filter-prefix

sergiodj:
Upstream uses `v4.0...` instead of `v4_0_...`.


AbduAzim424 (approved):
LGTMT

sergiodj:
No epoch bump since this just changes the tag filter prefix.

PR #78242, opened 2026-01-15
feat(datadog-agent): add msodbcsql18 to datadog-agent

developer-guy:
Adds Microsoft ODBC Driver 18 (msodbcsql18) required for SQL Server integration. The official Datadog Agent uses invoke omnibus.build which automatically includes msodbcsql18 via Ruby software definitions. 

Our Wolfi build uses invoke agent.build (Go binaries only) and cannot invoke omnibus without adding heavy Ruby dependencies. We replicate the upstream [msodbcsql18.rb](https://github.com/DataDog/datadog-agent/blob/main/omnibus/config/software/msodbcsql18.rb) logic directly in the melange pipeline. Note: When upgrading datadog-agent, check the upstream msodbcsql18.rb for version/SHA256 changes.

```
tar tvf  packages/aarch64/datadog-agent-7.74-7.74.0-r0.apk |  grep -i msodbc
drwxr-xr-x  0 root   root         0 Jan 15 19:53 opt/datadog-agent/embedded/msodbcsql
drwxr-xr-x  0 root   root         0 Jan 15 19:53 opt/datadog-agent/embedded/msodbcsql/etc
-rwxr-xr-x  0 root   root       149 Jan 15 19:53 opt/datadog-agent/embedded/msodbcsql/etc/odbcinst.ini
drwxr-xr-x  0 root   root         0 Jan 15 19:53 opt/datadog-agent/embedded/msodbcsql/include
-rw-r--r--  0 root   root     44040 Jan 15 19:53 opt/datadog-agent/embedded/msodbcsql/include/msodbcsql.h
drwxr-xr-x  0 root   root         0 Jan 15 19:53 opt/datadog-agent/embedded/msodbcsql/lib
lrwxrwxrwx  0 root   root         0 Jan 15 19:53 opt/datadog-agent/embedded/msodbcsql/lib/libmsodbcsql-18.3.so -> /opt/datadog-agent/embedded/msodbcsql/lib64/libmsodbcsql-18.3.so.3.1
drwxr-xr-x  0 root   root         0 Jan 15 19:53 opt/datadog-agent/embedded/msodbcsql/lib64
-rwxr-xr-x  0 root   root   2048280 Jan 15 19:53 opt/datadog-agent/embedded/msodbcsql/lib64/libmsodbcsql-18.3.so.3.1
drwxr-xr-x  0 root   root         0 Jan 15 19:53 opt/datadog-agent/embedded/msodbcsql/share
drwxr-xr-x  0 root   root         0 Jan 15 19:53 opt/datadog-agent/embedded/msodbcsql/share/doc
drwxr-xr-x  0 root   root         0 Jan 15 19:53 opt/datadog-agent/embedded/msodbcsql/share/doc/msodbcsql18
-rw-r--r--  0 root   root     12135 Jan 15 19:53 opt/datadog-agent/embedded/msodbcsql/share/doc/msodbcsql18/LICENSE.txt
-rw-r--r--  0 root   root     17740 Jan 15 19:53 opt/datadog-agent/embedded/msodbcsql/share/doc/msodbcsql18/RELEASE_NOTES
drwxr-xr-x  0 root   root         0 Jan 15 19:53 opt/datadog-agent/embedded/msodbcsql/share/resources
drwxr-xr-x  0 root   root         0 Jan 15 19:53 opt/datadog-agent/embedded/msodbcsql/share/resources/en_US
-rw-r--r--  0 root   root    200704 Jan 15 19:53 opt/datadog-agent/embedded/msodbcsql/share/resources/en_US/msodbcsqlr18.rll
```

### Verification against official Datadog Agent Docker image

Confirmed that the APK contains all msodbc files present in the official Docker image:

```
$ docker container run --rm --platform linux/arm64 --entrypoint="bash" datadog/agent:latest -c "find / 2>/dev/null | grep -i msodbc"
/opt/datadog-agent/embedded/msodbcsql
/opt/datadog-agent/embedded/msodbcsql/include
/opt/datadog-agent/embedded/msodbcsql/include/msodbcsql.h
/opt/datadog-agent/embedded/msodbcsql/lib
/opt/datadog-agent/embedded/msodbcsql/lib/libmsodbcsql-18.3.so
/opt/datadog-agent/embedded/msodbcsql/lib64
/opt/datadog-agent/embedded/msodbcsql/lib64/libmsodbcsql-18.3.so.3.1
/opt/datadog-agent/embedded/msodbcsql/etc
/opt/datadog-agent/embedded/msodbcsql/etc/odbcinst.ini
/opt/datadog-agent/embedded/msodbcsql/share
/opt/datadog-agent/embedded/msodbcsql/share/resources
/opt/datadog-agent/embedded/msodbcsql/share/resources/en_US
/opt/datadog-agent/embedded/msodbcsql/share/resources/en_US/msodbcsqlr18.rll
```

All files match. The APK additionally includes documentation files (LICENSE.txt, RELEASE_NOTES) which are not present in the official image.

### DIFF
--- a/datadog-agent-7.72.yaml
+++ b/datadog-agent-7.72.yaml
@@ -3,7 +3,7 @@ package:
   # This package has two git checkouts. For each new release, the commit SHA for
   # DataDog/integrations-core must also be updated.
   version: "7.72.4"
-  epoch: 5
+  epoch: 6
   description: "Collect events and metrics from your hosts that send data to Datadog."
   copyright:
     - license: Apache-2.0
@@ -18,9 +18,11 @@ package:
       - blkid
       - findutils
       - grep
+      - krb5 # Required by msodbcsql18 for Kerberos authentication
       - libpcap
       - libseccomp
       - shadow
+      - unixodbc # ODBC driver manager required by msodbcsql18
 
 capabilities:
   add:
@@ -29,6 +31,10 @@ capabilities:
 vars:
   python-version: "3.12"
   destd: /opt/datadog-agent
+  # Microsoft ODBC Driver 18 for SQL Server - required for SQL Server integration
+  # When upgrading datadog-agent, check the version used in the omnibus build:
+  # https://github.com/DataDog/datadog-agent/blob/main/omnibus/config/software/msodbcsql18.rb
+  msodbcsql18-version: "18.3.3.1-1"
 
 var-transforms:
   - from: ${{package.version}}
@@ -46,6 +52,16 @@ var-transforms:
     match: ^(\d+\.\d+)\.\d+$
     replace: "$1"
     to: major-minor-version
+  # Transform msodbcsql18-version (e.g., "18.3.3.1-1") to library suffix (e.g., "18.3.so.3.1")
+  - from: ${{vars.msodbcsql18-version}}
+    match: ^(\d+\.\d+)\.(\d+\.\d+)-\d+$
+    replace: "$1.so.$2"
+    to: msodbcsql18-libsuffix
+  # Extract major.minor version (e.g., "18.3.3.1-1" -> "18.3") for symlink naming
+  - from: ${{vars.msodbcsql18-version}}
+    match: ^(\d+\.\d+)\.\d+\.\d+-\d+$
+    replace: "$1"
+    to: msodbcsql18-major-minor
 
 environment:
   contents:
@@ -63,6 +79,7 @@ environment:
       # - clang-16-dev
       - cmake
       - coreutils
+      - curl
       - dpkg
       - elfutils-dev
       - findutils
@@ -82,10 +99,12 @@ environment:
       - linux-headers
       - ninja
       - openssf-compiler-options
+      - patchelf # Required to patch RPATH of msodbcsql18 library
       - procps-dev
       - py${{vars.python-version}}-build-base-dev
       - py${{vars.python-version}}-semver
       - systemd-dev
+      - unixodbc-dev # Required to build with ODBC support for SQL Server integration
       - util-linux-misc # unshare
       - wget # Required for downloading clang-12 and kernel headers from debian
   environment:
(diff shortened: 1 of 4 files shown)


max-allan-cgr:
Wouldn't it be easier to use our existing msodbc image and pull the package apart/move files around?

PS don't forget the odbc.ini as well as odbcinst.ini.

developer-guy:
We don’t have version 18.3.x version stream  of it, and if I’m not mistaken it isn’t up to date. It also does something similar—extracting the .deb package and moving files around—so I thought it would be beneficial to follow the upstream implementation in this file:
https://github.com/DataDog/datadog-agent/blob/main/omnibus/config/software/msodbcsql18.rb

developer-guy:
The same CVEs are showing up here as well (#77972). I’m not sure how the previous PR was merged despite the failing ci-cve-scan checks, but since these are the same CVEs reported earlier, I think it should be acceptable to merge this PR too.

developer-guy:
I also analyzed the file sizes of the newly added files:

```
tar tzvf packages/aarch64/datadog-agent-7.74-7.74.1-r2.apk 2>/dev/null \
  | grep -i msodbc \
  | grep -v '^d' \
  | grep -v '^l' \
  | awk '{sum += $5} END {printf "%d bytes\n", sum}'
```

The total size of the msodbc files is 2,323,048 bytes (~2.22 MB), which isn’t a significant increase.

datadog-agent-7.72.yaml
```diff
@@ -341,6 +360,48 @@ pipeline:
...
+
+      # Install the driver to the embedded directory
+      cp -r /tmp/msodbcsql18/opt/microsoft/msodbcsql18/* ${{targets.contextdir}}${{vars.destd}}/embedded/msodbcsql/
+
+      # Create symlink for library compatibility
+      ln -sf ${{vars.destd}}/embedded/msodbcsql/lib64/libmsodbcsql-${{vars.msodbcsql18-libsuffix}} ${{targets.contextdir}}${{vars.destd}}/embedded/msodbcsql/lib/libmsodbcsql-${{vars.msodbcsql18-major-minor}}.so
+
+      # Install license and documentation
+      mkdir -p ${{targets.contextdir}}${{vars.destd}}/embedded/msodbcsql/share
+      cp -r /tmp/msodbcsql18/usr/share/* ${{targets.contextdir}}${{vars.destd}}/embedded/msodbcsql/share/ || true
```
mritunjaysharma394: Why the || true block here?

datadog-agent-7.72.yaml
```diff
@@ -341,6 +360,48 @@ pipeline:
       # Install trace-loader binary
       install -Dm755 bin/trace-loader/trace-loader ${{targets.contextdir}}${{vars.destd}}/embedded/bin/trace-loader
 
+  # Install Microsoft ODBC Driver 18 for SQL Server integration
```
mritunjaysharma394, line 363: Is it possible to add a test to prevent future CRIs?

datadog-agent-7.72.yaml
```diff
@@ -341,6 +360,48 @@ pipeline:
...
+      extract: false
+
+  - name: Install msodbcsql18 ODBC driver
+    runs: |
+      mkdir -p /tmp/msodbcsql18
+      mkdir -p ${{targets.contextdir}}${{vars.destd}}/embedded/msodbcsql/lib
+
+      # Extract the deb package
+      if [ "${{build.arch}}" = "aarch64" ]; then
+        dpkg-deb -R msodbcsql18_${{vars.msodbcsql18-version}}_arm64.deb /tmp/msodbcsql18
```
mritunjaysharma394, line 387: Are we sure to untar debian package? Is it not possible to build a similar?

toabctl (approved):
LGTM but I wonder if the library should actually be a separate package independent from datadog-agent. and then datadog-agent would runtime depend on that and would symlink to the required directories. but approving anyway.

datadog-agent-7.72.yaml
toabctl: I don't think the source code is available for the ODBC Driver for SQL Server

developer-guy: exaclty, this is why we put that into extra.

datadog-agent-7.72.yaml
developer-guy: claude does that, let me quickly fix it.

datadog-agent-7.72.yaml
developer-guy: I'm not sure what we can do about it since it was a missing driver but I'm open to ideas.

toabctl (approved):

anushkamittal20 (approved):

mritunjaysharma394 (approved):

PR #78284, opened 2026-01-16
cinc-auditor: add train-k8s-container-mitre plugin and kubectl dep

stevebeattie:
Add the train-k8s-container-mitre gem as part of the package because it provides a functioning mechanism for querying containers in k8s environments. Unfortunately, due to the inconsistency of the gem name with the plugin name (the mitre version is a functional fork of the unpublished-as-a-gem and unmaintained inspec/train-k8s-container repo), this does not show up when one does:

```
cinc-auditor plugin list
```

The train-k8s-container plugin needs the kubectl utility available in the runtime environment to work, so include it as a runtime dependency.

Add an additional check to ensure the k8s-container plugin is actually available to cinc-auditor without actually having a k8s environment set up.

(The included-by-default gem plugin train-kubernetes does not seem to be functional as shipped:
https://github.com/inspec/train-kubernetes/issues/24 which matches the behavior I experienced in testing)

Ref: https://github.com/chainguard-dev/prodsec/issues/588

### DIFF
--- a/cinc-auditor.yaml
+++ b/cinc-auditor.yaml
@@ -4,7 +4,7 @@
 package:
   name: cinc-auditor
   version: "7.0.95"
-  epoch: 2
+  epoch: 3
   description: cinc-auditor is an OSS tool for applying inspec profiles
   copyright:
     - license: Apache-2.0
@@ -15,6 +15,8 @@ package:
     runtime:
       # needed for the shell wrapper script in /usr/bin
       - busybox
+      # both kubernetes transport (train-) plugins need kubectl installed
+      - kubectl
       # additional gem dependencies added in the upstream cinc-auditor
       # habitat/plan.sh script that can't be listed as part of the Gemfile
       # for some reason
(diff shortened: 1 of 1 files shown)


stevebeattie:
Rebased to main and also fix the commit signature.

antitree (approved):
I'll leave my protest that kubectl seems like an unnecessary dependency since you should be able to query the API directly but I don't think that's a hill worth dying on.  Built, works for me.

stevebeattie:
> I'll leave my protest that kubectl seems like an unnecessary dependency since you should be able to query the API directly but I don't think that's a hill worth dying on. Built, works for me.

Oh, I agree completely, and if this was my software, that's what I'd be working towards. But alas, from the README for the train-k8s-container project:

> Below are the two mandatory pre-requisites for this plugin to work.
>
> - train-k8s-container creates connection to k8s containers using the [kubeconfig](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/)
> - [kubectl](https://kubernetes.io/docs/tasks/tools/#kubectl) Both the kubeconfig and kubectl has to be present on the host machine from where this plugin is executed
> By default, it looks for the kubeconfig file in ~/.kube/config. It can be overridden by the ENV["KUBECONFIG"].

(from https://github.com/inspec/train-k8s-container/tree/82bb3cbbeb388c6f5dd46c3ea7e31555807f2739?tab=readme-ov-file#configuration )

**Updated**: bah, sorry that was the inspec project's unpublished gem repo; but MITRE's repo says similar things:

> Prerequisites
> - kubectl installed and in PATH
> - kubeconfig configured with cluster access (default: ~/.kube/config)
> - RBAC permissions to exec into target pods

https://github.com/mitre/train-k8s-container/?tab=readme-ov-file#prerequisites

PR #78301, opened 2026-01-16
apicurio-registry: add compat package

toabctl:
Fixes:

Related:


Dentrax (dismissed):
Thanks for the work, confirmed by comparing against to: `crane config apicurio/apicurio-registry`, and you also added a test for the entrypoint; LGTM!

Dentrax (approved):

PR #78349, opened 2026-01-16
Create BPF variants for the remaining LLVM packages

sergiodj:
Ref.: https://github.com/wolfi-dev/os/pull/78055

### DIFF
--- a/llvm-15.yaml
+++ b/llvm-15.yaml
@@ -1,7 +1,7 @@
 package:
   name: llvm-15
   version: 15.0.7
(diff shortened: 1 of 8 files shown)


sergiodj:
The missing epoch bump on `llvm-19.yaml` is due to the fact that I'm only fixing a typo in the file, so there's no need to do a full rebuild.

EyeCantCU (dismissed):

EyeCantCU (approved):

PR #78570, opened 2026-01-21
py3-l*: switch to git-checkout

cjwatson:
The `scanelf` hack doesn't seem to be needed any more.

I also noticed that Python 3.13 support could be enabled in `py3-llhttp` now, so I did that.


cjwatson:
Closing in favour of https://github.com/chainguard-dev/stereo/pull/12377.

PR #78578, opened 2026-01-23
libsemanage: disable expand-check for semanage commands

markusboehme:
Disable expand-check in semanage.conf for all semanage commands to improve compatibility with the SELinux policy of Fedora and its derivatives. This change aligns with semanage.conf on these distributions.


xnox (approved):

xnox:
yolo

xnox:
wrong target repository.

PR #78639, opened 2026-02-19
zookeeper: remove config in root `/` directory

mbrancato:
Fixes:
A bug where newer zk packages from wolfi have a `/conf` folder which should be in `/usr/share/java/zookeeper/conf` as older packages were. Creating new directories in root is not a standard practice.

Related:

commit for blame:
https://github.com/wolfi-dev/os/commit/b6871dc15d3d12cf1dc0bc4a7e4e3982e53b68ff

#### For new package PRs only

#### For new version streams

#### For security-related PRs

```
<!--ci-cve-scan:fail-any-->
```

#### For version bump PRs

#### For PRs that add patches

### DIFF
--- a/zookeeper-3.9.yaml
+++ b/zookeeper-3.9.yaml
@@ -61,7 +61,7 @@ pipeline:
       # Install Zookeeper
       mv apache-zookeeper-${{package.version}}-bin/bin ${{targets.destdir}}/usr/share/java/zookeeper
       mv apache-zookeeper-${{package.version}}-bin/lib ${{targets.destdir}}/usr/share/java/zookeeper
-      mv apache-zookeeper-${{package.version}}-bin/conf ${{targets.destdir}}/
+      mv apache-zookeeper-${{package.version}}-bin/conf ${{targets.destdir}}/usr/share/java/zookeeper
 
 subpackages:
   - name: ${{package.name}}-compat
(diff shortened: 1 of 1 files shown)


AnmolVirdi:
Hi @mbrancato, thank you for taking the time to raise this PR.
Just want to clarify that the change in question is intentional fix rather than a bug. We aligned the packaging with how upstream distributes and bundles `zookeeper`, where the `conf` directory resides at the root (`/`). For reference, the upstream image layout looks like this:
```
#/apache-zookeeper-3.8.6-bin# ls
bin  conf  docs  lib  LICENSE.txt  NOTICE.txt  README.md  README_packaging.md
#/apache-zookeeper-3.8.6-bin# ls conf/
#/apache-zookeeper-3.8.6-bin# ls ../conf/
configuration.xsl  logback.xml  zoo_sample.cfg
```

Following the upstream structure makes it easier for users to treat our `zookeeper` package as a drop-in replacement, which was the main goal behind this adjustment.

That said, I’m happy to discuss further if you think there are edge cases we should consider. Thanks again for the thoughtful contribution.

mbrancato:
@AnmolVirdi just do that in your apko config using the paths directive to symlink from `/conf` to the installed location for zookeeper. The package today still looks nothing like the "official" container which has pretty much all the folders in root. But still, that is a container issue (apko) not a package issue. Only the `/conf` folder was moved.

AnmolVirdi:
@mbrancato couple of pointers:
1. We do have package configuration that matches the "official" upstream. The `zookeeper-compat` sub-package in the same manifest takes care of it.
2. Is there any specific use-case which was blocked or negatively impacted by this change? If there is, I’d really appreciate understanding it better so we can address it appropriately. Else, we should keep it at par with upstream.

PR #78659, opened 2026-03-30
perl-yaml-syck: fix git tag prefix in checkout

antitree:
## Summary

- Fix `git-checkout` tag template: `v${{package.version}}` → `${{package.version}}`
- Remove `strip-prefix: v` and `tag-filter-prefix: v` from update config
- Bump epoch to 3

The upstream `toddr/YAML-Syck` repo uses tags without a `v` prefix (e.g. `1.36`, not `v1.36`). The `git-checkout` pipeline was failing with `fatal: Remote branch v1.36 not found in upstream origin` because of the incorrect prefix.

This was causing `perl-yaml-syck` builds to fail in melange CI (chainguard-dev/melange#2394) and likely in other pipelines too.

## Test plan

- [ ] Verify `git ls-remote --tags https://github.com/toddr/YAML-Syck.git` shows `1.36` (no `v` prefix)
- [ ] Verify package builds successfully with the corrected tag

### DIFF
--- a/perl-yaml-syck.yaml
+++ b/perl-yaml-syck.yaml
@@ -1,7 +1,7 @@
 package:
   name: perl-yaml-syck
   version: "1.36"
-  epoch: 2 # go/wolfi-rsc/perl-yaml-syck
+  epoch: 3 # go/wolfi-rsc/perl-yaml-syck
   description: Fast, lightweight YAML loader and dumper
   copyright:
     - license: MIT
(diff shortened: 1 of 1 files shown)


egibs (approved):

stevebeattie:
I think this should go through stereo and the sync process there.

antitree:
Wrong repo — this fix should go to stereo instead.

PR #78672, opened 2026-04-14
Remove issue template and PR details for adding package

mosabua:
As it stands now we do not accept new package requests. We are working on a clear policy to determine what packages we accept. Once we have that we plan to add the policy details and a corresponding template back to the project.


xnox:
Will chat in DMs

mosabua:
Thanks

PR #78697, opened 2026-05-07
py3-googleapis-common-protos: relax protobuf<6.0.0.dev0 upper bound in METADATA

cyphermox:
## Summary

- Wolfi ships `protobuf>=7.0` but `googleapis-common-protos 1.66.0` pins `protobuf<6.0.0.dev0` in its wheel METADATA, causing `pip check` to fail with `protobuf 7.x` installed
- Patch the `dist-info/METADATA` after install to remove `,<6.0.0.dev0` from the `Requires-Dist: protobuf` line, following the same pattern used by `mxnet.yaml` in stereo

## Test plan

- [ ] `make package/py3-googleapis-common-protos` succeeds
- [ ] `make test/py3-googleapis-common-protos` passes
- [ ] `pip check` no longer reports a protobuf conflict for this package

### DIFF
--- a/py3-googleapis-common-protos.yaml
+++ b/py3-googleapis-common-protos.yaml
@@ -1,7 +1,7 @@
 package:
   name: py3-googleapis-common-protos
   version: 1.66.0
-  epoch: 3 # go/wolfi-rsc/py3-googleapis-common-protos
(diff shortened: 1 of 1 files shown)


cyphermox:
Opened against wrong repo — fix belongs in chainguard-dev/stereo os/ instead.

PR #78702, opened 2026-05-12
php-8.5: enable and package opcache extension

jamesabell:
Adds OPcache support to the php-8.5 package, mirroring what `php-8.4.yaml` already does.

### Why

PHP 8.5 ships OPcache as a bundled Zend extension, but it has to be explicitly enabled at compile time via `--enable-opcache`. The current `php-8.5.yaml` has no opcache flag in its configure section, so `opcache.so` never gets built and the `php-8.5-opcache` subpackage doesn't exist. Downstream images can't run PHP 8.5 with opcache.

`php-8.4.yaml` enables it via `--enable-opcache=shared` plus an `opcache: Opcache` entry in the `range: extensions` data block. This PR ports both lines to 8.5.

### Changes

- `php-8.5.yaml`: add `--enable-opcache=shared` to the configure flags
- `php-8.5.yaml`: add `opcache: Opcache` to the `data.extensions.items` block (the existing `range: extensions` subpackage loop will then build `php-8.5-opcache` and `php-8.5-opcache-config` automatically)

### Verification

After this PR builds, the following should succeed:

```bash
apk add php-8.5-opcache
php -m | grep -i opcache    # should output: Zend OPcache (or similar)
```

### Pre-review Checklist

- [x] Mirrors the existing pattern in `php-8.4.yaml` exactly

### DIFF
--- a/php-8.5.yaml
+++ b/php-8.5.yaml
@@ -1,7 +1,7 @@
 package:
   name: php-8.5
   version: "8.5.6"
-  epoch: 0
+  epoch: 1
   description: "the PHP programming language"
   copyright:
     - license: PHP-3.01
@@ -136,6 +136,7 @@ pipeline:
         --with-gmp=shared  \
         --enable-intl=shared  \
         --with-ldap=shared  \
+        --enable-opcache=shared \
         --enable-soap=shared  \
         --with-xsl=shared \
         --with-zlib \
@@ -199,6 +200,7 @@ data:
       gettext: GetText
       intl: Internationalization
       mbstring: Multibyte String Functions
+      opcache: Opcache
       pcntl: pcntl
       pdo: PHP Data Objects
       phar: PHP Archive


jamesabell:
Context for reviewers: opcache was explicitly removed from `php-8.5.yaml` in commit 89a31210bc with the rationale:

> The main change here is that upstream made opcache built-in, which means that we no longer have to build it as a shared library.

I believe that premise is incorrect. PHP 8.5 didn't change opcache's compile-time requirement `--enable-opcache` (or `--enable-opcache=shared`) is still required for `opcache.so` to be built at all. Without the flag, opcache isn't compiled.

What did change in 8.5 is that `php.ini-production` ships with `opcache.enable=1` by default (i.e., enabled at config level rather than requiring explicit activation). That's a config default change, not a build-time change. It may have been the source of the confusion.

Current 8.5 package:

```bash
$ apk add php-8.5-opcache
  php-8.5-opcache (no such package):
    required by: world[php-8.5-opcache]

$ docker run --rm cgr.dev/chainguard/php:latest-8.5 php -m | grep -i opcache
# (no output — opcache.so is not in the main package either)
````

This PR ports the 8.4 pattern back, which gives us `opcache.so` as a shared extension and a `php-8.5-opcache` subpackage. Unless I am missing something about the intention with the removal.

jamesabell:
Closing — I was wrong. The official PHP 8.5 UPGRADING notes explicitly state:

> The --enable-opcache/--disable-opcache configure flags have been removed, and the build does not produce opcache.so or php_opcache.dll objects anymore.
>
> The Opcache extension is now always built into the PHP binary and is always loaded.

So this package definition is already correct as-is — opcache is statically compiled into the main \`php-8.5\` binary and is always loaded at runtime. There's no shared \`opcache.so\` in 8.5, so no \`php-8.5-opcache\` subpackage should exist.

The original issue I hit was that my downstream Dockerfile was trying to \`apk add php-8.5-opcache\`, which is no longer the right approach for 8.5. The fix is on my side (drop that line and rely on the bundled opcache).

Apologies for the noise, and thanks to the maintainer for catching this in the original commit message of 89a31210bc.

PR #78703, opened 2026-05-14
langfuse-3: package upgrade - fix build failure

### LINKED ISSUE #78689: [Wolfi Package Update]: langfuse-3
### Package name

langfuse-3

### Current version in Wolfi

3.164.0

### Requested version

3.172.0

### Upstream project URL

https://github.com/langfuse/langfuse

### Problem

New versions aren't being published. On a separate note Langfuse hosts their images with Node 24 while the Wolfi package uses Node 20, maybe worth correcting.

### Steps to reproduce

_No response_

### Root cause (if known)

_No response_

### Proposed solution

_No response_

### Testing performed

_No response_

### Acceptance criteria

- [x] The requested version is the **latest stable upstream release** (no pre-releases or RCs)
- [x] The upstream project uses an **OSI-approved license**
- [x] The change aligns with Wolfi’s packaging and security model
- [x] The package can be reasonably maintained over time
- [x] There are no known unresolved security or supply-chain concerns

ashishch432:
- Bump langfuse to 3.174.1, reset epoch to 0
- Upgrade Node.js 20 → 24 (build env, runtime, worker), this is the actual version used in upstream
- Replace fragile minimumReleaseAge.patch with guarded sed - warns if upstream changes the value, no longer requires a separate patch file. This was causing failures.
- Consolidate all pnpm overrides into a single pnpm pkg set call - the worker subpackage's duplicate overrides were a no-op (no pnpm install re-run) and some set older versions than the main build
- Remove dead/incorrect overrides from original and added patches as needed (eg. oci-* update for uuid latest version to be pickedup)
- Checked local melange build passes (except the few uses: checks which are not available in this repo)

Fixes: #78689

### Pre-review Checklist

```
<!--ci-cve-scan:fail-any-->
```

#### For version bump PRs


ashishch432:
Change applied via https://github.com/wolfi-dev/os/commit/c59f799203ec0a940db38bed167c3953753033d4

amouat:
@ashishch432 many thanks for bringing this to our attention and taking the time to create a PR. All should be fixed now.

PR #78708, opened 2026-05-22
librdkafka: Enable SASL support

alexnovak:
Modifies librdkafka to compile with sasl support. 

Looking at the source for the [cmakelists](https://github.com/confluentinc/librdkafka/blob/v2.14.1/CMakeLists.txt#L123-L162), some auth methods like SASL SCRAM are only enabled if both sasl and ssl libraries are present. Just having cyrus-sasl present at compile-time is sufficient to add support.

This PR adds cyrus-sasl-dev to the dependencies for librdkafka, and adds an additional test to confirm the SCRAM-SHA-512 sasl mechanism is available, and some others.

### DIFF
--- a/librdkafka.yaml
+++ b/librdkafka.yaml
@@ -78,6 +78,8 @@ update:
     - ^ar_test_.*
     - ^test-.*
     - -DEV.$
+    - -dev$
+    - -RC\d+$
   git:
     strip-prefix: v
 
--- a/librdkafka.yaml
+++ b/librdkafka.yaml
@@ -20,6 +20,7 @@ environment:
       - ca-certificates-bundle
       - cmake
       - curl-dev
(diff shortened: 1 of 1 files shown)


xnox:
Hi!

Thank you for your contribution. This PR looks good. We will mirror it internally, complete review and CI, and then it will be mirrored back here.

I will let you know once this is done.

Thank you!

ohashmi1:
Hello @xnox is there an update on this PR? When can we expect it to be merged?

ohashmi1:
@xnox any update?

alexnovak:
Hey @xnox! Sorry to bother you, but is there anything I can do to help carry this through? We've built our own apk on our end, but I'd love to cover this path for those who come after.

PR #78725, opened 2026-07-21
Ruby-3.4: Backport fiber use-after-free bugfix (Ruby #22196)

jontutcher:
Ruby 3.4.10 introduced a use-after-free in `fiber_switch` (a side effect of the [Bug #21955] backport in `148f263a`, which widened when a terminated fiber's stack is released). Under fiber-heavy workloads — e.g. the `async` gem — this causes intermittent segfaults. This patch has been requested by a customer, for whom reverting to 3.4.9 is undesirable because of CVEs in that version.

Related:
* https://bugs.ruby-lang.org/issues/22196 (upstream Ruby bug)
* https://github.com/ruby/ruby/pull/17957

The patch is the upstream commit verbatim (`b92c8c2`, applies cleanly to
v3_4_10) and includes upstream's regression test in `test/ruby/test_fiber.rb`. It should be removed when 3.4.11 is released. 

### Pre-review Checklist

- [x] Epoch bumped (1 → 2)
- Reported by a customer hitting this in production on `ruby:3.4.10-dev`

```
<!--ci-cve-scan:fail-any-->
```

#### For version bump PRs

#### For PRs that add patches


jontutcher:
For the record: wolfi-dev/os now receives changes only via automated export from Chainguard's internal package monorepo, so PRs here can't be merged directly. This change has been re-raised internally and will appear here via the regular sync once merged. (Fix for the Ruby 3.4.10 fiber use-after-free regression, https://bugs.ruby-lang.org/issues/22196.)

PR #78742, opened 2026-09-15
fix(python-3.11): require glibc>=2.44 at runtime

tomho-confluent:
## What

Add an explicit `glibc>=2.44` runtime dependency to the `python-3.11-base` subpackage.

## Why

Since the glibc toolchain transition (#285516, *"python-3.11: transition to newest version of glibc"*, epoch bump 2→3, `glibc-2.43-dev`→`glibc-2.44-dev`), `python-3.11` C extension modules are built against glibc 2.44 and reference `GLIBC_2.44` versioned symbols.

But `python-3.11-base` only carries the SCA-generated, **unversioned** `so:libc.so.6` runtime dependency. That is satisfied by *any* glibc, so on an image that already has an older glibc installed (e.g. `glibc-2.43`, still the newest in the `glibc` virtual and what many base images ship), `apk add python-3.11` **installs cleanly but fails at runtime**:

```
ImportError: /usr/lib/libm.so.6: version `GLIBC_2.44' not found
 (required by /usr/lib/python3.11/lib-dynload/math.cpython-311-aarch64-linux-gnu.so)
```

### Evidence — verified from the published aarch64 `.apk`s

Extracted `math.cpython-311-*.so` from each epoch and read its `GLIBC_*` symbol version references:

| `python-3.11-base` | built against | max `GLIBC_*` referenced |
|---|---|---|
| `3.11.16-r2` (last pre-transition) | glibc-2.43 | `GLIBC_2.38` |
| `3.11.16-r3` (first post-transition) | glibc-2.44 | **`GLIBC_2.44`** |
| `3.11.16-r6`, `-r7` | glibc-2.44 | **`GLIBC_2.44`** |

So `r2` resolves fine against a 2.43 base image; `r3` onward silently requires 2.44. Meanwhile the newest `glibc` in the virtual is still `2.43-r13`, while `glibc-2.44` (up to `2.44-r6`) exists and *provides* `so-ver:libc.so.6=2.44-r6` — apk just never gets told to prefer it.

### Reproduction

Any Wolfi-based image whose glibc is older than 2.44 that then does `apk add python3=~3.11` (or `python-3.11`), e.g. `python3 -c "import math"`.

## The fix

`python-3.11-base` genuinely **requires** glibc 2.44 (has since epoch 3) — this change just makes that requirement explicit so apk resolves it. It is not a version pin beyond what the compiled extensions already mandate.

## Scope / root cause

The **underlying gap** is that the SCA does not emit a *versioned* libc requirement (`so-ver:libc.so.6>=…`) for binaries referencing versioned glibc symbols — there are currently **zero** packages in the index depending on `so-ver:libc.so.6`, only the unversioned `so:libc.so.6`. The same latent break therefore likely affects any other package migrated to the `glibc-2.44` stream (the sibling interpreters included). A general fix in melange's SCA would cover all of them; I've scoped this PR to the one package with a concrete reproduction and flagged the rest here for maintainer judgement.

## Testing

I could not build/verify locally (no `melange`/`wolfictl` in my environment) — relying on repo CI to build `3.11.16-r8` and confirm the `glibc>=2.44` dependency resolves. Please treat as draft until CI is green.


tomho-confluent:
Closing in favor of #78743, which describes the root cause. The underlying gap is that SCA emits only the unversioned `so:libc.so.6` for binaries that reference versioned glibc symbols, so the systemic fix (emit `so-ver:libc.so.6>=X`) is preferable to a per-recipe `glibc>=2.44` pin. Happy to help with either direction there.

